PR #21982, opened 2022-09-09
feat(certificatemanager): deprecate DnsValidatedCertificate

corymhall:
Now that the official CloudFormation resource `AWS::CertificateManager::Certificate` (CDK's `Certificate` construct) supports DNS validation we do not want to recommend using the `DnsValidatedCertificate` construct.

The `DnsValidatedCertificate` construct uses CloudFormation custom resources to perform the certificate creation and this creates a lot of maintenance burden on our team (see the list of linked issues). Currently the primary use case for using `DnsValidatedCertificate` over `Certificate` is for cross region use cases. For this use case I have updated the README to have our suggested solution.

The example in the README is tested in this [integration test](https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/aws-cloudfront/test/integ.cloudfront-cross-region-cert.ts)

fixes #8934, #2914, #20698, #17349, #15217, #14519

----

### All Submissions:

* [ ] Have you followed the guidelines in our [Contributing guide?](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md)

### Adding new Unconventional Dependencies:

* [ ] This PR adds new unconventional dependencies following the process described [here](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md/#adding-new-unconventional-dependencies)

### New Features

* [ ] Have you added the new feature to an [integration test](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md)?
	* [ ] Did you use `yarn integ` to deploy the infrastructure and generate the snapshot (i.e. `yarn integ` without `--dry-run`)?

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #24543, opened 2023-03-09
revert(certificatemanager): deprecate DnsValidatedCertificate

stoyan-scava:
Undo deprecate DnsValidatedCertificate

DnsValidatedCertificate is the only construct that have automatically cleanup of
DNS validation records when the Certificate is deleted.

closes https://github.com/aws/aws-cdk/discussions/23952
closes https://github.com/aws/aws-cdk/discussions/23931

revert https://github.com/aws/aws-cdk/pull/21982


corymhall:
@stoyan-scava I've commented on this issue before, but hopefully I can explain our position here. Typically in the `aws-cdk` Custom Resources are used as glue code or to fill in minor gaps in CloudFormation functionality. In rare cases we use Custom Resources to actually create entire resources that don't have official CloudFormation resource coverage.

In the case of the DnsValidatedCertificate, it was created because the official `Certificate` resource did not support DNS validation. Over time this resource grew in functionality until it became clear that the expectation from users was that it would be a full fledged CloudFormation resource, but implemented as a Custom Resource. This is not a use case that we want to support in `aws-cdk`. There is already an official CloudFormation `Certificate` resource that now supports DNS validation. Any limitations in that resource should be raised with CloudFormation and should be added to the official resource. We do not want to continue to add new functionality to `DnsValidatedCertificate`. We could revert the deprecation, but then it would be deprecated in reality if not in name.

> DnsValidatedCertificate is the only construct that have automatically cleanup of
DNS validation records when the Certificate is deleted.

This should be something that the official `Certificate` resource [should handle](https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/837).

vincent-dm:
But the whole point of CDK is making CloudFormation usable! Pointing to other departments is really unhelpful, and it would be a much better customer experience if the CDK feature was deprecated only after CloudFormation was enhanced.

dendle:
@corymhall Please can we entertain making a specific construct that allows us to create certificates in us-east-1, to work around the glaring issue of Cloudfront requiring certificates to be hosted there, *before* we deprecate the only alternative?  Creating an entire stack for a certificate to exist in another region is nuts.

corymhall:
> Creating an entire stack for a certificate to exist in another region is nuts.

I am 100% onboard with doing something similar to how we handle `Lambda@Edge` where the cdk handles creating the stack in `us-east-1` for you. We've talked about doing something in core to handle this scenario so that we don't have to reimplement it for every use case, but haven't had the time to work on it. If someone wants to work on a PR to add it to certificatemanager, we would review it.

> CDK feature was deprecated only after CloudFormation was enhanced.

Deprecation in the `aws-cdk` means two things:
1. We will no longer add features and enhancements to the deprecated API.
2. The deprecated API will be removed in the next major release.

\#1 is the reason we deprecated `DnsValidatedCertificate`. As I explained, maintaining this construct requires us to keep feature parity (+ additional features) with the official `Certificate` construct. This is not the type of thing we want to support in core.

For \#2 we currently do not have any plans for a `v3`.

vincent-dm:
> For #2 we currently do not have any plans for a v3.

Ok, that's good to know. I didn't realize that with major release, you meant the number before the dot. I guess we are safe for a while indeed.

PR #24544, opened 2023-03-09
fix(pipelines): Ubuntu 5 images will be slow, move to Ubuntu 6

rix0rrr:
CodeBuild has moved the `STANDARD_5` images to the slow path, meaning they will not be cached on the host anymore. Every customer using CDK Pipelines has gotten an email about this.

Move the CDK Pipelines default image to `STANDARD_6`.

----


Naumel (approved):
10 matches of STANDARD_5_0
5 matches of STANDARD_6_0
13 matches of CDKP_DEFAULT_CODEBUILD_IMAGE
All looks good from my side √

Added the `do-not-merge` label in case you'd like another pair of eyes, since the number of changes files is not trivial.

philberesford-bjss:
FWIW, [AWS-CDK 2.69.0](https://github.com/aws/aws-cdk/releases/tag/v2.69.0) had breaking changes (incorrectly reported as bug-fix) since the new default of [Ubuntu 6 does not support the Python 3.9](https://docs.aws.amazon.com/codebuild/latest/userguide/available-runtimes.html) runtime.
Recording here for other engineers who come a cropper.

PR #24546, opened 2023-03-09
fix(redshift): Column ids were not being default assigned

### LINKED ISSUE #24545: redshift: no default id being assigned
### Describe the bug

Upon the recent release, column ids were introduced, however the default value of the `id` is not set.
The `id` should be set to the `name` property, however this is not the case. The `cdk diff` does not show this.

### Expected Behavior

Upon not setting an id in the `Table` construct.
```
[ ]   {
[ ]     "name": "col3",
[ ]     "dataType": "float",
[ ]     "comment": "A test column",
[-]     "encoding": "RAW"
[+]     "encoding": "RAW",
[+]     "id": "col3"
[ ]   }
```

### Current Behavior

```
[ ]   {
[ ]     "name": "col3",
[ ]     "dataType": "float",
[ ]     "comment": "A test column",
[-]     "encoding": "RAW"
[+]     "encoding": "RAW",
[ ]   }
```

### Reproduction Steps

Create a new table, without setting the `id` property of that table.
```typescript
new redshift.Table(stack, 'Table', {
  ...databaseOptions,
  tableColumns: [
    { name: 'col1', dataType: 'varchar(4)' },
    { name: 'col2', dataType: 'float' },
    { name: 'col3', dataType: 'float' },
  ],
});
```

### Possible Solution

Update the `addColumnIds` function to include the necessary functionality, which is a side effect to append on a column id if not present.

This should have been implemented in the recent update, however this was overlooked.

### CDK CLI Version

2.68.0

### Framework Version

_No response_

### Node.js Version

v16.16.0

### OS

Amazon Linux 2

### Language

Typescript, Python

### Language Version

TypeScript (3.7.5) | 3.7.16

### Other information

The PR that should have contained the fix is included here:
https://github.com/aws/aws-cdk/pull/24272

Rizxcviii:
The side effect to add the column ids, if not present, was not being performed. This functionality was supposed to be added, but was overlooked.

Closes #24545.

----

### DIFF
--- a/packages/aws-cdk-lib/.npmignore
+++ b/packages/aws-cdk-lib/.npmignore
@@ -5,6 +5,7 @@
 !*.d.ts
 !*.js
 test/
+**/test/**
 
 # Coverage
 coverage
--- a/packages/aws-cdk-lib/aws-s3/lib/bucket.ts
+++ b/packages/aws-cdk-lib/aws-s3/lib/bucket.ts
@@ -1983,7 +1983,22 @@ export class Bucket extends BucketBase {
 
   /**
    * Set up key properties and return the Bucket encryption property from the
-   * user's configuration.
+   * user's configuration, according to the following table:
+   *
+   * | props.encryption | props.encryptionKey | props.bucketKeyEnabled | bucketEncryption (return value) | encryptionKey (return value) |
+   * |------------------|---------------------|------------------------|---------------------------------|------------------------------|
+   * | undefined        | undefined           | e                      | undefined                       | undefined                    |
+   * | UNENCRYPTED      | undefined           | false                  | undefined                       | undefined                    |
+   * | undefined        | k                   | e                      | SSE-KMS, bucketKeyEnabled = e   | k                            |
+   * | KMS              | k                   | e                      | SSE-KMS, bucketKeyEnabled = e   | k                            |
+   * | KMS              | undefined           | e                      | SSE-KMS, bucketKeyEnabled = e   | new key                      |
+   * | KMS_MANAGED      | undefined           | e                      | SSE-KMS, bucketKeyEnabled = e   | undefined                    |
+   * | S3_MANAGED       | undefined           | false                  | SSE-S3                          | undefined                    |
+   * | UNENCRYPTED      | undefined           | true                   | ERROR!                          | ERROR!                       |
+   * | UNENCRYPTED      | k                   | e                      | ERROR!                          | ERROR!                       |
+   * | KMS_MANAGED      | k                   | e                      | ERROR!                          | ERROR!                       |
+   * | S3_MANAGED       | undefined           | true                   | ERROR!                          | ERROR!                       |
+   * | S3_MANAGED       | k                   | e                      | ERROR!                          | ERROR!                       |
    */
   private parseEncryption(props: BucketProps): {
     bucketEncryption?: CfnBucket.BucketEncryptionProperty,


packages/@aws-cdk/aws-redshift/test/table.test.ts
```diff
@@ -141,18 +141,39 @@ describe('cluster table', () => {
...
-  it('throws if column ids are similar', async () => {
-    const updatedTableColumns: redshift.Column[] = [
-      { id: 'col1', name: 'col1', dataType: 'varchar(4)' },
-      { id: 'col1', name: 'col2', dataType: 'float' },
-    ];
-
-    expect(
-      () => new redshift.Table(stack, 'Table', {
+  describe('columnId', () => {
+    it('throws if column ids are similar', async () => {
```
comcalvi, line 145:
```suggestion
    it('throws if column ids are not unique', async () => {
```

packages/@aws-cdk/aws-redshift/lib/table.ts
```diff
@@ -329,12 +330,22 @@ export class Table extends TableBase {
...
       if (column.id) {
         if (columnIds.has(column.id)) {
           throw new Error(`Column id '${column.id}' is not unique.`);
         }
         columnIds.add(column.id);
+      } else {
+        if (columnIds.has(column.name)) {
+          throw new Error(`Column name '${column.name}' is not unique amongst the column ids.`);
+        }
+        this.tableColumns[i] = {
```
comcalvi, line 344: This implies a strict ordering of code, namely that `addColumnIds` must be called after `this.tableColumns = props.tableColumns`. This is not an obvious contract, and one I'd prefer to avoid.

Could you move the `this.tableColumns = props.tableColumns` line to the top of this method?

comcalvi (changes requested):

packages/@aws-cdk/aws-redshift/lib/table.ts
Rizxcviii: Unfortunately, what you have mentioned was not possible, as `this.tableColumns` was meant to be initialised within either the constructor or the declaration. Therefore, I altered an alternative solution, for the `addColumnIds` function to return a new `Column[]`. This will be the same array, but with `ids` attached.

packages/@aws-cdk/aws-redshift/test/table.test.ts
Rizxcviii: This has been done

Rizxcviii:
Is it possible to get this reviewed?

packages/@aws-cdk/aws-redshift/test/table.test.ts
```diff
@@ -141,18 +141,39 @@ describe('cluster table', () => {
...
-      () => new redshift.Table(stack, 'Table', {
+  describe('columnId', () => {
+    it('throws if column ids are not unique', async () => {
+      const updatedTableColumns: redshift.Column[] = [
+        { id: 'col1', name: 'col1', dataType: 'varchar(4)' },
+        { id: 'col1', name: 'col2', dataType: 'float' },
+      ];
+
+      expect(
+        () => new redshift.Table(stack, 'Table', {
```
comcalvi, line 152:
```suggestion
      expect(() => new redshift.Table(stack, 'Table', {
```

packages/@aws-cdk/aws-redshift/lib/table.ts
comcalvi: having this "create and then immediately modify" pattern is confusing. Can you change `addColumnIds` to be `configureTableColumns` and then have `this.tableColumns = configureTableColumns(props.table.Columns)` ? This way, `configureTableColumns` will not modify any state (no setting of `this.tableColumns` in the method) and instead return the correct state.

comcalvi (changes requested):
Sorry for the delay on this review

packages/@aws-cdk/aws-redshift/lib/table.ts
Rizxcviii: @comcalvi This has been done

comcalvi (dismissed):
lgtm, nice work

TheRealAmazonKendra (approved):

PR #24549, opened 2023-03-09
fix(iam): roleName not validated in fromRoleName function

### LINKED ISSUE #24503: aws_iam: iam.Role.from_role_name should fail but returns weird info
### Describe the bug

If I accidentally from `from_role_name` using an role ARN like the code below.  

```python

role_arn = "arn:aws:iam::***:role/myrole"

role = iam.Role.from_role_name(
            scope=self,
            id="Role",
            role_name=role_arn,
        )
print(role.role_arn)
```

The function does not report any error and returns a valid role object with wrong ARN. The snippet will print

`arn:aws:iam::***:role/arn:aws:iam::***:role/myrole`

which of course is a invalid arn and cannot be used anywhere.

### Expected Behavior

Either

throw an exception such as "Role not found" or "Invalid Role Name"

OR

Return the correct role as it does, but with its attributes having valid values.

### Current Behavior

The function is returning an object with wrong ARN and not failing as expected.

### Reproduction Steps

Just include the above code in a CDK stack.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.67.0 (build b6f7f39)

### Framework Version

_No response_

### Node.js Version

v14.21.3

### OS

Amazon Linux 2

### Language

Python

### Language Version

Python 3.9.13

### Other information

_No response_

lpizzinidev:
The `validateRoleName` function validates the `roleName` parameter passed to the `fromRoleName` function following the pattern specified in the [docs](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateRole.html).

Closes #24503.

**Note**
I added the `IAM_IMPORTED_ROLE_STACK_SAFE_DEFAULT_POLICY_NAME` feature flag check because [this](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-iam/test/integ.imported-role.ts) integration test failed, generating a `roleName` in the format `${Token[TOKEN.26]}`.
I don't know if this is the expected behavior, a bug, or an error in my implementation.

----

### DIFF
--- a/CHANGELOG.v2.alpha.md
+++ b/CHANGELOG.v2.alpha.md
@@ -2,6 +2,25 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.72.0-alpha.0](https://github.com/aws/aws-cdk/compare/v2.71.0-alpha.0...v2.72.0-alpha.0) (2023-03-29)
+
+
+### ⚠ BREAKING CHANGES TO EXPERIMENTAL FEATURES
+
+* **servicecatalogappregistry:** This commit involves share replacement during the deployment of `ApplicationAssociator` due to share construct id update. After this change, frequent share replacements due to structural change in `Application` construct should be avoided. `Application.shareApplication` starts to require construct id (first argument) and share name (added in `ShareOption`) as input.
+* **ivs:** Renamed ChannelProps.name to ChannelProps.channelName
+* Renamed PlaybackKeyPairProps.name to PlaybackKeyPairProps.playbackKeyPairName
+* Channel now generates a physical name if one is not provided
+* PlaybackKeyPair now generates a physical name if one is not provided
+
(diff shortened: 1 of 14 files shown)


lpizzinidev:
> 
> ```
> ```
> 
> 

Exemption Request.
This PR introduces validation on a parameter and will throw an exception in case of a wrong pattern so a unit test should be enough to check the behavior.

TheRealAmazonKendra (changes requested):

So, yes, the failing integ test is a matter of an error in implementation. The feature flag shouldn't be needed but you will need to add handling for the value being a token. 

The best you can do here is to only validate when it isn't a token so I would add in an if statement checking if `roleName` is a token and only calling `validateRoleName` if it is not. Check out the `Token` class for our functions for checking if something is a token or not, you can use those here.

TheRealAmazonKendra:
Granted integ test exception.

TheRealAmazonKendra (approved):

PR #24550, opened 2023-03-09
feat(remodel): adds remodel package for repo restructure

MrArnoldPalmer:
Adds a package, @aws-cdk/remodel, that is a CLI tool that will restructure the packages within the repo to remove the v1 modules, and collapse them all into aws-cdk-lib. The aim of this is to make working on, building, testing, etc of the aws-cdk easier and faster.

The remodel tool does the following:
1. Copies the current aws-cdk contents into a new directory that it will work within
2. Runs ubergen to bundle all of the @aws-cdk/* service packages into aws-cdk-lib and rewrite their imports to be relative where necessary.
3. Copies all integration tests and snapshots into a new @aws-cdk-testing/framework-testing package
4. Runs ./transform.sh to generate the alpha packages within packages/individual-packages then moves them back into the @aws-cdk/* folder
5. Does general cleanup of leftover packages and reformat test files, build tooling, etc to support building the repository in its new structure.

### DIFF
--- a/tools/@aws-cdk/remodel/lib/template/buildspec-pr.yaml
+++ b/tools/@aws-cdk/remodel/lib/template/buildspec-pr.yaml
@@ -1,0 +1,31 @@
+version: 0.2
+
+# This buildspec is intended to be used by GitHub PR builds.
+
+phases:
+  install:
+    commands:
+
+      # baked in our image.
+      # this also takes care of launching the docker daemon.
+      - /root/ecr-proxy/start.sh
+
+      # CodeBuild always runs as root, allow npm to operate as such
+      - npm config set unsafe-perm true
+
+      # Install yarn if it wasn't already present in the image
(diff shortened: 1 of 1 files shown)


rix0rrr:
Scratch pad for notes after looking at https://github.com/aws/aws-cdk/tree/feat/remodel:

I see that a bunch of code I would have expected to see disapper is still in there:

* ubergen
* individual-packages
* individual-pkg-gen
* cdk-dasm (not strictly part of this task but if we're cleaning up anyway...)
* shouldn't `remodel` delete itself after running? :) 

I would have expected to see alpha packages under `@aws-cdk`, but I'm not seeing any. Am I looking at the right thing?

MrArnoldPalmer:
Closing this as not needed as the actual remodel PR was merged https://github.com/aws/aws-cdk/pull/24376

PR #24217, opened 2023-02-17
chore: build.sh script unnecessarily requires lerna to be globally installed

timmattison:
Either lerna needs to be installed as a dependency or run through npx. Otherwise build.sh fails in the middle of a build on a system without it installed.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #24556, opened 2023-03-09
chore: build.sh script requires lerna to be globally installed

madeline-k:
While setting up a new laptop, I realized that our build scripts require a global installation of `lerna`, but the contributing guide does not actually tell you to install it. 

The `build.sh` script runs the `lerna` command directly, which only works if `lerna` is globally installed. `lerna` is a dependency in the monorepo, which means after a `yarn install`, `lerna` is available in the monorepo, and a global install is not necessary. This PR changes build.sh to run `npx lerna` so that a global install is not required to run the build.sh script. 

----

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -80,7 +80,7 @@ fi
 
 echo "============================================================================================="
 echo "building..."
-time lerna run $bail --stream $runtarget || fail
+time npx lerna run $bail --stream $runtarget || fail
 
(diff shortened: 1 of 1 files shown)


madeline-k:
Closing since this was already done in https://github.com/aws/aws-cdk/pull/24217! Thanks @timmattison!

PR #24557, opened 2023-03-09

laverdet:

```
Resources
 └─ [~] Description
     ├─ [-] ?????
     └─ [+] 🤦🏻‍♂️
```

### DIFF
--- a/packages/@aws-cdk/cloudformation-diff/test/deep-equal.test.ts
+++ b/packages/@aws-cdk/cloudformation-diff/test/deep-equal.test.ts
@@ -1,0 +1,11 @@
+import { deepEqual } from '../lib/diff/util';
+
+test('mangled strings', () => {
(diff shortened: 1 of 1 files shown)


laverdet:
Exemption Request

TheRealAmazonKendra:
> Reference

You must provide a justification for the exemption you're requesting. Please let us know why you'd like this exemption.

laverdet:
@TheRealAmazonKendra thank you for the response. This change only affects the CLI output and therefore an integration test isn't possible. I'm happy to spend more time documenting or working through other prerequisites but I do want to make sure that, conceptually, this change is one that the team will accept. I have also included more information about the benefits and tradeoffs of this change in the original PR message.

laverdet:
@TheRealAmazonKendra could we reopen this? It is not abandoned and I am waiting on feedback from the team.

PR #24565, opened 2023-03-10

### LINKED ISSUE #24548: (module name): (short issue description)
### Describe the bug

`AppSync` now supports `JavaScript` resolvers which in `CDK` are can be populated trough the `code` prop which can either be `fromInline` or `fromAsset`.

The problem is that it seems `AppSync` can only take `.js` and will throw a `The code contains one or more errors.` exception when you use a custom bundling environment, since this emits a `.zip`.

### Expected Behavior

`AppSync` digests single file archives generated when using a custom `Docker` environment to bundle `JavaScript` `AppSync` resolvers.

### Current Behavior

`AppSync` throws a 'The code contains one or more errors.' exception when using a custom `Docker` environment to bundle `JavaScript` `AppSync` resolvers.

### Reproduction Steps

```ts
new Resolver(this, "TestResolver", {
  api: new GraphqlApi(this, "TestApi", {
    name: "TestApi",
    schema: SchemaFile.fromAsset(path.join(__dirname, "./schema.graphql")),
  }),
  typeName: "Query",
  fieldName: "test",
  runtime: FunctionRuntime.JS_1_0_0,
  code: Code.fromAsset(path.join(__dirname, "./"), {
    bundling: {
      image: DockerImage.fromBuild(path.join(__dirname, "./")),
      command: [
        "esbuild",
        "/asset-input/resolver.ts",
        "--target=node18",
        "--outfile=/asset-output/index.js",
      ],
    },
  }),
});
```

### Possible Solution

I see two possible solution for this problem:

1. Allow single file outputs from bundling with custom environment rather then only archives
2. Somehow make AppSync phrase the archive edited from a custom environment bundle

_No response_

### CDK CLI Version

2.67.0

### Framework Version

_No response_

### Node.js Version

18

### OS

aws/codebuild/amazonlinux2-x86_64-standard:4.0

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

p-mercury:
Closes #24548

I added and implemented `BundlingOutput.FILE` type to allow bundled assets to produce a single file.
I also wrote the unit test for this new feature.

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
@@ -20,7 +20,12 @@ describe('ci', () => {
       const execOptions = {
         captureStderr: true,
         onlyStderr: true,
-        modEnv: { CI: 'true' },
+        modEnv: {
+          CI: 'true',
+          JSII_SILENCE_WARNING_KNOWN_BROKEN_NODE_VERSION: 'true',
+          JSII_SILENCE_WARNING_UNTESTED_NODE_VERSION: 'true',
+          JSII_SILENCE_WARNING_DEPRECATED_NODE_VERSION: 'true',
+        },
       };
 
       const deployOutput = await fixture.cdkDeploy('test-2', execOptions);


p-mercury:
`Exemption Request` I added the new `BundlingOutput.FILE` feature to the `integ.js-resolver.ts` and the intention is that if it works correctly it should behave the same as a `Code.fromAsset` without bundling, so the snapshot not changing means it is working as intended.

TheRealAmazonKendra:
> `Exemption Request` I added the new `BundlingOutput.FILE` feature to the `integ.js-resolver.ts` and the intention is that if it works correctly it should behave the same as a `Code.fromAsset` without bundling, so the snapshot not changing means it is working as intended.

If the test is run, we see a lot more changes than what's showing up so I'm concerned that this did not actually run correctly based off just that change. Please create a new integ test for this to ensure that the asset deploys and is handled correctly.

TheRealAmazonKendra (changes requested):
> `Exemption Request` I added the new `BundlingOutput.FILE` feature to the `integ.js-resolver.ts` and the intention is that if it works correctly it should behave the same as a `Code.fromAsset` without bundling, so the snapshot not changing means it is working as intended.


If the test is run, we see a lot more changes than what's showing up so I'm concerned that this did not actually run correctly based off just that change. Please create a new integ test for this to ensure that the asset deploys and is handled correctly.

PR #24213, opened 2023-02-17
feat(eks): add helm flag --skip-crds

alexandersperling:
When installing Helm chart sometimes it is needed to skip the installation of custom resource definitions.
This MR adds the functionality to the helm chart construct.

In fact I did the following

* Add the possibility to skip the installation of crds in a helm chart via a new property
* Add tests for skipCrds property
* Update README

Closes #24296 

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #24583, opened 2023-03-12
feat(eks): Add atomic flag for helm construct

### LINKED ISSUE #22254: aws-eks: support `--atomic` flag for helm commands
### Describe the feature

Support the Helm CLI `--atomic` flag  as a prop to the aws-cdk.aws-eks.HelmChart construct. This would enable a user to perform their install/upgrade as an atomic operation, rolling back or uninstalling if the operation fails.

### Use Case

Currently, if the resources installed by helm do not successfully initialize and the `--wait` flag is true, then helm will fail to send a response back to the custom resource in time. Once the first invocation fails, subsequent attempts will fail with the following error:

```console
Error: UPGRADE FAILED: another operation (install/upgrade/rollback) is in progress
```
After the max attempts, the CloudFormation stack will be marked as `UPDATE_ROLLBACK_FAILED`. Forcing the user to manually rollback the helm release as described in [this issue on the helm repo](https://github.com/helm/helm/issues/7476)

### Proposed Solution

This could be supported with an additional, optional ResourceProperty passed to the custom resource handler and used in the constructed helm command.

https://github.com/aws/aws-cdk/blob/877ad7121b31fce49ac093695555bc637e3f010d/packages/%40aws-cdk/aws-eks/lib/kubectl-handler/helm/__init__.py#L149

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.28.1

MacOS, TypeScript

alexandersperling:
This PR adds the possibilty to use the atomic flag for helm charts.

Basically I took code from an PR which was closed sometime ago, refactored it and made changes for an integration test.

* Add atomic flag for helm chart construct
* Update unit test
* Update README
* Update integ test

Closes #22254

----


alexandersperling:
I tried running integration test, but got an error while deploying, which relates to the skip-crd helm chart.

I did not run the integration test updates when adding the skip-crd flag, so I'm not sure how they passed the last time.

Could someone take a look into this, rerun the integration tests and let me know, if any changes are needed for this? 

```
Failed resources:
aws-cdk-eks-helm-test | 18:35:10 | CREATE_FAILED        | Custom::AWSCDK-EKS-HelmChart          | Cluster/chart-test-skip-crd-installation/Resource/Default (ClustercharttestskipcrdinstallationB8323954) Received response status [FAILED] from custom resource. Message returned: Error: b'Release "lambda-chart-release" does not exist. Installing it now.\nError: release: already exists\n'
```

alexandersperling:
> What do you mean by you didn't run the integration test updates last time? Our integ tests work by deploying the pre-change template and then update it with the changes applied. If you didn't run the update, that test just plain never worked.

Thats what I thought as well.
In #24213 I added the flag for skipping crds and were not able to run integration tests (no bootstraped AWS account).
However I added a case in the integ test file and you were running the tests for me which updated the snapshots. Thats why I said, 'I did not run the integration test updates' and 'I'm not sure how they passed the last time.'.

> Also each test shouldn't contain multiple test cases so these should have each been individual tests anyway.

Okay, so I should better add separate test files for both of these cases? I was just adding them to the `integ.eks-helm-asset.ts` because it already contained more than one case, but if it's the preferred way to have separate files I can do that.

alexandersperling:
Clarification Request

Hi, I've updated the corresponding integ test file, but unfortunately I'm not able to run an integtest and update the snapshot.
Would be great if someone take this over, like described in the CONTRIBUTING.md

jeffb4:
@TheRealAmazonKendra I pulled up the codebuild logs from the last build and was unable to determine what had failed. There were a lot of docker/moby warnings and errors spamming the logs, was that the reason a failure was registered?

PR #24592, opened 2023-03-13
fix(rds): descriptive error for creating proxy for read replica

mskrip:
The bug was that the `ProxyTarget` requires `engine` on the instance it targets but `DatabaseInstanceReadReplica` did not inherit `engine` from its source properly.

The engine property is now passed to the object and creating ProxyTarget from read replica instance fails explicitly since [you can associate a proxy only with the writer DB instance, not a read replica](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy.html).

----


mskrip:
Exemption Request: This fix does not change any deployment behavior, only moves the error condition to a more appropriate place; explicit error instead of a weird attribute error.

ocrlgv:
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy.html
this page must be outdated or something. I successfully (in the UI) create an RDS Proxy to a read replica instance.
So it seems the engine should be exposed.
is there any workaround as I've tried using the `.addProxy` and creating an `rds.DatabaseProxy` but both complain the same message
```
Error: Could not determine engine for proxy target 'Base-au/businessDbReplica'. Please provide it explicitly when importing the resource
```

mskrip:
AWS RDS Console allows creating the proxy with Read replica as target for some reason, but it does not associate the database as its target. When I created the proxy and I check `$ aws rds describe-db-proxy-targets --db-proxy-name replica-proxy` I get output

```
{
    "Targets": []
}
```

Doing this through CloudFormation (by creating read replica object through `rds.DatabaseInstanceReadReplica.from_database_instance_attributes` (which takes in the `engine` attribute by the way) I get an error during deployment saying `DB Instance db-replica-instance is in an unsupported state - Instance should not be a read replica`, i.e. the same one I added in this PR on synth time instead of deploy time.

mskrip:
Can this be reopened? This PR was not stale and I have no further action I can take from my side to fix the failing checks.

PR #24598, opened 2023-03-13
chore: Skip memory config check for windows instance

homakk:
> REPLACE THIS TEXT BLOCK
>
>
> code you submit.
>

Closes #<issue number here>.

----

### DIFF
--- a/packages/@aws-cdk/aws-ecs/lib/base/task-definition.ts
+++ b/packages/@aws-cdk/aws-ecs/lib/base/task-definition.ts
@@ -442,7 +442,7 @@ export class TaskDefinition extends TaskDefinitionBase {
     this.ephemeralStorageGiB = props.ephemeralStorageGiB;
 
     // validate the cpu and memory size for the Windows operation system family.
-    if (props.runtimePlatform?.operatingSystemFamily?._operatingSystemFamily.includes('WINDOWS')) {
+    if (isWindowsOs(props.runtimePlatform?.operatingSystemFamily?._operatingSystemFamily)) {
       // We know that props.cpu and props.memoryMiB are defined because an error would have been thrown previously if they were not.
       // But, typescript is not able to figure this out, so using the `!` operator here to let the type-checker know they are defined.
       this.checkFargateWindowsBasedTasksSize(props.cpu!, props.memoryMiB!, props.runtimePlatform!);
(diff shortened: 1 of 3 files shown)


packages/@aws-cdk/aws-ecs/lib/base/task-definition.ts
```diff
@@ -1206,3 +1206,10 @@ export function isFargateCompatible(compatibility: Compatibility): boolean {
 export function isExternalCompatible(compatibility: Compatibility): boolean {
   return [Compatibility.EXTERNAL].includes(compatibility);
 }
+
+/**
+ * Return true for windows instance
+ */
+export function isWindowsInstance(operatingSystemFamily?: string): boolean | undefined {
+  return operatingSystemFamily?.startsWith('WINDOWS_SERVER_');
```
pallymore, line 1214: thanks for making a helper for this!

looks like line 445 above is also checking for WINDOWS - maybe we should make that line use the same helper also? (`.includes('WINDOWS')` might be more generic 

the name `isWindowsInstance` is slightly odd here as there are no "instances" in the task definition world - I think 
`isWindowsOS` is probably better. 

BTW - what do you think about adding a helper on the `OperatingSystemFamily` class directly? I'm not sure what's the convention here, since both places have `operatingSystemFamily` instances, I'm curious if there is a way to allow us to simply write `runtimePlatform.operatingSystemFamily.isWindows()` in both lines.

packages/@aws-cdk/aws-ecs/test/task-definition.test.ts
```diff
@@ -200,6 +200,36 @@ describe('task definition', () => {
       });
     });
 
+    test('A task definition for non windows instance with no memory config throws an error', () =>{
```
pallymore, line 203: this is a good test. 🎊

may I suggest adding another one with Windows OS here also? so we know the same values for windows are not throwing any errors 😬

pallymore (changes requested):

packages/@aws-cdk/aws-ecs/lib/base/task-definition.ts
homakk: ya changed isWindowsInstance to isWindowsOs.
Added same logic for line 445 as well.

and changed code which can use by Ec2 customers as well where Windows on EC2  only memoryLimitMib is provided,

packages/@aws-cdk/aws-ecs/test/task-definition.test.ts
homakk: Gotcha added new test case with windows OS

PR #24650, opened 2023-03-16
chore(cfnspec): remove Redshift patch

otaviomacedo:
This change removes the following patch from `cfnspec`, which was blocking the build after `VpcEndpoint` was removed from `AWS::Redshift::EndpointAccess`:

```json
"AWS::Redshift::EndpointAccess": {
  "patch": {
  "description": "This patch fixes all types that were previously typed as Json, and CfnSpec v101.0.0 added types to them, which is a breaking change.",
  "operations": [
    {
      "op": "remove",
      "path": "/Properties/VpcEndpoint/Type"
    },
    {
      "op": "add",
      "path": "/Properties/VpcEndpoint/PrimitiveType",
      "value": "Json"
    }
  ]
}
```
The latest spec adds three new modules: `aws-internetmonitor`, `aws-ivschat` and `aws-systemsmanagersap`. All three failed to build because a dependency on `jsii@v4.9-next` was missing. Dependency added.

Also, the property  `aws-events/CfnArchive.attrArchiveName` was removed, breaking the `aws-events/Archive` L2. Fixed by replacing it with `archive.ref`.

The rest of the changes in this PR were auto-generated.

----

### DIFF
--- a/packages/@aws-cdk/aws-lambda-python/README.md
+++ b/packages/@aws-cdk/aws-lambda-python/README.md
@@ -285,3 +285,24 @@ new python.PythonFunction(this, 'function', {
   },
 });
 ```
+
+## Troubleshooting
+
+### Containerfile: no such file or directory
+
+If you are on a Mac, using [Finch](https://github.com/runfinch/finch) instead of Docker, and see an error
+like this:
+
+```txt
+lstat /private/var/folders/zx/d5wln9n10sn0tcj1v9798f1c0000gr/T/jsii-kernel-9VYgrO/node_modules/@aws-cdk/aws-lambda-python-alpha/lib/Containerfile: no such file or directory
+```
+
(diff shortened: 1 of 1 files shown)


rix0rrr:
> The latest spec adds three new modules: aws-internetmonitor, aws-ivschat and aws-systemsmanagersap. All three failed to build because a dependency on jsii@v4.9-next was missing. Dependency added.

Why do packages need a direct dependency on `jsii 4.9`? Doesn't the `jsii` dependency normally come in via `cdk-build-tools` ?

If this is the new normal, did you also add the new config to `create-missing-libraries.ts` ?

otaviomacedo:
> If this is the new normal, did you also add the new config to `create-missing-libraries.ts` ?

This is not the new normal. It will go away soon, but for now we need to add this dependency for the build to succeed.

rix0rrr (approved):

otaviomacedo:
@mergifyio refresh

PR #24652, opened 2023-03-16
fix(bootstrap): missing CloudFormation permissions for DescribeStacks

### LINKED ISSUE #24371: Bootstrap template: missing new IAM permission cloudformation:ListStacks
### Describe the bug

We've received a warning from AWS about IAM roles changes

> On January 25, 2023, we updated our DescribeStacks API as part of our continuous improvements to ensure customers have granular control over the APIs they use.

> We require your action to update IAM Policies in your account by March 27, 2023 to avoid disruption to your DescribeStacks API calls.

> AWS CloudFormation now requires that customers have the cloudformation:ListStacks permission when calling DescribeStacks [1] without a target stack. This is in addition to the cloudformation:DescribeStacks permission that is already a requirement. You will need to add the cloudformation:ListStacks permission by March 27, 2023, on any calling principals using the DescribeStacks API without a target. CloudFormation has identified principals associated with your account that do not have ListStacks permissions making DescribeStacks calls without a target. These users or roles need their policies updated to include cloudformation:ListStacks along with the existing cloudformation:DescribeStacks permissions:

The role listed in the warning was `assumeed-role/cdk-hnb659fds-deploy-role-acct-region`

I've checked template for bootstrapping
https://github.com/aws/aws-cdk/blob/v1-main/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml

It is missing `cloudformation:ListStacks` permission

### Expected Behavior

bootstrap-template.yaml needs cloudformation:ListStacks

### Current Behavior

bootstrap template missing permissions to list stacks granularly 

### Reproduction Steps

It is not breaking yet.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.66.1

### Framework Version

_No response_

### Node.js Version

18x

### OS

osx

### Language

Typescript

### Language Version

_No response_

### Other information

None

danno-s:
**NOTE**: This change introduces bootstrap stack version `17`. You should rebootstrap all your accounts, as this new bootstrap stack versions adds permissions that CloudFormation may need, starting March 27th. We are working to figure out what CDK-related operations this will impact.

This change adds the ListStacks permission to `bootstrap-template.yml`. This is needed because on March 27th, CloudFormation will require DescribeStacks API Calls without defined targets to have this permission in addition to the DescribeStacks permission.

Closes #24371.

----


mrgrain:
Thanks for this @danno-s 
Can you also increase the version counter please?

danno-s:
Yeah of course! I'm on it

rix0rrr:
This change should not be a `chore`. This is pretty relevant from a user's point of view :)

rix0rrr:
Do not merge this yet. We are still figuring out if the CDK actually requires this permission or not.

rix0rrr:
Cross-posting the [comment I made on the linked issue](https://github.com/aws/aws-cdk/issues/24371#issuecomment-1473657755):

----------

So far, we've not been able to determine the cause of these untargeted calls, and while we use CDK ourselves every day, none of our accounts are affected by this new policy.

I would therefore like to ask you to dig into your CloudTrail logs to help us figure out when and how your `cdk-hnb659fds-deploy-role-$ACCT-$REGION` role was involved in making untargeted `DescribeStacks` calls.

Merging the proposed fix would weaken the security posture of all CDK users. Since it seems like CDK doesn't actually need these permissions by default, that's something we'd rather not do if it's not strictly necessary.

rix0rrr:
Original issue has been closed, as it seems the required permissions aren't actually necessary.

PR #24672, opened 2023-03-17
feat(servicecatalogappregistry): add attribute groups to an application

liwewang-amazon:
To associate a attribute group to an application created in `ApplicationAssociator`, customers have to use `AttributeGroup` construct separately to create and associate the attribute group separately. This makes the `AttributeGroup` and `AttributeGroupAssociation` created in another stack than `ApplicationAssociator` stack.
This commits provides an one-stop action, i.e. `Application.addAttributeGroup()`, to create and associate attribute group on `Application` Construct. This solution not only makes attribute group creation and association easier for customer who uses `Application` construct, but also lets customer to have attribute groups and attribute group associations for the `ApplicationAssociator` applications in the same stack.

`Application.addAttributeGroup()` has `id` in the parameters, for two reasons:
- consistent with the experience where customer can define logical ID when using `new AttributeGroup()`
- complexity of deciding logical ID from the attribute group input:
  - We have to make sure update attributes/description won't trigger create and then delete but update, which will cause name conflict exception. 
  - We also don't want to generate logical ID from attribute group name only, as if two `Application.addAttributeGroup()` method calls with the same name will result in construct ID conflict. This exposes implementation details and makes it hard to customers to debug and resolve.

BREAKING CHANGE: This commit contains destructive changes to the RAM Share.
Since the application RAM share name is calculated by the application construct, where one method is added. Integration test detects a breaking change where RAM share will be created. Integration test snapshot is updated to cater this destructive change.

----

### DIFF
--- a/packages/@aws-cdk/aws-servicecatalogappregistry/lib/application.ts
+++ b/packages/@aws-cdk/aws-servicecatalogappregistry/lib/application.ts
@@ -17,7 +17,7 @@ const APPLICATION_ALLOW_ACCESS_RAM_PERMISSION_ARN = 'arn:aws:ram::aws:permission
  */
 export interface AttributeGroupAssociationProps {
   /**
-   * name for attribute group.
+   * Name for attribute group.
    *
    */
   readonly attributeGroupName: string;


packages/@aws-cdk/aws-servicecatalogappregistry/README.md
```diff
@@ -154,7 +138,12 @@ const associatedApp = new appreg.ApplicationAssociator(app, 'AssociatedApplicati
 });
 
 // Associate application to the attribute group.
-customAttributeGroup.attributeGroup.associateWith(associatedApp.appRegistryApplication());
+associatedApp.appRegistryApplication().addAttributeGroup({
```
rix0rrr, line 141: `appRegistryApplication` should probably be a field, not a function? Make it a getter if you need to throw.

packages/@aws-cdk/aws-servicecatalogappregistry/lib/application.ts
```diff
@@ -114,6 +147,23 @@ abstract class ApplicationBase extends cdk.Resource implements IApplication {
     }
   }
 
+  /**
+   * Create an attribute group and associate this application with the created attribute group.
+   */
+  public addAttributeGroup(props: AttributeGroupAssociationProps): IAttributeGroup {
+    const attributeGroup = new AttributeGroup(this, props.attributeGroupLogicalId, {
```
rix0rrr, line 154: This is not a logical ID, it's a construct ID. Please at least make it optional?

```suggestion
    const attributeGroup = new AttributeGroup(this, props.constructId ?? 'Attributes', {
```

packages/@aws-cdk/aws-servicecatalogappregistry/lib/application.ts
```diff
@@ -12,6 +12,32 @@ import { CfnApplication, CfnAttributeGroupAssociation, CfnResourceAssociation }
...
+ */
+export interface AttributeGroupAssociationProps {
+  /**
+   * Logical Id of the attribute group to be created
+   */
+  readonly attributeGroupLogicalId: string;
+  /**
+   * Enforces a particular physical attribute group name.
+   */
+  readonly attributeGroupName: string;
```
rix0rrr, line 26: Why is this not optional?

rix0rrr (dismissed):

rix0rrr:
> BREAKING CHANGE: This commit contains destructive changes to the RAM Share.
Since the application RAM share name is calculated by the application construct, where one method is added. 

I don't fully understand this. Do you think it is reasonable behavior?

packages/@aws-cdk/aws-servicecatalogappregistry/lib/application.ts
liwewang-amazon: Attribute group name is the key to calculate attribute group construct id to make sure that multiple `addAttributeGroup` calls won't generate the same construct ID. Since name is unique, if `addAttributeGroup` is called twice with the same `attributeGroupName`, construct duplicate error will be thrown. Both `Names.uniqueNodeId()` and `Names.uniqueResourceName()` generate id from the path of the construct, i.e. path of the application. As a result,  two `addAttributeGroup` calls with different names\attributes\descriptions generate two `AttributeGroup` constructs with the same construct ID.

packages/@aws-cdk/aws-servicecatalogappregistry/lib/application.ts
liwewang-amazon: Removed it and calculate it from the attribute group name.

packages/@aws-cdk/aws-servicecatalogappregistry/lib/application.ts
liwewang-amazon: This should be required even if we use customer provided construct id for attribute group.

Attribute group name is the key human readable identifier our service offers for customer to find the attribute group, we want to make it explicit, so that it will can make use of the name to look for the attribute group in other platforms like console or SDK client.

rix0rrr (approved):

PR #21533, opened 2022-08-10
fix(ec2): unavailable kernel version for Amazon Linux 2022 fails deployment (under feature flag)

mrgrain:
Amazon Linux 2022 is in preview release

It appears that previous kernel versions are not retained in this phase.
While existing deployments will still work, it is not possible to create/update
instances with an old kernel version.

This change will change the kernel version for all customers to the default
version, unless they have opted into a specific version. Since the AL22 is in
preview, we deem this an acceptable breaking change.

https://docs.aws.amazon.com/linux/al2022/ug/naming-and-versioning.html

Fixes #21011
Replaces #21438

----

### All Submissions:

* [x] Have you followed the guidelines in our [Contributing guide?](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md)

### Adding new Unconventional Dependencies:

* [ ] This PR adds new unconventional dependencies following the process described [here](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md/#adding-new-unconventional-dependencies)

### New Features

* [ ] Have you added the new feature to an [integration test](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md)?
	* [ ] Did you use `yarn integ` to deploy the infrastructure and generate the snapshot (i.e. `yarn integ` without `--dry-run`)?

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

Issue #21011, opened 2022-07-06
@aws-cdk/aws-ec2: Wrong enum declare enum AmazonLinuxKernel


adonisd:
### Describe the bug

the enum `declare enum AmazonLinuxKernel` return 5.10 as latest kernel where it should be 5.15 for `AMAZON_LINUX_2022`
This leads to an error in CDK when trying to deploy:

```
failed: Error [ValidationError]: Unable to fetch parameters [/aws/service/ami-amazon-linux-latest/al2022-ami-kernel-5.10-x86_64] from parameter store for this account.
    at Request.extractError (/home/adonis/work/FOLDER/FOLDER-cicd/CORE/node_modules/node_modules/aws-sdk/lib/protocol/query.js:50:29)
    at Request.callListeners (/home/adonis/work/FOLDER/FOLDER-cicd/CORE/node_modules/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/home/adonis/work/FOLDER/FOLDER-cicd/CORE/node_modules/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/home/adonis/work/FOLDER/FOLDER-cicd/CORE/node_modules/node_modules/aws-sdk/lib/request.js:686:14)
    at Request.transition (/home/adonis/work/FOLDER/FOLDER-cicd/CORE/node_modules/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/home/adonis/work/FOLDER/FOLDER-cicd/CORE/node_modules/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /home/adonis/work/FOLDER/FOLDER-cicd/CORE/node_modules/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/home/adonis/work/FOLDER/FOLDER-cicd/CORE/node_modules/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/home/adonis/work/FOLDER/FOLDER-cicd/CORE/node_modules/node_modules/aws-sdk/lib/request.js:688:12)
    at Request.callListeners (/home/adonis/work/FOLDER/FOLDER-cicd/CORE/node_modules/node_modules/aws-sdk/lib/sequential_executor.js:116:18) {
  code: 'ValidationError',
  time: 2022-07-06T10:59:21.163Z,
[...]

PR #24690, opened 2023-03-19
feat(ec2): Allow for latest Amazon Linux 2023 Image when creating EC2s

sktan:
Amazon Linux 2023 is now generally available, and the SSM parameters exist to be able to reference this via the `ec2.AmazonLinuxGeneration` structs when creating EC2 instances.

https://aws.amazon.com/blogs/aws/amazon-linux-2023-a-cloud-optimized-linux-distribution-with-long-term-support/

This PR adds support for creating EC2 instances with Amazon Linux 2023 AMIs.

I've also added error messages for both Amazon Linux 2022 and Amazon Linux 2023 when specifying a storage optimised image as these do not exist in Parameter Store.

Amazon Linux 2022 and 2023 images were referencing invalid AMI SSM parameters, so I've also fixed that up to resolve to the `kernel-default` images.

----


packages/@aws-cdk/aws-ec2/lib/machine-image.ts
```diff
@@ -402,13 +402,16 @@ export class AmazonLinuxImage extends GenericSSMParameterImage {
     let virtualization: AmazonLinuxVirt | undefined;
     let storage: AmazonLinuxStorage | undefined;
 
-    if (generation === AmazonLinuxGeneration.AMAZON_LINUX_2022) {
-      kernel = AmazonLinuxKernel.KERNEL5_X;
+    if (generation === AmazonLinuxGeneration.AMAZON_LINUX_2022 || generation === AmazonLinuxGeneration.AMAZON_LINUX_2023) {
+      kernel = (props && props.kernel) || AmazonLinuxKernel.DEFAULT;
```
MrArnoldPalmer, line 406: So this seems like a breaking change. For some current users their kernel version goes from 5_X to DEFAULT. Is that intentional and what are the implications of that? What are the implications generally of the SSM parameter name changing? Does it matter?

Additionally I'm not sure if it was intentional that when props.kernel was defined, and AMAZON_LINUX_2022 was the generation, then the kernel version got set to 5_X.

MrArnoldPalmer (commented):
Just some questions

packages/@aws-cdk/aws-ec2/lib/machine-image.ts
corymhall: See this comment for how we want to handle this

https://github.com/aws/aws-cdk/issues/21011#issuecomment-1372170948

MrArnoldPalmer:
So, discussing this with @corymhall, I think we need to implement one of the solutions he talks about in his [comment here](https://github.com/aws/aws-cdk/issues/21011#issuecomment-1372170948). I think there isn't an easy fix implementing `kernel-default` without breaking people currently. Here was a previous attempt: https://github.com/aws/aws-cdk/pull/21533

That being said I'm going to close this for now. For those wishing to use newer generation/kernel versions, specify the SSM parameter name using `ec2.MachineImage.fromSsmParameter` as mentioned in [this comment](https://github.com/aws/aws-cdk/issues/21011#issuecomment-1223795077)

PR #24340, opened 2023-02-26
refactor(apigateway): divide setupPermissions methods to small functions

bun913:
Closes #24339 

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #24693, opened 2023-03-20
refactor(ecs): replace instanceOf into BottleRocketImage.isBottleRocketImage() method

### LINKED ISSUE #24692: (ecs): (replace instnaceOf with isBottleRocketImage)
### Describe the feature

I create `Role.isRole()` method at #24340 (https://github.com/aws/aws-cdk/pull/24340#discussion_r1120780639)

That's is for strict type check for `iam.Role` class.

I'll create `ecs.BottleRocketImage` type check like above.

```ts
export class BottleRocketImage implements ec2.IMachineImage {
  /**
   * Return whether the given object is a BottleRocketImage
   */
  public static isBottleRocketImage(x: any): x is BottleRocketImage {
    return x !== null && typeof x === 'object' && BR_IMAGE_SYMBOL in x;
  }

```

Fix the code that uses instanceOf as follows

https://github.com/aws/aws-cdk/blob/e8158af34eb6402c79edbc171746fb5501775c68/packages/%40aws-cdk/aws-ecs/lib/cluster.ts#L1292

### Use Case

Since instanceOf is not a product to be used for strict type determination, it is better to replace it.

fyi: https://github.com/aws/aws-cdk/pull/20564/files#diff-fe41afc44afd9acb16980754990f93acb596ed37662296575ea2fbe1c5531b50R203-R206

### Proposed Solution

I'll create isBottleRocketImage() method like below.

```ts
export class BottleRocketImage implements ec2.IMachineImage {
  /**
   * Return whether the given object is a BottleRocketImage
   */
  public static isBottleRocketImage(x: any): x is BottleRocketImage {
    return x !== null && typeof x === 'object' && BR_IMAGE_SYMBOL in x;
  }
}

Object.defineProperty(BottleRocketImage.prototype, BR_IMAGE_SYMBOL, {
  value: true,
  enumerable: false,
  writable: false,
});
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.69.0

macOS Monterey

bun913:
Closes #24692.

----


MrArnoldPalmer:
@bun913 what is the use case for this? To make our own code more type safe I can see but what is the use case for the public API?

bun913:
@MrArnoldPalmer 

As you say, I created this PR for my conversion to TypeSafe. I am planning to make small improvements in my spare time, as I have learned that it is not appropriate to use `instansOf` when I am doing another refactoring before.

Please see below conversation.

https://github.com/aws/aws-cdk/pull/24340#discussion_r1120780639

MrArnoldPalmer (dismissed):
Cool, ty @bun913 !

corymhall (approved):
re-approving

PR #24702, opened 2023-03-20
fix(toolkit): RWLock.acquireRead is not re-entrant

RomainMuller:
If multiple threads of the same process attempt to acquire the same reader lock, the a race condition occurs, and the first thread to release the reader lock will release ALL the locks.

Introduce a counter so that each acquire attempt uses a different file name, ensuring that the read lock is reentrant.

----

### DIFF
--- a/packages/aws-cdk/lib/api/util/rwlock.ts
+++ b/packages/aws-cdk/lib/api/util/rwlock.ts
@@ -170,7 +170,7 @@ async function readFileIfExists(filename: string): Promise<string | undefined> {
 let tmpCounter = 0;
 async function writeFileAtomic(filename: string, contents: string): Promise<void> {
   await fs.mkdir(path.dirname(filename), { recursive: true });
-  const tmpFile = `${filename}.${++tmpCounter}`;
+  const tmpFile = `${filename}.${process.pid}_${++tmpCounter}`;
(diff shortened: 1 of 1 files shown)


packages/aws-cdk/lib/api/util/rwlock.ts
```diff
@@ -156,9 +167,10 @@ async function readFileIfExists(filename: string): Promise<string | undefined> {
   }
 }
 
+let tmpCounter = 0;
 async function writeFileAtomic(filename: string, contents: string): Promise<void> {
   await fs.mkdir(path.dirname(filename), { recursive: true });
-  const tmpFile = `${filename}.${process.pid}`;
+  const tmpFile = `${filename}.${++tmpCounter}`;
```
rix0rrr, line 173: Needs both the counter and pid in here together methinks. Otherwise multiple processes might collide on the tmpcounter.

rix0rrr (changes requested):

packages/aws-cdk/lib/api/util/rwlock.ts
RomainMuller: The `filename` already includes the `pid` here... I'm relying on that there.

rix0rrr (approved):

PR #24707, opened 2023-03-20
feat(cloudwatch): added defaultInterval prop to cw-dashboard

pattasai:
This PR adds defaultInterval to cloudwatch dashboard, which allows interval duration in relative time eg. 7 days.

```ts
const dashboard = cw.Dashboard(stack, 'Dash', {
      defaultInterval: cdk.Duration.days(7),
    });
```
Here, the dashboard would show the metrics for the last 7 days.

Closes #<issue number here>.

----


packages/@aws-cdk/aws-cloudwatch/lib/dashboard.ts
```diff
@@ -107,15 +115,19 @@ export class Dashboard extends Resource {
       }
     }
 
+    if (props.start !== undefined && props.defaultInterval !== undefined) {
```
corymhall, line 118:
```suggestion
    if (props.start && props.defaultInterval) {
```

packages/@aws-cdk/aws-cloudwatch/lib/dashboard.ts
```diff
@@ -107,15 +115,19 @@ export class Dashboard extends Resource {
...
     const dashboard = new CfnDashboard(this, 'Resource', {
       dashboardName: this.physicalName,
       dashboardBody: Lazy.string({
         produce: () => {
           const column = new Column(...this.rows);
           column.position(0, 0);
           return Stack.of(this).toJsonString({
-            start: props.start,
-            end: props.end,
+            start: props.defaultInterval !== undefined ? `-${props.defaultInterval?.toIsoString()}` : props.start,
```
corymhall, line 129:
```suggestion
            start: props.defaultInterval ? `-${props.defaultInterval?.toIsoString()}` : props.start,
```

packages/@aws-cdk/aws-cloudwatch/lib/dashboard.ts
```diff
@@ -107,15 +115,19 @@ export class Dashboard extends Resource {
...
+            end: props.defaultInterval !== undefined ? undefined : props.end,
```
corymhall, line 130:
```suggestion
            end: props.defaultInterval ? undefined : props.end,
```

corymhall (dismissed):
@pattasai looks great! Just a couple of cosmetic suggestions.

packages/@aws-cdk/aws-cloudwatch/lib/dashboard.ts
rix0rrr: In some cases it is safe to omit `!== undefined`, and in some cases it leads to bugs (for example, in the case of nullable integers, and occasionally in the case of nullable strings depending on what you want to do with the string).

When you're not *super* experienced yet, it's probably safer to always write `!== undefined`. So I support the current use.

packages/@aws-cdk/aws-cloudwatch/lib/dashboard.ts
rix0rrr: Here it would be safe to leave out, but also it doesn't hurt that much. Readability might be *slightly* better without.

rix0rrr (approved):

Issue #24696, opened 2023-03-20
Use AWS Python CDK to create EKS Cluster and Nodegroup


Bruce-Lu674:
### Describe the bug

Hi, folks

I met a promble when use aws python cdk to create eks cluster. Please find information below:

My local env:
(.venv) [ec2-user@ip-10-0-1-73 python-cdk]$ cdk --version
2.67.0 (build b6f7f39)
(.venv) [ec2-user@ip-10-0-1-73 python-cdk]$ python3 --version
Python 3.7.10
(.venv) [ec2-user@ip-10-0-1-73 python-cdk]$ cat /proc/version 
Linux version 5.10.144-127.601.amzn2.x86_64 (mockbuild@ip-10-0-44-229) (gcc10-gcc (GCC) 10.3.1 20210422 (Red Hat 10.3.1-1), GNU ld version 2.35-21.amzn2.0.1) #1 SMP Thu Sep 29 01:11:59 UTC 2022

Here is the core code:

node_role = iam.Role.from_role_arn(self, 'eks-node-role-arn-lookup', 'arn:aws-cn:iam::xxxxxxxxxxx:role/eks-node-role')

cluster.add_nodegroup_capacity(
    nodegroup_name,
    nodegroup_name=nodegroup_name,
    instance_types=[ec2.InstanceType(instance_type)],
    min_size=1,
    max_size=3,
    capacity_type=capacity_type,
    disk_size=disk_size,
    ami_type=ami_type
	node_role=node_role
)

I manually create the Node Role, and the cdk will deploy successfully, but when i remove the node_role parameter, like these:
cluster.add_nodegroup_capacity(
    nodegroup_name,
    nodegroup_name=nodegroup_name,
    instance_types=[ec2.InstanceType(instance_type)],
    min_size=1,
    max_size=2,
    capacity_type=capacity_type,
    disk_size=disk_size,
    ami_type=ami_type
)

Below error messages will be thrown :

Resource handler returned message: "Following required service principals [ec2.amazonaws.com.cn] were not found in the trust relations
hips of nodeRole arn:aws-cn:iam::4123xxxxxxx:role/eks-cluster-stack-eksgitlabrunnerclusterNodegroupg-1EPH8PW36YZ3A (Service: Eks, Sta
[...]

PR #22589, opened 2022-10-20
fix(region-info): EC2 service principal is incorrect in GovCloud regions

kaizencc:
According to these [docs](https://dca-docs-aws.amazon.com/c2s/latest/userguide/ec2.html), "The AWS service principal for Amazon EC2 is ec2.amazonaws.com, but the former service principal of ec2.c2s.ic.gov is still supported for backward compatibility."

PR #24709, opened 2023-03-20
fix(region-info): EC2 service principal in Chia

pahud:
EC2 service principal in Chia should be `ec2.amazonaws.com.cn` rather than `ec2.amazonaws.com`. 

This feature was previously broke from https://github.com/aws/aws-cdk/pull/22589 and this PR fixes that.

Closes https://github.com/aws/aws-cdk/issues/24696

----

### DIFF
--- a/packages/@aws-cdk/aws-ecs/test/cluster.test.ts
+++ b/packages/@aws-cdk/aws-ecs/test/cluster.test.ts
@@ -151,7 +151,7 @@ describe('cluster', () => {
               Action: 'sts:AssumeRole',
               Effect: 'Allow',
               Principal: {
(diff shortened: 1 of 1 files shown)


pahud:
Exemption Request

pahud:
I can't find the document about this, but per the IAM console from AWS China, the service principal should be `ec2.amazonaws.com.cn`.

image

ItielOlenick:
Still an issue on 2.74.0 (latest python)

djessup:
Ya'll seem to break China like this every few months or so. AWS Partitions are a thing, China and GovCloud are the notable examples. I don't know what the situation is in AWS C2S (aka "spook-cloud"), but probably it's a similar deal. Given how deep-rooted and necessary this concept seems to be, I can't see it going away anytime soon. 

Next time somebody has a bright idea to optimise or get rid of `region-info` somehow - please, just don't.

PR #24718, opened 2023-03-21
chore: detect property type renames to protect against breakage

rix0rrr:
Upstream service teams may rename property types in the resource specification. This is strictly speaking not breaking from the point of view of the spec, because the names of the property types don't appear anywhere in the code a user would normally type (i.e., in a CloudFormation template).

However, CDK generates classes for these types, and so the name *does* matter and changing it is breaking.

To detect these instances, we check that during an upgrade, all old property type names are still present. If not, the reason is probably that they renamed a type.

Note that this is not a 100% guaranteed to catch all scenarios (I'm sure you can think of changes that would be breaking and still pass this check), but it's at least very likely to catch honest mistakes in commonly expected scenarios.

For those interested in how it works:

* During the spec upgrade, we have both the old and the new spec available.
* We iterate over all objects keys in the property types of the old spec, looking like: `{ "PropertyTypes": { "AWS::ElastiCache::ReplicationGroup.LogDeliveryConfigurationRequest": { ... }, ... }` object, and make sure that the keys are also present in the property types of the new spec.

Also change the `copy/paste` operation pair of a previous patch into a `move` operation, so that if the type definition changes in the future we won't accidentally keep it at an old definition.

----


rix0rrr:
Example output:

```
┌───────────────────────────────────────────────────────────────────────────────────────┐
│                                                                                       ▐█
│  PROPERTY TYPES HAVE DISAPPEARED                                                      ▐█
│                                                                                       ▐█
│  Some type names have disappeared from the old specification.                         ▐█
│                                                                                       ▐█
│  This probably indicates that the service team renamed one of the types. We have      ▐█
│  to keep the old type names though: renaming them would constitute a breaking change  ▐█
│  to consumers of the L1 resources.                                                    ▐█
│                                                                                       ▐█
└─▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▟█

See what the renames were, check out this PR locally and add a JSON patch file for these types:

(Example)

{
  "patch": {
    "description": "Undoing upstream property type renames of <SERVICE> because <REASON>",
    "operations": [
      {
        "op": "move",
        "from": "/PropertyTypes/AWS::ElastiCache::ReplicationGroup.<NEW_TYPE_NAME_HERE>",
        "path": "/PropertyTypes/AWS::ElastiCache::ReplicationGroup.LogDeliveryConfigurationRequest"
      }
    ]
  }
}             
```

rix0rrr:
For those unhappy with the look of the box drawing characters on GitHub 🥹, this is what it looks like in an actual terminal:

image

PR #24732, opened 2023-03-21
docs(elasticloadbalancing): update readme on internetFacing

pattasai:
This PR adds the documentation in the readme about `internetFacing` on classic load balancer . 

Closes #<issue number here>.

----


packages/@aws-cdk/aws-elasticloadbalancing/README.md
```diff
@@ -18,7 +18,9 @@ classic load balancers.
 
 Load balancers send traffic to one or more AutoScalingGroups. Create a load
 balancer, set up listeners and a health check, and supply the fleet(s) you want
-to load balance to in the `targets` property.
+to load balance to in the `targets` property. We also need to set `internetFacing` as `true` on
```
TheRealAmazonKendra, line 21: Can this be phrased without `we` here? this doesn't quite match the tone of the readme.

TheRealAmazonKendra (changes requested):

packages/aws-cdk-lib/aws-elasticloadbalancing/README.md
```diff
@@ -8,7 +8,9 @@ classic load balancers.
 
 Load balancers send traffic to one or more AutoScalingGroups. Create a load
 balancer, set up listeners and a health check, and supply the fleet(s) you want
-to load balance to in the `targets` property.
+to load balance to in the `targets` property. Additionally, you need to set `internetFacing`
+as `true` on the load balancer(this is an internet facing load balancer), otherwise the
+load balancer would not be accessible from the internet.
```
rix0rrr, line 13: You're phrasing this as "you MUST do this otherwise it will not be accessible from the internet". If it had been mandatory to always put `internetFacing: true`, then we would have done it for you 🤓 that's the point of CDK!

It's not *necessary* for the load balancer to be accessible from the internet. You may create a load balancer for internal services that just want to talk to each other on a private network (and for safety reasons that is the default).

So that the very least, you should phrase this as "If you want the load balancer to be accessible from the internet, set `internetFacing: true`."

rix0rrr (changes requested):

rix0rrr (approved):

PR #24745, opened 2023-03-22
feat(ses): event destinations for configuration sets

jogold:
Add a L2 for `ConfigurationSetEventDestination`.

Only Amazon SNS and Amazon CloudWatch have been implemented.

By default, the destination is enabled and all events are published.

Also add a `addEventDestination()` convenience method to `ConfigurationSet`.

----


packages/@aws-cdk/aws-ses/lib/configuration-set-event-destination.ts
```diff
@@ -0,0 +1,279 @@
...
+   * @default - a CloudFormation generated name
+   */
+  readonly configurationSetEventDestinationName?: string;
+
+  /**
+   * Whether Amazon SES publishes events to this destination
+   *
+   * @default true
+   */
+  readonly enabled?: boolean;
```
TheRealAmazonKendra, line 36: Is there a use case for adding the destination and not enabling it? This seems unnecessary but maybe I'm missing something.

packages/@aws-cdk/aws-ses/lib/configuration-set-event-destination.ts
```diff
@@ -0,0 +1,279 @@
...
+  /**
+   * The name of an Amazon CloudWatch dimension associated with an email sending metric.
+   */
+  readonly name: string;
+
+  /**
+   * The default value of the dimension that is published to Amazon CloudWatch
+   * if you do not provide the value of the dimension when you send an email.
+   */
+  readonly default: string;
```
TheRealAmazonKendra, line 180: I'm wondering if calling this `defaultValue` would be more clear.

packages/@aws-cdk/aws-ses/lib/configuration-set-event-destination.ts
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/**
+ * A configuration set event destination
+ */
+export class ConfigurationSetEventDestination extends Resource implements IConfigurationSetEventDestination {
+  /**
+   * Use an existing configuration set
+   */
+  public static fromConfigurationSetEventDestinationId(
```
TheRealAmazonKendra, line 224: Is there a case to also have a `fromConfigurationSetEventDestinationName`?

TheRealAmazonKendra (changes requested):
Overall this is fantastic. Just a few small comments/questions inline.

packages/@aws-cdk/aws-ses/lib/configuration-set-event-destination.ts
jogold: Can be used in staging/dev/test environments where it can be "manually" enabled for testing purposes (so you don't pay for 100% of the time enabled in those environments).

packages/@aws-cdk/aws-ses/lib/configuration-set-event-destination.ts
jogold: Changed to `defaultValue`.

packages/@aws-cdk/aws-ses/lib/configuration-set-event-destination.ts
jogold: The `id` is the name I think but the doc is not really clear...

image


Anyway I don't think this resource will be referenced by another CF resource, even the `fromConfigurationSetEventDestinationId()` will not be used I think... (but makes the linter happy)

packages/aws-cdk-lib/aws-ses/lib/configuration-set-event-destination.ts
```diff
@@ -0,0 +1,279 @@
...
+import { Construct } from 'constructs';
+import { IConfigurationSet } from './configuration-set';
+import { CfnConfigurationSetEventDestination } from './ses.generated';
+
+/**
+ * A configuration set event destination
+ */
+export interface IConfigurationSetEventDestination extends IResource {
+  /**
+   * The ID of the configuration set even destination
```
jogold, line 13:
```suggestion
   * The ID of the configuration set event destination
```

otaviomacedo (approved):

PR #24725, opened 2023-03-21
chore(custom-resource): Using NODEJS_18_X with AwsCustomResource

WinterYukky:
Details will be given in due course.

Closes #23989.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #24747, opened 2023-03-22
feat(custom-resource): Using latest runtime with AwsCustomResource

WinterYukky:
Reopen #24725 

I will create PR that is region check mechanism before merge this PR.

Closes #<issue number here>.

----


TheRealAmazonKendra:
Hey @WinterYukky, I wouldn't invest more time in this one because we can't actually upgrade all regions to the latest version. We have to add some additional handling. I'm working on a change for this to get us as up-to-date for each region we can.

WinterYukky:
Thank you @TheRealAmazonKendra for your comment.

I understand that not all regions can be up to date and I have included the code for that in this PR. Specifically, an implementation that manages the latest runtimes by region using `@aws-cdk/region-info`. I also include an implementation that dynamically switches which SDK to use depending on the runtime, as Nodejs 16 and Nodejs 18 runtimes have different versions of the AWS SDK installed internally.

PR #24753, opened 2023-03-22
feat(glue): add ExternalTable for use with connections

### LINKED ISSUE #24741: glue: glue connections to be assigned to glue tables
### Describe the feature

When creating a table, you are able to add an additional parameter to the Table properties called `connectionName` which will allow developers to assign a Glue Data Connection to the `Connection` property.
There should be a field to allow you to directly manipulate the `connectionName` property. In addition, there could potentially be a need to allow direct access of the `location` property, as crawlers tend to change the location property, if crawling a third party database.

### Use Case

Databases that contain structured data, that may never change (or change very rarely) would not need the use of a crawler to update the schema. This could be good in reducing cost of the end user.
In addition, when running glue jobs, developers can directly use the table, as this will also contain the connection needed.

### Proposed Solution

- Add a `connection` property that could potentially take an existing glue connection, or it's associating connection name.
- Allow dynamic changing of the `location` name, if the glue data source does not reside in S3. This could either be a separate feature request, or could be included here.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.68.0

Amazon Linux 2

Rizxcviii:
Changing the table structure to include an initial `TableBase` abstract class, allowing different tables of different data sources to be created from. Initially there are two, `S3Table` and `ExternalTable`.

- `S3Table`: The current table structure that has been used throughout the previous versions of the CDK
- `ExternalTable`: The new glue table that will be used to store metadata about external data sources. This subclass will contain an `externalDataLocation` property to explicitly specify the `Location` property of the underlying `CfnTable` L1 construct
- `Table`: This is now `@deprecated` to shift the usage towards `S3Table`

Closes #24741.

----


Rizxcviii:
Ignore the request lol, I got it to work

Rizxcviii:
@corymhall @TheRealAmazonKendra Can this get reviewed? Thank you.

packages/@aws-cdk/aws-glue-alpha/README.md
```diff
@@ -219,7 +219,23 @@ new glue.Table(this, 'MyTable', {
...
+  database: myDatabase,
+  columns: [{
+    name: 'col1',
+    type: glue.Schema.STRING,
+  }],
+  dataFormat: glue.DataFormat.JSON,
+});
+```
+
+Using `externalDataLocation` will override the `bucket` and `s3Prefix` properties.
```
mrgrain, line 238: As I understand it `externalDataLocation` only works together with `connection`, right?
If that's the case, your two new examples probably should be combined into one.

packages/@aws-cdk/aws-glue-alpha/lib/table.ts
```diff
@@ -421,7 +455,8 @@ export class Table extends Resource implements ITable {
    *
    * @param grantee the principal
    */
-  public grantRead(grantee: iam.IGrantable): iam.Grant {
+  public grantRead(grantee: iam.IGrantable): iam.Grant | undefined {
+    if (!this.bucket) return;
```
mrgrain, line 459: This is I don't think this is accurate. you probably only want to skip the S3 stuff here.

mrgrain (changes requested):
Thanks for this, this will be an important addition to the alpha module.

I'm not quite sure about the design however. The special cases inside the `grant*()` methods make me think we should have separate subclasses for this. I even leaning towards `S3Table` being the special case of the regular table that just takes a location string.

What do you think?

packages/@aws-cdk/aws-glue-alpha/README.md
Rizxcviii: You're right actually. I however read in the documentation that [nested properties should rather be flattened](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#flat)

packages/@aws-cdk/aws-glue-alpha/lib/table.ts
Rizxcviii: I see what you mean, there's a lot of functionality missing for this. I'll get to work on it

packages/@aws-cdk/aws-glue-alpha/README.md
Rizxcviii: Okay, I'll get to work on this.

Rizxcviii: This has been done

Rizxcviii:
> Thanks for this, this will be an important addition to the alpha module.
> 
> I'm not quite sure about the design however. The special cases inside the `grant*()` methods make me think we should have separate subclasses for this. I'm even leaning towards `S3Table` being the special case of the regular table that just takes a location string.
> 
> What do you think?

Thinking about it, it does make sense, and I was leaning towards what you are saying whilst I was developing this. As glue connections was a feature that was somewhat needed, I tried to find a middle ground at the moment, to implement a good solution that could be built into the CDK. But this could definitely be implemented. Depending on whether this should be implemented now or later, that's up to you.

mrgrain:
> But this could definitely be implemented. Depending on whether this should be implemented now or later, that's up to you.

There is some internal push on getting the module stabilized. I've reached out to the people to see if the table construct is on their radar. It _might_ be worth getting this through an RFC on it's own.

packages/@aws-cdk/aws-glue-alpha/lib/table.ts
Rizxcviii: This has been updated

Rizxcviii:
> > But this could definitely be implemented. Depending on whether this should be implemented now or later, that's up to you.
> 
> There is some internal push on getting the module stabilized. I've reached out to the people to see if the table construct is on their radar. It _might_ be worth getting this through an RFC on it's own.

Okay cool, for now I've left out the separate class, and have instead implemented a check for S3 buckets to begin with inside the functions

Rizxcviii:
> > Okay cool, for now I've left out the separate class, and have instead implemented a check for S3 buckets to begin with inside the functions
> 
> Thanks @Rizxcviii . We'd like to go with the separate classes.
> 
> * Deprecate current `Table`.
> * Add new `S3Table`
> * Add new `ConnectedTable` (or maybe `ExternalTable` is better 🤔 )

When you refer to "deprecate", are you referring to deleting the construct entirely, or specifying that the construct is "deprecated" in the documentation?
Also, I like `ExternalTable`

rix0rrr:
@Rizxcviii we don't want to disturb existing users unnecessarily. Let's `@deprecate` them, and make sure that when we stabilize we don't add them back in.

Rizxcviii:
@rix0rrr @mrgrain I've implemented separate classes. Is there any chance in reviewing please?

packages/@aws-cdk/aws-glue-alpha/lib/table-deprecated.ts
mrgrain, line 1: Thanks for that! There's a few changes you now need to revert on the old deprecated table to keep it the same as before.

packages/@aws-cdk/aws-glue-alpha/lib/s3-table.ts
mrgrain, line 1: Same here. I would have expected this `S3Table` to conform mostly to the previous interface. It doesn't need location now, does it?

mrgrain (changes requested):

packages/@aws-cdk/aws-glue-alpha/lib/s3-table.ts
Rizxcviii: This was a rather opinionated change from my side. I thought it would be nice to have from the users perspective as the location property is generated within the constructor and not explicitly provided by the user in all cases. Given that it's an abstract member, it could be enforced for further subclasses that are derived from `TableBase`. Although again this is opinionated, and I can revert this if needed

mrgrain:
Let me rephrase my review a bit more succinct. In this PR I'm expecting to see the following:
- A new `ExternalTable` class that implements your suggested feature
- A new `S3Table` that has _exactly_ the same interface as the current `Table`.
- A new abstract `BaseTable` that collates all shared functionality and abstract methods of between `S3Table` and `ExternalTable`
- Deprecation of the old `Table` construct and related interfaces. This could be done in two ways:
   - Just keep the old code around, with no changes except for the `@deprecated` tag
   - Add a new stub class that just extends `S3Table` but is otherwise empty and has the `@deprecated` tag (similar might be needed for some interfaces)

Rizxcviii:
> Let me rephrase my review a bit more succinct. In this PR I'm expecting to see the following:
> 
> * A new `ExternalTable` class that implements your suggested feature
> * A new `S3Table` that has _exactly_ the same interface as the current `Table`.
> * A new abstract `BaseTable` that collates all shared functionality and abstract methods of between `S3Table` and `ExternalTable`
> * Deprecation of the old `Table` construct and related interfaces. This could be done in two ways:
>   
>   * Just keep the old code around, with no changes except for the `@deprecated` tag
>   * Add a new stub class that just extends `S3Table` but is otherwise empty and has the `@deprecated` tag (similar might be needed for some interfaces)

Hi @mrgrain sorry for the continuous questioning, but with point number 2 about the `S3Table` interface, is interface inheritance permissible in that case?

mrgrain:
> Hi @mrgrain sorry for the continuous questioning, but with point number 2 about the S3Table interface, is interface inheritance permissible in that case?

Interface inheritance is permissible. However I'd like to see the new `S3Table` holding the source of truth since that's the class we will be using going forward.

Rizxcviii:
I've created a subclass of `S3Table`, however I didn't inherit the interface. The `TableProps` Interface already exists, so there is a name collision. However, the `S3TableProps` interface contains exactly the same props as the current `Table`, therefore the constructor will just ask for that interface.

packages/@aws-cdk/aws-glue-alpha/lib/s3-table.ts
mrgrain: Ah I see. Let's keep this PR focused on `ExternalTable` and the rename.
You can open a second one with improvements to `S3Table` afterwards. Does that sound okay to you?

Rizxcviii: my bad, I just saw this change requested right now, let me begin working on it

packages/@aws-cdk/aws-glue-alpha/lib/table-deprecated.ts
Rizxcviii: This should be done now

packages/@aws-cdk/aws-glue-alpha/lib/s3-table.ts
Rizxcviii: This has been done, I've removed the `location` prop.

Rizxcviii:
I have now removed the `location` prop, as it was correct to keep this outside of the PR.

mrgrain:
@Rizxcviii Very close now! There's still a couple of unexpected changes to the deprecated Table I'd like reverted:

```
API elements with incompatible changes:
warn - PROP @aws-cdk/aws-glue-alpha.Table.encryption: type Optional<@aws-cdk/aws-glue-alpha.TableEncryption> (formerly @aws-cdk/aws-glue-alpha.TableEncryption): output type is now optional [changed-type:@aws-cdk/aws-glue-alpha.Table.encryption]
warn - INITIALIZER @aws-cdk/aws-glue-alpha.Table.<initializer>: argument props, takes @aws-cdk/aws-glue-alpha.S3TableProps (formerly @aws-cdk/aws-glue-alpha.TableProps): @aws-cdk/aws-glue-alpha.TableProps does not extend @aws-cdk/aws-glue-alpha.S3TableProps [incompatible-argument:@aws-cdk/aws-glue-alpha.Table.<initializer>]
warn - PROP @aws-cdk/aws-glue-alpha.TableProps.bucket: has been removed [removed:@aws-cdk/aws-glue-alpha.TableProps.bucket]
warn - PROP @aws-cdk/aws-glue-alpha.TableProps.encryption: has been removed [removed:@aws-cdk/aws-glue-alpha.TableProps.encryption]
warn - PROP @aws-cdk/aws-glue-alpha.TableProps.encryptionKey: has been removed [removed:@aws-cdk/aws-glue-alpha.TableProps.encryptionKey]
warn - PROP @aws-cdk/aws-glue-alpha.TableProps.s3Prefix: has been removed [removed:@aws-cdk/aws-glue-alpha.TableProps.s3Prefix]
```

You can check that list yourself by running `yarn compat` inside the package directory.

packages/@aws-cdk/aws-glue-alpha/test/integ.table.ts
```diff
@@ -116,6 +134,23 @@ new glue.Table(stack, 'MyTableWithStorageDescriptorParameters', {
...
 });
 
+new glue.ExternalTable(stack, 'MyTableWithCustomLocation', {
+  database,
+  connection,
+  tableName: 'custom_location_table',
+  columns,
+  dataFormat: glue.DataFormat.JSON,
+  externalDataLocation: 'default_db.public.test',
+});
```
mrgrain, line 144: Can you please split this out into a new separate test?

mrgrain (dismissed):

mrgrain (changes requested):
See other comments.

Rizxcviii:
> @Rizxcviii Very close now! There's still a couple of unexpected changes to the deprecated Table I'd like reverted:
> 
> ```
> API elements with incompatible changes:
> warn - PROP @aws-cdk/aws-glue-alpha.Table.encryption: type Optional<@aws-cdk/aws-glue-alpha.TableEncryption> (formerly @aws-cdk/aws-glue-alpha.TableEncryption): output type is now optional [changed-type:@aws-cdk/aws-glue-alpha.Table.encryption]
> warn - INITIALIZER @aws-cdk/aws-glue-alpha.Table.<initializer>: argument props, takes @aws-cdk/aws-glue-alpha.S3TableProps (formerly @aws-cdk/aws-glue-alpha.TableProps): @aws-cdk/aws-glue-alpha.TableProps does not extend @aws-cdk/aws-glue-alpha.S3TableProps [incompatible-argument:@aws-cdk/aws-glue-alpha.Table.<initializer>]
> warn - PROP @aws-cdk/aws-glue-alpha.TableProps.bucket: has been removed [removed:@aws-cdk/aws-glue-alpha.TableProps.bucket]
> warn - PROP @aws-cdk/aws-glue-alpha.TableProps.encryption: has been removed [removed:@aws-cdk/aws-glue-alpha.TableProps.encryption]
> warn - PROP @aws-cdk/aws-glue-alpha.TableProps.encryptionKey: has been removed [removed:@aws-cdk/aws-glue-alpha.TableProps.encryptionKey]
> warn - PROP @aws-cdk/aws-glue-alpha.TableProps.s3Prefix: has been removed [removed:@aws-cdk/aws-glue-alpha.TableProps.s3Prefix]
> ```
> 
> You can check that list yourself by running `yarn compat` inside the package directory.

This is now fixed. I ran `yarn compat` after the fix and there appears to be no issues

mrgrain (dismissed):

Rizxcviii:
@mrgrain There was an error regarding permissions to update the branch, and it recommended that I update the branch locally, I didn't know that it would remove your approval, my bad :'(

mrgrain (approved):

mrgrain:
No worries! Sometimes there are conflicts and than this is unavoidable.

Issue #24743, opened 2023-03-22
eks: policy does not exist or is not attachable in China regions


pahud:
### Describe the bug

When I deployed eks cluster to `cn-north-1` I got the error:

>Policy arn:aws-cn:iam::aws:policy/AmazonElasticContainerRegistryPublicReadOnly does not exist or is not attachable. (Service: AmazonIdentityManagement; Status Code: 404; Error Code: NoSuchEntity; Request ID: 084210a1-3251-4868-9420-f5e47d62f66a; Proxy: null)

<img width="1245" alt="image" src="https://user-images.githubusercontent.com/278432/226941421-a20f8847-ca2f-4db6-8f6b-7c020052096b.png">

### Expected Behavior

successful deployment

### Current Behavior

fail to deploy with the error message above.

### Reproduction Steps

```ts
      const cluster = new eks.Cluster(this, 'Cluster', {
        vpc,
        version: eks.KubernetesVersion.V1_25,
        kubectlLayer: new KubectlLayer(this, 'KubectlLayer'),
        defaultCapacity: 0,
      })
```

### Possible Solution

ECR Public is not available in AWS China regions and the `AmazonElasticContainerRegistryPublicReadOnly` is not available. We probably should exclude it when deploying in China regions. And this brings up another issue - if the eks cluster in China is trying to helm install charts from ECR public, it should not run the `aws ecr-public` commands like this:

https://github.com/aws/aws-cdk/blob/a51346ebe1115e900d77d436b4a162871dbb2109/packages/%40aws-cdk/aws-eks/lib/kubectl-handler/helm/__init__.py#L115

We probably should fix that in a separate PR.

### CDK CLI Version

2.69.0

### Node.js Version

v19.6.0

### OS

mac os x

### Language

Typescript

pahud:
PR underway.

wongma7:
it also affects govcloud https://github.com/aws/aws-cdk/issues/24808.

I'm not sure if I can downgrade CDK to work around the issue because the cluster is 1.25 and changelog says I need 2.69.0 for 1.25 support

PR #24759, opened 2023-03-23
fix(eks): policy does not exist or is not attachable in China regions

pahud:
As ECR Public is not available in China regions and govcloud, `AmazonElasticContainerRegistryPublicReadOnly` is not available and should not be attached to the role. This PR implements a `RegionInfo` method to determine if ECR public is available in current region and conditionally attach `AmazonElasticContainerRegistryPublicReadOnly` policy to the role.

Closes https://github.com/aws/aws-cdk/issues/24743 https://github.com/aws/aws-cdk/issues/24808

----


pahud:
Exemption Request

packages/@aws-cdk/aws-eks/lib/private/service-available.ts
```diff
@@ -0,0 +1,10 @@
+import { Token } from '@aws-cdk/core';
+import { RegionInfo } from '@aws-cdk/region-info';
+
+export function EcrPublicAvailable(region: any): boolean {
+  if (region && !Token.isUnresolved(region)) {
```
TheRealAmazonKendra, line 5: Is there any handling we can do for when a token is being used? It seems like it's a fairly common use case and this fix would still result in an error for those users.

pahud: Yes thank you for the heads up. I will fix this soon.

pahud:
close in favor of https://github.com/aws/aws-cdk/pull/25170

PR #24760, opened 2023-03-23
fix(servicecatalogappregistry): RAM Share is replaced on every change to Application

liwewang-amazon:
`Application.shareApplication` can be called multiple times which generates different resources within the same `Application` construct. The share construct id generation was based on child length of `Application` construct. This makes every change to the `Application` construct can unnecessarily replace the RAM share when the share actually needs no update.

Similar to the solution for `addAttributeGroup` introduced in the [previous update](https://github.com/aws/aws-cdk/pull/24672), this commit starts to require both construct id and share name in the `Application.shareApplication` method input, which are used to create RAM share construct.

For `ApplicationAssociator`, `Application.shareApplication` is called for cross-account stack associations. Share construct id depends on the node unique id of the application in `ApplicationAssociator` and the node unique id of the stack to be associated. This reduces unnecessary share replacement in `ApplicationAssociator` stack.

BREAKING CHANGE: This commit involves share replacement during the deployment of `ApplicationAssociator` due to share construct id update. After this change, frequent share replacements due to structural change in `Application` construct should be avoided. `Application.shareApplication` starts to require construct id (first argument) and share name (added in `ShareOption`) as input.

----


rix0rrr (changes requested):
Please make sure that your PR title confirms to the conventional commit standard (`fix`, `feat`, `chore`) and that it is written in a style that will reflect correctly in the change log (See [Contributing Guide, Pull Requests](https://github.com/aws/aws-cdk/blob/v1-main/CONTRIBUTING.md#step-4-pull-request))

rix0rrr (approved):

PR #24770, opened 2023-03-23
fix(s3): deprecate unencrypted encryption configuration for s3 bucket

clueleaf:
S3 now applies SSE-S3 by default if server side encryption is not configured.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/serv-side-encryption.html

This PR simply deprecates `UNENCRYPTED` configuration of `BucketEncryption`, and changes some docs because they do not seem precise anymore.
Behavior of S3 bucket creation is not modified. Many other constructs create S3 buckets internally and updating all those related constructs will result in a huge impact.

This PR also deprecates `TableEncryption.UNENCRYPTED` of AWS Glue and creates an S3 bucket with `S3_MANAGED` if table encryption is not specified.

----


clueleaf:
Exemption Request
Integration test files are not modified because behavior is not changed.

TheRealAmazonKendra:
> Exemption Request Integration test files are not modified because behavior is not changed.

Unfortunately, unless the tests are updated, the build will not pass so we cannot grant this exemption.

TheRealAmazonKendra:
Actually, in this case it's failing because the tests aren't using `testDeprecated`. For any tests within the S3 package, please use `testDeprecated`. For those that explicitly set `BucketEncryption.UNENCRYPTED` outside the package, please update to an appropriate value and/or leave that field unset.

clueleaf:
@TheRealAmazonKendra 
Thank you for your guide. Now I am using `testDeprecated` for S3 unit tests with `BucketEncryption.UNENCRYPTED`.

I also deprecated `TableEncryption.UNENCRYPTED` of Glue service because users can explicitly specify S3 bucket encryption status.
For other services, removed `BucketEncryption` if it is set to `UNENCRYPTED`.

packages/@aws-cdk/aws-glue/lib/table.ts
```diff
@@ -46,6 +46,13 @@ export interface ITable extends IResource {
  * @see https://docs.aws.amazon.com/athena/latest/ug/encryption.html
  */
 export enum TableEncryption {
+  /**
+   * Previous option. Tables can not be unencrypted now.
```
otaviomacedo, line 50: Given that `aws-glue` is still alpha, we should consider removing this property altogether, so we don't carry this dead weight up to graduation.

I'd like to hear what other contributors think.

clueleaf: It sounds fair to me.
Waiting for some more comments and I'll proceed.

TheRealAmazonKendra: Agreed on this.

packages/aws-cdk-lib/aws-cloudtrail/test/cloudtrail.test.ts
```diff
@@ -258,8 +258,8 @@ describe('cloudtrail', () => {
         trailName: 'KmsKeyTrail',
         encryptionKey: key,
       });
-      new Trail(stack, 'UnencryptedTrail', {
-        trailName: 'UnencryptedTrail',
+      new Trail(stack, 'DefauleTrail', {
+        trailName: 'DefauleTrail',
```
TheRealAmazonKendra, line 262:
```suggestion
      new Trail(stack, 'DefaultTrail', {
        trailName: 'DefaultTrail',
```

packages/aws-cdk-lib/aws-cloudtrail/test/cloudtrail.test.ts
```diff
@@ -275,7 +275,7 @@ describe('cloudtrail', () => {
         },
       });
       Template.fromStack(stack).hasResourceProperties('AWS::CloudTrail::Trail', {
-        TrailName: 'UnencryptedTrail',
+        TrailName: 'DefauleTrail',
```
TheRealAmazonKendra, line 278:
```suggestion
        TrailName: 'DefaultTrail',
```

TheRealAmazonKendra (changes requested):

packages/@aws-cdk/aws-glue/lib/table.ts
clueleaf: Removed `UNENCRYPTED` property.

packages/aws-cdk-lib/aws-s3/test/bucket.test.ts
otaviomacedo, line 1: The behaviour of `Bucket` hasn't changed, so we shouldn't change the tests, either. If users pass `UNENCRYPTED`, we still want it to work the same way as before, which means we still need these tests in place. 

Having said that, the test coverage for this feature has a lot of gaps, at the moment, such as the `S3_MANAGED` case. So, please either convert them to _new_ test cases or don't modify the tests at all (apart from replacing `test` with `testDeprecated` where appropriate).

otaviomacedo (dismissed):
Thanks, @clueleaf. I think we're almost there. Just check out my comment about the unit tests, which I didn't catch in the first pass.

packages/aws-cdk-lib/aws-s3/test/bucket.test.ts
clueleaf: Thanks. Modified `S3_MANAGED` tests to new test cases.

otaviomacedo (approved):

PR #20871, opened 2022-06-25
feat(batch): add secrets props to job definition

yoshizawa56:
Add a secrets property to batch.JobDefinitionContainer. This interface is almost the same as ecs.ContainerDefinitionOptions.

This is reopen PR of #19506

closes #10976

----

### All Submissions:

* [x] Have you followed the guidelines in our [Contributing guide?](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md)

### Adding new Unconventional Dependencies:

* [ ] This PR adds new unconventional dependencies following the process described [here](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md/#adding-new-unconventional-dependencies)

### New Features

* [x] Have you added the new feature to an [integration test](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md)?
	* [x] Did you use `yarn integ` to deploy the infrastructure and generate the snapshot (i.e. `yarn integ` without `--dry-run`)?

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #24775, opened 2023-03-23
feat(batch): new L2 Constructs

comcalvi:
Adds a new set of L2s for Batch. See https://github.com/aws/aws-cdk-rfcs/pull/484/files for details.

This contains breaking changes to the alpha module! To use the previous API, pin your version of `aws-batch-alpha` to `2.71.0`. 

BREAKING CHANGE: `ComputeEnvironment` has been removed and replaced by `ManagedEc2EcsComputeEnvironment`, `ManagedEc2EksComputeEnvironment`, and `UnmanagedComputeEnvironment`.

`JobDefinition` has been removed and replaced by `EcsJobDefinition`, `EksJobDefinition`, and `MultiNodeJobDefinition`

----


TheRealAmazonKendra:
@Mergifyio update

packages/@aws-cdk/aws-batch/lib/ecs-container-definition.ts
```diff
@@ -0,0 +1,949 @@
...
+
+  /**
+   * The directory within the Amazon EFS file system to mount as the root directory inside the host.
+   * If this parameter is omitted, the root of the Amazon EFS volume is used instead.
+   * Specifying `/` has the same effect as omitting this parameter.
+   * The maximum length is 4,096 characters.
+   *
+   * @default - root of the EFS File System
+   */
+  public readonly rootDirectory?: string;
```
rix0rrr, line 167: For a simple integration class, not sure we need all these. I suppose it doesn't hurt either now that you've written all of it already, but people aren't usually going to be interacting with this object.

packages/@aws-cdk/aws-batch/lib/ecs-container-definition.ts
```diff
@@ -0,0 +1,949 @@
...
+Object.defineProperty(HostVolume.prototype, HOST_VOLUME_SYMBOL, {
+  value: true,
+  enumerable: false,
+  writable: false,
+});
+
+/**
+ * A container that can be run with ECS orchestration
+ */
+export interface IEcsContainerDefinition extends IConstruct {
```
rix0rrr, line 273: Is a containerdefinition a construct? Does it live on its own in the construct tree somewhere?

rix0rrr (commented):
My brain stopped working after a while :) this is what I've got so far

packages/@aws-cdk/aws-batch/README.md
```diff
@@ -23,394 +23,602 @@
...
+*Note*: Batch does not allow specifying instance types or classes with different architectures.
+For example, `InstanceClass.A1` cannot be specified alongside `InstanceClass.OPTIMAL`,
+because `A1` uses ARM and `OPTIMAL` uses x86_64.
+You can specify both `InstanceType.OPTIMAL` alongside several different instance types in the same compute environment:
 
-Compute Environments implement the `IConnectable` interface, which means you can use
-connections on other CDK resources to manipulate the security groups and allow access.
+```ts
+const computeEnv = new batch.ManagedEC2ComputeEnvironment(this, 'myEc2ComputeEnv', {
+  instanceTypes: [batch.InstanceType.of(ec2.InstanceClass.M5AD, ec2.InstanceSize.LARGE)],
```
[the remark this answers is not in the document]
comcalvi: woops, the idea was to extend but I've scrapped it all in favor of a `useOptimalInstanceTypes` flag that will handle the `'optimal'` instance class recognized by batch

packages/@aws-cdk/aws-batch/lib/compute-environment-base.ts
```diff
@@ -0,0 +1,119 @@
...
+   * scale instances up or down.
+   *
+   * If you change a ComputeEnvironment from enabled to disabled while it is executing jobs,
+   * Jobs in the `STARTED` or `RUNNING` states will not
+   * be interrupted. The ComputeEnvironment will no longer scale instances down;
+   * that is, any instances that were scaled up will remain if the ComputeEnvironment is disabled.
+   *
+   * To ensure you aren't billed for unused capacity, remove the ComputeEnvironment from your stack.
+   */
+  readonly enabled: boolean;
```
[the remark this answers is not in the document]
comcalvi: We set it to `true` and give empty values where needed:

```
  public static fromManagedEc2EcsComputeEnvironmentArn(
    scope: Construct, id: string, managedEc2EcsComputeEnvironmentArn: string,
  ): IManagedEc2EcsComputeEnvironment {
    const stack = Stack.of(scope);
    const computeEnvironmentName = stack.splitArn(managedEc2EcsComputeEnvironmentArn, ArnFormat.SLASH_RESOURCE_NAME).resourceName!;

    class Import extends Resource implements IManagedEc2EcsComputeEnvironment {
      public readonly computeEnvironmentArn = managedEc2EcsComputeEnvironmentArn;
      public readonly computeEnvironmentName = computeEnvironmentName;
      public readonly enabled = true;
      public readonly instanceClasses = [];
      public readonly instanceTypes = [];
      public readonly maxvCpus = 1;
      public readonly connections = { } as any;

      public addInstanceClass(_instanceClass: ec2.InstanceClass): void {
        throw new Error(`cannot add instance class to imported ComputeEnvironment '${id}'`);
      }
      public addInstanceType(_instanceType: ec2.InstanceType): void {
        throw new Error(`cannot add instance type to imported ComputeEnvironment '${id}'`);
      }
    }

    return new Import(scope, id);
  }
```

packages/@aws-cdk/aws-batch/lib/compute-environment-base.ts
```diff
@@ -0,0 +1,119 @@
...
+  public abstract readonly computeEnvironmentArn: string;
+
+  protected resourceProps: CfnComputeEnvironmentProps;
+
+  constructor(scope: Construct, id: string, props?: ComputeEnvironmentProps) {
+    super(scope, id, {
+      physicalName: props?.computeEnvironmentName,
+    });
+
+    this.computeEnvironmentName = props?.computeEnvironmentName;
```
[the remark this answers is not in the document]
comcalvi:
```
 this.computeEnvironmentName = props?.computeEnvironmentName;
```

has been removed

packages/@aws-cdk/aws-batch/README.md
```diff
@@ -23,394 +23,602 @@
...
+});
+jobDefn.containerDefinition.addHostPathVolume({
+  name: 'hostPath',
+  hostPath: '/sys',
+  mountPath: '/Volumes/hostPath',
+});
+jobDefn.containerDefinition.addSecretVolume({
+  name: 'secret',
+  optional: true,
+  mountPath: '/Volumes/secret',
```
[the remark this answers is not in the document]
rix0rrr: I'd rather keep it consistent

packages/@aws-cdk/aws-batch/lib/compute-environment-base.ts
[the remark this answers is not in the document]
rix0rrr: I understand, but who *needs* it? What other construct cannot operate without this exposed?

Because it seems to be designed for use by subclasses, but the subclass initialization is awkward currently.

packages/@aws-cdk/aws-batch/lib/ecs-container-definition.ts
rix0rrr: I suppose there's also nothing AGAINST containerdefinition being a construct. Oh hell, why not.

packages/@aws-cdk/aws-batch/lib/ecs-container-definition.ts
```diff
@@ -0,0 +1,939 @@
...
+
+  /**
+   * Whether or not to use the AWS Batch job IAM role defined in a job definition when mounting the Amazon EFS file system.
+   * If specified, `enableTransitEncryption` must be `true`.
+   *
+   * @see https://docs.aws.amazon.com/batch/latest/userguide/efs-volumes.html#efs-volume-accesspoints
+   *
+   * @default false
+   */
+  readonly useJobDefinitionRole?: boolean;
```
rupard-aws, line 140: Batch calls this role the `JobRole[Arn]` , Batch doesn't really use the term JobDefinitionRole in its docs.

packages/@aws-cdk/aws-batch/lib/eks-job-definition.ts
```diff
@@ -0,0 +1,220 @@
...
+      ...this.resourceProps,
+      type: 'container',
+      jobDefinitionName: props.jobDefinitionName,
+      eksProperties: {
+        podProperties: {
+          containers: [
+            this.container._renderContainerDefinition(),
+          ],
+          dnsPolicy: this.dnsPolicy,
+          hostNetwork: this.useHostNetwork,
```
rupard-aws, line 170: in the EKS interface defs, `useHostNetwork` is defaulting to `false`.  This is all a bit misleading.  Batch is defaulting `hostNetwork` to `true` (which is different than Kube defaults) because 

> Most [batch] workloads are egress-only and don't require the overhead of IP allocation for each pod for incoming connections.

The default in CDK should match Batch's default for `hostNetwork` which means `null` or not-set is true in Batch.

packages/@aws-cdk/aws-batch/lib/job-definition-base.ts
```diff
@@ -0,0 +1,311 @@
...
+ * Abstract base class for JobDefinitions
+ *
+ * @internal
+ */
+export abstract class JobDefinitionBase extends Resource implements IJobDefinition {
+  public readonly abstract jobDefinitionArn: string;
+  public readonly abstract jobDefinitionName: string;
+
+  public readonly parameters?: { [key:string]: any };
+  public readonly propagateTags?: boolean;
```
rupard-aws, line 260: `propagateTags` is not supported for EKS/Kube Jobs.

packages/@aws-cdk/aws-batch/lib/managed-compute-environment.ts
```diff
@@ -0,0 +1,1222 @@
...
+
+  private readonly instanceProfile: iam.CfnInstanceProfile;
+
+  constructor(scope: Construct, id: string, props: ManagedEc2EcsComputeEnvironmentProps) {
+    super(scope, id, props);
+
+    this.images = props.images;
+    this.allocationStrategy = determineAllocationStrategy(id, props.allocationStrategy, this.spot);
+    this.spotBidPercentage = props.spotBidPercentage;
+    this.spotFleetRole = props.spotFleetRole ?? (this.spot ? createSpotFleetRole(this) : undefined);
```
rupard-aws, line 633: This is _only_ required if `spot` and `allocationStrategy` is set explicitly to BEST_FIT or null (to api).  The spot SCO and BFP do not require it.

packages/@aws-cdk/aws-batch/lib/managed-compute-environment.ts
```diff
@@ -0,0 +1,1222 @@
...
+    this.kubernetesNamespace = props.kubernetesNamespace;
+    this.eksCluster = props.eksCluster;
+
+    this.images = props.images;
+    this.allocationStrategy = determineAllocationStrategy(id, props.allocationStrategy, this.spot);
+    if (this.allocationStrategy === AllocationStrategy.BEST_FIT) {
+      throw new Error(`ManagedEc2EksComputeEnvironment '${id}' uses invalid allocation strategy 'AllocationStrategy.BEST_FIT'`);
+    }
+    this.spotBidPercentage = props.spotBidPercentage;
+    this.spotFleetRole = props.spotFleetRole ?? (this.spot ? createSpotFleetRole(this) : undefined);
```
rupard-aws, line 998: EKS CE does not support BEST_FIT SPOT, therefore spotFleetRole is never needed in for any EKS CE allocation type.

packages/@aws-cdk/aws-batch/lib/managed-compute-environment.ts
```diff
@@ -0,0 +1,1222 @@
...
+  * However, if you can continue to roll it back,
+  * you can return the stack to its original settings and then try to update it again.
+  *
+  * The properties which require a replacement of the Compute Environment are:
+  *
+  * `allocationStrategy`, `spotBidPercentage`, // TODO
+  * @see: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-batch-computeenvironment.html#cfn-batch-computeenvironment-replacecomputeenvironment
+  * @see: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-continueupdaterollback.html
+  *
+  * @default true
```
rupard-aws, line 133: Batch may want this value to be `false` by default for new CFN based CEs since replacement is the legacy behavior once Batch released dynamic infra updates in CEs, https://docs.aws.amazon.com/batch/latest/userguide/updating-compute-environments.html

I will follow up once I have more info, FYI for now...

rupard-aws (changes requested):
Thanks for your epic effort here :)

packages/@aws-cdk-testing/alpha-module-stable-tests/README.md
```diff
@@ -0,0 +1,31 @@
+# CDK CLI integration test
```
rix0rrr, line 1: Does this package need to exist? They look like unit tests only, any particular reason these tests couldn't just go into the alpha module itself (or the alpha integration module itself)? 

It already depends on `aws-cdk-lib` anyway, so putting unit tests there wouldn't add a dependency arrow.

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/placement-group.test.ts
```diff
@@ -0,0 +1,118 @@
+import { Template } from '@aws-cdk/assertions';
```
rix0rrr, line 1: This is  a unit test file, not an integ test file.

packages/@aws-cdk-testing/alpha-module-stable-tests/README.md
comcalvi: these are unit tests, and they are unit tests that depend on both `aws-cdk-lib` and an alpha module (here, `aws-batch`). They can't go in `aws-cdk-lib`, because then `aws-cdk-lib` depends on `aws-batch`, and that is a circular dependency that Lerna yells at me for:

```lerna ERR! ECYCLE Dependency cycles detected, you should fix these!
lerna ERR! ECYCLE aws-cdk-lib -> @aws-cdk/aws-batch-alpha -> @aws-cdk/integ-runner -> aws-cdk -> aws-cdk-lib
```

We could put them in the alpha module itself. Is that preferable?

packages/@aws-cdk/aws-batch/lib/managed-compute-environment.ts
rupard-aws: after syncing with some Batch SMEs, we concluded that `ReplaceComputeEnvironment` should default to `false` because that is the better experience long-term for ComputeEnvironments created by these new constructs.

rix0rrr (approved):

rjuengling-hf:
Prior to this change container secrets [could be added](https://pypi.org/project/aws-cdk.aws-batch-alpha/2.73.0a0/#job-definition) to a `JobDefinition` this way:
```
db_secret = secretsmanager.Secret(self, "secret")

batch.JobDefinition(self, "batch-job-def-secrets",
    container=batch.JobDefinitionContainer(
        image=ecs.EcrImage.from_registry("docker/whalesay"),
        secrets={
            "PASSWORD": ecs.Secret.from_secrets_manager(db_secret, "password")
        }
    )
)
```

I suppose now the secrets need to be set via `EcsEc2ContainerDefinitiont` & co. I have tried a few things but have not found a way to do this. Could you please provide an example such as the one I quoted, for the new API?

rjuengling-hf:
Thanks. What I am not seeing in your example is the mapping from a secret to an environment variable set in the Docker container. In the example I quoted a secret is made available in the Docker container per environment variable `PASSWORD`.

How is the secret to be accessed in the Docker container with the new API?

comcalvi:
I think this is the feature you're talking about: https://github.com/aws/aws-cdk/pull/20871

This is not supported in the new APIs, so I can add support for it; should just be changing the type from `Secret[]` to { [key: string]: Secret }`

PR #24780, opened 2023-03-24
fix(batch): deny zero maxvCpus in compute environment instead of implicitly setting it to 256

daniel-j-h:
In the batch compute environment, the maxVcpu property is optional and defaults to 256.

If users set maxvCpus to zero they expect the compute environment to not run any computation, i.e. not enabled.

Instead at the moment when the user passes maxVcpu of zero we set if to 256 implicitly due to an unfortunate initialization issue.

We just ran into this issue where our on-demand compute environment suddenly ran with 256 vcpus unnoticed, even though in the code it said `maxVcpus: 0`. This changeset flags and denies this implicit behavior.

----

https://github.com/aws/aws-cdk/blob/02c9d98ae30c1e57e8fa95ed143701f3f43e6f04/packages/%40aws-cdk/aws-batch/lib/compute-environment.ts#L201-L207

https://github.com/aws/aws-cdk/blob/02c9d98ae30c1e57e8fa95ed143701f3f43e6f04/packages/%40aws-cdk/aws-batch/lib/compute-environment.ts#L406

----

### DIFF
--- a/packages/@aws-cdk/aws-batch/lib/compute-environment.ts
+++ b/packages/@aws-cdk/aws-batch/lib/compute-environment.ts
@@ -403,7 +403,7 @@ export class ComputeEnvironment extends Resource implements IComputeEnvironment,
         ec2KeyPair: props.computeResources.ec2KeyPair,
         imageId: props.computeResources.image && props.computeResources.image.getImage(this).imageId,
         launchTemplate: props.computeResources.launchTemplate,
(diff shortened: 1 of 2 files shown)


TheRealAmazonKendra:
Thank you for your contribution. This functionality is being addressed in a previously existing [PR](https://github.com/aws/aws-cdk/pull/24775) so I'm going to go ahead and close this one.

PR #24781, opened 2023-03-24
fix(lambda-nodejs): pnpm installs frozen lockfile in a CI environment

Tucker-Eric:
[pnpm automatically enables `--frozen-lockfile` in CI environments](https://pnpm.io/cli/install#--frozen-lockfile) which breaks `NodejsFunction` local bundling.

This change appends the `--no-prefer-frozen-lockfile` flag as described in https://github.com/pnpm/pnpm/issues/1994#issuecomment-609403673

----

### DIFF
--- a/CHANGELOG.v2.alpha.md
+++ b/CHANGELOG.v2.alpha.md
@@ -2,6 +2,25 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.72.0-alpha.0](https://github.com/aws/aws-cdk/compare/v2.71.0-alpha.0...v2.72.0-alpha.0) (2023-03-29)
+
+
+### ⚠ BREAKING CHANGES TO EXPERIMENTAL FEATURES
+
+* **servicecatalogappregistry:** This commit involves share replacement during the deployment of `ApplicationAssociator` due to share construct id update. After this change, frequent share replacements due to structural change in `Application` construct should be avoided. `Application.shareApplication` starts to require construct id (first argument) and share name (added in `ShareOption`) as input.
+* **ivs:** Renamed ChannelProps.name to ChannelProps.channelName
+* Renamed PlaybackKeyPairProps.name to PlaybackKeyPairProps.playbackKeyPairName
+* Channel now generates a physical name if one is not provided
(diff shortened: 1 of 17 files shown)


Tucker-Eric:
Looking for an Exemption Request for the integration test as well.

TheRealAmazonKendra:
I'm removing the exemption request label for now. Please use a quote reply to my last message to add it back when you respond to my question so I see that this needs another look.

Thanks for you work here!

Tucker-Eric:
> > Looking for an Exemption Request for the integration test as well.
> 
> To provide the exemption request, can you please describe how this was manually tested? I am inclined to agree that this is difficult to write an automated test for.

Hey @TheRealAmazonKendra! Appreciate your help with this!

This was manually tested locally using the command dumped in the error output of the current command.
image

Removing/replacing the asset staging path with a local one and rerunning the same commands in order (with the `CI=true` env variable to mimmic a CI environment) the following commands were run.

without `--no-prefer-frozen-lockfile`
```bash
mkdir asset-staging && \
    echo '{"dependencies":{"pg":"8.10.0"}}' > "asset-staging/package.json" && \
    cp "pnpm-lock.yaml" "asset-staging/pnpm-lock.yaml" && \
    cd asset-staging && \
    CI=true pnpm install --config.node-linker=hoisted --config.package-import-method=clone-or-copy && \
    rm "node_modules/.modules.yaml"
```

with `--no-prefer-frozen-lockfile`
```bash
mkdir asset-staging && \
    echo '{"dependencies":{"pg":"8.10.0"}}' > "asset-staging/package.json" && \
    cp "pnpm-lock.yaml" "asset-staging/pnpm-lock.yaml" && \
    cd asset-staging && \
    CI=true pnpm install --config.node-linker=hoisted --config.package-import-method=clone-or-copy --no-prefer-frozen-lockfile && \
    rm "node_modules/.modules.yaml"
```
image

TheRealAmazonKendra (approved):

TheRealAmazonKendra:
@Mergifyio reset

ofhouse:
I there any ETA when this fix will be published?
Tested it with `aws-cdk-lib@2.72.1` but it seems like this was not part of the release: https://unpkg.com/browse/aws-cdk-lib@2.72.1/aws-lambda-nodejs/lib/package-manager.js

PR #24786, opened 2023-03-25
fix(events-targets): parameter JobName can exceed limit of 128 characters

### LINKED ISSUE #24654: (aws-events): Automatically Generated JobName Must Be Shorter Than 128 Char Long
### Describe the bug

In `AWS::Events::Rule`, `BatchParameters`, the `JobName` must be shorter than 128 character long. Otherwise it fails with:
```
Parameter JobName is not valid. Reason: Length must be in between 1 and 128. (Service: AmazonCloudWatchEvents; Status Code: 400; Error Code: ValidationException; Request ID: 1234567890; Proxy: null)
```

### Expected Behavior

Ideally `BatchJob` (in `aws-events-targets`) should generate a shorter name, or the very least the name should be checked in `events.Rule` to fail during `synth` phase.

### Current Behavior

It fails during deployment.

### Reproduction Steps

My stack is pretty big, so not sure, but I guess creating a stack with longer name and leave it to CDK to generate the `BatchJob.jobName` should be enough to reproduce it. Let me know if you need a sample project.

### Possible Solution

Ideally `BatchJob` (in `aws-events-targets`) should generate a shorter name, or the very least the name should be checked in `events.Rule` to fail during `synth` phase.

Interestingly there's no such requirement on CFN documentation.

### CDK CLI Version

2.69.0

### Framework Version

_No response_

### Node.js Version

18

### OS

MacOS

### Language

Typescript

### Language Version

4.9.0

### Other information

Happy to send a PR if I know how to fix it. Thanks.

lpizzinidev:
`BatchJob` should generate a `JobName` with a length smaller than 128 characters, otherwise, it will fail during `synth`.

Closes #24654.

----


TheRealAmazonKendra (changes requested):
Thank you for your contribution! Overall this looks good. However, we don't need a feature flag in this case because this is a non-stateful resource.

otaviomacedo (approved):

TheRealAmazonKendra (approved):

PR #23514, opened 2022-12-30
fix(s3): logging bucket blocks KMS_MANAGED encryption

biffgaut:
----
closes #23513

### All Submissions:

* [ x] Have you followed the guidelines in our [Contributing guide?](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md)

### Adding new Construct Runtime Dependencies:

* [ ] This PR adds new construct runtime dependencies following the process described [here](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md/#adding-construct-runtime-dependencies)

### New Features

* [ ] Have you added the new feature to an [integration test](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md)?
	* [ ] Did you use `yarn integ` to deploy the infrastructure and generate the snapshot (i.e. `yarn integ` without `--dry-run`)?

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #23385, opened 2022-12-18
fix(s3): buckets with SSE-KMS silently fail to receive logs

laurelmay:
AWS S3 Server Access Logging does not support logging to buckets that
use SSE-KMS, only to buckets without default encryption or to buckets
that use SSE-S3. At least in some cases, this misconfiguration can be
caught within the CDK (when logging to the same bucket or when the
target bucket is using a KMS CMK).

This will still fail to catch scenarios where the target bucket is using
SSE-KMS using a KMS-managed key because the `encryptionKey` property is
not set on the Bucket in that scenario.

This may be a breaking change for some users; what is currently a mostly
silent misconfiguration will become an error when synthesizing.

----

### All Submissions:

* [X] Have you followed the guidelines in our [Contributing guide?](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md)

### Adding new Construct Runtime Dependencies:

* [ ] This PR adds new construct runtime dependencies following the process described [here](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md/#adding-construct-runtime-dependencies)

### New Features

* [ ] Have you added the new feature to an [integration test](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md)?
	* [ ] Did you use `yarn integ` to deploy the infrastructure and generate the snapshot (i.e. `yarn integ` without `--dry-run`)?

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #24789, opened 2023-03-26
fix(s3): KMS encryption works fine for server access logging target buckets

zdf1230:
KMS encryption works fine for server access logging target buckets with proper permission being setup

There were 2 changes wants to solve the issue that buckets with SSE-KMS silently fail to receive logs.
https://github.com/aws/aws-cdk/pull/23514 & https://github.com/aws/aws-cdk/pull/23385

The changes would let it fail early in build time when SSE-KMS being set for server access logging bucket. But it actually works when user use SSE-KMS with customized encryption key. Just need to add KMS Read Write policy for `logging.s3.amazonaws.com` to the key and add S3 Put Object permission for `logging.s3.amazonaws.com`.

So this change is removing the condition failing for the SSE-KMS with customized encryption key case.

However, it is not possible to know which encryption type for the server access logging bucket, so the only checking can be applied after this change merged is failing when logging to self case using `BucketEncryption.KMS_MANAGED`.

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-ecr/test/integ.basic.js.snapshot/aws-ecr-integ-stack.template.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-ecr/test/integ.basic.js.snapshot/aws-ecr-integ-stack.template.json
@@ -5,7 +5,19 @@
    "Properties": {
     "LifecyclePolicy": {
      "LifecyclePolicyText": "{\"rules\":[{\"rulePriority\":1,\"selection\":{\"tagStatus\":\"any\",\"countType\":\"imageCountMoreThan\",\"countNumber\":5},\"action\":{\"type\":\"expire\"}}]}"
-    }
+    },
+    "RepositoryPolicyText": {
+      "Statement": [
+       {
+        "Action": "ecr:GetDownloadUrlForLayer",
+        "Effect": "Allow",
+        "Principal": {
+         "AWS": "*"
+        }
+       }
+      ],
+      "Version": "2012-10-17"
+     }
    },
    "UpdateReplacePolicy": "Retain",
    "DeletionPolicy": "Retain"
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-ecr/test/integ.basic.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-ecr/test/integ.basic.ts
@@ -1,12 +1,17 @@
 import * as cdk from 'aws-cdk-lib';
 import { IntegTest } from '@aws-cdk/integ-tests-alpha';
 import * as ecr from 'aws-cdk-lib/aws-ecr';
+import * as iam from 'aws-cdk-lib/aws-iam';
 
 const app = new cdk.App();
 const stack = new cdk.Stack(app, 'aws-ecr-integ-stack');
 
 const repo = new ecr.Repository(stack, 'Repo');
 repo.addLifecycleRule({ maxImageCount: 5 });
+repo.addToResourcePolicy(new iam.PolicyStatement({
+  actions: ['ecr:GetDownloadUrlForLayer'],
+  principals: [new iam.AnyPrincipal()],
+}));
 
 new cdk.CfnOutput(stack, 'RepositoryURI', {
   value: repo.repositoryUri,
--- a/packages/aws-cdk-lib/aws-ecr/README.md
+++ b/packages/aws-cdk-lib/aws-ecr/README.md
@@ -126,3 +126,18 @@ const repository = new Repository(this, 'MyTempRepo', {
   autoDeleteImages: true,
 });
 ```
+
+## Managing the Resource Policy
+
+You can add statements to the resource policy of the repository using the
+`addToResourcePolicy` method. However, be advised that you must not include
+a `resources` section in the `PolicyStatement`.
+
+```ts
+declare const repository: ecr.Repository;
+repository.addToResourcePolicy(new iam.PolicyStatement({
+  actions: ['ecr:GetDownloadUrlForLayer'],
+  // resources: ['*'], // not currently allowed!
+  principals: [new iam.AnyPrincipal()],
+}));
+```
--- a/packages/aws-cdk-lib/aws-ecr/lib/repository.ts
+++ b/packages/aws-cdk-lib/aws-ecr/lib/repository.ts
@@ -4,6 +4,7 @@ import * as events from '../../aws-events';
 import * as iam from '../../aws-iam';
 import * as kms from '../../aws-kms';
 import {
+  Annotations,
   ArnFormat,
   IResource,
   Lazy,
@@ -424,7 +425,7 @@ export interface OnCloudTrailImagePushedOptions extends events.OnEventOptions {
  */
 export interface OnImageScanCompletedOptions extends events.OnEventOptions {
   /**
-   * Only watch changes to the image tags spedified.
+   * Only watch changes to the image tags specified.
    * Leave it undefined to watch the full repository.
    *
    * @default - Watch the changes to the repository with all image tags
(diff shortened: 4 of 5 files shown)


zdf1230:
Exemption Request

Integration tests succeeded and this change won't change snapshot.

packages/@aws-cdk/aws-s3/test/bucket.test.ts
```diff
@@ -356,30 +356,30 @@ describe('bucket', () => {
...
   });
 
-  test('logs to self, KMS key with no specific encryption specified throws error', () => {
+  test('logs to self, KMS key with no specific encryption specified does not throw error', () => {
     const stack = new cdk.Stack();
     const key = new kms.Key(stack, 'TestKey');
     expect(() => {
       new s3.Bucket(stack, 'MyBucket', { encryptionKey: key, serverAccessLogsPrefix: 'test' });
-    }).toThrow(/SSE-S3 is the only supported default bucket encryption for Server Access Logging target buckets/);
+    }).not.toThrowError();
```
TheRealAmazonKendra, line 382: These should test against the expected template output, not just that is doesn't throw an error.

packages/@aws-cdk/aws-s3/lib/bucket.ts
```diff
@@ -2105,17 +2105,10 @@ export class Bucket extends BucketBase {
...
-        props.encryption === BucketEncryption.KMS_MANAGED ||
-        props.encryption === BucketEncryption.KMS )) ||
-      // Another bucket is being used that is configured for default SSE-KMS
-      props.serverAccessLogsBucket?.encryptionKey
-    ) {
-      throw new Error('SSE-S3 is the only supported default bucket encryption for Server Access Logging target buckets');
+    // Proper permission can be added to the key and bucket for BucketEncryption.KMS case.
+    // But for BucketEncryption.KMS_MANAGED, the key permission cannot be modified.
+    if (!props.serverAccessLogsBucket && props.encryption === BucketEncryption.KMS_MANAGED) {
+      throw new Error('Default bucket encryption with KMS managed key is not supported for Server Access Logging target buckets');
```
TheRealAmazonKendra, line 2111: If this is the case, we should have integ tests where `encryption` is set to `BucketEncryption.KMS` with an `encryptionKey` and without an `encryptionKey` set.

packages/@aws-cdk/aws-s3/test/bucket.test.ts
```diff
@@ -356,30 +356,30 @@ describe('bucket', () => {
...
+    }).toThrow(/Default bucket encryption with KMS managed key is not supported for Server Access Logging target buckets/);
   });
 
-  test('logs to self, KMS encryption without key throws error', () => {
+  test('logs to self, KMS encryption without key does not throw error', () => {
     const stack = new cdk.Stack();
     expect(() => {
       new s3.Bucket(stack, 'MyBucket', { encryption: s3.BucketEncryption.KMS, serverAccessLogsPrefix: 'test' });
-    }).toThrow(/SSE-S3 is the only supported default bucket encryption for Server Access Logging target buckets/);
+    }).not.toThrowError();
```
TheRealAmazonKendra, line 366: These should test against the expected template output, not just that is doesn't throw an error.

packages/@aws-cdk/aws-s3/test/bucket.test.ts
```diff
@@ -356,30 +356,30 @@ describe('bucket', () => {
...
   });
 
-  test('logs to self, KMS encryption with key throws error', () => {
+  test('logs to self, KMS encryption with key does nor throw error', () => {
     const stack = new cdk.Stack();
     const key = new kms.Key(stack, 'TestKey');
     expect(() => {
       new s3.Bucket(stack, 'MyBucket', { encryptionKey: key, encryption: s3.BucketEncryption.KMS, serverAccessLogsPrefix: 'test' });
-    }).toThrow(/SSE-S3 is the only supported default bucket encryption for Server Access Logging target buckets/);
+    }).not.toThrowError();
```
TheRealAmazonKendra, line 374: These should test against the expected template output, not just that is doesn't throw an error.

packages/@aws-cdk/aws-s3/test/bucket.test.ts
```diff
@@ -399,40 +399,40 @@ describe('bucket', () => {
...
   });
 
-  test('logs to separate bucket, KMS encryption without key throws error', () => {
+  test('logs to separate bucket, KMS encryption without key does not throw error', () => {
     const stack = new cdk.Stack();
     const logBucket = new s3.Bucket(stack, 'testLogBucket', { encryption: s3.BucketEncryption.KMS });
     expect(() => {
       new s3.Bucket(stack, 'MyBucket', { serverAccessLogsBucket: logBucket });
-    }).toThrow(/SSE-S3 is the only supported default bucket encryption for Server Access Logging target buckets/);
+    }).not.toThrowError();
```
TheRealAmazonKendra, line 417: These should test against the expected template output, not just that is doesn't throw an error.

packages/@aws-cdk/aws-s3/test/bucket.test.ts
```diff
@@ -399,40 +399,40 @@ describe('bucket', () => {
...
 
-  test('logs to separate bucket, KMS encryption with key throws error', () => {
+  test('logs to separate bucket, KMS encryption with key does not throw error', () => {
     const stack = new cdk.Stack();
     const key = new kms.Key(stack, 'TestKey');
     const logBucket = new s3.Bucket(stack, 'testLogBucket', { encryptionKey: key, encryption: s3.BucketEncryption.KMS });
     expect(() => {
       new s3.Bucket(stack, 'MyBucket', { serverAccessLogsBucket: logBucket });
-    }).toThrow(/SSE-S3 is the only supported default bucket encryption for Server Access Logging target buckets/);
+    }).not.toThrowError();
```
TheRealAmazonKendra, line 426: These should test against the expected template output, not just that is doesn't throw an error.

packages/@aws-cdk/aws-s3/test/bucket.test.ts
```diff
@@ -399,40 +399,40 @@ describe('bucket', () => {
...
 
-  test('logs to separate bucket, KMS key with no specific encryption specified throws error', () => {
+  test('logs to separate bucket, KMS key with no specific encryption specified does not throw error', () => {
     const stack = new cdk.Stack();
     const key = new kms.Key(stack, 'TestKey');
     const logBucket = new s3.Bucket(stack, 'testLogBucket', { encryptionKey: key });
     expect(() => {
       new s3.Bucket(stack, 'MyBucket', { serverAccessLogsBucket: logBucket });
-    }).toThrow(/SSE-S3 is the only supported default bucket encryption for Server Access Logging target buckets/);
+    }).not.toThrowError();
```
TheRealAmazonKendra, line 435: These should test against the expected template output, not just that is doesn't throw an error.

TheRealAmazonKendra (changes requested):
Thank you for your contribution. Please see my comments inline.

TheRealAmazonKendra:
Removing request for exemption per my comments in my review.

packages/@aws-cdk/aws-s3/lib/bucket.ts
zdf1230: The case you talked about is checked under function `parseEncryption`.
```
    // if encryption key is set, encryption must be set to KMS.
    if (encryptionType !== BucketEncryption.KMS && props.encryptionKey) {
      throw new Error(`encryptionKey is specified, so 'encryption' must be set to KMS (value: ${encryptionType})`);
    }
```

The change I modified here is mean to fail the build when the server access logs bucket being set as itself and the bucket encryption is set to KMS_MANAGED. In this case, though the resource can be successfully deployed, we cannot modified the encryption key owned by KMS. The key and the bucket need proper permission to make the whole logging workflow work. So, people won't able to get access log from the bucket or the log would be unreadable.

This change won't affect snapshot at all.

packages/@aws-cdk/aws-s3/test/bucket.test.ts
zdf1230: The original logic would throw the error for the case which should not be. And none of these changes are affecting template output. They just mean to fail the build earlier to prevent some non-working situation.

My change is just remove some failed checking condition which is actually working case.

As my change won't affect generating template output, I don't think there is anything I can do there.

zdf1230:
Thanks for the comments. I replied them. Please let me know if anything is uncleared. As I commented, I still think I need the Exemption Request.

zdf1230:
Clarification Request

PR #24791, opened 2023-03-26
feat(ecr-assets): Support Docker BuildKit secret from environment

### LINKED ISSUE #24691: (cdk-assets): docker buildkit secrets from environement
### Describe the feature

#23778 Introduced the ability to pass secrets within the [build process thanks to BuildKit](https://docs.docker.com/engine/reference/commandline/buildx_build/#secret). Such feature supports secrets from two souces:
1. **File**: docker will read the file and pass the content as the secret
2. **Env**: docker will read the secret from the Environment and pass it

The PR aboved implemented (1) but not (2)

### Use Case

As a developer,

I want to be able to use the Environment to pass secrets,

so I don't have to create ephimeral files from secrets that I already have in the environment. Specifically in Github Action pipelines.

### Proposed Solution

In line with what [is proposed in here](https://github.com/aws/aws-cdk/issues/14395#issuecomment-1157820199), I suggest to
```ts
/**
 * buildArgs?: {
 *    [key: string]: string | undefined;
 * };
 */

// Such
const asset = new DockerImageAsset(stack, 'blah', { directory, buildSecrets: { mySecret: undefined } });
```

This way we keep the same interface and we take the `undefined` semantic to fetch the secret from the environment.

Spcifically we can modify [this line](https://github.com/aws/aws-cdk/blob/74512fa339e0a2937213f519c109ef1207e9d0c6/packages/cdk-assets/lib/private/docker.ts#L58) such
```ts
// Before
 ...flatten(Object.entries(options.buildSecrets || {}).map(([k, v]) => ['--secret', `id=${k},${v}`])),

// After
...flatten(Object.entries(options.buildSecrets || {}).map(([k, v]) => ['--secret', v===undefined ? `id=${k}`:`id=${k},${v}`])),
```

### Other Information

This could close #14395

### Acknowledgements

### CDK version used

2.69.0

MacOS Ventura 13.2.1

danwiltshire:
Closes #24691

----


packages/@aws-cdk/core/test/bundling.test.ts
```diff
@@ -603,9 +603,12 @@ describe('bundling', () => {
 
   test('ensure correct Docker CLI arguments are returned', () => {
     // GIVEN
+    process.env.TEST_FIXTURE_ENV = 'abc';
     const fromSrc = DockerBuildSecret.fromSrc('path.json');
+    const fromEnvironment = DockerBuildSecret.fromEnvironment('TEST_FIXTURE_ENV');
 
     // THEN
     expect(fromSrc).toEqual('src=path.json');
+    expect(fromEnvironment).toEqual('env=TEST_FIXTURE_ENV');
```
TheRealAmazonKendra, line 612: Could you please add a test case where the env variable isn't correctly set but `fromEnvironment` is called?

packages/@aws-cdk/core/lib/bundling.ts
```diff
@@ -21,6 +21,21 @@ export class DockerBuildSecret {
...
+   * @param environmentVariable The environment variable name
+   * @returns The latter half required for `--secret`
+   */
+  public static fromEnvironment(environmentVariable: string): string {
+    const value = process.env[environmentVariable];
+
+    if (value) {
+      return `env=${environmentVariable}`;
+    } else {
+      throw new Error(`Environment variable ${environmentVariable} is not set`);
```
TheRealAmazonKendra, line 36: At what point does this error message get surfaced? Will it be in a context that this message will be clear enough to a user?

TheRealAmazonKendra (changes requested):
Thank you for your contribution! Please see my comments inline. My only real concern, besides wanting another test is how the error message will be surfaced and if the user will have enough context where it does surface to ensure they know where it's coming from.

danwiltshire:
Closing this PR after the repo refactor, will create new.

PR #24796, opened 2023-03-27
fix(globalaccelerator): parameter name can exceed limit of 64 characters

### LINKED ISSUE #24325: aws-cdk-lib.aws_globalaccelerator: Name attribute not generated according to rules
### Describe the bug

It seems that the Accelerator "Name" attribute is generated incorrectly and cause errors that prevent the creation of the accelerator if the automatically generated name is too long.

### Expected Behavior

I expected the resource to be created successfully and with a compliant name.

### Current Behavior

The error message returned was:

> Resource handler returned message: "Model validation failed (#/Name: 2 schema violations found) #/Name: expected maxLength: 64, actual: 68 (#/Name) #/Name: failed validation constraint for keyword [pattern] (#/Name)" (RequestToken: e4dc2c84-20e3-c18e-d9b1-649ca1fe3a1e, HandlerErrorCode: InvalidRequest)

### Reproduction Steps

I have a stack named `ZpayGlobalSyteCentral-Stack` and an accelerator named `Accelerator`.

```
    const accelerator = new Accelerator(this, 'Accelerator');
```

```
    pipeline.addStage(new ZpayGlobalStage(this, 'NpayGlobalSyteCentral', {
```

The generated unique name is 68 characters long, where as 64 characters is the limit for a GlobalAccelerator name.

### Possible Solution

It seems that the current name is generated by `cdk.Names.uniqueId`:

https://github.com/aws/aws-cdk/blob/539d03684453a73acbd6f1734ea99292b7c6d956/packages/%40aws-cdk/aws-globalaccelerator/lib/accelerator.ts#L90

This is incorrect to generate a resource name, as explained by the documentation. Instead, the generation should happen via `cdk.Names.uniqueResourceName`:

```
      name: props.acceleratorName ?? cdk.Names.uniqueResourceName(this, { maxLength: 64, allowedSpecialCharacters: '.-' }),
```

Above code is not tested and is not a significant contribution from me, so can be used freely and without attribution.

https://i.imgur.com/MS1ysj0.mp4

### CDK CLI Version

2.66.0 (build c96c17d)

### Framework Version

_No response_

### Node.js Version

v19.5.0

### OS

Linux codespaces-af5ba0 5.4.0-1103-azure #109~18.04.1-Ubuntu SMP Wed Jan 25 20:53:00 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Language

Typescript

### Language Version

4.9.5

### Other information

_No response_

lpizzinidev:
The global accelerator should automatically generate a Name with a length smaller than 64 characters if the `acceleratorName` property is not specified, otherwise, it will fail during synth.
If `acceleratorName` is specified, it should validate its length to be in the expected range (1-64 characters).

Closes #24325.

----

### DIFF
--- a/packages/aws-cdk-lib/pipelines/README.md
+++ b/packages/aws-cdk-lib/pipelines/README.md
@@ -1291,6 +1291,15 @@ We therefore expect you to mind the following:
   come with CDK Pipelines will expect `package-lock.json` and `yarn.lock` to
   ensure your dependencies are the ones you expect.
 
+- CDK Pipelines runs on resources created in your own account, and the configuration
+  of those resources is controlled by developers submitting code through the pipeline.
+  Therefore, CDK Pipelines by itself cannot protect against malicious
+  developers trying to bypass compliance checks. If your threat model includes
+  developers writing CDK code, you should have external compliance mechanisms in place like
+  [AWS CloudFormation Hooks](https://aws.amazon.com/blogs/mt/proactively-keep-resources-secure-and-compliant-with-aws-cloudformation-hooks/)
+  (preventive) or [AWS Config](https://aws.amazon.com/config/) (reactive) that
+  the CloudFormation Execution Role does not have permissions to disable.
+
 - Credentials to production environments should be short-lived. After
   bootstrapping and the initial pipeline provisioning, there is no more need for
   developers to have access to any of the account credentials; all further


packages/@aws-cdk/aws-globalaccelerator/lib/accelerator.ts
```diff
@@ -85,9 +86,16 @@ export class Accelerator extends cdk.Resource implements IAccelerator {
   constructor(scope: Construct, id: string, props: AcceleratorProps = {}) {
     super(scope, id);
 
+    const maxResourceNameLength = 64;
+    const name = cdk.FeatureFlags.of(this).isEnabled(cxapi.GLOBAL_ACCELERATOR_SAFE_RESOURCE_NAME) ?
+      cdk.Names.uniqueResourceName(this, {
+        separator: '-',
+        maxLength: maxResourceNameLength,
+        allowedSpecialCharacters: '/_-',
+      }) : props.acceleratorName ?? cdk.Names.uniqueId(this);
```
TheRealAmazonKendra, line 95: We'll still have a problem here if `acceleratorName` is longer than 64 characters. I'm not seeing anywhere that we validate the length of it. We should probably add that to this change.

packages/@aws-cdk/aws-globalaccelerator/lib/accelerator.ts
```diff
@@ -85,9 +86,16 @@ export class Accelerator extends cdk.Resource implements IAccelerator {
   constructor(scope: Construct, id: string, props: AcceleratorProps = {}) {
     super(scope, id);
 
+    const maxResourceNameLength = 64;
+    const name = cdk.FeatureFlags.of(this).isEnabled(cxapi.GLOBAL_ACCELERATOR_SAFE_RESOURCE_NAME) ?
```
TheRealAmazonKendra, line 90: Because this isn't used on a stateful resource, I'm not actually sure we need the feature flag here. Checking in with other team members on this.

TheRealAmazonKendra (changes requested):
Thank you for your contribution! Overall, this looks great. Please see my comments inline.

packages/@aws-cdk/aws-globalaccelerator/lib/accelerator.ts
lpizzinidev: Should I trim the length automatically`(props.acceleratorName ?? cdk.Names.uniqueId(this)).slice(-64);` or check the value and throw an error in case it exceeds the character limit?

packages/@aws-cdk/aws-globalaccelerator/lib/accelerator.ts
lpizzinidev: Ok, let me know. I've added the feature flag because the issue is similar to [this one](https://github.com/aws/aws-cdk/issues/24654), where a feature flag was [requested](https://github.com/aws/aws-cdk/issues/24654#issuecomment-1473826663).

lpizzinidev: I've removed the FF.
Once this has been reviewed and accepted I will align the [other PR](https://github.com/aws/aws-cdk/pull/24786) as well.

TheRealAmazonKendra (changes requested):
Looks like there are a few issues to resolve. Our restructure caused some conflicts but also the build failed due to the integ test needing to be updated.

lpizzinidev:
@TheRealAmazonKendra 
Should I move the integration test to the `packages/@aws-cdk-testing/framework-integ` folder?
Also, the documentation has not been updated to reflect the last changes.
Can you explain how to run the integration test with the new project structure?

TheRealAmazonKendra (approved):

PR #24812, opened 2023-03-28
fix(ec2): VPC Flow Log record fields are not available

### LINKED ISSUE #24807: (ec2): VPC Flow Log `interface-id` format is incorrect
### Describe the bug

In the `LogFormat` enum-like class, `INTERFACE_ID` has an invalid value.

https://github.com/aws/aws-cdk/blob/c13a0f1a3820c1b8622c3de687ae52834f8ea614/packages/%40aws-cdk/aws-ec2/lib/vpc-flow-logs.ts#L417-L420

### Expected Behavior

When specifying `ec2.LogFormat.INTERFACE_ID`, the stack should create a valid VPC Flow Logs configuration. `${interface-id` should be `${interface-id}`.

### Current Behavior

The CloudFormation stack fails to create because the format string is incorrect.

### Reproduction Steps

```ts
    const vpc = new ec2.Vpc(this, 'Vpc', {
      ipAddresses: ec2.IpAddresses.cidr('10.10.10.0/24'),
      flowLogs: {
        vpc: {
          destination: ec2.FlowLogDestination.toCloudWatchLogs(),
          logFormat: [
            ec2.LogFormat.INTERFACE_ID,
          ],
        },
      },
    });
```

### Possible Solution

All the constants should be defined using the `ec2.LogFormat.field` static method instead of the constructor; this will ensure the formatting is correct (and a `}` isn't missing).

Workaround: use `ec2.LogFormat.field('interface-id')` instead.

### CDK CLI Version

2.70.0

### Framework Version

_No response_

### Node.js Version

16

### OS

Linux

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

laurelmay:
There are a lot of VPC Flow Log record fields that are not currently
available as named attributes (from various versions). [Recently][1],
CloudWatch Logs got support for all existing versions of log record
fields. This makes sure [all][2] those available.

This also corrects an issue with `interface-id` which made it impossible
to create logs for CloudWatch Logs since it was missing the closing
brace. This is done by making use of the `field` static method for nearly
all existing fields that are just a single field. Hopefully this prevents
errors in the future.

[1]: https://aws.amazon.com/about-aws/whats-new/2023/03/amazon-cloudwatch-logs-amazon-vpc-flow-logs-metadata/
[2]: https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs.html#flow-logs-fields

Closes #24807.

----


TheRealAmazonKendra (approved):
Looks good! Thanks!

PR #24830, opened 2023-03-28
chore(eks): Add missing ami types for node groups

### LINKED ISSUE #24803: aws-eks: Add missing ami types for node groups
### Describe the feature

The list of AMI types provided in CDK is different than those offered in Cloudformation - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-eks-nodegroup.html#cfn-eks-nodegroup-amitype

### Use Case

I saw this issue in the aws-cdk slack channel, but the user is trying to use windows amis - `WINDOWS_CORE_2019_x86_64 | WINDOWS_CORE_2022_x86_64 | WINDOWS_FULL_2019_x86_64 | WINDOWS_FULL_2022_x86_64`

These are supported via Cloudformation but missing in cdk since the Enum is manually defined.

### Proposed Solution

Updated the enum to include these values

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.70.0

Mac

cgamache:
Adds Windows amiTypes in aws-eks. These are supported via Cloudformation but missing in cdk since the Enum is manually defined.

Closes #24803

----

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/lib/staging/maven.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/staging/maven.ts
@@ -1,5 +1,6 @@
 /* eslint-disable no-console */
 import * as path from 'path';
+import { pathExists } from 'fs-extra';
 import { LoginInformation } from './codeartifact';
 import { parallelShell } from './parallel-shell';
 import { UsageDir } from './usage-dir';
@@ -29,6 +30,9 @@ export async function uploadJavaPackages(packages: string[], login: LoginInforma
   await parallelShell(packages, async (pkg, output) => {
     console.log(`⏳ ${pkg}`);
 
+    const sourcesFile = pkg.replace(/.pom$/, '-sources.jar');
+    const javadocFile = pkg.replace(/.pom$/, '-javadoc.jar');
+
     await shell(['mvn',
       `--settings=${settingsFile(usageDir)}`,
       'org.apache.maven.plugins:maven-deploy-plugin:3.0.0:deploy-file',
@@ -36,8 +40,8 @@ export async function uploadJavaPackages(packages: string[], login: LoginInforma
       '-DrepositoryId=codeartifact',
       `-DpomFile=${pkg}`,
       `-Dfile=${pkg.replace(/.pom$/, '.jar')}`,
-      `-Dsources=${pkg.replace(/.pom$/, '-sources.jar')}`,
-      `-Djavadoc=${pkg.replace(/.pom$/, '-javadoc.jar')}`], {
+      ...await pathExists(sourcesFile) ? [`-Dsources=${sourcesFile}`] : [],
+      ...await pathExists(javadocFile) ? [`-Djavadoc=${javadocFile}`] : []], {
       output,
       modEnv: {
         // Do not try to JIT the Maven binary


cgamache:
I am not able to run integration tests. Please advise how the README should change.

TheRealAmazonKendra:
> Clarification Request: I am not able to run integration tests. Please advise how the README should change.

This should actually be a `chore` not a `feat`. Those requirements will disappear.

cgamache:
Hi @TheRealAmazonKendra ! ... Thanks for the tip from before. The build passed, and it now only lacks an approving review.

packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts
```diff
@@ -479,6 +508,21 @@ function isGpuInstanceType(instanceType: InstanceType): boolean {
...
+ * https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html
+ * @param instanceType The EC2 instance type
+ */
+function isWindowsSupportedInstanceType(instanceType: InstanceType): boolean {
+  // capture the family, generation, capabilities, and size portions of the instance type id
+  const instanceTypeFormat = instanceType.toString().match(/^[a-z]+\d{1,2}[a-z]*\.[a-z0-9]+$/);
+  if (instanceTypeFormat == null) {
+    throw new Error('Malformed instance type identifier');
+  }
+  const forbiddenInstanceTypes:RegExp = /^(?:c3|c4|d2|i2|m4|m6a|r3)\.[a-z0-9]+$/;
```
otaviomacedo, line 522: This kind of thing breaks the encapsulation that `InstanceType` is supposed to provide. I think it's better to move this logic inside `InstanceType` somehow. An idea off the cuff:

```ts
public sameInstanceClassAs(other: InstanceType): boolean {
  // parse this.instanceTypeIdentifier and other.instanceTypeIdentifier 
  // and compare the parts before the '.'
}
```

but you may come up with a more elegant solution along these lines.

cgamache: This function was modeled after `function isGpuInstanceType(instanceType: InstanceType): boolean` ...  

Expanding the capability of InstanceTypes with a function like 

```TypeScript
    public sameInstanceClassAs(other: InstanceType): boolean {
        const instanceClass: RegExp = /^([a-z]+\d{1,2}[a-z]*)\.([a-z0-9]+)$/;
        const instanceClassId = this.instanceTypeIdentifier.match(instanceClass);
        const otherInstanceClassId = other.toString().match(instanceClass);
        if (instanceClassId == null || otherInstanceClassId == null) {
          throw new Error('Malformed instance type identifier');
        }
        return instanceClassId[1] === otherInstanceClassId[1];
    }
```

and co-opting the new method is a good idea. It does not help solve the immediate problem of the lack of support for Windows Containers in CDK for EKS. Perhaps we can merge this change and then create another issue and PR to clean up the instance type checks?

cgamache: The purpose of this pull request is to address issue https://github.com/aws/aws-cdk/issues/24803 which deals strictly with adding values to NodegroupAmiType that are missing in CDK but supported in CloudFormation. I set up some guard rails to ensure the users of CDK would not inadvertently select an unsupported instance type for their Windows NodeGroups. The guard rails function present in the PR mimics the preexisting guard rails function for GPU-based NodeGroups. The proposed patch shouldn't be faulted for re-using a pattern already present in the codebase to solve an immediate problem.

I am not denying that those functions could be improved. It is a good thing you suggest.

However, I propose that this PR be merged as-is, and a separate issue for cleaning up the instance type checking routines be submitted.

Each PR would represent an atomic concern. This is a good practice.

TheRealAmazonKendra (changes requested):
I'm in agreement with @otaviomacedo here about moving the logic inside InstanceType. While we may have another location that does what you've done here, it's not a great design choice and we should have caught it in that instance as well. 

On second glance at this, we actually have more complexity in the logic about which ami's work where so my original comment was wrong in thinking that we didn't need integ tests here. Apologies for not looking closer in the first place and providing misleading information.

cgamache:
PR updated with requested changes, @TheRealAmazonKendra  ... Please let me know if there is anything that I missed.

🤞 this should be ready for integration tests.

otaviomacedo:
Thanks, @cgamache. Please go ahead and write the integration test cases, so we can generate the snapshots for you.

otaviomacedo (dismissed):
Thanks, @cgamache. Please go ahead and write the integration test cases, so we can generate the snapshots for you.

cgamache:
@otaviomacedo @TheRealAmazonKendra Integration test written and added to the PR.

Bad news though-- the build is now failing. I'm not sure if the un-run integration test is the culprit. It builds and unit tests locally without failures.

Nothing looks out of place in the build logs until the failure at the end. And the failure message isn't specific as to what has gone wrong. I do see that it noticed the "NEW" integration test.

Please advise. Thanks very much!

otaviomacedo (approved):

PR #24834, opened 2023-03-28
chore: dont close PRs with exemption request

laurelmay:
Users that have made an exemption request are likely waiting for feedback from the core team after `pr-linter` requested changes.

----


kylelaker:
For an example, https://github.com/aws/aws-cdk/pull/24486#issuecomment-1487520799 and [the run that generated](https://github.com/aws/aws-cdk/actions/runs/4547126159/jobs/8016620812#step:2:269) that message based off requested changes.

TheRealAmazonKendra:
As a general policy we can't let user comments directly impact our workflows. I am, on the other hand, working my way though our backlog of PRs today so I'll be taking a look at your PR.

kylelaker:
Oh yeah, that makes sense. I guess since the label causes the check to be entirely bypassed even if an important check is failing and since it can be re-applied, this is sorta suboptimal.

PR #24855, opened 2023-03-29
docs(servicecatalogappregistry): update README.md to track latest change

liwewang-amazon:
This commit updates the `README.md` to be in sync with the latest change. Deprecated methods are removed, new methods are added.

----

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,7 +100,7 @@ time, you can execute the following to build it and it's dependencies.
 
 ```console
 $ cd packages/aws-cdk-lib
-$ ../../../scripts/buildup
+$ ../../scripts/buildup
 ```
 
 Note: The `buildup` command is resumable. If your build fails, you can fix the issue and run `buildup --resume` to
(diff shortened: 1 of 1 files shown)


packages/@aws-cdk/aws-servicecatalogappregistry/README.md
```diff
@@ -204,10 +217,15 @@ const associatedApp = new appreg.ApplicationAssociator(app, 'AssociatedApplicati
 });
 ```
 
+### Associate cross-region stack
```
yingdon-amazon, line 220: Just curious, if we do not support this feature, do we need to mention it here?

liwewang-amazon: I think it is needed since customer cdk app may contain cross region stacks, and introduction of ApplicationAssociator could not work as expected. This gives the clear expectation.

yingdon-amazon: make sense, explaining clearly is always good

TheRealAmazonKendra (approved):
Thanks for the update!

PR #24856, opened 2023-03-29
feat(stepfunctions-tasks): add elasticmapreduce:AddTags permission for EmrCreateCluster state with tags

### LINKED ISSUE #24842: aws-stepfunctions-tasks: Additional elasticmapreduce:AddTags permission required to launch EMR clusters with tags from April 24th onwards
### Describe the bug

We received the following notice from Amazon EMR service via AWS Health Dashboard:

> Amazon Elastic Map Reduce (EMR) will soon require permission for the “elasticmapreduce:AddTags” IAM action, which supports tagging when creating an Amazon EMR cluster. Beginning April 24, 2023, this permission will be required in addition to permission for the “elasticmapreduce:RunJobFlow” IAM action, in order to create a cluster with tags.

We are currently using AWS CDK aws-stepfunctions-tasks module EmrCreateCluster step / construct to create EMR clusters as part of Step Functions based EMR workflows. As per https://github.com/aws/aws-cdk/blob/40590d5fdc9f01c9fed553f31f788b7cd9aad1fe/packages/%40aws-cdk/aws-stepfunctions-tasks/lib/emr/emr-create-cluster.ts#L301, that construct does not grant the state machine role the soon to be required elasticmapreduce:AddTags permission in case the cluster is created with tags.

### Expected Behavior

elasticmapreduce:AddTags permission is added to the state machine role if EmrCreateCluster step with tags is added to a state machine

### Current Behavior

elasticmapreduce:AddTags permission is not added to the state machine role if EmrCreateCluster step with tags is added to a state machine

### Reproduction Steps

```
import { StateMachine } from "aws-cdk-lib/aws-stepfunctions";
import { EmrCreateCluster } from "aws-cdk-lib/aws-stepfunctions-tasks";

new StateMachine(this, "StateMachine", {
  definition: new EmrCreateCluster(this, "Cluster", {
    name: "test",
    instances: {},
    tags: {
      "key": "value"
    }
  })
})
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.70.0 (build c13a0f1)

### Framework Version

_No response_

### Node.js Version

v14.21.3

### OS

Ubuntu 20.04

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

sjakthol:
Beginning April 24, 2023, Amazon Elastic Map Reduce (EMR) will start to require permission for the "elasticmapreduce:AddTags" IAM action when creating an EMR cluster with tags. This commit updates the EmrCreateCluster state to add elasticmapreduce:AddTags action to the state machine execution role when tags have been defined.

Closes #24842.

----


packages/@aws-cdk/aws-stepfunctions-tasks/lib/emr/emr-create-cluster.ts
```diff
@@ -319,6 +319,16 @@ export class EmrCreateCluster extends sfn.TaskStateBase {
       );
     }
 
+    // Allow the StateMachine to tag clusters if tags are defined
+    if (this.props.tags) {
+      policyStatements.push(
+        new iam.PolicyStatement({
+          actions: ['elasticmapreduce:AddTags'],
+          resources: ['*'],
```
TheRealAmazonKendra, line 327: Is it possible to scope this down further?

TheRealAmazonKendra (dismissed):
Thank you for your contribution! Please see my comments inline. Overall this looks great!

packages/@aws-cdk/aws-stepfunctions-tasks/lib/emr/emr-create-cluster.ts
sjakthol: As per https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonelasticmapreduce.html, AddTags action supports four kinds of EMR resources. Clusters are the only ones this state needs to be able to tag. The resource could be limited to all cluster resources (wildcard still required on the clusterid part of the ARN).

As for conditions, the action seems to support the usual RequestTag style condition keys. We could, in theory, add a condition to this statement that limits the tag keys and their values to the values specified in the tags prop. However, this gets complex pretty quickly since e.g. tag values can be dynamic jsonpath references to fields in the state object and tag keys can (maybe) even be CDK tokens. Constructing RequestTag conditions that support all these edge cases tricky very fast. Plus, if the user is allowed to use this CDK construct and create a role for the state machine, they could anyway specify any tags they wish in the props and hence the conditions generated based on them would allow any tags chosen by the user.  

A condition I would maybe like to see here is something that would allow tagging to be limited to happen during RunJobFlow call only. EC2 has ec2:CreateAction condition key which can be used to do that. With this kind of limitation you would not be able to go around tagging existing clusters if you happen to have access to execute a state machine that accepts tag keys / values from the state object. However, EMR does not seem to support such condition so that cannot be done.

Anyway, I'll change the resource to only allow tagging on clusters and leave out the additional conditions due to complications explained above. Thanks for the comment!

packages/@aws-cdk/aws-stepfunctions-tasks/test/emr/integ.emr-create-cluster-with-tags.ts
```diff
@@ -0,0 +1,29 @@
...
+
+new StateMachine(stack, 'SM', {
+  definition: step,
+});
+
+new IntegTest(app, 'EmrCreateClusterTest', {
+  testCases: [stack],
+});
+
+app.synth();
```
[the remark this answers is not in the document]
sjakthol: Thanks. Removed in latest followup commit.

sjakthol:
Rebased on top of latest changes to main branch. Unfortunately I was not able to retain the commit history I had in this PR but had to recreate the changes in fresh commit on top of main branch instead. Comparing to previous state of this PR, the rebase included the following changes:
* Reapplied changes to aws-stepfunctions-tasks/lib/emr/emr-create-cluster.ts in the new location of this file
* Moved integration test to the new integration test package
No changes in content of the PR was made in this process.

corymhall (dismissed):

corymhall (approved):

PR #24871, opened 2023-03-30
feat(cloudfront): add types for cors http methods in response headers policy

cchanche:
### Motivation

When creating a custom response header policy, it is unclear what is supposed to be inserted **exactly** in the `corsBehavior.accessControlAllowMethods`.

The [cloudformation docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-responseheaderspolicy-accesscontrolallowmethods.html#properties) is very clear on the subject : 

> The list of HTTP methods. Valid values are:
> - `GET`
> - `DELETE`
> - `HEAD`
> - `OPTIONS`
> - `PATCH`
> - `POST`
> - `PUT`
> - `ALL`
>
> `ALL` is a special value that includes all of the listed HTTP methods.

Other related properties (`accessControlAllowHeaders` and `accessControlAllowOrigins`) are compatible with the wildcard pattern (`['*']`), and one could easily assume the same for `accessControlAllowMethods`.

Specifying a wildcard for `accessControlAllowMethods` would compile fine since the current type is `string[]`, but cloudformation fails to deploy the generated template.

### Fixes and improvents

- new `enum` : `HttpMethods`. The base `enum` was randomly stolen from `@aws-cdk/aws-stepfunctions-tasks/lib/eks/call.ts`, and completed with the `ALL` and `OPTIONS` values and docs.

- Replace the type of `ResponseHeadersCorsBehavior['accessControlAllowMethods']` (`string[]` -> `HttpMethods[]`)

BREAKING CHANGE: Type of ResponseHeadersCorsBehavior['accessControlAllowMethods'] is now an enum
* **@aws-cdk/aws-cloudfront:** Type ResponseHeadersCorsBehavior['accessControlAllowMethods'] as an enum


cchanche:
Oh I see now these are actual breaking changes. I read through the [contributing guide section](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#breaking-changes) again, and I am not sure how to deal with this other than flagging PR for what it is.

PR #24881, opened 2023-03-30
fix(bootstrap): Add explicit policy version to bootstrap template

seanmcl:
It's good to be explicit about the version. The default is 2008 which doesn't support policy variables. If we leave it off and accidentally include one we will silently be able to create one but the policy will not work.


MrArnoldPalmer:
@seanmcl so reading docs I don't see how this could break existing users, changing the version doesn't require replacement of the BucketPolicy resource. Just wanna double check there isn't some sharp corner I'm not thinking of.

PR #24884, opened 2023-03-31
feat(ec2): add support for al2023

### LINKED ISSUE #24873: aws-ec2: Support Amazon Linux 2023
### Describe the feature

Amazon Linux 2023 has been released on March 15, 2023: https://aws.amazon.com/about-aws/whats-new/2023/03/amazon-linux-2023/
Currently, CDK supports Amazon Linux 2022 which was in preview and not a thing anymore.
We should support Amazon Linux 2023 as both latest Amazon Linux version and specifically.

### Use Case

Create an EC2 instance with AL2023

### Proposed Solution

Add AL2023 to the enumeration of supported Amazon Linux versions.
Use the corresponding AMIs to provision the EC2 instance.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.72.0

macOS 13.3

lpizzinidev:
Add support for the new [Amazon Linux 2023](https://aws.amazon.com/it/about-aws/whats-new/2023/03/amazon-linux-2023/) OS for EC2.

Closes #24873.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ec2/README.md
+++ b/packages/aws-cdk-lib/aws-ec2/README.md
@@ -1060,7 +1060,7 @@ new ec2.Instance(this, 'Instance4', {
   vpc,
   instanceType,
   machineImage: ec2.MachineImage.latestAmazonLinux({
-    generation: ec2.AmazonLinuxGeneration.AMAZON_LINUX_2022,
+    generation: ec2.AmazonLinuxGeneration.AMAZON_LINUX_2023,
   }),
 });
 


lpizzinidev:
Exemption Request.

I don't think that this requires an integration test.
However, let me know if adding or editing existing tests is necessary.

packages/@aws-cdk/aws-ec2/README.md
```diff
@@ -1074,6 +1074,15 @@ new ec2.Instance(this, 'Instance4', {
...
 
+// AWS Linux 2023
+new ec2.Instance(this, 'Instance4', {
+  vpc,
+  instanceType,
+  machineImage: ec2.MachineImage.latestAmazonLinux({
+    generation: ec2.AmazonLinuxGeneration.AMAZON_LINUX_2023,
+  }),
+});
+
```
WinterYukky, line 1085: Please remove the example for "AWS Linux 2022" as it is unnecessary, and update the comment for "AWS Linux 2023" to say "Amazon Linux 2023" instead.

packages/@aws-cdk/aws-ec2/lib/machine-image.ts
```diff
@@ -483,6 +484,11 @@ export enum AmazonLinuxGeneration {
    * Amazon Linux 2022
    */
   AMAZON_LINUX_2022 = 'al2022',
```
WinterYukky, line 486: Please deprecate the use of AMAZON_LINUX_2022.

WinterYukky (changes requested):
The added code looks good, but I think it would be necessary to mark Amazon Linux 2022 as deprecated.

packages/@aws-cdk/aws-ec2/lib/machine-image.ts
```diff
@@ -416,13 +416,14 @@ export class AmazonLinuxImage extends GenericSSMParameterImage {
     let virtualization: AmazonLinuxVirt | undefined;
     let storage: AmazonLinuxStorage | undefined;
 
-    if (generation === AmazonLinuxGeneration.AMAZON_LINUX_2022) {
+    if (generation === AmazonLinuxGeneration.AMAZON_LINUX_2022 || generation === AmazonLinuxGeneration.AMAZON_LINUX_2023) {
       kernel = AmazonLinuxKernel.KERNEL5_X;
```
sktan, line 420: You'll find that when this is run and deployed, the SSM parameter won't actually resolve due to Amazon Linux 2023 using Kernel 6.1 rather than Kernel 5.10 like Amazon LInux 2022 is currently coded for.

But also, Amazon Linux 2022 will not deploy via this mechanism, during the preview -> general availability phase, the kernel version was bumped from 5.10 to 5.15 (the SSM param for 5.10 does not exist).

There was an attempt to see if we can re-align it along with introducing 2023, which seems to have opened a bigger can of worms.

https://github.com/aws/aws-cdk/pull/24690

lpizzinidev: Thanks for the feedback.

I guess that this PR should be closed or kept on hold until [this](https://github.com/aws/aws-cdk/issues/21011#issuecomment-1372170948) is implemented.

Unless an alignment of `KERNEL5_X` to `5.15` and the creation of `KERNEL6_X` for `Amazon Linux 2023` can be an acceptable workaround for now.

MrArnoldPalmer:
@lpizzinidev I think we should close this until we implement one of the solutions in the [comment you linked](https://github.com/aws/aws-cdk/issues/21011#issuecomment-1372170948). We have had a few of these come through so I'll talk with the team about bumping this up on our priority list. If you're interested in taking a crack at something along those lines definitely open a PR and see how if feels to you.

PR #24896, opened 2023-03-31
chore: upgrade lerna and use nx

corymhall:
Updating lerna to the latest version which now uses `nx` under the hood. Apart from just updating the version, this also allows us to take advantage of nx caching. For now this has only been enabled locally.

I'm not 100% sure about the `nx` configuration since our repo is a little weird, but I've been running locally with this configuration for probably 6 months and it has worked fine for me.

Side note - the repo restructure has caused some weirdness with our tests. I had to make some updates to existing tests, and I do not know why.

Closes #<issue number here>.

----


lerna.json
```diff
@@ -1,18 +1,29 @@
 {
-  "lerna": "^4.0.0",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "packages": [
```
corymhall, line 4: For some reason if you don't specify each package then lerna can't figure it out. I _think_ it might be because we have packages with the same name in different locations.

corymhall:
I ran this through the test pipeline and the build passed.

MrArnoldPalmer (approved):

comcalvi (approved):

PR #24911, opened 2023-04-03

p-mercury:
Closes https://github.com/aws/aws-cdk/issues/24548

I added and implemented the BundlingOutput.FILE type which allows bundled assets to produce single files as output, which in useful for targets which can not parse directories.


packages/@aws-cdk/aws-appsync/.gitignore
MrArnoldPalmer, line 1: Looks like a leftover before we moved everything to `aws-cdk-lib`.

MrArnoldPalmer (changes requested):
Just an extra file but otherwise I think this makes sense to me.

packages/@aws-cdk/aws-appsync/.gitignore
p-mercury: Yes, I missed the move, I fixed all the issues caused by it and removed the leftover ignore file.

PR #24916, opened 2023-04-03
feat: upgrade default CR runtime version

MrArnoldPalmer:
Adds a regional fact for the default custom resource NodeJS runtime version. Currently for 'aws' and 'aws-cn' partitions uses NodeJS 16.x and NodeJS 14.x for all other partitions. This should allow continued usage of our custom resources in gov cloud and other non-commercial regions while avoiding usage of Node14 which is EOL for others.

----


packages/aws-cdk-lib/region-info/build-tools/generate-static-data.ts
```diff
@@ -73,6 +74,8 @@ export async function main(): Promise<void> {
 
     registerFact(region, 'APPMESH_ECR_ACCOUNT', APPMESH_ECR_ACCOUNTS[region]);
 
+    registerFact(region, 'DEFAULT_CR_NODE_VERSION', ['aws', 'aws-cn'].includes(partition) ? Runtime.NODEJS_16_X.toString() : Runtime.NODEJS_14_X.toString());
```
rix0rrr, line 77: To make this slightly easier to manage in the future, can we do a map of `partition` -> `runtime` ?

packages/aws-cdk-lib/region-info/build-tools/generate-static-data.ts
```diff
@@ -12,6 +12,7 @@ import {
   RULE_CLASSIC_PARTITION_BECOMES_OPT_IN,
 } from '../lib/aws-entities';
 import { Default } from '../lib/default';
+import { Runtime } from '../../aws-lambda';
```
rix0rrr, line 15: Here's a risk of the current repo layout: dependencies that shouldn't be.

`region-info` is a fundamental library and shouldn't depend on construct libraries. Can you hardcode the strings here?

And also: can you think of a mechanism to help us prevent these kinds of circular dependencies? I've been on a project where nothing was enforced and the dependencies were the most goddamn exquisite plate of spaghetti you've ever seen, and when we finally wanted to pick stuff apart it was extremely hard.

packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/aws-custom-resource.ts
```diff
@@ -405,11 +406,18 @@ export class AwsCustomResource extends Construct implements iam.IGrantable {
 
     this.props = props;
 
+    // Runtime regional fact should always return a known runtime string that lambda.Runtime
+    // can index off, but for type safety we also default it here.
+    const runtimeName = cdk.Stack.of(scope).regionalFact(FactName.DEFAULT_CR_NODE_VERSION);
+    const runtime = runtimeName
+      ? new lambda.Runtime(runtimeName, lambda.RuntimeFamily.NODEJS, { supportsInlineCode: true })
+      : lambda.Runtime.NODEJS_16_X;
```
rix0rrr, line 414: Yes, this. But then also for all other custom resources that we have, right?

So the route53 resolver and the s3 bucket deleter etc.?

rix0rrr (changes requested):

packages/aws-cdk-lib/region-info/build-tools/generate-static-data.ts
MrArnoldPalmer: since region-info is a duplicated package, maybe we can enforce it during build of the @aws-cdk/region-info version (which is copied from here). The linter will enforce that all modules that it imports are it its dependencies and we can add a lint rule for those duplicated packages that they can't take any internal dependencies. Meaning nothing within our repository?

Semi related as the new build mechanism for the @aws-cdk/region-info version https://github.com/aws/aws-cdk/pull/24929

packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/aws-custom-resource.ts
MrArnoldPalmer: I got some but I think there are still more. Will do another pass.

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
```diff
@@ -783,7 +782,7 @@ export class Repository extends RepositoryBase {
     // images in the repository and the ability to get all repositories to find the arn needed on delete.
     const provider = CustomResourceProvider.getOrCreateProvider(this, AUTO_DELETE_IMAGES_RESOURCE_TYPE, {
       codeDirectory: path.join(__dirname, 'auto-delete-images-handler'),
-      runtime: CustomResourceProviderRuntime.NODEJS_14_X,
+      runtime: CustomResourceProvider.regionalDefaultRuntime(this),
```
rix0rrr, line 785: I think the term "regional default" leaves open too many unknowns.

- What language is it? (Get Node in there somewhere)
- Who is it for? (This is definitely just for us, people can take a dependency on it, but they shouldn't be scandalized if and when we change it without notice)


`builtInCustomResourceNodeRuntime` ? Bit long...

packages/@aws-cdk/aws-redshift/lib/private/database-query.ts
```diff
@@ -34,7 +35,7 @@ export class DatabaseQuery<HandlerProps> extends Construct implements iam.IGrant
       code: lambda.Code.fromAsset(path.join(__dirname, 'database-query-provider'), {
         exclude: ['*.ts'],
       }),
-      runtime: lambda.Runtime.NODEJS_14_X,
+      runtime: AwsCustomResource.regionalDefaultRuntime(this),
```
rix0rrr, line 38: Is it on `AwsCustomResource` or on `CustomResourceProvider` ?

Honestly, I'm fine with you sticking it somewhere in `core` or `lambda`. Maybe as a global function, so it doesn't appear in jsii?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
```diff
@@ -19,7 +20,7 @@ export interface EvaluateExpressionProps extends sfn.TaskStateBaseProps {
   /**
    * The runtime language to use to evaluate the expression.
    *
-   * @default lambda.Runtime.NODEJS_14_X
+   * @default lambda.Runtime.NODEJS_16_X
```
rix0rrr, line 23: Better not promise this, eh :)

packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
```diff
@@ -100,16 +100,35 @@ export const EBS_ENV_ENDPOINT_HOSTED_ZONE_IDS: { [region: string]: string } = {
...
+// Compute this from our Partition list so new ones are added automatically
+export const CR_DEFAULT_RUNTIME_MAP: { [partition: string]: string } = Object.values(Partition)
+  .reduce((accum, partition) => {
+    const hasNode16 = [Partition.Default, Partition.Cn, Partition.UsGov].includes(partition);
+
+    return {
+      ...accum,
+      [partition]: hasNode16 ?  'nodejs16.x' : 'nodejs14.x',
+    };
+  }, {});
```
rix0rrr, line 130: Honestly I'd prefer

```
const RUNTIME_MAP: Record<Partition, string> = {
  [Partition.Default]: 'node16.x',
  [Partition.Cn]: 'node16.x',
  [Partition.UsGov]: 'node16.x',
  [Partition.UsIso]: 'node14.x',
  [Partition.UsIsoB]: 'node14.x',
};
```

Yes there's some repetition, but it's also eminently readable.

packages/@aws-cdk/aws-redshift/lib/private/database-query.ts
MrArnoldPalmer: So both, because there are two different types here unfortunately. We wrap the runtime strings in a separate type for CustomResourceProviderRuntime 😞

alexherriott:
Hi,

Do you have an estimated Date for when this change will be available in a CDK release?

Thanks

packages/aws-cdk-lib/region-info/build-tools/generate-static-data.ts
MrArnoldPalmer: removed this dependency for now though.

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
MrArnoldPalmer: changed to `builtInCustomResourceNodeRuntime` and `builtInCustomResourceProviderNodeRuntime` respectively.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
MrArnoldPalmer: yeah made this more explicit that it defaults to the region's default? Kinda vague still. I guess the default default is node16?

packages/@aws-cdk/aws-amplify-alpha/lib/branch.ts
```diff
@@ -237,7 +237,7 @@ class AmplifyAssetDeploymentProvider extends NestedStack {
           __dirname,
           'asset-deployment-handler/index.js',
         ),
-        runtime: lambda.Runtime.NODEJS_14_X,
+        runtime: lambda.Runtime.NODEJS_16_X,
```
rix0rrr, line 240: Is this safe? Does Amplify not exist in GovCloud?

Shouldn't this be the same new floating runtime?

packages/aws-cdk-lib/aws-lambda/lib/runtime.ts
```diff
@@ -284,4 +284,9 @@ export class Runtime {
       other.family === this.family &&
       other.supportsInlineCode === this.supportsInlineCode;
   }
+
+  // Used to construct default custom resource runtime by region
+  public static fromRegionDefaultString(name: string): Runtime {
```
rix0rrr, line 289: Should have "node" in the name probably

rix0rrr (approved):

packages/aws-cdk-lib/aws-lambda/lib/runtime.ts
MrArnoldPalmer: removing this altogether actually since it's only used in one place noe.

PR #20564, opened 2022-06-01
chore: add `eslint` rule to forbid `instanceof`

rix0rrr:
----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #24920, opened 2023-04-04
fix(core): allow override with cross-stack references

### LINKED ISSUE #18882: PropertyOverrides: Tokens not properly resolved in cross stack scenarios
### What is the problem?

`Token`s that a are passed from one stack into another are not properly resolved in methods like `CfnResource.addPropertyOverride`.

### Reproduction Steps

```typescript
import * as cdk from "aws-cdk-lib";
import * as ec2 from "aws-cdk-lib/aws-ec2";
import * as lambda from "aws-cdk-lib/aws-lambda";

const app = new cdk.App();

const stackA = new cdk.Stack(app, "StackA");
const vpc = new ec2.Vpc(stackA, "Vpc");

const stackB = new cdk.Stack(app, "StackB");
const fn = new lambda.Function(stackB, "Lambda", {
    runtime: lambda.Runtime.NODEJS_14_X,
    handler: "index.js",
    code: lambda.Code.fromInline("hello"),
});
const cfnFn = (fn.node.defaultChild as cdk.CfnResource);
cfnFn.addPropertyOverride("VpcConfig", {
    subnetIds: vpc.selectSubnets({subnetType: ec2.SubnetType.PUBLIC}).subnetIds,
});
```

The generated `StackB` should have an `AWS::Lambda::Function` with this `VpcConfig`:
```json
...
"VpcConfig": {
  "subnetIds": [
    {
      "Fn::ImportValue": "StackA:ExportsOutputRefVpcPrivateSubnet1Subnet536B997AFD4CC940"
    },
    {
      "Fn::ImportValue": "StackA:ExportsOutputRefVpcPrivateSubnet2Subnet3788AAA1380949A3"
    }
  ]
}
...
```

### What actually happened?

```json
...
"VpcConfig": {
  "subnetIds": [
    {
      "Ref": "VpcPrivateSubnet1Subnet536B997A"
    },
    {
      "Ref": "VpcPrivateSubnet2Subnet3788AAA1"
    }
  ]
}
...
```

### CDK CLI Version

2.10.0 (build e5b301f)

### Framework Version

_No response_

### Node.js Version

v14.18.3

### OS

Linux

### Language

Typescript

### Language Version

_No response_

### Other information

There must be a simple workaround, because if one uses the selected subnets in `StackB` as properties for another resource inside the same Stack, all `Token`s, including the ones of the `addPropertyOverride`, get resolved properly.

tmokmss:
closes #18882

The problem is described in the original issue, and below is what I found as the root cause and how it can be fixed.

---

Previously when we used a cross-stack reference in override, it was not resolved as an `Fn::ImportValue`.

To make it an `Fn::ImportValue`, we need to get every token in an app and find _references_ (tokens that references resources outside of its stack) from them. The related code is here:

https://github.com/aws/aws-cdk/blob/810d736a8d20638e778c5773507f0edb12733a49/packages/%40aws-cdk/core/lib/private/refs.ts#L139-L140

To get all the tokens in an app, we use `RememberingTokenResolver`, which _remembers_ every token it has found during resolution. So basically this resolver must be used on every resolution to find all the tokens.

https://github.com/aws/aws-cdk/blob/810d736a8d20638e778c5773507f0edb12733a49/packages/%40aws-cdk/core/lib/private/resolve.ts#L270-L276

However, the resolver is not used specifically when we resolve tokens in **raw overrides**. Actually the current interface of `postProcess` function of `PostResolveToken` class makes It difficult to use an external resolver.

https://github.com/aws/aws-cdk/blob/810d736a8d20638e778c5773507f0edb12733a49/packages/%40aws-cdk/core/lib/cfn-resource.ts#L374-L380

That is why, in this PR, we move the resolution process outside of the `postProcess`, allowing to resolve tokens in raw overrides with the `RememberingTokenResolver` resolver.

This change also simplifies the current implementation of deepMerge as a side product.

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-lambda/test/python-lambda-handler/requirements.txt
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-lambda/test/python-lambda-handler/requirements.txt
@@ -1,1 +1,1 @@
-requests==2.26.0
+requests==2.31.0
--- a/packages/@aws-cdk/aws-lambda-python-alpha/test/lambda-handler-custom-build/requirements.txt
+++ b/packages/@aws-cdk/aws-lambda-python-alpha/test/lambda-handler-custom-build/requirements.txt
@@ -4,4 +4,4 @@ chardet==3.0.4
 idna==2.10
 urllib3==1.26.7
 # Requests used by this lambda
-requests==2.26.0
+requests==2.31.0
--- a/packages/@aws-cdk/aws-lambda-python-alpha/test/lambda-handler-dockercopy/Pipfile
+++ b/packages/@aws-cdk/aws-lambda-python-alpha/test/lambda-handler-dockercopy/Pipfile
@@ -4,4 +4,4 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [packages]
-requests = "==2.26.0"
+requests = "==2.31.0"
--- a/packages/@aws-cdk/aws-lambda-python-alpha/test/lambda-handler-dockercopy/Pipfile.lock
+++ b/packages/@aws-cdk/aws-lambda-python-alpha/test/lambda-handler-dockercopy/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "6cfaa5a495be5cf47942a14b04d50e639f14743101e621684e86449dbac8da61"
+            "sha256": "09535b2c80bf51574bb5186192ca9c8daac60752580f99bba70f11e9fb257c64"
         },
         "pipfile-spec": 6,
         "requires": {},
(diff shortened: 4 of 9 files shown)


tmokmss:
Exemption Request

Existing integration tests already cover the changes by this PR.

packages/aws-cdk-lib/core/lib/cfn-resource.ts
```diff
@@ -644,66 +609,11 @@ function deepMerge(target: any, ...sources: any[]) {
 
     for (const key of Object.keys(source)) {
       const value = source[key];
-      if (typeof(value) === 'object' && value != null && !Array.isArray(value)) {
+      if (typeof(value) === 'object' && value != null && !Array.isArray(value) && !(value instanceof Intrinsic)) {
```
TheRealAmazonKendra, line 612: `instanceof` is not safe in all contexts. See this [PR](https://github.com/aws/aws-cdk/pull/20564/files) for how we do this in other places.

TheRealAmazonKendra (changes requested):
Please make sure that your PR body describes the problem the PR is solving, and the design approach and alternatives considered. Explain why the PR solves the problem. A link to an issue is helpful, but does not replace an explanation of your thought process (See [Contributing Guide, Pull Requests](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request)).

From the context you've provided I don't totally understand the problem being solved or how your solution fixes it. Please update the body of your PR to provide this context.

We also definitely need an integration test on this. While we don't have any in core, you can pick any module to put this into practice with. We should also run this through the testing pipeline so I'm going to add that label here as well.

packages/aws-cdk-lib/core/lib/cfn-resource.ts
tmokmss: I don't understand why instanceof should not be used here. The PR has no description and it is not even merged. Why do you think so?

tmokmss:
Clarification Request

Why I should not use `instanceof`?

tmokmss:
Exemption Request

Currently integ-runner does not seem to support multiple-stacks deployment, which makes it difficult to test this scenario.

Also, this change only affects synth phase, not deployment phase, so I believe we don't need any integration test here.

```
ERROR      test/integ.cfn-resource 4.098s
      "cdk-integ" can only operate on apps with a single stack.

  If your app has multiple stacks, specify which stack to select by adding this to your test source:

      /// !cdk-integ STACK ...

  Available stacks: Stack1 Stack2 (wildcards are also supported)
```

tmokmss:
I just changed a random integ test to keep the PR open. I'll revert it after a review.

packages/aws-cdk-lib/core/lib/resolvable.ts
```diff
@@ -158,6 +158,8 @@ export class DefaultTokenResolver implements ITokenResolver {
       // The token might have returned more values that need resolving, recurse
       resolved = context.resolve(resolved);
       resolved = postProcessor.postProcess(resolved, context);
+      // resolve again since postProcess might have added more tokens (e.g. overriding)
+      resolved = context.resolve(resolved);
```
corymhall, line 162: It looks like we are passing the `context` through to the `postProcess`, but it
isn't making its way all the way through to the `processor`. Is there a way we
could pass it all the way through and use that to replace the
`Tokenization.resolve` function?

corymhall (approved):
Couple of questions

corymhall (changes requested):
Meant to request changes.

packages/aws-cdk-lib/core/lib/resolvable.ts
tmokmss: Thanks, I'll check that way

tmokmss: @corymhall 
I investigated it, and please correct me if my understanding is wrong 🙏 

If we take that approach, we can use `context.resolve` here instead of `Tokenization.resolve`, and it can also settle the issue.

https://github.com/aws/aws-cdk/blob/04323c44da6b8821a361a36bb2feadde77d54c35/packages/aws-cdk-lib/core/lib/cfn-resource.ts#L441-L447

The disadvantage is that, to pass the context, we have to change the signature of `_toCloudFormation` function something like `_toCloudFormation(context?)`. It requires changes in many files and I don't think it is a preferred approach. 

Isn't it more reasonable to resolve tokens outside of the postProcess function?

corymhall: I was thinking of adding it to the `PostResolveToken` `processor` method as an optional property

https://github.com/aws/aws-cdk/blob/b1e649a24375ca13a96c1b667ef27e92577b329f/packages/aws-cdk-lib/core/lib/util.ts#L82.

So then it would be something like
```ts
[this.logicalId]: new PostResolveToken({
            ...
          }, (resourceDef, context) => {
```

rix0rrr: I like Cory's solution. `PostResolveToken` is private and so is free to have its API changed. This will (hopefully?) cleanly achieve the desired solution.

tmokmss: @corymhall @rix0rrr 
Thanks, I think I have to study about CDK tokens more 😄  Will try the approach anyway.

One question is, why not resolve tokens after the deepMerge? (i.e. my original approach)
It can remove some hacks currently implemented like `removeEmpty: false` or `MERGE_EXCLUDE_KEYS`, and simplify the existing logic.

Is it performance-wise reason (to reduce recursion), or is there any edge case not to work properly that I am not aware of?

tmokmss:
Hi @corymhall could you review this again? Idk why it is regarded as CHANGES REQUESTED state.

corymhall (approved):

PR #24927, opened 2023-04-04
chore(release): revert `1.199.0` merge-back 

iliapolo:
Reverts aws/aws-cdk#24893

The merge back mistakenly bumped to version `1.199.0` even though the release associated with it was for `1.198.1`.
It happened because th `1.199.0` release got triggered before `1.198.1` finished, and the merge back build reads the current version from the latest `v1-release` branch, instead of the commit that triggered it. 

Long story short, the [latest v1 version](https://www.npmjs.com/package/aws-cdk?activeTab=versions) is `1.198.1`, so thats what we should have in the code.


iliapolo:
Not ready yet.

Naumel (approved):

PR #24938, opened 2023-04-05
docs(elbv2): clarify target group documentation

laurelmay:
The Target Group documentation improperly described the `port` property
as where the _listener_ listened for traffic; however, this property is
for the _target group_ and has no bearing on any created listeners.
Additionally, `protocol` is not just optional for Lambda functions, it
is entirely ignored.

This updates the wording to be slightly closer to the documentation on
the `AWS::ElasticLoadBalancingV2::TargetGroup` resource documentation
from CloudFormation.

----


TheRealAmazonKendra (approved):
Thanks!

PR #24940, opened 2023-04-05
chore(deps): bump rosetta to `1.80.0`

iliapolo:
Pulling the rosetta [fix](https://github.com/aws/jsii/pull/4034) that should fix the pipelines. Didn't do this through the upgrade workflow [PR](https://github.com/aws/aws-cdk/pull/24933) because it looks like there are other dependencies failing the build we don't want to be blocked on.


TheRealAmazonKendra (approved):
Approving, assuming the build succeeds.

PR #24949, opened 2023-04-05
feat(step-functions-tasks): Support PropagatedTagSource property for EcsRunTask

### LINKED ISSUE #12999: aws-step-function-tasks: Support PropagateTags for EcsRunTask
Hello!

When launching an ECS task with Step Functions, it is possible to specify the `PropagateTags` parameter (cf. this [documentation](https://docs.aws.amazon.com/step-functions/latest/dg/connect-ecs.html)) but I do not see it in the [properties](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-stepfunctions-tasks.EcsRunTaskProps.html) of the EcsRunTask construct.

### Use Case

This is mainly for tracking costs associated to the tasks running in Fargate.

### Proposed Solution

At the moment, the workaround is to convert the EcsRunTask to a custom task and to add the `PropagateTags` field directly. It is working nicely but you lose the benefits of EcsRunTask. That means you need to create the IAM role yourself because CDK does not "recognize" the step as an ECS task anymore.

```
ecs_task_as_json = sfn_tasks.EcsRunTask(...).to_state_json()
ecs_task_as_json["Parameters"]["PropagateTags"] = "TASK_DEFINITION"
ecs_task = sfn.CustomState(self, task_name, state_json=ecs_task_as_json)
```

Having an enum rather than string and a default value set to `TASK_DEFINITION` will be a good approach?

### Other

N/A

---

msysh:
Currently, when launching an ECS task with Step Functions, it is impossible to propagate tag to task. To solve this issue, add property for PropagatedTagSource for EcsRunTask

Closes #12999

----


comcalvi (approved):
lgtm, nice work!

PR #24986, opened 2023-04-07

p-mercury:
Closes https://github.com/aws/aws-cdk/issues/24548

I added and implemented the BundlingOutput.FILE type which allows bundled assets to produce single files as output, which in useful for targets which can not parse directories.

### DIFF
--- a/CHANGELOG.v2.alpha.md
+++ b/CHANGELOG.v2.alpha.md
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.76.0-alpha.0](https://github.com/aws/aws-cdk/compare/v2.75.1-alpha.0...v2.76.0-alpha.0) (2023-04-19)
+
 ## [2.75.1-alpha.0](https://github.com/aws/aws-cdk/compare/v2.75.0-alpha.0...v2.75.1-alpha.0) (2023-04-18)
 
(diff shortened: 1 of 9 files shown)


packages/aws-cdk-lib/core/test/staging.test.ts
```diff
@@ -1260,6 +1260,53 @@ describe('staging', () => {
...
+        image: DockerImage.fromRegistry('alpine'),
+        command: [DockerStubCommand.SUCCESS],
+        outputType: BundlingOutput.FILE,
+      },
+    });
+
+    // THEN
+    const assembly = app.synth();
+    expect(fs.readdirSync(assembly.directory)).toEqual([
+      'asset.f5160e037189d688a526fefb9c0ef62e8638115ba7ccc90011e6e0aecb05f39d',
```
corymhall, line 1283: It looks like it is still staging both the directory and the file, can you check on that?

packages/@aws-cdk-testing/framework-integ/test/aws-appsync/test/integ.js-resolver-bundled.js.snapshot/asset.c8a056b7b029cede879af4833f48bcc456748b2f8456f85ba76393466c08693a/index.js
corymhall, line 1: It doesn't look like the `.js` asset is being committed which may be due to .gitignore rules. Can you check?

corymhall (changes requested):
@p-mercury thanks for the PR! This looks great, just have 2 minor questions.

corymhall (changes requested):
putting back in request changes to reflect status.

Issue #24984, opened 2023-04-07
(ecs): Allow passing executionRole to imported TaskDefinitions


blimmer:
### Describe the feature

Currently, when importing a TaskDefinition, either via `TaskDefinition.fromTaskDefinitionAttributes` or `Fargate.fromFargateTaskDefinitionAttributes`, you can pass a `taskRole`, but not an `executionRole`.

In both cases, the `CommonTaskDefinitionAttributes` interface defines the attributes that can be passed when importing: https://github.com/aws/aws-cdk/blob/a98a98147534f89a219521a2e51a6a1e25a2ac06/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts#L231-L253

As you can see, you can pass `taskRole` but not `executionRole`.

### Use Case

I have a shared TaskDefinition that I use in other CDK apps. This TaskDefinition is triggered by an EventBridge rule. I share the TaskDefinition ARN, Task Role ARN (and Execution Role ARN) via `CfnOutput`s.

Since I use a custom image stored in ECR, my TaskDefinition has both a Task Role and an Execution Role. When creating the EventBridge rule, the `EcsTask` target needs to allow the `events` `ServicePrincipal` to `iam:PassRole` _both_ the Task Role and Execution Role for EventBridge to successfully `RunTask`.

The `EcsTask` target already has this logic: https://github.com/aws/aws-cdk/blob/a98a98147534f89a219521a2e51a6a1e25a2ac06/packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts#L206-L213

However, because I can't pass the `executionRole` when I import the Task Definition, the logic to allow `PassRole` isn't added, and the EventBridge invocation fails.

I can work around this issue by forcing `taskDefinition.executionRole` assignment after the import:

```ts
(importedEcsTask.executionRole as any) = importedExecutionRole;
```

### Proposed Solution

Currently, these `.from` imports rely on the `ImportedTaskDefinition` class: https://github.com/aws/aws-cdk/blob/a98a98147534f89a219521a2e51a6a1e25a2ac06/packages/aws-cdk-lib/aws-ecs/lib/base/_imported-task-definition.ts

This class already exposes an `executionRole` property: https://github.com/aws/aws-cdk/blob/a98a98147534f89a219521a2e51a6a1e25a2ac06/packages/aws-cdk-lib/aws-ecs/lib/base/_imported-task-definition.ts#L53-L57

[...]

PR #24987, opened 2023-04-07
fix(ecs): allow passing execution role to imported TaskDefinitions

blimmer:
See #24984 for details.

TLDR; there's not a way currently exposed to define the `executionRole` on an imported `TaskDefinition`. This change allows optionally passing an `executionRole`.

Fixes issue #24984.

----


packages/aws-cdk-lib/aws-ecs/test/external/external-task-definition.test.ts
```diff
@@ -617,4 +617,34 @@ describe('external task definition', () => {
       deviceType: 'eia2.medium',
     })).toThrow('Cannot use inference accelerators on tasks that run on External service');
   });
+
+  test('can import an External Task Definition using attributes', () => {
```
blimmer, line 621: Import was previously untested.

blimmer:
Exemption Request - there are no integ tests for `aws-ecs` (from what I can tell). Can I skip this and rely on the unit tests instead?

corymhall (dismissed):

corymhall (approved):

PR #24998, opened 2023-04-07
fix(sqs): explicitly set default encryption to SQS_MANAGED

### LINKED ISSUE #24993: (aws-sqs): default encryption of `SSE-SQS` is not explicitly set
### Describe the bug

Right now, an SQS Queue [has `SQS_MANAGED (SSE-SQS)` enabled by default](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_sqs.Queue.html#encryption). This, however, is [implicit](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-sqs/lib/queue.ts#L399) and should be explicit, setting the same properties as if `SQS_MANAGED` was [explicitly set](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-sqs/lib/queue.ts#L434).

### Expected Behavior

I expect the default Queue to create a CFN Resource with the the following property: `sqsManagedSseEnabled: true`

### Current Behavior

There is no `sqsManagedSseEnabled: true` property set, instead it assumes SSE-SQS is enabled *at the region*.

### Reproduction Steps

Create a SQS queue and examine its CFN output.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.70

### Framework Version

_No response_

### Node.js Version

16.3.0

### OS

Mac

### Language

Python

### Language Version

_No response_

### Other information

_No response_

joel-aws:
Currently, `SQS_MANAGED` encryption is the implicit default in new queues. This, however, can causes issues with deployments into other regions (e.g., gov cloud), where the CFN behavior is not necessarily the same.

By making the encryption default explicit, it is easier to test for the conditions and alert or make changes based on it.

Closes #24993


joel-aws:
https://github.com/Mergifyio update

jumic:
As already mentioned in the corresponding issue, AWS has changed the [default behaviour for newly created queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-configure-sse-existing-queue.html).
> To protect the data in a queue’s messages, Amazon SQS has server-side encryption (SSE) enabled by default for all newly created queues.

Existing queues that were created before are still unencrypted. If we explicitly set `sqsManagedSseEnabled` to `true` this will change encryption setting for those queues. [This blog entry](https://aws.amazon.com/blogs/compute/announcing-server-side-encryption-with-amazon-simple-queue-service-managed-encryption-keys-sse-sqs-by-default/) describes that `sqsManagedSseEnabled` has to be set to `false` when anonymous requests are used.

From my point of view, this change is a breaking change. If `sqsManagedSseEnabled` will be set to `true` by default in CDK and queues were created before SSE-SQS was a default in AWS, anonymous requests will fail. Therefore it should be considered to add a feature flag if this PR will be accepted.

joel-aws:
@jumic -- agree with your points. Have to move this to draft, but will get to adding a feature flag later.

PR #24999, opened 2023-04-08
fix(ecs): potential race condition on TaskRole default policy update with CfnService

### LINKED ISSUE #24880: aws-ecs: Race condition in Ec2Service & FargateService between updating the TaskRole default policy and the CfnService
### Describe the bug

Within [Ec2Service](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.Ec2Service.html) and [FargateService](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.FargateService.html), there is no dependency added between the TaskRole default policy and the CfnService.

If updates are made to the TaskRole default policy (e.g. adding new inline policies), CloudFormation will update both the CfnService and the TaskRole default policy at the same time, and this can result in ECS tasks spinning up without the proper permissions being updated in IAM for the task role yet.

### Expected Behavior

The TaskRole and its children (e.g. the default policy) should _always_ deploy first within a CloudFormation stack before the ECS service.

### Current Behavior

The TaskRole default policy and the ECS service deploy at the same time.

### Reproduction Steps

Every usage of [Ec2Service](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.Ec2Service.html) and [FargateService](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.FargateService.html) is vulnerable to this deployment race condition risk, if they ever make edits to the task role default policy.

### Possible Solution

```
this.node.addDependency(this.taskDefinition.taskRole);
```
within the [BaseService](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.BaseService.html) constructor.

Lambda L2 Function construct does this already (see [here](https://github.com/aws/aws-cdk/blob/v2.72.0/packages/@aws-cdk/aws-lambda/lib/function.ts#L831)).

_No response_

### CDK CLI Version

2.72.0

### Framework Version

_No response_

### Node.js Version

v16.9.1

### OS

Mac OSX

### Language

Typescript

### Language Version

TypeScript 4.9.5

### Other information

_No response_

lpizzinidev:
Prevents potential race conditions on TaskRole default policy update in EC2 and Fargate services by adding a dependency on the TaskRole.
This will update the TaskRole and its children first and the service after.

Closes #24880.

----


packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -561,6 +561,8 @@ export abstract class BaseService extends Resource
       ...additionalProps,
     });
 
+    this.node.addDependency(this.taskDefinition.taskRole);
```
corymhall, line 564: From the linked issue it looks like we should be adding a dependency on the `Policy` not the role. There should already be an implicit dependency on the role and in cases where the policy is updated the role itself will probably not be updated so will not be in the dependency chain.

corymhall (dismissed):

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
lpizzinidev: Thanks for the feedback.

The [default policy](https://github.com/aws/aws-cdk/blob/3fe166f8f74784eb04b6e048659e46785483d786/packages/aws-cdk-lib/aws-iam/lib/role.ts#L413) is a `private` property in the role.
Should I create a public getter to retrieve it and add a dependency with:
```
this.node.addDependency(this.taskDefinition.taskRole.getDefaultPolicy);
```
?

corymhall: I don't think dependencies work that way. If I declare an explicit dependency on resource A, CloudFormation does not view that as adding an explicit dependency on every resource that resource A depends on. It only depends on resource A, so if resource A is not updated then it doesn't have to wait.

But we can always run tests to confirm :)

corymhall: We might be able to accomplish this globally though if we can assume that anytime someone adds a dependency on a role that they also want to add a dependency on the default policy (I think that is a safe assumption)

I think by adding something like this to `Role`
```ts
Dependable.implement(this, {
  dependencyRoots: [this, this.defaultPolicy],
});
```

Whenever you do `addDependency(role)` it will also add a dependsOn for the policy.

iancaffey: > I don't think dependencies work that way. If I declare an explicit dependency on resource A, CloudFormation does not view that as adding an explicit dependency on every resource that resource A depends on. It only depends on resource A, so if resource A is not updated then it doesn't have to wait.

For sure, but I'm only talking about child constructs of the Role. The default policy is scoped to the role, so addDependency(role) takes care of it.

corymhall: ah you are right. @lpizzinidev can you add a test that asserts the dependency is added on the policy as well?

packages/aws-cdk-lib/aws-ecs/test/base-service.test.ts
```diff
@@ -44,6 +44,27 @@ describe('When import an ECS Service', () => {
...
+
+    // WHEN
+    new ecs.FargateService(stack, 'FargateService', {
+      cluster,
+      taskDefinition,
+    });
+
+    // THEN
+    Template.fromStack(stack).hasResource('AWS::ECS::Service', {
+      DependsOn: ['FargateTaskDefTaskRole0B257552'],
```
corymhall, line 65: The `DependsOn` should also contain the policy resource.

corymhall (changes requested):

iancaffey:
@lpizzinidev @corymhall any chance we can get this PR across the finish line? :)

lpizzinidev:
@iancaffey @corymhall 
I was looking into this, the problem is that for `FargateService` the default policy is not added to the dependencies array.
Could it be that the construct is not creating a default policy?
I see that `defaultPolicy` is initialized in the [`addToPrincipalPolicy`](https://github.com/aws/aws-cdk/blob/37caaabd9d28dd7bb7d0499cc8606e1a382b32fa/packages/aws-cdk-lib/aws-iam/lib/role.ts#L521-L525) but I can't find a call to the method in `FargateService`.
I tried adding the dependency manually in the `Role` as suggested [here](https://github.com/aws/aws-cdk/pull/24999#discussion_r1179129964) without luck (also, this step should be unnecessary as reported [here](https://github.com/aws/aws-cdk/pull/24999#discussion_r1179166059)).

iancaffey:
You'll see a lot of `taskDefinition.addToTaskRolePolicy(...)`. Or callers might directly reference `taskDefinition.taskRole` when adding grants. 

But yeah, the default props don't end up triggering adding any permissions that would end up initializing the policy. 

Could you manually add task role permissions in the unit test to trigger initializing the policy?

PR #25005, opened 2023-04-09
fix(elasticloadbalancingv2): ALB listeners with multiple forwardi…

### LINKED ISSUE #24805: elbv2: addTargetGroups add only 1 target group.
### Describe the bug

ApplicationListener.addTargets(id, props) fails to add more than 1 target group to the rule created. When adding more than one target group via this function the resulting template only has the second targetgroup on the array added to the loadbalancer listener rule. 

### Expected Behavior

All target groups are added with the same weight to the created rule. per the documentation.

Load balance incoming requests to the given load balancing targets.

This method implicitly creates an ApplicationTargetGroup for the targets involved, and a 'forward' action to route traffic to the given TargetGroup.

If you want more control over the precise setup, create the TargetGroup and use addAction yourself.

It's possible to add conditions to the targets added in this way. At least one set of targets must be added without conditions.

### Current Behavior

Only the second targetgroup in the array of target groups for the props is added. 

### Reproduction Steps

```ts
for (let index = 0; index < totalContainers; index++) {
      var containerDefinition = taskDefinition.addContainer('UnityContainer_' + index, {
        containerName: 'unity-render-' + index,
        image: ecs.ContainerImage.fromEcrRepository(repo),
        portMappings: [{ containerPort: 21075 }],
        memoryReservationMiB: 4096,
        cpu: 2048,
        environment:{wb_environment:"dev"},
        secrets: {wb_secret: ssmSecret},
        privileged: true,
        logging: ecs.LogDriver.awsLogs({ streamPrefix: 'unity-render', logGroup: containerLogs })
      });
      containerDefinition.addMountPoints({ containerPath: '/tmp', sourceVolume: 'tmp', readOnly: false })
     let targetGroup = new elbv2.ApplicationTargetGroup(ecsService, `ecsTarget-${process.env.CDK_ENV}-${index}`, {
        port: 80,
        targets: [ecsService.loadBalancerTarget({
          containerName: containerDefinition.containerName,
          containerPort: 21075,
        })],
        targetType: elbv2.TargetType.INSTANCE,
        targetGroupName: `ecsTarget-${process.env.CDK_ENV}-${index}`,
        deregistrationDelay: cdk.Duration.seconds(45),
        healthCheck: {
          interval: cdk.Duration.seconds(15),
          path: '/test/ok',
          timeout: cdk.Duration.seconds(5)
        },
        vpc: vpc,
      });
      ecsTargets.push(ecsTarget); 
     }
const http80Listener = lb.addListener(`HTTP-80-${process.env.CDK_ENV}`, {
      port: 80,
      open: true,
    });

 http80Listener.addTargetGroups('test',{
      targetGroups: ecsHttpTargets,
      conditions: [
        elbv2.ListenerCondition.pathPatterns(['/mypath/*'])
      ],
      priority:10
    })
```

### Possible Solution

Resulting template adds both targetgroups to the rule

_No response_

### CDK CLI Version

2.69.0 (build 60a5b2a)

### Framework Version

aws-cdk-lib@2.69.0

### Node.js Version

v16.13.1

### OS

22.3.0 Darwin Kernel Version 22.3.0: Thu Jan  5 20:50:36 PST 2023; root:xnu-8792.81.2~2/RELEASE_ARM64_T6020 arm64

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

Styerp:
…ng targets need weights specified

The [ELB docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html#forward-actions) on this aren't explicit, but the examples do _not_ show a configuration with multiple targets, but no weighting. This change assigns a weight to the action, if one wasn't specified.

If specifying stickiness, but no weight, the template will change.

Fixes #24805.

Results in CFN like
  LBListener49E825B4:
    Type: AWS::ElasticLoadBalancingV2::Listener
    Properties:
      DefaultActions:
        - ForwardConfig: TargetGroups: - TargetGroupArn: Ref: TargetGroupOne7810CAFB Weight: 1 - TargetGroupArn: Ref: TargetGroupTwo593946E1 Weight: 1 Type: forward

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "jsii-diff": "1.78.1",
     "jsii-pacmak": "1.78.1",
     "jsii-reflect": "1.78.1",
-    "jsii-rosetta": "~5.0.0",
+    "jsii-rosetta": "~5.0.6",
     "lerna": "^6.6.1",
     "patch-package": "^6.5.1",
     "semver": "^6.3.0",
--- a/packages/@aws-cdk/aws-kinesisfirehose-alpha/package.json
+++ b/packages/@aws-cdk/aws-kinesisfirehose-alpha/package.json
@@ -111,6 +111,11 @@
       "*:*"
     ]
   },
+  "jsiiRosetta": {
+    "exampleDependencies": {
+      "@aws-cdk/aws-kinesisfirehose-destinations-alpha": "*"
+    }
+  },
   "pkglint": {
     "exclude": [
       "naming/package-matches-directory",
--- a/packages/aws-cdk-lib/assertions/README.md
+++ b/packages/aws-cdk-lib/assertions/README.md
@@ -121,7 +121,7 @@ template.hasResource('Foo::Bar', {
 });
 ```
 
-You can also assert the definitions of all resources of a type using the 
+You can also assert the definitions of all resources of a type using the
 `allResources()` API.
 
 ```ts
@@ -146,9 +146,9 @@ that matches specific properties. The following code asserts that a template con
 an Output with a `logicalId` of `Foo` and the specified properties -
 
 ```ts
-const expected = { 
+const expected = {
   Value: 'Bar',
-  Export: { Name: 'ExportBaz' }, 
+  Export: { Name: 'ExportBaz' },
 };
 template.hasOutput('Foo', expected);
 ```
@@ -177,7 +177,7 @@ The APIs `hasMapping()`, `findMappings()`, `hasCondition()`, and `hasCondtions()
 
 The expectation provided to the `hasXxx()`, `findXxx()` and `templateMatches()`
 APIs, besides carrying literal values, as seen in the above examples, also accept
-special matchers. 
+special matchers.
 
 They are available as part of the `Match` class.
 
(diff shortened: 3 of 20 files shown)


Styerp:
From a design perspective:

I followed the intent of the construct in making the weights equal. There may be nuances around other actions and their weights I did not consider when choosing 1 as the weight. I'm not sure if what weights add up to is important or if it's just what % of the whole they are, for example. Any guidance or feedback welcome. 

From a rules of contribution perspective:

I haven't been able to get the integ test to actually run from within the aws-cdk repo, so I can't generate the snapshot. I can run it in a different project, link the cdk, re-install @aws-cdk/integ-runner (because it has broken references right now), then run the integ tests. The CFN template for the integration test is ostensibly valid, but deployment fails before the change. Thoughts?

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener-rule.ts
```diff
@@ -254,9 +254,9 @@ export class ApplicationListenerRule extends Construct {
       this.configureAction(props.action);
     }
 
-    (props.targetGroups || []).forEach((group) => {
-      this.configureAction(ListenerAction.forward([group]));
-    });
+    if(props.targetGroups) {
+        this.configureAction(ListenerAction.forward(props.targetGroups));
```
otaviomacedo, line 258: Wouldn't only this change be enough?

otaviomacedo (approved):

otaviomacedo (approved):

PR #25008, opened 2023-04-10

laverdet:
I am reopening this from https://github.com/aws/aws-cdk/pull/24557

> This PR has been deemed to be abandoned, and will be automatically closed. Please create a new PR for these changes if you think this decision has been made in error.

---

```
Resources
 └─ [~] Description
     ├─ [-] ?????
     └─ [+] 🤦🏻‍♂️
```


laverdet:
@TheRealAmazonKendra I am following up on your comment from 3 weeks ago here: https://github.com/aws/aws-cdk/pull/24557#issuecomment-1487877689

This change only affects the CLI output and therefore an integration test isn't possible. I'm happy to spend more time documenting or working through other prerequisites but I do want to make sure that, conceptually, this change is one that the team will accept.

laverdet:
@TheRealAmazonKendra Hi there just another ping to my previous comment. Please let me know if there is someone else I can loop in if you don't have the bandwidth.

laverdet:
@TheRealAmazonKendra Just checking in again. The PR is not abandoned. Please let me know if there's anything I can do to help.

laverdet:
@rix0rrr would you mind taking a look at this? I'm happy to adjust the approach, add more tests, or do what else needs to be done. I'm not getting any feedback from the team so I'm not sure how to proceed. The diff noise with non-ASCII information in `cdk diff` makes it difficult to find meaningful changes to our stacks.

laverdet:
@aws-cdk-automation @rix0rrr @TheRealAmazonKendra thank you, the PR is not abandoned. I'll open a new instance tomorrow where we can continue our discussion.

PR #25079, opened 2023-04-12
chore(s3): documented how the bucket encryption configuration is computed

otaviomacedo:
----


packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1983,7 +1983,22 @@ export class Bucket extends BucketBase {
...
   /**
    * Set up key properties and return the Bucket encryption property from the
-   * user's configuration.
+   * user's configuration, according to the following table:
+   *
+   * | props.encryption | props.encryptionKey | props.bucketKeyEnabled | bucketEncryption (return value) | encryptionKey (return value) |
+   * |------------------|---------------------|------------------------|---------------------------------|------------------------------|
+   * | undefined        | undefined           | e                      | undefined                       | undefined                    |
+   * | UNENCRYPTED      | undefined           | false                  | undefined                       | undefined                    |
+   * | undefined        | k                   | e                      | SSE-KMS, bucketKeyEnabled = e   | k                            |
```
corymhall, line 1992: What does `e` represent?

otaviomacedo: It's just a variable being reused in the other columns. What I'm trying to show in this line, for example, is that whatever the value of `props.bucketKeyEnabled`, it will be used in the output `bucketEncryption`. In the lines where it's not reused, it just means that it doesn't matter. Same goes for `k`.

I'm not sure what better way to represent this.

PR #25089, opened 2023-04-12
chore(cdk): Speed up typescript app compile time

miguel-a-calles-mba:
> REPLACE THIS TEXT BLOCK
>
>
> code you submit.
>

Closes #None.

The cdk init command can generate a typescript cdk app. The cdk app uses ts-node for things like cdk deploys. As a developer, it seems to take forever for the ts-node to start the cdk deploy process. Adding SWC to the tsconfig file allows ts-node to use a rust-based transpiler when doing a cdk deploy. The SWC speeds up the cdk deploy time by a lot.

I have not noticed any type checking issues in vscode nor any transpile issues in deployed lambda functions.

Ref: https://typestrong.org/ts-node/docs/swc/
Ref: https://aws.plainenglish.io/speed-up-aws-cdk-deploys-up-to-80-c47afad1c18c

----


corymhall (changes requested):
@miguel-a-calles-mba can you update the PR description to include your motivation behind this change and what performance improvements you see?

I've thought about doing something like this before, but we would lose type checking which I'm not sure we should do.

miguel-a-calles-mba:
> @miguel-a-calles-mba can you update the PR description to include your motivation behind this change and what performance improvements you see?
> 
> I've thought about doing something like this before, but we would lose type checking which I'm not sure we should do.

Hi @corymhall, I added a description explaining the rationale.

miguel-a-calles-mba:
My CI/CD pipeline execution time dropped from 345 secs to 74 secs	 when I added the ts-node SWC setting.

corymhall (dismissed):
@miguel-a-calles-mba after discussing this with the team, disabling type checking is not something that we want to turn on by default for all users. If users know what they are doing and want to enable it then they can update their config.

We can definitely update our documentation to have instructions on doing so because I do think this information would be valuable to a lot of users. If you want to take a stab at updating the documentation let me know.

miguel-a-calles-mba:
@corymhall, sounds good. Is there a specific section that should be updated?

corymhall:
> @corymhall, sounds good. Is there a specific section that should be updated?

I would add it to this section https://github.com/aws/aws-cdk/tree/main/packages/aws-cdk#configuration

corymhall (approved):

Issue #16118, opened 2021-08-18
(aws-dynamodb): L2 Construct for GlobalTable


cmckni3:
Create an L2 construct for `AWS::DynamoDB::GlobalTable` to abstract away creation details.

Handling the setup of streams, encryption, etc., in the L1 construct can be cumbersome.

### Use Case

DynamoDB replication between regions.

### Proposed Solution

Create L2 construct called `GlobalTable` similar to the `Table` construct.

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

madeline-k:
Link: https://github.com/aws/aws-cdk/issues/15957

madeline-k:
Thanks for opening this feature request! There are a couple of issues open asking for the same thing, will link them here:
#15957, https://github.com/aws/aws-cdk/issues/15258. I am marking this as a p2, which means we won't be able to prioritize it right now. But we always welcome contributions! It looks like there are a few people in the community interested in adding this. Please take a look at our contributing guide: https://github.com/aws/aws-cdk/blob/master/CONTRIBUTING.md

ethanrucinski:
I have started working on an L2 construct for this. @cmckni3 if you happen to have any specific use cases or requests let me know if it might be possible to share! I'd love any feedback if I'm able to make my way along with this.

ethanrucinski:
Hey do we have an ETA for this construct? We're trying to decide if we should publish our own internally or wait for this to become available

barriem:
Would be awesome to get the new L2 construct, keeping an eye on this ticket!

nakedible-p:
Big question here: are you planning on keeping the current method of just setting `replicationRegions` on a DynamoDB `Table` to create a global table - or are you planning on exposing a different construct `GlobalTable` for global tables as CloudFormation does?

I might able to contribute if this decision is clear.

gagipro:
+1
issue with node JS14.x deprecation on custom resources and not respecting enterprise security rules.
also removing replication through cdk doesn't seem supported.

mrgrain:
Ref: https://github.com/aws/aws-cdk-rfcs/blob/master/text/0510-dynamodb-global-table.md

anshu49746:
Hi,

Any ETA when the Global Table L2 construct will be available for usage?

anshu49746:
[...]

PR #25091, opened 2023-04-12
feat: first commit of globaltable l2

pattasai:
This PR creates L2 construct for globalTable of Dynamodb

Resolves https://github.com/aws/aws-cdk/issues/16118

----


packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,2152 @@
...
+export interface SystemErrorsForOperationsMetricOptions extends cloudwatch.MetricOptions {
+
+    /**
+     * The operations to apply the metric to.
+     *
+     * @default - All operations available by DynamoDB tables will be considered.
+     */
+    readonly operations?: Operation[];
+
+}
```
rix0rrr, line 38: I appreciate that you copy/pasted all of this out to get started.

Now that we're working on landing this as "official" code in the CDK repository, it makes sense to start cleanup.

For all the definitions that are shared between `table.ts` and `global-table.ts`, it makes sense to move them out to a separate file (call it `shared.ts` for example), and start referencing them from there.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,2152 @@
...
+    /**
+     * Server-side KMS encryption with a master key managed by AWS.
+     */
+    AWS_MANAGED = 'AWS_MANAGED',
+}
+
+/**
+ * Represents the table schema attributes.
+ */
+export interface SchemaOptions {
```
rix0rrr, line 140: I don't like this type. Remember how the existence of this type, shared as it was everywhere, made it impossible for us to make a change we needed to make to address one issue? We talked about this.

I hope you're not using it below :). If you are, just copy/paste these two fields into where you were using it before.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,2152 @@
...
+
+        this.billingMode = props.capacity?.billMode || BillingMode.PAY_PER_REQUEST
+
+        const { sseSpecification, encryptionKey } = this.parseEncryption(props);
+
+        let streamSpecification: CfnTable.StreamSpecificationProperty | undefined;
+        if (props.replicas) {
+            if (props.stream && props.stream !== StreamViewType.NEW_AND_OLD_IMAGES) {
+                throw new Error('`stream` must be set to `NEW_AND_OLD_IMAGES` when specifying `replicationRegions`');
+            }
```
rix0rrr, line 1332: This is true. Do we also check this when `addReplica` is called?

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,2152 @@
...
+            streamSpecification = { streamViewType: StreamViewType.NEW_AND_OLD_IMAGES };
+
+            this.billingMode = props.billingMode ?? BillingMode.PAY_PER_REQUEST;
+        } else {
+            this.billingMode = props.billingMode ?? BillingMode.PAY_PER_REQUEST;
```
rix0rrr, line 1337: This line occurs twice. It could occur only once, and doesn't have to be inside the `if (props.replicas)` check, right?

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,2152 @@
...
+     * @default RemovalPolicy.RETAIN
+     */
+    readonly removalPolicy?: RemovalPolicy;
+
+    /**
+     * Regions where replica tables will be created
+     *
+     * @default - no replica tables are created
+     */
+    readonly replicationRegions?: string[];
```
rix0rrr, line 264: Shouldn't there be properties associated with replicas?

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,2152 @@
...
+     * The name of TTL attribute.
+     * @default - TTL is disabled
+     */
+    readonly timeToLiveAttribute?: string;
+
+    /**
+     * When an item in the table is modified, StreamViewType determines what information
+     * is written to the stream for this table.
+     *
+     * @default - streams are disabled unless `replicationRegions` is specified
```
rix0rrr, line 248: 👍

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,2152 @@
...
+     *
+     * DO NOT UNSET this property if adding/removing multiple replicationRegions
+     * in one deployment, as CloudFormation only supports one region replication
+     * at a time. CDK overcomes this limitation by waiting for replication to
+     * finish before starting new replicationRegion.
+     *
+     * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-globaltable.html#cfn-dynamodb-globaltable-replicas
+     * @default true
+     */
+    readonly waitForReplicationToFinish?: boolean;
```
rix0rrr, line 287: This seems like a holdover from when we used to do the replication ourselves. Is this still necessary?

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,2152 @@
...
+            resourceName: 'AWSServiceRoleForApplicationAutoScaling_DynamoDBTable',
+        }));
+    }
+
+    /**
+     * Creates replica tables
+     *
+     * @param regions regions where to create tables
+     */
+    private createReplicaTables(regions: string[], timeout?: Duration, waitForReplicationToFinish?: boolean) {
```
rix0rrr, line 1768: This whole function needs to disappear, doesn't it?

rix0rrr (commented):
I like where this is going! 

I'm noticing a whole bunch of code that is either no longer being used, or that should now be shared between `table.ts` and `global-table.ts`, so it's kinda hard to review on its own.

I assume you'll be doing a cleanup at some point, right?

mrgrain:
Drafting this until someone can work on it again.

kaizencc:
Superseded by #26563

PR #25095, opened 2023-04-13
fix(servicecatalogappregistry): Remove deprecated resource in Application Construct

yingdon-amazon:
BREAKING CHANGE: this change will deprecated **associateStack** and **associateAttributeGroup** in Application Construct. 
The user who are using these two method need to update to use alternative method.
For associateStack, the alternative method is **associateApplicationWithStack**
For associateAttributeGroup, the alternative method is **AttributeGroup.associateWith**

The user who are using these two method need to update to use alternative method. For associateStack, the alternative method is **associateApplicationWithStack** For associateAttributeGroup, the alternative method is **AttributeGroup.associateWith**

Purpose of this PR:

we need to remove deprecated resource before we moving into stable version The method that we remove is: associateStack and associateAttributeGroup

CHANGES:

1. in lib/application.ts, we remove these two methods and update their corresponding interface
2. in test/ application.test.ts & test/integ.application.ts, we update application.test.ts and integ.application.ts to remove these two methods' related test


rix0rrr (dismissed):

santanugho:
`associateStack` and `associateAttributeGroup` is already available in the main branch for cdk `v1`.
Check [this code](https://code.amazon.com/packages/Aws-streaming-data-solution/blobs/f1a7d4f69348de6397dec82eaa8a7ad2c93e7fc9/--/source/lib/app-registry.ts?__session=eyJ0YWJzIjpbeyJmaWxlIjoic291cmNlL3BhY2thZ2UuanNvbiJ9LHsiZmlsZSI6InNvdXJjZS9saWIvYXBwLXJlZ2lzdHJ5LnRzI0w2NCJ9XX0%3D#L64) which is using `associateStack` from the [main branch](https://code.amazon.com/packages/Aws-streaming-data-solution/blobs/15f67cc21029f834f9b6cad6da6cb0c0504c02a6/--/source/package.json?__session=eyJ0YWJzIjpbeyJmaWxlIjoic291cmNlL3BhY2thZ2UuanNvbiNMNDgifSx7ImZpbGUiOiJzb3VyY2UvbGliL2FwcC1yZWdpc3RyeS50cyNMNjQifV19#L48).

By removing these two functions completely will mean these methods will never be available in any of the latest CDK version and if any customer wants to upgrade their  CDK dependency, then they cant do that without changing the underlying code.

rix0rrr (approved):

PR #25097, opened 2023-04-13
feat(ecs): Update ECS IBaseService interface to include metric functions

infanton:
The [cdk-monitoring-constructs](https://github.com/cdklabs/cdk-monitoring-constructs) package added [this commit](https://github.com/cdklabs/cdk-monitoring-constructs/commit/0f97cdaa63d5240f0ac8e36bc6d4ee35fda98ed8#diff-af1c768a7595564edd2a06c45cf934f8334176e066ff74a4f7a6351180c99532R137) to accept imported load balancer and target groups; however, it is unable to accept imported fargate service because the returned interface from `aws-cdk-lib/aws-ecs/lib/base` does not include metrics.

----


cmonsalve-truxnow:
Could we open this PR again? it doesn't make sense that IBaseService and IFargateService don't support metrics when importing resources.

PR #24459, opened 2023-03-05
fix(ec2): Vpc lookup is setting region from stack

daschaa:
Fixes #22178.

This PR tackles the issue that the VPC `fromLookup` method was not setting the region correctly, even if passed in the lookup method.
The issue was, that in the dummy response from the lookup method, the region was not set.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

Issue #22178, opened 2022-09-21
aws_ec2: vpc.fromLookup doesn't set env.region correctly


cartalla:
### Describe the bug

The context is that the bug fix for #20496 and #20530 did not resolve the problem with the Route53.add_vpc() call not setting the region for the VPC correctly. The addVpc function is now using vpc.env.region if it is set, but the problem is that it is not set correctly to match the value in cdk.context.json and set the region from the stack instead.

### Expected Behavior

The template should have the correction region set for the VPC.

```
  "PrivateDnsE0FF4F9B": {
   "Type": "AWS::Route53::HostedZone",
   "Properties": {
    "Name": "slurmiad.local.",
    "VPCs": [
     {
      "VPCId": "vpc-nnnnnnnnnnnnnnnnn",
      "VPCRegion": "us-east-1"
     },
     {
      "VPCId": "vpc-nnnnnnnnnnnnnnnnn",
      "VPCRegion": "us-west-2"
     },
     {
      "VPCId": "vpc-nnnnnnnnnnnnnnnnn",
      "VPCRegion": "eu-west-1"
     }
    ]
   },
   "Metadata": {
    "aws:cdk:path": "slurmiad/PrivateDns/Resource"
   }
  },
```

### Current Behavior

The resulting CFN template has the correct VpcIds, but the incorrect regions:

```
  "PrivateDnsE0FF4F9B": {
   "Type": "AWS::Route53::HostedZone",
   "Properties": {
    "Name": "slurmiad.local.",
    "VPCs": [
     {
      "VPCId": "vpc-nnnnnnnnnnnnnnnnn",
      "VPCRegion": "us-east-1"
     },
     {
      "VPCId": "vpc-nnnnnnnnnnnnnnnnn",
      "VPCRegion": "us-east-1"
     },
     {
      "VPCId": "vpc-nnnnnnnnnnnnnnnnn",
      "VPCRegion": "us-east-1"
     }
    ]
   },
   "Metadata": {
    "aws:cdk:path": "slurmiad/PrivateDns/Resource"
   }
  },
```

### Reproduction Steps

I'm creating a Route53.HostedZone for use in 3 VPCs that are located in 3 different regions.
The VPCs aren't part of the stack and are created in CDK using 

```
                ec2.Vpc.from_lookup(
                    self, f"Vpc{region_dict['Region']}",
                    region = region_dict['Region'],
                    vpc_id = region_dict['VpcId'])
```

[...]

PR #25098, opened 2023-04-13
fix(ec2): Vpc lookup is setting region from stack

daschaa:
Reopens: [#24459](https://github.com/aws/aws-cdk/pull/24459)
Fixes [#22178](https://github.com/aws/aws-cdk/issues/22178)

This PR tackles the issue that the VPC fromLookup method was not setting the region correctly, even if passed in the lookup method.
The issue was, that in the dummy response from the lookup method, the region was not set.

----


daschaa:
@comcalvi I had to reopen this, because the original PR was closed due to staleness. Sorry for the inconvenience!

PR #25112, opened 2023-04-13
fix(appmesh): add missing `port` property

### LINKED ISSUE #22452: aws appmesh(module name): (short issue description)
### Describe the bug

Unable to add match port 

I'm defining a gateway route like 
 const route = gateway?.addGatewayRoute(
        `${this.props.containerApplicationName}-web-route`,
        {
          routeSpec: appmesh.GatewayRouteSpec.http({
            routeTarget: containerApplicationService,
            match: {
              hostname: GatewayRouteHostnameMatch.exactly(
                `${
                  this.props.containerApplicationName
                }.domainName`
              ),
              rewriteRequestHostname: true,
            },
          }),
        }
      );
I get this error Gateway route must define a match port if the parent Virtual Gateway has multiple listeners
I'm using the same virtual gateway for multiple for multiple virtual service routers.

### Expected Behavior

automatically associate router ports

### Current Behavior

Gateway route must define a match port if the parent Virtual Gateway has multiple listeners. (Se
rvice: AWSAppMesh; Status Code: 400; Error Code: BadRequestException; Request ID: 2da86d7b-7348-
4345-b25d-6804389a0e01; Proxy: null)

### Reproduction Steps

const containerPorts = props.containers!.map((container) => {
      return { containerPort: parseInt(container.containerApi.port) };
    });
    GenericContainerServiceConstruct.externalPortMappings = containerPorts;

const gateway = new appmesh.VirtualGateway(scope, `${id}-virtual-gateway`, {
      mesh: GenericContainerServiceConstruct.mesh,
      listeners: containerPorts .map((port) => {
        return appmesh.VirtualGatewayListener.http({ port });
      }),
    });

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.4

### Framework Version

_No response_

### Node.js Version

16

### OS

AMI linux 2

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

sumupitchayan:
Closes #22452 

Adds `port` property to the `HttpGatewayRouteMatch` interface along with a unit test to validate the change.

----


corymhall (changes requested):
@sumupitchayan great job! The only thing I think we should add is to add some validation that will throw an error if the user has not set `port` when required. The linked issues seems to indicate that the match port is required when the virtual gateway has >=2 listeners or else it will throw an error like:

```Gateway route must define a match port if the parent Virtual Gateway has multiple listeners.```

I think we can add a validation somewhere in [this class](https://github.com/aws/aws-cdk/blob/428bfc2dc566cff17b5856b2b346e473ea3642c4/packages/aws-cdk-lib/aws-appmesh/lib/gateway-route.ts#L111). It might look something like:

```ts
    this.node.addValidation({
      validate() {
        if (Resource.isOwnedResource(props.virtualGateway)) {
          const cfnGateway = props.virtualGateway.node.defaultChild as CfnVirtualGateway;
          const listeners = (cfnGateway.spec as CfnVirtualGateway.VirtualGatewaySpecProperty).listeners;
          if (Array.isArray(listeners) && listeners.length > 1) {
            return ['some error message'];
          }

        }
        return [];
      },
    });
```

sumupitchayan:
Thanks for the feedback @corymhall ! Just added the validation - let me know if you spot anything else to improve!

sumupitchayan:
> Thanks for the feedback @corymhall ! Just added the validation - let me know if you spot anything else to improve!

xzx

> @sumupitchayan great job! The only thing I think we should add is to add some validation that will throw an error if the user has not set `port` when required. The linked issues seems to indicate that the match port is required when the virtual gateway has >=2 listeners or else it will throw an error like:
> 
> `Gateway route must define a match port if the parent Virtual Gateway has multiple listeners.`
> 
> I think we can add a validation somewhere in [this class](https://github.com/aws/aws-cdk/blob/428bfc2dc566cff17b5856b2b346e473ea3642c4/packages/aws-cdk-lib/aws-appmesh/lib/gateway-route.ts#L111). It might look something like:
> 
> ```ts
>     this.node.addValidation({
>       validate() {
>         if (Resource.isOwnedResource(props.virtualGateway)) {
>           const cfnGateway = props.virtualGateway.node.defaultChild as CfnVirtualGateway;
>           const listeners = (cfnGateway.spec as CfnVirtualGateway.VirtualGatewaySpecProperty).listeners;
>           if (Array.isArray(listeners) && listeners.length > 1) {
>             return ['some error message'];
>           }
> 
>         }
>         return [];
>       },
>     });
> ```

corymhall:
> Thanks for the feedback @corymhall ! Just added the validation - let me know if you spot anything else to improve!

Can you also add a test for that validation case? You can assert that a validation throws an error by doing

```ts
expect(() => app.synth()).toThrow(/the error message/);
```

corymhall (approved):
@sumupitchayan great job 🚀

PR #25113, opened 2023-04-13
chore: upgrade jest to ^29

corymhall:
Now that we are on typescript >4.2 we can upgrade to jest 29.

- Updated some snapshots because the snapshot format in jest 29 changed.
- Cleaned up some test output by spying logs

All other updates were just as a result of the version update

----

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/README.md
+++ b/packages/@aws-cdk-testing/cli-integ/README.md
@@ -3,13 +3,7 @@
 
 ---
 
-![cdk-constructs: Experimental](https://img.shields.io/badge/cdk--constructs-experimental-important.svg?style=for-the-badge)
-
-> The APIs of higher level constructs in this module are experimental and under active development.
-> They are subject to non-backward compatible changes or removal in any future version. These are
-> not subject to the [Semantic Versioning](https://semver.org/) model and breaking changes will be
-> announced in the release notes. This means that while you may use them, you may need to update
-> your source code when upgrading to a newer version of this package.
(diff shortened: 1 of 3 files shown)


packages/aws-cdk-lib/core/test/validation/validation.test.ts
```diff
@@ -50,6 +53,8 @@ describe('validations', () => {
     app.synth();
     expect(process.exitCode).toEqual(1);
 
+    // eslint-disable-next-line no-console
```
vinayak-kukreja, line 56: I believe this can be removed :)

vinayak-kukreja (changes requested):

packages/aws-cdk-lib/core/test/validation/validation.test.ts
corymhall, line 56:
```suggestion
```

corymhall: Good catch! Updated.

packages/aws-cdk-lib/core/test/validation/validation.test.ts
```diff
@@ -50,6 +53,7 @@ describe('validations', () => {
     app.synth();
     expect(process.exitCode).toEqual(1);
 
+    console.log(consoleErrorMock.mock.calls[0][0]);
```
corymhall, line 56:
```suggestion
```

vinayak-kukreja (approved):

PR #25116, opened 2023-04-13
feat(kms): add required aliasname prefix to aliasnames with tokens

### LINKED ISSUE #25033: (kms): add required aliasname prefix to aliasnames that don't start with a token
### Describe the feature

Currently, we add the `REQUIRED_ALIAS_PREFIX` to the alias name if it is not present on non-token names only:

```ts
if (!Token.isUnresolved(aliasName)) {
      if (!aliasName.startsWith(REQUIRED_ALIAS_PREFIX)) {
        aliasName = REQUIRED_ALIAS_PREFIX + aliasName;
      }

      if (aliasName === REQUIRED_ALIAS_PREFIX) {
        throw new Error(`Alias must include a value after "${REQUIRED_ALIAS_PREFIX}": ${aliasName}`);
      }

      if (aliasName.toLocaleLowerCase().startsWith(DISALLOWED_PREFIX)) {
        throw new Error(`Alias cannot start with ${DISALLOWED_PREFIX}: ${aliasName}`);
      }

      if (!aliasName.match(/^[a-zA-Z0-9:/_-]{1,256}$/)) {
        throw new Error('Alias name must be between 1 and 256 characters in a-zA-Z0-9:/_-');
      }
    }
```

This means that you're on your own if you specify a tokenized aliasName. If you fail to specify the correct required prefix in the resolved token, you get the worlds worst error message at deploy time.

```bash
"Model validation failed (#/AliasName: failed validation constraint for keyword [pattern])"
```

### Use Case

I could be writing something like this:

```ts
new kms.Key(this, 'key', {
  alias: `MyKey${this.account}`,
});
```

If this is specified in a fixed-env account, then `this.account` actually gets resolved to the account number and the alias is not a token. Thus, the `REQUIRED_ALIAS_PREFIX` gets added and deploy succeeds.

If this is specified in a env-agnostic account, then `this.account` is still a token at this point and the `REQUIRED_ALIAS_PREFIX` is _not_ added. In this case, we get the cryptic deploy time error message from kms.

### Proposed Solution

We could improve this experience (given how busted the kms error message is) by deconstructing the token and checking to see if the _start_ of the string is tokenized or not. If it is not, we can safely check to see if it begins with the `REQURIED_ALIAS_PREFIX` and add it in if it is not.

We can do this deconstruction with `Tokenization.reverseString()` I believe.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.73

MacOs

daschaa:
Closes #25033.

----


daschaa:
Exemption Request: Since it is not mentioned in the Readme that the Alias is extended with the prefix if not set, I guess no additional entry has to be added for this feature.
Please let me know if i should extend the readme. :)

corymhall (approved):

Issue #21364, opened 2022-07-28
LookupMachineImage: Causes Invalid assembly manifest error


Console32:
### Describe the bug

I want to switch from a static AMI Map to a lookup. But when I use a Lookup Image, stack results in errors

current:
```
const image = MachineImage.genericLinux({
      [props.awsRegion]: props.ami,
    });
```

desired:
```
    const image = new LookupMachineImage({ name: 'whatever' } );
```

### Expected Behavior

cdk synth works

### Current Behavior

```
Error: Invalid assembly manifest:
instance.missing[0].props.account is not of a type(s) string
instance.missing[0].props.account is not of a type(s) string
instance.missing[0].props.account is not of a type(s) string
instance.missing[0].props requires property "domainName"
instance.missing[0].props.account is not of a type(s) string
instance.missing[0].props requires property "parameterName"
instance.missing[0].props.account is not of a type(s) string
instance.missing[0].props requires property "filter"
instance.missing[0].props.account is not of a type(s) string
instance.missing[0].props requires property "serviceName"
instance.missing[0].props.account is not of a type(s) string
instance.missing[0].props requires property "loadBalancerType"
instance.missing[0].props.account is not of a type(s) string
instance.missing[0].props requires property "loadBalancerType"
instance.missing[0].props.account is not of a type(s) string
instance.missing[0].props.account is not of a type(s) string
instance.missing[0].props requires property "aliasName"
instance.missing[0].props requires property "pluginName"
instance.missing[0].props is not any of <#/definitions/AmiContextQuery>,<#/definitions/AvailabilityZonesContextQuery>,<#/definitions/HostedZoneContextQuery>,<#/definitions/SSMParameterContextQuery>,<#/definitions/VpcContextQuery>,<#/definitions/EndpointServiceAvailabilityZonesContextQuery>,<#/definitions/LoadBalancerContextQuery>,<#/definitions/LoadBalancerListenerContextQuery>,<#/definitions/SecurityGroupContextQuery>,<#/definitions/KeyContextQuery>,<#/definitions/PluginContextQuery>

```

### Reproduction Steps

I am not able to reproduce this in a small example, can assist in pinning down the issue though. Just not sure where to start.

### CDK CLI Version

2.33.0 (build 859272d)

### Node.js Version

v16.15.1

### OS

Windows 11

### Language

Typescript

Console32:
Changed Images to some prebuilt AMI's in a shared account and do now get the same error. Any clue where to start?

[...]

PR #25134, opened 2023-04-14
fix: stack account id throws error if not a string

vinayak-kukreja:
If the account id passed in as stack environment is not a string, then that leads to errors in some lookup functions. This PR is adding a type check for stack's account id and region. These are only allowed to be `string` datatype. 

Closes https://github.com/aws/aws-cdk/issues/21364

----


vinayak-kukreja:
Exemption Request

I believe the unit test already cover the scenario that causes the error. If an env account id or region is not a string, now it should not be able to deploy so integ test would not make sense here. Let me know if that is not the case.

PR #12416, opened 2021-01-08
fix(aws-rds): correct Policy resource for Proxy::grantConnect()

jdvornek:
fixes #12415

To generate the correct policy, the DatabaseProxy ARN is parsed
and the resulting components are used along with a new parameter
to grantConnect.

The unit test was updated and passes. Caveat lector, I was not
able to get a full docker build or a full local build to work on
my box.

I'm not sure if this should be considered a breaking change. While
it technically alters the functionality of a published function,
the current behavior provides no utility.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #25141, opened 2023-04-15
fix(rds): Correct ARN in IAM policy for IAM database access

### LINKED ISSUE #11851: (awards): grantConnect for IAM authentication provides invalid permissions
`DatabaseInstance` has a method [grantConnect](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-rds.DatabaseInstance.html#grantwbrconnectgrantee) for granting `connect` access to instance using IAM based authentication. 

However, the db resource ARN in the produced IAM policy is incorrect and doesn't work. Based on the [documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.IAMPolicy.html) The format for the resource in the IAM policy should be: `arn:aws:rds-db:region:account-id:dbuser:DbiResourceId/db-user-name`

The actual resource produced by `grantConnect` is having format: `arn:aws:rds:region:account-id:db:DBInstanceId`. Also, the function doesn't provide any parameter to define the db username to be used in the policy. 

### Reproduction Steps

```
import { Stack, Construct, StackProps } from '@aws-cdk/core';
import {
  DatabaseInstance,
  DatabaseInstanceEngine,
  PostgresEngineVersion,
  Credentials,
} from '@aws-cdk/aws-rds';

import { IVpc } from '@aws-cdk/aws-ec2';
import { User } from '@aws-cdk/aws-iam';

export interface MyStackProps extends StackProps {
  vpc: IVpc;
}
  constructor(scope: Construct, id: string, props: MyStackProps) {
    super(scope, id, props);

    const db = new DatabaseInstance(this, 'Instance', {
      engine: DatabaseInstanceEngine.postgres({ version: PostgresEngineVersion.VER_12_4 }),
      credentials: Credentials.fromGeneratedSecret('testuser'),
      vpc: props.vpc,
      port: 5432,
      iamAuthentication: true,
    });

    const user = new User(this, 'TestUser', {
      userName: 'testuser',
    });

    db.grantConnect(user);
  }
}

```

To create a IAM policy where the resource ARN would be according to the documentation i.e.  `arn:aws:rds-db:region:account-id:dbuser:DbiResourceId/db-user-name`

### What actually happened?

Instead of the correct policy, the generated template contains following definition:
```
           {
              "Action": "rds-db:connect",
              "Effect": "Allow",
              "Resource": {
                "Fn::Join": [
                  "",
                  [
                    "arn:",
                    {
                      "Ref": "AWS::Partition"
                    },
                    ":rds:",
                    {
                      "Ref": "AWS::Region"
                    },
                    ":",
                    {
                      "Ref": "AWS::AccountId"
                    },
                    ":db:",
                    {
                      "Ref": "InstanceC1063A87"
                    }
                  ]
                ]
              }
            }
```

In addition that the format of the ARN is incorrect, also wrong DB identifier is used. The template uses the `DB Instance id` but the correct identifier is the `DB Resource id`. 
 

### Environment

  - **CDK CLI Version  :** 1.75.0
  - **Framework Version:** 1.75.0
  - **Node.js Version:** 12.18.3
  - **OS               :** OSx
  - **Language (Version):**  Typescript 4.1.2

### Other

The support for the grantConnect was requested in this [issue](https://github.com/aws/aws-cdk/issues/1558) and added in this [PR](https://github.com/aws/aws-cdk/pull/9887).

A [comment](https://github.com/aws/aws-cdk/issues/1558#issuecomment-454988630) in the original issue still stands i.e. that the DB Resource Id is not accessible in Cloudformation. 

--- 

This is :bug: Bug Report

akash1810:
The [IAM policy for IAM database access](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.IAMPolicy.html) takes the form of:

```
arn:aws:rds-db:region:account-id:dbuser:DbiResourceId/db-user-name
```

Following https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/105, this change updates the ARN used in `grantConnect` to this format.

Additionally, update the signature of `grantConnect` to take an optional `dbUser`, which is defaulted to the master username of the database, obtained via the available `Secret`.

This signature change also matches `grantConnect` in `DatabaseProxy`. See #12416.

Closes #11851.

----


akash1810:
Clarification Request.

packages/aws-cdk-lib/aws-rds/lib/instance.ts
```diff
@@ -130,6 +147,7 @@ export abstract class DatabaseInstanceBase extends Resource implements IDatabase
       public readonly instanceEndpoint = new Endpoint(attrs.instanceEndpointAddress, attrs.port);
       public readonly engine = attrs.engine;
       protected enableIamAuthentication = true;
+      public readonly instanceResourceId = attrs.instanceResourceId;
```
akash1810, line 150: This is a new, required property. Is it a breaking change for `fromDatabaseInstanceAttributes`?

akash1810: > This is a new, required property. Is it a breaking change for `fromDatabaseInstanceAttributes`?

Ahh the AWS CodeBuild report suggests it is:

```log
err  - IFACE aws-cdk-lib.aws_rds.DatabaseInstanceAttributes: newly required property 'instanceResourceId' added: input to aws-cdk-lib.aws_rds.DatabaseInstanceBase.fromDatabaseInstanceAttributes [strengthened:aws-cdk-lib.aws_rds.DatabaseInstanceAttributes]
Some packages seem to have undergone breaking API changes. Please avoid.
```

Advice on how to proceed welcomed 🙏🏽.

corymhall: I think you can change this to an optional property in `IDatabaseInstance` and `DatabaseInstanceBase` and `fromDatabaseInstanceAttributes`.

akash1810: > I think you can change this to an optional property in `IDatabaseInstance` and `DatabaseInstanceBase` and `fromDatabaseInstanceAttributes`.

Done in b612e7ec9ba873e5d652001c10e7bcbe3e7afdad. I'm not a huge fan of this as it means one has to ensure the property is not `undefined` when accessing it, which adds bloat IMO:

```ts
const db = new DatabaseInstance();
const resourceId: string | undefined = db.instanceResourceId;

if (!resourceId) {
  // unhappy path
}
else {
  // happy path
}
```

Alternatively, one can use [non-null assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator), however I've worked in projects where this is yields warnings from the linter.

Additionally, this being optional makes it the [only optional property](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.DatabaseInstance.html#properties). Not a problem in general, however might be an indication of not being desired?

akash1810:
Exemption Request

Unit tests have been updated, and an additional one added.

packages/aws-cdk-lib/aws-rds/lib/instance.ts
corymhall: Yeah it's not great, but it's the only way to avoid breaking changes.

corymhall:
> 
> ```
> ```
> 

Integration tests are now added in `@aws-cdk-testing/framework-integ`. In this case I think the integ test would only be useful if we created an assertion test where something (i.e. lambda function) was able to connect to the database, but I think the effort to create that case is too large so I'm exempting it.

corymhall (approved):
🚀

corymhall:
@akash1810 you need to authorize modification on the head branch in order for mergify to merge this.

akash1810:
> @akash1810 you need to authorize modification on the head branch in order for mergify to merge this.

Ah. Done that now. Thanks!

corymhall:
@Mergifyio requeue

ChrisLahaye:
We are experiencing some issues after upgrading to v2.77.0 which I believe is due to this change.

When using `databaseInstance.grantConnect(iamRole)` the following error is thrown:
```
Error: Resolution error: Resolution error: Resolution error: Resolution error: Synthing a secret value to Resources/${Token[xxx.ServiceAccount.Role.DefaultPolicy.Resource.LogicalID.2000]}/Properties/policyDocument/Statement/0/Resource. Using a SecretValue here risks exposing your secret. Only pass SecretValues to constructs that accept a SecretValue property, or call AWS Secrets Manager directly in your runtime code. Call 'secretValue.unsafeUnwrap()' if you understand and accept the risks.
```

I think this happens due to the following new code:
```
dbUser = this.secret.secretValueFromJson('username').toString();
```

Perhaps this should be unsafeUnwrap instead of toString?

nick-asher:
Don't suppose there's any appetite to populate this property when fetching a database via FindDatabases()? There's no clear way defined to actually populate this property in this commit unless I'm missing something? The one comment in the docs about it being logged in CloudTrail doesn't feel like a sensible way to fetch the Id.

PR #25160, opened 2023-04-17
fix(aws-cdk-lib): compiled .js files are no longer being minified

madeline-k:
Prior to the aws-cdk repo remodel, there was a script executed at the end of build to minify the `.js` files. After the remodel, that script was excluded. In this change, I am adding it back in as a pre-package step. There is no need to perform this as part of the build. The `minify-sources` script takes about 2 seconds to run. 

The size of aws-cdk-lib v 2.74 is 222 MB. This change will bring the overall size down to 205 MB. 

----


MrArnoldPalmer (approved):

madeline-k:
The PR Linter should not be complaining for this one. I am working on updating that

madeline-k:
PR Linter change to unblock this PR: https://github.com/aws/aws-cdk/pull/25181

PR #25165, opened 2023-04-17
fix(servicecatalogappregistry): Revert deprecated method removing PR to keep deprecated method in alpha version

yingdon-amazon:
Revert PR of deprecating method AssociateAttributeGroup and AssociateStack since we would like to allow our v1 customer to continue to use deprecated method in v2 alpha version to offer them more time  to switch to stable version. We will remove these deprecated method in stable version

This reverts commit [9222f217aa5f199dd88addc9d0dee0f3f900b5a2.](https://github.com/aws/aws-cdk/commit/9222f217aa5f199dd88addc9d0dee0f3f900b5a2)

----

### DIFF
--- a/packages/@aws-cdk/aws-servicecatalogappregistry-alpha/lib/application.ts
+++ b/packages/@aws-cdk/aws-servicecatalogappregistry-alpha/lib/application.ts
@@ -57,12 +57,28 @@ export interface IApplication extends cdk.IResource {
   readonly applicationName?: string;
 
   /**
+   * Associate this application with an attribute group.
+   *
+   * @param attributeGroup AppRegistry attribute group
+   */
+  associateAttributeGroup(attributeGroup: IAttributeGroup): void;
+
+  /**
    * Create an attribute group and associate this application with the created attribute group.
    *
    * @param id name of the AttributeGroup construct to be created.
    * @param attributeGroupProps AppRegistry attribute group props
    */
   addAttributeGroup(id: string, attributeGroupProps: AttributeGroupAssociationProps): IAttributeGroup;
+
+  /**
+   * Associate this application with a CloudFormation stack.
+   *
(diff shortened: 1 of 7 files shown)


corymhall:
> Revert PR of deprecating method AssociateAttributeGroup and AssociateStack since we would like to allow our v1 customer to continue to use deprecated method in v2 alpha version to offer them more time to switch to stable version. We will remove these deprecated method in stable version

I'm not sure I understand. Are you saying this will allow customers to more easily migrate from CDK v1 => v2?

yingdon-amazon:
> > Revert PR of deprecating method AssociateAttributeGroup and AssociateStack since we would like to allow our v1 customer to continue to use deprecated method in v2 alpha version to offer them more time to switch to stable version. We will remove these deprecated method in stable version
> 
> I'm not sure I understand. Are you saying this will allow customers to more easily migrate from CDK v1 => v2?

yes, you are right. the main goal of getting it back is to offer a better customer experience when v1 customer is doing migration from v1 to v2

**Here is some context about this change:**
Recently, we found that we do not mark these two methods as deprecated in v1 cdk. Also, we cannot update v1 cdk now to send deprecation notification to them. Thus, v1 customer cannot receive any notification for the methods’ deprecation forever.
Therefore, if we directly remove these methods in v2, then v1 customers have to update their code to depreciate these two methods. We think that this migration’s customer experience is not good.

In order to offer a better customer experience to customers, we decide to keep deprecated method in alpha version. Thus, if v1 customs do not want to update their code to deprecate the methods when they want to migrate to v2, they can use our v2 alpha version temporarily

corymhall:
@yingdon-amazon I don't think this is something we want to do. The decision was made to remove these methods so we should stick to that. And anyway you will have to remove these methods before going to stable anyway so the latest alpha version will always have these methods removed.

Issue #24236, opened 2023-02-20
aws_rds: DatabaseCluster deletion_protection does not default to false


rittneje:
### Describe the bug

The CDK docs say the following:

> deletion_protection (Optional[bool]) – Indicates whether the DB cluster should have deletion protection enabled. Default: - true if removalPolicy is RETAIN, false otherwise

### Expected Behavior

If I first deploy a stack with that flag set to `True`, and then I deploy a stack update with the flag not set at all, it should disable deletion protection, as the docs claim.

### Current Behavior

CloudFormation instead stops tracking the deletion protection property.

### Reproduction Steps

Deploy a stack with a `DatabaseCluster` like so:

```python
aws_rds.DatabaseCluster(stack, "Database", ..., deletion_protection=True, ...)
```

Then remove the `deletion_protection` argument from the constructor. Look at the output of `cdk diff`. Then deploy the stack and confirm in the RDS console that deletion protection is still enabled.

### Possible Solution

CDK is not passing the `DeletionProtection` property in the CloudFormation template at all if the `deletion_protection` parameter is not passed. So the docs are really describing CloudFormation's default behavior, which is only applicable to creating a new resource.

Instead, either CDK MUST always set the `DeletionProtection` flag in the stack template to abide by its current documentation, or the documentation needs to be amended to distinguish resource creation from update.

### CDK CLI Version

2.60.0

### Node.js Version

v18.13.0

### OS

Alpine 3.17

### Language

Python

### Language Version

3.11.1

pahud:
Yes. Database replacement is something we definitely should avoid. I'll bring this to the team for more visibility. Any additional comments or ideas are appreciated here.

pahud:
According to this, I think by default `deletionProtection` is actually `undefined`, which is equivalent to `false` only when the resource is initially created:

https://github.com/aws/aws-cdk/blob/8492dfefd209dfcceb528bc30dea4be55eaa66c7/packages/aws-cdk-lib/aws-rds/lib/private/util.ts#L77-L83

However, if you update the resource from explicit `deletionProtection: true` to `deletionProtection: undefined` and `removalPolicy != RETAIN`, cloudformation seems to just leave it as is and not set it to `false`, which explains we are seeing this value remains `true` in the console.

I think we should update the document to clarify this.

PR #25169, opened 2023-04-18
docs(rds): improve the doc for deletionProtection

pahud:
The undefined default of `deletionProtection` property is confusing especially when we update the cluster by removing this property. This PR improves the description.

Closes https://github.com/aws/aws-cdk/issues/24236

----


packages/aws-cdk-lib/aws-rds/lib/cluster.ts
```diff
@@ -103,7 +103,9 @@ interface DatabaseClusterBaseProps {
   /**
    * Indicates whether the DB cluster should have deletion protection enabled.
    *
-   * @default - true if ``removalPolicy`` is RETAIN, false otherwise
+   * @default - true if ``removalPolicy`` is RETAIN, `undefined` otherwise, which will not enable the protection
+   * when initially create the cluster. To update the cluster and disable it, you will need to
+   * explicitly update this value from `true` to `false`.
```
corymhall, line 108:
```suggestion
   * @default - true if `removalPolicy` is RETAIN, `undefined` otherwise, which will not enable deletion protection.
   * To disable deletion protection after it has been enabled, you must explicitly set this value to `false`.
```

corymhall (approved):
Just a minor suggestion. Feel free to take it or leave it and remove the
do-not-merge label :)

Issue #24808, opened 2023-03-27
aws-eks: KubectlProvider fails to create/update in govcloud


wongma7:
### Describe the bug

In govcloud (aws-us-gov partition), "[Amazon ECR public registries](https://docs.aws.amazon.com/AmazonECR/latest/public/public-registries.html) aren't supported." https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-ecr.html

The managed policy "AmazonElasticContainerRegistryPublicReadOnly" does not exist there.

https://github.com/aws/aws-cdk/pull/24104 added this managed policy "AmazonElasticContainerRegistryPublicReadOnly" and now KubectlProvider HandlerServiceRole fails to create/update.

### Expected Behavior

KubectlProvider HandlerServiceRole successfully creates/updates in govcloud same as in other AWS partitions

### Current Behavior

KubectlProvider HandlerServiceRole fails to create/update in govcloud

### Reproduction Steps

Create a 1.25 EKS cluster in govcloud*
```
const cluster = new eks.Cluster(this, 'hello-eks', {
  version: eks.KubernetesVersion.V1_25,
});
```

### Possible Solution

Wrap this:
```
    // For OCI helm chart public ECR authorization.
    this.handlerRole.addManagedPolicy(
      iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonElasticContainerRegistryPublicReadOnly'),
    );
```
in a partition check

### CDK CLI Version

~~2.27.0 (build 8e89048)~~

    "aws-cdk": "^2.69.0",

### Framework Version

    "aws-cdk": "^2.69.0",

### Node.js Version

v18.14.0

### OS

linux

### Language

Typescript

pahud:
closed by https://github.com/aws/aws-cdk/pull/25215

PR #25170, opened 2023-04-18
fix(eks): policy does not exist or is not attachable in China and GovCloud regions

pahud:
Reopening this PR and closing the [previous one](https://github.com/aws/aws-cdk/pull/24759). 

As ECR Public is not available in China regions and GovCloud, `AmazonElasticContainerRegistryPublicReadOnly` IAM managed policy would not be available in those affected regions and should not be attached to the role. This PR implements a RegionInfo method to determine if ECR public is available in the deploying region and conditionally attach `AmazonElasticContainerRegistryPublicReadOnly` to the role. If region can't be determined in synth time, we assume ECR Public is not available to avoid potential failure.

Closes https://github.com/aws/aws-cdk/issues/24743 https://github.com/aws/aws-cdk/issues/24808 https://github.com/aws/aws-cdk/issues/25178

----


packages/aws-cdk-lib/region-info/build-tools/generate-static-data.ts
```diff
@@ -56,6 +56,7 @@ export async function main(): Promise<void> {
     registerFact(region, 'DOMAIN_SUFFIX', domainSuffix);
 
     registerFact(region, 'CDK_METADATA_RESOURCE_AVAILABLE', AWS_CDK_METADATA.has(region) ? 'YES' : 'NO');
+    registerFact(region, 'ECR_PUBLIC_AVAILABLE', partition === 'aws' ? 'YES' : 'NO');
```
iliapolo, line 59: Did you consider using the `AWS::Partition` pseudo param in conjunction with `Fn::If` to make this more environment agnostic? its a bummer the user must specify a `region` in the stack in order to get this functionality.

packages/aws-cdk-lib/region-info/lib/region-info.ts
```diff
@@ -70,6 +70,13 @@ export class RegionInfo {
     return Fact.find(this.name, FactName.CDK_METADATA_RESOURCE_AVAILABLE) === 'YES';
   }
 
+  /**
+   * Whether Amazon ECR Public is available in this region or not.
+   */
+  public get ecrPublicAvailable(): boolean {
+    return Fact.find(this.name, FactName.ECR_PUBLIC_AVAILABLE) === 'YES';
```
iliapolo, line 77: So this would mean if a user doesn't specify a region in the stack, which is the common case, we now change the behavior and remove the policy and remove the ecr authorization - this might have a some implications no? We should default to the common case, and make the user "work" a little harder for the uncommon one.

PR #25181, opened 2023-04-18
chore(prlint): allow 'aws-cdk-lib' as a pr title scope

madeline-k:
If a PR applies to aws-cdk-lib as a whole, we want to label it as such. Today, if you have a PR titled "fix(aws-cdk-lib): some change", the PR linter will include this message:

```
❌ The title of the pull request should omit 'aws-' from the name of modified packages. Use 'cdk-lib' instead of 'aws-cdk-lib'.
```

----


tools/@aws-cdk/prlint/test/lint.test.ts
```diff
@@ -142,8 +156,8 @@ describe('commit message format', () => {
   })
 });
 
-describe.only('ban breaking changes in stable modules', () => {
-  test.only('breaking change in stable module', async () => {
+describe('ban breaking changes in stable modules', () => {
+  test('breaking change in stable module', async () => {
```
madeline-k, line 160: For some unknown reason, this test was marked as `only` which caused jest to skip all other tests in this file. Removing that here so that the unit tests in this file actually get run.

MrArnoldPalmer: Whoops, that would be because I broke this test and was fixing it for remodel =( sorry

Issue #25178, opened 2023-04-18
eks: missing package.json in the integ test


pahud:
### Describe the bug

The `package.json` is missing in [aws-eks/test/sdk-call-integ-test-docker-app/app/](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/framework-integ/test/aws-eks/test/sdk-call-integ-test-docker-app/app) and the [DockerImageAssets](https://github.com/aws/aws-cdk/blob/8ca4c09817d03a094ac395f9ad1adace931f74b4/packages/%40aws-cdk-testing/framework-integ/test/aws-eks/test/integ.eks-service-account-sdk-call.ts#L18-L20) bundling from the integ testing is failing.

https://github.com/aws/aws-cdk/blob/8ca4c09817d03a094ac395f9ad1adace931f74b4/packages/%40aws-cdk-testing/framework-integ/test/aws-eks/test/sdk-call-integ-test-docker-app/app/Dockerfile#L8

The `package.json` was still existing back to [v2.72.1](https://github.com/aws/aws-cdk/blob/v2.72.1/packages/@aws-cdk/aws-eks/test/sdk-call-integ-test-docker-app/app/package.json) but for some reason it was removed. The integ test will fail if this file is missing.

### Expected Behavior

The integ test([sdk-call-integ-test-docker-app](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/framework-integ/test/aws-eks/test/sdk-call-integ-test-docker-app)) should succeed.

### Current Behavior

The integ test([sdk-call-integ-test-docker-app](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/framework-integ/test/aws-eks/test/sdk-call-integ-test-docker-app)) fails due to missing `package.json` and can't bundle the docker image assets.

### Reproduction Steps

`yarn integ test/sdk-call-integ-test-docker-app.js --force` to run the integ test.

### Possible Solution

Add the missing `package.json` back.

### CDK CLI Version

2.72.1

### Node.js Version

v18.15.0

### OS

mac os x

### Language

Typescript

pahud:
closed by https://github.com/aws/aws-cdk/pull/25448

Issue #24358, opened 2023-02-28
(custom-resources): StateNotFoundError: State functionActiveV2 not found in ap-south-2


pahud:
### Describe the bug

When deploying 10+ custom resources in `ap-south-2`, it fails with `StateNotFoundError: State functionActiveV2 not found` error as below:

1. It's fine to deploy 10+ custom resources in other AWS regions such as `us-east-1` or `ap-northeast-1`. Only `ap-south-2` fails in this case.
2. When deploying with <10 custom resources, `ap-south-2` will be fine with no error.

### Expected Behavior

The provided code above should deploy in `ap-south-2`.

### Current Behavior

<img width="1172" alt="image" src="https://user-images.githubusercontent.com/278432/221716895-f414c22f-2da8-47a9-95e8-e77cddf7f42b.png">

### Reproduction Steps

```ts
import * as cdk from "aws-cdk-lib";
import { Construct } from "constructs";
import * as lambda from "aws-cdk-lib/aws-lambda";
import { CustomResource } from "aws-cdk-lib";
import * as cr from "aws-cdk-lib/custom-resources";

export class DummyLambdaStack extends cdk.Stack {
  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
    super(scope, id, props);

    const onEventHandler = new lambda.Function(this, 'OnEventHandler', {
      runtime: lambda.Runtime.NODEJS_18_X,
      code: lambda.Code.fromInline(`exports.onEvent = () => { return }`),
      handler: 'index.onEvent',
    })

    const provider = new cr.Provider(this, "Provider", {
      onEventHandler,
    });

    const numResources = 10
    for (let i = 0; i < numResources; i++) {
      new CustomResource(this, `CR${i}`, {
        serviceToken: provider.serviceToken,
      });
    }
  }
}
```

### Possible Solution

There might be some restrictions in `ap-south-2`.

### CDK CLI Version

2.66.1 (build 539d036)

### Node.js Version

v16.17.0

### OS

Linux

### Language

Typescript

pahud:
Same error if I change the runtime to Python 3.9

```ts
const onEventHandler = new lambda.Function(this, 'OnEventHandler', {
  runtime: lambda.Runtime.PYTHON_3_9,
  code: lambda.Code.fromInline(`def on_event(event, context): return {}`),
  handler: 'index.on_event',
})
```

zelu-zuehlke:
[...]

PR #25215, opened 2023-04-20
fix(eks): policy does not exist or is not attachable in China and GovCloud regions

pahud:
Reopening this PR because #25170 was closed by accident.

As ECR Public is not available in China regions and GovCloud, `AmazonElasticContainerRegistryPublicReadOnly` IAM managed policy would not be available in those affected regions and should not be attached to the role. This PR implements a CfnCondition to determine if ECR public is available based on `Aws.Partition` of the deploying region and conditionally attach `AmazonElasticContainerRegistryPublicReadOnly` to the kubectl-provider handler role. 

This PR has been tested in the following regions:

- [x] *cn-north-1
- [x] *cn-northwest-1
- [x] us-east-1

* I can confirm the role is created correctly in cn regions but due to 
   - #24358 
   - #24696  
The cluster and nodegroup are still failing to create in CN.

Closes https://github.com/aws/aws-cdk/issues/24743 https://github.com/aws/aws-cdk/issues/24808 https://github.com/aws/aws-cdk/issues/25178


packages/aws-cdk-lib/aws-eks/lib/kubectl-handler/helm/__init__.py
```diff
@@ -110,11 +115,16 @@ def get_oci_cmd(repository, version):
     elif public_registry['registry'] is not None:
         logger.info("Found AWS public repository, will use default region as deployment")
         region = os.environ.get('AWS_REGION', 'us-east-1')
-
-        cmnd = [
-            f"aws ecr-public get-login-password --region us-east-1 | " \
-            f"helm registry login --username AWS --password-stdin {public_registry['registry']}; helm pull {repository} --version {version} --untar"
-            ]
+        
+        if is_ecr_public_available(region):
```
corymhall, line 119: I'm not sure I understand what is happening here. From what I can follow this code would only be executed if the repository that is passed in is a public repository,  in which case just running `helm pull` wouldn't work right?

packages/aws-cdk-lib/aws-eks/lib/kubectl-provider.ts
```diff
@@ -165,10 +165,17 @@ export class KubectlProvider extends NestedStack implements IKubectlProvider {
...
-    );
+    /**
+     * For OCI helm chart public ECR authorization. As ECR public is only available in `aws` partition,
+     * we conditionally attach this policy when the AWS partition is `aws`.
+     */
+    const policyArn = iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonElasticContainerRegistryPublicReadOnly').managedPolicyArn;
+    const hasEcrPublicCondition = new CfnCondition(this, 'HasEcrPublic', {
+      expression: Fn.conditionEquals(Aws.PARTITION, 'aws'),
+    });
+    (this.handlerRole.node.defaultChild as iam.CfnRole).addPropertyOverride('ManagedPolicyArns.3',
```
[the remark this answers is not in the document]
pahud: AWS China doesn't have its own "ECR Public" service in either BJS(cn-north-1) or ZHY(cn-northwest-2) and all IAM actions starting with `ecr-public:*` are not available so we can't create a custom managed policy wrapping up all `ecr-public:*` actions. However, users in China still can docker pull images from ECR public at us-east-1, but only in unauthenticated way, which means they should not run `aws ecr-public get-login-password` in China unless they have AWS global credentials pre-configured.

packages/aws-cdk-lib/aws-eks/lib/kubectl-handler/helm/__init__.py
pahud: This means now that we are not in `aws` partition and probably in `aws-cn` partition but I need to helm install a chart located in ECR Public in us-east-1, which means:

1. I will need to helm pull that image straight from us-east-1 in `unauthenticated` way because I can't authenticate it in non `aws` partitions.
2. I can't run [helm registry login](https://helm.sh/docs/helm/helm_registry_login/) because I need to pull that image as unauthenticated.

packages/aws-cdk-lib/aws-eks/lib/kubectl-provider.ts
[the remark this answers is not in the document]
pahud: It worked! And I have tested in both `us-east-1` and `cn-north-1`.

packages/aws-cdk-lib/aws-eks/lib/kubectl-handler/helm/__init__.py
corymhall: ahh ok thanks for the clarification.

corymhall (approved):
🚀

justin007755:
Hello @pahud ,

We are still encountering below when using latest cdk version to create eks and corresponding resources like helm chart etc, and I tested cdk-2.65.0 which looks good, however, it's hard for us to use this cdk version considering other facts, so do we have a ETA or workaround for this issue?

-----------------------------------

2023-05-19 14:12:02 UTC+0800	HandlerServiceRoleFCDC14AE	
CREATE_FAILED   Policy arn:aws-cn:iam::aws:policy/AmazonElasticContainerRegistryPublicReadOnly does not exist or is not attachable. (Service: AmazonIdentityManagement; Status Code: 404; Error Code: NoSuchEntity; Request ID: 8a2723e1-3330-40e4-af9c-d45b6e6aa3b3; Proxy: null)

-----------------------------------

PR #25216, opened 2023-04-20
chore: DeletePolicy should default to retain

Naumel:
Considering the still volatile nature of the `import` to Cfn and CDK, the default for the required `DeletePolicy` should be to `Retain`. Any mistake that would create new resources can easily be removed with the stack, but we should not jeopardize existing, possible stateful ones.

[CONTRIBUTING GUIDE]: https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md
[DESIGN GUIDELINES]: https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md

Additional typos corrected.

----


Naumel:
exec. id: 001ff64d-00b4-447a-8580-20f24eb6f09e

PR #24316, opened 2023-02-24
fix(core): Configureautoscaling calling autoscaling connections securitygroups in Cluster (under feature flag)

homakk:
Configureautoscaling calling autoscaling connections securitygroups in Cluster (under feature flag)
image
>
> Describe the reason for this change, what the solution is, and any
> important design decisions you made. 
>
> Remember to follow the [CONTRIBUTING GUIDE] and [DESIGN GUIDELINES] for any
> code you submit.
>
> [CONTRIBUTING GUIDE]: https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md
> [DESIGN GUIDELINES]: https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md

Closes #22764
----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #25217, opened 2023-04-20
fix(core): Configureautoscaling calling autoscaling connections securitygroups in Cluster (under feature flag)

### LINKED ISSUE #22764: aws-ecs: Cluster.addAsgCapacityProvider doesn't add security groups to Cluster.connections like addAutoScalingGroup did
### Describe the issue

The docs for [cluster.addAutoScalingGroup](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-ecs.Cluster.html#addwbrautowbrscalingwbrgroupautoscalinggroup-optionsspan-classapi-icon-api-icon-deprecated-titlethis-api-element-is-deprecated-its-use-is-not-recommended%EF%B8%8Fspan) point towards using [cluster.addAsgCapacityProvider](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-ecs.Cluster.html#addwbrasgwbrcapacitywbrproviderprovider-options) instead, as the former is deprecated, and gone in v2.

They are nearly direct replacements.

The obvious - but wrong - replacement I started with was as follows:
Before:
```typescript
myCluster.addAutoScalingGroup(myAsg);
```
After:
```typescript
const myCapacityProvider = new ecs.AsgCapacityProvider(this, 'MyCapacityProvider', {
    autoScalingGroup: myAsg,
});
myCluster.addAsgCapacityProvider(myCapacityProvider);
```

However, this resulted in surprise issues with security groups elsewhere in my code (some code assumed an ECS Service had at least one security group in `.connections.securityGroups`, but it had zero, so errored.).

Turns out this is because `cluster.addAutoScalingGroup` adds security groups from the ASG's connections into the ECS Cluster's connections, but `cluster.addAsgCapacityProvider` does not.

See https://github.com/aws/aws-cdk/blob/c4e07a65f9580d3e39d4077d872bee6ad17d8924/packages/%40aws-cdk/aws-ecs/lib/cluster.ts#L385

So, in my example above, I needed to add:
```typescript
myCluster.connections.connections.addSecurityGroup(...myAsg.connections.securityGroups)
```

I don't really understand why this isn't done in `configureAutoScalingGroup` like all the other common setup is. I think it would be fine to add, as it won't break for someone who has already added the workaround like I have, as it seems that re-adding security groups to connections no-op (they don't get duplicated).

---

So I think one (or more) of the following could resolve this:
- the docs pointing from `addAutoScalingGroup` to `addAsgCapacityProvider` should highlight the ways in which they are different, and guide towards how to do a direct replacement.
- the options that can be passed to `addAsgCapacityProvider` should allow setting a `addSecurityGroupsFromAsgConnectionsToClusterConnections` or something, and when set to `true`, this does add security groups to the clusters connections.
- `Cluster.configureAutoScalingGroup` should call `this.connections.connections.addSecurityGroup(...autoScalingGroup.connections.securityGroups);`

---

Other related issues and PRs:
- https://github.com/aws/aws-cdk/pull/15107 this appears to have been an attempt to fix the issue, but the PR stalled
    - and the linked issue https://github.com/aws/aws-cdk/issues/14764
- https://github.com/aws/aws-cdk/issues/16260#issuecomment-937918710
- https://github.com/aws/aws-cdk/issues/15733

### Links

addAutoScalingGroup: https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-ecs.Cluster.html#addwbrautowbrscalingwbrgroupautoscalinggroup-optionsspan-classapi-icon-api-icon-deprecated-titlethis-api-element-is-deprecated-its-use-is-not-recommended%EF%B8%8Fspan

cluster.addAsgCapacityProvider: https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-ecs.Cluster.html#addwbrasgwbrcapacitywbrproviderprovider-options

homakk:
Configureautoscaling calling autoscaling connections securitygroups in Cluster (under feature flag)
image
>
>
> code you submit.
>

Closes #22764

----

### DIFF
--- a/packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
+++ b/packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
@@ -17,6 +17,7 @@ Flags come in three types:
 
 | Flag | Summary | Since | Type |
 | ----- | ----- | ----- | ----- |
+| [@aws-cdk/aws-ecs-patterns:addSecurityGroupToAsgCapacityProviders](#aws-cdkaws-ecs-patternsaddsecuritygrouptoasgcapacityproviders) | Add security group through "configureAutoScalingGroup" | V2·NEXT | (default) |
 | [@aws-cdk/aws-route53-patters:useCertificate](#aws-cdkaws-route53-pattersusecertificate) | Use the official `Certificate` resource instead of `DnsValidatedCertificate` | V2·NEXT | (default) |
 | [@aws-cdk/core:newStyleStackSynthesis](#aws-cdkcorenewstylestacksynthesis) | Switch to new stack synthesis method which enables CI/CD | 2.0.0 | (fix) |
 | [@aws-cdk/core:stackRelativeExports](#aws-cdkcorestackrelativeexports) | Name exports based on the construct paths relative to the stack, rather than the global construct path | 2.0.0 | (fix) |
@@ -61,6 +62,7 @@ The following json shows the current recommended set of flags, as `cdk init` wou
 ```json
 {
   "context": {
+    "@aws-cdk/aws-ecs-patterns:addSecurityGroupToAsgCapacityProviders": true,
     "@aws-cdk/aws-lambda:recognizeLayerVersion": true,
     "@aws-cdk/core:checkSecretUsage": true,
     "@aws-cdk/core:target-partitions": [
@@ -318,6 +320,26 @@ Encryption can also be configured explicitly using the `encrypted` property.
 | (default in v2) | `true` |  |
 
 **Compatibility with old behavior:** Pass the `encrypted: false` property to the `FileSystem` construct to disable encryption.
+
+
+### @aws-cdk/aws-ecs-patterns:addSecurityGroupToAsgCapacityProviders
+
+*Add security group through "configureAutoScalingGroup"* (default)
+
+ConfigureAutoScalingGroup currently does not connect the ASG security group to the cluster's security group.
+The result of this is that on new deployments, EC2 instances which have been autoscaled can have their security groups
+reconfigured and lose connectivity to the ECS cluster. This feature flag enables the correct behavior. 
+
+If this flag is not set, cluster.addAsgCapacityProvider() does not correctly configure the autoscaling group's
+Security Groups to work with the ECS cluster. If the flag is set, the ASG is correctly configured.
+
+
+| Since | Default | Recommended |
+| ----- | ----- | ----- |
+| (not in v1) |  |  |
+| V2·NEXT | `false` | `true` |
+
+**Compatibility with old behavior:** You can use `configureAutoScalingGroup()`, to add secuirty group.
 
 
 ### @aws-cdk/aws-route53-patters:useCertificate


packages/@aws-cdk/cx-api/FEATURE_FLAGS.md
corymhall, line 1: You shouldn't need to add anything to `@aws-cdk/cx-api`. The build should copy over everything from `aws-cdk-lib/cx-api`

packages/aws-cdk-lib/aws-ecs/lib/cluster.ts
```diff
@@ -474,6 +475,10 @@ export class Cluster extends Resource implements ICluster {
   }
 
   private configureAutoScalingGroup(autoScalingGroup: autoscaling.AutoScalingGroup, options: AddAutoScalingGroupCapacityOptions = {}) {
+    const enableEcsAddSecurityGroup = FeatureFlags.of(this).isEnabled(cxapi.ECS_ADD_SECURITY_GROUP_TO_ASG_CAPACITY_PROVIDERS);
```
corymhall, line 478: I'm not sure this needs a feature flag. Can you explain what is the breaking change here?

corymhall (changes requested):

packages/aws-cdk-lib/aws-ecs/lib/cluster.ts
homakk: https://github.com/aws/aws-cdk/pull/24316
This is the actually ticket which contain all the conversation which is abandoned due to no review or no updates received for this for past 4 weeks.

Without providing feature flag the integ tests present in ecs-patterns are breaking. As this is a p1 fix but it may be breaking for other customers-- we asked @TheRealAmazonKendra is there any guidance on whether we need to add a feature flag here.

She said this need feature flag. After I created a feature flag the integ test are not breaking in other modules

packages/@aws-cdk/cx-api/FEATURE_FLAGS.md
homakk: Ya but this PR raised before the release. So at that time I worked on @aws-cdk/cx-api. After new release I made changes for aws-cdk-lib/cx-api files. I'm not sure wether both are required or not. So I didn't changed anything to  to @aws-cdk/cx-api.

Now I just updated those changes.

packages/aws-cdk-lib/aws-ecs/lib/cluster.ts
corymhall: I see in that PR that breaking changes were mentioned, but I don't see any discussion of what those breaking changes were. Can you list them here?

homakk: This are breaking in aws-ecs-patterns at that time 
image

corymhall: But is that actually a breaking change? Just because the template changed doesn't mean it would break anyone.

homakk: When we create a new feature flag this will affect the behavior of the tests in modules which depend on aws-ecs. So the cached snapshot will not have the feature flag enabled, but the new snapshot will.
The alb-ecs-service-command-entry-point test which is failing in aws-ecs-patterns  due to a weird regression. the cached snapshot doesn’t deploy, possibly due to a mutation of the image or a change in ecs behavior. it was odd; neither Austin nor Yurui could really figure out what happened it seems. but finally Yurui supplied Austin with the new command/entry-point to use where this was a particularly tricky one it seems. and fix it by running yarn integ --update-on-failed --disable-update-workflowto make sure that the new snapshot deployed.


But without feature flag all the integ test in aws-ecs-patterns which contain autoscaling got break at that time. I dont have the exact screenshot related that as per Kendra Neil guidence I created a feature flag to solve this one

PR #25229, opened 2023-04-20
feat(appmesh): access log format support for app mesh

EdwardXF:
**Implement access logging feature for AppMesh CDK.**
* Access logging feature allows customer to choose between Text and Json format for the virtual nodes and virtual gateways. Customers can also specify patterns supported by Envoy. (https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage) Example CFN template: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-appmesh-virtualnode-loggingformat.html

### All Submissions:

* [Y] Have you followed the guidelines in our [Contributing guide?](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md)

### Adding new Unconventional Dependencies:

* [N] This PR adds new unconventional dependencies following the process described [here](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md/#adding-new-unconventional-dependencies)

### New Features

* [Y] Have you added the new feature to an [integration test](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md)?
	* [Y] Did you use `yarn integ` to deploy the infrastructure and generate the snapshot (i.e. `yarn integ` without `--dry-run`)?

----


packages/aws-cdk-lib/aws-appmesh/lib/shared-interfaces.ts
```diff
@@ -143,28 +143,138 @@ class FileAccessLog extends AccessLog {
...
+        text: this.text,
+      },
+    };
+  }
+}
+
+/**
+  * Key Value pair json reference for json Format
+  */
+class JsonFormatRef {
```
otaviomacedo, line 268: Why do you need to create this new type?

packages/aws-cdk-lib/aws-appmesh/lib/shared-interfaces.ts
```diff
@@ -143,28 +143,138 @@ class FileAccessLog extends AccessLog {
...
+  }
+}
+
+class TextLoggingFormat extends LoggingFormat {
+  /**
+  * Json pattern for the output logs
+  *
+  * @default - no text format specified
+  */
+  private readonly text?: string;
```
otaviomacedo, line 250: Why is this optional?

otaviomacedo (dismissed):

packages/aws-cdk-lib/aws-appmesh/lib/shared-interfaces.ts
EdwardXF: The whole field is optional. And you can have either text or json format

packages/aws-cdk-lib/aws-appmesh/lib/shared-interfaces.ts
EdwardXF: Deleted that, that's part of our api design for other parts

packages/aws-cdk-lib/aws-appmesh/lib/shared-interfaces.ts
otaviomacedo: Sure, but if want the `text` field in the L1 to be undefined, why bother creating an instance of this class in the first place? In fact, you are already enforcing the text to be defined in the public interface (`LoggingFormat.fromText()`). Do the same here. 

Same for the JSON case.

EdwardXF: Yeah, you are right, being silly here, updated

otaviomacedo (approved):

otaviomacedo (approved):

PR #25230, opened 2023-04-21
fix(eks): The cluster's admin role uses a root account principal

comcalvi:
The code here: https://github.com/aws/aws-cdk/blob/0d156a810a7a049e03f2d84582f12b7a231dea2e/packages/aws-cdk-lib/aws-eks/lib/cluster-resource.ts#L122

results in a root principal being used as the cluster's admin role.

----


iliapolo:
Closing in favor of https://github.com/aws/aws-cdk/pull/25268

PR #25239, opened 2023-04-21
chore(aws-cdk-lib): Run pre and post package steps for aws-cdk-lib from pack.sh

madeline-k:
Until this PR, the pre and post package steps defined in `aws-cdk-lib`'s `package.json` were not actually executed during the pack step of the pipeline. The ultimate purpose of this change is to make sure that the `.js` source files are minified in the release pipeline, resulting in a smaller overall package size. 

----


tools/@aws-cdk/cdk-build-tools/lib/package-info.ts
```diff
@@ -142,14 +142,14 @@ export interface CDKBuildOptions {
   };
 
   /**
-   * An optional command (formatted as a list of strings) to run before building
+   * Optional commands (formatted as a list of strings, which will be joined together with the && operator) to run before building
```
madeline-k, line 145: I did not change the logic for the pre and post build steps, but it was inaccurately described here, so I updated the comments.

pack.sh
```diff
@@ -38,16 +38,24 @@ function lerna_scopes() {
 
 scripts/run-rosetta.sh --infuse --pkgs-from $TMPDIR/jsii.txt
 
+# Execute any pre-package steps for the jsii modules here:
```
otaviomacedo, line 41: I'm not very familiar with `cdk-package`, but from a quick look, it seems that it supports many targets at once (essentially calling `pacmak` the same way as in lines 47-49). So instead of creating `pre-only` and `post-only` options, wouldn't it be simpler to call `cdk-package` instead of `pacmak` directly, and let it run the pre and post steps normally?

In other words, replacing these three separate steps with:

```bash
npx lerna run --scope aws-cdk-lib package -- --verbose --targets $(cat $TMPDIR/jsii.txt)
```

madeline-k: I've implemented and tested both approaches, and I think we should stick with my current approach, even though it is more complex in the scripting code. The two major downsides to using lerna are:
1. 5 minutes is added to the pack time
2. Modifications to the nx.json config are required. And the change forces us to pack ALL dependencies of the packages that need to be packed, including packages in `tools` that don't need to be packed. 

I've made the change and then reverted it so you can take a look if you want: https://github.com/aws/aws-cdk/pull/25239/commits/b50c981a68cab989e249302ecdc68857e55a2f34

PR #25241, opened 2023-04-21
fix(integ-tests): allow multiple AwsApiCalls with the same action and different parameters

### LINKED ISSUE #25014: (integ-test): multiple AwsApiCalls with the same action creates duplicate names
### Describe the bug

If you have > 1 `awsApiCall` calls on the same API & Action, then there will be a collision in the constructs' names.

### Expected Behavior

Sometimes you have to do the same API call against different resources, so I would expect different construct names.

### Current Behavior

Any `AwsApiCall` with the same API + action instantiates constructs with the same name.

### Reproduction Steps

```typescript
import { AccountRootPrincipal, Role } from 'aws-cdk-lib/aws-iam';
import { Key } from 'aws-cdk-lib/aws-kms';
import { App, CfnOutput, RemovalPolicy, Stack } from 'aws-cdk-lib';
import { DeduplicationScope, FifoThroughputLimit, Queue, QueueEncryption } from 'aws-cdk-lib/aws-sqs';

const app = new App();

const stack = new Stack(app, 'aws-cdk-sqs');

const dlq = new Queue(stack, 'DeadLetterQueue');
const queue = new Queue(stack, 'Queue', {
  deadLetterQueue: { queue: dlq, maxReceiveCount: 5 },
});
const fifo = new Queue(stack, 'FifoQueue', {
  fifo: true,
  encryptionMasterKey: new Key(stack, 'EncryptionKey', { removalPolicy: RemovalPolicy.DESTROY }),
});
const highThroughputFifo = new Queue(stack, 'HighThroughputFifoQueue', {
  fifo: true,
  fifoThroughputLimit: FifoThroughputLimit.PER_MESSAGE_GROUP_ID,
  deduplicationScope: DeduplicationScope.MESSAGE_GROUP,
});
const kmsManagedEncryptedQueue = new Queue(stack, 'kmsManagedEncryptedQueue', {
  encryption: QueueEncryption.KMS_MANAGED,
});
const unencryptedQueue = new Queue(stack, 'UnencryptedQueue', {
  encryption: QueueEncryption.UNENCRYPTED,
});
const ssl = new Queue(stack, 'SSLQueue', { enforceSSL: true });

const role = new Role(stack, 'Role', {
  assumedBy: new AccountRootPrincipal(),
});

dlq.grantConsumeMessages(role);
queue.grantConsumeMessages(role);
fifo.grantConsumeMessages(role);
highThroughputFifo.grantConsumeMessages(role);
kmsManagedEncryptedQueue.grantConsumeMessages(role);
unencryptedQueue.grantConsumeMessages(role);
ssl.grantConsumeMessages(role);

new CfnOutput(stack, 'QueueUrl', { value: queue.queueUrl });

const integtest = new integ.IntegTest(app, 'SqsTest', {
  testCases: [stack],
});

// GIVEN a defaults-only SQS Queue
// WHEN we query it for the `sqsManagedSseEnabled` attribute
const defaultQueueAttrs = integtest.assertions.awsApiCall('SQS', 'getQueueAttributes', {
  QueueUrl: queue.queueUrl,
  AttributeNames: ['SqsManagedSseEnabled'],
});
// THEN we expect it to return `true`
defaultQueueAttrs.expect(integ.ExpectedResult.stringLikeRegexp('true'));


// GIVEN a SQS Queue with KMS
// WHEN we query it for the `sqsManagedSseEnabled` attribute
const kmsQueueAttrs = integtest.assertions.awsApiCall('SQS', 'getQueueAttributes', {
  QueueUrl: kmsManagedEncryptedQueue.queueUrl,
  AttributeNames: ['SqsManagedSseEnabled'],
});
// THEN we expect it to return `false`
kmsQueueAttrs.expect(integ.ExpectedResult.stringLikeRegexp('false'));

app.synth();
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.70.0

### Framework Version

_No response_

### Node.js Version

16.3.0

### OS

Mac

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

guiyom-e:
This PR allows multiple AwsApiCalls with the same action and different parameters.
Closes #25014

To be able to test the new behavior, a new integration test has been added in aws-redshift-alpha package.

Next possible step (to be discussed): add an optional `callNumber` in `awsApiCall` method to be able to use this method twice with the same action and same parameters.


packages/@aws-cdk/integ-tests-alpha/lib/assertions/private/deploy-assert.ts
```diff
@@ -59,7 +59,12 @@ export class DeployAssert extends Construct implements IDeployAssert {
   }
 
   public awsApiCall(service: string, api: string, parameters?: any, outputPaths?: string[]): IApiCall {
-    return new AwsApiCall(this.scope, `AwsApiCall${service}${api}`, {
+    let hash = 'NoParameter';
```
corymhall, line 62:
```suggestion
    let hash = '';
```

Can we default to no hash instead.

corymhall (changes requested):
Looks great! Just one minor change.

packages/@aws-cdk/integ-tests-alpha/lib/assertions/private/deploy-assert.ts
guiyom-e: Yes, sure, and it's simpler! I've just pushed changes.

corymhall (dismissed):
Thanks for the update! 🚀

guiyom-e:
@corymhall the merge train github action fails with error `refusing to allow a GitHub App to create or update workflow .github/workflows/github-merit-badger.yml without workflows permission
err-code: 8A27E`
Is it an known error that can be fixed ? Or do I need to rebase manually the PR ?

corymhall (approved):

PR #25259, opened 2023-04-24
fix(backup): `BackupVault.fromBackupVaultArn` parses wrong arn format

### LINKED ISSUE #25212: (aws-backup): (BackupVault.fromBackupVaultArn fails to parse backup vault ARN)
### Describe the bug

If you try to include a reference to an already defined backup vault through its ARN, and then use the vault in a backup plan rule, you get an error when you try to deploy the stack which uses it.

**Error: Failed to create ChangeSet cdk-deploy-change-set on datalake-redshift: FAILED, Template error: Fn::Select  cannot select nonexistent value at index 1**

### Expected Behavior

No error, referencing the backup vault construct in a backup plan should work.

### Current Behavior

A stack trace when the change set is about to be created, which includes the error:

**Error: Failed to create ChangeSet cdk-deploy-change-set on datalake-redshift: FAILED, Template error: Fn::Select  cannot select nonexistent value at index 1**

### Reproduction Steps

```ts
const vault = BackupVault.fromBackupVaultArn('arn:aws:backup:eu-north-1:123456789012:backup-vault:blablabla')
const plan = new backupPlan(scope, 'plan', { backupVault: vault });
```

Workaround currently is to change the ARN to use a slash between "backup-vault" and the backup vault name, before calling fromBackupVaultArn().

### Possible Solution

Change fromBackupVaultArn to use `ArnFormat.COLON_RESOURCE_NAME` instead of `ArnFormat.SLASH_RESOURCE_NAME`.

Looking at the generated CloudFormation where it is used, one can see the following below. The problem here is that "Fn::Select" uses index 1, after splitting by "/". However, there is no "/" separator in a backup vault ARN, only ":". So the index does not work.

```json
"BackupPlanRule": [
      {
       "EnableContinuousBackup": true,
       "Lifecycle": {
        "DeleteAfterDays": 7
       },
       "RuleName": "datalake_backup_plan-PITR-rule",
       "TargetBackupVault": {
        "Fn::Select": [
         1,
         {
          "Fn::Split": [
           "/",
           {
            "Fn::Select": [
             5,
             {
              "Fn::Split": [
               ":",
               {
                "Ref": "backupvaultarnparamParameter"
               }
              ]
             }
            ]
           }
          ]
         }
        ]
```

This is verified by looking at the code for the BackupVault in aws-backup module (below from release 2.76.0), where **fromBackupVaultArn()** uses `ArnFormat.SLASH_RESOURCE_NAME` instead of `ArnFormat.COLON_RESOURCE_NAME`.

```ts
/**
 * A backup vault
 */
export class BackupVault extends BackupVaultBase {
  /**
   * Import an existing backup vault by name
   */
  public static fromBackupVaultName(scope: Construct, id: string, backupVaultName: string): IBackupVault {
    const backupVaultArn = Stack.of(scope).formatArn({
      service: 'backup',
      resource: 'backup-vault',
      resourceName: backupVaultName,
      arnFormat: ArnFormat.COLON_RESOURCE_NAME,
    });

    return BackupVault.fromBackupVaultArn(scope, id, backupVaultArn);
  }

  /**
   * Import an existing backup vault by arn
   */
  public static fromBackupVaultArn(scope: Construct, id: string, backupVaultArn: string): IBackupVault {
    const parsedArn = Stack.of(scope).splitArn(backupVaultArn, ArnFormat.SLASH_RESOURCE_NAME);

```

### CDK CLI Version

2.76.0

### Framework Version

_No response_

### Node.js Version

16

### OS

macOS

### Language

Python

### Language Version

3.8

### Other information

_No response_

lpizzinidev:
`BackupVault.fromBackupVaultArn` parsed ARNs using the `ArnFormat.SLASH_RESOURCE_NAME` format.
This fix changes it to the [expected](https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsbackup.html#awsbackup-resources-for-iam-policies) `ArnFormat.COLON_RESOURCE_NAME` format.

Closes #25212 .

----


lpizzinidev:
Exemption Request.
Let me know if integration tests are needed.

corymhall (approved):

PR #25267, opened 2023-04-24
fix(api-gateway): add validation to `variables` property on `Stage` resource

### LINKED ISSUE #3635: [apigateway] Add validations to the 'variables' property on Stage resource
* **I'm submitting a ...**
  - [ ] :beetle: bug report
  - [x] :rocket: feature request
  - [ ] :books: construct library gap
  - [ ] :phone: security issue or vulnerability => Please see [policy](https://github.com/aws/aws-cdk/security/policy)
  - [ ] :question: support request => Please see note at the top of this template.

* **What is the current behavior?**

API Gateway Stage variables must match a regex to be valid ([See the documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html#cfn-apigateway-stage-variables)). When adding variables using CDK, this regex is not enforced, so it can synthesize a CloudFormation template that cannot be deployed.

```typescript
    // const myFunction: IFunction;

    const versionAlias = lambda.Version.fromVersionAttributes(this, 'VersionInfo', {
        lambda: myFunction,
        version: '${stageVariables.lambdaAlias}'
    });

    const api: apiGateway.RestApi = new apiGateway.LambdaRestApi(this, 'RestApi', {
        restApiName: 'my-test-api',
        handler: versionAlias,
        deployOptions: {
          variables: {
            lambdaAlias: myFunction.latestVersion // == '$LATEST'
          }
        }
      });
```

* **What is the expected behavior (or behavior of feature suggested)?**

CDK could throw an error if the variable values do not match the regex, preventing invalid templates from being created.

* **What is the motivation / use case for changing the behavior or adding this feature?**

In a Blue/Green Deployment use case, I might have `blue`, `green` and `production` stages on my RestAPI, using Stage variables to define which Lambda alias each stage integrates with. It is currently illegal to use the `$LATEST` alias in a Stage variable.

* **Please tell us about your environment:**

  - CDK CLI Version: 1.3.0
  - Module Version: `@aws-cdk/aws-apigateway@1.3.0`
  - OS: OSX Mojave 10.14.5
  - Language: TypeScript

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. associated pull-request, stackoverflow, gitter, etc)

sumupitchayan:
Closes #3635 

From the original issue: API Gateway Stage variables must match a regex to be valid ([See the documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html#cfn-apigateway-stage-variables)). When adding variables using CDK, this regex is not enforced, so it can synthesize a CloudFormation template that cannot be deployed.

This PR fixes this issue by adding a validation that the strings match that regex along with a unit test for these changes.

----

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ let us know if it's not up-to-date (even better, submit a PR with your  correcti
 
 The following steps describe how to set up the AWS CDK repository on your local machine.
 The alternative is to use [Gitpod](https://www.gitpod.io/), a Cloud IDE for your development.
-See [Gitpod section](#gitpod-alternative) on how to set up the CDK repo on Gitpod.
+See [Gitpod section](#gitpod) on how to set up the CDK repo on Gitpod.
 
 ### Setup
 
@@ -92,7 +92,7 @@ CDK can be found at the location `packages/aws-cdk-lib/aws-iam`.
 The repo also contains the `tools/` directory that holds custom build tooling (modeled as private npm packages)
 specific to the CDK.
 
-### Build
+### Building aws-cdk-lib
 
 The full build of all of the packages within the repository can take a few minutes, about 20  when all tests are run.
 Most contributions only require working on a single package, usually `aws-cdk-lib`. To build this package for the first
@@ -162,7 +162,7 @@ Please follow the [setup instructions](https://code.visualstudio.com/docs/remote
 
 With VS Code setup, you will be prompted to open the `aws-cdk` repo in a Dev Container, or you can choos "Dev Containers: Reopen in Container" from the VS Code command palette.
 
-### Gitpod (Alternative)
+### Gitpod
 
 You may also set up your local development environment using [Gitpod](http://gitpod.io) -
 a service that allows you to spin up an in-browser Visual Studio Code-compatible editor,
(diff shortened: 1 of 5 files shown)


packages/aws-cdk-lib/aws-apigateway/lib/lambda-api.ts
```diff
@@ -68,6 +69,22 @@ export class LambdaRestApi extends RestApi {
       this.root.addMethod = addMethodThrows;
       this.root.addProxy = addProxyThrows;
     }
+
+    this.node.addValidation({
+      validate() {
+        if (props.deployOptions?.variables) {
+          for (let key in props.deployOptions.variables) {
+            // Checks that variable Stage values match regex
+            const regexp = /[A-Za-z0-9-._~:/?#&amp;=,]+/;
```
corymhall, line 78: Is this regex correct? The docs have this one `[A-Za-z0-9-._~:/?#&=,]+`

packages/aws-cdk-lib/aws-apigateway/lib/lambda-api.ts
```diff
@@ -68,6 +69,22 @@ export class LambdaRestApi extends RestApi {
       this.root.addMethod = addMethodThrows;
       this.root.addProxy = addProxyThrows;
     }
+
+    this.node.addValidation({
+      validate() {
+        if (props.deployOptions?.variables) {
```
corymhall, line 75: Totally options, but a style suggestion

```ts
validate() {
  for (const value of Object.values(props.deployOptions.variables ?? {}) {
    ...
  }
}
```

corymhall (changes requested):

packages/aws-cdk-lib/aws-apigateway/lib/lambda-api.ts
sumupitchayan: Typo on my end - thanks for catching this!

corymhall (approved):

Issue #25272, opened 2023-04-24
bootstrap: Template can currently only be deployed into Commercial Cloud when `TrustedAccounts` set


meeber:
### Describe the bug

#24588 introduced ARNs with hardcoded partitions of "aws" (i.e., Commercial Cloud) on lines 477 & 480 of bootstrap-template.yaml, thus preventing deployment into other partitions (e.g., GovCloud) when the `TrustedAccounts` parameter is set to one or more Account IDs.

### Expected Behavior

To be able to deploy the CDK Bootstrap Template into partitions other than Commercial Cloud (such as GovCloud) when setting the `TrustedAccounts` parameter.

### Current Behavior

Deployment fails in CloudFormation with error: Partition "aws" is not valid for resource "arn:aws:kms:*:<account_id>:*".

### Reproduction Steps

Deploy version 17 of the CDK Bootstrap Template into a partition other than Commercial Cloud and set the TrustedAccounts` parameter to one or more Account IDs.

### Possible Solution

Replace the hardcoded "aws" partition in the ARNs with either "*" or "${AWS::Partition}"; I didn't test to see if the latter option works inside the join function.

### CDK CLI Version

2.76.0

### Node.js Version

14.19.1

### OS

CentOS 7.9.2009

### Language

Python

### Language Version

3.9.10

indrora:
Related: CN users have issues with hardcoded ARN parts: https://github.com/aws/aws-cdk/issues/23642

PR #25273, opened 2023-04-24
fix: hardcoded partition in bootstrap template

MrArnoldPalmer:
Removes hardcoded partition in bootstrap template introduced in https://github.com/aws/aws-cdk/pull/24588.

Fix: #25272

----


rix0rrr (changes requested):
Doesn’t work. We are not allowed to use intrinsics in the separator field. We can remove this change altogether, I’m told it doesn’t help

Issue #22442, opened 2022-10-10
aws-eks: Nodegroup.nodegroupName is not the nodegroup's name


plumdog:
### Describe the bug

The result of `myNodegroup.nodegroupName` is something like `mycluster/mynodegroup`. This is not the nodegroup's name.

Eg, if I try to pass this to an SDK call to describe the nodegroup, I get:
```
$ aws eks describe-nodegroup --cluster-name mycluster --nodegroup-name mycluster/mynodegroup

An error occurred (InvalidParameterException) when calling the DescribeNodegroup operation: The nodegroup name parameter contains invalid characters. It should begin with letter or digit and can have any of the following characters: the set of Unicode letters, digits, hyphens and underscores.
```

### Expected Behavior

`myNodegroup.nodegroupName` to give the name of the nodegroup.

### Current Behavior

`myNodegroup.nodegroupName` is the "physical resource ID", see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-eks-nodegroup.html#aws-resource-eks-nodegroup-return-values, because it uses `.ref` not `attrNodegroupName`. See eg https://github.com/aws/aws-cdk/blob/v2.33.0/packages/@aws-cdk/aws-eks/lib/managed-nodegroup.ts#L455

### Reproduction Steps

Roughly, something like:
```typescript
const vpc = new ec2.Vpc(this, 'Vpc');
const cluster = new eks.Cluster(this, 'Cluster', {
    clusterName: 'mycluster',
});
const nodegroup = cluster.addNodegroup(...);
new CfnOutput(this, 'NodegroupName', {
    value: nodegroup.nodegroupName,
});
```
Then call `aws eks describe-nodegroup --cluster-name mycluster --nodegroup-name [value from output]` and see that it errors.

### Possible Solution

Change `.nodegroupName` to be `.attrNodegroupName`.

Maybe add `.nodegroupPhysicalResourceId` as `.ref`?

### Additional Information/Context

This would be a breaking change.

### CDK CLI Version

2.33.0

### Node.js Version

16

### OS

Linux

### Language

Typescript

pahud:
Hi @plumdog 

Yes you are right. This seems to be a bug. I am assigning this to myself as p2 and the PR will be underway shortly.

https://github.com/aws/aws-cdk/blob/280b876d426522fd043f13ff0b188e29acaa0fbf/packages/%40aws-cdk/aws-eks/lib/managed-nodegroup.ts#L455

PR #25278, opened 2023-04-25
fix(eks): nodegroupName incorrectly assigned

bhaskerv:
This PR fixes the EKS nodegroupName attribute that was incorrectly being assigned.
Old: cluster-name/nodegroup-name
New: nodegroup-name

Fixes: https://github.com/aws/aws-cdk/issues/22442

----


packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts
```diff
@@ -476,7 +476,7 @@ export class Nodegroup extends Resource implements INodegroup {
       resource: 'nodegroup',
       resourceName: this.physicalName,
     });
-    this.nodegroupName = this.getResourceNameAttribute(resource.ref);
+    this.nodegroupName = resource.getAtt('NodegroupName').toString();
```
corymhall, line 479:
```suggestion
    this.nodegroupName = this.getResourceNameAttribute(resource.attrNodegroupName);
```

Can we do this instead?

corymhall (changes requested):

PR #24026, opened 2023-02-06
fix(cdk-assets):  asset concurrency leaves a corrupted archive

rix0rrr:
This is a re-roll of https://github.com/aws/aws-cdk/pull/23677 which was reverted in https://github.com/aws/aws-cdk/pull/23994 because the `randomUUID()` function from the original solution was not available in Node versions below 14.17 (and we advertise compatibility with Node 14.*).

We didn't actually need a UUID, just any random string, so replace it with a function that generates a random string in a different way.

----------

Resolves #23290

A very simple fix for the issue where builds with `--concurrency` specified can lead to corrupt archives. Rather than use the outputFile as the basis for the temp file name we simply use a random string.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

Issue #23290, opened 2022-12-09
(cdk-assets): Parallel zipping same asset with `cdk deploy --concurrency 2` has chance to create and permanently upload corrupted asset to S3


vaclavbarta:
### Describe the bug

I suspect that there is possible race condition in cdk bundling code, this code do not handle any parallel compression of the same asset at same time correctly (bundle into same file)

https://github.com/aws/aws-cdk/blame/2ed006e50b15dfca96395d442ccee648abdbb374/packages/cdk-assets/lib/private/archive.ts#L11-L17 problem seems line 14

I can suggest some unique/randomized temporaryOutputFile name postfix to prevents racecondition when used with `cdk deploy --concurrency 2 ...` or more concurrency value.

At the same time, it would be appropriate to somehow ensure that such an error does not permanently break the app or even any other installations of other apps inside whole account in the future (see possible solutions)

Have a nice day ;)

### Expected Behavior

Paralell zipping with `--concurrency 2` option will not produce corrupted assets, do not upload it into asset bucket and recovery from this error in future. More attention to the atomicity and independency of the operations during parallel processing would be good

### Current Behavior

Yesterday we deploy our aws-cdk project with github pipelines, flow is
```shell
cdk synth
cdk deploy --all --concurrency 2 --app 'cdk.out/assembly-stage-prod/'
```

Project also use Stage construct and every stage contains two independent Stacks with API Gateways and NodejsFunctions constructs with [logRetention props](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.NodejsFunction.html#:~:text=function%27s%20execution%20environment.-,logRetention%3F,-RetentionDays)
```javascript
  const func = new NodejsFunction(this, 'func', {
      runtime: lambda.Runtime.NODEJS_16_X,
      entry: path.join(__dirname, 'src/lambda.ts'),
      handler: 'handler',
      logRetention: RetentionDays.ONE_MONTH,
    });
```

After code change, which adds using logRetention, our pipeline fails with this error and hangs about 30 minutes
```log
[0%] start: Publishing 624c688f90a2958730a01ea851f96aeebfec60adb98217e2590973a0613788d9:***-eu-central-1
[...]

PR #25293, opened 2023-04-25
fix(cdk-assets): concurrent asset building may occurs `ENOENT: no such file or directory, unlink`

tockn:
There are still similar issues of #23290  occurring despite the fixes for #23677 and #24026. In `moveIntoPlace` during asset building, there is a process to unlink if there is already a file with the same name in the destination directory. However, the check for the existence of this file can sometimes result in a race condition and trigger an ENOENT error.

Group 103

This pull request resolves this issue by adding a random string to the file name of the packaged file itself.

----


tockn:
Exemption Request

tockn:
@rix0rrr Hi. Thank you for the maintenance work you've done. Would it be possible for you to review this pr?

tockn:
@corymhall Thank you for the maintenance. Would it be possible for you to review this for me?

tockn:
@corymhall Is there any updates? 🙇‍♂️

corymhall (changes requested):
@tockn 

I'm not sure I understand why the issue is occurring in the first place. Can you explain more the steps that are occurring that lead to the issue? How can it lead to a race condition?

tockn:
@corymhall thank you for your response!

To be honest, I don't fully understand the exact reasons why this problem occurs, nor do I know how to reproduce it. However, it has been observed that when deploying many stacks containing Lambda functions using the concurrency option in cdk deploy, this issue tends to occur probabilistically.

In #23290, an `ENOENT: ... rename` error occurred, rather than an `ENOENT: ... unlink` error, which was resolved by #23677 and #24026. This implies that there is a potential for parallel execution of zip operations with the same assetId.

In the moveIntoPlace function related to the problem, the following steps exist:

1. Check whether a file exists at the path specified by target.
2. If it exists, delete (unlink) it.
3. Rename the file from source to target.

The `ENOENT: ... rename` issue in #23290, I believe, occurred according to the procedure shown in the following diagram.

image


Regarding the `ENOENT: ... unlink` issue this time, I think it is occurring due to a similar process as shown in the following diagram.

image


Therefore, in this PR, I am attempting to resolve this by adding a random string to the outputFile passed to the zipDirectory function, as a way to avoid this problem.

tockn:
@corymhall 

> a single asset can be used by multiple stacks and each stack will attempt to deploy their own assets in parallel.

I see, I understand now, thank you!

> The team has recently https://github.com/aws/aws-cdk/pull/25536 the asset publishing mechanism

That's great! I'm very much looking forward to it!

rix0rrr:
I think your analysis is correct, but I don't like the proposed fix. I'd prefer fixing the `moveIntoPlace` function.

rix0rrr:
I'm proposing this change instead: #25869

Issue #16223, opened 2021-08-25
(ecr):  Repository.fromRepositoryArn does not validate ARN which leads to wasted time


markusl:
Ok, so the following code with the new CDK Pipelines produces an unexpected result:

```ts
    const ecrRepoArn = 'arn:aws:ecr:eu-central-1:1234567891234:base-image-repo-from-ecr';
    const pipeline = new pipelines.CodePipeline(this, id, {
      pipelineName,
      crossAccountKeys: true,
      dockerEnabledForSelfMutation: true,
      dockerCredentials: [pipelines.DockerCredential.ecr(
        [
          ecr.Repository.fromRepositoryArn(this, 'ecr', ecrRepoArn)
        ]
      )],
```

The code will produce an error when running the pipeline:

`error  : [100%] fail: docker --config /tmp/cdkDockerConfignkKBqd build --tag cdkasset-699e5231c66d33f2bbdcd748964cd36874b9dc39c369f2f6d521381bd97afda0 . exited with error code 1: pull access denied for 1234567891234.dkr.ecr.eu-central-1.amazonaws.com/base-image-repo-from-ecr, repository does not exist or may require 'docker login': denied: User: arn:aws:sts::0987654376:assumed-role/PipelinePi-L9HGIT7CY91W/AWSCodeBuild-6054835c-9a09-4c3c-8f34-401787bd9553 is not authorized to perform: ecr:BatchGetImage on resource: arn:aws:ecr:eu-central-1:1234567891234:repository/base-image-repo-from-ecr`

Can you spot the problem? Neither did I. So, I navigated to the ECR console to look for the ARN, but there are no ARNs. Googling ECR ARN produces AWS documentation page https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_Repository.html which does not have any ARNs.

After navigating to the IAM console, I see the following code which looks about right:

```json
        {
            "Action": [
                "ecr:BatchCheckLayerAvailability",
                "ecr:GetDownloadUrlForLayer",
                "ecr:BatchGetImage"
            ],
            "Resource": "arn:aws:ecr:eu-central-1:1234567891234:base-image-repo-from-ecr",
            "Effect": "Allow"
        },
```

However, if you look *very closely* you can spot an error: `Ln 47, Col 24 Invalid ARN Resource: Resource ARN does not match the expected ARN format. Update the resource portion of the ARN. Learn more `.

[...]

PR #25302, opened 2023-04-25
feat(ecr): validate repository arn in fromRepositoryArn

vinayak-kukreja:
This PR is adding a regex to check the `arn` structure of an ECR repository. 

Closes https://github.com/aws/aws-cdk/issues/16223

----


vinayak-kukreja:
Exemption Request

Only a regex is being added which would validate before deployment, so integration test does not make sense here.

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
```diff
@@ -828,6 +832,31 @@ export class Repository extends RepositoryBase {
   }
 }
 
+function isValidEcrRepoArn(arn: string): boolean {
+  const arnRegex = /^arn:(\w+):ecr:([a-z]+-[a-z]+-\d{1}):(\d{12}):repository\/([A-Za-z]+.*)$/;
+  const arnInfo = arnRegex.exec(arn);
+
+  if (arnInfo === null) {
+    throw new Error(`There was an error parsing repository arn: ${arn}.`);
```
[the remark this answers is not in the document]
vinayak-kukreja: It would return `null` if the regex match fails which would mean that the arn does not align with the regex. 

More here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#return_value


I can change the error to say,
```
The repository arn does not match the regex and is invalid: ${repositoryArn}.
```

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
```diff
@@ -544,6 +546,27 @@ export class Repository extends RepositoryBase {
       throw new Error('"repositoryArn" is a late-bound value, and therefore "repositoryName" is required. Use `fromRepositoryAttributes` instead');
     }
 
+    const arnRegex = /^arn:(\w+):ecr:([a-z]+-[a-z]+-\d{1}):(\d{12}):repository\/([A-Za-z]+.*)$/;
```
kaizencc, line 549: instructions unclear 😂 I meant that the `validateRepoArn` function can be a function that lives inside `fromRepositoryArn`. Like this:

```ts
public static fromRepositoryArn(repositoryArn: string) {
  if (Token.isUnresolved(repositoryArn)) {
    throw new Error();  
  }

  validateRepositoryArn();

  const repositoryName = repositoryArn.split('/').slice(1).join('/');

  class Import extends RepositoryBase {}

  return new Import(scope, id, {
    environmentFromArn: repositoryArn,
  });

  function validateRepositoryArn() {
    // all your validations here
  }
}
```

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
```diff
@@ -1,5 +1,8 @@
 import { EOL } from 'os';
 import * as path from 'path';
+import { IConstruct, Construct } from 'constructs';
+import { CfnRepository } from './ecr.generated';
+import { LifecycleRule, TagStatus } from './lifecycle';
```
kaizencc, line 5: Wait, why these imports?

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
```diff
@@ -559,6 +563,29 @@ export class Repository extends RepositoryBase {
     return new Import(scope, id, {
       environmentFromArn: repositoryArn,
     });
+
+    function validateRepositoryArn() {
```
kaizencc, line 567: after looking closely at what #16223 is complaining about, I don't think that this function is all that helpful.

I don't think very many people are getting the partition or region wrong (and even in that case you are only validating that there is a typo, so if i meant `us-east-1` but wrote `us-east-2` I pass the validation).

Instead, the complaint is the not-that-obvious fact that `repository/` is part of the arn. Your regex checks that, but if it does fail, you get sent to `The regex validation failed for repository arn: ${repositoryArn}.` and doesn't actually tell me a) what regex or b) what part of the regex failed.

---

I think the important (and only) thing to test here is that `repository/` is part of the `<REPO-NAME>` portion of the arn. So this is my suggestion:

```ts
const splitArn = repositoryArn.split(':');
if (!splitArn[splitArn.length - 1].startsWith('repository/')) {
  throw new Error(`Repository arn should be in the format 'arn:<PARTITION>:ecr:<REGION>:<ACCOUNT>:repository/<NAME>', got ${repositoryArn}`);
}
```

Does that make sense? Honestly this is pretty standard amongst all arns so  I'm not sure if this is that important honestly. But I support adding this in just in case.

kaizencc (changes requested):
Almost there :)

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
vinayak-kukreja: Ohh, these just are moved up by eslint. I believe Momo recently fixed all of these after repo restructure.

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
vinayak-kukreja: I see, that does make sense. Would update this.

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
```diff
@@ -575,6 +577,14 @@ export class Repository extends RepositoryBase {
...
+      const splitArn = repositoryArn.split(':');
+
+      if (!splitArn[splitArn.length - 1].startsWith('repository/')) {
+        throw new Error(`Repository arn should be in the format 'arn:<PARTITION>:ecr:<REGION>:<ACCOUNT>:repository/<NAME>', got ${repositoryArn}.`);
```
kaizencc, line 585: Commenting here that `<NAME>` cannot include `:` so there should be no valid arns that include `:` after `repository/`.

runev-admin:
> This PR is adding a regex to check the `arn` structure of an ECR repository.
> 
> Closes #16223
> 
> _By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license_

The PR actually breaks the scenario below:
```
    const repoArn = ssm.StringParameter.valueFromLookup(
      this,
      configs.RepoArnParameterName
    );

    taskDefinition
      .addContainer("Container", {
        image: ecs.ContainerImage.fromEcrRepository(
          ecr.Repository.fromRepositoryArn(this, "EcrRepo", repoArn),
          releaseVersion
        ),
  ....
```

**Error: Repository arn should be in the format 'arn::ecr:::repository/', got dummy-value-for-/xxx/us-east-1/xx/xxx/EcrRepoArn.**

You can't validate the ARN at synth time. repoArn will be a dummy value then!

PR #25326, opened 2023-04-26
chore: fix patch build

rix0rrr:
The patch file that's in here is unnecessary, and in fact harmful: because the rename it applies is not necessary, it copies `undefined` values into perfectly fine types.

Remove patch and add guard rails agains this.

----


mrgrain (approved):
Ah yes, the patch needs to go in after the spec has been updated.

How did the pass the PR build though 🤔

rix0rrr:
I'm wondering that as well

rix0rrr:
Haaa and now the check fails because the spec has an error at the old commit

rix0rrr:
Oh it's because `git rev-parse --abbrev-ref HEAD` returns `HEAD` if we aren't on a branch, and then the next `git checkout HEAD` which should restore the commit ptr doesn't do anything

Issue #25196, opened 2023-04-19
core: refresh the contributing guide


pahud:
### Describe the feature

The [contributing guide](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) has been missing a lot of information since the repo remodel and is getting very difficult for the community to contribute the pull requests. I think we should update the contributing guide to clarify some important steps including:

1. How to build the core package(aws-cdk-lib).
2. How to run `yarn watch` when we start our hacking on individual module(such as aws-eks).
3. After we modify the source, how to deploy a simple app with our fix in our AWS account to validate it.
4. How to run all unit tests on a specific module(e.g. aws-eks).
5. How to run a specific unit test on a specific module(e.g. aws-eks).
6. How to run all integ tests on specific module(e.g. aws-eks).
7. How to run a specific integ test and update the snapshots on failed.
8. How to run awslint or any other linting tools before submitting the PR.
9. How to avoid common errors such as JavaScript heap out of memory error described in https://github.com/aws/aws-cdk/issues/25092

### Use Case

After the repo remodel, contributors now feel confused and the contributing guide is out-of-date.

### Proposed Solution

Update the contributing guide with relevant steps and guidance.

### Other Information

This is my personal note for a recent PR to `aws-eks` but I expect a formal guidance.

```sh
# in aws-cdk root directory
$ export NODE_OPTIONS=--max-old-space-size=8192
$ yarn install
$ cd packages/aws-cdk-lib
$ ../../scripts/buildup 
# or
$ npx lerna run build --scope=aws-cdk-lib
# now run `yarn watch` in `aws-cdk-lib`
$ yarn watch
# cd to individual module directory and start your hack
$ cd aws-eks
# now, run the unit tests for this individual module
$ yarn test aws-eks
# now validate the existing integ tests in framework-integ
$ cd <cdk_root>/packages/@aws-cdk-testing/framework-integ
# we need buildup first
$ ../../../scripts/buildup 
# now let's run all existing integ tests for aws-eks
$ yarn integ --directory test/aws-eks/test/
# If any single integ test fails, run with `--update-on-failed`. This triggers a real deployment and snapshot update.
$ yarn integ --update-on-failed test/aws-eks/test/integ.xxx.js
# test again without `--update-on-failed`
$ yarn integ test/aws-eks/test/integ.xxx.js
# when we fix all individual integ tests, run against the directory again
$ yarn integ --directory test/aws-eks/test/
[...]

PR #25328, opened 2023-04-26
fix(ecs): Allow scheduling DAEMON services even if no EC2 capacity attached to cluster (#25306)

### LINKED ISSUE #25306: ecs: Don't error if cluster lacks EC2 capacity when adding an Ec2Service
### Describe the feature

This is basically a duplicate of #5214, but I believe I've found a valid use case for not throwing a compile-time error when adding an EC2 service to an ECS cluster that lacks capacity.

Suppose:

1. I want to add a DAEMON service to my cluster. 
2. My stack organization consists of:
  a. A "shared resource" stack containing the ECS cluster, VPC, and other such resources
  b. One or more "functional" stacks, each of which provides both the services that will be managed by the cluster and the EC2 capacity on which it runs

Assume the DAEMON service will run on all EC2 instances attached to the cluster in the functional stacks described in 2(b) above. 

To avoid repetition, I want the DAEMON service to be declared in the shared resource stack so that it can be run on all EC2 capacity described in the functional stacks automatically.  Currently I cannot do this because the EC2 capacity has not been provisioned yet - but it will be later in the deployment process.

### Use Case

The use case here is an group of applications that consists of EC2 services, each of which will be run on separate EC2 ASGs, but which will have a common DAEMON service associated with each of them.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

N/A

N/A

mifisignal:
It can be useful to allow DAEMON services to be scheduled to a cluster even if there is no EC2 capacity attached to it yet, under the assumption that capacity will be attached to the cluster later. We no longer fail validation if this happens.

Fixes #25306

----


otterley:
I'm currently having trouble building test cases of this bug reported here:

https://github.com/aws/aws-cdk/issues/25196#issuecomment-1524060934

otterley:
Exemption Request - no integration test is needed here because this fix only addresses a compile-time validation issue. No changes to the rendered stack templates are made.

packages/aws-cdk-lib/aws-ecs/lib/ec2/ec2-service.ts
```diff
@@ -249,7 +249,7 @@ export class Ec2Service extends BaseService implements IEc2Service {
    */
   private validateEc2Service(): string[] {
     const ret = new Array<string>();
-    if (!this.cluster.hasEc2Capacity) {
+    if (!this.daemon && !this.cluster.hasEc2Capacity) {
```
corymhall, line 252: I'm not sure I see how this is any different from the closed issue that is related. The ask is to be able to add services to a cluster that don't have any capacity. Sounds like you are proposing that you found a loophole in the original issue because we never specifically mentioned `daemon` services 😁

I think what we need instead is a new way of managing capacity in clusters. You need capacity in a cluster so we don't want to remove that check, but we should have a way to tell the cluster that you have capacity.

corymhall (changes requested):

otterley:
@pahud are we good to merge?

packages/aws-cdk-lib/aws-ecs/lib/ec2/ec2-service.ts
otterley: @corymhall I completely understand your point of view. As I said in https://github.com/aws/aws-cdk/issues/25306#issuecomment-1524057191 - I don't necessarily believe we should fail validation for lack of capacity. I purposely kept this PR small in scope because I think it's more easily defensible, whereas the larger question still needs to be resolved.

corymhall: @otterley I disagree with your comment. If we have reason to believe that a users configuration will fail once deployed we should fail synthesis.

The bottom line is that multiple CDK team members have said the type of change that we will support. This PR falls under the same use case.

corymhall (approved):

PR #25330, opened 2023-04-27
feat(eks): alb controller include versions 2.4.2 - 2.5.1

### LINKED ISSUE #25307: eks: Update AlbControllerVersions to add missing versions between 2.4.1 and 2.5.x
### Describe the feature

aws-eks' build-in support for AWS Load Balancer Controller has only versions up to 2.4.1 (released in March 2022). Releases 2.4.2->2.5.1 releases and their associated policies are missing.

Release list: https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases

### Use Case

I would like to use a newer controller version without manual configuration.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.76.0

N/A

colifran:
## Motivation:
We should provide users with all available ALB controller versions for use with aws-eks. This change does not prohibit users from using previous ALB controller versions. Instead, this adds support for versions 2.4.2 - 2.5.1. Previous ALB controller versions can be specified by using the static "of" method as part of the AlbControllerVersion class, e.g., AlbControllerVersion.of().

## Testing:
Updated existing ALB controller integrity test to use ALB controller version 2.5.1. 

Closes #25307

----

### DIFF
--- a/packages/@aws-cdk/aws-msk-alpha/lib/cluster-version.ts
+++ b/packages/@aws-cdk/aws-msk-alpha/lib/cluster-version.ts
@@ -3,7 +3,11 @@
  */
 export class KafkaVersion {
   /**
+   * **Deprecated by Amazon MSK. You can't create a Kafka cluster with a deprecated version.**
+   *
    * Kafka version 1.1.1
+   *
+   * @deprecated use the latest runtime instead
    */
   public static readonly V1_1_1 = KafkaVersion.of('1.1.1');
 
--- a/packages/aws-cdk-lib/aws-route53/lib/record-set.ts
+++ b/packages/aws-cdk-lib/aws-route53/lib/record-set.ts
@@ -156,7 +156,13 @@ export interface RecordSetOptions {
   readonly zone: IHostedZone;
 
   /**
-   * The domain name for this record.
+   * The subdomain name for this record. This should be relative to the zone root name.
+   *
+   * For example, if you want to create a record for acme.example.com, specify
+   * "acme".
+   *
+   * You can also specify the fully qualified domain name which terminates with a
+   * ".". For example, "acme.example.com.".
    *
    * @default zone root
    */
@@ -376,7 +382,7 @@ export class AaaaRecord extends RecordSet {
  */
 export interface CnameRecordProps extends RecordSetOptions {
   /**
-   * The domain name.
+   * The domain name of the target that this record should point to.
    */
   readonly domainName: string;
 }


packages/aws-cdk-lib/aws-eks/test/alb-controller.test.ts
```diff
@@ -69,3 +69,267 @@ test('throws when a policy is not defined for a custom version', () => {
     version: AlbControllerVersion.of('custom'),
   })).toThrowError("'albControllerOptions.policy' is required when using a custom controller version");
 });
+
+test('can configure custom repository with alb controller version 2.4.2', () => {
```
kaizencc, line 73: I don't know that all these tests are useful. It's just testing that a certain string shows up in the template, which has already been tested elsewhere. The important thing when adding these new versions is the integ test -- making sure that the versions you are adding can actually be deployed to cloudformation.

Secondly, if we did want all these tests, we can avoid copy/paste fairly easily. This is where `test.each` comes in handy. Something like:

```ts
test.each(['2.4.1', '2.4.2'])('testname', (version) => {
   AlbController.create(stack, {
     cluster,
     version: AlbControllerVersion.of(version), // we know this is equivalent to AlbControllerVersion.V2_4_2
     repository: 'custom',
});
```

But again, I think we can just do away with all these tests and instead add/update an integ test to make sure that the new versions work.

packages/@aws-cdk-testing/framework-integ/test/aws-eks/test/integ.alb-controller.ts
```diff
@@ -20,7 +20,7 @@ class EksClusterAlbControllerStack extends Stack {
       vpc,
       ...getClusterVersionConfig(this),
       albController: {
-        version: eks.AlbControllerVersion.V2_4_1,
+        version: eks.AlbControllerVersion.V2_5_1,
```
kaizencc, line 23: the fact that this integ test works is good enough for me. We can be unit-test exempt (or just have one unit test if you really really want).

packages/aws-cdk-lib/aws-eks/lib/addons/alb-iam_policy-v2.4.2.json
kaizencc, line 1: I assume you are just copying prior art here but I can't help but ask wtf are these addons. Why can't they be added the normal way? And why can't they be resued given that there's no version specific entity in the policy?

kaizencc (changes requested):
I have some minor thoughts but other than that this looks good!

packages/aws-cdk-lib/aws-eks/lib/addons/alb-iam_policy-v2.4.2.json
colifran: I'm not completely sure. These are taken from here based on the specific version - https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/install/iam_policy.json.

I saw @iliapolo on some of the prior commits. He may be able to shed some light on this for us.

kaizencc (commented):
Lets give until Monday for @iliapolo to take a look, because I do find it weird to have a different policy per version and I didn't have time to investigate further. 

In the meantime, I think the PR description could use an update (and note that users could specify these versions on their own via the `AlbControllerVersion.of()` method so we're not actually denying any uses these versions atm.

packages/aws-cdk-lib/aws-eks/lib/addons/alb-iam_policy-v2.4.2.json
kaizencc: I see that too, but I don't see specific versions for 2.4.2 - 2.5.1. Looks like the latest was 2.4.1. I'm 99% sure this is okay, but I do want someone to take a second look. @otaviomacedo, since you have some experience in eks can you take a look? Previous PRs that did something similar: #18147, #19653

colifran: I should have been more clear. The IAM policies that were added aren't copied from what we already have for v2.4.1. The new IAM policies that were added are what was specifically released for v2.4.2 - v2.5.1. I used the above linked github repo and went back through each version we're adding support for and added the IAM policy specific to that version.

kaizencc: gotcha. specifically if you go to https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/install/iam_policy.json, the versions are specified under _tags_ and not _branches_.

kaizencc (approved):
Looks good!

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
```diff
@@ -87,6 +87,46 @@ export class AlbControllerVersion {
...
+
+  /**
+   * v2.5.0
+   */
+  public static readonly V2_5_0 = new AlbControllerVersion('v2.5.0', false);
+
+  /**
+   * v2.5.1
+   */
+  public static readonly V2_5_1 = new AlbControllerVersion('v2.5.1', false);
```
djnalluri, line 128: I have tried using `.of('v2.5.1')` with a custom policy. It is able to deploy, however, the pods cannot start up because they don't have the required permissions for the K8s Lease API that were added in version 1.5.0 of the [helm chart](https://github.com/aws/eks-charts/commit/0541e76e0d425c01141c016eca90338bb6d348fa). I would suggest adding an extra constructor parameter for the chart version. The appropriate chart can then be picked per ALB version instead of the hardcoded `1.4.1`.

djnalluri (changes requested):
These changes may not be enough to support version 2.5.0 and above. The [release notes](https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/tag/v2.5.0) mention incompatibilities with deployment manifests from previous versions.

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
colifran: Thanks for catching this! I will work on addressing it.

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
```diff
@@ -130,6 +170,198 @@ export enum AlbScheme {
...
+
+  /**
+   * v1.4.1
+   */
+  V1_4_1 = '1.4.1',
+
+  /**
+   * v1.4.0
+   */
+  V1_4_0 = '1.4.0',
```
kaizencc, line 237: Anyone using these old versions of helm? I think we can get by with an enum-like class here too:

```typescript
class HelmChartVersion {
  public static V1_4_1 = new HelmChartVersion('1.4.1');
  public static V1_5_1 = new HelmChartVersion('1.5.1');

  public constructor(public version: string) {}
}
```

Let people just specify their other versions I think

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
kaizencc: Because of this, it's not good enough to default helm version to 1.4.1 everywhere. That would mean that if someone wanted to do this:

```ts
new AlbController(this, 'blah', {
  version: AlbControllerVersion.V2_5_1,
});
```

Somewhere along the line it would fail because it's not compatible with helm 1.4.1. So we'd force the user to have to do something like:

```ts
new AlbController(this, 'blah', {
  version: AlbControllerVersion.V2_5_1,
  helmChart: HelmChartVersion.V1_5_0,
});
```

Even though we know what helm chart version would be compatible. So we should be able to add the default in for the user.

In practice that means that we add a prop to the constructor of AlbControllerVersion called `helmChartVersion`.

```ts
class AlbControllerVersion {
  public static readonly V2_5_1 = new AlbControllerVersion('v2.5.1', HelmChartVersion.V1.5.0, false);
  private constructor(public readonly version: string, public readonly helmChart, public readonly custom: boolean) {}
}

// where we need it, we can do
props.albControllerVersion.helmChart;
```

None of this should be a breaking change since the constructor for AlbContrllerVersion is private

kaizencc (changes requested):
A couple more comments :)

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
colifran: Thanks, Kaizen. This makes sense. I've what I think are the necessary updates for the AlbControllerVersion class to support the addition of helmChartVersion in the constructor

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
```diff
@@ -140,6 +372,11 @@ export interface AlbControllerOptions {
    */
   readonly version: AlbControllerVersion;
 
+  /**
+   * The version of the helm chart to use.
+   */
+  readonly helmChartVersion: HelmChartVersion;
```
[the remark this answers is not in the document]
colifran: Thanks for catching this - I've removed this and added it to the AlbControllerVersion constructor to allow this to support current users.

packages/aws-cdk-lib/aws-eks/test/alb-controller.test.ts
```diff
@@ -67,5 +69,6 @@ test('throws when a policy is not defined for a custom version', () => {
   expect(() => AlbController.create(stack, {
     cluster,
     version: AlbControllerVersion.of('custom'),
+    helmChartVersion: HelmChartVersion.V1_4_1,
```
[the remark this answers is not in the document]
colifran: Removed.

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
colifran: Good point. I opted against using an enum-like class here though. I'm wondering if this would be too much abstraction for what would essentially just be a string? Instead, I decided to hard code the helm chart version for the static alb controller versions that create instances of the alb controller class similar to the already hard coded alb controller version that is passed to the constructor.

kaizencc (commented):
@colifran one last thing. i think we need to add a unit test that tests that when you specify `AlbControllerVersion.V2_5_1` you get the corresponding helm chart 1.5.2. That should be doable by looking through the properties of `Custom::AWSCDK-EKS-HelmChart` in the template. 

You're doing this already in the integ test so its not like I think it won't succeed. I just think it's important to have a unit test for this actually new functionality you are adding.

packages/aws-cdk-lib/aws-eks/test/alb-controller.test.ts
```diff
@@ -69,3 +69,37 @@ test('throws when a policy is not defined for a custom version', () => {
...
+  });
+
+  AlbController.create(stack, {
+    cluster,
+    version: AlbControllerVersion.V2_5_1,
+    repository: 'custom',
+  });
+
+  Template.fromStack(stack).hasResourceProperties(HelmChart.RESOURCE_TYPE, {
+    Version: '1.5.2',
```
kaizencc, line 87:
```suggestion
    Version: '1.5.2', // The helm chart version associated with AlbControllerVersion.V2_5_1
```

kaizencc (approved):

PR #25345, opened 2023-04-27
fix(elasticloadbalancingv2): the bucket policy for ELB access logging is too permissive

tmokmss:
Currently when we enable ELB (both ALB and NLB) access logging (`logAccessLogs` method), it sets bucket policy as below:

```js
          {
            Action: [
              's3:PutObject',
              's3:PutObjectLegalHold',
              's3:PutObjectRetention',
              's3:PutObjectTagging',
              's3:PutObjectVersionTagging',
              's3:Abort*',
            ],
            Action: 's3:PutObject',
            Effect: 'Allow',
            Principal: { AWS: { 'Fn::Join': ['', ['arn:', { Ref: 'AWS::Partition' }, ':iam::127311923021:root']] } },
            Resource: {
              'Fn::Join': ['', [{ 'Fn::GetAtt': ['AccessLoggingBucketA6D88F29', 'Arn'] }, '/AWSLogs/',
                { Ref: 'AWS::AccountId' }, '/*']],
            },
          },
```

However, [the doc](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html#attach-bucket-policy) says they only need `PutObject` action. Because this policy is too permissive, it also contradicts the Security Hub policy [`[S3.6] S3 permissions granted to other AWS accounts in bucket policies should be restricted`](https://docs.aws.amazon.com/securityhub/latest/userguide/s3-controls.html#s3-6).

By this fix, we will allow only `PutObject` action to be more compliant with the policy.

----


corymhall (approved):
🙏🏻

PR #25347, opened 2023-04-27
feat(rds): add missing PerformanceInsightRetention options

ahammond:
We have a requirement to set PerformanceInsights retention to 2 months. Cfn supports it. Cdk should, too.


ahammond:
Exemption Request this doesn't seem like it needs integration testing, but... ???

corymhall (approved):

PR #25350, opened 2023-04-28
fix(s3): KMS encryption works fine for server access logging target buckets

zdf1230:
The previous changes(https://github.com/aws/aws-cdk/pull/23514 & https://github.com/aws/aws-cdk/pull/23385) about failing early when certain encryption type being used is not correct. In fact, KMS encryption works fine for server access logging target buckets with proper permission being setup.

So this change is removing the condition failing for the SSE-KMS with customized encryption key case.

However, it is not possible to know which encryption type for the server access logging bucket, so the only checking can be applied after this change merged is failing when logging to self case using BucketEncryption.KMS_MANAGED.

After this fix, the only condition would be failed pre-checking is __Log to self and using KMS_MANAGED encryption type__

This change only fix the checking condition, so this change won't affect snapshot at all. Hence, Exemption Request.

----


zdf1230:
As described in the description, this change only fix the checking condition, and won't affect snapshot generation at all. Hence, Exemption Request.

corymhall (dismissed):
I think since we've had a couple of PRs that assert different behavior we need some more documentation/testing. Can you provide links to documentation the states that you can set up logging with a KMS encrypted bucket?

If it is indeed possible then can we somehow automatically setup the correct permissions to allow it? Can we also add an integration test to assert that it works?

zdf1230:
> I think since we've had a couple of PRs that assert different behavior we need some more documentation/testing. Can you provide links to documentation the states that you can set up logging with a KMS encrypted bucket?
> 
> If it is indeed possible then can we somehow automatically setup the correct permissions to allow it? Can we also add an integration test to assert that it works?

Thanks for the response. It is a very good point that we can automatically setup the correct permissions. Added that and the integration tests as well. Deployed in my personal account and verified it works.

Though I cannot find any doc shown that SSE-KMS works. But the reason I know it works fine because we original use SSE-KMS approach in our services. The previous change caused error when we build with newer version of `aws-cdk`. Hence I am here to fix this to enable us using latest version of `aws-cdk`

corymhall (approved):

PR #25355, opened 2023-04-28
chore: upgrade dependencies to remove `vm2` vulnerability

rix0rrr:
Not that anybody uses the PAC resolver feature, but that feature uses `vm2` and `vm2` has vulnerabilities reported against it every other week.

Upgrade all dependencies to remove the use of vulnerable versions of that dependency.

----


mrgrain (approved):
> It's necessary because the types for  the workerpool module reference the type Transferable all of a sudden.

> And that type is only defined in lib.dom.d.ts (not in @types/node) for whatever reason

Issue #23628, opened 2023-01-10
aws-eks: Imported cluster yields invalid long stack name


J11522:
### Describe the bug

We try adding manifests to an imported cluster, which creates a new stack to host the kubectl provider.
This stack has a name with > 128 characters and fails validation.

As far as I can tell, the problem lies within the [generateStackName](https://github.com/aws/aws-cdk/blob/1485f48c515f81fa585a325e02f1602f56443605/packages/@aws-cdk/core/lib/stack.ts#L1400) method.
The fix to the original [issue](https://github.com/aws/aws-cdk/issues/20124) made changes to the [makeStackName](https://github.com/aws/aws-cdk/pull/20528/files#diff-caa054ee92a558f0c56b9e350128d032b8504f4d30193f70a9a8a4a2e707c990R1318) function and applied the length limit there.
This ensures that the stack id created by generateStackId is no longer than 128 characters.
However, the generated stack id is prefixed with `${assembly.stageName}-` when available, result in a too long stack name

### Expected Behavior

The stack name for the kubectlProvider stack is valid

### Current Behavior

The stack name for the kubectlProvider stack is longer than 128 characters and therefore invalid.

### Reproduction Steps

//Import cluster in a staged application with long names
this.cluster = Cluster.fromClusterAttributes(this, 'ClusterImport', {
      clusterName
      openIdConnectProvider,
      kubectlSecurityGroupId,
      kubectlPrivateSubnetIds,
      kubectlRoleArn,
      vpc,
    });
//Add manifest, resulting in creation of kubectlProvider
    this.cluster.addManifest('SampleNamespace', {
      apiVersion: 'v1',
      kind: 'Namespace',
      metadata: { name: 'SampleNamespace' },
    });

### CDK CLI Version

2.59.0 (build b24095d)

### Node.js Version

18.13 LTS

### OS

Ventura 13.1

### Language

Typescript

PR #24443, opened 2023-03-03
fix(core): Add stage prefix to stack name shortening process

QustodioPablo:
There is an issue in the stack name generation process where the prefix generated from assembly's stage name is not taken into account when shortening a stack name to meet the requirement of being equal or less than 128 characters longs.

This can lead to generating an invalid stack name greater than 128 characters: since the stack name is shortened to 128 characters, when the prefix is added, the limit is exceeded.

Current solution:
- Add the prefix into the first ID if any prefix is present.

Alternative solution:
- Add the prefix at the start of the ID list. Doing this would require to add more conditionals and consider more corner cases to keep the names consistent with the old algorithm.
 

Fixes #23628

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #25359, opened 2023-04-28
fix(core): Add stage prefix to stack name shortening process

QustodioPablo:
There is an issue in the stack name generation process where the prefix generated from assembly's stage name is not taken into account when shortening a stack name to meet the requirement of being equal or less than 128 characters longs.

This can lead to generating an invalid stack name greater than 128 characters: since the stack name is shortened to 128 characters, when the prefix is added, the limit is exceeded.

Current solution:
- Adding a feature flag
  - With the feature flag on, the prefix is processed within the generateUniqueName function.
  - With the feature flag off, stack name generation is not changed

Fixes https://github.com/aws/aws-cdk/issues/23628

NOTE: This PR was previously opened, but it was merged before I was able to add a feature flag, which ended up adding breaking changes and the changes of the PR were rolled back. Old PR: https://github.com/aws/aws-cdk/pull/24443


QustodioPablo:
Exemption Request: The previous PR was exempt from adding integration tests

QustodioPablo:
Clarification Request: Can a maintainer confirm if this PR needs integration tests?

crbernal (approved):

andypapas (approved):

ygneo (approved):

ygneo:
👍

ericgimenezgalera (approved):

MariposaGentil (approved):

cas-jc (approved):

lauralongares (approved):

lauralongares (approved):

packages/aws-cdk-lib/core/lib/private/unique-resource-name.ts
```diff
@@ -59,7 +67,7 @@ export function makeUniqueResourceName(components: string[], options: MakeUnique
   // in order to support transparent migration of cloudformation templates to the CDK without the
   // need to rename all resources.
   if (components.length === 1) {
-    const topLevelResource = removeNonAllowedSpecialCharacters(components[0], separator, options.allowedSpecialCharacters);
+    const topLevelResource = prefix + removeNonAllowedSpecialCharacters(components[0], separator, options.allowedSpecialCharacters);
```
corymhall, line 70: Shouldn't this prefix also be included in the `removeNonAllowedSpecialCharacters`?
```suggestion
    const topLevelResource = removeNonAllowedSpecialCharacters(prefix + components[0], separator, options.allowedSpecialCharacters);
```

packages/aws-cdk-lib/core/lib/stack.ts
```diff
@@ -1644,9 +1649,14 @@ export function rootPathTo(construct: IConstruct, ancestor?: IConstruct): IConst
  * has only one component. Otherwise we fall back to the regular "makeUniqueId"
  * behavior.
  */
-function makeStackName(components: string[]) {
-  if (components.length === 1) { return components[0]; }
-  return makeUniqueResourceName(components, { maxLength: 128 });
+function makeStackName(components: string[], prefix: string='') {
+  if (components.length === 1) {
+    const stack_name = prefix + components[0];
```
corymhall, line 1654: why not just let `makeUniqueResourceName` do this?

packages/aws-cdk-lib/core/test/stage.test.ts
corymhall, line 1: Can you add a couple more tests:

- stage & stack with 1 component and < 128
- stage and stack with > 1 component and < 128
- stage and stack with > 1 component and > 128

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -789,6 +790,21 @@ export const FLAGS: Record<string, FlagInfo> = {
...
+
+  //////////////////////////////////////////////////////////////////////
+  [INCLUDE_PREFIX_IN_UNIQUE_NAME_GENERATION]: {
+    type: FlagType.BugFix,
+    summary: 'Include the stack prefix in the stack name generation process',
+    detailsMd: `
+      This flag prevents the prefix of a stack from making the stack's name longer than the 128 character limit.
+
+      If the flag is set, the prefix is included in the stack name generation process.
+      If the flag is not set, then the prefix of the stack is prepended to the generated stack name.
```
corymhall, line 802: Can you add a note here explaining what will happen if you enable this feature flag on an existing stack.

corymhall (changes requested):
@QustodioPablo just a couple of questions, otherwise looks good.

packages/aws-cdk-lib/core/lib/private/unique-resource-name.ts
QustodioPablo: This is done to keep the new name as close as possible to the original behavior. When a shortened stack name had a prefix, the final result would be "Prefix-ShortenedStackName". If the prefix is included in the removal of special characters, the name generated would be "PrefixShortenedStackName".

packages/aws-cdk-lib/core/lib/stack.ts
QustodioPablo: This is done also to make the stack names to be as close as possible to the original behavior. If we go through the makeUniqueResourceName, we would lose all non allowed special character (dashes for example) from the stack name if the component has any in its name. So we don't want single component stacks to go through the makeUniqueResourceName in order to preserve its original name.

packages/aws-cdk-lib/core/test/stage.test.ts
QustodioPablo: All of them with the feature flag on, right?

packages/aws-cdk-lib/cx-api/lib/features.ts
QustodioPablo: Sure!

packages/aws-cdk-lib/core/lib/private/unique-resource-name.ts
corymhall: ah yeah you are right 👍

packages/aws-cdk-lib/core/test/stage.test.ts
corymhall: yep!

QustodioPablo: One of the suggested tests found a case where the name generated would be different when using this FF, the same scenario that caused the previous PR to be reverted: The dash from the prefix was being removed when a multi component stack is processed in the makeUniqueResourceName function.

Fixed it

packages/aws-cdk-lib/cx-api/lib/features.ts
QustodioPablo: Done

QustodioPablo:
@TheRealAmazonKendra I reopened this PR adding a feature flag, could you add an exception for integration tests as in the previous PR?

otaviomacedo (dismissed):

PR #25363, opened 2023-04-28
fix(s3-deployment): doesn't work in ADC regions

rix0rrr:
The AWS CLI that we use to run `aws s3 sync` comes with its own certificate bundle, which doesn't include the certificates used in ADC regions.

Fortunately, Lambda has curated a CA bundle already, we just need to force the CLI to use it.

----


rix0rrr:
Do not merge until we get confirmation from Lambda team that it's safe to do this in all regions.

Naumel (approved):

PR #25368, opened 2023-04-28
feat(lambda-event-sources): S3EventSource requires Bucket instead of IBucket

### LINKED ISSUE #4323: (aws-lambda-event-sources): `S3EventSource` should take `IBucket` instead of `Bucket`
### Reproduction Steps

Cannot import an existing s3 bucket when creating Lambda s3 events:
1. `aws_s3.Bucket.from_bucket_name(...)` or `aws_s3.Bucket.from_bucket_arn(...)`, etc. returns `aws_s3.IBucket` type
1. However, `aws_lambda_event_sources.S3EventSource` only takes `aws_s3.Bucket` type as a parameter
1. Note: `aws_s3.Bucket` implements `aws_s3.IBucket`

### Error Log

```
File "/usr/local/lib/python3.6/site-packages/jsii/_runtime.py", line 66, in __call__
    inst = super().__call__(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/aws_cdk/aws_lambda_event_sources/__init__.py", line 199, in __init__
    jsii.create(S3EventSource, self, [bucket, props])
  File "/usr/local/lib/python3.6/site-packages/jsii/_kernel/__init__.py", line 208, in create
    overrides=overrides,
  File "/usr/local/lib/python3.6/site-packages/jsii/_kernel/providers/process.py", line 331, in create
    return self._process.send(request, CreateResponse)
  File "/usr/local/lib/python3.6/site-packages/jsii/_kernel/providers/process.py", line 316, in send
    raise JSIIError(resp.error) from JavaScriptError(resp.stack)
jsii.errors.JSIIError: Object of type @aws-cdk/aws-s3.IBucket is not convertible to @aws-cdk/aws-s3.Bucket
```

### Environment

  - **CLI Version      :** v1.10.1
  - **Framework Version:** v1.10.1
  - **OS               :** Mac OS
  - **Language         :** Python

### Other

`aws_lambda_event_sources.S3EventSource` should take `aws_s3.IBucket` as a parameter's type.

--- 

This is :bug: Bug Report

### LINKED ISSUE #23940: aws-lambda-event-sources: S3EventSource Requires Bucket, but only needs an IBucket
### Describe the bug

I want to use an externally created bucket as an event notification source for a lambda

### Expected Behavior

An IBucket can be used as the notification source 

### Current Behavior

Type errors.  An `IBucket` cannot be assigned to a `Bucket`. 

### Reproduction Steps

```
import * as core from "aws-cdk-lib";
import * as event_sources from "aws-cdk-lib/aws-lambda-event-sources";
import * as s3 from "aws-cdk-lib/aws-s3";

const app = new core.App()
const ibucket = s3.Bucket.fromBucketName(app, "IBucket", "IBucket");
const eventSource = new event_sources.S3EventSource(ibucket, {events: [s3.EventType.OBJECT_CREATED]})
```

> TS2345: Argument of type 'IBucket' is not assignable to parameter of type 'Bucket'.

### Possible Solution

Since `addEventNotification` is the only method used by S3EventSource and is available on an IBucket, I think it should be safe to loosen the type restriction to IBucket. 

Coercing the type and deploying works.

_No response_

### CDK CLI Version

2.40.0

### Framework Version

_No response_

### Node.js Version

16

### OS

OSX

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

Styerp:
Closes #23940 
Closes #4323 
----


Styerp:
Exemption Request

I don't think this needs a readme update, since this makes it work how folks expect, but I can modify one if there's a reasonable spot.

Caveats called out in other threads are already covered by other validations and readmes.

corrjo:
Would love to see this merged and get rid of my forced casting to s3.Bucket 🙃

TheRealAmazonKendra:
We cannot accept changes without a passing build. It looks like the snapshots have been incorrectly generated to be `.ts` instead of `.js`. Please delete them and rerun the tests without the typescript language flag.

Styerp:
[API surface changes](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#api-surface-changes) says changes would cause builds to fail -- that will not happen here. Underlying objects didn't access `Bucket` methods on the object  passed into this function, either directly or indirectly. All properties and objects being used were already available on the IBucket, which was the driving rationale to ease this input type.

While this is ostensibly an API Surface Change, it is NOT a _breaking_ surface change.

[Behavior changes](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#behavior-changes) say "even though the user didn't change their code, the CloudFormation template that gets synthesized is now different." The integration tests prove this is not the case -- the CFN generated is identical.

Build will continue to fail because of the "Breaking API Changes". Guidance on how y'all handle this would be great.

PR #25374, opened 2023-04-30
fix(ecs): memory issue in validateTaskDefinition()

### LINKED ISSUE #25275: ecs: task definition validation container bug
### Describe the bug

When I add a container definition to a fargate task definition, I am prompted to include `memory` or `memoryReservation` as a required field. This is not a container-level requirement if memory is already specified in the task.

### Expected Behavior

I expect the CDK to let me create a container without specifying `memory`. 

### Current Behavior

```
  [EcsWorkspaceStack/nginxTaskDef] ECS Container web must have at least one of 'memoryLimitMiB' or 'memoryReservationMiB' specified
    at validateTree (/Users/petderek/experiments/ecs-workspace/node_modules/aws-cdk-lib/core/lib/private/synthesis.ts:356:11)
    at synthesize (/Users/petderek/experiments/ecs-workspace/node_modules/aws-cdk-lib/core/lib/private/synthesis.ts:49:5)
    at App.synth (/Users/petderek/experiments/ecs-workspace/node_modules/aws-cdk-lib/core/lib/stage.ts:217:33)
    at process.<anonymous> (/Users/petderek/experiments/ecs-workspace/node_modules/aws-cdk-lib/core/lib/app.ts:195:45)
```

### Reproduction Steps

Something like this should repro:

```
        const basicNginxTask = new aws_ecs.TaskDefinition(this, "nginxTaskDef", {
            compatibility: aws_ecs.Compatibility.EC2_AND_FARGATE,
            cpu: "256",
            executionRole: task_execution_role,
            family: "nginx",
            memoryMiB: "512",
            networkMode: aws_ecs.NetworkMode.AWS_VPC,
        });
        basicNginxTask.addContainer("nginxContainer", {
            containerName: "web",
            essential: true,
            image: aws_ecs.ContainerImage.fromRegistry("public.ecr.aws/nginx/nginx:stable"),
            portMappings: [{containerPort: 80}],
            // uncomment this to 'fix' and compile
            //memoryReservationMiB: 512,
        });
```

### Possible Solution

One solution would be to check if there is `memory` set at in the parent before validating the existence of `memory` in the container.

The correct behavior is mentioned in the `register-task-definition` documentation.

> If using the Fargate launch type, this parameter is optional.
> If using the EC2 launch type, you must specify either a task-level memory value or a container-level memory value. If you specify both a container-level memory and memoryReservation value, memory must be greater than memoryReservation . If you specify memoryReservation , then that value is subtracted from the available memory resources for the container instance where the container is placed. Otherwise, the value of memory is used.

### CDK CLI Version

2.76.0 (build 78c411b)

### Framework Version

_No response_

### Node.js Version

v18.16.0

### OS

mac

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

jackhannan:
If you where to add a container definition to a fargate task definition if you where to include `memory` or `memoryResorvation` which at least one is required you get an error. I added a first if statement to see if  the parent container has `memory` and then see if the other containers have `memory` .

`private validateTaskDefinition(): string[] {
    const ret = new Array<string>();

    if (isEc2Compatible(this.compatibility)) {
      // EC2 mode validations

      // Container sizes
      for (const container of this.containers) {
        if (container.memoryLimitSpecified && container === this.containers[0]) {
          break;
        } else if (!container.memoryLimitSpecified) {
          ret.push(`ECS Container ${container.containerName} must have at least one of 'memoryLimitMiB' or 'memoryReservationMiB' specified`);
        }
      }

    }`

https://github.com/jackhannan-hello/aws-cdk/blob/8598a44350a932ddc7e05f99b089e35e4c4a9fcf/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts#L735-L750

I did this so that if the parent has a specified memory it will exit the loop checking for memory so that if one of your sub containers has no memory specified it does not throw an error. If there is no specified memory in the parent it will go through all of the sub containers to make sure it has memory specified.

Closes #25275.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
+++ b/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
@@ -11,6 +11,7 @@ import { PlacementConstraint } from '../placement';
 import { ProxyConfiguration } from '../proxy-configuration/proxy-configuration';
 import { RuntimePlatform } from '../runtime-platform';
 import { measureMemory } from 'node:vm';
+import { aws_memorydb } from '../../..';
 
 /**
  * The interface for all task definitions.
@@ -737,12 +738,12 @@ export class TaskDefinition extends TaskDefinitionBase {
     if (isEc2Compatible(this.compatibility)) {
       // EC2 mode validations
 
-      if (!this.containers[0].memoryLimitSpecified) {
-        // Container sizes
-        for (const container of this.containers) {
-          if (!container.memoryLimitSpecified) {
-            ret.push(`ECS Container ${container.containerName} must have at least one of 'memoryLimitMiB' or 'memoryReservationMiB' specified`);
-          }
+      // Container sizes
+      for (const container of this.containers) {
+        if (container.memoryLimitSpecified && container === this.containers[0]) {
+          break;
+        } else if (!container.memoryLimitSpecified) {
+          ret.push(`ECS Container ${container.containerName} must have at least one of 'memoryLimitMiB' or 'memoryReservationMiB' specified`);
         }
       }
 


peterwoodworth (changes requested):
Hey @jackhannan-hello, thanks for submitting this

You've commited a fresh `package-lock.json` to the root, did you use `yarn install --frozen-lockfile` to install your packages? There shouldn't be a fresh package.json here

Additionally, you've added two dependencies that don't seem to get used

```
import { measureMemory } from 'node:vm';
import { aws_memorydb } from '../../..';
```

I don't think we need these, and they can be removed.

Regarding testing, could you please add a unit test which checks that this synthesizes correctly when correctly used, and another unit test which throws when misconfigured? Additionally, be sure to run the integ test you've created with our integ-runner CLI. The contributing guide should shed some light on this 🙂

PR #25379, opened 2023-05-01
feat(route53): HostedZone's default period at the end should be optional 

### LINKED ISSUE #22406: HostedZone: Default period at the end should be optional
### Describe the bug

In the [HostedZone constructor](https://github.com/aws/aws-cdk/blob/f169d7f73d7ee4b7165299ae45399438393f7463/packages/%40aws-cdk/aws-route53/lib/hosted-zone.ts#L157-L176), a period is added to the end of the name on line 163. 

This changes the user's input from `github.com` to `github.com.`. Although Route53 considers these to be identical, they cannot co-exist together. Thus, for anybody moving from CFN templates to CDK, who did not have a `.` at the end of their name, this line breaks their CloudFormation deployment.

Consider a user who has a CloudFormation template with the following:

```
CustomHostedZone:
    Type: AWS::Route53::HostedZone
    Properties:
        Name: github.com
        ...
```

Who then re-creates their resource in CDK, keeping the same LogicalId for the migration

```
new HostedZone(this, "CustomHostedZone", {
    zoneName: 'github.com'
    ...
}
overrideLogicalId("CustomHostedZone");
```

Because the `.` is added, CloudFormation will see this as a name change, and try to re-create the resource. However, Route53 will see the same domain name and block the creation because the domain already exists. Because this added `.` is by default, the user is only left with hacky solutions to deploy their code.

### Expected Behavior

Give the user an option whether or not the `.` should be added at the end

### Current Behavior

The user has no control over the behavior.

### Reproduction Steps

Create a hosted zone with a CFN template with no trailing `.`, and then try to create the same HostedZone, with the LogicalId overwritten, using CDK.

### Possible Solution

At this point, I believe the only viable solution is a flag (`addTrailingDot`?) added to the properties which is defaulted to true. Unfortunately, removing this behavior outright will break anybody's code who didn't have pre-existing resources. 

If `github.com.` exists in a user's Route53 domain list, and the `.` is removed from a later release of this package, then their CDK will start deploying as `github.com`, which will still cause CloudFormation to try to re-create the resource, causing the same issue.

_No response_

### CDK CLI Version

2.43.1

### Framework Version

_No response_

### Node.js Version

14

### OS

Mac

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

watany-dev:
new flag `addTrailingDot`

The addTrailingDot option provides users with the choice to add a trailing dot (.) to the given domain name. In DNS, appending a dot at the end of a domain name indicates that the domain name is a fully qualified domain name (FQDN).
This function returns the value as is if the domain name is already an FQDN or has a trailing dot, and otherwise treats it as having a trailing dot. The existing specification was rejecting domain names with trailing dots, but this modification makes it unnecessary, so the validation check has been removed.

Closes #22406

----


packages/aws-cdk-lib/aws-route53/lib/hosted-zone.ts
```diff
@@ -132,11 +139,6 @@ export class HostedZone extends Resource implements IHostedZone {
       props: query,
     }).value;
 
-    // CDK handles the '.' at the end, so remove it here
-    if (response.Name.endsWith('.')) {
```
corymhall, line 136: I think we need to keep this here. It looks like we are explicitly keeping the `Name` value within cdk without the `.` and I don't think we should change this behavior.

packages/aws-cdk-lib/aws-route53/lib/hosted-zone.ts
```diff
@@ -159,8 +161,11 @@ export class HostedZone extends Resource implements IHostedZone {
 
     validateZoneName(props.zoneName);
 
+    // addTrailingDot プロパティが false でなければ末尾にドットを追加
```
corymhall, line 164: Can you remove these other characters.

corymhall (changes requested):
Looks good! Just  a couple of comments.

corymhall (dismissed):

corymhall (approved):

PR #25375, opened 2023-04-30
feat(lambda): response payload streaming

watany-dev:
[AWS Lambda introduces response payload streaming](https://aws.amazon.com/jp/about-aws/whats-new/2023/04/aws-lambda-response-payload-streaming/)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #25386, opened 2023-05-01
feat: Add invokeMode to FunctionUrl.

grasmash:
Lambda has released a new invokeMode parameter for Function URL, enabling response to be buffered or streamed back from the function. This change introduces the new parameter to the FunctionUrl class following the patterns established by existing parameters like `FunctionUrlCorsOptions`.

----


packages/aws-cdk-lib/aws-lambda/lib/function-url.ts
```diff
@@ -59,6 +59,23 @@ export enum HttpMethod {
...
+
+  /**
+   *  Invocation results are available when the payload is complete.
+   */
+  BUFFERED = 'BUFFERED',
+
+  /**
+   * Your function streams payload results as they become available.
+   */
+  STREAMED = 'STREAMED',
```
valerena, line 76: This should probably be [`RESPONSE_STREAM`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-url.html#cfn-lambda-url-invokemode)

grasmash:
I'm not sure this actually requires a new integration test .. I see no existing integration tests for Function URLs. Is that right?

grasmash:
The existing function URL test has been updated. I don't think this requires a new test. Therefore, making an Exemption Request.

grasmash:
The existing function URL test has been updated. I don't think this requires a new test. Therefore, making an Exemption Request.

watany-dev:
私のPRと重複した内容のようです

https://github.com/aws/aws-cdk/pull/25375

grasmash:
Yes and yours has better testing :) I'll close this.

PR #25398, opened 2023-05-02
feat(neptune): adds the snapshotIdentifier property for restoring databases

### LINKED ISSUE #24972: neptune: construct should allow restoration from a snapshot
### Describe the feature

The snapshotIdentifier would work very similar to the underlying RDS feature (https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.CfnDBCluster.html). This would take in an ARN or another identifier and restore the Neptune database from the snapshot that matches the identifier.

### Use Case

Users require the ability to save snapshots of their Neptune database and restore from that snapshot in the case of a critical failure. This field would allow restoration from a snapshot while deploying applications that utilize cloud formation.

### Proposed Solution

Inclusion of the snapshotIdenfier property built into RDS - https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.CfnDBCluster.html

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.73.0

Mac OSX, Linux

nrusson:
Allows neptune snapshot identifiers to be passed in as part of a construct. Restores the given snapshot instead of creating a cluster from scratch.
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.CfnDBCluster.html

Closes #24972.

----

### DIFF
--- a/packages/@aws-cdk/aws-neptune-alpha/lib/cluster.ts
+++ b/packages/@aws-cdk/aws-neptune-alpha/lib/cluster.ts
@@ -260,6 +260,14 @@ export interface DatabaseClusterProps {
   readonly securityGroups?: ec2.ISecurityGroup[];
 
   /**
+   * If a snapshot is specified, a target DB cluster is created from the source
+   * DB snapshot with a default configuration and default security group.
+   *
+   * @default a snapshot ARN to restore from
+   */
+  readonly snapshotIdentifier?: string;
+
+  /**
    * The removal policy to apply when the cluster and its instances are removed
    * or replaced during a stack update, or when the stack is deleted. This
    * removal policy also applies to the implicit security group created for the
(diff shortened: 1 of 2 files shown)


peterwoodworth (commented):
I think you need to PascalCase `SnapshotIdentifier` on line 600 to get the test to pass 🙂 

As for the design, I had imagined that this feature would be a bit more similar to the RDS constructs when discussing on the other issue - we actually have entirely different constructs when created from snapshot in RDS. 

I'm not familiar enough with Neptune to know what the best design will be here - what you have _might_ be the right design here, but this doesn't really offer anything that a one line escape hatch wouldn't also offer, and we will probably be able to come up with a better API. Looking at the CloudFormation documentation, it seems that [`SourceDBClusterIdentifier`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-neptune-dbcluster.html#cfn-neptune-dbcluster-sourcedbclusteridentifier) is important to specify. Additionally, this documentation [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-neptune-dbinstance.html#cfn-neptune-dbinstance-dbsnapshotidentifier) seems to state that this feature isn't possible?

I think this requires a more thorough design discussion, which we probably won't be able to do for a while as we focus on graduating other alpha modules first. Let's continue discussion on the original issue, thanks a ton for your willingness to contribute 🙂

PR #25411, opened 2023-05-03
feat(redshift): Configuration for Redshift User Groups

### LINKED ISSUE #25394: redshift: Redshift User Groups
### Describe the feature

Add an API for creating Redshift User Groups

### Use Case

Sometimes, infra developers may not need to create individual users, but entire groups, and then manage and maintain access to these groups through the Redshift Console. 
If there are a vast number of users within a Redshift Data Warehouse, managing them within a single CDK application may be cumbersome, so these can be managed through a minimal amount of user groups, which can then be managed through the Redshift console, etc.

### Proposed Solution

Add a separate API within the Redshift Construct to access and maintain a group of `Redshift.User`. The Redshift Data API allows the execution of SQL code, which also includes the addition of running `GROUP` commands. Therefore, we can manage and maintain through the existing Redshift Data API already set up.

### Other Information

I am currently working on this, and it seems there will be a need to refactor the `UserTablePrivileges` internal construct, which I can create a separate `refactor` PR for.

### Acknowledgements

### CDK version used

2.76.0

Amazon Linux 2

Rizxcviii:
Adds User Groups into Redshift. This is done by utilising the `GROUP` command supplied by Redshift to create and manage user groups. The user groups themselves are not stateful, therefore when a user group is altered in any way; it is simply a matter of deleting and recreating the user group, rather than making use of `ALTER` commands. User groups can also be assigned privileges. Therefore, we make use of the `UserTablePrivileges` internal construct.
This therefore means that this PR contains a refactor, that changes the way in which the `username` is now recorded. `username`s are now changed to use `accessors` instead, which will determine what type of accessor is being used (User, User Group, etc.). This will enable reusability of future additions to the schema.
Therefore, for this PR, please pay attention to the separate refactor of the `UserTablePrivileges`, and the addition of the User Group construct.

`UserTablePrivileges` refactor:
- username is now changed to accessor, which is an object using its name and type of accessor (user, user group, etc.).
- username still being checked, for initial updates, but has been marked as deprecated to inform future developers of the change.

`UserGroup` feature:
- Test
- integration test
- addition of its privileges
- similar to `User` construct, if you would like to use it for reference.

Closes #25394.

----

### DIFF
--- a/packages/@aws-cdk/aws-redshift-alpha/test/integ.database.js.snapshot/aws-cdk-redshift-cluster-database.assets.json
+++ b/packages/@aws-cdk/aws-redshift-alpha/test/integ.database.js.snapshot/aws-cdk-redshift-cluster-database.assets.json
@@ -27,7 +27,7 @@
         }
       }
     },
-    "a80ba6dcee89bb6f4e42ff7991925527e80dc655ec5678a3917460a015c33f9a": {
+    "a2fbc809aa587013f91da85e316a2f70eb26aecf9ad010a5d307dc121b82f478": {
       "source": {
         "path": "aws-cdk-redshift-cluster-database.template.json",
         "packaging": "file"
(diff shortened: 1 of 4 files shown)


corymhall (commented):
@Rizxcviii thanks for taking the time to create the PR. To be transparent the team will most likely not be able to review this PR anytime soon. I would also argue that we should remove all of the custom resource code that manages the cluster before graduating this module, but that might require a bigger team discussion.

PR #25412, opened 2023-05-03
fix(core): allow functions in DeletionPolicy and UpdateReplacePolicy

bertrandda:
With [`AWS::LanguageExtensions`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-languageextension-transform.html) it's possible to use functions in `DeletionPolicy` and `UpdateReplacePolicy`.

When we try to include a template with a function in DeletionPolicy attribut (or UpdateReplacePolicy) in CDK stack, it fails with error
```
Error: Unrecognized DeletionPolicy '[object Object]'
```

This PR add a call to function parser when parsed cloudformation template contains DeletionPolicy as object.

----


bertrandda:
`Exemption Request`
Changes are only on a helpers function, I added related unit test but I don't think a integration test is necessary.

corymhall (changes requested):
@bertrandda in order to use language extensions don't you need to include the transform in the template?

Can you also add an integration test to make sure that the resulting template deploys? You can add the integ test to this folder https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/framework-integ/test/aws-cloudformation/test

bertrandda:
> @bertrandda in order to use language extensions don't you need to include the transform in the template?
> 
> Can you also add an integration test to make sure that the resulting template deploys? You can add the integ test to this folder https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/framework-integ/test/aws-cloudformation/test

I added an integ tests with a basic template

packages/@aws-cdk-testing/framework-integ/test/aws-cloudformation/test/integ.import-template.ts
```diff
@@ -0,0 +1,17 @@
+import * as core from 'aws-cdk-lib';
+import * as inc from 'aws-cdk-lib/cloudformation-include';
+import * as integ from '@aws-cdk/integ-tests-alpha';
+
+const app = new core.App();
+
+const stack = new core.Stack(app, 'ImportStack');
+
+new inc.CfnInclude(stack, 'ImportStack', {
```
corymhall, line 9: We need a test that doesn't include `CfnInclude`. We also want to test that cdk will synth and deploy a valid template.

corymhall (changes requested):

packages/@aws-cdk-testing/framework-integ/test/aws-cloudformation/test/integ.import-template.ts
bertrandda: I am not sure what do you expected. Just a basic stack creation (with for example a sqs queue)?

PR #25426, opened 2023-05-04
chore: update nodejs to 18 in gitpod

watany-dev:
I love using Gitpod! I saw that node18 support is needed at https://github.com/aws/aws-cdk/pull/25381 and understood the situation.

----


pahud:
Just opened your PR branch and the prebuild with node18 was doing great.

It's not mandatory to opt in 18 but now that we have a PR here I don't see any reason not 😄.

image

.gitpod.yml
```diff
@@ -3,7 +3,7 @@ github:
     pullRequestsFromForks: true
     addComment: true
 
-image: jsii/superchain:1-buster-slim-node16
+image: jsii/superchain:1-buster-slim-node18
```
pahud, line 6: According to https://github.com/aws/jsii/pull/4040 maybe we should use ecr public instead.


```suggestion
image: public.ecr.aws/jsii/superchain:1-buster-slim-node18
```

pahud (commented):
Let's use ecr public.

watany-dev:
Thank you for your review. Certainly this should be obtained from ECR Public!

mrgrain (approved):

PR #25436, opened 2023-05-04
chore(aws-cdk-lib): ignore cfn-types-2-classes.json

mrgrain:
This is a generated file that does not need to be present in the repo.

Also changed the ordering of services.
Previously this was ordered by UTF-16 code unit values. Now it's human friendly, in case anyone ever needs to look at it. This also helps with the transition to the new l1 generation.

----


packages/aws-cdk-lib/.npmignore
```diff
@@ -32,6 +32,9 @@ junit.xml
 
 !*.lit.ts
 
+# keep class map for cloudformation-include
+!cloudformation-include/cfn-types-2-classes.json
```
mrgrain, line 36: I've run `yarn package` locally and confirmed the file is correctly included.

rix0rrr (approved):

PR #25441, opened 2023-05-04
feat(ec2): added support for network interfaces on ec2 instances by providing an associatePublicIpAddress property

### LINKED ISSUE #17127: [aws-ec2]: Add autoAssignPublicIp for ec2.Instance
### Description

Add boolean property of `autoAssignPublicIp` to ec2.Instance properties. 

### Use Case

AutoScalingGroup does this, but in my case the instance is not in an ASG, and my subnet is configured to not map public IPs by default.

### Proposed Solution

I'll leave this to you.

### Other information

_No response_

### Acknowledge

colifran:
## Motivation

When creating and launching an EC2 instance, a public IPv4 address will be assigned by default for any instances being launched into a default public subnet. Conversely, any EC2 instance being launched into a nondefault public subnet will not be automatically assigned a public IPv4 address. The decision to automatically assign or not assign a public IPv4 address is based on a subnet property which is true by default for default public subnets and false by default for nondefault public subnets. This property can be controlled by specifying that the 'associatePublicIpAddress' be true for an EC2 instance. This property can be exposed via the 'networkInterfaces' property on the underlying L1 CfnInstance construct. Furthermore, any network interface that has an 'associatePublicIpAddress' set to true must also be the primary network interface for the EC2 instance and a primary network interface will always have a device index of 0. The work in this PR will allow a user to automatically have a public IPv4 address assigned to an EC2 instance that they are launching into a nondefault public subnet or stop the default subnet behavior of automatically assigning a public IPv4 address.

## Important Changes

The changes made in this PR start by exposing the 'networkInterfaces' property on the underlying L1 CfnInstance. Next, I added 'associatePublicIpAddress' as an optional boolean property that is part of the 'InstanceProps' interface. Importantly, if 'associatePublicIpAddress' is set to true or false, then this means we need to launch the EC2 instance with a configured primary network interface. If 'associatePublicIpAddress' is set to true or false, a network interfaces array is created with the specified network interface configuration for the primary network interface. The subnetId and securityGroupIds are also configured for the network interface since they must be defined on the network interface level when launching an EC2 instance with a configured network interface. I updated the L1 CfnInstance to set subnetId and securityGroupIds to undefined in the event that the network interfaces array is defined. 

Closes #17127

----


packages/aws-cdk-lib/aws-ec2/README.md
```diff
@@ -1547,6 +1547,36 @@ const aspect = new ec2.InstanceRequireImdsv2Aspect();
...
+const instance = new ec2.Instance(this, 'Instance', {
+  vpc,
+  vpcSubnets: { subnetGroupName: 'public-subnet-1' },
+  instanceType: ec2.InstanceType.of(ec2.InstanceClass.T3, ec2.InstanceSize.NANO),
+  machineImage: new ec2.AmazonLinuxImage({ generation: ec2.AmazonLinuxGeneration.AMAZON_LINUX_2 }),
+  detailedMonitoring: true,
+  associatePublicIpAddress: true,
+});
+
+instance.node.addDependency(vpc);
```
corymhall, line 1577: I think you may be able to accomplish in this PR this by using the `internetConnectivityEstablished` dependable.

example
https://github.com/aws/aws-cdk/blob/53d24f1a1357410746052bbf25b5b89b985f7b80/packages/aws-cdk-lib/aws-rds/lib/cluster.ts#L878-L878

```ts
this.instance.node.addDependency(internetConnected);
```

packages/aws-cdk-lib/aws-ec2/lib/instance.ts
```diff
@@ -398,14 +407,22 @@ export class Instance extends Resource implements IInstance {
       });
     }
 
+    // network interfaces array is set to configure the primary network interface if associatePublicIpAddress is true or false
+    const networkInterfaces = props.associatePublicIpAddress !== undefined
```
corymhall, line 411: I wonder if we should add `DeleteOnTermination`. The docs don't state what the default is so I'm not sure. What do you think?

packages/aws-cdk-lib/aws-ec2/README.md
```diff
@@ -1547,6 +1547,36 @@ const aspect = new ec2.InstanceRequireImdsv2Aspect();
 Aspects.of(this).add(aspect);
 ```
 
+### Associating a Public IP Address with an Instance
+
+All subnets have an attribute that determines whether instances launched into that subnet are assigned a public IPv4 address. This attribute is set to true by default for default public subnets. Thus, an EC2 instance launched into a default public subnet will be assigned a public IPv4 address. Nondefault public subnets have this attribute set to false by default and any EC2 instance launched into a nondefault public subnet will not be assigned a public IPv4 address automatically. To automatically assign a public IPv4 address to an instance launched into a nondefault public subnet, you can set the `associatePublicIpAddress` property on the `Instance` construct to true. Alternatively, to not automatically assign a public IPv4 address to an instance launched into a default public subnet, you can set `associatePublicIpAddress` to false. Importantly, if you set `associatePublicIpAddress` and there is a VPC defined in the same stack, you must make sure the `Instance` has a dependency on the VPC to declare a dependency on the VPC-gateway attachment.
```
corymhall, line 1552: Can you add a note here about what happens if you want to switch this value for an existing instance?

corymhall (changes requested):
@colifran looks great! Just a couple of minor comments/questions.

packages/aws-cdk-lib/aws-ec2/lib/instance.ts
colifran: This is a good point. I just launched an instance via the console and the primary network interface has DeleteOnTermination set to true. To ensure we get the same behavior I think it would be good to explicitly set this as true.

packages/aws-cdk-lib/aws-ec2/README.md
colifran: I've updated this section in the README to include a note about what happens when associatePublicIpAddress is included, removed, or updated on an existing instance.

packages/aws-cdk-lib/aws-ec2/README.md
colifran: Good suggestion. I've added this into the instance constructor and updated the README to remove instance.node.addDependency(vpc).

corymhall (approved):
🚀

PR #25446, opened 2023-05-04
docs: add note about stored context when looking up VPC AZs

### LINKED ISSUE #16250: (aws_cdk.aws_ec2): Ability to get stable list of n subnets
Apologies if there is already a way to do this.

I need a way to select a given number of subnets and know I'll always get the same subnets. I don't want to have to make the consumers of these subnets have to know exactly which subnets are available to achieve this.

### Use Case

For putting an Elasticsearch domain in a VPC, I need to have a certain number of subnets available depending on how many instances are in the domain.

For example, if I have 2 data nodes, I need to specify two subnets.

I can do something like (in Python):

            vpc_subnets=aws_ec2.SubnetSelection(
                subnets=vpc.private_subnets[:2]
            )

To get two subnets from the list of private subnets. However, as far as I can see, I have no ordering guarantee from vpc.private_subnets. So I have the danger at any point that the ordering of the subnets may change, and my Elasticsearch domain end up moving subnets unintentionally.

### Proposed Solution

What I'd really like to be able to do is say something like.

            vpc_subnets=aws_ec2.SubnetSelection(
                subnet_type=SubnetType.PRIVATE,
                subnet_count=2,
            )

And know that as long as the list of subnets matching my selection criteria does not change, I'll always get the same two subnets.

peterwoodworth:
Adding this note to be explicitly clear

Closes #16250 

----

### DIFF
--- a/packages/@aws-cdk/aws-servicecatalogappregistry-alpha/lib/application.ts
+++ b/packages/@aws-cdk/aws-servicecatalogappregistry-alpha/lib/application.ts
@@ -57,12 +57,28 @@ export interface IApplication extends cdk.IResource {
   readonly applicationName?: string;
 
   /**
+   * Associate this application with an attribute group.
+   *
+   * @param attributeGroup AppRegistry attribute group
+   */
+  associateAttributeGroup(attributeGroup: IAttributeGroup): void;
+
+  /**
    * Create an attribute group and associate this application with the created attribute group.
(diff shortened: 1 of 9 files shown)


peterwoodworth:
Linter failed with `Error: Not Found` 🤔

corymhall (approved):

PR #25339, opened 2023-04-27
fix(eks): Runtime EOL handling 

watany-dev:
the following runtime support

- Python3.7 
    - 27 Jun 2023
- Node.js 14
   - 30 Apr 2023

Closes #25213

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #25448, opened 2023-05-04
chore(eks): fix integ tests

### LINKED ISSUE #25438: eks: integ.eks-helm-asset.ts is failing
### Describe the bug

In `integ.eks-helm-asset.ts`, two helm charts with exactly the same release name might lead to race condition with the error and fail the integ test.

```
 Error: b'Release "lambda-chart-release" does not exist. Installing it now.\nError: release: already exists\n'
```

https://github.com/aws/aws-cdk/blob/aff7ff463d421eca383e45464a1028bab4cff78c/packages/%40aws-cdk-testing/framework-integ/test/aws-eks/test/integ.eks-helm-asset.ts#L71-L89

### Expected Behavior

`yarn integ` on this integ test should succeed with no error.

### Current Behavior

```
 Error: b'Release "lambda-chart-release" does not exist. Installing it now.\nError: release: already exists\n'
```

### Reproduction Steps

Just `yarn integ` on this test with `--force` to force the deployment.

### Possible Solution

I tried to build dependency between them but the error still occurred. This is because helm chart install is async and when the custom resource returns it does mean the release has completely installed. Generally, we don't add two helm charts with the same release name like that in real environment. We probably should replace one of them with another helm chart to avoid that.

_No response_

### CDK CLI Version

2.77.0

### Framework Version

_No response_

### Node.js Version

1.16

### OS

mac os x

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

pahud:
This PR fixes 2 integ tests from aws-eks.

- `integ.eks-helm-assets` is failing because 2 helm charts are trying to install with the same release name and this is failing with race condition.  This PR replaces one of them with another helm chart from ACK to avoid that and ensures the integ test will succeed.
- `integ.eks-service-account-sdk-call` is missing `package.json` and `sdk-call.js` as described in https://github.com/aws/aws-cdk/issues/25178 and `docker build` is failing due to missing files. We restored them from [v2.72.1](https://github.com/aws/aws-cdk/tree/v2.72.1/packages/%40aws-cdk/aws-eks/test/sdk-call-integ-test-docker-app/app) with minor update on the test.

Now both integ tests can successfully run with no error.

Closes #25438 #25178 

----


pahud:
image

watany-dev:
Indeed, those two things were not going well, and this integration was stalled.
https://github.com/aws/aws-cdk/pull/25339

PR #25464, opened 2023-05-05
chore(prlinter): fixing some errors in the needs review workflow

corymhall:
Fixing some of the remaining issues with the new needs review workflow

- `aws-cdk-automation` does show up as MEMBER so need to filter it out
- If the PR linter failed then we never made it to the assessReview. Make sure we assess review even if the linter fails

----


kaizencc (approved):
Yay

PR #25465, opened 2023-05-05
chore(eks): exclude sdk-call.js in .gitignore

pahud:
The `sdk-call.js` was required but ignored in https://github.com/aws/aws-cdk/pull/25448 because of [this .gitignore rule](https://github.com/aws/aws-cdk/blob/df3ae95b1ea6e4160b877b92993d93b9fbf45226/packages/%40aws-cdk-testing/framework-integ/.gitignore#L2). This file is required and should not be ignored.

This PR adds a `.gitignore` at `packages/@aws-cdk-testing/framework-integ/test/aws-eks/test/sdk-call-integ-test-docker-app/app/.gitignore` to make sure this js will not be excluded.

Closes #<issue number here>.

----


iliapolo:
@pahud This will be fixed as part of https://github.com/aws/aws-cdk/pull/25473 (as we also identified the same problem). 

But I don't think the `.gitignore` file is needed...I guess it was needed when it used to be in `aws-cdk-lib`, where `.js` files are excluded. But since its now in `@aws-cdk-testing/framework-integ`, they aren't excluded anymore. 

Closing this. Let me know if I missed something.

PR #25468, opened 2023-05-06
fix(ec2): non-agnostic stack throws error with availability zones defined in VPC

### LINKED ISSUE #21690: (ec2): cannot provide availabilityZones to VPC
### Describe the bug

I cannot use the availabilityZones prop for VPC without it throwing an error about my stack not being able to use those AZs. I can only use `dummy1a` etc.

Error message:
```
Given VPC 'availabilityZones' us-east-1a must be a subset of the stack's availability zones dummy1a,dummy1b,dummy1c
```

The [docs](https://docs.aws.amazon.com/cdk/api/v1/docs/aws-ec2-readme.html#control-over-availability-zones) state that:
> The number of Availability Zones that are available depends on the region and account of the Stack containing the VPC. If the [region and account are specified](https://docs.aws.amazon.com/cdk/latest/guide/environments.html) on the Stack, the CLI will [look up the existing Availability Zones](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#using-regions-availability-zones-describe) and get an accurate count.

However, specifying the environment on the stack will not force a lookup for viable AZs even when supplying AZs to the availabilityZones prop

The described workaround overriding `get availabilityZones()` works, however that should not be necessary to be able to specify the AZs to use in a new VPC

### Expected Behavior

VPC to use supplied AZs

### Current Behavior

Throws error because the usable Stack AZs are dummy values

### Reproduction Steps

Specify a new VPC

```
    new Vpc(this, 'vpc', {
      availabilityZones: ['us-east-1a']
    })
```

In the stack I've supplied an account number as well as the region. Interestingly enough, _not_ supplying the environment to the stack will lead to successful synth

### Possible Solution

The context lookup is not triggered for Stack.availabilityZones, but it should be triggered in this case, or there should be no check to make sure that the availability zones are valid.

The issue goes away if you specify the availability zones based off the availability zones from the stack

```
    const vpc = new ec2.Vpc(this, 'TestVPC', {
      cidr: '10.0.0.0/16',
      availabilityZones: cdk.Stack.of(this).availabilityZones.sort().slice(0,1)
    })
```

This will ensure a lookup occurs

### CDK CLI Version

2.38

### Framework Version

_No response_

### Node.js Version

.

### OS

.

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

colifran:
## Changes Made

The existing code calls a get availabilityZones method off of the stack. If no context is provided, then two possible return values are possible:
1) The stack is deemed agnostic, i.e., the stack has no defined environment, or the environment only contains just an account or just a region. In this case, the return value will be an array of unresolved tokens.
2) The stack is deemed non-agnostic. In this case, the return value will be ['dummy1a', 'dummy1b', 'dummy1c']

In the event of case 1, the filter method is called off of the array of unresolved tokens and attempts to create a new array of 'resolved stack AZs'. However, since the array that filter is called off of contains unresolved tokens, the result is an empty array. Moving forward, the value 'areGivenAzsSubsetOfStack' will be true only because 'resolvedStackAzs' is of length 0.

In the event of case 2, the filter method is called off of the array containing dummy values - ['dummy1a', 'dummy1b', 'dummy1c']. Since each element in this array will be considered resolved, the resulting array will also just be ['dummy1a', 'dummy1b', 'dummy1c']. Moving forward from this, the value 'areGivenAzsSubsetOfStack' will be false assuming the user is providing actual availability zones, i.e., 'us-east-1a', 'us-east-1b', etc. The previously mentioned error will then be thrown which is unhelpful for the user as the validation done in this case isn't helpful.

As a result, the only useful validation that can occur is in the case that a context has been provided that contains one of two keys:
1) availability-zones:account=${account}:region=${region}
2) aws:cdk:availability-zones:fallback

I've updated the validation logic to check for one of these two keys - it first checks for key 1 and then checks for the availability-zones fallback key (key 2). If either of the two keys is found then 'stackAzs' will be equal to the associated key's value. If neither key is found then 'stackAzs' will be undefined. From here, we check if 'stackAzs' is undefined, and if it is then no useful validation can be done so 'areGivenAzsSubsetOfStackAzs' will be true and we will move on without validation. If 'stackAzs' is not undefined, then we can check that each given availability zone exists within the'stackAzs' array. The error mentioned above will only be thrown if 'areGivenAzsSubsetOfStackAzs' if false.

Closes #21690

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
+++ b/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
@@ -1451,7 +1451,7 @@ export class Vpc extends VpcBase {
 
     if (props.availabilityZones) {
       // If given AZs and stack AZs are both resolved, then validate their compatibility.
-      const resolvedStackAzs = stack.resolvedAvailabilityZones;
+      const resolvedStackAzs = this.resolveStackAvailabilityZones(stack.availabilityZones);
       const areGivenAzsSubsetOfStack = resolvedStackAzs.length === 0 ||
         props.availabilityZones.every(az => Token.isUnresolved(az) ||resolvedStackAzs.includes(az));
       if (!areGivenAzsSubsetOfStack) {
@@ -1719,6 +1719,19 @@ export class Vpc extends VpcBase {
         Account: Stack.of(this).account,
       },
     });
+  }
+
+  /**
+   * Returns the list of resolved availability zones found in the provided stack availability
+   * zones.
+   *
+   * Note: A resolved availability zone refers to an availability zone that is not a token
+   * and is also not a dummy value.
+   */
+  private resolveStackAvailabilityZones(stackAvailabilityZones: string[]): string[] {
+    const dummyValues = ['dummy1a', 'dummy1b', 'dummy1c'];
+    // if an az is resolved and it is not a 'dummy' value, then add it to array as a resolved az
+    return stackAvailabilityZones.filter(az => !Token.isUnresolved(az) && !dummyValues.includes(az));
   }
 }
 
--- a/packages/aws-cdk-lib/core/lib/stack.ts
+++ b/packages/aws-cdk-lib/core/lib/stack.ts
@@ -805,21 +805,6 @@ export class Stack extends Construct implements ITaggable {
   }
 
   /**
-   * Returns the list of resolved AZs that are available in the AWS environment
-   * (account/region) associated with this stack. If there are no resolved AZs,
-   * then an empty array will be returned.
-   *
-   * If no availability zones exist in the context, then they will be reported
-   * as missing and the CLI will resolve them by calling EC2 `DescribeAvailabilityZones`
-   * on the target environment.
-   */
-  public get resolvedAvailabilityZones(): string[] {
-    const dummyValues = ['dummy1a', 'dummy1b', 'dummy1c'];
-    // if an az is resolved and it is not a 'dummy' value, then add it to array as a resolved az
-    return this.availabilityZones.filter(az => !Token.isUnresolved(az) && !dummyValues.includes(az));
-  }
-
-  /**
    * Register a file asset on this Stack
    *
    * @deprecated Use `stack.synthesizer.addFileAsset()` if you are calling,


colifran:
Exemption Request

I'm not sure updating any existing integration tests or adding a new one make sense for this. Hard coding the availabilityZones prop doesn't make sense since the integration tests deploy in parallel in 3 different regions which would cause a failure. I think the current integration test integ.vpc-azs.ts is sufficient and the unit tests I added for this prove the validation logic works correctly.

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1450,14 +1450,15 @@ export class Vpc extends VpcBase {
     Tags.of(this).add(NAME_TAG, props.vpcName || this.node.path);
 
     if (props.availabilityZones) {
-      // If given AZs and stack AZs are both resolved, then validate their compatibility.
-      const resolvedStackAzs = stack.availabilityZones.filter(az => !Token.isUnresolved(az));
```
iliapolo, line 1454: The call to `stack.availabilityZones` is what ensures the CLI will perform the lookup, if you remove this, what guarantees the context will be created?

packages/aws-cdk-lib/aws-ec2/test/vpc.test.ts
```diff
@@ -708,25 +708,73 @@ describe('vpc', () => {
       }).toThrow(/Vpc supports 'availabilityZones' or 'maxAzs', but not both./);
     });
 
-    test('with availabilityZones set correctly', () => {
```
iliapolo, line 711: Changing existing tests is usually a sign of breaking changes...can you explain why its needed?

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
colifran: Thanks for pointing this out. I didn't catch this in the code. With that being said, this should obviously be re-included in the code to ensure the context will be created.

packages/aws-cdk-lib/aws-ec2/test/vpc.test.ts
colifran: The reason for any changes in the unit tests were due to tests being run against 'dummy' values. From what I can tell, the only time we can effectively evaluate whether the provided availability zones are valid for the stack are when the context has the availability zones. Otherwise, we're checking against tokens or 'dummy' values. Given your previous comment I reverted the unit tests that I changed.

packages/aws-cdk-lib/aws-ec2/test/vpc.test.ts
```diff
@@ -719,13 +719,38 @@ describe('vpc', () => {
 
     test('with availabilityZones set to zones different from stack', () => {
       const stack = getTestStack();
+      // we need to create a context with availability zones, otherwise we're checking against dummy values
+      stack.node.setContext(
+        `availability-zones:account=${stack.account}:region=${stack.region}`,
+        ['us-east-1a', 'us-east1b', 'us-east-1c'],
+      );
```
colifran, line 726: The reason for this change in an existing test is because we need to check the provided availability zones against well-defined availability zones, i.e., from a context. If availability zones aren't provided in a context, then we're checking against tokens or 'dummy' values which is unnecessary. With the changes I made, this test no longer throws because we've eliminated any validation when 'dummy' values are returned. Here, I've added AZs in a context so that we're still checking provided AZs against stack AZs.

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1451,12 +1451,9 @@ export class Vpc extends VpcBase {
...
       // If given AZs and stack AZs are both resolved, then validate their compatibility.
-      const resolvedStackAzs = stack.availabilityZones.filter(az => !Token.isUnresolved(az));
-      const areGivenAzsSubsetOfStack = resolvedStackAzs.length === 0 // stack AZs are tokenized, so we cannot validate it
-        || props.availabilityZones.every(
-          az => Token.isUnresolved(az) // given AZ is tokenized, such as in integ tests, so we cannot validate it
-            || resolvedStackAzs.includes(az));
-      if (!areGivenAzsSubsetOfStack) {
+      const resolvedStackAzs = stack.resolvedAvailabilityZones;
+      const areGivenAzsSubsetOfStack = props.availabilityZones.every(az => Token.isUnresolved(az) || resolvedStackAzs.includes(az));
+      if (resolvedStackAzs.length > 0 && !areGivenAzsSubsetOfStack) {
```
colifran, line 1456: This length check is unneeded - I'm updating

packages/aws-cdk-lib/core/lib/stack.ts
```diff
@@ -804,6 +804,21 @@ export class Stack extends Construct implements ITaggable {
...
+  /**
+   * Returns the list of resolved AZs that are available in the AWS environment
+   * (account/region) associated with this stack. If there are no resolved AZs,
+   * then an empty array will be returned.
+   *
+   * If no availability zones exist in the context, then they will be reported
+   * as missing and the CLI will resolve them by calling EC2 `DescribeAvailabilityZones`
+   * on the target environment.
+   */
+  public get resolvedAvailabilityZones(): string[] {
```
corymhall, line 816: After thinking about it some more I think this should just be a private method on the `Vpc` construct. It seems too specific of a use case to put on the `Stack` since most users should not need to care about the `dummy` values.

corymhall (changes requested):

packages/aws-cdk-lib/core/lib/stack.ts
colifran: Update made and changes pushed. I made this a private method as part of the Vpc construct. It's no longer a getter though. I decided the best way to structure this was as "resolveStackAvailabilityZones" which would then be passed the stack availability zones to be resolved.

corymhall (approved):

PR #25473, opened 2023-05-07
fix(eks): overly permissive trust policies

iliapolo:
The *CreationRole* and the *default MastersRole* use the account root principal in their trust policy, which is overly permissive. Instead, use the specific lambda handler roles that need it, and remove the default masters role.

BREAKING CHANGE: A masters role is no longer provisioned by default. Use the `mastersRole` property to explicitly pass a role that needs cluster access. In addition, the creation role no longer allows any identity (with the appropriate `sts:AssumeRole` permissions) to assume it. 

----

### DIFF
--- a/pack.sh
+++ b/pack.sh
@@ -107,9 +107,3 @@ if find dist/ | grep -F "${marker}"; then
   echo "This is expected for builds in a development environment but should not happen in CI builds!"
   exit 1
 fi
-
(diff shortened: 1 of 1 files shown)


iliapolo:
@comcalvi The EKS tests finally pass! :)

But looks like there are a bunch of step-functions tests that need updating as well.

PR #25482, opened 2023-05-08
chore(eks): test app should only exist in framework-integ

corymhall:
The integ tests were moved to `framework-integ` as part of the repo remodel so the test app no longer is needed in `aws-cdk-lib`. 

----


corymhall:
> Hi @corymhall
> 
> We should have added it back in #25448
> 
> And we probably should update the .gitignore in #25465 can you take a look?

ahh so we don't need this test app to exist in `aws-cdk-lib/aws-eks` at all? It just needs to exist in `framework-integ`?

pahud:
@corymhall 

No we don't and yes I believe it's only used in here:

https://github.com/aws/aws-cdk/blob/4038274b812a68993bf6b6cb5b0a3fe971a059ad/packages/%40aws-cdk-testing/framework-integ/test/aws-eks/test/integ.eks-service-account-sdk-call.ts#L20-L22

corymhall:
@Mergifyio queue

PR #25486, opened 2023-05-08
feat(efs): add grantRead and grantReadWrite, grantRootAccess to FileSystem

WinterYukky:
## What change
I added `grantRead()` and `grantReadWrite()`, `grantRootAccess()` to `efs.FileSystem` as Beta1 method.

## Why need this change?
To make IAM authentication easier for clients. 

Currently, v2.78.0 has implemented `grant()` method in `efs.FileSystem`. However, EFS can't restrict only granted client even when customers only use the `grant()` method. Because EFS default file system policy grants full access to any anonymous client that can connect to the file system using a mount target.  To avoid this issue, customers must set file system policies that not grant anonymous clients, to EFS. In this PR, when using the `grantXxx` method that allows IAM authentication for clients, a file system policy that does not allow anonymous clients is set to `efs.FileSystem` by default to suit the customer's use case. Next example is grant read and write access to EC2 Instance.

```ts
declare const client: ec2.Instance;
const fileSystem = new efs.FileSystem(this, 'FileSystem', {
  vpc: new ec2.Vpc(this, 'VPC'),
});
fileSystem.grantReadWrite(client);
```

## How do I continue to allow anonymous access?
You can use `allowAnonymousAccess` props for allow anonymous access.

```ts
declare const client: ec2.Instance;
const fileSystem = new efs.FileSystem(this, 'FileSystem', {
  vpc: new ec2.Vpc(this, 'VPC'),
  allowAnonymousAccess: true,
});
fileSystem.grantRead(client);
```

## Others

The file system policies created to prevent anonymous clients are based on the AWS Management Console.
image

----


packages/aws-cdk-lib/aws-efs/README.md
```diff
@@ -96,14 +96,38 @@ fileSystem.addToResourcePolicy(statement);
 ### Permissions
 
 If you need to grant file system permissions to another resource, you can use the `.grant()` API.
-As an example, the following code gives `elasticfilesystem:ClientWrite` permissions to an IAM role.
+As an example, the following code gives `elasticfilesystem:Backup` permissions to an IAM role.
```
WinterYukky, line 99: To differentiate it from client authentication use cases, we changed the `grant()` example from `elasticfilesystem:clientWrite` to `elasticfileSystem:Backup'.

packages/@aws-cdk-testing/framework-integ/test/aws-efs/test/integ.efs.permission.ts
WinterYukky, line 1: This integration test tests the following:

1. Clients granted by `grantReadWrite()` write files during deployment
2. Assert that anonymous client mounts fail
3. Assert that a client authorized by `grantRead()` will read the file

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -301,10 +318,75 @@ abstract class FileSystemBase extends Resource implements IFileSystem {
...
    */
   public grant(grantee: iam.IGrantable, ...actions: string[]): iam.Grant {
-    return iam.Grant.addToPrincipal({
+    return iam.Grant.addToPrincipalOrResource({
+      grantee: grantee,
+      actions: actions,
+      resourceArns: [this.fileSystemArn],
+      resource: this,
+    });
+  }
```
WinterYukky, line 327: Change the grant function from iam.Grant.addToPrincipal to iam.Grant.addToPrincipalOrResource in grant method, like a same resource that has resource policy.

Examples
- https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-s3/lib/bucket.ts#L968-L988
- https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-sns/lib/topic-base.ts#L134C1-L141
- https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-kms/lib/key.ts#L164-L196

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -301,10 +318,75 @@ abstract class FileSystemBase extends Resource implements IFileSystem {
...
+   * Grant the client actions defined in actions to the given grantee on this File System resource.
+   * If this method is used and the allowAnonymousAccess props are not specified,
+   * anonymous access to this file system is prohibited.
+   *
+   * @param grantee The principal to grant right to
+   * @param actions The client actions to grant
+   * @param conditions The conditions to grant
+   * @internal
+   */
+  public _grantClient(grantee: iam.IGrantable, actions: ClientAction[], conditions?: Record<string, Record<string, unknown>>): iam.Grant {
```
corymhall, line 339: Does this need to be `public` or can it be `private`?

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -301,10 +318,75 @@ abstract class FileSystemBase extends Resource implements IFileSystem {
...
+      resource: this,
+      conditions,
+    });
+  }
+
+  /**
+   * Grant read permissions for this file system to an IAM principal.
+   * @param grantee The principal to grant read to
+   */
+  public grantReadBeta1(grantee: iam.IGrantable): iam.Grant {
```
corymhall, line 354: I think it is safe to remove the `Beta1` from these. Based on the API I can't
foresee any breaking changes that we would need to make.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -409,7 +491,26 @@ export class FileSystem extends FileSystemBase {
       throughputMode: props.throughputMode,
       provisionedThroughputInMibps: props.provisionedThroughputPerSecond?.toMebibytes(),
       backupPolicy: props.enableAutomaticBackups ? { status: 'ENABLED' } : undefined,
-      fileSystemPolicy: Lazy.any({ produce: () => this._fileSystemPolicy }),
+      fileSystemPolicy: Lazy.any({
+        produce: () => {
+          const allowAnonymousAccess = props.allowAnonymousAccess ?? !this._grantedClient;
```
corymhall, line 496: I think this should be the default behavior. Would you mind/be able to create a
new feature flag and make this the new default?

corymhall (dismissed):
Looks great! Especially love the integ test! Just a couple of comments

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
WinterYukky: Does that mean that the behavior of deny anonymous clients should be made the default even when not using `grantClient`?

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
WinterYukky: I was planning to use this method on AccessPoint, so I made it public. Currently, it doesn't need to be public, so I will change to private until open the PR that add grant API to AccessPoint :)

WinterYukky: Incidentally, `_grantClient` is a method created to determine whether a client has been granted or not. If denying anonymous clients is the default behavior, then this method differs from `grant()` only by being able to specify conditions. If that's the case, I thought it would be okay to create a grant method that can specify conditions as `grantBeta1()`, but what do you think?

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
corymhall: yes, but only with the feature flag.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -409,7 +509,28 @@ export class FileSystem extends FileSystemBase {
       throughputMode: props.throughputMode,
       provisionedThroughputInMibps: props.provisionedThroughputPerSecond?.toMebibytes(),
       backupPolicy: props.enableAutomaticBackups ? { status: 'ENABLED' } : undefined,
-      fileSystemPolicy: Lazy.any({ produce: () => this._fileSystemPolicy }),
+      fileSystemPolicy: Lazy.any({
+        produce: () => {
+          const denyAnonymousAccessFlag = FeatureFlags.of(this).isEnabled(cxapi.EFS_DENY_ANONYMOUS_ACCESS) ?? false;
+          const denyAnonymousAccessByDefault = denyAnonymousAccessFlag || this._grantedClient;
```
kaizencc, line 515: help me understand why `this._grantedClient` is included here?

kaizencc (changes requested):
Hi @WinterYukky just a couple minor comments :)

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
WinterYukky: Originally, my design was to reject anonymous clients only when using `grantXXX` (e.g. `grantRead`). Currently, an implementation that rejects anonymous clients by default based on feature flags has been added. Therefore, it is also possible to remove `_grantedClient` and switch to feature flags only.
However, even if a user who has upgraded the CDK from an older version starts using `grantXXX`, anonymous clients will not be rejected unless the feature flag is enabled, so I implemented an implementation that determines the permission of an anonymous client with both the feature flag and `_grantedClient` according to the original purpose. Nonetheless, I'm going to try not to use `_grantedClient` if it's unnecessary because it affects maintainability. What do you think?

rix0rrr (approved):

rix0rrr (approved):

PR #25320, opened 2023-04-26
chore(cfnspec): fix the cfnspec update

rix0rrr:
The cfnspec job currently doesn't create a PR anymore, because the validation is wrong, and fails the job that creates the PR.

Do the additional "did types get renamed" validation only at build time in a PR, not in the job that creates the PR, so we get more visibility on the changes and it's easier to pick up the work to fix it.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #25494, opened 2023-05-09
chore(cfnspec): fix the cfnspec update (#25320)

corymhall:
The cfnspec job currently doesn't create a PR anymore, because the validation is wrong, and fails the job that creates the PR.

Do the additional "did types get renamed" validation only at build time in a PR, not in the job that creates the PR, so we get more visibility on the changes and it's easier to pick up the work to fix it.

Co-authored-by: rix0rrr <rix0rrr@gmail.com>

----


rix0rrr (approved):

corymhall:
@Mergifyio queue

PR #25495, opened 2023-05-09
fix(lambda): validation for FunctionUrlCorsOptions.maxAge

Tietew:
AWS::Lambda::Url's Cors.MaxAge has a maximum value of 86400 (secs)
see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-url-cors.html#cfn-lambda-url-cors-maxage

This PR adds validation for it.

Note: No maximum value for S3, CloudFront (ResponseHeadersPolicy), and API Gateway.

----


Tietew:
Exemption Request

This PR just adds a runtime validation, which is out of scope of integration tests.

packages/aws-cdk-lib/aws-lambda/lib/function-url.ts
```diff
@@ -229,6 +229,10 @@ export class FunctionUrl extends Resource implements IFunctionUrl {
   }
 
   private renderCors(cors: FunctionUrlCorsOptions): CfnUrl.CorsProperty {
+    if (cors.maxAge && !cors.maxAge.isUnresolved() && cors.maxAge.toSeconds() > 86400) {
+      throw new Error(`FunctionUrl CORS maxAge should be less than or equal to 86400 secs (got ${cors.maxAge})`);
```
corymhall, line 233:
```suggestion
      throw new Error(`FunctionUrl CORS maxAge should be less than or equal to 86400 secs (got ${cors.maxAge.toSeconds()})`);
```

Not sure what the value would be otherwise.

corymhall (changes requested):

corymhall (approved):

Naumel (approved):

PR #25525, opened 2023-05-10

### LINKED ISSUE #25309: (cdk-core): CDK diff does not work with umlauts
### Describe the bug

When doing a CDK diff it always detect a change in the tags because the „ü“ in this case results in a „?“. 
Although the Tags on the resource are correct including the umlaut as specified.

### Expected Behavior

Diff command should be capable to handle all allowed chars for tags.

### Current Behavior

[~] AWS::IAM::Role VirtualGuideiaaInstance/Cert/Certificate/CertificateRequestorFunction/ServiceRole VirtualGuideiaaInstanceCertCertificateCertificateRequestorFunctionServiceRole3B312417 
 └─ [~] Tags
     └─ @@ -5,7 +5,7 @@
        [ ] },
        [ ] {
        [ ]   "Key": "customer",
        [-]   "Value": "Messe M?nchen"
        [+]   "Value": "Messe München"
        [ ] },
        [ ] {
        [ ]   "Key": "environment",

### Reproduction Steps

const app = new App();
// ...
Tags.of(app).add("Ära", "München")

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.76.0 (build 78c411b)

### Framework Version

2.76.0

### Node.js Version

v16.17.0

### OS

Mac OS 13.3

### Language

Typescript

### Language Version

3.9.10

### Other information

_No response_

laverdet:
I am reopening this from https://github.com/aws/aws-cdk/pull/25008

> This PR has been deemed to be abandoned, and will be automatically closed. Please create a new PR for these changes if you think this decision has been made in error.

and following up on my comments here:
https://github.com/aws/aws-cdk/pull/24557#issuecomment-1492965718
https://github.com/aws/aws-cdk/pull/24557#issuecomment-1500747296
https://github.com/aws/aws-cdk/pull/25008#issuecomment-1501441939
https://github.com/aws/aws-cdk/pull/25008#issuecomment-1518734007
https://github.com/aws/aws-cdk/pull/25008#issuecomment-1526860767
https://github.com/aws/aws-cdk/pull/25008#issuecomment-1530705196
https://github.com/aws/aws-cdk/pull/25008#issuecomment-1532033381
https://github.com/aws/aws-cdk/pull/25008#issuecomment-1541047453

@aws-cdk-automation @rix0rrr @TheRealAmazonKendra - I'm happy to adjust the approach, add more tests, or do what else needs to be done. I'm not getting any feedback from the team so I'm not sure how to proceed. The diff noise with non-ASCII information in cdk diff makes it difficult to find meaningful changes to our stacks.

🗿🗞️📬 **Crucially, this change only affects the CLI output and therefore an integration test isn't possible.**

---

```
Resources
 └─ [~] Description
     ├─ [-] ?????
     └─ [+] 🤦🏻‍♂️
```

fixes #25309


laverdet:
Exemption Request - This change only affects the CLI output and therefore an integration test isn't possible. I'm happy to spend more time documenting or working through other prerequisites but I do want to make sure that, conceptually, this change is one that the team will accept. I have also included more information about the benefits and tradeoffs of this change in the original PR message.

laverdet:
@skinny85 any chance I could get some eyes on this?

skinny85:
@laverdet unfortunately, I'm no longer on the CDK team, so I can't help with the PR 🙂.

laverdet:
@corymhall Is there any chance you or someone on the team could take a quick look at this? It is a 3-line CLI-only QOL improvement and I'm a little confused as to why I haven't gotten any communication from the team in the 2 months since I opened #24557. If this is not something the team wants then that's ok but right now I'm caught in an awkward holding pattern.

packages/@aws-cdk/cloudformation-diff/test/deep-equal.test.ts
```diff
@@ -0,0 +1,11 @@
+import { deepEqual } from '../lib/diff/util';
+
+test('mangled strings', () => {
+  expect(deepEqual('foo', 'foo')).toBeTruthy();
+  expect(deepEqual('????', '文字化け')).toBeTruthy();
```
otaviomacedo, line 5: Equality, being an equivalence relation, should be reflexive, symmetric and transitive. In particular, for the symmetric property, the following should hold:

`deepEqual('????', '文字化け')` $\iff$ `deepEqual('文字化け', '????')`

which is not the case. The left-hand side is true and the right-hand side is false. But even if we fix this (and assuming transitivity also holds), we get the following nonsensical implication:

```ts
// Given that the following statements are true:
deepEqual('文字化け', '????')
deepEqual('????', '🤦🏻‍♂️')

// Then, by transitivity:
deepEqual('文字化け', '🤦🏻‍♂️')
```

laverdet:
Please don't close it again. It is clearly not abandoned.

PR #25529, opened 2023-05-11
fix(iam): Modify addManagedPolicy to compare ARN instead of instance reference

d0819:
## Issue
When creating a role, the following warning message appeared:
```
Policy large: 11 exceeds 10 managed policies attached to a Role, this requires a quota increase
```

This was caused by the same managed policy being added multiple times.

Although there was only one managed policy in the created template, it appears that the managedPolicies field of the Role class has multiple instances of the same managed policy added to it.

----


d0819:
Exemption Request

This change only affects the `managedPolicies` field that is held within the Role class, and does not have any impact on the generated template. Therefore, I believe that no changes are necessary for testing.

packages/aws-cdk-lib/aws-iam/lib/role.ts
```diff
@@ -551,7 +551,7 @@ export class Role extends Resource implements IRole {
     if (this._precreatedRole) {
       return this._precreatedRole.addManagedPolicy(policy);
     } else {
-      if (this.managedPolicies.find(mp => mp === policy)) { return; }
+      if (this.managedPolicies.some(mp => mp.managedPolicyArn === policy.managedPolicyArn)) { return; }
```
corymhall, line 554: We don't need an integration test, but we do need a unit test.

corymhall (changes requested):

packages/aws-cdk-lib/aws-iam/lib/role.ts
d0819: I tried writing a unit test. Could you please take a look at it?

I added `addToPrincipalPolicy()` to this test because this warning only appears when `defaultPolicy` exists.

otaviomacedo (approved):

PR #25536, opened 2023-05-11
feat(cli): assets can now depend on stacks

rix0rrr:
Introduce a work graph, in which building assets, publishing assets, and deploying stacks are nodes. Each can have their own sets of dependencies which will be respected in a parallel deployment.

This change supports an upcoming change for asset publishing.

----


packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
kaizencc, line 1:
```suggestion
```

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -186,24 +187,31 @@ export class CdkToolkit {
     }
 
     const stacks = stackCollection.stackArtifacts;
-    const assetBuildTime = options.assetBuildTime ?? AssetBuildTime.ALL_BEFORE_DEPLOY;
+    const assetBuildTime = options.assetBuildTime ?? AssetBuildTime.ALL_BEFORE_DEPLOY; // TODO: deal with this
```
kaizencc, line 190:
```suggestion
```

We shouldn't need to care about `assetBuildTime` in this context anymore

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -281,7 +289,7 @@ export class CdkToolkit {
           rollback: options.rollback,
           hotswap: options.hotswap,
           extraUserAgent: options.extraUserAgent,
-          buildAssets: assetBuildTime !== AssetBuildTime.ALL_BEFORE_DEPLOY,
+          buildAssets: assetBuildTime !== AssetBuildTime.ALL_BEFORE_DEPLOY, // TODO: remove this
```
kaizencc, line 292: I flagged this before because I believe this no longer makes sense to have for `deployStack`. We probably want to hardcode it to `buildAssets: false` or remove it altogether.

packages/aws-cdk/lib/util/work-graph-builder.ts
```diff
@@ -0,0 +1,121 @@
...
+        for (const entry of manifest.entries) {
+          const parentStack = parentStacks.get(artifact);
+          if (parentStack === undefined) {
+            throw new Error('Found an asset manifest that is not associated with a stack');
+          }
+          this.addAsset(parentStack, artifact, manifest, entry);
+        }
+      } else if (cxapi.NestedCloudAssemblyArtifact.isNestedCloudAssemblyArtifact(artifact)) {
+        const assembly = new cxapi.CloudAssembly(artifact.fullPath);
+        // FIXME: make artifact IDs unique
```
kaizencc, line 85: still FIXME?

packages/aws-cdk/lib/util/work-graph.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  public node(id: string) {
+    const ret = this.nodes[id];
+    if (!ret) {
+      throw new Error(`No node with id ${id} among ${Object.keys(this.nodes)}`);
+    }
+    return ret;
+  }
+
+  public absorb(graph: WorkGraph) {
```
kaizencc, line 51: what an ugly word lol. Is there any semantic reason you used this over `merge` or just its a cooler word

packages/aws-cdk/test/api/cloudformation-deployments.test.ts
```diff
@@ -884,31 +846,32 @@ test('readCurrentTemplateWithNestedStacks() succesfully ignores stacks without m
...
+//       }),
+//     }),
+//   });
+//   const expectedEnvironment = expect.objectContaining({
+//     account: 'account',
+//     name: 'aws://account/region',
+//     region: 'region',
+//   });
+//   expect(buildAssets).toBeCalledWith(expectedAssetManifest, sdkProvider, expectedEnvironment, undefined);
+// });
```
kaizencc, line 874: are these my changes that still exist in this PR? we are deleting some tests here because they no longer make sense in the new model, right?

packages/aws-cdk/test/cdk-toolkit.test.ts
```diff
@@ -585,63 +586,37 @@ describe('deploy', () => {
...
+    //       hotswap: HotswapMode.FULL_DEPLOYMENT,
+    //     });
+
+    //     expect(mockBuildAssets).toHaveBeenCalledWith(expect.anything(), expect.objectContaining({
+    //       buildOptions: expect.objectContaining({
+    //         parallel: false,
+    //       }),
+    //     }));
+    //   });
+    // });
```
kaizencc, line 619: does this mean we will soon remove these tests, or just commented out while iterating and forgot to uncomment?

packages/aws-cdk/test/work-graph.test.ts
```diff
@@ -0,0 +1,364 @@
...
+  assetDependencies?: string[];
+  id: string;
+  type: 'stack' | 'asset' | 'tree'| 'nested';
+  name?: number;
+  displayName?: string;
+}
+
+function createArtifacts(artifacts: TestArtifact[]) {
+  return artifacts;
+}
```
kaizencc, line 321: huh?

packages/aws-cdk/test/work-graph.test.ts
```diff
@@ -0,0 +1,364 @@
...
+// Not great to have actual sleeps in the tests, but they mostly just exist to give the async workflow
+// a chance to start new tasks.
+const SLOW = 200;
+
+/**
+ * Repurposing unused stack attributes to create specific test scenarios
+ * - stack.name          = deployment duration
+ * - stack.displayName   = error message
+ */
+describe('DeployAssets', () => {
```
kaizencc, line 17:
```suggestion
describe('WorkGraph', () => {
```

kaizencc (commented):
Mostly bookkeeping things to make sure we take care of before releasing

packages/aws-cdk-lib/cx-api/lib/cloud-assembly.ts
```diff
@@ -228,7 +228,7 @@ export class CloudAssembly {
       }
     }
 
-    return topologicalSort(result, x => x.id, x => x._dependencyIDs);
+    return topologicalSort(result, x => x.id, x => x._dependencyIDs); // TODO: remove redundant toposort
```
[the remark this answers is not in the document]
rix0rrr: I'm not comfortable outright removing it. I'll make it an option.

packages/aws-cdk/lib/cdk-toolkit.ts
rix0rrr: We do need it because we'll derive `prebuildAssets: true|false` from this.

packages/aws-cdk/lib/util/work-graph.ts
rix0rrr: Yes. In my mind, it's ambiguous whether `merge` will mutate the LHS in-place, or return a new value.

`absorb` is very obviously mutating the object.

packages/aws-cdk/test/work-graph.test.ts
rix0rrr: A good reason why I don't like top-level `describe()`s.

They usually end up mirroring the file name, for no additional benefit.

packages/aws-cdk/lib/util/work-graph-builder.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+   * Oof, see this parameter list
+   */
+  // eslint-disable-next-line max-len
+  private addAsset(parentStack: cxapi.CloudFormationStackArtifact, assetArtifact: cxapi.AssetManifestArtifact, assetManifest: AssetManifest, asset: IManifestEntry) {
+    const buildId = `${this.idPrefix}${asset.id}-build`;
+
+    // Add the build node, but only one per "source"
+    // FIXME: May need to take directory into account for this key
```
kaizencc, line 46: Has this FIXME been fixed?

packages/aws-cdk/lib/util/work-graph.ts
```diff
@@ -0,0 +1,304 @@
...
+    this.addNodes(...Object.values(graph.nodes));
+  }
+
+  public hasFailed(): boolean {
```
kaizencc, line 87: this used to be public because it was being used in deploy.ts but now it is only being used in the private `forAllArtifacts` method. I think we should make it private as well.

packages/aws-cdk/lib/util/work-graph.ts
```diff
@@ -0,0 +1,304 @@
...
+    return Object.values(this.nodes).some((n) => n.deploymentState === DeploymentState.FAILED);
+  }
+
+  public hasNext(): boolean {
```
kaizencc, line 91: Similarly, `hasNext()` used to be public for reasons but now is only being used in a single unit test. Feels like it should be private as well now

packages/aws-cdk/lib/util/work-graph.ts
```diff
@@ -0,0 +1,304 @@
...
+          await actions.publishAsset(x);
+          break;
+      }
+    });
+  }
+
+  /**
+   * Return the set of unblocked nodes
+   */
+  public ready(): ReadonlyArray<WorkNode> {
```
kaizencc, line 115: same for `ready()`, this is being used in one test

packages/aws-cdk/lib/util/work-graph.ts
```diff
@@ -0,0 +1,304 @@
...
+      }
+
+      start();
+
+      function start() {
+        graph.updateReadyPool();
+
+        for (let i = 0; i < graph.readyPool.length; ) {
+          const node = graph.readyPool[i];
+          // if (node.deploymentState !== DeploymentState.QUEUED) { continue; }
```
kaizencc, line 150:
```suggestion
```

packages/aws-cdk/test/cdk-toolkit.test.ts
kaizencc: was previously marked as resolved but i still see the commented out tests and I feel like if for some reason we wanted to keep these in here, then we should document better why that is the case.

packages/aws-cdk/lib/util/work-graph-builder.ts
rix0rrr: No

packages/aws-cdk/lib/util/work-graph.ts
rix0rrr: I actually kinda like it. But you're right, we don't need it for now.

PR #25543, opened 2023-05-11
chore(aws-cdk-lib): refactor submodule gen

mrgrain:
Submodule generation is split between `cfn2ts` and `aws-cdk-lib/scripts/gen.ts`.
This PR re-organizes the responsibilities between the packages.
`cfn2ts` is now only responsible for generating L1 files (including Canned Metrics and Augmentations).
Any further code gen related to the specific submodule setup in `aws-cdk-lib` has been moved there.
This will make it easier to replace `cfn2ts` with a new version.

Also fixes a few minor issue on the way:
- use alphabetical order in all generated artifacts (previously was ordered by Unicode code point)
- update `lazy-index.ts` again
- move submodule specific code gen in one place
- add README to submodules

With this clean-up version in place, we can now go and add submodule generation as a step of the spec update.
Additionally, spec patches can now be less strict and will continue even on failure.
This is to facilitate the spec bump job, which will do that.
PR builds continue to be strict and will fail.

----

### DIFF
--- a/.github/ISSUE_TEMPLATE/notice.yml
+++ b/.github/ISSUE_TEMPLATE/notice.yml
@@ -1,0 +1,58 @@
+---
+name: "❗ Notice"
+description: Post a notice for a high impact issue
+title: "❗ NOTICE (module name): (short notice description)"
+labels: [p0, management/tracking]
+body:
+  - type: dropdown
+    attributes:
+      label: Status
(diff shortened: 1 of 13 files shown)


Naumel (approved):
Code looks good to me, adding the label in case a second pair of eyes is needed

PR #25556, opened 2023-05-12
feat(toolkit): notificationArns option in cdk import

### LINKED ISSUE #23548: (cli): Support notification-arns in cdk import
### Describe the feature

When using the `cdk import` command (https://github.com/aws/aws-cdk-rfcs/issues/52), there doesn't appear to be a way to pass the `--notification-arns` flag as you can do in a normal deployment. Since an import operation creates a cloudformation change set and `--notification-arns` is already a valid flag on the `cdk deploy` command, it's reasonable to expect that this paramer could be passed for `cdk import` as well.

### Use Case

When updating Cloudformation stacks, processes that trigger off Cloudformation notifications and need a way to be able to pass the SNS topic so they can react to import operations triggered via CDK.

### Proposed Solution

Since `--notification-arns` is already present and supported on `cdk deploy`, this should be implemented the same way to that someone who is used to using that flag on the deploy command can apply it to `cdk import` as well. This would be an optional parameter so it shouldn't negatively impact any existing workflows.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.59.0 and 1.186.1

N/A - New Feature Request

nosredna123:
closes #23548

> REPLACE THIS TEXT BLOCK
>
>
> code you submit.
>

Closes #<issue number here>.

----


andersonmgomes:
Clarification Request
I have searched for references to previous command options in the readme and test files, such as the "--create-resource-mapping" option, but have not found any. Based on this, I believe making any changes to the "notification-arns" option is not mandatory. However, I would appreciate your confirmation on whether I need to make any exclusive modifications to this option.

corymhall (changes requested):
We have an integ test here that tests resource import

https://github.com/aws/aws-cdk/blob/53d24f1a1357410746052bbf25b5b89b985f7b80/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts#L1135-L1135

Can you add a new test that tests resource import with notification-arns? For reference
https://github.com/aws/aws-cdk/blob/53d24f1a1357410746052bbf25b5b89b985f7b80/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts#L469-L469

PR #25566, opened 2023-05-12
chore(msk): deprecate kafka version 1.1.1

### LINKED ISSUE #25565: (msk): kafka version 1.1.1. not longer supported
### Describe the bug

When trying to deploy a Kafka cluster with version `KafkaVersion.V1_1_1` I get the error

```
Resource handler returned message: "[ClientRequestToken: fab03ef7-2048-88e5-b49b-a58d0c91af5a] Unsupported KafkaVersion [1.1.1]. Valid values: [2.8.2.tiered, 2.6.2, 2.7.1, 2.8.0, 2.6.3, 2.7.2, 2.8.1, 2.5.1, 2.6.0, 2.6.1, 2.7.0, 2.3.1, 2.2.1, 2.4.1.1, 3.3.2, 3.1.1, 3.2.0, 3.3.1, 3.4.0]
```

### Expected Behavior

The enum `KafkaVersion` contains only supported versions.

### Current Behavior

`KafkaVersion.V1_1_1` is still implemented.

### Reproduction Steps

```
class KafkaVersionTest extends Stack {
  constructor(scope: Construct, id: string, { version }: {version: KafkaVersion}) {
    super(scope, id, { env });
    const vpc = Vpc.fromLookup(this, 'Vpc', { isDefault: true });

    new msk.Cluster(this, 'Cluster', {
      clusterName: `integ-test-v111`,
      kafkaVersion: KafkaVersion.V1_1_1,
      vpc,
    });
  }
}
```

### Possible Solution

Remove the unsupported versions from the class. This should not break backwards compatibility because they can't be deployed anyways.

_No response_

### CDK CLI Version

2.79.1

### Framework Version

_No response_

### Node.js Version

18

### OS

Mac

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

daschaa:
Closes #25565

As described in the docs, Apache Kafka 1.1.1 is no longer supported (only for existing clusters): [https://docs.aws.amazon.com/msk/latest/developerguide/supported-kafka-versions.html#1.1.1](https://docs.aws.amazon.com/msk/latest/developerguide/supported-kafka-versions.html#1.1.1)

----


peterwoodworth (changes requested):
We would want to mark as deprecated instead. This configuration is valid for clusters that are still on this version, and removing it outright would break those cases.

daschaa:
@peterwoodworth Thank you for your  feedback! I agree that marking this as deprecated is the better option here.
Should we also log something at synthesis time or just add the deprecated flag to the JSDoc string?

peterwoodworth:
Just the deprecated flag in the docstring should be good I think

daschaa:
@peterwoodworth I reverted the changes and added a deprecation note instead.

packages/@aws-cdk/aws-msk-alpha/lib/cluster-version.ts
```diff
@@ -4,6 +4,7 @@
 export class KafkaVersion {
   /**
    * Kafka version 1.1.1
+   * @deprecated Kafka clusters with version 1.1.1. can no longer be created.
```
kaizencc, line 7:
```suggestion
   * **Deprecated by Amazon MSK. You can't create a Kafka cluster with a deprecated version.**
   *
   * Kafka version 1.1.1
   *
   * @deprecated use the latest runtime instead
```

kaizencc (approved):
Thanks @daschaa

PR #25577, opened 2023-05-13
fix(appsync): default resultsCacheTtl 0 seconds instead of 5 minutes

Oatelaus:
Within the documentation for this field it is said

```
    /**
     * How long the results are cached.
     * Disable caching by setting this to 0.
     *
     * @default Duration.minutes(5)
     */
```

But currently not setting this field creates an authoriser with a cache of 0.

Following the approach seen within the APIGateway pair I have added a default timeout which matches the documentation.

https://github.com/aws/aws-cdk/blob/5ccc56975c323ea19fd0917def51184e13f440d9/packages/%40aws-cdk/aws-apigatewayv2-authorizers-alpha/lib/http/lambda.ts#L103

----


Oatelaus:
I didn't create an issue for this. I do not know if this is a hard requirement - please let me know if so. Thanks for reading.

Oatelaus:
I've seen that this doesn't work. I'll close it, fix it, take a breathe and reopen when I'm ready.

😆

PR #25581, opened 2023-05-15
docs(ssm): explain that SecretValue.ssmSecure() lives in core now

### LINKED ISSUE #22593: aws_ssm: valueForSecureStringParameter is deprecated and the alternative isn't clearly documented
### Describe the issue

From here:
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ssm.StringParameter.html#static-valuewbrforwbrsecurewbrstringwbrparameterscope-parametername-versionspan-classapi-icon-api-icon-deprecated-titlethis-api-element-is-deprecated-its-use-is-not-recommended%EF%B8%8Fspan

This text, for `valueForSecureStringParameter`:
> Deprecated: Use SecretValue.ssmSecure() instead, it will correctly type the imported value as a SecretValue and allow importing without version.

There's no mention of what `SecretValue.ssmSecure()` is, which module it belongs to or how to use it. The documentation should point properly point to a non-deprecated alternative.

### Links

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ssm.StringParameter.html#static-valuewbrforwbrsecurewbrstringwbrparameterscope-parametername-versionspan-classapi-icon-api-icon-deprecated-titlethis-api-element-is-deprecated-its-use-is-not-recommended%EF%B8%8Fspan

Zishanwang1992:
Closes #22593.

----


packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
```diff
@@ -579,7 +579,8 @@ export class StringParameter extends ParameterBase implements IStringParameter {
    * @param scope Some scope within a stack
    * @param parameterName The name of the SSM parameter
    * @param version The parameter version (required for secure strings)
-   * @deprecated Use `SecretValue.ssmSecure()` instead, it will correctly type the imported value as a `SecretValue` and allow importing without version.
+   * @deprecated Use `SecretValue.ssmSecure()` instead, it will correctly type the imported value as a `SecretValue` and allow importing without version. [Click here for more
+   * details](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.SecretValue.html#static-ssmwbrsecureparametername-version).
```
kaizencc, line 583:
```suggestion
   * @deprecated Use `SecretValue.ssmSecure()` instead, it will correctly type the imported value as a `SecretValue` and allow importing without version. `SecretValue` lives in the core `aws-cdk-lib` module.
```

kaizencc (approved):
Thanks @Zishanwang1992! The reason why I changed it up is because some places won't be able to render the markdown syntax correctly, and i dont think we need to link to the docs so much as just mention where the new method lives

PR #25600, opened 2023-05-15
chore: update eslint rule to allow max 1 empty line

kaizencc:
There should be no reason to have back-to-back new lines which our current eslint rules allow. I made the rule stricter and fixed instances where we were breaking the stricter rule.

----


kaizencc:
I added `cli-integ-tested` to get past the linter but I didn't actually integ test the cli changes. That is because the only thing this PR changes is removal of excess new lines. No code was changed from this PR

rix0rrr:
What's the reason behind this change?

kaizencc:
> What's the reason behind this change?

Just personal preference. I suppose we should take this up as a vote, since the value we get is "i dislike seeing extra new lines that are unnecessary and try to fix them when i see them" but the unvalue is that we will receive more linter errors now.

mrgrain:
No opinion here. But while you're add it, can you force a newline at the end of each file? ;P

comcalvi (approved):

PR #25604, opened 2023-05-15
fix(lambda-nodejs): ignore noEmit in tsconfig when pre-compiling

### LINKED ISSUE #25603: lambda-nodejs: NodejsFunction not ignoring noEmit in tsconfig when pre-compiling
### Describe the bug

When using the `preCompilation: true` bundling option to `NodejsFunction`  with a tsconfig that contains the compiler option `"noEmit": true`, no compiled .js files are generated by the pre-compilation step before bundling. This may cause `esbuild` to bundle the original .ts files instead, which is not the intended outcome when using pre-compilation.

Some projects may enable `noEmit` in their tsconfig so that they can use `tsc` to perform type checking without generating needless output files, since `NodejsFunction` will do that automatically as part of pre-compilation during deployment.

When `NodejsFunction` builds the `tsc` options from the tsconfig for pre-compilation, it should ignore any `noEmit` compiler option so that output files will always be generated.

### Expected Behavior

A setting of `"noEmit": true` in tsconfig has no effect on the pre-compilation behavior of `NodejsFunction`. Pre-compilation will still output the compiled .js files before bundling.

### Current Behavior

A setting of `"noEmit": true` in tsconfig prevents the pre-compilation option of `NodejsFunction` from working correctly, as no .js files are generated. In this case, `esbuild` bundles the original .ts files, which is not the desired outcome.

### Reproduction Steps

Using a `tsconfig.json` such as:
```json
{
  "compilerOptions": {
    "target": "es2021",
    "module": "commonjs",
    "lib": [
      "es2021",
      "dom"
    ],
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true,
    "esModuleInterop": true,
    "isolatedModules": true,
    "strict": true,
    "allowUnusedLabels": false,
    "noEmit": true,
    "noFallthroughCasesInSwitch": true,
    "noImplicitOverride": true,
    "noImplicitReturns": true,
    "typeRoots": [
      "./node_modules/@types"
    ]
  },
  "exclude": [
    "node_modules",
    "cdk.out"
  ]
}
```

with a `NodejsFunction` such as:
```ts
new NodejsFunction(this, 'some-function', {
    bundling: {
        preCompilation: true,
    },
})
```

will show this issue, where no pre-compiled .js files are generated.

### Possible Solution

Add `'noEmit'` to the `excludedCompilerOptions` in the `getTsconfigCompilerOptions()` utility function.

_No response_

### CDK CLI Version

2.79.1 (build 2e7f8b7)

### Framework Version

_No response_

### Node.js Version

v18.16.0

### OS

macOS 12.6.5

### Language

Typescript

### Language Version

Typescript (5.0.4)

### Other information

_No response_

leestkly:
When building the options to `tsc` for pre-compilation from tsconfig, exclude any `noEmit` compiler option in case it has been set to `true`, which would prevent generation of the compiled .js files. Added `'noEmit'` to the list of excluded options when processing tsconfig.

Closes #25603.

----


leestkly:
I un-skipped the tests for `getTsconfigCompilerOptions()` in order to demonstrate the change. I hope that's ok. Let me know if I should re-skip them.

corymhall (approved):

corymhall (approved):

PR #25610, opened 2023-05-16
chore(core): `ITaggableV2` is safe to implement for L1s

rix0rrr:
`ITaggable` defines a property: `tags: TagManager`.

This definition conflicts with many L1 resources: they have `<cfnProperty>: <PropType>` properties on them, and if a property is named `tags` they conflict.

The old behavior is to have `tags: TagManager` for a select set of resources we recognize, and `tags: CfnResource.TagProperty[]` for all other resources.

There is occlusion whichever way we choose these properties.

Introduce a new interface, `ITaggable2 { tagManager: TagManager }`. I'm going to state there is *very little chance* of upstream properties being called `TagManager`, and so we can have both the `tags: CnfResource.TagProperty[]` property as well as the `tagManager: TagManager` property.

The plan is to generate one of the following L1 patterns:

```ts
// Old
class CfnSomething implements ITaggable {
  readonly tags: TagManager;  // Backwards compatible
  tagsRaw: CfnSomething.TagProperty[]; // Also allow direct access
}

// New
class CfnSomething implements ITaggable2 {
  readonly tags: CfnSomething.TagProperty[];  // Backwards compatible
  readonly cdkTagManager: TagManager;  // Allow access to the TagManager
}
```

----


corymhall (approved):
I don't see this called out, but I'm assuming that this is a two step process where this PR is setting up the prerequisites and a future PR will update the codegen to generate the `new` version of the resource (for new resources)? If that is the case then feel free to remove the do-not-merge label.

rix0rrr:
It is that! But before we merge @mrgrain suggested we do a round of consensus gathering around the naming

PR #25625, opened 2023-05-17
fix(cdk-lib): Pass lookupRoleArn to NestedStackSynthesizer

### LINKED ISSUE #25171: pipelines: Cannot perform lookup in cross-account nested stack
### Describe the bug

When a resource is created in a NestedStack in a cross-account pipeline, the Build/Synth phase failed with he following error message: Need to perform AWS calls for account 222222222222, but the current credentials are for 111111111111.

111111111111 is the pipeline account
222222222222 is the target account where the resource should be deployed.

If the same resource is moved to the parent stack everything works.

### Expected Behavior

Resource creation should work in NestedStack in cross-account pipelines.

### Current Behavior

Resource creation failed in NestedStack in cross-account pipelines.

### Reproduction Steps

Create a pipeline.
Create a stage in a different account.
Create a NestedStack inside de stage stack.
Create a resource in the Nested Stack.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.74.0

### Framework Version

_No response_

### Node.js Version

v16.20.0

### OS

Amazon-Linux 2023

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

### LINKED ISSUE #25624: aws-cdk-lib: NestedStacks dont perform lookups on a cross-account setup
### Describe the bug

When setting a cross-account CDK project, if one of the Stacks being deployed on a target account (different than the deployment account) has a NestedStacks which needs to perform a lookup operation (for example, to create a VPC), then CDK is having credential issues, not being able of doing those lookups and interrupting the synth process.

This is because the NestedStack's synthesizer doesn't receive the lookupRoleArn from the parent stack synthesizer, so the NestedStack tries with local credentials (of the deployment account) instead of assuming a cross-account role (on the target account) as regular non-nested Stack would do.

A workaround for this is to manually define the required context on cdk.context.json so CDK doesn't try to fetch the information.

This is of special interest on CDK's projects being deployed by [cdk-pipelines](https://docs.aws.amazon.com/cdk/v2/guide/cdk_pipeline.html)

### Expected Behavior

An CDK's NestedStack should have the parent's lookupRoleArn so it can perform a succesful lookup and add the information to the context provider.

### Current Behavior

When trying an cdk synth, the following error is being thrown:

`[Error at /PipelineStack/App/AppStack/Nested-Stack] Need to perform AWS calls for account XXXXXXXX, but the current credentials are for YYYYYYYY`

Where YYYYYYY is the deployment account (where the synth is being made) and XXXXXXXX is the target account.

### Reproduction Steps

**pipeline.ts**
```
#!/usr/bin/env node
import { PipelineStack } from '../lib/pipeline-stack';

const app = new cdk.App();
new PipelineStack(app, 'PipelineStack', {
  /* If you don't specify 'env', this stack will be environment-agnostic.
   * Account/Region-dependent features and context lookups will not work,
   * but a single synthesized template can be deployed anywhere. */

  /* Uncomment the next line to specialize this stack for the AWS Account
   * and Region that are implied by the current CLI configuration. */
  // env: { account: process.env.CDK_DEFAULT_ACCOUNT, region: process.env.CDK_DEFAULT_REGION },

  /* Uncomment the next line if you know exactly what Account and Region you
   * want to deploy the stack to. */
  env: { account: '123456789012', region: 'us-east-1' },

  /* For more information, see https://docs.aws.amazon.com/cdk/latest/guide/environments.html */
});

```
**pipeline-stack.ts**
```
import { CodePipeline, CodePipelineSource, ShellStep } from 'aws-cdk-lib/pipelines';

const env = { account: '1234567890', region: 'us-east-1' };

export class PipelineStack extends cdk.Stack {
    super(scope, id, props);
    
    const pipeline = new CodePipeline(this, 'Pipeline', {
      pipelineName: 'Pipeline',
      crossAccountKeys: true,
      synth: new ShellStep('Synth', {
        input: CodePipelineSource.gitHub('OWNER/REPO', 'main'),
        commands: ['npm ci', 'npm run build', 'npx cdk synth']
      })
    });
    
    pipeline.addStage(new AppStage(this, "App", {
      env: env
    }));
    
  }
}

export class AppStage extends cdk.Stage {
    constructor(scope: Construct, id: string, props?: cdk.StageProps) {
      super(scope, id, props);

      const lambdaStack = new AppStack(this, 'AppStack');
    }
}


export class AppStack extends cdk.Stack {
    super(scope, id, props);
    
    const nested = new NestedStack(this);
  }
}

class NestedStack extends cdk.NestedStack {
  constructor(scope: Construct) {
    super(scope, 'Nested-Stack');
    const vpc = new ec2.Vpc(this, 'VPC_nested', {
      ipAddresses: ec2.IpAddresses.cidr('10.0.0.0/16'),
    })
  }
}
```

### Possible Solution

Add the variable lookupRoleArn on the Synthesizers Interfaces and invoke the function synthesizeTemplate of class NestedStackSynthesizer accordingly

_No response_

### CDK CLI Version

2.79.1 (build 2e7f8b7)

### Framework Version

_No response_

### Node.js Version

v16.20.0

### OS

Amazon Linux 2

### Language

Typescript, Python

### Language Version

_No response_

### Other information

https://github.com/aws/aws-cdk/issues/21690

IgnacioAcunaF:
NestedStack's synthesizer doesn't receive the lookupRoleArn from the parent stack synthesizer, so the NestedStack tries with local credentials (of the deployment account) instead of assuming a cross-account role (on the target account) as regular non-nested Stack would do.

This PR aims to add lookupRoleArn reference to the StackSynthesizer class and IStackSynthesizer, so it can be use on the NestedStack to explicitly set an IAM role in case of parent stack having one already defined, so CDK uses the role instead of local credentials.

Closes #25171.

----

### DIFF
--- a/packages/aws-cdk-lib/core/lib/stack-synthesizers/default-synthesizer.ts
+++ b/packages/aws-cdk-lib/core/lib/stack-synthesizers/default-synthesizer.ts
@@ -343,6 +343,13 @@ export class DefaultStackSynthesizer extends StackSynthesizer implements IReusab
     return this.qualifier;
   }
 
+  /**
+   * The role used to lookup for this stack
+   */
+  public get lookupRole(): string | undefined {
+    return this.lookupRoleArn;
+  }
+
   public bind(stack: Stack): void {
     super.bind(stack);
 
--- a/packages/aws-cdk-lib/core/lib/stack-synthesizers/nested.ts
+++ b/packages/aws-cdk-lib/core/lib/stack-synthesizers/nested.ts
@@ -34,6 +34,6 @@ export class NestedStackSynthesizer extends StackSynthesizer {
   public synthesize(session: ISynthesisSession): void {
     // Synthesize the template, but don't emit as a cloud assembly artifact.
     // It will be registered as an S3 asset of its parent instead.
-    this.synthesizeTemplate(session);
+    this.synthesizeTemplate(session, this.parentDeployment.lookupRole);
   }
 }
--- a/packages/aws-cdk-lib/core/lib/stack-synthesizers/stack-synthesizer.ts
+++ b/packages/aws-cdk-lib/core/lib/stack-synthesizers/stack-synthesizer.ts
@@ -27,6 +27,13 @@ export abstract class StackSynthesizer implements IStackSynthesizer {
     return undefined;
   }
 
+  /**
+   * The role used to lookup for this stack
+   */
+  public get lookupRole(): string | undefined {
+    return undefined;
+  }
+
   private _boundStack?: Stack;
 
   /**
--- a/packages/aws-cdk-lib/core/lib/stack-synthesizers/types.ts
+++ b/packages/aws-cdk-lib/core/lib/stack-synthesizers/types.ts
@@ -14,6 +14,13 @@ export interface IStackSynthesizer {
   readonly bootstrapQualifier?: string;
 
   /**
+   * The role used to lookup for this stack
+   *
+   * @default - no role
+   */
+  readonly lookupRole?: string;
+
+  /**
    * Bind to the stack this environment is going to be used on
    *
    * Must be called before any of the other methods are called, and can only be called once.


peterwoodworth:
@IgnacioAcunaF can you please either add tests, or [make a request](https://github.com/aws/aws-cdk/pull/25625#issuecomment-1551944271)? We won't be able to provide a meaningful review otherwise

IgnacioAcunaF:
@peterwoodworth sure, will work on that!

PR #25642, opened 2023-05-18
fix(lambda-nodejs): cannot use .mts, .cts, and .cjs entry files

### LINKED ISSUE #21635: Lambda: Can't use typescript module files
### Describe the bug

We've tried to compile ESM output (.mjs) files from .mts (TypeScript Module) files via CDK "NodejsFunction". We don't want to change everything in our project to be JavaScript module code. So the only option is to use .mjs, or in our case, .mts files.
But someone thought it is a good idea to prohibit any other file extensions than .jsx?, .tsx? or .mts. So we're constantly getting an error that .mts is not valid. (which it is and should be allowed)

https://github.com/aws/aws-cdk/blob/28914bdb50b7d11ca02f2d815b292297521489c1/packages/aws-cdk-lib/aws-lambda-nodejs/lib/function.ts#L165

As soon as I remove the referenced line from the functions.js it compiles and deploys fine ...

### Expected Behavior

At least .mts should be allowed as well if there is .ts and .mjs allowed.

https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions

Maybe it is a good idea to allow .cjs and .cts as well these days 😅 

### Current Behavior

Error "Only JavaScript or TypeScript entry files are supported." is thrown even for perfectly valid TypeScript files. Probably even for perfectly valid JavaScript files with the ending .cjs.

### Reproduction Steps

Stack code to create lambda using mts file and ESM format.
```
import * as lambda from "aws-cdk-lib/aws-lambda-nodejs";
import { Runtime as LambdaRuntime } from "aws-cdk-lib/aws-lambda";

new lambda.NodejsFunction(this, 'DummyLambda', {
    entry: path.join('src', `DummyLambda.mts`),
    runtime: LambdaRuntime.NODEJS_16_X,
    bundling: {
        charset: lambda.Charset.UTF8,
        format: lambda.OutputFormat.ESM,
        target: "node16.15",
        @see https://github.com/evanw/esbuild/issues/584#issuecomment-740170568
        esbuildArgs: {
            "--main-fields": "module,main",
        },
    },
});
```

Contents of specified DummyLambda.mts can be any typescript code. Doenst matter for generating the error. The file isn't even required to exist.

### Possible Solution

Regarding to https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions at least .mts should be allowed. I would suggest to allow .cjs and .cts as well.

_No response_

### CDK CLI Version

2.37.1

### Framework Version

_No response_

### Node.js Version

All

### OS

All

### Language

Typescript

### Language Version

All

### Other information

_No response_

ersims:
This adds support for using TypeScript Module (`.mts`), TypeScript CommonJS (`.cts`) and JavaScript CommonJS (`.cjs`) entry files, as well as adding these to the default entry file search. 

This is a must for projects mixing ESM and CommonJS (quite common as not everything can/should be fully ESM yet) or where explicit file extension is desired.

I tried to follow the existing convention as best I could to keep it as simple as possible. I am not sure why the tests for the existing entry files are skipped as they seem to be working correctly, at least locally, so I enabled the new tests I added.

Closes #21635.

----


corymhall (dismissed):

corymhall (approved):

packages/@aws-cdk-testing/framework-integ/test/aws-lambda-nodejs/test/integ.modules.ts
```diff
@@ -0,0 +1,60 @@
...
+      entry: path.join(__dirname, 'integ-handlers/cjs-handler.cjs'),
+    });
+  }
+}
+
+const app = new App();
+const mtsEntryTestCase = new TypeScriptModuleStack(app, 'cdk-integ-lambda-nodejs-modules-mts');
+const ctsEntryTestCase = new TypeScriptCommonJsStack(app, 'cdk-integ-lambda-nodejs-modules-cts');
+const cjsEntryTestCase = new JavaScriptCommonJsStack(app, 'cdk-integ-lambda-nodejs-modules-cjs');
+
```
corymhall, line 42:
```suggestion
```

corymhall (approved):

PR #25643, opened 2023-05-18
feat(batch-alpha): tag instances launched from your managed CEs

comcalvi:
`Tags.of()` now applies to `ComputeResources.Tags`

----


otaviomacedo (commented):
Could we achieve the same result by calling this new property `tags` (and relaying it to the `tags` property in the L1), so that users could use the `Tags` class?

```ts
Tags.of(computeEnvironment).add('key', 'value')
```

PR #25644, opened 2023-05-18
feat(cli): logging can be corked

kaizencc:
🍾

This PR extends #25536 by fixing issues with logging.

- Asset building and publishing are now completely separate tasks, so there is never a need for the "Building and Publishing" message in cdk-assets. I've removed a good chunk of unnecessary private props in `AssetPublisher` and now we simply print `Building` when building an asset and `Publishing` when publishing an asset. No combos anymore.
- Asset build/publish can now happen concurrently with stack deployments when there are no dependencies between the two, but if `--require-approval` is set (which it is by default), sensitive stack deployments prompt the user for a `y/n` response before deployment. Additional asset related messages may come in at this time, cluttering the log. The solution here is to implement a cork that is turned on when prompting the user and turned off after user input. When using the helper function `withCorkedLogging(callback)`, logs will instead be stored in memory and released when the cork is popped.

Testing:

There's not a great way to test these changes in code since they should only affect logging. Instead, I hope the following photos suffice:

Before the lock change, logging looked like this:
Screen Shot 2023-05-18 at 4 59 35 PM

Now it looks like this in the same scenario: 
 
Screen Shot 2023-05-18 at 4 49 39 PM

The screenshots also show the logs that say `Building` and `Publishing` separately rather than `Building and Publishing` as it did before.

----

### DIFF
--- a/packages/aws-cdk/lib/logging.ts
+++ b/packages/aws-cdk/lib/logging.ts
@@ -31,6 +31,7 @@ function uncorkLogging() {
   CORK_COUNTER -= 1;
   if (!corked()) {
     logBuffer.forEach(([stream, str]) => stream.write(str + '\n'));
+    logBuffer.splice(0);
   }
 }
 


packages/aws-cdk/lib/logging.ts
```diff
@@ -7,6 +7,11 @@ const { stdout, stderr } = process;
 
 type WritableFactory = () => Writable;
 
+// LOGGER_LOCKED is an array rather than a boolean because it needs to be modified by other
```
kaizencc, line 10:
```suggestion
// LOG_LOCK is an array rather than a boolean because it needs to be modified by other
```

packages/aws-cdk/lib/logging.ts
```diff
@@ -7,6 +7,11 @@ const { stdout, stderr } = process;
 
 type WritableFactory = () => Writable;
 
+// LOG_LOCK is an array rather than a boolean because it needs to be modified by other
+// parts of the CLI and imported variables are always immutable.
+export const LOG_LOCK = [false];
```
kaizencc, line 12: this is almost certainly the hackiest thing i've ever come up with. whats the better way to do something like this?

kaizencc: ok i thought of a better way immediately lol

packages/aws-cdk/lib/logging.ts
```diff
@@ -7,6 +7,14 @@ const { stdout, stderr } = process;
 
 type WritableFactory = () => Writable;
 
+let LOG_LOCK = false;
+
+export function setLogLock(lock: boolean) {
+  LOG_LOCK = lock;
```
rix0rrr, line 13: Using a boolean is not re-entrancy safe. Make it a counter, increment by 1 every time someone locks and decrement by 1 every time someone unlocks.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -254,6 +256,11 @@ export class CdkToolkit {
           }
 
           const confirmed = await promptly.confirm('Do you wish to deploy these changes (y/n)?');
+
+          setLogLock(false);
+          // Calls the logger and prints any logs that happened during the lock
+          print('');
```
rix0rrr, line 262: `setLogLock(false)` could/should flush buffered messages straight away, so you don't need the extra `print`.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -238,6 +238,8 @@ export class CdkToolkit {
       if (requireApproval !== RequireApproval.Never) {
         const currentTemplate = await this.props.deployments.readCurrentTemplate(stack);
         if (printSecurityDiff(currentTemplate, stack, requireApproval)) {
+          // Lock the logger from logging temporarily
+          setLogLock(true);
```
rix0rrr, line 242: This feature is uncommonly (but amusingly) called "corking" [one](https://www.educative.io/answers/what-is-stream-module-cork-in-nodejs) [two](https://lists.freedesktop.org/archives/pulseaudio-discuss/2009-December/005767.html).

Can we call it `corkLogging()` and `uncorkLogging()` ?

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -254,6 +256,11 @@ export class CdkToolkit {
           }
 
           const confirmed = await promptly.confirm('Do you wish to deploy these changes (y/n)?');
+
+          setLogLock(false);
```
rix0rrr, line 260: This needs to be in a `try/finally` or the cleanup might not happen in case of an exception.

You could also make:

```ts
async function withCorkedLogging<A>(block: () => Promise<A>): Promise<A> {
  corkLogging();
  try {
    return block();
  } finally {
    uncorkLogging();
  }
}
```

rix0rrr (approved):
I like the solution! Needs some changes before it can be merged but conditionally approved

PR #25646, opened 2023-05-19
feat(stepfunctions): add getters for context object fields

### LINKED ISSUE #25415: stepfunctions: easy access to fields in execution context object
### Describe the feature

Add functions for accessing fields like `name` or `id` from the execution context object. We already have a function for accessing the Task Token: https://github.com/aws/aws-cdk/blob/96bb8ce6b3a45daf47d4d4dbf91fd6b69988bb7b/packages/aws-cdk-lib/aws-stepfunctions/lib/fields.ts#L75

### Use Case

For accessing the task token, we can do something like this:

```
sfn.JsonPath.taskToken
```

However, when it comes to accessing something like the name of the Execution, we have to do something like this:

```
sfn.JsonPath.stringAt("$$.Execution.Name")
```

### Proposed Solution

Add functions for getting all of the remaining fields from the context object.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.72.0

macOS 12.6.5

colifran:
This PR adds additional getters to allow easy access to context object fields. We currently have getters allowing easy access to the entire context object and the context object task token field. This adds getters for the following additional context object fields:
- Execution id
- Execution input
- Execution name
- Execution role arn
- Execution start time
- State entered time
- State name
- State retry count
- State machine id
- State machine name

Closes #25415

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
+++ b/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
@@ -1451,13 +1451,11 @@ export class Vpc extends VpcBase {
 
     if (props.availabilityZones) {
       // If given AZs and stack AZs are both resolved, then validate their compatibility.
-      const resolvedStackAzs = stack.availabilityZones.filter(az => !Token.isUnresolved(az));
-      const areGivenAzsSubsetOfStack = resolvedStackAzs.length === 0 // stack AZs are tokenized, so we cannot validate it
-        || props.availabilityZones.every(
-          az => Token.isUnresolved(az) // given AZ is tokenized, such as in integ tests, so we cannot validate it
-            || resolvedStackAzs.includes(az));
+      const resolvedStackAzs = this.resolveStackAvailabilityZones(stack.availabilityZones);
+      const areGivenAzsSubsetOfStack = resolvedStackAzs.length === 0 ||
+        props.availabilityZones.every(az => Token.isUnresolved(az) ||resolvedStackAzs.includes(az));
       if (!areGivenAzsSubsetOfStack) {
-        throw new Error(`Given VPC 'availabilityZones' ${props.availabilityZones} must be a subset of the stack's availability zones ${stack.availabilityZones}`);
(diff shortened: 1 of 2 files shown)


packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/lambda/integ.invoke.ts
```diff
@@ -49,6 +49,18 @@ const checkJobState = new LambdaInvoke(stack, 'Check the job state', {
   resultSelector: {
     status: sfn.JsonPath.stringAt('$.Payload.status'),
   },
+  payload: sfn.TaskInput.fromObject({
```
corymhall, line 52: Can you add an assertion here to assert that the state machine execution
succeeded? The reason that the assertion is needed is because this test
currently only tests that the state machine is able to be deployed, but there is
not any validation that the state machine is actually valid and will execute.

You can find an example here
https://github.com/corymhall/aws-cdk/blob/53d24f1a1357410746052bbf25b5b89b985f7b80/packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/aws-sdk/integ.call-aws-service-sfn.ts#L38-L49

You can probably also have the lambda function validate that all those fields
are set.

corymhall (changes requested):
Just 1 comment on the integration test.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/lambda/integ.invoke.ts
```diff
@@ -75,4 +95,20 @@ new cdk.CfnOutput(stack, 'stateMachineArn', {
   value: sm.stateMachineArn,
 });
 
+const integ = new IntegTest(app, 'IntegTest', {
+  testCases: [stack],
+});
+const res = integ.assertions.awsApiCall('StepFunctions', 'startExecution', {
+  stateMachineArn: sm.stateMachineArn,
+});
+const executionArn = res.getAttString('executionArn');
```
corymhall, line 104: Nice!

corymhall (approved):
🚀

PR #25660, opened 2023-05-20
feat(events-targets): support assignPublicIp flag to EcsTask

### LINKED ISSUE #9233: [aws-events-targets] Add flag for Auto-assign public IP to EcsTask
With ECS tasks I ran into an issue where ECS is unable to pull images from ECR even though the task is set to run in a PUBLIC subnet. In the console it appears that it is possible to set the Auto-assign public IP for an Event target. 

### Use Case

I have a cloudwatch event that is set to run every week that fires an ECS task. The ECS task needs to be able to pull the ECR image and communicate with external resources and therefore needs a public IP address. 

My current event and target: 

```typescript
    const encryptionEventRule = new Rule(this, `${this.projectTag}-encryption-event-rule`, {
      schedule: Schedule.expression("rate(7 days)"),
    });

    encryptionEventRule.addTarget(
      new EcsTask({
        cluster,
        taskDefinition: encryptionTaskDefinition,
        taskCount: 1,
        subnetSelection: VPC.selectSubnets({
          subnetType: SubnetType.PUBLIC,
        }),
      })
    );
```

### Proposed Solution

Just like ECS services I believe a flag for the Auto-assign public IP can be added

In an ECS Service
```typescript
    const service = new FargateService(this, SERVICE_NAME, {
      serviceName: SERVICE_NAME,
      taskDefinition: taskDefinition,
      cluster: cluster,
      desiredCount: SERVICE_DESIRED_CONTAINER_COUNT,
      assignPublicIp: SERVICE_ASSIGN_PUBLIC_IP,
      securityGroup: serviceSecurityGroup,
      vpcSubnets: VPC.selectSubnets({
        subnetType: SubnetType.PUBLIC,
      }),
    });
```

My proposal for an ECS Task event target:
```typescript
    encryptionEventRule.addTarget(
      new EcsTask({
        cluster,
        taskDefinition: encryptionTaskDefinition,
        taskCount: 1,
        assignPublicIp: SERVICE_ASSIGN_PUBLIC_IP,
        subnetSelection: VPC.selectSubnets({
          subnetType: SubnetType.PUBLIC,
        }),
      })
    );
```

### Other

The CloudWatch console allows for setting the Auto-assign public IP value when creating an Event Rule that triggers ECS tasks as described [here](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_AwsVpcConfiguration.html).

---

ymhiroki:
This feature supports `assignPublicIp` to `EcsTask`.

It specifies whether the task's elastic network interface receives a public IP address.
You can enable it only when `LaunchType` is `FARGATE`.
In this commit, the choice logic of the `LaunchType` keeps the backwards compatibility.

Closes #9233 

----


packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts
```diff
@@ -201,14 +211,19 @@ export class EcsTask implements events.IRuleTarget {
     const tagList = this.tags;
 
     const subnetSelection = this.props.subnetSelection || { subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS };
-    const assignPublicIp = subnetSelection.subnetType === ec2.SubnetType.PUBLIC ? 'ENABLED' : 'DISABLED';
+
+    const assignPublicIp = (this.assignPublicIp ?? subnetSelection.subnetType == ec2.SubnetType.PUBLIC) ? 'ENABLED' : 'DISABLED';
+    const launchType = this.taskDefinition.isEc2Compatible ? 'EC2' : 'FARGATE';
+    if (assignPublicIp == 'ENABLED' && launchType != 'FARGATE') {
```
corymhall, line 217: Won't this throw an error anytime the subnet selection is public?

packages/aws-cdk-lib/aws-events-targets/test/ecs/event-rule-target.test.ts
```diff
@@ -537,74 +537,6 @@ test('Can use same fargate taskdef multiple times in a rule', () => {
   }))).not.toThrow();
 });
 
-test('Isolated subnet does not have AssignPublicIp=true', () => {
```
corymhall, line 540: We should not need to remove any tests.

corymhall (changes requested):

packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts
```diff
@@ -201,14 +211,19 @@ export class EcsTask implements events.IRuleTarget {
     const tagList = this.tags;
 
     const subnetSelection = this.props.subnetSelection || { subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS };
-    const assignPublicIp = subnetSelection.subnetType === ec2.SubnetType.PUBLIC ? 'ENABLED' : 'DISABLED';
+
+    const assignPublicIp = (this.assignPublicIp ?? subnetSelection.subnetType == ec2.SubnetType.PUBLIC) ? 'ENABLED' : 'DISABLED';
```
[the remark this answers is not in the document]
ymhiroki: The public IP addresses can be enabled even in the private subnets. (of course, it's impossible to communicate with the Internet and a meaningless operation.)
I think we shouldn't throw an error in the case since it's the valid setting.

packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts
ymhiroki: No, I added [an integration test case with the `public` subnet selection](https://github.com/aws/aws-cdk/blob/9f9e59731328a8bd8fc6d3aef4f444df73e71ebc/packages/%40aws-cdk-testing/framework-integ/test/aws-events-targets/test/ecs/integ.event-fargate-task.ts#L53) and it passed the test.
The task with `ecs.FargateTaskDefinition`  isn't `isEc2Compatible`  and its `launchType` is `FARGATE`.

packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts
corymhall, line 215:
```suggestion
    const assignPublicIp = (this.assignPublicIp ?? subnetSelection.subnetType === ec2.SubnetType.PUBLIC) ? 'ENABLED' : 'DISABLED';
```

packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts
corymhall, line 217:
```suggestion
    if (assignPublicIp === 'ENABLED' && launchType !== 'FARGATE') {
```

corymhall (changes requested):

Issue #23394, opened 2022-12-19
(events): L2 for EventBridge Scheduler


tmokmss:
### Describe the feature

EventBridge Scheduler was released on 2022/11. It is basically a powerful version of the existing EventBridge scheduled rules, with more scalability, timezone support, time window schedules, etc. [Introducing Amazon EventBridge Scheduler](https://aws.amazon.com/blogs/compute/introducing-amazon-eventbridge-scheduler/)

It would be great if we have a user friendly interface for this resource just like the existing L2 constructs in `aws-events` module. CloudFormation reference is [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-scheduler-schedule.html).

**NOTE**: L1 constructs are already available: [doc](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_scheduler-readme.html)

### Use Case

It will improve most of the use cases that currently use EventBridge scheduled rules.

### Other Information

The RFC process is completed: https://github.com/aws/aws-cdk-rfcs/pull/486

See the referenced PRs to check the progress of implementation.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.52.0

### Environment details (OS name and version, etc.)

n/a

mascur:
Hi @tmokmss,

Thank you for this feature request! I think this would be a great addition! If you would like to contribute then you'll need to submit an [RFC](https://github.com/aws/aws-cdk-rfcs) first [because this is a new L2 construct](https://github.com/aws/aws-cdk/blob/v1-main/CONTRIBUTING.md#step-2-design).

I've marked this issue as a `p2`. We most likely won't be able to get to this issue. However, if you or another community member would like to submit a PR we might be able to review that. You can refer to the [Contribution Guide here](https://github.com/aws/aws-cdk/blob/v1-main/CONTRIBUTING.md) to get started.

We use +1s to help prioritize our work, and are happy to revaluate this issue based on community feedback. You can reach out to the [cdk.dev](http://cdk.dev/) community on [Slack](https://cdk.dev/) to solicit support for reprioritization.

aldex32:
following

yaojiach:
+1

yaojiach:
+1

filletofish:
Hey all, have started a RFC for this construct: https://github.com/aws/aws-cdk-rfcs/issues/474.

[...]

PR #25663, opened 2023-05-21
feat(scheduler): ScheduleTargetInput

Jacco:
This PR contains implementation of ScheduleTargetInput. While a schedule is the main resource in Amazon EventBridge Scheduler, this PR adds ScheduleTargetInput on which ScheduleTargetBase depends.

Every Schedule has a target that determines what extra information is sent to the target when the schedule is triggered. Also 4 ContextAttributes can be used that will be resolved at trigger-time.

To be able to create sensible unit tests, also the a start is made to add the `Schedule` and the `LambdaInvoke` target as described in the RFC.

Also added a small fix to 2 of the unit tests of the previous PR for this module.

----


packages/@aws-cdk/aws-scheduler-alpha/test/input.test.ts
```diff
@@ -0,0 +1,142 @@
...
+              { 'Ref': 'AWS::AccountId' },
+              '"}'
+            ]]
+          },
+        },
+      },
+    });
+  });
+
+  test('create an input with fromText with ContextrAttribute', () => {
```
filletofish, line 99:
```suggestion
  test('create an input with fromText with ContextAttribute', () => {
```

packages/@aws-cdk/aws-scheduler-alpha/test/schedule-expression.test.ts
```diff
@@ -94,7 +94,7 @@ describe('schedule expression', () => {
   });
 
   test('one-time expression string has expected date', () => {
-    const x = ScheduleExpression.at(new Date(2022, 10, 20, 19, 20, 23));
+    const x = ScheduleExpression.at(new Date(Date.UTC(2022, 10, 20, 19, 20, 23)));
```
filletofish, line 97: Thanks! Good catch!

packages/@aws-cdk/aws-scheduler-alpha/lib/input.ts
```diff
@@ -0,0 +1,120 @@
...
+export abstract class ScheduleTargetInput {
+  /**
+   * Pass text to the target, it is possible to embed `ContextAttributes` and other
+   * cdk references.
+   *
+   * The target input value will be a single string: the string you pass
+   * here. Do not use this method to pass a complex value like a JSON object to
+   * a target. Use `ScheduleTargetInput.fromObject()` instead.
+   * 
+   * @param text text to use as input for the target
```
filletofish, line 16: Should we separate visually param name and description? Could also add `-`?
```suggestion
   * @param text Text to use as the input for the target
```

filletofish (changes requested):

kaizencc (commented):
Minor comments, but I dont really have a problem with most of this

packages/@aws-cdk/aws-scheduler-alpha/test/schedule-expression.test.ts
filletofish: Yeah, we discussed it internally in Slack (pinged you again) and @Jacco suggestion was: 

```
public static at(year: int, monthIndex: int, day?: int, hours?: int, minutes?: int, seconds?: int, milliseconds?: int, timeZone?: TimeZone): ScheduleExpression {
   ...
}
```

Jacco:
About `at`

This is from the AutoScaling module:

```ts
  /**
   * Construct a Schedule from a moment in time
   */
  public static at(moment: Date): Schedule {
    return new LiteralSchedule(`at(${formatISO(moment)})`);
  }
```

It implicitly converts the date passed in to UTC with the local timeZone (local to cdk). Probably also not very intuitive.

How about the following:

```ts
  public static at(localYear: number, localMonth: number, localDay: number, localHour: number, localMinute: number, localSeconds: int, timeZone?: TimeZone): ScheduleExpression {
   ...
  }
```

packages/@aws-cdk/aws-scheduler-alpha/test/input.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    Template.fromStack(stack).hasResource('AWS::Scheduler::Schedule', {
+      Properties: {
+        Target: {
+          Input: '"Test=<aws.scheduler.schedule-arn>"',
+        },
+      },
+    });
+  });
+
+  test('create an input with fromObject with ContextrAttribute', () => {
```
kaizencc, line 116:
```suggestion
  test('create an input with fromObject with ContextAttribute', () => {
```

packages/@aws-cdk/aws-scheduler-alpha/test/schedule-expression.test.ts
kaizencc: I still think that `at` should just mirror what we have elsewhere. If we want to change how `at` works cdk-wide fine, but thats a different conversation

packages/@aws-cdk/aws-scheduler-alpha/lib/schedule.ts
```diff
@@ -0,0 +1,63 @@
...
+import { CfnSchedule } from 'aws-cdk-lib/aws-scheduler';
+import { Construct } from 'constructs';
+import { ScheduleExpression } from './schedule-expression';
+
+/**
+ * DISCLAIMER: WORK IN PROGRESS, INTERFACE MIGHT CHANGE
+ * 
+ * This unit is not yet finished. Only rudimentary Schedule is implemented in order
+ * to be able to create some sensible unit tests
+ */
```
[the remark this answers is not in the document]
Jacco: Agree that this is possible. I moved the other stuff except for the interface ISchedule to the private directory. I do not agree though that this was the best way forward since this all needs to be undone in the coming PRs. It seems an unnecessary hassle for an experimental module.

I wanted to introduce these incomplete-but-working classes (Schedule and Target) to also be able to add a working integration test in this PR with the implemented features. For now we did not test that the currently implemented features Expression and Input actually work through deployment with the generated CloudFormation.

Also the rosetta tests currently break because the examples reference Schedule.

kaizencc (dismissed):

kaizencc (approved):

kaizencc (approved):

Issue #8827, opened 2020-07-01
[aws-apigateway] impossible to remove default method authorization 


erik-sab:
It seems not possible to remove authorization for API Gateway methods if it is defined in `defaultMethodOptions` on `RestApi` level.

### Reproduction Steps

First I create RestApi Gateway (https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-apigateway.RestApi.html) with custom authorizer set by default for all methods:
```java
        var apiGw = RestApi.Builder.create(scope, "Stack-RestApi")
                .defaultMethodOptions(MethodOptions.builder()
                        .apiKeyRequired(Boolean.FALSE)
                        .authorizationType(AuthorizationType.CUSTOM)
                        .authorizer(authorizer)
                        .build())
...
```

And then in resources stack I try to create documentation Method (https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-apigateway.Method.html) with security disabled:
```java
        Method.Builder builder = Method.Builder.create(scope, "Stack-ApiInfoMethodGET")
                .options(MethodOptions.builder()
                        .apiKeyRequired(Boolean.FALSE)
                        .authorizationType(AuthorizationType.NONE)
                        .authorizer(null) // tried to reset authorizer also
                        .build())
...
```

### Error Log

```
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.6.0:java (default-cli) on project cdk-stacks: An exception occured while executing the Java class. Stack-ApiGwResource/Stack-ApiInfoResource/GET - Authorization type is set to NONE which is different from what is required by the authorizer [CUSTOM]
[ERROR] Error: Stack-ApiGwResource/Stack-ApiInfoResource/GET - Authorization type is set to NONE which is different from what is required by the authorizer [CUSTOM]
[ERROR]     at new Method (/tmp/jsii-kernel-TH8eSk/node_modules/@aws-cdk/aws-apigateway/lib/method.js:27:19)
[ERROR]     at /tmp/jsii-java-runtime16647868641746308035/jsii-runtime.js:7906:49
[ERROR]     at Kernel._wrapSandboxCode (/tmp/jsii-java-runtime16647868641746308035/jsii-runtime.js:8382:19)
[ERROR]     at Kernel._create (/tmp/jsii-java-runtime16647868641746308035/jsii-runtime.js:7906:26)
[...]

PR #25682, opened 2023-05-23
fix(apigateway): allow overriding apiKeyRequired on methods

iRoachie:
RestApi has the ability to set the apiKeyRequired option for all methods via defaultMethodOptions.

Setting this option on a method should override the value set in defaultMethodOptions, but it doesn't work.

This commit fixes the behaviour and adds a test.

Mentioned in #8827

----


packages/aws-cdk-lib/aws-apigateway/test/lambda-api.test.ts
```diff
@@ -299,7 +299,7 @@ describe('lambda api', () => {
       ResourceId: { Ref: 'lambdarestapiproxyE3AE07E3' },
       AuthorizationType: 'NONE',
       AuthorizerId: Match.absent(),
-      ApiKeyRequired: Match.absent(),
+      ApiKeyRequired: false,
```
iRoachie, line 302: This should be the correct value given that it's set [here](https://github.com/aws/aws-cdk/blob/5673b75c7ac1d4e122fd441c43cf42703047c6e0/packages/aws-cdk-lib/aws-apigateway/lib/resource.ts#L303)

vinayak-kukreja:
Hey, this looks good to me. Thank you for your contributing.

Getting a second opinion from @corymhall.

corymhall (approved):

corymhall (approved):

PR #25697, opened 2023-05-23
feat(ec2): add addSecurityGroup method to launth template

### LINKED ISSUE #18712: (aws-ec2): Launch template support for multiple security groups
### Description

Hi,

I noticed when migrating my EC2 launch template definition to CDK (Typescript project) that there is currently no option to provide multiple security groups for a launch template. When creating a launch template in the AWS Console this is possible to do.

I'm not sure if i am missing something here but it does not seem possible to add multiple security groups during creation or assigning additional security groups

### Use Case

It would be nice to be able to re-use existing security groups instead of having to make a new (duplicate) one which is just a bunch of existing security groups combined.

### Proposed Solution

[`LaunchTemplateProps` ](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.LaunchTemplateProps.html) currently only is able to process a single `securityGroup`.  It would be nice if this would accept `ISecurityGroup` or `ISecurityGroup[]`

Another option would to be able to add additional security groups to the [`LaunchTemplate`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.LaunchTemplate.html) after instantiating [just like with EC2 instances ](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.Instance.html) by using `addSecurityGroup(securityGroup)`

EDIT: Alternative is to use `CfnLaunchTemplate` since this does support multiple security groups.

### Other information

_No response_

### Acknowledge

- [x] This feature might incur a breaking change

kuredev:
[LaunchTemplateProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.LaunchTemplateProps.html) is able to process a single securityGroup.
Currently we are required to use connections when we wanted to use multiple security groups. https://github.com/aws/aws-cdk/issues/18712#issuecomment-1026975615
I implemented addSecurityGroup method to make this easier.

Closes #18712

----


packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts
```diff
@@ -786,6 +786,18 @@ export class LaunchTemplate extends Resource implements ILaunchTemplate, iam.IGr
...
   }
 
+  /**
+   * Add the security group to the instance.
+   *
+   * @param securityGroup: The security group to add
+   */
+  public addSecurityGroup(securityGroup: ISecurityGroup): void {
+    if (!this._connections) {
+      throw new Error('LaunchTemplate can only be added a securityGroup if another securityGroup is initialized in the constructor.');
```
otaviomacedo, line 796: Why?

kuredev: Thank you for your comment.
Because I understood that `_connections` is initialized only in the constructor.
https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts#L585

If `_connections` wasn't initialized in the constructor, an error occurs when trying to call `addSecurityGroup` here.

otaviomacedo (approved):

kuredev:
@Mergifyio requeue

otaviomacedo (approved):

PR #25699, opened 2023-05-23
chore(ecs): add security groups from the ASG's connections into the ECS Cluster's connection when using `addAsgCapacityProvider`

### LINKED ISSUE #22764: aws-ecs: Cluster.addAsgCapacityProvider doesn't add security groups to Cluster.connections like addAutoScalingGroup did

KollaAdithya:
Automatically add security groups from the ASG's connections into the ECS Cluster's connection when using `addAsgCapacityProvider`.
```
const myCapacityProvider = new ecs.AsgCapacityProvider(this, 'MyCapacityProvider', {
    autoScalingGroup: myAsg,
});
myCluster.addAsgCapacityProvider(myCapacityProvider);
```

Closes #22764 .

----


packages/aws-cdk-lib/aws-ecs/lib/cluster.ts
```diff
@@ -474,6 +474,9 @@ export class Cluster extends Resource implements ICluster {
   }
 
   private configureAutoScalingGroup(autoScalingGroup: autoscaling.AutoScalingGroup, options: AddAutoScalingGroupCapacityOptions = {}) {
+    if (autoScalingGroup.connections?.securityGroups) {
```
bvtujo, line 477: Is this always true? Will this condition always be triggered?

packages/aws-cdk-lib/aws-ecs/lib/cluster.ts
```diff
@@ -474,6 +474,9 @@ export class Cluster extends Resource implements ICluster {
...
+      this.connections.connections.addSecurityGroup(...autoScalingGroup.connections.securityGroups);
+    }
```
bvtujo, line 479: I ask because I just want to make sure that there is a tangible change to at least one integ test template generated by this line of code. If so, then we're all good and I'll approve; if not, we need to make sure we can figure out what change we expect to be applied to the clsuter security groups or autoscaling group connections.

KollaAdithya: These changes are tested over here my modifying this integration test and this will modify snapshot to add security groups from ASG connections to Cluster connections.
https://github.com/aws/aws-cdk/pull/25699/files#diff-05228637eeb8241609924d2fc7acce96e8aa522e1aafe36ca7d5a47c3b73ab16R53

bvtujo (approved):
LGTM

otaviomacedo (approved):

otaviomacedo (dismissed):

KollaAdithya:
Hey @otaviomacedo ! Thanks for the approval but why is the `mergify ` check failing.

PR #25708, opened 2023-05-23
fix(core): prevent the error when the condition is split into groups of 10 and 1 in `Fn.conditionOr()`

### LINKED ISSUE #25696: Fn.conditionOr generates invalid CFN snippet
### Describe the bug

I have a piece of code that uses `Fn.conditionOr` with a list of conditions. When the list is longer than 10 elements, CDK is smart enough to break it up into chunks of 10. Otherwise, CloudFormation would complain that an Fn::Or cannot have more than 10 elements. The problem I'm running into is for a list of 11 elements. CDK generates two Fn::Or expressions: One with 10 elements and one with 1 element. When deployment this stack, CloudFormation complains that an Fn::Or must contain at least 2 elements.
Error message from CFN: `Template error: every Fn::Or object requires a list of at least 2 and at most 10 boolean parameters.`

### Expected Behavior

Given a list of `conditions` of length 11 (let's say of type Fn::Equals) and this code snippet:
`Fn.conditionOr(...conditions)`, CDK should break up the list in a way that's valid according to CloudFormation's rules. For example into length 9 and 2:

```json
"Fn::Or": [
 {
  "Fn::Or": [
   {
    "Fn::Equals": [
     "1",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "2",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "3",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "4",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "5",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "6",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "7",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "8",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "9",
     "B"
    ]
   }
  ]
 },
 {
  "Fn::Or": [
   {
    "Fn::Equals": [
     "10",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "11",
     "B"
    ]
   }
  ]
 }
]

```

### Current Behavior

Given a list of `conditions` of length 11 (let's say of type Fn::Equals) and this code snippet:
`Fn.conditionOr(...conditions)`, CDK generates a template snippet like this:

```json
"Fn::Or": [
 {
  "Fn::Or": [
   {
    "Fn::Equals": [
     "1",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "2",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "3",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "4",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "5",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "6",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "7",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "8",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "9",
     "B"
    ]
   },
   {
    "Fn::Equals": [
     "10",
     "B"
    ]
   }
  ]
 },
 {
  "Fn::Or": [
   {
    "Fn::Equals": [
     "11",
     "B"
    ]
   }
  ]
 }
]

```

### Reproduction Steps

See example above

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.62.2

### Framework Version

_No response_

### Node.js Version

v16.20.0

### OS

Ubuntu 20

### Language

Typescript

### Language Version

4.9.4

### Other information

_No response_

wafuwafu13:
Closes #25696

>The problem I'm running into is for a list of 11 elements. CDK generates two Fn::Or expressions: One with 10 elements and one with 1 element. When deployment this stack, CloudFormation complains that an Fn::Or must contain at least 2 elements.

reproduce code: https://github.com/aws/aws-cdk/issues/25696#issuecomment-1560136915
approach: https://github.com/aws/aws-cdk/issues/25696#issuecomment-1559887661

----


wafuwafu13:
Exemption Request

I add test case in packages/aws-cdk-lib/core/test/condition.test.ts

packages/aws-cdk-lib/core/lib/cfn-fn.ts
```diff
@@ -334,6 +334,10 @@ export class Fn {
     if (conditions.length === 1) {
       return conditions[0] as ICfnRuleConditionExpression;
     }
+    // prevent the error "Fn::Or object requires a list of at least 2" when the condition is split into groups of 10 and 1
+    if (conditions.length % 10 === 1) {
+      return Fn.conditionOr(..._inGroupsOf(conditions, 9).map(group => new FnOr(...group)));
```
tmokmss, line 339: I think this can break when `conditions.length % 90 == 1` :)

packages/aws-cdk-lib/core/test/condition.test.ts
```diff
@@ -59,4 +59,54 @@ describe('condition', () => {
...
+                { 'Fn::Equals': ['f', '6'] },
+                { 'Fn::Equals': ['g', '7'] },
+                { 'Fn::Equals': ['h', '8'] },
+                { 'Fn::Equals': ['i', '9'] },
+              ],
+            },
+            {
+              'Fn::Or': [
+                { 'Fn::Equals': ['j', '10'] },
+                { 'Fn::Equals': ['k', '11'] },
```
MalikAtalla-AWS, line 104: This is totally a nice-to-have, but in terms of the simplicity of the resulting CFN snippet, I think it would be preferable to simply drop the last wrapping `Fn::Or`, like this

```
    'Fn::Or': [
      {
        'Fn::Or': [
          { 'Fn::Equals': ['a', '1'] },
          { 'Fn::Equals': ['b', '2'] },
          { 'Fn::Equals': ['c', '3'] },
          { 'Fn::Equals': ['d', '4'] },
          { 'Fn::Equals': ['e', '5'] },
          { 'Fn::Equals': ['f', '6'] },
          { 'Fn::Equals': ['g', '7'] },
          { 'Fn::Equals': ['h', '8'] },
          { 'Fn::Equals': ['i', '9'] },
          { 'Fn::Equals': ['j', '10'] }
        ],
      },
      {
          'Fn::Equals': ['k', '11']
      },
    ]
```

packages/aws-cdk-lib/core/lib/cfn-fn.ts
wafuwafu13: Yes, but would it be necessary to consider conditions for numbers larger than 91 or 181...?

packages/aws-cdk-lib/core/test/condition.test.ts
wafuwafu13: Yes, but I think the brevity of the patch is better now :)

packages/aws-cdk-lib/core/lib/cfn-fn.ts
wafuwafu13: I agree that if we can't be sure, I should respond.
But in this case, I don't have a simple solution to solve this 91 or 181 or more problem...
And I'd prefer not to add more complex patch and test(should check the case of 91 and doesn't use snapshot for condition now) for funky code of 91 or 181 or more conditions.

tmokmss: I see. I feel the solution @MalikAtalla-AWS proposed is also simple without any edge cases, but let's wait for comments from the maintainers.

corymhall (dismissed):

packages/aws-cdk-lib/core/test/condition.test.ts
wafuwafu13: I've found that this can be achieved with a simple code!

packages/aws-cdk-lib/core/lib/cfn-fn.ts
```diff
@@ -334,6 +334,10 @@ export class Fn {
     if (conditions.length === 1) {
       return conditions[0] as ICfnRuleConditionExpression;
     }
+    // prevent the error "Fn::Or object requires a list of at least 2" when the condition is split into groups of 10 and 1
```
otaviomacedo, line 337: Just a matter of preference, but this seems more readable:

```ts
public static conditionOr(...conditions: ICfnConditionExpression[]): ICfnRuleConditionExpression {
  if (conditions.length === 0) {
    throw new Error('Fn.conditionOr() needs at least one argument');
  }
  if (conditions.length === 1) {
    return conditions[0] as ICfnRuleConditionExpression;
  }
  if (conditions.length <= 10) {
    return new FnOr(...conditions);
  }
  return Fn.conditionOr(..._inGroupsOf(conditions, 10).map(group => Fn.conditionOr(...group)));
}
```

The three base cases are stated explicitly, falling back to the recursive case at the end.

otaviomacedo (dismissed):

PR #25717, opened 2023-05-24
chore(lambda-nodejs): stop bundling lambda provided sdk by default

### LINKED ISSUE #25492: aws_lambda_nodejs: Using Lambda Provided SDK by default in NodejsFunction leads to higher cold starts
### Describe the feature

The [BundlingOptions](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.BundlingOptions.html#externalmodules) in NodejsFunction construct removes AWS SDK dependencies by default.

This uses Lambda Provided SDK in the resulting function. This has higher cold start than a bundled function with AWS SDK dependencies included.

This happens, because the Node.js runtime has to do module resolution and go through multiple files while reading dependency code in the bundled function which uses Lambda Provided SDK. When SDK in bundled with the function code, the cold starts are lower as the as Node.js runtime has to read single file without any module resolution.

### Use Case

The default NodeJsFunction created by CDK Lambda should have low cold starts.

Lambda customers are usually sensitive to [cold starts](https://www.google.com/search?q=lambda+cold+starts). The bundling of the source code already reduces the function size, which is way below the [Lambda Quotas](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html) limit of 50 MB.

For repro, refer README of https://github.com/trivikr/aws-cdk-lambda-nodejs-function-cold-start-latency

Here are the benchmark numbers in ms:
```console
{
  'NodejsFunction default (uses Lambda Provided SDK)': 1227.1435,
  'NodejsFunction custom (uses Customer Deployed SDK)': 929.441
}
```

#### Notes:

* The benchmarks were run on count 1000 to ensure they’re reliable.
* As of 2023-05-09, it ensured that the SDK version is same in both cases, i.e. 3.188.0
* The numbers show median value for all cold start invocations. For other statistics, refer to output of measure-cold-start Step Function execution.

### Proposed Solution

The default NodeJsFunction created by CDK Lambda should not exclude AWS SDK dependencies.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.76.0

N/A

fynnfluegge:
This PR disables bundling of Lambda provided SDK by default which leads to lower cold starts since aws-sdk is removed from asset bundle as extensively explained and benchmarked in the related issue.

Breaking change. An opt im property is added to bundle the aws-sdk as it has been the default before.

Closes #25492.

----


corymhall (changes requested):
@fynnfluegge thanks for the PR! In general I like this change, I think we should be bundling everything by default. I'm worried about making this change the default without a feature flag though.

I tested the existing integration tests and the aws-sdk v2 test timed out until I increased the `memorySize` of the function. This is just one breaking point, but it makes me think that there could be more that we are not thinking of.

1. Lets put this behind a feature flag so that we don't change the behavior for existing users
2. Is there a minimum required memory setting for `aws-sdk`?

packages/@aws-cdk-testing/framework-integ/test/aws-lambda-nodejs/test/integ.dependencies.ts
```diff
@@ -37,6 +37,9 @@ class SdkV3TestStack extends Stack {
     this.lambdaFunction = new lambda.NodejsFunction(this, 'external-sdk-v3', {
       entry: path.join(__dirname, 'integ-handlers/dependencies-sdk-v3.ts'),
       runtime: Runtime.NODEJS_18_X,
+      bundling: {
```
corymhall, line 40: Can you remove this so that the integ test will test the new behavior?

Can you also add `diffAssets: true` to the `IntegTest`

corymhall (changes requested):
@fynnfluegge thanks for the PR! In general I like this change, I think we should be bundling everything by default. I'm worried about making this change the default without a feature flag though.

I tested the existing integration tests and the aws-sdk v2 test timed out until I increased the `memorySize` of the function. This is just one breaking point, but it makes me think that there could be more that we are not thinking of.

1. Lets put this behind a feature flag so that we don't change the behavior for existing users
2. Is there a minimum required memory setting for `aws-sdk`?

fynnfluegge:
> @fynnfluegge thanks for the PR! In general I like this change, I think we should be bundling everything by default. I'm worried about making this change the default without a feature flag though.
> 
> I tested the existing integration tests and the aws-sdk v2 test timed out until I increased the `memorySize` of the function. This is just one breaking point, but it makes me think that there could be more that we are not thinking of.
> 
> 1. Lets put this behind a feature flag so that we don't change the behavior for existing users
> 2. Is there a minimum required memory setting for `aws-sdk`?

Hey @corymhall thanks for reviewing! 
1. Feature flag sounds reasonable, will put the changes behind it ✅
2. Not sure 🤔 Will take a look at this

packages/@aws-cdk-testing/framework-integ/test/aws-lambda-nodejs/test/integ.dependencies.ts
corymhall: I think it may be because there is no lock file. I would try adding a new sub folder for this handler with a minimal `package.json` and `package-lock.json`.

fynnfluegge: That's exactly what I did 🤔

corymhall: Ok I think for this test we have to have the bundling actually perform an npm install.

```ts
    this.lambdaFunction = new lambda.NodejsFunction(this, 'external-sdk-v3', {
      entry: path.join(__dirname, 'integ-handlers/v3/dependencies-sdk-v3.ts'),
      runtime: Runtime.NODEJS_18_X,
      depsLockFilePath: path.join(__dirname, 'integ-handlers/v3/package-lock.json'),
      bundling: {
        bundleLambdaProvidedAwsSdk: true,
        commandHooks: {
          beforeInstall() {
            return [];
          },
          afterBundling() {
            return [];
          },
          beforeBundling(inputDir, _outputDir) {
            return [`cd ${inputDir}`, 'npm install --ci'];
          },
        },
      },
    });
```

fynnfluegge: Okay that seems to work ✅
So I want to create the new snapshots now but that leads me into a rabbit hole 😄 
When running `yarn integ --update-on-failed aws-lambda-nodejs/test/integ.dependencies.js` it says for the test with the bundled sdk:
Screenshot 2023-06-09 at 16 14 51

It is the `Response object is too long` error which relates to several open issues. Any idea how to do a workaround for this?

PR #25725, opened 2023-05-24
feat(lambda): provide support for AWS Parameters and Secrets Extension for Lambda

### LINKED ISSUE #23187: (lambda): Make it easy to use "AWS Parameters and Secrets Lambda Extension" with Lambda Functions
### Describe the feature

The [AWS Parameters and Secrets Lambda Extension](https://docs.aws.amazon.com/secretsmanager/latest/userguide/retrieving-secrets_lambda.html) is a great way to securely fetch Secrets Manager Secrets from Lambda functions. Today, I have to reference the documentation and hard-code the Layer ARN from [this table](https://docs.aws.amazon.com/secretsmanager/latest/userguide/retrieving-secrets_lambda.html#retrieving-secrets_lambda_ARNs) to add the layer to my function.

It would be nice if there were a convenience method that automatically determined the proper ARN for the Lambda's architecture and deployment region. It would also be great if the available [configuration environment variables](https://docs.aws.amazon.com/secretsmanager/latest/userguide/retrieving-secrets_lambda.html#retrieving-secrets_lambda_env-var) were able to be set via CDK.

### Use Case

A common pattern for exposing Secrets Manager secrets to Lambda functions is by setting them as environment variables. However, this is insecure because the variable is set in plain text in the console. Additionally, updates to the secret outside of CloudFormation do not update the environment variable. Recent versions of CDK warn you about this, requiring an explicit call to `unsafeUnwrap` on the Secret.

One way to work around this issue is to call the Secrets Manager API (e.g. via `aws-sdk`) in your Lambda function. However, this can get expensive because it results in a lot of retrievals of secret values. You then have to implement your own caching behavior if you want to reduce cost.

The AWS Parameters and Secrets Lambda Extension solves both of these problems. The secret is fetched at runtime and it handles caching for you automatically.

### Proposed Solution

```ts
const myFunc = new lambda.Function(this, 'Function', {
  // existing parameters
});
myFunc.attachParametersAndSecretsExtension();
```

Then, the logic for the function would look something like (pseudocode):

```ts
interface ParametersAndSecretsExtensionConfig {
  parametersSecretsExtensionCacheEnabled?: boolean
  parametersSecretsExtensionCacheSize?: number
  // and the rest of the config properties from: https://docs.aws.amazon.com/secretsmanager/latest/userguide/retrieving-secrets_lambda.html#retrieving-secrets_lambda_env-var
}

class Function {
  // existing logic

  attachParametersAndSecretsExtension(config: ParametersAndSecretsExtensionConfig = {}) {
    this.addLayers(this.getParametersAndSecretsExtensionLayerArn);

    for (const configKey in config) {
      this.addEnvironment(constantCase(configKey), config[configKey].toString());
    }
  }

  private getParametersAndSecretsExtensionLayerArn(): ILayerVersion {
    // logic similar to https://github.com/aws/aws-cdk/blob/f2855dc12b1ff2c1a24f6fbd9de20d9b7079127a/packages/@aws-cdk/region-info/build-tools/fact-tables.ts#L244
    // it returns the proper layer for the region and architecture of the lambda
    const layerArn = lookupArnFromFactTable(this.architecture);

    return LayerVersion.fromLayerArn(layerArn)
  }
}
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.53.0

MacOS, TypeScript

colifran:
This PR provides support for the AWS Parameters and Secrets Extension for Lambda functions. This extension will allow users to retrieve and cache AWS Secrets Manager secrets and AWS Parameter Store parameters in Lambda functions without using an SDK.

Note: Design was updated from previous PR which had a circular dependency causing go build to break.

Closes #23187 

----


packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -1149,6 +1160,33 @@ Environment variables can be marked for removal when used in Lambda@Edge by sett
...
+  private configureParamsAndSecretsExtension(props: FunctionProps): void {
+    if (props.paramsAndSecrets === undefined) {
+      return;
+    }
+
+    // grant permissions to lambda execution role to allow access to provided secrets
+    props.paramsAndSecrets.secrets?.forEach(secret => {
+      secret.grantRead(this);
+      if (secret.encryptionKey) {
+        secret.encryptionKey.grantDecrypt(this);
```
rix0rrr, line 1175: Feels like `secret.grantRead()` should grant this automatically. Does it not?

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -1149,6 +1160,33 @@ Environment variables can be marked for removal when used in Lambda@Edge by sett
...
+      }
+    });
+
+    // grant permission to lambda execution role to allow access to provided parameters
+    props.paramsAndSecrets.parameters?.forEach(param => {
+      param.grantRead(this);
+      // TODO: Figure out how to determine if param is secure string for kms:Decrypt
```
rix0rrr, line 1182: Again, if the param can know itself whether it's a secure string or not, then `grantRead()` should be able to grant that permission automatically.

packages/aws-cdk-lib/aws-lambda/lib/params-and-secrets-layers.ts
```diff
@@ -0,0 +1,269 @@
...
+   * @default - Logging level will be `info`
+   */
+  readonly logLevel?: ParamsAndSecretsLogLevel;
+
+  /**
+   * The maximum number of connection for HTTP clients that the Parameters and
+   * Secrets Extension uses to make requests to Parameter Store or Secrets
+   * Manager. There is no maximum limit. Minimum is 1.
+   *
+   * Note: This is a per-client configuration.
```
rix0rrr, line 106: I'm not sure people will understand what this means.

Maybe say something like "Every running copy of this Lambda Function may open this many connections, so the total number of connections may exceed this number".

packages/aws-cdk-lib/aws-lambda/lib/params-and-secrets-layers.ts
```diff
@@ -0,0 +1,269 @@
...
+
+    if (this.options.secretsManagerTtl !== undefined && this.options.secretsManagerTtl.toSeconds() > 300) {
+      throw new Error(`Maximum TTL for a cached secret is 300 seconds - provided: ${this.options.secretsManagerTtl.toSeconds()} seconds`);
+    }
+
+    if (this.options.parameterStoreTtl !== undefined && this.options.parameterStoreTtl.toSeconds() > 300) {
+      throw new Error(`Maximum TTL for a cached parameter is 300 seconds - provided: ${this.options.parameterStoreTtl.toSeconds()} seconds`);
+    }
+
+    return {
```
rix0rrr, line 236: This is lovely! <3 Exactly what CDK is for

rix0rrr (commented):
Looks good!

I have one slight concern: I wonder if this makes more sense as an external construct, rather than a built-in option on Lambda. So for example: 

```ts
new ParametersAndSecretsExtension(this, 'Extension', {
  function: fn,
  // ...options here...
});
```

It's not as discoverable, which is a pity, but it is a pattern that's easier to replicate for other future layers, and also 3rd party layers.

packages/aws-cdk-lib/aws-lambda/lib/params-and-secrets-layers.ts
colifran: Good suggestion. I'll make this change.

packages/aws-cdk-lib/aws-lambda/lib/function.ts
colifran: Great catch. Just checked the code and it does grant decrypt as part of grantRead when an encryptionKey is specified.

packages/aws-cdk-lib/aws-lambda/lib/function.ts
colifran: Same as my above comment. Thanks for pointing this out and grantRead here also grantsDecrypt when an encryptionKey is provided.

colifran:
@rix0rrr I can't directly respond to your first comment for some reason. That said, I made the initial design choice for several reasons:

1. We use a similar design pattern for lambda insights and ADOT lambda extensions. My thinking was that maintaining a consistent user experience when we're adding support for different lambda extensions would be important.
2. To me, it makes sense to add a lambda extension to a lambda function and then use _bind internally to bind the extension to the lambda. That said, I wonder if passing a function to a lambda extension would convey that the function is part of the extension vs. the extension being part of the lambda function.

I think you bring up a very valid point in considering making this an external construct rather than a built-in option. Do the two points I made above change your mind at all?

otaviomacedo (approved):

Issue #25732, opened 2023-05-25
IAM: addAccountCondition adds condition which is not AWS:SourceAccount


mdavis-xyz:
### Describe the issue

The description for [addAccountCondition](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-iam.PolicyStatement.html#addwbraccountwbrconditionaccountid) says:

> Add a condition that limits to a given account.

Based on this, I assumed that it would add 

```
      "Condition": {
        "StringEquals": {
          "AWS:SourceAccount": "1234"
        }
      }
```

But it does not. It adds:

```
      "Condition": {
        "StringEquals": {
          "sts:ExternalId": "1234"
        }
      }
```

These are for very different use cases.

In general, any account can set any value for `"sts:ExternalId"`. An attacker who knows my account number could just set my account number as the ExternalId when assuming their own role, which would satisfy this constraint. So this condition doesn't actually relate to an account, just a passphrase.

My use case is that I'm granting S3 permission to publish to my SNS topic, with a topic policy. But I don't want other accounts to publish to my topic. I want to restrict it with an *account condition*.

When AWS assumes a role (e.g. S3 publishing notifications to my SNS topic), it does *not* set the assume role passphrase to the account number. So this condition actually ends up blocking legitimate traffic from my account.

I feel like this function is misnamed.

I'm guessing that changing the behavior of this function from "sts:ExternalId" to "AWS:SourceAccount" is a backwards-incompatible change that will mess with other users.

So what I propose is that the documentation should just say that this doesn't actually add a source account condition. And it should say how you should add a source account condition.

Something like:

> Add a `"sts:ExternalId"` condition that the caller used an external ID when assuming their role which is equal to the specified account number, to prevent a confused deputy attack.
> Note that this is different to a condition limiting the source account. To do that, either use the account's principal with `addAwsAccountPrincipal`,  or use `statement.addCondition("StringEquals", { "AWS:SourceAccount": Aws.ACCOUNT_ID }) `.

(I also think it would be nice to make the argument to `addAccountCondition` and `addAwsAccountPrincipal` optional, defaulting to the current account. And to add another function `addSourceAccountCondition`, which uses `"AWS:SourceAccount"`)

### Links

https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-iam.PolicyStatement.html#addwbraccountwbrconditionaccountid

pahud:
In your use case:

[...]

PR #25744, opened 2023-05-25
chore(iam): support addSourceAccountCondition and addSourceArnCondition

pahud:
The [addAccountCondition](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_iam.PolicyStatement.html#addwbraccountwbrconditionaccountid) method essentially create a `StringEquals` condition with `sts:ExternalId` which is used for [Cross-account confused deputy prevention](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html#mitigate-confused-deputy). This PR adds `addSourceArnCondition` and `addSourceAccountCondition` methods used for [Cross-service confused deputy prevention](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html#cross-service-confused-deputy-prevention) and improves the doc on the methods.

Closes https://github.com/aws/aws-cdk/issues/25732

----


mdavis-xyz:
In the docs for the new `addSourceArnCondition` function, maybe mention that the argument can contain wildcards? i.e. it doesn't have to be an ARN, it can be a pattern that ARNs are matched against.

otaviomacedo (approved):

PR #25749, opened 2023-05-26
fix(ssm): cannot import a ssm parameter with a name containing unresolved token

### LINKED ISSUE #17094: (ssm): can not import parameter name from another stack
### What is the problem?

cdk generates input parameters in a stack when importing parameter store values.

### Reproduction Steps

```ts
    configSqs = (scope: cdk.Construct) =>
        ssm.StringParameter.fromStringParameterAttributes(
            scope,
            'import-sqs-string-param',
            {
                simpleName: true,
                parameterName: cdk.Fn.importValue('some-exported-value-holding-the-param-name'),
            }
        );

        const sqsConfig = configSqs(this);
        sqsConfig.grantRead(this.ingestLambda);

```

This renders perfectly in terms of IAM:

```json
            {
              "Action": [
                "ssm:DescribeParameters",
                "ssm:GetParameters",
                "ssm:GetParameter",
                "ssm:GetParameterHistory"
              ],
              "Effect": "Allow",
              "Resource": {
                "Fn::Join": [
                  "",
                  [
                    "arn:",
                    {
                      "Ref": "AWS::Partition"
                    },
                    ":ssm:eu-west-1:123456789012:parameter/",
                    {
                      "Fn::ImportValue": "some-exported-value-holding-the-param-name"
                    }
                  ]
                ]
              }
            }
```

But i also generates a parameter construct in the stack (I have no idea why):

```json
  "Parameters": {
    "importsqsstringparamParameter": {
      "Type": "AWS::SSM::Parameter::Value<String>",
      "Default": {
        "Fn::ImportValue": "some-exported-value-holding-the-param-name"
      }
    },
```

CDK not to create input parameters on the CF-stack, Why does it create an input parameter to the cloudformation stack?

PS

I know you can explicity reference the parameter cross stack, but then you get into the automagically creation of exports problem

### What actually happened?

CDK creates input parameters on the CF-stack which breaks when deploying with: "Template format error: Every Default member must be a string".

### CDK CLI Version

1.124.0

### Framework Version

_No response_

### Node.js Version

12

### OS

ubuntu

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

tmokmss:
Previously, when we import a SSM parameter by `ssm.StringParameter.fromStringParameterAttributes`, we use `CfnParameter` to get the value. 

```json
  "Parameters": {
    "importsqsstringparamParameter": {
      "Type": "AWS::SSM::Parameter::Value<String>",
      "Default": {
        "Fn::ImportValue": "some-exported-value-holding-the-param-name"
      }
    },
```

However, `Parameters.<Name>.Default` only allows a concrete string value. If it contains e.g. intrinsic functions, we get an error like this from CFn: `Template format error: Every Default member must be a string.`

This PR changes the behavior of `fromStringParameterAttributes` method. Now it uses `CfnDynamicReference` instead of `CfnParameter` if a parameter name contains unresolved tokens.

Since previously the case when `Token.isUnresolved(attrs.parameterName) == true` just resulted in a deployment error, this is not a breaking change.

Closes #17094

----


packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
```diff
@@ -473,9 +473,15 @@ export class StringParameter extends ParameterBase implements IStringParameter {
 
     const type = attrs.type ?? attrs.valueType ?? ParameterValueType.STRING;
 
-    const stringValue = attrs.version
-      ? new CfnDynamicReference(CfnDynamicReferenceService.SSM, `${attrs.parameterName}:${Tokenization.stringifyNumber(attrs.version)}`).toString()
-      : new CfnParameter(scope, `${id}.Parameter`, { type: `AWS::SSM::Parameter::Value<${type}>`, default: attrs.parameterName }).valueAsString;
+    let stringValue: string;
+    if (attrs.version) {
+      stringValue = new CfnDynamicReference(CfnDynamicReferenceService.SSM, `${attrs.parameterName}:${Tokenization.stringifyNumber(attrs.version)}`).toString();
+    } else if (Token.isUnresolved(attrs.parameterName)) {
```
corymhall, line 479: I think there are cases where this could be a breaking change. There could be cases where `Token.isUnresolved` returns `true`, but the value in the template is still a resolved string.

For example

```ts
const parameterName = Lazy.string({ produce: () => 'myParameterName' });
```

`Token.isUnresolved(parameterName)` will return true, but the value that will be output will be 

```json
{
  "Type": "AWS::SSM::Parameter::Value<STRING>",
  "Default": "myParameterName",
}
```

We might need to use `Tokenization.isResolvable` instead.

packages/aws-cdk-lib/aws-ssm/test/parameter.test.ts
```diff
@@ -963,3 +963,40 @@ test('fails if parameterName is undefined and simpleName is "false"', () => {
   // THEN
   expect(() => new ssm.StringParameter(stack, 'p', { simpleName: false, stringValue: 'foo' })).toThrow(/If "parameterName" is not explicitly defined, "simpleName" must be "true" or undefined since auto-generated parameter names always have simple names/);
 });
+
```
corymhall, line 966: Can you add a test here that uses a lazy value? A good example would be a cross env bucketname.

```ts
const bucket = new s3.Bucket(this, 'Bucket', {
  bucketName: cdk.PhysicalName.GENERATE_IF_NEEDED,
});

const stack = new cdk.Stack(scope, 'Cross', {
  env: { region: 'us-east-1', account: process.env.CDK_DEFAULT_ACCOUNT },
});
ssm.StringParameter.fromStringParameterAttributes(stack, 'param', {
  parameterName: bucket.bucketName,
  simpleName: true,
});
```

corymhall (dismissed):

packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
tmokmss: Hi @corymhall thanks! I confirmed you are right (I added a test case and it failed.)

Now I tried `Tokenization.isResolvable` but for some reason it was not evaluated as `true` for a cross stack reference. I'll dive deep the reason later.

Perhaps we need to resolve the token first, and check if it contains CFn intrinsics, as you've done on `cfn-resources.ts`.

packages/aws-cdk-lib/aws-ssm/test/parameter.test.ts
tmokmss: Note: I used iam.Role instead as i don't know if it is allowed to add a new dependency on other modules.

packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
tmokmss: `Tokenization.isResolvable(attrs.parameterName)` won't work as expected because `typeof(attrs.parameterName)` is always `'string'`, which fails this evaluation.

https://github.com/aws/aws-cdk/blob/1d7a9a80b08d41ce8759bed9286adaa8259c2bc8/packages/aws-cdk-lib/core/lib/token.ts#L319-L321


So to determine whether a parameter name can be resolved to a concrete string, we can actually resolve it. Like this:

```ts
if (typeof Stack.of(scope).resolve(attrs.parameterName) != 'string') 
```

However, if we resolve the token here, it means a lazy string isn't actually evaluated lazily enough. Instead it's evaluated immediately to determine the condition. I'm not sure if it will break something though. A similar approach is also used in s3-deployment module, so maybe it's just fine.

https://github.com/aws/aws-cdk/blob/1d7a9a80b08d41ce8759bed9286adaa8259c2bc8/packages/aws-cdk-lib/aws-s3-deployment/lib/render-data.ts#L18-L19

Alternatively, we can create a method like `isLazyString` to determine if a string is not a lazy value without resolving it.  But then we cannot cover the case when `Lazy.string` produces cfn intrinsics like this:

```ts
Lazy.string({ produce: () => someResource.ref });
```

Another consideration is if we should really care this edge case. I think almost all of the devs don't use a Lazy string as a ssm parameter name to import it. At least we can fix it if someone would report an issue.

Hope you can suggest which way we should go, thanks! @corymhall

corymhall: It seems like what we are really trying to do is determine if the value is an intrinsic or not. In that case I think it might be best to just add a  `isIntrinsic` method to [Intrinsic](https://github.com/aws/aws-cdk/blob/53d24f1a1357410746052bbf25b5b89b985f7b80/packages/aws-cdk-lib/core/lib/private/intrinsic.ts#L36-L36)

tmokmss: Thanks for the insight! I agree with that the only case we have to handle is when parameterName contains intrinsics.

But `attrs.parameterName` is just a string (like `${Token[TOKEN.274]}`) before resolution. Is it possible to determine if the string is actually an instrinsic without resolving it?

tmokmss: It seems we can get backing tokens from a string by `TokenMap.instance().splitString(string).tokens` without actually resolving it. On second thought, however, if we also want to handle the case `Lazy.string({ produce: () => someResource.ref })`, we have to resolve the token anyway to determine if it's an intrinsic. But the disadvantage is that it has a side effect of resolving a lazy token in an unexpected timing.

@corymhall Do you think we really have to handle the lazy token edge case? I don't think it is a critical problem and is worth complicating the implementation, and would rather go with the original patch. To avoid a breaking change, maybe we can just add a feature flag.

corymhall: @tmokmss yeah there is no way to handle the `Lazy` case so lets let that fallback to the old behavior. Lets take an approach where we explicitly cover certain scenarios and everything else just falls back to the existing functionality. The one I can think of is an `isIntrinsic`

```ts
Token.isUnresolved(attrs.parameterName) && Intrinsic.isInstrinsic(Tokenization.reverseString(attrs.parameterName).firstToken)

```

tmokmss: @corymhall Thanks. I tried it and found that it won't work as expected :(
Because in the test case below, the parameterName is also a `Intrinsic` (to be precise, it's a `Reference` that extends `Intrinsic`), failing the test.

https://github.com/aws/aws-cdk/blob/54bcad2b172946e7879e0296b19c656a4716eca7/packages/aws-cdk-lib/aws-ssm/test/parameter.test.ts#L1045-L1055

So I guess what we need is a way to
1. determine if the token contains a CFn intrinsic
2. without resolving the token, to avoid side effects to Lazy tokens

I'll try to find how to achieve this, but let me know if you have a good idea.

corymhall: What if we just don't handle the `Reference` case and only handle where `isIntrinsic() && !isReference()`?

tmokmss: @corymhall Thanks. I confirmed it won't work either. The `parameterName` is actually also a `Reference` in other test cases like `When a parameter name contains tokens, use dynamic reference instead` too. I guess it's because of this function:

https://github.com/aws/aws-cdk/blob/c135656bb0b6de9cce639218a83acf958f9bca4e/packages/aws-cdk-lib/core/lib/resource.ts#L254

To get the underlying token, it seems we must first resolve the reference, because the intrinsic is lazily produced. 

I'm now wondering whether switching a [parameter](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#aws-ssm-parameter-types) to a [dynamic reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html) should really be considered as a breaking change. As far as I read the docs, there seems to be no remarkable difference between them. Given it's also very rare to use a lazy token for parameter names, we can tolerate the change, maybe under a feature flag.

corymhall: `When a parameter name contains tokens, use dynamic reference instead` that case is what I am saying that we should not cover. It should just fall back to the old behavior and will not work in some cases (like before). Something like this should work though

```ts
ssm.StringParameter.fromStringParameterAttributes(this, 'param', {
  parameterName: Fn.getAtt(), // or Fn.conditionOr(), etc.
})
```

corymhall: There are some limitations https://github.com/aws/aws-cdk/pull/22239#issuecomment-1262069499

tmokmss: @corymhall Thanks, I didn't know that! 

So how about letting users choose which they use, parameter or dynamic reference? We'll add a property like `forceDynamicReference?: boolean` (default to false) to `CommonStringParameterAttributes`. This is kind of a leaky abstraction, but it should at least solve all the problem above.

corymhall: @tmokmss lets do both. Lets cover what we can (my comment above) and then allow the user to force the dynamic reference. Neither should need a feature flag.

tmokmss: @corymhall Now dynamic reference is used only when `cdk.Fn` is explicitly used. It can address the original use case reported on the issue. 

Other parameter name containing a token, we cannot decide if it contains CFn intrinsics or not without resolving the token, which is not ideal for lazy tokens. In that case, we just use CfnParameter, and letting a user to force dynamic reference by passing a flag.

I made the change to only `fromStringParameterAttributes`, not `fromListParameterAttributes`. I think we can do this in another PR.

corymhall (approved):

corymhall (approved):

PR #25775, opened 2023-05-29
fix(neptune): change removalPolicy of instance created within neptune cluster cdk

gopuneet:
change removalPolicy of instance created within neptune cluster cdk to DESTROY instead of inheriting from cluster

This fixes 2 issues faced by users of Neptune CDK:
1. The instance not being deleted when decreasing the count specified in `instances?` prop of [DatabaseCluster](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-neptune-alpha.DatabaseCluster.html) construct since it was inheriting the default RemovalPolicy of cluster resulting in instance also using `RemovalPolicy.RETAIN`.
2. Being unable to specify the `RemovalPolicy.SNAPSHOT` for the neptune DBCluster even though it is supported ([aws docs reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html)) because the policy was being inherited by the instance resulting in the error `Resource type AWS::Neptune::DBInstance does not support Snapshot DeletionPolicy`.

This aligns the behaviour of instances created via the `instances?` prop to match the default behaviour of [AWS::Neptune::DBInstance](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-neptune-dbinstance.html) by setting the default DeletionPolicy / RemovalPolicy to Delete / DESTROY.

----

### DIFF
--- a/packages/@aws-cdk/aws-neptune-alpha/lib/cluster.ts
+++ b/packages/@aws-cdk/aws-neptune-alpha/lib/cluster.ts
@@ -645,7 +645,7 @@ export class DatabaseCluster extends DatabaseClusterBase implements IDatabaseClu
       instance.node.addDependency(internetConnectivityEstablished);
 
       // Neptune instance's default deletion policy should be DESTROY to align with the CFN behaviour and
-      // to allow customers to set cluster's RemovalPolicy to SNAPSHOT since instance doesn't support SNAPSHOT.
+      // to allow users to set cluster's RemovalPolicy to SNAPSHOT since instance doesn't support SNAPSHOT.
       instance.applyRemovalPolicy(RemovalPolicy.DESTROY, {
         applyToUpdateReplacePolicy: true,
       });


packages/@aws-cdk/aws-neptune-alpha/lib/cluster.ts
```diff
@@ -644,7 +644,9 @@ export class DatabaseCluster extends DatabaseClusterBase implements IDatabaseClu
       // things in the right order.
       instance.node.addDependency(internetConnectivityEstablished);
 
-      instance.applyRemovalPolicy(props.removalPolicy, {
+      // Neptune instance's default deletion policy should be DESTROY to align with the CFN behaviour and
+      // to allow customers to set cluster's RemovalPolicy to SNAPSHOT since instance doesn't support SNAPSHOT.
+      instance.applyRemovalPolicy(RemovalPolicy.DESTROY, {
```
phanindhra876, line 649: Won't this set removal policy to DESTROY irrespective of what is set by the user?

packages/@aws-cdk/aws-neptune-alpha/test/cluster.test.ts
```diff
@@ -33,8 +33,8 @@ describe('DatabaseCluster', () => {
     });
 
     Template.fromStack(stack).hasResource('AWS::Neptune::DBInstance', {
-      DeletionPolicy: 'Retain',
-      UpdateReplacePolicy: 'Retain',
+      DeletionPolicy: 'Delete',
+      UpdateReplacePolicy: 'Delete',
     });
```
phanindhra876, line 38: Add a testcase where defaults are picked when Deletion and UpdateReplace policy are not set and vice-versa.

packages/@aws-cdk/aws-neptune-alpha/lib/cluster.ts
gopuneet: Yes, that is the intention.

This is the construct for cluster and users trying to set SNAPSHOT deletion policy for cluster are not able to because it gets inherited by the instance(s) created via cluster prop at this line and results in the error `Resource type AWS::Neptune::DBInstance does not support Snapshot DeletionPolicy` thus preventing users from using that deletion policy.

The instance(s) auto-created via the cluster prop `instances?` should use the DESTROY removal policy as it is the default behaviour in CFN.
If users want to customize the instance removal policy (instead of using default), they are expected to use the [DatabaseInstance](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-neptune-alpha.DatabaseInstance.html) construct.

packages/@aws-cdk/aws-neptune-alpha/test/cluster.test.ts
gopuneet: As explained in https://github.com/aws/aws-cdk/pull/25775#discussion_r1209796704, the DeletionPolicy (and UpdateReplacePolicy) for the instance(s) auto-created via the cluster prop `instances?` are expected to always use `Delete` as default. The policy specified by the user for the cluster construct should only be applied to the cluster.

phanindhra876 (approved):

PR #25778, opened 2023-05-30
fix(aws-cdk-lib): attribute `FindingsFilterListItems` on AWS::Macie::FindingsFilter does not work

mrgrain:
The attribute was removed in a recent CFN Spec update.
We had then patched the attribute and type back in to avoid a breaking change. 
Now I have confirmed with the service team that this attribute has actually never worked and the removal is indeed correct.
Update CDK to reflect the fix made in CloudFormation.

----


Naumel (approved):

mrgrain:
> mrgrain/fix-remove-macie-patch

@macrombilux Yes?

PR #25782, opened 2023-05-30
feat(lambda-event-sources): S3EventSource requires Bucket instead of IBucket

Styerp:
Cloning https://github.com/aws/aws-cdk/pull/25368
Closes https://github.com/aws/aws-cdk/issues/23940
Closes https://github.com/aws/aws-cdk/issues/4323

----


Styerp:
Reposting https://github.com/aws/aws-cdk/pull/25368#issuecomment-1559440120

[API surface changes](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#api-surface-changes) says changes would cause builds to fail -- that will not happen here. Underlying objects didn't access `Bucket` methods on the object  passed into this function, either directly or indirectly. All properties and objects being used were already available on the IBucket, which was the driving rationale to ease this input type.

While this is ostensibly an API Surface Change, it is NOT a _breaking_ surface change.

[Behavior changes](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#behavior-changes) say "even though the user didn't change their code, the CloudFormation template that gets synthesized is now different." The integration tests prove this is not the case -- the CFN generated is identical.

Build will continue to fail because of the "Breaking API Changes". Guidance on how y'all handle this would be great. Clarification Request

otaviomacedo:
> @rix0rrr Any idea why `jsii-diff` would be failing here:
> 
> ```
> API elements with incompatible changes:
> err  - PROP aws-cdk-lib.aws_lambda_event_sources.S3EventSource.bucket: type aws-cdk-lib.aws_s3.IBucket (formerly aws-cdk-lib.aws_s3.Bucket): aws-cdk-lib.aws_s3.IBucket does not extend aws-cdk-lib.aws_s3.Bucket [changed-type:aws-cdk-lib.aws_lambda_event_sources.S3EventSource.bucket]
> ```
> 
> This seems like an okay change to me.

The problem here is that `bucket` is not only a constructor argument, but also a public property. With this change, the following would break:

```ts
const bucket = new Bucket(...);
const source = new S3EventSource(bucket, ...);

source.bucket.addCorsRule(...); // <-- Consumers expect this to be a Bucket
```

I'm not sure if there's a valid use case for accessing the `bucket` property like this (it was probably made public by accident). But technically, it's a breaking change.

Issue #24822, opened 2023-03-28
Auto delete images upon ECR repo removal - CDK v2.70.0 not working


missourian55:
### Describe the bug

Tested this much awaited feature in CDKv2.70.0 and it is still not working as expected.

Error
```
Resource handler returned message: "The repository with name 'test-app' in registry with id '1234567890' 
cannot be deleted because it still contains images (Service: Ecr, Status Code: 400,
Request ID: 842c370d-f12344XXXXX-ed47d07f0443)" 
(RequestToken: c64e0c80-9xxxxxxca5006c6c, HandlerErrorCode: GeneralServiceException)
```

ecr: add option to auto delete images upon ECR repository removal (https://github.com/aws/aws-cdk/issues/24572) ([7de5b00](https://github.com/aws/aws-cdk/commit/7de5b00dcf24c4f6721317860c7e42c485e3ca58)), closes https://github.com/aws/aws-cdk/issues/15932 https://github.com/aws/aws-cdk/issues/12618 https://github.com/aws/aws-cdk/issues/15932

### Expected Behavior

Delete of ECR repo with images should succeed if removalPolicy "Destroy" and autoDelete "true"

```
const bucket = new ecr.Repository(this, 'Repo', {
  repositoryName: 'delete-even-if-contains-images',
  removalPolicy: cdk.RemovalPolicy.DESTROY,
  autoDeleteImages: true,
});
```

### Current Behavior

Error
```
Resource handler returned message: "The repository with name 'test-app' in registry with id '1234567890' 
cannot be deleted because it still contains images (Service: Ecr, Status Code: 400,
Request ID: 842c370d-f12344XXXXX-ed47d07f0443)" 
(RequestToken: c64e0c80-9xxxxxxca5006c6c, HandlerErrorCode: GeneralServiceException)
```

### Reproduction Steps

Create a ECR repo similar to below using `cdk deploy` and add images to it. Afterwards, execute `cdk destroy` which fails with exception mentioned above

```
const bucket = new ecr.Repository(this, 'Repo', {
  repositoryName: 'delete-even-if-contains-images',
  removalPolicy: cdk.RemovalPolicy.DESTROY,
  autoDeleteImages: true,
});
```

### CDK CLI Version

CDK v2.70.0 

### Node.js Version

v19.8.0

### OS

mac-os M1

### Language

Java

### Language Version

JDK 17

pahud:
[...]

PR #25789, opened 2023-05-31
fix(ecr): auto delete images on ECR repository containing manifest list

hkford:
Fixes https://github.com/aws/aws-cdk/issues/24822

As I commented on https://github.com/aws/aws-cdk/issues/24822#issuecomment-1531165305, auto delete container images in ECR repository fails when it has container manifest list. I fix custom resource Lambda function to delete tagged images first.

----

### DIFF
--- a/packages/cdk-assets/lib/private/archive.ts
+++ b/packages/cdk-assets/lib/private/archive.ts
@@ -58,17 +58,18 @@ function writeZipFile(directory: string, outputFile: string): Promise<void> {
 }
 
 /**
- * Rename the file to the target location, taking into account that we may see EPERM on Windows
- * while an Antivirus scanner still has the file open, so retry a couple of times.
+ * Rename the file to the target location, taking into account:
+ *
+ * - That we may see EPERM on Windows while an Antivirus scanner still has the
+ *   file open, so retry a couple of times.
+ * - This same function may be called in parallel and be interrupted at any point.
  */
 async function moveIntoPlace(source: string, target: string, logger: Logger) {
   let delay = 100;
   let attempts = 5;
   while (true) {
     try {
-      if (await pathExists(target)) {
-        await fs.unlink(target);
-      }
+      // 'rename' is guaranteed to overwrite an existing target, as long as it is a file (not a directory)
       await fs.rename(source, target);
       return;
     } catch (e: any) {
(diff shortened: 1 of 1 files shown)


hkford:
Exemption Request - All integration tests remain unchanged so I think we do not need to update snapshots in this PR.

corymhall (dismissed):
Can you add `diffAssets: true` to the integ test here? Since this integ test tests custom resources this should have been set to true and will show the diff.

https://github.com/aws/aws-cdk/blob/53d24f1a1357410746052bbf25b5b89b985f7b80/packages/@aws-cdk-testing/framework-integ/test/aws-ecr/test/integ.repository-auto-delete-images.ts#L19-L19

tmokmss:
@hkford 
You should run the integration test locally in your AWS environment to update the snapshots and make sure it works. :)
c.f. https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#integration-tests

packages/aws-cdk-lib/aws-ecr/lib/auto-delete-images-handler/index.ts
```diff
@@ -40,14 +40,23 @@ async function emptyRepository(params: ECR.ListImagesRequest) {
   const listedImages = await ecr.listImages(params).promise();
 
   const imageIds = listedImages?.imageIds ?? [];
+  const imageIdsTagged = imageIds.filter((imageId) => 'imageTag' in imageId) ?? [];
```
corymhall, line 43: Looks like you would be processing the tagged images twice. Can we instead create two lists that don't have overlap?

Maybe something like

```ts
const imageIds = [];
const imageIdsTagged = [];
(listedImages.imageIds ?? []).forEach(imageId => {
  if ('imageTag' in imageId) {
    imageIdsTagged.push(imageId);
  } else {
    imageIds.push(imageId);
  }
});
```

packages/aws-cdk-lib/aws-ecr/test/auto-delete-images-handler.test.ts
corymhall, line 1: Can you add a test where some images are tagged and some are not (with mixed ordering) so that we can assert that it deletes the tagged images first?

corymhall (changes requested):
Looks good! Just a couple of minor comments.

corymhall (dismissed):

corymhall (approved):

PR #25802, opened 2023-06-01
chore: avoid importing directly from 'aws-cdk-lib' in alpha packages

thdxr:
now that `aws-cdk-lib` exports the `core` sub-package, the alpha packages should import from there instead of `aws-cdk-lib`

otherwise any usage of these alpha packages will cause node to traverse every single file in `aws-cdk-lib`


WarrenBuffering:
lgtm

snewell92:
@mrgrain this improves upon #25543 if you wanna take a look 😊

corymhall:
@thdxr thanks for picking this up! Looks like there are just some linting issues with the build. Once it is passing ping me and I'll approve.

thdxr:
I cannot figure out how to run the linters autofix

there seems to be tsconfig files missing in all of these packages that the linter is looking for - am i doing something wrong?

corymhall:
> I cannot figure out how to run the linters autofix
> 
> there seems to be tsconfig files missing in all of these packages that the linter is looking for - am i doing something wrong?

You should be able to run `yarn build --fix` in the package.

corymhall (approved):

PR #25822, opened 2023-06-01
fix(kms): aliasName references alias itself (under feature flag)

### LINKED ISSUE #25761: (s3): dependency not added when key alias provided as encryptionKey to bucket
### Describe the bug

Our team has an S3 `Bucket` construct that uses a KMS `Alias` construct as its (the `Bucket`'s) `encryptionKey`. When the CDK code is compiled down to CloudFormation, the `BucketEncryption`.`ServerSideEncryptionConfiguration`.`KMSMasterKeyID` compiles down to something like

```
{
 "Fn::Join": [
  "",
  [
   "arn:",
   {
    "Ref": "AWS::Partition"
   },
   ":kms:<REGION>:<ACCOUNT_ID>:alias/<ALIAS_NAME>"
  ]
 ]
}
```

instead of something like

```
{
 "Ref": "<ALIAS_LOGICAL_ID>"
}
```

which was unexpected.

This impacted us in production in the following way. We had already deployed our `Bucket` previously, and writes to the `Bucket` are on the critical path for some of our APIs. We wanted to deploy the `Key` and corresponding `Alias` as well, so we did so. However, during the deployment for about 2 minutes, the APIs for which the `Bucket` is on the critical path were giving Internal Server Errors due to `AmazonS3Exception: Alias <ALIAS_ARN> is not found.`. We found that this was because, without a logical dependency between `Bucket` and `Alias`, the `Bucket`'s `encryptionKey` was updated before the `Alias` had been created.

### Expected Behavior

We expected the `BucketEncryption`.`ServerSideEncryptionConfiguration`.`KMSMasterKeyID` to compile down to something like

```
{
 "Ref": "<ALIAS_LOGICAL_ID>"
}
```

so that there would be a logical dependency between the `Bucket` and the `Alias`, which is what you would expect from looking at the CDK code.

### Current Behavior

The `BucketEncryption`.`ServerSideEncryptionConfiguration`.`KMSMasterKeyID` compiled down to something like

```
{
 "Fn::Join": [
  "",
  [
   "arn:",
   {
    "Ref": "AWS::Partition"
   },
   ":kms:<REGION>:<ACCOUNT_ID>:alias/<ALIAS_NAME>"
  ]
 ]
}
```

instead.

### Reproduction Steps

Create a stack consisting only of the following CDK constructs:

```
const keyPolicy = new PolicyDocument({
  statements: [
    new PolicyStatement({
      sid: 'FooStmt',
      'actions': [
        'kms:Decrypt',
        'kms:GenerateDataKey'
      ],
      principals: [new AnyPrincipal()],
      resources: ['*'],
    })
  ]
})

const key = new Key(this, 'FooBucketKey', {
  enableKeyRotation: true,
  enabled: true,
  policy: keyPolicy,
  removalPolicy: RemovalPolicy.RETAIN
})

const alias = new Alias(this, 'FooBucketKeyAlias', {
  aliasName: 'fooBucketKeyAlias',
  targetKey: key
})

const bucket = new Bucket(this, 'FooBucket', {
  bucketName: 'fooBucket',
  removalPolicy: RemovalPolicy.RETAIN,
  blockPublicAccess: BlockPublicAccess.BLOCK_ALL,
  versioned: true,
  encryption: BucketEncryption.KMS,
  encryptionKey: alias,
  enforceSSL: true
})
```

### Possible Solution

When attaching an `IKey` to a resource, check if the `IKey` is an `Alias`, and if so, compile down to a reference instead of an ARN.

_No response_

### CDK CLI Version

2.14.0

### Framework Version

_No response_

### Node.js Version

14.21.3

### OS

Amazon Linux 2

### Language

Typescript

### Language Version

Typescript (4.3.0)

### Other information

_No response_

kaizencc:
Fixes #25761

The existing behavior was that `alias.aliasName` was always a raw string, so properties like `keyId` and `keyArn` depended on a raw string and not a reference to the alias itself. This behavior is preserved.

Under a new feature flag, `const KMS_ALIAS_NAME_REF = '@aws-cdk/aws-kms:aliasNameRef'`, we instead use a reference to the `aliasName` output itself, which means that properties that depend on `aliasName` now depend on the `alias`. In turn, the `alias` depends on the `key`. This allows the expected behavior where specifying something like `alias.keyArn()` depends on the key.

----


otaviomacedo (approved):

ottokruse:
This change breaks my existing code, because of the `Ref` I now get a CloudFormation circular dependency. I want the raw string, not the `Ref`. Is there a way to get the raw string?

ottokruse:
This is because I want to use the name of the alias in a condition statement of my key policy

ottokruse:
FYI doing this to get the raw string again:

```
key.node.defaultChild as cdk.aws_kms.CfnAlias).aliasName
```

PR #25840, opened 2023-06-02
feat(ecs): enable alarm-based rollbacks

bvtujo:
This PR enables the new Deployment Alarms feature in ECS for Service L2s [blog post](https://aws.amazon.com/blogs/containers/automate-rollbacks-for-amazon-ecs-rolling-deployments-with-cloudwatch-alarms/).

This PR contains changes after the first round of revision by the CDK team. 

Continuation of #25346

**Why are so many integration tests impacted by this change? There are 2 reasons:**
1. This PR changes the ECS L2s to set the default configuration for the `CfnService.deploymentConfiguration.alarms` property to:
```
alarmNames: [],
rollback: false,
enable: false,
```
This is necessary, because adding deployment alarms, deploying your CFN stack, then removing the deployment alarms from the CFN template, and deploying again WILL NOT remove the deployment alarms from the service. To remove previously configured deployment alarms, you must explicitly use the configuration shown above. Making this update will cause no interruption to existing ECS services, and does not trigger any update to the service itself during the CFN update.

The ECS UpdateService API is stateful, meaning that if a field is not present in the CloudFormation object, it will be ignored in the update. This was originally implemented due to the problems with desiredCount resetting to bad values after autoscaling changed it, but requires us to set an explicit disableDeploymentAlarms() method to cause the service update to behave correctly.

2. Most ECS integ tests have not been executed in a long time. So, many of the changes are bring the snapshots up to date with the latest integ test format, or lambda layers, etc.

**Detailed list of changes:**
- entire alarm based rollback task into single commit
- resolving merge conflicts
- integ files modules updated to resolve merge conflicts
- regarding merge conflicts issue
- integ files import changes for merge conflicts
- adding this change for PR not to abondended
- chore: refactor createAlarm & remove unnecessary unit tests
- chore: fix api but failing unit tests
- chore: fix validation
- chore: readme changes
- chore: fix integ test, remove unnecessary ec2 test
- chore: refactor enableAlarms for clarity
- test: fix integ test, finally
- docs: update docstrings for new props and methods

----


packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -607,10 +692,169 @@ export abstract class BaseService extends Resource
...
-  /**   * Enable Service Connect
+  /**
+   * Enable Deployment Alarms which take advantage of arbitrary alarms and configure them after service initialization.
+   * If you have already enabled deployment alarms, this function can be used to tell ECS about additional alarms that
+   * should interrupt a deployment.
+   *
+   * New alarms specified in subsequent calls of this function will be appended to the existing list of alarms.
+   *
+   * Subsequent calls of this function will respect the Alarm Behavior you set previously, unless you specify a new
+   * `behavior`.
```
madeline-k, line 709: [p0] This goes against the design principles of CDK constructs. While you use imperative languages to write your CDK apps, the resources are defined in a declarative way. The order in which you call functions should not impact the resulting infrastructure. Instead, we will just throw an error if the user tries to mix different Alarm Behaviors. What is the intent behind that anyway? Why should the last specified behavior be the one that "wins"?

madeline-k (dismissed):
Hey @bvtujo, thanks for your patience on this one. It was a bit of a challenge. 

All of my feedback I have implemented in another branch, which you can merge into your branch if you are aligned with it. Checkout the commit here: https://github.com/aws/aws-cdk/commit/8017b6f117260190180f3a17f11cc06141a48809?diff=split

madeline-k (dismissed):
This looks great! Thanks for incorporating all of my feedback. I have a few tiny comments remaining on the documentation.

madeline-k (approved):
🎉

PR #25841, opened 2023-06-03
fix(s3): fail fast for s3 lifecycle configuration when ExpiredObjectDeleteMarker specified with ExpirationInDays, ExpirationDate, or TagFilters.

### LINKED ISSUE #25824: aws-cdk/aws-s3: cdk can generate an invalid lifecycle configuration for s3:Bucket
### Describe the bug

Synthed buckets can end up with an invalid lifecycle configuration which fails to be applied by CFN when adding the `expiredObjectDeleteMarker` configuration.

For example, according to the CFN docs for [ExpiredObjectDeleteMarker](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-lifecycleconfig-rule.html#cfn-s3-bucket-rule-expiredobjectdeletemarker):

> This cannot be specified with `ExpirationInDays`, `ExpirationDate`, or `TagFilters`.

However, if the developer sets `expire` and `expiredObjectDeleteMarker`, CDK synths the code with no errors.

### Expected Behavior

An invalid lifecycle configuration should fail to synth.

### Current Behavior

When an invalid lifecycle configuration is specified, CDK synths the artifacts without raising any warning or errors. The error only occurs when applying the template in CFN. You get this fantastically cryptic error in the CFN logs, and the stack rolls back:

```
The XML you provided was not well-formed or did not validate against our published schema (Service: Amazon S3; Status Code: 400; Error Code: MalformedXML;)
```

### Reproduction Steps

This code would generate a template that would fail to be applied in CFN:

```typescript
new s3.Bucket(this, 'bucket', {
      lifecycleRules: [
        {
          enabled: true,
          expiration: Duration.days(14),
          noncurrentVersionExpiration: Duration.days(7),
          abortIncompleteMultipartUploadAfter: Duration.days(1),
          expiredObjectDeleteMarker: true,
        },
      ],
    });
```

### Possible Solution

If the user adds an invalid combination of lifecycle configurations, we could do any of these:

* prominently warn the user that the configuration will fail in CFN
* just fail the build completely

_No response_

### CDK CLI Version

2.79.1

### Framework Version

_No response_

### Node.js Version

19.1.0

### OS

Linux

### Language

Typescript

### Language Version

4.9.5

### Other information

_No response_

Zishanwang1992:
Closes #25824.

----


Zishanwang1992:
Exemption Request: This fix is to fail fast. The integ test has no difference compared to unit test. The unit test has been added for the scenario.

abelmokadem:
Just a random thought. Why not fix this in the type definition of Lifecycle Rule?

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -2085,6 +2085,11 @@ export class Bucket extends BucketBase {
 
     function parseLifecycleRule(rule: LifecycleRule): CfnBucket.RuleProperty {
       const enabled = rule.enabled ?? true;
+      if ((rule.expiredObjectDeleteMarker != null && rule.expiredObjectDeleteMarker != undefined)
+      && (rule.expiration || rule.expirationDate || self.parseTagFilters(rule.tagFilters))) {
+        // ExpiredObjectDeleteMarker cannot be specified with ExpirationInDays, ExpirationDate, or TagFilters.
+        throw new Error('ExpiredObjectDeleteMarker cannot be specified with expiration, ExpirationDate, or TagFilters.');
+      }
```
corymhall, line 2092:
```suggestion
      if ((rule.expiredObjectDeleteMarker)
      && (rule.expiration || rule.expirationDate || this.parseTagFilters(rule.tagFilters))) {
        // ExpiredObjectDeleteMarker cannot be specified with ExpirationInDays, ExpirationDate, or TagFilters.
        throw new Error('ExpiredObjectDeleteMarker cannot be specified with expiration, ExpirationDate, or TagFilters.');
      }
```

corymhall (changes requested):
Thanks for the PR! Just 2 minor comments

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
Zishanwang1992: The reason I do not use the "if ((rule.expiredObjectDeleteMarker))" is the type of "expiredObjectDeleteMarker" is boolean. If its value is false, the validation here will not work.  I assume "false" could also not be specified with ExpirationInDays, ExpirationDate, or TagFilter according to public doc.

See if it makes sense.

Zishanwang1992: Btw, the "this.parseTagFilters(rule.tagFilters)" will not compile, and there is same "this.parseTagFilters(rule.tagFilters)" usage at line 2112/2117. My change is to stay consistent with existing code.

See if it makes sense.

packages/aws-cdk-lib/aws-s3/test/rules.test.ts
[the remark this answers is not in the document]
Zishanwang1992: Will add different combinations.

Zishanwang1992:
> Just a random thought. Why not fix this in the type definition of Lifecycle Rule?

I am not sure if we can do this in the type definition of Lifecycle Rule, because there is complex validaiton logic. We can easily do this when Lifecycle rule gets initialized and the current code shall be the only place to initialize the Lifecycle rule.

packages/aws-cdk-lib/aws-s3/test/rules.test.ts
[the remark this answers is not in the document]
Zishanwang1992: @corymhall Could you help review this PR?

packages/aws-cdk-lib/aws-s3/test/rules.test.ts
```diff
@@ -25,6 +25,51 @@ describe('rules', () => {
     });
   });
 
+  test('ExpiredObjectDeleteMarker cannot be specified with ExpirationInDays, ExpirationDate, or TagFilters.', () => {
```
corymhall, line 28:
```suggestion
  test('ExpiredObjectDeleteMarker cannot be specified with ExpirationInDays.', () => {
```

packages/aws-cdk-lib/aws-s3/test/rules.test.ts
```diff
@@ -25,6 +25,51 @@ describe('rules', () => {
...
+        expiredObjectDeleteMarker: true,
+      }],
+    });
+
+    expect(() => {
+      Template.fromStack(stack).toJSON();
+    }).toThrow('ExpiredObjectDeleteMarker cannot be specified with expiration, ExpirationDate, or TagFilters.');
+  });
+
+  test('ExpiredObjectDeleteMarker cannot be specified with ExpirationInDays, ExpirationDate, or TagFilters.', () => {
```
corymhall, line 42:
```suggestion
  test('ExpiredObjectDeleteMarker cannot be specified with ExpirationDate.', () => {
```

packages/aws-cdk-lib/aws-s3/test/rules.test.ts
```diff
@@ -25,6 +25,51 @@ describe('rules', () => {
...
```
corymhall, line 56:
```suggestion
  test('ExpiredObjectDeleteMarker cannot be specified with TagFilters.', () => {
```

corymhall (approved):

PR #25845, opened 2023-06-04
feat(ecr): repo.grantPush

watany-dev:
For example, I realized that the role given to perform a push from Github Actions to ECR is excessive if using grantPullPush. The Readme was temporarily updated to fulfill the conditions of a 'feat' commit.

----


watany-dev:
Do we need a new integ-test?

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
```diff
@@ -361,8 +366,33 @@ export abstract class RepositoryBase extends Resource implements IRepository {
...
+      scope: this,
+    });
 
+    return ret;
+  }
+
+  /**
+   * Grant the given identity permissions to use the images in this repository
+   */
+  public grantPush(grantee: iam.IGrantable) {
```
kaizencc, line 388: This feels reasonable to me but I don't have too much context in ecr so just going to make this comment here. 

If we go this route, remember to change `grantPullPush` to call `grantPush` -- there's no reason to duplicate the code in both functions.

But then we also have to audit `ecr:BatchCheckLayerAvailability`, because that's not part of the permissions for `grantPullPush`. Why is it part of `grantPush`? And if it is necessary, it should be a part of `grantPullPush` as well.

watany-dev: There was an issue with the 'getauthorizationtoken' being duplicated when executing 'grantPull' and 'grantPush' within 'grantPullPush'. I have extracted the other Actions into S3 Constructs and ensured readability by commonizing them. I would like to discuss whether this answer is appropriate. I am not sure about the details of 'BatchCheckLayerAvailability', but it is an Action in accordance with the official documentation below. I have made it compatible with both 'Pull' and 'PullPush'. 
https://docs.aws.amazon.com/ja_jp/AmazonECR/latest/userguide/image-push.html#image-push-iam

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
```diff
@@ -373,17 +379,37 @@ export abstract class RepositoryBase extends Resource implements IRepository {
...
-    return this.grant(grantee,
-      'ecr:PutImage',
-      'ecr:InitiateLayerUpload',
-      'ecr:UploadLayerPart',
-      'ecr:CompleteLayerUpload',
+    const ret = this.grant(grantee,
+      ...perm.REPO_PULL_ACTIONS,
+      ...perm.REPO_PUSH_ACTIONS,
     );
+    iam.Grant.addToPrincipal({
```
otaviomacedo, line 405: To deal with the duplication issue, you can declare both sets of actions as `Set<string>`, both of which containing `'ecr:GetAuthorizationToken'`. Then you can transform them to arrays before passing them to `this.grant()`.

otaviomacedo (changes requested):

packages/aws-cdk-lib/aws-ecr/lib/perms.ts
```diff
@@ -0,0 +1,15 @@
+export const REPO_PULL_ACTIONS: string[] = [
```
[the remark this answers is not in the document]
watany-dev: fixed.

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
watany-dev: Thank you for the suggestion. Unfortunately, the ecr:GetAuthorizationToken action is invoked not against a registry, but against the entire ECR service, requiring its resource to be set as '*'. Therefore, I believe it's difficult to handle it along with other Actions within the grant. With this in mind, if there are any necessary refactorings, I would appreciate your guidance.

packages/aws-cdk-lib/aws-ecr/lib/repository.ts
tmokmss: I'd like to confirm if the duplication is really an issue here. `iam.Grant` is wise enough to get rid of any duplication, e.g. even if you allow `ecr:GetAuthorizationToken` twice, the policy is synthesized without any error and it has just a single allow policy. Afaik it should be valid and safe to call `grantPull` and `grantPush` in `grantPullPush`.

otaviomacedo (approved):

Issue #24072, opened 2023-02-08
(pipelines): (Add feature to display all changesets for stage before manual approval step)


Nico-DB:
### Describe the feature

I would like to "run" all changesets of a stage at first, then have a manual approval step and then perform the execute change set in the original run-order. 

### Use Case

The use case is, that we would like to check the changesets for all stacks of an environment, to be able to perform a qualified  manual approval.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.63.2

### Environment details (OS name and version, etc.)

Linux a-3pq6x6lu5kukl 5.15.86-53.137.amzn2.x86_64 #1

tim-finnigan:
Hi @Nico-DB thanks for creating this feature request. Could you provide more details as far as your use case and what changes you're proposing? You are welcome to go ahead and create a PR, here is the Contributing guide for reference: https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md. I can't guarantee that the team would consider implementing the changes you're proposing but it would help to have some code to review in order to better understand the proposed use case and solution.

Nico-DB:
Hi @tim-finnigan ,

i have attached a picture showing the use-case i mentioned. This is done via the codepipeline construct.

![Example-manual-approval](https://user-images.githubusercontent.com/64137458/217794564-8f7be0d2-02d0-4047-aae0-083cced6ef92.png)

tim-finnigan:
Thanks for following up — if there is any more information anyone wants to share as far as use cases for this feature, please leave a comment here. Others can also 👍 the issue if interested in this feature.

PR #25852, opened 2023-06-05
feat(pipelines): Add feature to display all changesets for stage before manual approval step + added postPrepare Step #25413

Nico-DB:
Describe the feature

I would like to "run" all changesets of a stage at first, then have a manual approval step and then perform the execute change set in the original run-order.

Use Case

The use case is, that we would like to check the changesets for all stacks of an environment, to be able to perform a qualified manual approval.

Closes https://github.com/aws/aws-cdk/issues/24072.
----


Nico-DB:
@rix0rrr I have created some new integ tests for the pipelines construct. Since the cdk-pipelines-demo repo is private, i cannot generate the snapshots. Can you please run the tests?

Nico-DB:
@hoegertn Here is the new pr

Nico-DB:
Exemption Request:
I am not able to adapt the needed snapshots since the pipelines do point to a privat repository by [rix0rrr](https://github.com/rix0rrr)

Nico-DB:
This should stay open.

PR #25858, opened 2023-06-05
feat(stepfunctions): allow setting comment on state machine

mschelstastic:
Just a small PR to allow setting top level comment on a state machine

### DIFF
--- a/packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
+++ b/packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
@@ -314,7 +314,7 @@ export interface GraphqlApiProps {
   /**
    * GraphQL schema definition. Specify how you want to define your schema.
    *
-   * Schema.fromFile(filePath: string) allows schema definition through schema.graphql file
+   * SchemaFile.fromAsset(filePath: string) allows schema definition through schema.graphql file
    *
    * @default - schema will be generated code-first (i.e. addType, addObjectType, etc.)
    *
--- a/packages/aws-cdk-lib/aws-certificatemanager/lib/certificate.ts
+++ b/packages/aws-cdk-lib/aws-certificatemanager/lib/certificate.ts
@@ -94,9 +94,9 @@ export interface CertificateProps {
   readonly transparencyLoggingEnabled?: boolean;
 
   /**
-   * The Certifcate name.
+   * The Certificate name.
    *
-   * Since the Certifcate resource doesn't support providing a physical name, the value provided here will be recorded in the `Name` tag
+   * Since the Certificate resource doesn't support providing a physical name, the value provided here will be recorded in the `Name` tag
    *
    * @default the full, absolute path of this construct
    */
--- a/packages/aws-cdk-lib/aws-certificatemanager/test/certificate.test.ts
+++ b/packages/aws-cdk-lib/aws-certificatemanager/test/certificate.test.ts
@@ -407,7 +407,7 @@ describe('Transparency logging settings', () => {
   });
 });
 
-describe('Certifcate Name setting', () => {
+describe('Certificate Name setting', () => {
   test('the Name tag is defaulted to path', () => {
     const stack = new Stack(undefined, 'TestStack');
 
--- a/packages/aws-cdk-lib/aws-cloudfront/lib/web-distribution.ts
+++ b/packages/aws-cdk-lib/aws-cloudfront/lib/web-distribution.ts
@@ -555,7 +555,7 @@ export class ViewerCertificate {
   }
 
   /**
-   * Generate a viewer certifcate configuration using
+   * Generate a viewer certificate configuration using
    * the CloudFront default certificate (e.g. d111111abcdef8.cloudfront.net)
    * and a `SecurityPolicyProtocol.TLS_V1` security policy.
    *


curquhart:
this is "ready" but the build is failing for what appears to be an unrelated reason... going to mark as draft until i have the cycles to look into that more thoroughly (also going to update from main in case it was just a flake)

curquhart:
build succeeded on re-run so good to go :)

curquhart:
I probably won't have time to look into the merge conflict until next week. Let me know if I should just close this until then or not (donno if the new DefinitionBody.fromChainable stuff sets the sfn comment or not so need to test, but def don't have time this week.

curquhart:
Just wanted to follow up on this- would appreciate if someone has time to look at it thanks :)

packages/aws-cdk-lib/aws-stepfunctions/lib/state-graph.ts
```diff
@@ -35,6 +35,16 @@ export class StateGraph {
    */
   public timeout?: Duration;
 
+  /**
+   * Set a comment to render into the graph JSON.
+   *
+   * Read/write. Only makes sense on the top-level graph, subgraphs
+   * do not support this feature.
+   *
+   * @default No comment
```
vinayak-kukreja, line 44: **NIT**: Could you change this to `@default - No comment`.

packages/aws-cdk-lib/aws-stepfunctions/lib/state-machine.ts
```diff
@@ -119,6 +119,13 @@ export interface StateMachineProps {
    */
   readonly timeout?: Duration;
 
+  /**
+   * Comment that describes this state machine
+   *
+   * @default No comment
```
vinayak-kukreja, line 125: Similar to earlier comment :)

vinayak-kukreja (changes requested):
Hey, this looks great. Thank you for contributing.
I have just some NITs. 

Tagging @iliapolo for a second opinion.

packages/aws-cdk-lib/aws-stepfunctions/lib/state-machine.ts
curquhart: heh yup, np, updated :D

vinayak-kukreja:
Thank you for addressing the feedback. :) LGTM, syncing with Eli to get a second opinion.

packages/aws-cdk-lib/aws-stepfunctions/lib/state-graph.ts
```diff
@@ -35,6 +35,16 @@ export class StateGraph {
    */
   public timeout?: Duration;
 
+  /**
+   * Set a comment to render into the graph JSON.
+   *
+   * Read/write. Only makes sense on the top-level graph, subgraphs
+   * do not support this feature.
```
iliapolo, line 42: I understand you followed the same pattern as the `timeout` property - but this behavior makes me a little uncomfortable. Lets see if we can make it better?

See other comments.

packages/aws-cdk-lib/aws-stepfunctions/lib/state-machine.ts
```diff
@@ -707,6 +714,7 @@ export class ChainDefinitionBody extends DefinitionBody {
   public bind(scope: Construct, sfnPrincipal: iam.IPrincipal, sfnProps: StateMachineProps): DefinitionConfig {
     const graph = new StateGraph(this.chainable.startState, 'State Machine definition');
     graph.timeout = sfnProps.timeout;
+    graph.comment = sfnProps.comment;
```
iliapolo, line 717: What if instead we do:

```ts
const graphJson = Stack.of(scope).toJsonString(graph.toGraphJson());
return {
  definitionString: { Comment: sfnProps.comment, ...graphJson },
};
```

Basically having the state machine augment the graph with properties that are only relevant to the state machine as a whole - and not to sub-graphs.

curquhart:
> @curquhart leaving a request changes review to reflect the current state of this PR and remove the `needs-review` label. We'll take another look when outstanding comments are addressed!

sure thing. Sorry haven't addressed yet, I'm on vaca at the moment, should address it next week.

packages/aws-cdk-lib/aws-stepfunctions/lib/state-machine.ts
curquhart: finally updated,

```
    const graphJson = graph.toGraphJson();
    return {
      definitionString: Stack.of(scope).toJsonString({ ...graphJson, Comment: sfnProps.comment }),
    };
```

how's that? I included comment after since the graph always sets it (even if undefined) and {something: 'foo', ... {something:undefined}} will override with undefined

packages/aws-cdk-lib/aws-stepfunctions/lib/state-graph.ts
```diff
@@ -113,6 +123,7 @@ export class StateGraph {
       StartAt: this.startState.stateId,
       States: states,
       TimeoutSeconds: this.timeout && this.timeout.toSeconds(),
+      Comment: this.comment,
```
kaizencc, line 126: I don't think changes to `StateGraph` are necessary anymore. `StateGraph` is exposed but not meant to be used externally. No one should be setting `comment` on `StateGraph` and internally we are not doing that anymore either. So there's no need to have this prop.

(This means that the ordering between `...graphJson` and `Comment: sfnProps.comment` won't matter anymore, but its fine to have graphJson in front).

curquhart: haha... I forgot i added it to state-graph, totally makes sense to take it out - updated

kaizencc (approved):
Thanks @curquhart! Looks good

PR #25859, opened 2023-06-05
chore: experimental cdk migrate command

HBobertz:
cdk migrate command.

See readme for additional description of command. Default input path is a file named "template.txt" in the parent directory for now. Default output path is the current directory. If you have any issues with the actually generated code, please create an issue on https://github.com/iph/noctilucent

Please refer to noctilucent or command help supported languages to see what languages are supported.

----


packages/aws-cdk/THIRD_PARTY_LICENSES
```diff
@@ -2228,6 +2228,32 @@ IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
Naumel, line 2254: This seems standard. Do we have to involve anyone else to review the wording?

packages/aws-cdk/generate.sh
```diff
@@ -13,4 +13,34 @@ cat > build-info.json <<HERE
...
+  # Check out the submodule if it's not there already
+  if [ ! -f "vendor/noctilucent/Cargo.toml" ]; then
+    git -C ./vendor clone https://github.com/iph/noctilucent.git
+    git -C ./vendor/noctilucent reset --hard HEAD
+    git -C ./vendor/noctilucent git fetch && git checkout 6da7c9fade55f8443bba7b8fdfcd4ebfe5208fb1
+  fi
+
+  # update the package to the pinned commit hash
+  git -C vendor/noctilucent reset --hard HEAD
+  git -C vendor/noctilucent fetch && git -C vendor/noctilucent checkout 6da7c9fade55f8443bba7b8fdfcd4ebfe5208fb1
```
Naumel, line 29: You only need one line in the if, the other two get executed no matter what, right?

packages/aws-cdk/lib/commands/migrate.ts
```diff
@@ -0,0 +1,71 @@
...
+import * as path from 'path';
+import { initializeProject, availableInitTemplates } from '../../lib/init';
+import { warning } from '../logging';
+import * as nocti from '../vendor/noctilucent';
+
+/** The list of languages supported by the built-in noctilucent binary. */
+export const MIGRATE_SUPPORTED_LANGUAGES: readonly string[] = nocti.supported_languages();
+
+export async function cliMigrate(
+  inputpath: string = process.cwd() + '/../temp.txt',
```
Naumel, line 11: ~~What's the chance of (eventual) name collision?~~

This hardcodes a bunch of input that should be accepted from the command line. Will this actually output things in "~/my-dir" if the cli comand was `cdk migrate -l java -inputpath ~/template.json -outputpath ~/my-dir`?

packages/aws-cdk/lib/commands/migrate.ts
```diff
@@ -0,0 +1,71 @@
...
+  language = MIGRATE_SUPPORTED_LANGUAGES[0],
+  generateOnly = false,
+  outputpath = process.cwd(),
+) {
+  warning('This is a very experimental feature, use at your own risk.');
```
Naumel, line 16:
```suggestion
  warning('This is an experimental feature. We make no guarantees about the outcome or stability of the functionality.');
```

packages/aws-cdk/lib/commands/migrate.ts
```diff
@@ -0,0 +1,71 @@
...
+  const type = 'default'; // "default" is the default type (and maps to 'app')
```
Naumel, line 17: I had to double-take: why is this an option here, but not exposed as an argument?
Even if right now it has a single value, it can be expanded in the future (based on the comment)

Naumel (commented):
I have a couple of questions, if I just got the wrong understanding, we can clear it up and approve the changes.

packages/aws-cdk/lib/commands/migrate.ts
HBobertz: Yup it works, these are just setting the default. We could just delete the /../temp.txt default for the input path if we wanted, I just left it in there for now

packages/aws-cdk/THIRD_PARTY_LICENSES
HBobertz: I'm honestly not sure, romain is the one who included this. I'm not even sure it's totally necessary since this gets auto genned anyway

packages/aws-cdk/generate.sh
HBobertz: derp ill fix this

RomainMuller (approved):

PR #25868, opened 2023-06-06
fix(app-mesh): Missing port property in gRPC routers matchers

### LINKED ISSUE #25810: (app-mesh): Missing port property in GrpcRouteMatch construct
### Describe the bug

While the `CfnRoute` give us the ability to define a gRPC match route rule specifying the port, the L2 construct `GrpcRouteMatch` does not have that property (only have `service_name`, `method_name` and `metadata`), leveraging us to use L1 constructs instead if we want to use the port matcher.

### Expected Behavior

`GrpcRouteMatch` L2 construct also let us specify the `port` matcher option.

### Current Behavior

`GrpcRouteMatch` L2 construct does not have the property `port`.

### Reproduction Steps

Example in python; the following is not valid code:

```python
appmesh.RouteSpec.grpc(
    match=appmesh.GrpcRouteMatch(port=1234),
    weighted_targets=[appmesh.WeightedTarget(virtual_node=node)],
)
```

We need to use L1 construct, to get the same result:

```python
appmesh.CfnRoute.RouteSpecProperty(
    grpc_route=appmesh.CfnRoute.GrpcRouteProperty(
        action=appmesh.CfnRoute.GrpcRouteActionProperty(
            weighted_targets=[appmesh.CfnRoute.WeightedTargetProperty(
                virtual_node=node.virtual_node_name,
                weight=1,
            )],
        ),
        match=appmesh.CfnRoute.GrpcRouteMatchProperty(port=1234),
    ),
)
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.81.0

### Framework Version

2.81.0

### Node.js Version

v16.14.2

### OS

MacOs Ventura

### Language

Python

### Language Version

python 3.11.1

### Other information

I have checked through the CDK docs that typescript CDK version have the same issue, so I guess this applies to all languages available.

neovasili:
As described in the related issue, `GrpcRouteMatch` L2 construct was missing `port` property which is already present in the L1 construct `CfnRoute`.

This PR adds the missing `port` property to `GrpcRouteMatch` L2 construct and also adds `port` property to `GrpcGatewayRouteMatch` and `HttpRouteMatch` L2 constructs that were also missing it.

The PR includes unit and integration tests expansion to cover this new property plus a reference to the property in the appmesh README file.

Closes #25810.

----


yarn.lock
corymhall, line 1: yarn.lock shouldn't need to be updated.

corymhall (changes requested):
@neovasili thanks for the PR! just 1 minor thing.

yarn.lock
neovasili: updated! 😄

corymhall (approved):

neovasili:
nice! **Allow edits from maintainers** is already enabled 😄

PR #25876, opened 2023-06-06
feat(s3-deployment): create `DeployTimeSubstitutedFile` to allow substitutions in file

### LINKED ISSUE #1461: Parameterized swagger support for API gateway
I know that Swagger integration for the API-gateway resources is on the CDK roadmap.

I think it can bring a lot of value, since at the moment, if you want to use (using CFN) a swagger file you can either inline it and reference CFN resources, or import it and not reference anything.

Typically, you'd want both - you want to separate the swagger file from your CFN/CDK code so that you can use all the fancy tools (UI editor / client generation / etc), but also usually you'd need to reference CFN resources (e.g. lambda integrations).

With CDK it can be possible to have a templated external swagger file, and use string replacements for the referenced resources.

Took this offline with @eladb who suggested something in the lines of:
```
new apigateway.RestApi(this, 'MyAPI', {
  swagger: Swagger.load('/path/to/swagger', {
    xxx: myLambdaFunction.functionName,
    yyy: myAuthenticator.functionArn
  }
});
```

I think it could bring huge value to CDK users as you can use the "API first" methodology and leverage all the existing swagger tools.

sumupitchayan:
Closes #1461 

The `DeployTimeSubstitutedFile` construct allows you to upload a file and specify substitutions to be made in it, which will be resolved during deployment.

For example, if you wanted to create a REST API from a Swagger file spec but want to reference other CDK resources in your API spec, you can now do so in-line:

```ts
const bucket: Bucket;
const myLambdaFunction: lambda.Function;

const deployment = new s3deploy.DeployTimeSubstitutedFile(this, 'MyApiFile', {
  source: 'my-swagger-spec.yaml',
  destinationBucket: bucket,
  substitutions: {
    xxxx: myLambdaFunction.functionArn,
    yyyy: 'mySubstitution',
  },
});

const api = new apigateway.SpecRestApi(this, 'books-api', {
  apiDefinition: apigateway.ApiDefinition.fromBucket(deployment.bucket, deployment.objectKey),
});
```

Where 'xxxx' and 'yyyy' are the examples of placeholder text you can add in your local file spec to be substituted by surrounding the placeholder with double curly braces, for example writing: `{{ xxxx }}` in your file where you want a substitution.

----


packages/@aws-cdk-testing/framework-integ/test/aws-apigateway/test/integ.parametrized-swagger.api.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+}
+
+const app = new cdk.App();
+
+const testCase = new Test(app, 'test-apigateway-parameterized-swagger');
+new IntegTest(app, 'apigateway-parameterized-swagger', {
+  testCases: [testCase],
+});
```
otaviomacedo, line 48: To really make sure this is working end-to-end, we also need an [assertion](https://docs.aws.amazon.com/cdk/api/v1/docs/integ-tests-readme.html#assertions) to make sure that the variables are replaced.

otaviomacedo (changes requested):

packages/aws-cdk-lib/aws-s3-deployment/README.md
```diff
@@ -355,6 +355,36 @@ The value in `topic.topicArn` is a deploy-time value. It only gets resolved
...
+  destinationBucket: destinationBucket,
+  substitutions: {
+    variable-name: myLambdaFunction.functionName,
+  },
+});
+```
+
+Nested variables, like `{{ {{ foo }} }}` or `{{ foo {{ bar }} }}`, are not supported by this
+construct. In the first case of a single variable being is double nested `{{ {{ foo }} }}`, only 
+the `{{ foo }}` would be replaced by the substitution, and the extra brackets would remain in the file.
```
otaviomacedo, line 384: What if you have the following substitutions?

```ts
{
  foo: 'bar',
  bar: 'zee',
}
```

We might need a smarter regex. Something like:

```ts
s.replace(/{{\s*(\w+)\s*}}/g, function(match, expr) {
    return substitutions[expr] ?? expr;
})
```

This will be faster and only replace the innermost variable.

packages/aws-cdk-lib/aws-s3-deployment/test/bucket-deployment.test.ts
```diff
@@ -1500,6 +1500,86 @@ test('if any source has markers then all sources have markers', () => {
   });
 });
 
+test('DeployTimeSubstitutedFile can be used to add substitutions in a file', () => {
```
otaviomacedo, line 1503: We also need a test case for a variable that appears in the file but not in the substitutions map.

otaviomacedo (changes requested):

packages/@aws-cdk-testing/framework-integ/test/aws-apigateway/test/integ.parametrized-swagger.api.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const testCase = new Test(app, 'test-apigateway-parameterized-swagger');
+const integ = new IntegTest(app, 'apigateway-parameterized-swagger', {
+  testCases: [testCase],
+});
+
+const invoke = integ.assertions.invokeFunction({
+  functionName: lambdaFunctionArn,
+});
+invoke.expect(ExpectedResult.objectLike({
```
otaviomacedo, line 49: This doesn't prove anything relevant to this feature. The function could be deployed correctly, but the Rest API (which is the resource that depends on the substitutions being correct) could still be using an unsubstituted Open API file. At the very least, the assertion has to call the Rest API. You'll need to be creative and construct an API in such a way that its response contains the substituted variable. Response templates, perhaps?

otaviomacedo (changes requested):

otaviomacedo (approved):

ChrisLane:
This is great. Could the same be done for an `aws_s3.Asset`?

My use-case is that I have a third-party construct that accepts an S3 path for a script to run.
I have an `Asset` that deploys the script and I pass the S3 path from that to the third-party construct but then within the script I need to download a file that itself is deployed to S3 via CDK. This is why I need the replacement in the asset, to maintain the dependency to replace the construct when either of the files change.

alex-cook-finres:
Does this actually work as-is or does it need other changes? If I actually substitute any parameters I end up getting an "Invalid OAS input" failure error. If I hard-code values with the same code (just no substitutions in the file) it works fine. Any help would be great as this seems like a nice way to actually use my API spec compared to other options I've seen out there. 

```typescript
    const deployment = new s3deploy.DeployTimeSubstitutedFile(
      this,
      "MyApiFile",
      {
        source: path.join(__dirname, "../../openapi-spec/api-definition.yaml"),
        destinationBucket: bucket,
        destinationKey: "api-definition.yaml",
        substitutions: {
          LAMBDA_ARN: lambdaFunction.functionArn,
          AWS_REGION: this.region,
        },
      }
    );

    const api = new apigateway.SpecRestApi(this, "CdkApi", {
      apiDefinition: apigateway.ApiDefinition.fromBucket(
        deployment.deployedBucket, // also tried with deployment.bucket
        deployment.objectKey
      ),
    });
```

PR #19656, opened 2022-03-31
fix(core): reduce CFN template indent size to save bytes

PatMyron:
also: https://github.com/aws/serverless-application-model/pull/2368, https://github.com/cloudtools/troposphere/pull/2028, https://github.com/awslabs/goformation/pull/445
https://github.com/aws/aws-cdk/issues/18694
https://github.com/aws/aws-cdk/pull/18886
[CloudFormation templates can currently only be 1MB](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html)

Simply reducing indentation from 2 to 1 should remove ~1/4 of the template file size for everyone by default while still preserving indentation formatting. Beyond improved defaults, those wishing to reduce readability for further reduced file size could opt into using something like [`jq`](https://stedolan.github.io/jq/) on their own for now

Issue #8712, opened 2020-06-24
[cli] cdk synth should allow for an option to compact the template file (remove whitespaces)


mcliff1:
When producing JSON output cdk synth should have an option to minimize the output template file size.

### Use Case
AWS Cloudformation  has a limit on the template body size for an template on S3 at 460,800 bytes.

I had attempted to inject this type of filtering after cdk synth in the `cdk deploy` flow, but quickly got lost.

This will allow deployment of larger stacks.

### Proposed solution
Add an option on CDK Synth that will strip out all white spaces from the output template JSON (in my use case this reduced the size from 500k to 250k).
 
Possibly this could be an option for *cdk.json*

### Other

Using Nested Stacks is the most common way to reduce stack size, in my use case I was doing a lot of with Cross Account CodePipelines which created many permissions objects and was very sensitive to repacking in nested stacks.

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

rrrix:
In another life in another project, I was using the AWS Amplify CLI, which also outputs everything as very spacey CloudFormation JSON and had hit that same CloudFormation template file size limit. 

I ended up added a "hook" (hack?) in the Amplify CLI to use `cfn-flip` (https://github.com/awslabs/aws-cfn-template-flip) which converts everything to YAML, and has a `--clean` option (which allows it to fixup Intrinsic Functions to use the shorthand YAML versions). 

Perhaps that might be an option worth looking at? 

A quick test was able to reduce a random template I have from `64K` to `38K` (~40% reduction):

```sh
$ cfn-flip -y -c pipeline.template.json pipeline.template.yaml
$ ll pipeline.template.*
-rw-r--r--  1 user  staff    64K Jun 23 18:24 pipeline.template.json
-rw-r--r--  1 user  staff    38K Jun 24 12:23 pipeline.template.yaml
```

mcliff1:
> In another life in another project, I was using the AWS Amplify CLI, which also outputs everything as very spacey CloudFormation JSON and had hit that same CloudFormation template file size limit.
> 
> I ended up added a "hook" (hack?) in the Amplify CLI to use `cfn-flip` (https://github.com/awslabs/aws-cfn-template-flip) which converts everything to YAML, and has a `--clean` option (which allows it to fixup Intrinsic Functions to use the shorthand YAML versions).
> 
> Perhaps that might be an option worth looking at?
> 
[...]

PR #25892, opened 2023-06-07
feat(core): add option to suppress indentation in templates

### LINKED ISSUE #18694: (cdk cli): Create Smaller CloudFormation Templates
### Description

Allow options when deploying using the cdk cli that would result in physically smaller CloudFormation template.

Currently the template that is generated and deployed is formatted as JSON with whitespace.

From our testing, simply removing the whitespace can result in a template that is roughly 50% of the size. 

Outputting the template as YAML also decreases the template size by roughly the same amount.

### Use Case

We have a CDK project that has several lambda functions behind an API. We just ran into the limit where a CloudFormation template has to be less than 1MB. 

The number of resources we have is relatively low, but due to the auto generated names of resources the template size is quite large. 

If we could minify the template before deploying, or if we could have CDK deploy the template as YAML, we would be able to halve the size of the template and deploy twice as many resources.

### Proposed Solution

This would be a combination of previously proposed solutions.

https://github.com/aws/aws-cdk/issues/8712
The above issue suggests having an option to compact the whitespace. This seems like a reasonable proposal.

https://github.com/aws/aws-cdk/issues/2965
The above issue mentions including the option of writing YAML to `cdk.out`. This appears to have been rejected on the premise that the folder is for machines to read, and that the only benefit of YAML is that it's human readable.

This ignores the fact that the YAML format results in a significantly smaller file size, and that AWS CloudFormation has a hard limit on template file sizes.

### Other information

_No response_

### Acknowledge

jesterhazy:
Fixes #18694, #8712 

This change adds an option to suppress indentation in CloudFormation template files. Suppressing indentation will reduce the file size of templates.

Indentation can be set by enabling for specific Stacks using the new `suppressTemplateIndentation` property in `StackProps`, or globally using the new `@aws-cdk/core:suppressTemplateIndentation` context key.

This PR provides additional template size reduction beyond the indentation change in  #19656.

@rix0rrr @mackalex @PatMyron 

----


jesterhazy:
note: integration tests have NOT been run

jesterhazy:
Exemption Request: this change doesn't seem to warrant an integration test.

jesterhazy:
Clarification Request: which readme file would you expect to be update for this?

packages/aws-cdk-lib/core/lib/stack.ts
```diff
@@ -1046,7 +1073,8 @@ export class Stack extends Construct implements ITaggable {
         Annotations.of(this).addInfo(`Number of resources: ${numberOfResources} is approaching allowed maximum of ${this.maxResources}`);
       }
     }
-    fs.writeFileSync(outPath, JSON.stringify(template, undefined, 1));
+    const indent = this._suppressTemplateIndentation ? undefined : 1;
+    fs.writeFileSync(outPath, JSON.stringify(template, undefined, indent));
```
corymhall, line 1077: Can we check the size of the template and if it is too large throw a warning that you should set the `suppressTemplateIndentation` flag or split the resources between stacks.

packages/aws-cdk-lib/core/test/stack.test.ts
corymhall, line 1: Can you add another test that asserts the output is still indented when nothing is set? It doesn't look like an existing test covers that case.

corymhall (changes requested):
@jesterhazy actually after looking into this more I think you are on the right track. It looks like this is the best that we can do (for some reason I was thinking there was still something more we could do). 

I only have 2 minor comments and then I'm good with this.

jesterhazy:
@corymhall what needs to happen to move this forward?

packages/aws-cdk-lib/core/lib/stack.ts
```diff
@@ -1047,7 +1075,20 @@ export class Stack extends Construct implements ITaggable {
         Annotations.of(this).addInfo(`Number of resources: ${numberOfResources} is approaching allowed maximum of ${this.maxResources}`);
       }
     }
-    fs.writeFileSync(outPath, JSON.stringify(template, undefined, 1));
+
+    const indent = this._suppressTemplateIndentation ? undefined : 1;
+    const templateData = JSON.stringify(template, undefined, indent);
+
+    if (templateData.length > TEMPLATE_BODY_MAXIMUM_SIZE) {
+      throw new Error(`Template size for stack '${this.node.path}' exceeds limit: ${templateData.length}/${TEMPLATE_BODY_MAXIMUM_SIZE}`);
```
corymhall, line 1083: Lets change this to a warning too. It will still fail fast on deployment.

corymhall (dismissed):
Can you also add a simple integration test (or update an existing one) that sets the `suppressTemplateIndentations: true`. I just want to have a test that actually deploys a template without indentation.

packages/aws-cdk-lib/core/lib/stack.ts
jesterhazy: ok

corymhall (approved):

PR #25895, opened 2023-06-07
feat: update alpha modules to sdk v3

MrArnoldPalmer:
Updates alpha modules with custom resources to use NodeJS 18 and v3 of the AWS SDK. Since these CRs are pretty self-contained and safe to update, their handler code and runtime versions can be updated without any breakage for users.

----


packages/@aws-cdk/aws-amplify-alpha/lib/asset-deployment-handler/handler.ts
```diff
@@ -49,22 +48,22 @@ export class AmplifyAssetDeploymentHandler extends ResourceHandler {
...
         appId: this.props.AppId,
         branchName: this.props.BranchName,
         sourceUrl: assetUrl,
-      })
-      .promise();
+      });
 
     return {
-      AmplifyJobId: deployment.jobSummary.jobId,
+      AmplifyJobId: deployment.jobSummary?.jobId,
```
otaviomacedo, line 66: When is `deployment.jobSummary` not defined? And could this possibly be a breaking change?

corymhall (changes requested):
I think we should update all the integ test that use these custom resources to
have `diffAssets: true` so that we have to rerun the integ tests.

packages/@aws-cdk/aws-amplify-alpha/lib/asset-deployment-handler/handler.ts
MrArnoldPalmer: This is just a side effect of better typings in the new versions of the SDKs. My understanding is the same failure cases existed before where the response was possibly undefined but just not typed explicitly and therefore not handled.

packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/lambda-handler/sdk.ts
```diff
@@ -29,18 +29,23 @@ export function flatten(object: object): { [key: string]: any } {
   );
 }
 
+function getServiceClient(service: string): any {
+  const clientPackageName = `@aws-sdk/client-${service.toLowerCase()}`;
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const pkg = require(clientPackageName);
+    return new pkg[service]();
```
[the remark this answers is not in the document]
MrArnoldPalmer: for safety, we should. But it is actually nullable runtime wise just not in TS for whatever reason?

corymhall (approved):

PR #25906, opened 2023-06-08
feat(app-staging-synthesizer): clean up staging resources on deletion

### LINKED ISSUE #25722: (app-staging-synthesizer): clean up staging resources on deletion
### Describe the feature

When using the `AppStagingSynthesizer`, staging resources that live alongside your cdk app do not get deleted when the app gets deleted. This is because these resources are not empty, and CloudFormation cannot delete un-empty S3 buckets and ECR repos.

We solve this problem in the CDK with custom resources, but these custom resources are assets themselves that need to live in a staging bucket. This creates a chicken-and-egg problem for the `AppStagingSynthesizer`, because there's nowhere to put the cleanup custom resource asset.

### Use Case

Users using `cdk deploy` and `cdk destroy` on the `AppStagingSynthesizer` as part of development. We don't destroy the staging resources, so the user has to manage their deletion themselves. Otherwise, if you `cdk deploy` a resource with the same name, `cdk deploy` will fail.

### Proposed Solution

Potentially, we can minimize these custom resources and provide them as inline code directly in the template. If we can do that, we can clean up staging resources on deletion and remove a big friction point for users.

### Other Information

_No response_

### Acknowledgements

### CDK version used

CDK v1.83.0

mac

kaizencc:
closes #25722

----

### DIFF
--- a/packages/@aws-cdk/app-staging-synthesizer-alpha/lib/default-staging-stack.ts
+++ b/packages/@aws-cdk/app-staging-synthesizer-alpha/lib/default-staging-stack.ts
@@ -18,10 +18,26 @@ import {
   INLINE_CUSTOM_RESOURCE_CONTEXT,
 } from 'aws-cdk-lib/core';
 import { StringSpecializer } from 'aws-cdk-lib/core/lib/helpers-internal';
+import * as cxapi from 'aws-cdk-lib/cx-api';
+import { Construct } from 'constructs';
 import { BootstrapRole } from './bootstrap-roles';
 import { FileStagingLocation, IStagingResources, IStagingResourcesFactory, ImageStagingLocation } from './staging-stack';
 
 export const DEPLOY_TIME_PREFIX = 'deploy-time/';
+
+/**
+ * This is a dummy construct meant to signify that a stack is utilizing
+ * the AppStagingSynthesizer. It does not do anything, and is not meant
+ * to be created on its own. This construct will be a part of the
+ * construct tree only and not the Cfn template. The construct tree is
+ * then encoded in the AWS::CDK::Metadata resource of the stack and
+ * injested in our metrics like every other construct.
+ */
+export class UsingAppStagingSynthesizer extends Construct {
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+  }
+}
 
 /**
  * User configurable options to the DefaultStagingStack.
@@ -137,6 +153,12 @@ export class DefaultStagingStack extends Stack implements IStagingResources {
           throw new Error(`Stack ${stack.stackName} must be part of an App`);
         }
 
+        // Because we do not keep metrics in the DefaultStagingStack, we will inject
+        // a dummy construct into the stack using the DefaultStagingStack instead.
+        if (cxapi.ANALYTICS_REPORTING_ENABLED_CONTEXT) {
+          new UsingAppStagingSynthesizer(stack, `UsingAppStagingSynthesizer/${stack.stackName}`);
+        }
+
         const stackId = `StagingStack-${appId}-${context.environmentString}`;
         return new DefaultStagingStack(app, stackId, {
           ...options,
@@ -209,7 +231,10 @@ export class DefaultStagingStack extends Stack implements IStagingResources {
     super(scope, id, {
       ...props,
       synthesizer: new BootstraplessSynthesizer(),
+      analyticsReporting: false, // removing AWS::CDK::Metadata construct saves ~3KB
     });
+    // removing path metadata saves ~2KB
+    this.node.setContext(cxapi.PATH_METADATA_ENABLE_CONTEXT, false);
 
     // For all resources under the default staging stack, we want to inline custom
     // resources because the staging bucket necessary for custom resource assets
--- a/packages/@aws-cdk/app-staging-synthesizer-alpha/test/app-staging-synthesizer.test.ts
+++ b/packages/@aws-cdk/app-staging-synthesizer-alpha/test/app-staging-synthesizer.test.ts
@@ -121,6 +121,12 @@ describe(AppStagingSynthesizer, () => {
     expect(depStack.stackName).toEqual(`StagingStack-${APP_ID}`);
   });
 
+  test('stack has dummy construct for metrics', () => {
+    // WHEN
+    const dummyConstruct = stack.node.tryFindChild(`UsingAppStagingSynthesizer/${stack.stackName}`);
+    expect(dummyConstruct).toBeDefined();
+  });
+
   test('add file asset', () => {
     // WHEN
     const location = stack.synthesizer.addFileAsset({
--- a/packages/@aws-cdk/app-staging-synthesizer-alpha/test/integ.synth-default-resources.js.snapshot/manifest.json
+++ b/packages/@aws-cdk/app-staging-synthesizer-alpha/test/integ.synth-default-resources.js.snapshot/manifest.json
@@ -78,6 +78,15 @@
             "type": "aws:cdk:logicalId",
             "data": "lambdaecr2615DAF68"
           }
+        ],
+        "UsingDefaultStagingStacksynthesizedefaultresources": [
+          {
+            "type": "aws:cdk:logicalId",
+            "data": "UsingDefaultStagingStacksynthesizedefaultresources",
+            "trace": [
+              "!!DESTRUCTIVE_CHANGES: WILL_DESTROY"
+            ]
+          }
         ]
       },
       "displayName": "synthesize-default-resources"
--- a/packages/@aws-cdk/app-staging-synthesizer-alpha/test/integ.synth-default-resources.js.snapshot/tree.json
+++ b/packages/@aws-cdk/app-staging-synthesizer-alpha/test/integ.synth-default-resources.js.snapshot/tree.json
@@ -8,6 +8,14 @@
         "id": "synthesize-default-resources",
         "path": "synthesize-default-resources",
         "children": {
+          "UsingAppStagingSynthesizer--synthesize-default-resources": {
+            "id": "UsingAppStagingSynthesizer--synthesize-default-resources",
+            "path": "synthesize-default-resources/UsingAppStagingSynthesizer--synthesize-default-resources",
+            "constructInfo": {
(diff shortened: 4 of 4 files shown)


packages/@aws-cdk/app-staging-synthesizer-alpha/scripts/airlift-custom-resource-bindings.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+scriptdir=$(cd $(dirname $0) && pwd)
+customresourcedir=${scriptdir}/../../../custom-resource-bindings
```
rix0rrr, line 4: Would be nicer to use something like:

```
customresourcedir=$(node -p 'path.dirname(require.resolve("@aws-cdk/custom-resource-bindings/package.json"))')
```

here
```

File: packages/custom-resource-bindings/package.json [hunk 2]
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@aws-cdk/custom-resource-bindings",
+  "description": "CDK Custom Resources",
+  "version": "0.0.0",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
```
kaizencc, line 6: note: delete these

packages/custom-resource-bindings/scripts/build-and-minify-sources.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+
+const bindingsDir = path.join(__dirname, '..', 'lib');
+const testDir = path.join(__dirname, '..', 'test');
+
+recFolderStructure(bindingsDir);
+recFolderStructure(testDir);
+
+console.log(entryPoints);
```
kaizencc, line 26: delete

packages/custom-resource-bindings/scripts/build-and-minify-sources.js
```diff
@@ -0,0 +1,42 @@
...
+  const test = ep.includes('.test.ts');
+  esbuild.build({
+    entryPoints: [ep],
+    outfile: `${ep.slice(0, ep.lastIndexOf('.'))}.js`,
+    format: 'cjs',
+    platform: 'node',
+    minify: test ? false : true,
+    minifyWhitespace: test ? false : true,
+    minifySyntax: test ? false : true,
+    minifyIdentifiers: test ? false : true,
```
rix0rrr, line 38: Comments?

packages/custom-resource-bindings/scripts/build-and-minify-sources.js
```diff
@@ -0,0 +1,42 @@
...
+      entryPoints.push(fileOrDir);
+    }
+  }
+}
+
+const bindingsDir = path.join(__dirname, '..', 'lib');
+const testDir = path.join(__dirname, '..', 'test');
+
+recFolderStructure(bindingsDir);
+recFolderStructure(testDir);
```
rix0rrr, line 24: Why are we building the tests?

rix0rrr: I think maybe do a regular `tsc` pass, and a regular test pass, and independently `esbuild` the JavaScript into minified bundles?

packages/aws-cdk-lib/core/lib/custom-resource-provider/custom-resource-provider.ts
```diff
@@ -322,13 +319,56 @@ export class CustomResourceProvider extends Construct {
...
+    this.codeHash = codeInfo.inlineCode ? '' : codeInfo.staging.assetHash;
+  }
+
+  /**
+   * Returns information on the code for the custom resource. If the code is to
+   * be inlined, no further information is gathered. If the code is to be uploaded
+   * as an asset, the asset gets created in this function.
+   */
+  private getCodeInfo(props: CustomResourceProviderProps, stack: Stack): {
+    inlineCode: true,
```
rix0rrr, line 337: This variant of the return object could have the code in it already.

packages/aws-cdk-lib/core/lib/custom-resource-provider/custom-resource-provider.ts
```diff
@@ -322,13 +319,56 @@ export class CustomResourceProvider extends Construct {
...
-    this.codeHash = staging.assetHash;
+    this.codeHash = codeInfo.inlineCode ? '' : codeInfo.staging.assetHash;
+  }
+
+  /**
+   * Returns information on the code for the custom resource. If the code is to
+   * be inlined, no further information is gathered. If the code is to be uploaded
+   * as an asset, the asset gets created in this function.
+   */
+  private getCodeInfo(props: CustomResourceProviderProps, stack: Stack): {
```
rix0rrr, line 336: "info" ? 😝

packages/custom-resource-bindings/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@aws-cdk/custom-resource-bindings",
```
rix0rrr, line 2: `custom-resource-handlers` ?

packages/custom-resource-bindings/scripts/build-and-minify-sources.js
```diff
@@ -0,0 +1,42 @@
...
+const entryPoints = [];
+function recFolderStructure(fileOrDir) {
+  if (fs.statSync(fileOrDir).isDirectory()) {
+    const items = fs.readdirSync(fileOrDir);
+    for (const i of items) {
+      console.log(i);
+      recFolderStructure(path.join(fileOrDir, i));
+    }
+  } else {
+    if (path.extname(fileOrDir) === '.ts') {
```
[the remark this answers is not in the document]
kaizencc: I think we require the resulting file to be `index.js`

packages/custom-resource-bindings/scripts/build-and-minify-sources.js
kaizencc: theres no option for comments but they are already removed in the minified file

packages/@aws-cdk/app-staging-synthesizer-alpha/lib/default-staging-stack.ts
```diff
@@ -86,6 +87,14 @@ export interface DefaultStagingStackOptions {
...
    */
   readonly imageAssetVersionCount?: number;
+
+  /**
+   * Auto deletes objects in the staging S3 bucket and images in the
+   * staging ECR repositories.
+   *
+   * @default false
+   */
+  readonly autoDeleteStagingAssets?: boolean;
```
rix0rrr, line 97: Does this need to be an option? I would argue the whole point of this change is that this can be the default behavior, which is what everyone wants.

packages/aws-cdk-lib/scripts/airlift-custom-resource-handlers.sh
```diff
@@ -0,0 +1,21 @@
...
+
+scriptdir=$(cd $(dirname $0) && pwd)
+customresourcedir=${scriptdir}/../../custom-resource-handlers
+awscdklibdir=${scriptdir}/..
+
+list_custom_resources() {
+  for dir in $(ls $customresourcedir/lib); do
+    [ -d $customresourcedir/lib/$dir ] && echo lib/$dir/$(ls $customresourcedir/lib/$dir)
+  done
+}
```
rix0rrr, line 11: This thing is weird. What is this doing?

First of all, instead of

```
for dir in $(ls $customresourcedir/lib);
```

You could  do:

```
  for dir in $customresourcedir/lib/*;
```

Second of all, it seems to be assuming there's only ever one file inside each of the directories. If there were 2, you would get:

```
echo lib/$dir/index.js helper.js
```

Which is probably not what you intended.

So why not assume that all entrypoints must always be called `index.ts` and go:


```
list_custom_resources() {
  echo lib/*/*/index.ts
}
```

?

packages/aws-cdk-lib/scripts/airlift-custom-resource-handlers.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+scriptdir=$(cd $(dirname $0) && pwd)
+customresourcedir=${scriptdir}/../../custom-resource-handlers
```
rix0rrr, line 4: I'd rather you not make assumptions on where the SOURCE directory for this lives, but instead use Node's module resolution mechanism to find the path of this library:

```
customresourcedir=$(node -p "path.dirname(require.resolve('@aws-cdk/custom-resource-handlers/package.json'))")
```

packages/custom-resource-handlers/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+        "noImplicitReturns": true,
+        "noFallthroughCasesInSwitch": true,
+        "resolveJsonModule": true,
+        "composite": true,
+        "incremental": true
+      },
+    "include": [
+        "**/*.ts",
+        "**/*.d.ts",
+        "lib/init-templates/*/*/add-project.hook.ts"
```
rix0rrr, line 22:
```suggestion
```

packages/custom-resource-handlers/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    ],
+    "exclude": [
+        "lib/init-templates/*/typescript/**/*.ts"
```
rix0rrr, line 25:
```suggestion
```

packages/custom-resource-handlers/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@aws-cdk/custom-resource-handlers",
+  "description": "CDK Custom Resources",
```
rix0rrr, line 3:
```suggestion
  "description": "CDK Custom Resources",
  "private": true,
```

rix0rrr (approved):
Conditionally approved

packages/aws-cdk-lib/scripts/airlift-custom-resource-handlers.sh
kaizencc: a big part of it is just getting the file path after `lib` because that is how i am going to store it in the generated folder as well. but with a few modifications, this works

packages/custom-resource-handlers/package.json
kaizencc: its at the very bottom but i'll move it up

Issue #25835, opened 2023-06-02
eks: After upgrading to 2.80, imported cluster resources fail to deploy


okoskine:
### Describe the bug

After upgrading CDK to 2.80, the EKS cluster creation role no longer can be assumed by separate handlers in stacks that import the EKS cluster as explained in https://github.com/aws/aws-cdk/issues/25674.

The suggested fix of importing the whole kubectlProvider results in `Modifying service token is not allowed.` error when trying to deploy the stack.

### Expected Behavior

Kubernetes resources in the importing stack should still deploy in 2.80+ after adjusting the cluster importing as mentioned in https://github.com/aws/aws-cdk/issues/25674

### Current Behavior

cdk8s chart in the importing stack fails to deploy after the upgrade.

> StackNameRedacted: creating CloudFormation changeset...
> 2:20:09 PM | UPDATE_FAILED        | Custom::AWSCDK-EKS-KubernetesResource     | KubeClusterCommonK8sChartA8489958
> Modifying service token is not allowed.
> 
>  ❌  StackNameRedacted failed: Error: The stack named StackNameRedacted failed to deploy: UPDATE_ROLLBACK_COMPLETE: Modifying service token is not allowed.
>     at FullCloudFormationDeployment.monitorDeployment (/workdir/node_modules/aws-cdk/lib/index.js:397:10236)
>     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
>     at async Object.deployStack2 [as deployStack] (/workdir/node_modules/aws-cdk/lib/index.js:400:149977)
>     at async /workdir/node_modules/aws-cdk/lib/index.js:400:135508
> 
>  ❌ Deployment failed: Error: The stack named StackNameRedacted failed to deploy: UPDATE_ROLLBACK_COMPLETE: Modifying service token is not allowed.
>     at FullCloudFormationDeployment.monitorDeployment (/workdir/node_modules/aws-cdk/lib/index.js:397:10236)
>     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
>     at async Object.deployStack2 [as deployStack] (/workdir/node_modules/aws-cdk/lib/index.js:400:149977)
>     at async /workdir/node_modules/aws-cdk/lib/index.js:400:135508
> 
> The stack named StackNameRedacted failed to deploy: UPDATE_ROLLBACK_COMPLETE: Modifying service token is not allowed.
> 

### Reproduction Steps

Export (app A, stack A)

```typescript
[...]

PR #25908, opened 2023-06-08
fix(eks): imported clusters can't deploy manifests

pahud:
The imported eks clusters are having invalid service token and can't deploy new k8s manifests or helm charts.  This PR fixes that issue.

- [x] Update README and doc string. `functionArn` should be the custom resource provider's service token rather than the kubectl provider lambda arn. No breaking change in this PR.
- [x] Add a new integration test to ensure the imported cluster can always create manifests and helm charts.

Closes https://github.com/aws/aws-cdk/issues/25835

----


packages/aws-cdk-lib/aws-eks/README.md
```diff
@@ -713,10 +713,12 @@ By default, CDK will create a new python lambda function to apply your k8s manif
 
 ```ts
 const handlerRole = iam.Role.fromRoleArn(this, 'HandlerRole', 'arn:aws:iam::123456789012:role/lambda-role');
+const serviceToken = lambda.Function.fromFunctionName(this, 'ProviderOnEventFunc', 'ProviderframeworkonEvent-XXX').functionArn;
 const kubectlProvider = eks.KubectlProvider.fromKubectlProviderAttributes(this, 'KubectlProvider', {
   functionArn: 'arn:aws:lambda:us-east-2:123456789012:function:my-function:1',
   kubectlRoleArn: 'arn:aws:iam::123456789012:role/kubectl-role',
   handlerRole,
+  serviceToken,
```
otaviomacedo, line 721: If users will have to change their code anyway, why can't we tell them to pass the service token (as created above) to `functionArn`?

pahud: Ahhhh! Yes you are right! lol

pahud:
Hi @otaviomacedo please take a look again. Thanks.

otaviomacedo (approved):

dancmeyers:
I'm afraid this doesn't fix the reported error. See #25938

PR #25910, opened 2023-06-08
feat(autoscaling): deprecate launch configurations (under feature flag)

### LINKED ISSUE #23165: aws-autoscaling: Use launch templates instead of deprecated launch configurations
### Describe the feature

We just got an email from AWS saying launch configurations are deprecated. New accounts won't allow creating launch configurations after December 31st, 2023. No new features or instance types will be available after December 31st, 2022.

Unless given a launch template, [AutoScalingGroup](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_autoscaling.AutoScalingGroup.html) always creates a deprecated launch configuration.

### Use Case

Deploying existing CDK apps that use [AutoScalingGroup](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_autoscaling.AutoScalingGroup.html) without manually specifying a manually created launch template would fail soon. Using new instance types will be impossible without first manually creating a launch template.

### Proposed Solution

It would be great if [AutoScalingGroup](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_autoscaling.AutoScalingGroup.html) could start creating launch templates instead of launch configurations.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.52.0

N/A

colifran:
This PR focuses on supporting the deprecation of launch configurations when creating an auto scaling group. All new accounts created after December 31, 2023 will no longer be able to create launch configurations. An equivalent launch template will now be created by the AutoScalingGroup construct instead of a launch configuration.

Closes #23165

----

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudwatch/README.md
+++ b/packages/aws-cdk-lib/aws-cloudwatch/README.md
@@ -274,13 +274,6 @@ The most important properties to set while creating an Alarms are:
 
 To create a cross-account alarm, make sure you have enabled [cross-account functionality](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Cross-Account-Cross-Region.html) in CloudWatch. Then, set the `account` property in the `Metric` object either manually or via the `metric.attachTo()` method.
 
-Please note that it is **not possible** to:
-
-- Create a cross-Account alarm that as `evaluationPeriods > 1`. The reason is that the only
-  way to pass an AccountID is to use the [`Metrics` field of the Alarm resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudwatch-alarm.html#cfn-cloudwatch-alarm-metrics). If we use the `Metrics` field, the CloudWatch event that is
-  used to evaluate the Alarm doesn't have a `SampleCount` field anymore ("[When CloudWatch evaluates alarms, periods are aggregated into single data points](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Create-alarm-on-metric-math-expression.html)"). The result is that the Alarm cannot evaluate more than 1 period.
-- Create a cross-Region alarm ([source](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Cross-Account-Cross-Region.html)).
-
 ### Alarm Actions
 
 To add actions to an alarm, use the integration classes from the


colifran:
Still in process - do not review.

packages/@aws-cdk-testing/framework-integ/test/aws-ecs-patterns/test/ec2/integ.alb-ecs-service-command-entry-point.ts
```diff
@@ -4,8 +4,9 @@ import * as ecs from 'aws-cdk-lib/aws-ecs';
 import * as cdk from 'aws-cdk-lib';
 import * as integ from '@aws-cdk/integ-tests-alpha';
 import * as ecsPatterns from 'aws-cdk-lib/aws-ecs-patterns';
+import { AUTOSCALING_DISABLE_LAUNCH_CONFIG } from 'aws-cdk-lib/cx-api';
 
-const app = new cdk.App();
+const app = new cdk.App({ postCliContext: { [AUTOSCALING_DISABLE_LAUNCH_CONFIG]: false } });
```
colifran, line 9: This integ test is unsuccessful as it is, i.e., without any changes. The behavior is the same regardless of whether a launch configuration or a launch template is used. I'm continuing to investigate, but didn't want this to be a blocker since the issue is unrelated to the changes in this PR.

packages/@aws-cdk/integ-tests-alpha/lib/assertions/api-call-base.ts
```diff
@@ -125,6 +125,7 @@ export abstract class ApiCallBase extends Construct implements IApiCall {
   protected expectedResult?: string;
   protected flattenResponse: string = 'false';
   protected stateMachineArn?: string;
+  protected outputPaths: string[] = [];
```
colifran, line 128: Changes related to outputPaths were to fix a bug found where using getAtt and getAttString were not updating the outputPaths. This was an unrelated fix that I decided to address as part of this PR since it was impacting the aws-autoscaling-hooktargets integ test called integ.queue-hook

packages/aws-cdk-lib/aws-autoscaling/lib/auto-scaling-group.ts
```diff
@@ -1280,51 +1282,71 @@ export class AutoScalingGroup extends AutoScalingGroupBase implements
         throw new Error('Setting \'instanceType\' is required when \'launchTemplate\' and \'mixedInstancesPolicy\' is not set');
       }
 
+      Tags.of(this).add(NAME_TAG, this.node.path);
```
colifran, line 1285: I'd like for someone to take a second look at this. Should we only be adding this tag as part of a launch configuration? The PR this was added in mentions that this was added to support tagging of AutoScalingGroups and Security Groups so I'm thinking this isn't specific to launch configurations or launch templates, but AutoScalingGroups as a whole.

packages/@aws-cdk-testing/framework-integ/test/aws-autoscaling-hooktargets/test/integ.queue-hook.js.snapshot/integ-autoscalinghook-queue.template.json
```diff
@@ -451,39 +451,85 @@
...
+        "Value": "integ-autoscalinghook-queue/Group/LaunchTemplate"
+       }
       ]
      }
-    ],
-    "UserData": {
-     "Fn::Base64": "#!/bin/bash"
-    }
-   },
-   "DependsOn": [
```
corymhall, line 476: It's losing the dependency on the `Role`. It still has the implicit dependency on the instance profile because of the reference, but there needs to be an explicit dependency on the role.

packages/aws-cdk-lib/aws-autoscaling/lib/auto-scaling-group.ts
```diff
@@ -1280,51 +1282,71 @@ export class AutoScalingGroup extends AutoScalingGroupBase implements
...
         assumedBy: new iam.ServicePrincipal('ec2.amazonaws.com'),
       });
-
       this.grantPrincipal = this._role;
 
       if (props.ssmSessionPermissions) {
         this.role.addManagedPolicy(iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonSSMManagedInstanceCore'));
       }
 
-      const iamProfile = new iam.CfnInstanceProfile(this, 'InstanceProfile', {
```
corymhall, line 1302: We need a way to keep using this with the new `LaunchTemplate`

corymhall (changes requested):
@colifran looks great! Just a couple of minor comments.

packages/@aws-cdk-testing/framework-integ/test/aws-autoscaling-hooktargets/test/integ.queue-hook.js.snapshot/integ-autoscalinghook-queue.template.json
colifran: Good catch. I've added a launch template dependency on the role.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/ecs/integ.ec2-run-task.js.snapshot/aws-sfn-tasks-ecs-ec2-integ.template.json
```diff
@@ -497,7 +497,7 @@
     ]
    }
   },
-  "Ec2ClusterDefaultAutoScalingGroupInstanceProfileDB232471": {
+  "Ec2ClusterDefaultAutoScalingGroupLaunchTemplateProfileA74899D4": {
```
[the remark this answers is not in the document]
colifran: I came up with a solution to address this. I'm updating the integ tests to reflect this change.

packages/aws-cdk-lib/aws-autoscaling/lib/auto-scaling-group.ts
colifran: I came up with a solution to keep using this with the new `LaunchTemplate` - this will be reflected when all changes are pushed

packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
```diff
@@ -1008,4 +1010,28 @@ is not viable in some productive setups.
 | 2.84.0 | `false` | `true` |
 
 
+### @aws-cdk/aws-autoscaling:disableDefaultLaunchConfigCreation
```
corymhall, line 1013: Since we are now creating a default launch template I think we should change this name since it implies that when enabled we won't create anything.

Maybe `defaultLaunchTemplateInsteadOfLaunchConfig`. It's kind of verbose, so if you have anything better...

packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts
```diff
@@ -786,6 +817,13 @@ export class LaunchTemplate extends Resource implements ILaunchTemplate, iam.IGr
     }
   }
 
+  /**
+   * @internal
+   */
+  public _addIamInstanceProfileArn(iamInstanceProfileArn: string) {
```
corymhall, line 823: I feel like we should just bit the bullet and create a `InstanceProfile` L2. It would be a super simple L2 and would allow us to just pass it in through properties of the launch template

corymhall (changes requested):

packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts
colifran: I created an `InstanceProfile` L2 and have the functionality we want working using the L2. I need to finish testing the `InstanceProfile` L2 and then I will push these changes.

packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts
```diff
@@ -404,6 +403,20 @@ export interface LaunchTemplateProps {
...
+  /**
+   * Whether instances should have a public IP addresses associated with them.
+   *
+   * @default - Use subnet settings
+   */
+  readonly associatePublicIpAddress?: boolean;
+
+  /**
+   * The instance profile used to pass role information to EC2 instances.
+   *
```
corymhall, line 416: Add a note here about providing either this or role (and do the same for role)

packages/aws-cdk-lib/aws-iam/lib/instance-profile.ts
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * Properties of an Instance Profile
+ */
+export interface InstanceProfileProps {
+  /**
+   * The name of the role to associate with the InstanceProfile. Only one role can
+   * be assigned to an EC2 instance at a time, and all applications on the instance
+   * share the same role and permissions.
+   */
+  readonly role: IRole;
```
corymhall, line 39: I keep going back and forth on this, but I'm leaning towards making this optional and creating a role for you by default.

corymhall (changes requested):
Looking good! Just a couple minor things

packages/aws-cdk-lib/aws-iam/lib/instance-profile.ts
colifran: This makes sense. I'll get this updated.

packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts
colifran: Updated.

corymhall (approved):

PR #25920, opened 2023-06-09
fix(cloudfront): avoid to sort TTLs when using Tokens in CachePolicy

Tietew:
Closes #25795.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

Issue #25795, opened 2023-05-31
(aws-cdk-lib.aws_cloudfront): CachePolicy TTL value from SSM parameter not resolved correctly


enpatrik:
### Describe the bug

CloudFront CachePolicy min/max/default TTL settings does not work with tokenized values (like SSM parameter).

### Expected Behavior

The CachePolicy settings to be set with the value from the SSM parameter.

### Current Behavior

The SSM parameter references are not resolved correctly.

### Reproduction Steps

Example1 (only using SSM parameter for defaultTtl):
```typescript
import * as cdk from 'aws-cdk-lib';
import { Construct } from 'constructs';
import * as ssm from 'aws-cdk-lib/aws-ssm';
import * as cloudfront from 'aws-cdk-lib/aws-cloudfront';

export class BugStack extends cdk.Stack {
  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
    super(scope, id, props);

    const defaultTtl = cdk.Token.asNumber(ssm.StringParameter.valueForStringParameter(this,'/Default'))

    new cloudfront.CachePolicy(this, 'CachePolicy', {
      defaultTtl: cdk.Duration.seconds(defaultTtl),
    })
  }
}
```
Result1:
```javascript
"CachePolicy": {
 "Type": "AWS::CloudFront::CachePolicy",
 "Properties": {
  "CachePolicyConfig": {
   "DefaultTTL": 0, // Should reference SSM parameter
   "MaxTTL": 31536000,
   "MinTTL": 0,
   // ...
```

---

Example2 (using SSM parameters for all ttl values):
```typescript
const minTtl = cdk.Token.asNumber(ssm.StringParameter.valueForStringParameter(this,'/Min'))
const maxTtl = cdk.Token.asNumber(ssm.StringParameter.valueForStringParameter(this,'/Max'))
const defaultTtl = cdk.Token.asNumber(ssm.StringParameter.valueForStringParameter(this,'/Default'))

new cloudfront.CachePolicy(this, 'CachePolicy', {
    minTtl: cdk.Duration.seconds(minTtl),
    maxTtl: cdk.Duration.seconds(maxTtl),
    defaultTtl: cdk.Duration.seconds(defaultTtl),
})
```

Result:
```javascript
"CachePolicy": {
 "Type": "AWS::CloudFront::CachePolicy",
 "Properties": {
  "CachePolicyConfig": {
   "DefaultTTL": {
[...]

PR #25922, opened 2023-06-09
fix(autoscaling): AutoScalingGroup maxCapacity defaults to minCapacity when using Token

Tietew:
Related to #25920.

In `AutoScalingGroup`, `maxCapacity` defaults to `Math.max(minCapacity, 1)` even when `minCapacity` is a Token.
Because Token number is always negative number, `maxCapacity` will be set to `1` when `maxCapacity` is `undefined` and `minCapacity` is a Token.

see also https://github.com/aws/aws-cdk/issues/25795#issuecomment-1571580559

----


Tietew:
Exemption Request

This PR modifies only runtime checks.

otaviomacedo (approved):

PR #25943, opened 2023-06-12
fix(backup): validation for vault name fails when parameters are referred in the name

### LINKED ISSUE #21735: aws-backup: Attribute BackupVault.props.backupVaultName fails a validation when used with parameters
### Describe the bug

Trying to put a CF-parameter into the name of backup vault fails.

### Expected Behavior

should run (allow the usage of tokens/parameters)

### Current Behavior

fails with:
```
Error: Expected vault name to match pattern `^[a-zA-Z0-9-_]{2,50}$`
    at new BackupVault (D:\Sources\HRAR\iac\node_modules\aws-cdk-lib\aws-backup\lib\vault.js:1:2076)
    ...
```

### Reproduction Steps

Run:
```typescript
const stage = new CfnParameter(this, 'Stage', { type: 'String', description: 'Stage' });

new backup.BackupVault(this, 'BackupVault', {
    backupVaultName: `backupVault-${stage.valueAsString}`
});
```

### Possible Solution

Remove the regEx validation from https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-backup/lib/vault.ts:264 

```typescript
if (props.backupVaultName && !/^[a-zA-Z0-9\-_]{2,50}$/.test(props.backupVaultName)) {
  throw new Error('Expected vault name to match pattern `^[a-zA-Z0-9\-_]{2,50}$`');
}
```

I think the validation is misplaced here, because it should support expressions like `bla-${Token[TOKEN.285]}-bla`.

_No response_

### CDK CLI Version

2.38.1 (build a5ced21)

### Framework Version

2.38.1

### Node.js Version

v16.15.1

### OS

Windows 10

### Language

Typescript

### Language Version

4.7.4

### Other information

_No response_

tam0ri:
Validation for Backup vault name fails when parameters are referred in the name now. Current implementation simply validates vault name with regular expression described in [CFn reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-backup-backupvault.html#cfn-backup-backupvault-backupvaultname), so it does not consider the special characters for parameters. This PR solves the issue by checking props.backupVaultName is resolved.

Closes #21735

----


corymhall (approved):

jumic (approved):
Thanks. Looks good.

jumic (approved):

packages/@aws-cdk-testing/framework-integ/test/aws-backup/test/integ.backup.ts
```diff
@@ -46,9 +57,6 @@ class TestStack extends Stack {
         moveToColdStorageAfter: Duration.days(30),
         deleteAfter: Duration.days(120),
       }],
-      recoveryPointTags: {
-        stage: 'prod',
-      },
```
jogold, line 51: Why is this removed from the existing integ test?

tam0ri: I don't remember the intention at the time, but it might be by my mistake. These lines were introduced few days before I submitted this PR, so I guess that some miss operation in my local environment at the time. Anyway, thank you for your nice catching! I'll revert this change and regenerate snapshots.

jogold (approved):

scanlonp (approved):
Looks good!

Issue #25794, opened 2023-05-31
aws_batch_alpha: FargateComputeEnvironment compute_environment_name not being respected


roketworks:
### Describe the bug

When creating a `FargateComputeEnvironment` using the construct and passing in `compute_environment_name` the resulting cloudformatation does not have the `ComputeEnvironmentName` property set. 

### Expected Behavior

`ComputeEnvironmentName` property set on the resulting `AWS::Batch::ComputeEnvironment` resource in the outputted CloudFormation. 

### Current Behavior

`ComputeEnvironmentName` property is not set on the resulting `AWS::Batch::ComputeEnvironment` resource in the outputted CloudFormation. 

This only seems to affect the `FargateComputeEnvironment` construct and not others. 

### Reproduction Steps

```python
# Resulting CloudFormation will not have ComputeEnvironmentName set
self.fargate_compute_env = batch.FargateComputeEnvironment(
    self, 'FargateComputeEnv',
    compute_environment_name='fargate',
    enabled=True,
    vpc=vpc,
)

# Resulting CloudFormation will have ComputeEnvironmentName set
self.gpu_compute_env = batch.ManagedEc2EcsComputeEnvironment(
    self, 'GPUComputeEnv',
    compute_environment_name='gpu',
    enabled=True,
    maxv_cpus=196,
    vpc=vpc,
    instance_types=[
        ec2.InstanceType.of(ec2.InstanceClass.G5, ec2.InstanceSize.XLARGE2),
        ec2.InstanceType.of(ec2.InstanceClass.G4DN, ec2.InstanceSize.XLARGE4)
    ],
)

```

### Additional Information/Context

```yaml
  FargateComputeEnv155F00A9:
    Type: AWS::Batch::ComputeEnvironment
    Properties:
      Type: managed
      ComputeResources:
        MaxvCpus: 256
        SecurityGroupIds:
          - Fn::GetAtt:
              - FargateComputeEnvSecurityGroup914E8A54
              - GroupId
        Subnets:
          - Ref: VpcIsolatedSubnet1SubnetE48C5737
          - Ref: VpcIsolatedSubnet2Subnet16364B91
        Type: FARGATE
        UpdateToLatestImageVersion: true
      ReplaceComputeEnvironment: false
      State: ENABLED
      UpdatePolicy: {}
    Metadata:
      aws:cdk:path: IdentificationTrainingInfra/FargateComputeEnv/Resource
.......
  TrainingComputeEnvEDCE302F:
[...]

PR #25944, opened 2023-06-12
fix(batch): computeEnvironmentName is not set in FargateComputeEnvironment

akshaypilankar:
computeEnvironmentName property was missing in FargateComputeEnvironment due to which ComputeEnvironmentName property set on the resulting AWS::Batch::ComputeEnvironment resource in the outputted CloudFormation.
Updated managed-compute-environment to reflect the fix made in FargateComputeEnvironment.

Closes https://github.com/aws/aws-cdk/issues/25794.


comcalvi (approved):
lgtm, thanks for reopening this!

comcalvi:
Adding the exempt-integ label because the snapshot was actually wrong before and I didn't catch it, this update to the snapshot is correcting it and we don't need to change the actual test code.

akshaypilankar:
@comcalvi Thank you

PR #25950, opened 2023-06-13
fix(rds): add writer instance id to database cluster's instance ids

### LINKED ISSUE #25947: RDS: DatabaseCluster Instance Identifiers Attribute
### Describe the feature
In the new writer/reader API, only readers' IDs are returned: [aws-cdk/packages/aws-cdk-lib/aws-rds/lib/cluster.ts](https://github.com/aws/aws-cdk/blob/006b542844726afb9e52ee728e6cd14ab9f975b7/packages/aws-cdk-lib/aws-rds/lib/cluster.ts#L666)

Related issue: https://github.com/aws/aws-cdk/issues/25898

### Use Case

We need the writer's instance ID to subscribe to its events.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.82.0

macOS Ventura 13.3.1

argarinpauljohn:
Adding the writer instance's id to the instance ids being returned. Currently it only returns just the readers'.
Closes #25947.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-rds/lib/cluster.ts
+++ b/packages/aws-cdk-lib/aws-rds/lib/cluster.ts
@@ -644,6 +644,7 @@ abstract class DatabaseClusterNew extends DatabaseClusterBase {
       subnetGroup: this.subnetGroup,
       promotionTier: 0, // override the promotion tier so that writers are always 0
     });
+    instanceIdentifiers.push(writer.instanceIdentifier);
     (props.readers ?? []).forEach(instance => {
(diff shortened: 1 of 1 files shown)


corymhall (changes requested):
Good catch! Can you also add a unit test?

amscheaws:
Probably also need to update the documentation for the cluster's property? It says,
```
Identifiers of the replicas.
```
Should this be,
```
Identifiers of the writer and readers.
```
Thanks.

PR #25964, opened 2023-06-13
fix(ecr): autoDeleteImages fails on multiple repositories

kaizencc:
When setting `autoDeleteImages: true` for multiple repositories in the same stack, permissions to do the actual deleting only get added to the first one. This is because the policy statement is added inside of the `getOrCreateProvider` method, and that method ensures that the provider is only created once.

Instead, this adds the policy statement on the provider itself, regardless of whether it was created or referenced.

----


packages/aws-cdk-lib/aws-ecr/test/repository.test.ts
```diff
@@ -976,4 +976,76 @@ describe('repository', () => {
...
+                },
+                {
+                  Effect: 'Allow',
+                  Action: [
+                    'ecr:BatchDeleteImage',
+                    'ecr:DescribeRepositories',
+                    'ecr:ListImages',
+                    'ecr:ListTagsForResource',
+                  ],
+                  Resource: [
```
kaizencc, line 1025: wouldn't it be great to combine the resources of this policy together? yes. afaik this would require a global variable to keep track of all the autoDelete repos and then lazily add them in.

i didn't do it here because we don't do it in s3 and i think we should do it in both, or neither

rix0rrr (approved):

rix0rrr (approved):

PR #26024, opened 2023-06-16
chore(app-staging-synthesizer): reduce the size of DefaultStagingStack template by 5KB

kaizencc:
Since `DefaultStagingStack` is a stack that creates staging resources, it is bound to the CloudFormation [limit](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html) of 51KB. This means that there is an upper limit on how many ECR repositories (and in turn, unique images) are allowed in the CDK App. In an effort to increase this limit, this PR does the following:

- Remove path metadata to save ~2KB
- Remove `AWS::CDK::Metadata` construct to save ~3KB
- Insert a dummy `UsingDefaultStagingStack` construct to the stacks that _use_ the `DefaultStagingStack` for analytics purposes.

In all, this should decrease the size of the DefaultStagingStack template by ~5KB, or ~10% of the overall size allowed.

----


packages/@aws-cdk/app-staging-synthesizer-alpha/lib/default-staging-stack.ts
```diff
@@ -17,11 +18,24 @@ import {
...
 import { FileStagingLocation, IStagingResources, IStagingResourcesFactory, ImageStagingLocation } from './staging-stack';
 
 export const DEPLOY_TIME_PREFIX = 'deploy-time/';
 
+/**
+ * This is a dummy construct meant to signify that a stack is utilizing
+ * the DefaultStagingStack. It does not do anything, and is not meant
+ * to be created on its own.
+ */
+export class UsingDefaultStagingStack extends CfnWaitConditionHandle {
```
kaizencc, line 33: I'm ready for you to tell me what the right construct to extend is. AFAIK, we need to extend an actual CFN resource, and `WaitConditionHandle` looks like the simplest. Not sure how users react when this is added though.

rix0rrr: > AFAIK, we need to extend an actual CFN resource

We do not. An empty class that extends `Construct` is fine.

rix0rrr: I think I'd like this to represent

```suggestion
export class UsingAppStagingSynthesizer extends CfnWaitConditionHandle {
```

rix0rrr (changes requested):

rix0rrr (approved):

PR #26027, opened 2023-06-17
fix(ecs): validation for task definition fails when task-level memory is defined but container-level memory and memoryReservation are not defined with EC2 compatibility

### LINKED ISSUE #25275: ecs: task definition validation container bug

tam0ri:
Currently, validation for ECS task definition fails when task-level memory is defined but container-level memory and memoryReservation are not defined with EC2 compatibility. On the other hand, if we specify task-level memory, we can omit container-level memory and memoryReservation parameters from ECS API perspective. This PR solves the issue by skipping the validation when task-level memory is defined.

Closes #25275

----


packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
```diff
@@ -736,9 +739,11 @@ export class TaskDefinition extends TaskDefinitionBase {
       // EC2 mode validations
 
       // Container sizes
-      for (const container of this.containers) {
-        if (!container.memoryLimitSpecified) {
-          ret.push(`ECS Container ${container.containerName} must have at least one of 'memoryLimitMiB' or 'memoryReservationMiB' specified`);
+      if (!this._memory) {
```
corymhall, line 742: We should also account for the scenario where both are provided.

> If using the EC2 launch type, you must specify either a task-level memory value or a container-level memory value. If you specify both a container-level memory and memoryReservation value, memory must be greater than memoryReservation . If you specify memoryReservation , then that value is subtracted from the available memory resources for the container instance where the container is placed. Otherwise, the value of memory is used.

corymhall (changes requested):
Looks great! Just 1 comment

packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
tam0ri: Hi @corymhall, thank you for your review!

Please let me clearify your comment. What do you mean `the scenario where both are provided`? Is the scenario where both a task-level memory value and a container-level memory value are specified? Currently I don't understand what we should account for additionaly in the scenario.

In my understanding,

pattern 1: only task-level memory value is specified.
--> In that case, we don't have to check whether container-level memory value is specified.

pattern 2: only container-level memory value is specified.
--> In that case, we should check whether either memoryLimitMiB or memoryReservationMiB is specified. This validation is already implemented.

pattern 3: both a task-level memory value and a container-level memory value are specified.
--> In that case, we don't have to check whether container-level memory value is specified.

tam0ri: In my understanding, this is already validated here.
https://github.com/aws/aws-cdk/blob/f8a94d8fc3a43711c4079d83d1abf0817209fc42/packages/aws-cdk-lib/aws-ecs/lib/container-definition.ts#L490-L492

corymhall: ah yep  you are right.

corymhall (approved):

PR #26031, opened 2023-06-17
fix(elbv2): correct wrong timeout validation

### LINKED ISSUE #26023: aws-cdk-lib/aws-elasticloadbalancingv2: custom health check timeouts sould be supported for Network Load Balancer health checks
### Describe the bug

After NLB updating November 2022, We can change HealthCheckTimeoutSeconds between 2–120 seconds 

Elastic Load Balancing capabilities for application availability
https://aws.amazon.com/about-aws/whats-new/2022/11/elastic-load-balancing-capabilities-application-availability/ 
>Network Load Balancer (NLB) Health Check Improvements: NLB allows customers to define health check intervals, specify HTTP response codes that determine target health, and configure the number of consecutive health check responses before a target is either health or unhealthy. For details, see the NLB health check documentation here.

Health checks for your target groups - Elastic Load Balancing
https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html
>HealthCheckTimeoutSeconds
>
>The amount of time, in seconds, during which no response from a target means a failed health check. The range is 2–120 seconds. The default values are 6 seconds for HTTP and 10 seconds for TCP and HTTPS health checks.
>
>6 seconds for HTTP health checks and 10 seconds for TCP and HTTPS health checks.

However the latest CDK is still not updated  yet.
https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts#L283-L288
```
if (healthCheck.timeout && healthCheck.timeout.toSeconds() !== NLB_HEALTH_CHECK_TIMEOUTS[healthCheck.protocol]) {
      ret.push([
        'Custom health check timeouts are not supported for Network Load Balancer health checks.',
        `Expected ${NLB_HEALTH_CHECK_TIMEOUTS[healthCheck.protocol]} seconds for ${healthCheck.protocol}, got ${healthCheck.timeout.toSeconds()}`,
      ].join(' '));
    }
```

### Expected Behavior

The option HealthCheckTimeoutSeconds should be available.

### Current Behavior

When I tried to deploy using custom health check timeout, It cause following error.
```
$ cdk deploy

  [NlbStack/NLB/NLBListener/NLBTargetsGroup] Custom health check timeouts are not supported for Network Load Balancer health checks. Expected 10 seconds for TCP, got 2
    at validateTree (/home/ec2-user/nlb/node_modules/aws-cdk-lib/core/lib/private/synthesis.js:3:12)
    at synthesize (/home/ec2-user/nlb/node_modules/aws-cdk-lib/core/lib/private/synthesis.js:1:953)
    at App.synth (/home/ec2-user/nlb/node_modules/aws-cdk-lib/core/lib/stage.js:1:2052)
    at process.<anonymous> (/home/ec2-user/nlb/node_modules/aws-cdk-lib/core/lib/app.js:1:1448)
    at Object.onceWrapper (node:events:628:26)
    at process.emit (node:events:513:28)
    at process.emit (node:domain:489:12)
    at process.emit.sharedData.processEmitHook.installedValue [as emit] (/home/ec2-user/nlb/node_modules/@cspotcode/source-map-support/source-map-support.js:745:40)
```

### Reproduction Steps

```typescript
import { Construct } from 'constructs'
import { App, Stack, StackProps } from 'aws-cdk-lib'
import * as cdk from 'aws-cdk-lib'
import * as ec2 from 'aws-cdk-lib/aws-ec2'
import * as elbv2 from 'aws-cdk-lib/aws-elasticloadbalancingv2'
import * as elbv2Targtes from 'aws-cdk-lib/aws-elasticloadbalancingv2-targets'

export class NlbStack extends Stack {
    super(scope, id, props);

    const vpc = new ec2.Vpc(this, 'VPC', {
      maxAzs: 2,
      natGateways: 0,
      subnetConfiguration: [{
        name: 'PublicSubnet',
        subnetType: ec2.SubnetType.PUBLIC,
      }],
    })

    const securityGroup = new ec2.SecurityGroup(this, 'SecurityGroup', {
      vpc,
      allowAllOutbound: true,
    })
    securityGroup.addIngressRule(
        ec2.Peer.anyIpv4(),
        ec2.Port.tcp(80),
    )
    const bastion = new ec2.BastionHostLinux(this, 'EC2', {
      vpc,
      securityGroup,
      instanceType: ec2.InstanceType.of(
          ec2.InstanceClass.BURSTABLE4_GRAVITON,
          ec2.InstanceSize.MICRO,
      ),
    })

    const nlb = new elbv2.NetworkLoadBalancer(this, 'NLB', {
      vpc,
      internetFacing: true,
      vpcSubnets: {
        subnetType: ec2.SubnetType.PUBLIC,
      },
    })
    const nlbListener = nlb.addListener(
        'NLBListener',
        {
          port: 80
        },
    )
    nlbListener.addTargets('NLBTargets', {
      protocol: elbv2.Protocol.TCP,
      port: 80,
      targets: [
          new elbv2Targtes.InstanceTarget(
              bastion.instance,
              80
          )
      ],
      healthCheck: {
        protocol: elbv2.Protocol.TCP,
        port: '80',
        interval: cdk.Duration.seconds(6),
        timeout: cdk.Duration.seconds(2),   //  Custom health check timeouts should be supported.
        healthyThresholdCount: 2,
        unhealthyThresholdCount: 2,
      },
    })
  }
}

const app = new App();
new NlbStack(app, 'my-test-stack', {
    env: { account: '', region: '' },
});
```

### Possible Solution

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts#L283-L288

It seems that we need to modify the validation for the URL above.

_No response_

### CDK CLI Version

2.84.0 (build f7c792f)

### Framework Version

_No response_

### Node.js Version

v16.20.0

### OS

Amazon Linux 2

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

tmyoda:
Following the update of Network Load Balancer (NLB) in November 2022, the range for setting HealthCheckTimeoutSeconds for NLB is now between 2 and 120 seconds. However, the CDK has yet to be updated to reflect this change.

https://aws.amazon.com/about-aws/whats-new/2022/11/elastic-load-balancing-capabilities-application-availability/
https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html

This PR addresses this by modifying the validation, enabling us to set the HealthCheckTimeoutSeconds for the NLB.

I have modified and added the following validation in this PR:
- Ensure that the HealthCheckTimeoutSeconds is between 2 and 120.
- Ensure that HealthCheckTimeoutSeconds is not greater than HealthCheckIntervalSeconds.

Closes #26023.

----


packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts
```diff
@@ -280,11 +280,17 @@ export class NetworkTargetGroup extends TargetGroupBase implements INetworkTarge
...
-      ].join(' '));
+
+    const lowHealthCheckTimeout = 2;
+    const highHealthCheckTimeout = 120;
+    if (healthCheck.timeout) {
+      const timeoutSeconds = healthCheck.timeout.toSeconds();
+      if (timeoutSeconds < lowHealthCheckTimeout || timeoutSeconds > highHealthCheckTimeout) {
+        ret.push(`Health check timeout '${timeoutSeconds}' not supported. Must be a number between ${lowHealthCheckTimeout} and ${highHealthCheckTimeout}.`);
+      }
+      if (healthCheck.interval && healthCheck.interval.toSeconds() < timeoutSeconds) {
```
colifran, line 291: Can we double check this? The console experience doesn't allow a health check to be configured unless the interval is strictly larger than the timeout. I think this should be:

`if (healthCheck.interval && healthCheck.interval.toSeconds() <= timeoutSeconds)`

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts
```diff
@@ -280,11 +280,17 @@ export class NetworkTargetGroup extends TargetGroupBase implements INetworkTarge
...
+        ret.push(`Health check timeout '${timeoutSeconds}' must not be greater than the interval '${healthCheck.interval.toSeconds()}'`);
```
colifran, line 292: Given my above comment, let's update this to be something like,

`Health check timeout `${timeoutSeconds}` must be strictly less than the interval `${healthCheck.interval.toSeconds()}`

colifran:
@tmyoda awesome job with this. I left a few minor comments that I'd like to double check with you.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts
```diff
@@ -280,11 +280,17 @@ export class NetworkTargetGroup extends TargetGroupBase implements INetworkTarge
...
     }
-    if (healthCheck.timeout && healthCheck.timeout.toSeconds() !== NLB_HEALTH_CHECK_TIMEOUTS[healthCheck.protocol]) {
-      ret.push([
-        'Custom health check timeouts are not supported for Network Load Balancer health checks.',
-        `Expected ${NLB_HEALTH_CHECK_TIMEOUTS[healthCheck.protocol]} seconds for ${healthCheck.protocol}, got ${healthCheck.timeout.toSeconds()}`,
-      ].join(' '));
+
+    const lowHealthCheckTimeout = 2;
+    const highHealthCheckTimeout = 120;
+    if (healthCheck.timeout) {
```
colifran, line 286: Could we update the timeout property of the HealthCheck interface to reflect this new range?

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts
tmyoda: I put `healthCheck.interval.toSeconds() < timeoutSeconds` because I confirmed that I can make Target groups which is Timeout = Interval.


However I checked the console and it's true and I think your suggestion (`healthCheck.interval.toSeconds() <= timeoutSeconds`) would be more natural.

target-group-100


I will fix this.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts
tmyoda: I will update this error message.
Thank you very much for the kind review!

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts
tmyoda: I understand that you mentioned the interface below.

https://github.com/aws/aws-cdk/blob/327035644b94ccee354cf34e1665f6eedfbf939d/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-target-group.ts#L110-L114

Please let me know if my understanding is wrong.
Thanks

colifran: Yep, you got it! I looked at the timeout description in that interface and it looks like it is outdated. This PR would be a good time to update this and align with  the CloudFormation documentation. I think we could do something like this for the description: "The amount of time during which no response from a target means a failed health check. Must be 2 to 120 seconds". The default flag would also need to be updated. I think for this we could do the following:

`@default 6 seconds if the protocol is HTTP, 5 seconds if protocol is 'GENEVE', 30 seconds if target type is 'lambda', 10 seconds for TCP, TLS, or HTTPS`.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html#cfn-elasticloadbalancingv2-targetgroup-healthchecktimeoutseconds

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts
tmyoda: Sorry, I realized that the change `healthCheck.interval.toSeconds() <= timeoutSeconds)` could potentially be affect a customer who is using `Timeout = Interval`

For instance, if a customer is using `HealthCheckIntervalSeconds = 10`, for HTTPS, HealthCheckTimeoutSeconds would be 10 (default value). This is a case of `Timeout = Interval`.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts
tmyoda: Thanks a lot for your kind suggestion!
I will align with the CFn docs

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts
colifran: Good catch. I agree.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-target-group.ts
tmyoda: Based on your comment above, I will not update the error message.

tmyoda:
Hi @colifran 
I have closed all comments. Could you please review it again?
Thank you.

colifran (approved):
Looks great! Thank you for your contribution.

PR #26040, opened 2023-06-19
fix(lambda): add instrument handler option to adotInstrumentation to support python lambda functions

### LINKED ISSUE #24666: lambda: AdotLambdaExecWrapper is incorrect for Python
### Describe the bug

The ADOT Lambda auto-instrumentation does not work for python function, as the exec wrapper does not match the expected values for python adot

### Expected Behavior

As per the ADOT documentation on auto-instumenting a python lambda function, the `AWS_LAMBDA_EXEC_WRAPPER` envar should be set to the value `/opt/otel-instrument`.

### Current Behavior

The envar is set to one of 3 values, per the enum:
```typescript
  REGULAR_HANDLER = '/opt/otel-handler',
  PROXY_HANDLER = '/opt/otel-proxy-handler',
  STREAM_HANDLER = '/opt/otel-stream-handler',
```

### Reproduction Steps

```typescript
new lambda.Function(this, "MyFunction", {
  runtime: lambda.Runtime.PYTHON_3_9,
  handler: "index.handler",
  code: lambda.Code.fromAsset(path.join(__dirname, "lambda-handler")),
  adotInstrumentation: {
    layerVersion: lambda.AdotLayerVersion.fromPythonSdkLayerVersion(
      lambda.AdotLambdaLayerPythonSdkVersion.LATEST
    ),
    execWrapper: lambda.AdotLambdaExecWrapper.REGULAR_HANDLER,
  },
});
```

### Possible Solution

* Provide an enum value that works for python (or a way to provide a custom value to match the type)
* Detect python functions and just set the value to be `/opt/otel-instrument`
* Work with ADOT to make the python value match other languages

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.AdotInstrumentationConfig.html

### CDK CLI Version

2.69

### Framework Version

_No response_

### Node.js Version

n/a

### OS

n/a

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

randyridgley:
Small change to add an enum value to AdotLambdaExecWrapper. The python layer for adot requires the path *'/opt/otel-instrument'* which differs from the layer needed for other languages which is *'/opt/otel-handler'*. In the link below in the section **Enable auto-instrumentation for your Lambda function** step 5 requires the *'/opt/otel-instrument'* path. The existing work around is to provide a lambda layer yourself and the appropriate env variable for *AWS_LAMBDA_EXEC_WRAPPER* without using the adotInstrumentation parameters. This fix simplifies dev requirements.

[Python Lambda Otel](https://aws-otel.github.io/docs/getting-started/lambda/lambda-python)

Closes #24666.

----


randyridgley:
I need clarification. The aws-lambda [README](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-lambda/README.md) has a section on [Adot Integration Tests](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-lambda/README.md#:~:text=the%20integration%20test) but the file does not exist. I'm also not sure if an integration test is required on this update based on the nature of the change. If this all looks good I would ask for an `Exemption Request` for this PR. Please advise.

randyridgley:
I decided to add an integration test and updated the README but for the life of me cannot run the integration test locally. Can someone help me understand what the command would be to build up the integration testing runner and how to run it. Do integration tests belong in @aws-cdk-testing/framework-integ or in the packages/aws-cdk-lib/aws-lambda/test directory? I see integ.X in both places.

packages/aws-cdk-lib/aws-lambda/test/integ.adot.ts
```diff
@@ -0,0 +1,27 @@
+import * as path from 'path';
```
corymhall, line 1: You can remove this file. The test should go in `framework-integ`.

packages/aws-cdk-lib/aws-lambda/lib/adot-layers.ts
```diff
@@ -184,6 +184,11 @@ export enum AdotLambdaExecWrapper {
    * propagation for HTTP requests.
    */
   STREAM_HANDLER = '/opt/otel-stream-handler',
+
+  /**
+   * Wrapping python lambda handlers see https://aws-otel.github.io/docs/getting-started/lambda/lambda-python
+   */
+  INSTRUMENT_HANDLER = '/opt/otel-instrument',
```
corymhall, line 191: This is so unfortunate 😒.

From the documentation it looks like `/opt/otel-instrument` is the only possible
value for python? If that is the case let just not even add an option and set
the value here based on the runtime.

https://github.com/aws/aws-cdk/blob/53d24f1a1357410746052bbf25b5b89b985f7b80/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L1161-L1161

corymhall (changes requested):

packages/aws-cdk-lib/aws-lambda/lib/adot-layers.ts
randyridgley: Not sure I follow on this one. The `this._layers.map` could be any number of layers even beyond the adot ones. What exactly would I filter on? I could maybe validate in `configureAdotInstrumentation` check for the `AdotLayerVersion` and if it's python set the `AWS_LAMBDA_EXEC_WRAPPER` env variable to `opt/otel-instrument`. Is that what you were thinking? I like either forcing the env variable or throwing an error if they select the wrong handler.

packages/aws-cdk-lib/aws-lambda/test/integ.adot.ts
randyridgley: Removed

packages/aws-cdk-lib/aws-lambda/lib/adot-layers.ts
randyridgley: In the `configureAdotInstrumentation` function I could do something like below to check the layer and throw and error if it's not `INSTRUMENT_HANDLER`. I would like to keep ADOT work separate from the generic lambda layer/env work. Or instead of throwing the error I could swap to the instrument handler but I worry about an end user setting a value and us swapping it out and what ramifications it could have in the future. The throwing of the error would be clear and it's a one time change.

```
    if (props.adotInstrumentation.layerVersion instanceof AdotLambdaLayerPythonSdkVersion
      && props.adotInstrumentation.execWrapper != AdotLambdaExecWrapper.INSTRUMENT_HANDLER) {
      throw new Error('Python Adot Lambda layer requires AdotLambdaExecWrapper.INSTRUMENT_HANDLER');
    }
```
Thoughts?

randyridgley: I decided to go with throwing an error when a python function is using the Adot layer when not using the EXEC_WRAPPER INSTRUMENT_HANDLER. I wanted to keep the logic for all adot within the configuration function `configureAdotInstrumentation` to isolate that work. This seemed like the best approach going forward. This last commit is ready for review.

corymhall (approved):

corymhall (dismissed):

randyridgley:
I had to kick off the code build project again cause it was failing in an area I didn't touch. I reviewed the failure and couldn't see exactly why it was failing. Updating the branch fixed the codebuild failure.

corymhall (approved):

PR #26049, opened 2023-06-20
feat(lambda): propagate tags to log group

### LINKED ISSUE #17533: lambda: allow propagating Function tags to LogGroup
### Description

When a lambda function is created via CDK, its LogGroup is not created right away. Instead it typically happens the first time the lambda executes. Consequently, the LogGroup does not have any of the function's tags applied to it. It would be nice if there were a way to mark the function to propagate tags to the LogGroup, perhaps via some custom resource. Ideally, this would be exposed as a mutable property so it can be toggled by an Aspect.

### Use Case

We want to be able to easily find LogGroups corresponding to particular lambda function tags via the resource tagging API. In particular, this will allow us to easily find and delete these LogGroups when they are no longer needed.

### Proposed Solution

Add a flag to the aws_lambda.Function construct indicating whether tags should be propagated. If so, create a custom resource to create the LogGroup if necessary, and then adjust its tags. It should account for adding new tags, deleting old tags, and modifying existing tags. It may be possible to combine this custom resource with the one that manages the LogGroup retention.

### Other information

_No response_

### Acknowledge

colifran:
This PR implements the ability to allow tags to propagate to the lambda log group via an optional property called 'propagateTagsToLogGroup' that can be set when a lambda function is configured. This functionality was implemented using the existing custom resource setup to allow users to set a custom log retention for a lambda log group.

Closes #17533

----


packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts
```diff
@@ -121,6 +121,64 @@ async function setRetentionPolicy(logGroupName: string, region?: string, options
...
+  // The same as in createLogGroupSafe(), here we could end up with the race
+  // condition where a log group is either already being created or its retention
+  // policy is being updated. This would result in an OperationAbortedException,
+  // which we will try to catch and retry the command a number of times before failing
+  let retryCount = options?.maxRetries == undefined ? 10 : options.maxRetries;
+  const delay = options?.retryOptions?.base == undefined ? 10 : options.retryOptions.base;
+  do {
+    try {
+      const cloudwatchlogs = new AWS.CloudWatchLogs({ apiVersion: '2014-03-28', region, ...options });
+      const tagsOnLogGroup = (await cloudwatchlogs.listTagsLogGroup({ logGroupName }).promise()).tags ?? {};
```
corymhall, line 142: It looks like `ListTagsLogGroup` is [on the path to
deprecation](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CloudWatchLogs.html#listTagsLogGroup-property)
and they recommend using `ListTagsForResource`

Same thing for `tagLogGroup` and `untagLogGroup`

packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts
```diff
@@ -121,6 +121,64 @@ async function setRetentionPolicy(logGroupName: string, region?: string, options
...
+        }
+        tagsKeys.push(tag.Key);
+      }
+
+      const tagsToDelete = tagsOnLogGroup
+        ? Object.keys(tagsOnLogGroup).filter(tag => !tagsKeys.includes(tag))
+        : [];
+
+      if (Object.keys(tagsToSet).length > 0) {
+        await cloudwatchlogs.tagLogGroup({ logGroupName, tags: tagsToSet }).promise();
```
corymhall, line 158: It looks like you can associate as many as 50 tags with a CloudWatch Logs
resource. We should handle this somehow. We could probably start off by just
throwing a validation error in the construct if > 50. I'm not sure how likely
that is so if we receive issue then we could handle > 50 at that point (I would
guess it's not that likely)

packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts
```diff
@@ -138,7 +196,11 @@ export async function handler(event: AWSLambda.CloudFormationCustomResourceEvent
       // Act on the target log group
       await createLogGroupSafe(logGroupName, logGroupRegion, retryOptions);
       await setRetentionPolicy(logGroupName, logGroupRegion, retryOptions, parseInt(event.ResourceProperties.RetentionInDays, 10));
+      if (event.ResourceProperties.PropagateTags === 'true') {
```
corymhall, line 199: We also need to handle the `event.RequestType === 'Delete`.

Here is a scenario:
I create a Lambda Function and set propagate tags
```ts
new Function(this, 'Function', { propagateTags: true });
```

Later I decide that I don't want to propagate tags anymore.
```ts
new Function(this, 'Function', { propagateTags: false });
```

This will remove the log retention custom resource which will delete the log
group.

Honestly I think this might also be a bug in the existing implementation. We
should only be destroying the log group if the lambda function is being
destroyed.

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -934,15 +934,80 @@ See [the AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/invocat
 Lambda functions automatically create a log group with the name `/aws/lambda/<function-name>` upon first execution with
 log data set to never expire.
 
-The `logRetention` property can be used to set a different expiration period.
+The `logRetention` property can be used to set a different expiration period. In addition to this, `Tags` applied to a lambda `Function` can be propagated to the associated `logGroup` by setting the `propagateTagsToLogGroup` property to true on the `Function` construct. `Tags` that don't already exist on the `logGroup` for the `Function` will be added as new `Tags` when `propagateTagsToLogGroup` is set to true. As an example, consider the following `Function` and assume that the associated `logGroup` has no `Tags`:
+
```
corymhall, line 938: I would also add an example of adding tags specifically to the log group or
excluding tags from the log group. i.e. `Tags.of(function).add('env', 'beta', {
    excludeResourceTypes: ['Custom::LogsRetention']})`

packages/aws-cdk-lib/aws-logs/lib/log-retention.ts
```diff
@@ -73,13 +80,14 @@ export interface LogRetentionRetryOptions {
...
  */
-export class LogRetention extends Construct {
-
+export class LogRetention extends Construct implements cdk.ITaggable {
   /**
    * The ARN of the LogGroup.
    */
   public readonly logGroupArn: string;
 
+  public readonly tags: cdk.TagManager = new cdk.TagManager(cdk.TagType.KEY_VALUE, 'Custom::LogRetention');
```
corymhall, line 89: I'm thinking we might want to change the type here from `Custom::LogRetention`
to `AWS::Logs::LogGroup` since we are tagging the log group and not the custom
resource itself. It would also allow the user to more intuitively filter for log
groups when tagging.

packages/@aws-cdk-testing/framework-integ/test/aws-logs/test/integ.log-retention.ts
```diff
@@ -1,21 +1,27 @@
...
+
+    new LogRetention(this, 'MyFirstLambda', {
+      logGroupName: 'logRetentionLogGroup1',
       retention: RetentionDays.ONE_DAY,
       removalPolicy: RemovalPolicy.DESTROY,
+      propagateTags: false,
     });
 
-    new LogRetention(this, 'MyLambda2', {
+    // the order here is specifically setup to make sure the resource provider will be granted
```
corymhall, line 18: I don't understand this comment?

packages/@aws-cdk-testing/framework-integ/test/aws-logs/test/integ.log-retention.ts
```diff
@@ -1,21 +1,27 @@
...
 import { IntegTest } from '@aws-cdk/integ-tests-alpha';
 import { LogRetention, RetentionDays } from 'aws-cdk-lib/aws-logs';
 
 class LogRetentionIntegStack extends Stack {
   constructor(scope: App, id: string, props?: StackProps) {
     super(scope, id, props);
 
-    new LogRetention(this, 'MyLambda', {
-      logGroupName: 'logRetentionLogGroup',
+    Tags.of(this).add('env', 'prod');
```
corymhall, line 9: We should add assertions for both integ tests to assert that the tags have been
added.

corymhall (changes requested):
Overall it's looking pretty good!

packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts
```diff
@@ -121,6 +121,64 @@ async function setRetentionPolicy(logGroupName: string, region?: string, options
...
+ * @param region the region of the log group
+ * @param options CloudWatch API SDK options
+ */
+async function setLogGroupTags(logGroupName: string, tags: AWS.CloudWatchLogs.Tags[], region?: string, options?: SdkRetryOptions) {
+  // The same as in createLogGroupSafe(), here we could end up with the race
+  // condition where a log group is either already being created or its retention
+  // policy is being updated. This would result in an OperationAbortedException,
+  // which we will try to catch and retry the command a number of times before failing
+  let retryCount = options?.maxRetries == undefined ? 10 : options.maxRetries;
+  const delay = options?.retryOptions?.base == undefined ? 10 : options.retryOptions.base;
```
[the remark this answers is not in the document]
colifran: I reused a lot of the existing logic from the other functions that the handler uses to configure log groups. I'll make sure to update this in all of them.

packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts
```diff
@@ -121,6 +121,64 @@ async function setRetentionPolicy(logGroupName: string, region?: string, options
...
+ * @param tags the tags to propagate to the log group
+ * @param region the region of the log group
+ * @param options CloudWatch API SDK options
+ */
+async function setLogGroupTags(logGroupName: string, tags: AWS.CloudWatchLogs.Tags[], region?: string, options?: SdkRetryOptions) {
+  // The same as in createLogGroupSafe(), here we could end up with the race
+  // condition where a log group is either already being created or its retention
+  // policy is being updated. This would result in an OperationAbortedException,
+  // which we will try to catch and retry the command a number of times before failing
+  let retryCount = options?.maxRetries == undefined ? 10 : options.maxRetries;
```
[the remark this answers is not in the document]
colifran: I reused a lot of the existing logic from the other functions that the handler uses to configure log groups. I'll make sure to update this in all of them.

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -934,15 +934,80 @@ See [the AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/invocat
...
+const lambdaFunction = new lambda.Function(this, 'MyFunction', {
+  code: new lambda.InlineCode('exports.handler = (event) => console.log(JSON.stringify(event));'),
+  handler: 'index.handler',
+  runtime: lambda.Runtime.NODEJS_14_X,
+  logRetention: logs.RetentionDays.ONE_MONTH,
+  functionName: 'MyFunction',
+  propagateTagsToLogGroup: true,
+});
+cdk.Tags.of(oneMonthFunction).add('env', 'beta');
```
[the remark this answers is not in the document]
colifran: Good catch. Updated. I will push this change with the other changes.

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -934,15 +934,80 @@ See [the AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/invocat
...
+
+const lambdaFunction = new lambda.Function(this, 'MyFunction', {
+  code: new lambda.InlineCode('exports.handler = (event) => console.log(JSON.stringify(event));'),
+  handler: 'index.handler',
+  runtime: lambda.Runtime.NODEJS_14_X,
+  logRetention: logs.RetentionDays.ONE_MONTH,
+  functionName: 'MyFunction',
+  propagateTagsToLogGroup: true,
+});
+cdk.Tags.of(oneMonthFunction).add('env', 'prod');
```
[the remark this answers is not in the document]
colifran: Good catch. Updated. I will push this change with the other changes.

packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts
colifran: I'll make sure this gets updated. I think we could just pass the log group ARN necessary for the resource version as part of the event passed to the lambda function. Shouldn't be too difficult to update.

packages/aws-cdk-lib/aws-logs/lib/log-retention.ts
colifran: I've updated this to be `AWS::Logs::LogGroup`

packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts
colifran: I've updated this to use tagResource and have updated the condition to check for a length greater than 0 and less than or equal to 50.

packages/aws-cdk-lib/aws-lambda/README.md
colifran: Good idea. I've added an example of this in the README.

packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts
colifran: I think this is covered by the following code:
https://github.com/aws/aws-cdk/blob/5f0d7c8d86cddb02c74581493f41283d6e4d4bcf/packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts#L218-L222
The log group will only be deleted if a delete request is made and the `RemovalPolicy` is destroy. I created the scenario you mentioned and the `LogGroup` was retained.

packages/@aws-cdk-testing/framework-integ/test/aws-logs/test/integ.log-retention.ts
colifran: I've added assertions to both log-retention integ tests.

mrgrain (approved):

mrgrain (changes requested):
I want to solve the issues of deleted log groups before me merge this.

Issue #18290, opened 2022-01-06
(aws-lambda-python): respect BUNDLING_STACKS config to skip python bundling


relm923:
### Description

Similar to the NodeJs Lambda, Python Lambda packaging should respect the `BUNDLING_STACKS` value and skip bundling appropriately. Currently Python Lambda ignores the config

### Use Case

Docker bundling of lambdas is slow and resource intensive. Its valuable to be able to skip packaging during snapshot tests and other commands that do not need the full bundling

### Proposed Solution

Match NodeJs lambda packaging, respect the `BUNDLING_STACKS` config and skip Docker bundling

### Acknowledge

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

corymhall:
As of the latest CDK version (`2.4.0`/`1.38.0`) this behavior in `PythonFunction` is the same as it is for `NodejsFunction`. You should no longer see bundling for a stack if bundling is skipped. Before `2.4.0` the bundling for `PythonFunction` actually occurred during the docker build (as part of `DockerImage.fromBuild()`) which is why bundling was never skipped.

You will still see the docker image being built every time, but it should be much faster than before since bundling is not occurring as part of the build. This is also the current behavior for all of the constructs that use Docker bundling.

I think it would be great if we could skip the docker build as well. We can repurpose this issue to track that if you want.

relm923:
Thanks for the clarification @corymhall. We are in fact seeing what you described. We leverage `NodeJsFunction` option for local/non-docker builds which allows us to fully skip bundling there.

Having the ability to fully skip the docker build using `BUNDLING_STACKS` and/or supporting building completely outside of Docker would be great

corymhall:
OK maybe I'll repurpose this issue to track allowing local bundling for `PythonFunction` and I'll use #14747 to track skipping the docker build altogether.

relm923:
Sounds goo!

relm923:
@corymhall just to clarify - even with `aws-cdk-lib v2.4.0` and `@aws-cdk/aws-lambda-python-alpha v2.4.0-alpha.0` we're still seeing the Docker image being pulled, `pipenv` being installed and all the lambda dependencies when running:
```sh
CDK_CONTEXT_JSON='{\"aws:cdk:bundling-stacks\":[\"NONE\"]}' cdk synth
```
Is this expected?

corymhall:
[...]

PR #26062, opened 2023-06-20
feat(lambda-python): add local bundling option

polothy:
Docker bundling is slow, but sometimes necessary. This just gives the end user the option to bundle locally if they want to, but still have the option to fallback to the Docker bundling.

Related to #18290 - but I'm not sure if it closes it.

----


polothy:
Not sure if I'll get around to integration tests for adding a pass-through property 😞

polothy:
Clarification Request

I got as far as running `yarn integ --update-on-failed test/integ.function.js` but I got worried about that it was doing to the AWS account and Ctr-C'ed out.

```
```

I don't have all those regions bootstrapped either.

corymhall (changes requested):
@polothy thanks for getting this started! For the python local bundling feature we need to provide the bundling logic rather than expecting the user to provide it. Something like what is done with the [lambda-nodejs](https://github.com/aws/aws-cdk/blob/939dc84b2eb18e9bcab3a0af95695cda151ee8a7/packages/aws-cdk-lib/aws-lambda-nodejs/lib/bundling.ts#L167-L169) construct.

polothy:
I wasn't planning to take this PR that far.  We can close it.

PR #26082, opened 2023-06-22
feat(opensearchservice): support for MultiAZWithStandBy (under feature flag)

### LINKED ISSUE #26026: Amazon OpenSearch Service: High Level Constructs For OpenSearch MultiAZWithStandBy Feature
### Describe the feature

The OpenSearch Team has recently launched MultiAZWithStandby Feature for OpenSearch Domain. [Feature Documentation Link](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/managedomains-multiaz.html#:~:text=around%20these%20configurations.-,Multi%2DAZ%20with%20Standby,-Multi%2DAZ%20with).  According to the CDK docs currently there is no high level CDK construct for this feature.[CDK Docs link for the MultiAZWIthStandbyFeature](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_opensearchservice.CfnDomain.ClusterConfigProperty.html#:~:text=OpenSearch%20Service%20.-,multiAzWithStandbyEnabled%3F,-Type%3A%20boolean)

### Use Case

The general recommendation is to use high level constructs and due to lack of High level construct for this feature, the CDK template needs to be migrated to use CFN constructs if there is a need to Create an OpenSearchDomain with MultiAZWithStandBy. This serves as a hinderance to the adoption of the MultiAZWithStandBy feature for AWS OpenSearch Domains.

### Proposed Solution

Add the MultiAZWithStandByAttribute to the capacityConfig object [CDK Doc for CapacityConfig Attribute](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_opensearchservice.CapacityConfig.html) . 
 A MultiAZWithStandBy based OpenSearchDomain created using CDK high level constructs should look like
const domain = new Domain(this, 'Domain', {
  version: EngineVersion.OPENSEARCH_1_0,
  capacity: {
    masterNodes: 3,
    **multiAZWithStandbyEnabled: true (possible values: [true,false])**
  },
});

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.84.0

macOS Ventura 13.4

lpizzinidev:
This fix adds support for the [`MultiAZWithStandbyEnabled`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-opensearchservice-domain-clusterconfig.html#:~:text=%3A%20No%20interruption-,MultiAZWithStandbyEnabled,Update%20requires%3A%20No%20interruption,-WarmCount) flag to the [`CapacityConfig`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_opensearchservice.CapacityConfig.html) interface.

If enabled, the `ENABLE_OPENSEARCH_MULTIAZ_WITH_STANDBY` feature flag set the default value of `MultiAZWithStandbyEnabled` to `true`

Closes #26026.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-opensearchservice/README.md
+++ b/packages/aws-cdk-lib/aws-opensearchservice/README.md
@@ -381,10 +381,10 @@ The domain can be configured to use [multi-AZ with standby](https://docs.aws.ama
 
 ```ts
 const domain = new Domain(this, 'Domain', {
-  version: opensearch.EngineVersion.OPENSEARCH_1_3,
+  version: EngineVersion.OPENSEARCH_1_3,
   ebs: {
     volumeSize: 10,
-    volumeType: EbsDeviceVolumeType.GENERAL_PURPOSE_SSD_GP3,
+    volumeType: ec2.EbsDeviceVolumeType.GENERAL_PURPOSE_SSD_GP3,
   },
   zoneAwareness: {
     enabled: true,


lpizzinidev:
Exemption Request.
Let me know if I should make changes to integration tests.

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
```diff
@@ -1556,6 +1566,7 @@ export class Domain extends DomainBase implements IDomain, ec2.IConnectable {
           : undefined,
         instanceCount,
         instanceType,
+        multiAzWithStandbyEnabled,
```
corymhall, line 1569:
```suggestion
        multiAzWithStandbyEnabled: props.capacity?.multiAzWithStandbyEnabled,
```

I don't think there is any reason to explicitly set this to false.

corymhall (changes requested):
We do need an integration test for this to ensure that we can create a domain
with this enabled (I don't think it will create currently).

I think we will also need a README entry that covers the requirements and
migration. Ideally we probably want a feature flag that will allow us to set
this to true by default (unless there is a reason that this shouldn't be the
default).

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
lpizzinidev: I've added a feature flag that defaults `multiAzWithStandbyEnabled` to `true`).
This is causing some integration tests to fail.
Let me know if the implementation is correct so I can proceed to fix the failing tests.

AmanRajAWS:
@lpizzinidev hey what are the current issues that you are blocked upon please let me know if you need any help with getting unblocked

mrgrain (dismissed):

mrgrain (approved):

PR #26102, opened 2023-06-23
feat(integ-tests): new HttpApiCall method to easily make http calls

corymhall:
This PR adds a new `httpApiCall` method (and associated construct) that makes it easy to make assertions against the results of an http call.

I've updated a couple of existing integration tests to use this new method.

----


packages/@aws-cdk-testing/framework-integ/test/aws-elasticloadbalancingv2-targets/test/integ.lambda-target.ts
```diff
@@ -39,5 +42,16 @@ def handler(event, context):
...
 
 const app = new App();
-new TestStack(app, 'TestStack');
-app.synth();
+const testCase = new TestStack(app, 'TestStack');
+const integ = new IntegTest(app, 'integ-test', {
+  testCases: [testCase],
+});
+
+const call = integ.assertions.httpApiCall(`http://${testCase.lb.loadBalancerDnsName}`, { });
```
mrgrain, line 50: This is very cool! 🚀

mrgrain (approved):

PR #26110, opened 2023-06-25
fix(sns-subscriptions): SQS queue encrypted by AWS managed KMS key is allowed to be specified as subscription and dead-letter queue

### LINKED ISSUE #19796: sns_subscriptions: Disallow SQS with AWS managed KMS key subscribing to SNS
### Describe the bug

According to https://aws.amazon.com/premiumsupport/knowledge-center/sns-topic-sqs-queue-sse-kms-key-policy/ , it's not possible to have SNS -> SQS when the SQS queue uses an AWS managed KMS key for encryption, but CDK makes it easy to create such a subscription.

### Expected Behavior

Either:

- subscriptions created via CDK to Just Work
- subscriptions that won't work (for obvious, known-at-synth-time reasons) to be flagged as invalid during synth, so I can fix it immediately

### Current Behavior

SNS fails to send any messages to SQS, with logged error messages like:

```json
{
  "notification": {...}
  "delivery": {
    "providerResponse": "{\"ErrorCode\":\"KMS.AccessDeniedException\",\"ErrorMessage\":\"null (Service: AWSKMS; Status Code: 400; Error Code: AccessDeniedException; Request ID: bb4fbbce-7269-4a16-956a-51927600b6a0; Proxy: null)\",\"sqsRequestId\":\"Unrecoverable\"}"
    ...
  }
}
```

### Reproduction Steps

```typescript
import {
  aws_sqs as sqs,
  aws_sns as sns,
  aws_sns_subscriptions as snsSubscriptions,
} from 'aws-cdk-lib';

class Stack extends cdk.Stack {
    super(scope, id, props);

    const t = new sns.Topic(this, 'T', {
      topicName: 't',
    });
    const q = new sqs.Queue(this, 'Q', {
      queueName: 'q',
      encryption: sqs.QueueEncryption.KMS_MANAGED,
    });
    const dlc = new sqs.Queue(this, 'DLQ', {
      queueName: 'dlq',
      encryption: sqs.QueueEncryption.KMS_MANAGED,
    });
    // this subscription will never do anything, nor will the DLQ
    t.addSubscription(
      new snsSubscriptions.SqsSubscription(q, {
        deadLetterQueue: dlq,
      })
    );
  }
}

const app = new cdk.App();
new Stack(app, 's', {});
```

(I haven't tested this reduced form, sorry)

### Possible Solution

It would be nice for CDK to emit an error during synth, since this subscription is never going to work, and thus it's better to tell the user immediately than force them to debug in a live environment.

_No response_

### CDK CLI Version

2.19.0 (build e0d3e62)

### Framework Version

2.19.0

### Node.js Version

v14.19.1

### OS

macOS

### Language

Typescript

### Language Version

4.6.3

### Other information

_No response_

tam0ri:
To send message to SQS queues encrypted by KMS from SNS, we need to grant SNS service-principal access to the key by key policy. From this reason, we need to use customer managed key because we can't edit key policy for AWS managed key. However, CDK makes it easy to create such a non-functional subscription.

To prevent CDK from making such a subscription, I added the validation which throw an error when SQS queue encrypted by AWS managed KMS key is specified as subscription or dead-letter queue.

Closes #19796

----


tam0ri:
Exemption Request because this change includes only additional validations. Please let me know if adding integration test is needed.

packages/aws-cdk-lib/aws-sns-subscriptions/lib/sqs.ts
```diff
@@ -38,6 +38,18 @@ export class SqsSubscription implements sns.ITopicSubscription {
     }
     const snsServicePrincipal = new iam.ServicePrincipal('sns.amazonaws.com');
 
+    // if the queue is encrypted by AWS managed KMS key (alias/aws/sqs),
+    // throw error message
+    if ((this.queue.node.defaultChild as sqs.CfnQueue).kmsMasterKeyId == 'alias/aws/sqs') {
```
corymhall, line 43: We also need to handle the case where the queue is imported `IQueue` in which
case there will not be a default child.

I think we may want to add a new attribute to the `IQueue`, something like
`encryptionType` that holds the type of encryption.

packages/aws-cdk-lib/aws-sns-subscriptions/lib/sqs.ts
```diff
@@ -38,6 +38,18 @@ export class SqsSubscription implements sns.ITopicSubscription {
...
+      throw new Error('SQS queue encrypted by AWS managed KMS key cannot be used as SNS subscription');
+    }
+
+    // if the dead-letter queue is encrypted by AWS managed KMS key (alias/aws/sqs),
+    // throw error message
+    if (this.props.deadLetterQueue && (this.props.deadLetterQueue.node.defaultChild as sqs.CfnQueue).kmsMasterKeyId == 'alias/aws/sqs') {
```
corymhall, line 49: Same comment as above

corymhall (changes requested):
Looks good! Just 1 comment

packages/aws-cdk-lib/aws-sns-subscriptions/lib/sqs.ts
tam0ri: Hi @corymhall, thank you for your review! I understand your concern.

If users import the queue via `fromQueueArn` method, CDK cannot determine whether the queue is encrypted by AWS managed KMS key or not.
https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-sqs/lib/queue.ts#L268-L287

So in my understanding, we can't validate it completely in that case. I think one possible solution is skipping the validation if `encryptionMasterKey` is undefined to prevent error due to missing default child. Does it make sense?

corymhall: I believe users can provide the key in `fromQueueAttributes` so if the `encryptionType` is undefined then we don't do the validation.

corymhall: Yeah I think so.

mrgrain:
@tam0ri Are you still able to address the outstanding comments?

tam0ri:
@mrgrain
Sorry for delay. I still have a motivation to address this. I develop in my spare time, but I have not been able to take a time recently. I'll try reserving a time for this in this week. Can I take some additional time?

mrgrain:
> @mrgrain
> Sorry for delay. I still have a motivation to address this. I develop in my spare time, but I have not been able to take a time recently. I'll try reserving a time for this in this week. Can I take some additional time?

Yes of course! Take as much time you need. If the automation closes this PR, just open a new one and feel free to tag me in it. 😃

tam0ri:
@corymhall Sorry for delay, I made some changes per your suggestion. Could you review it again?

corymhall (approved):

PR #26114, opened 2023-06-26
feat(app-mesh): support port property on weighted targets

### LINKED ISSUE #26083: (app-mesh): WeightedTarget construct is missing port property
### Describe the bug

While the `CfnRoute` give us the ability to define a route with weighted targets with a port, the L2 construct `WeightedTarget` does not have that property (only have `virtual_node` and `weight`), leveraging us to use L1 constructs instead if we want to use the port specification.

Port is required for routes to nodes that have multiple listeners.

### Expected Behavior

`WeightedTarget` L2 construct also let us specify the `port` option.

### Current Behavior

`WeightedTarget` L2 construct does not have the property `port`.

### Reproduction Steps

Example in python; the following is not valid code:

```python
appmesh.RouteSpec.grpc(
    weighted_targets=[appmesh.WeightedTarget(virtual_node=node, port=myPortNumber)],
    match=appmesh.GrpcRouteMatch(port=myPortNumber),
)
```

We need to use L1 construct, to get the same result:

```python
appmesh.CfnRoute.RouteSpecProperty(
    grpc_route=appmesh.CfnRoute.GrpcRouteProperty(
        action=appmesh.CfnRoute.GrpcRouteActionProperty(
            weighted_targets=[appmesh.CfnRoute.WeightedTargetProperty(
                virtual_node=node.virtual_node_name,
                weight=1,
                port=myPortNumber,
            )],
        ),
        match=appmesh.CfnRoute.GrpcRouteMatchProperty(port=myPortNumber),
    ),
)
```

### Possible Solution

Add the `Port` property to the construct interface, since is already a property in the CloudFormation specs.

_No response_

### CDK CLI Version

2.85.0

### Framework Version

2.85.0

### Node.js Version

v16.14.2

### OS

MacOs Ventura

### Language

Python

### Language Version

python 3.11.1

### Other information

I have checked through the CDK docs that typescript CDK version have the same issue, so I guess this applies to all languages available.

I can work in a fix if there is no reason to not work on this fix.

neovasili:
As described in the related issue, `WeightedTarget` L2 construct was missing `port` property which is already present in the L1 construct `CfnRoute`.

This PR adds the missing `port` property to `WeightedTarget` L2 construct.

The PR includes unit tests expansion to cover this new property appearance or absence.

Closes #26083.

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-cloudfront/test/integ.distribution-policies.js.snapshot/integ-distribution-policies.assets.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-cloudfront/test/integ.distribution-policies.js.snapshot/integ-distribution-policies.assets.json
@@ -1,7 +1,7 @@
 {
   "version": "32.0.0",
   "files": {
-    "01042f10dd3272da413b201384cdf825a7467030c0db8a2d5bcfe10b45a30ced": {
+    "17ab10021b77a5b225de54b10a02740bc6d638269786a643726871c66180f980": {
       "source": {
         "path": "integ-distribution-policies.template.json",
         "packaging": "file"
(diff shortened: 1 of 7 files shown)


mrgrain (approved):
🚀

PR #26116, opened 2023-06-26
fix(cdk-lib): Pass lookupRoleArn to NestedStackSynthesizer

IgnacioAcunaF:
NestedStack's synthesizer doesn't receive the lookupRoleArn from the parent stack synthesizer, so the NestedStack tries with local credentials (of the deployment account) instead of assuming a cross-account role (on the target account) as regular non-nested Stack would do.

This PR aims to add lookupRoleArn reference to the StackSynthesizer class and IStackSynthesizer, so it can be use on the NestedStack to explicitly set an IAM role in case of parent stack having one already defined, so CDK uses the role instead of local credentials.

Closes https://github.com/aws/aws-cdk/issues/25171.

----


IgnacioAcunaF:
Exemption Request: an integration test coudn't be added because it would be needed of two AWS accounts to execute (so it can actually switch and use the lookup role on the target account), which will break the logic of integration tests and snapshots reproducibility.

IgnacioAcunaF:
Exemption Request: an integration test coudn't be added because it would be needed of two AWS accounts to execute (so it can actually switch and use the lookup role on the target account), which will break the logic of integration tests and snapshots reproducibility.

IgnacioAcunaF:
Hi @peterwoodworth, hope you are doing great!
Could you help with the review?
Thanks in advance

corymhall (approved):

PR #26123, opened 2023-06-26
chore: build noctilucent WASM library in a container

HBobertz:
Due to the addition of noctilucent to cdk, contributors needed to download rust/rustup to be able to build the cdk.

This uses the pre-existing dependency on Docker/Finch to containerize the process in order to not incur any further dependencies for contributors to manage.

----

### DIFF
--- a/.github/workflows/handle-stale-discussions.yml
+++ b/.github/workflows/handle-stale-discussions.yml
@@ -1,0 +1,16 @@
+name: HandleStaleDiscussions
+on:
+  schedule:
+    - cron: '0 */4 * * *'
+  discussion_comment:
+    types: [created]
+
+jobs:
+  handle-stale-discussions:
+    name: Handle stale discussions
+    runs-on: ubuntu-latest
+    steps:
+      - name: Stale discussions action
+        uses: aws-github-ops/handle-stale-discussions@v1
+        env:
+          GITHUB_TOKEN:  ${{secrets.GITHUB_TOKEN}}
--- a/packages/@aws-cdk/app-staging-synthesizer-alpha/README.md
+++ b/packages/@aws-cdk/app-staging-synthesizer-alpha/README.md
@@ -48,106 +48,44 @@ in your account.
 
 ## Bootstrap Model
 
-Our current bootstrap model looks like this, when you run `cdk bootstrap aws://<account>/<region>` :
+In our default bootstrapping process, when you run `cdk bootstrap aws://<account>/<region>`, the following
+resources are created:
 
-```text
-┌───────────────────────────────────┐┌────────────────────────┐┌────────────────────────┐
-│                                   ││                        ││                        │
-│                                   ││                        ││                        │
-│          ┌───────────────┐        ││    ┌──────────────┐    ││    ┌──────────────┐    │
-│          │Bootstrap Stack│        ││    │  CDK App 1   │    ││    │  CDK App 2   │    │
-│          └───────────────┘        ││    └──────────────┘    ││    └──────────────┘    │
-│                                   ││                        ││                        │
-│                                   ││                        ││                        │
-│   ┌───────────────────────────┐   ││     ┌────────────┐     ││                        │
-│   │IAM Role for CFN execution │   ││┌────│  S3 Asset  │     ││                        │
-│   │    IAM Role for lookup    │   │││    └────────────┘     ││                        │
-│   │  IAM Role for deployment  │   │││                       ││                        │
-│   └───────────────────────────┘   │││                       ││     ┌─────────────┐    │
-│                                   │││            ┌──────────┼┼─────│  S3 Asset   │    │
-│                                   │││            │          ││     └─────────────┘    │
-│ ┌───────────────────────────────┐ │││            │          ││                        │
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -64,6 +64,7 @@ The following tools need to be installed on your system prior to installing the
 - [Python >= 3.6.5, < 4.0](https://www.python.org/downloads/release/python-365/)
 - [Docker >= 19.03](https://docs.docker.com/get-docker/)
   - the Docker daemon must also be running
+- [Rust >= 1.68, rustup >= 1.25.2](https://www.rust-lang.org/tools/install)
```
vinayak-kukreja, line 67: Thank you for adding this. 

How would users be informed of this change? I think an issue can be created and pinned informing the users about this, WDYT?

RomainMuller: Added a prerequisite test to inform users.

scripts/check-build-prerequisites.sh
```diff
@@ -74,6 +74,19 @@ else
     wrong_version
 fi
 
+# Rustup
+app="rustup"
+app_min="1.25.2"
+check_which $app $app_min
+app_v=$(${app} --version 2>/dev/null)
+echo -e "Checking rustup version... \c"
+if [ $(echo $app_v | grep -c -E "^1\.(2(5\.[2-9]\d*|[6-9]\d*\.\d*)|[3-9]\d+\.\d+)") -eq 1 ]
```
Naumel, line 83:
```suggestion
if [ $(echo $app_v | grep -c -E "^1\.(2(5\.[2-9]\d*|[6-9]\.\d*)|[3-9]\d+\.\d+)") -eq 1 ]
```

Naumel (dismissed):
I think there's an extra `\d*`
~~we need to check > 2.5, meaning there is only one digit after the 6-9, before it flips from 2.* to 3.*.~~
the dot is before the 2, not after 👍

scripts/check-build-prerequisites.sh
RomainMuller: Here's a railroad diagram of the original (i.e: my) version of the expression (courtesy of https://regexper.com):

image

rix0rrr (changes requested):
I don't think I'm cool with adding this dependency. Contributing is already hard, and this will make it even harder.

Why do we need it, exactly, and why can that not be addressed another way?

RomainMuller:
Rust is already in `jsii/superchain`, and has been for a LONG while because some Python `cryptography` module that we transitively depend on needs to be built from it's rust source... I didn't think this would be much of a problem since it's already a host prerequisite in many cases...

I guess we can instead leverage the existing `docker` dependency instead and do this stuff within a container if you'd prefer that...

packages/aws-cdk/generate.sh
```diff
@@ -15,30 +15,20 @@ cat > build-info.json <<HERE
...
-  cd ../../lib/vendor/noctilucent
-  rm package.json
-)
+# Build noctilucent package in a Docker/Finch VM
+NOCTILUCENT_GIT="https://github.com/iph/noctilucent.git"
+NOCTILUCENT_COMMIT_ID="6da7c9fade55f8443bba7b8fdfcd4ebfe5208fb1"
+if [ "$(cat lib/vendor/noctilucent/.version 2>/dev/null || echo '')" == "${NOCTILUCENT_GIT}:${NOCTILUCENT_COMMIT_ID}" ]
+then
+  echo "⏭️ Noctilucent WASM binary is up-to date, skipping build..."
+  echo "ℹ️ Delete lib/vendor/noctilucent/.version to force a rebuild."
```
rix0rrr, line 24: ❤️ clever

HBobertz: we were originally using git submodules but it turned out codepipeline didn't like that very much so we had to just get it ourselves

mrgrain: Submodules come with their own set of issues. If it's required for a build, we should not put it into a submodule.

The `noctilucent-wasm` npm package is still an option.

packages/aws-cdk/generate.sh
```diff
@@ -15,30 +15,20 @@ cat > build-info.json <<HERE
...
-  wasm-pack build --target nodejs                                               \
-    --out-dir="${pkgroot}/lib/vendor/noctilucent"                               \
-    --out-name=index                                                            
-
-  cd ../../lib/vendor/noctilucent
-  rm package.json
-)
+# Build noctilucent package in a Docker/Finch VM
+NOCTILUCENT_GIT="https://github.com/iph/noctilucent.git"
+NOCTILUCENT_COMMIT_ID="6da7c9fade55f8443bba7b8fdfcd4ebfe5208fb1"
```
mrgrain, line 20: I agree with @rix0rrr this feels like we are reinventing versioning and releases.

rix0rrr (approved):

scripts/check-build-prerequisites.sh
HBobertz: cool site. Based on this it seems romains expression works. Can you confirm elena?

PR #26130, opened 2023-06-27
feat(apprunner): make `Service` implement `IGrantable`

### LINKED ISSUE #26089: apprunner: Allow `instanceRole` to be obtained from `Service` instances.
### Describe the feature

Currently, `instanceRole` can only be set from the constructor.
If it is possible to retrieve it even after initialization like [`obtainExecutionRole`](https://github.com/aws/aws-cdk/blob/v2.85.0/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts#L704) in `TaskDefinition` of ECS, it would be useful to be able to grant privileges with the `grant` method.

### Use Case

Use as follows:

```typescript
declare const bucket: IBucket;

const service = new apprunner.Service(this, 'Service', {
    source: apprunner.Source.fromEcrPublic({
        imageConfiguration: { port: 8000 },
        imageIdentifier: 'public.ecr.aws/aws-containers/hello-app-runner:latest',
    }),
});

bucket.grantRead(service.obtainInstanceRole());
```

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.85.0

N/A

lpizzinidev:
Implementing `IGrantable` for cases when it's needed to grant permissions to a `Service` instance.

For example:
```
declare const bucket: IBucket;

const service = new apprunner.Service(this, 'Service', {
    source: apprunner.Source.fromEcrPublic({
        imageConfiguration: { port: 8000 },
        imageIdentifier: 'public.ecr.aws/aws-containers/hello-app-runner:latest',
    }),
});

bucket.grantRead(service);
```

Closes #26089.

----

### DIFF
--- a/.github/workflows/github-merit-badger.yml
+++ b/.github/workflows/github-merit-badger.yml
@@ -17,4 +17,4 @@ jobs:
           badges: '[beginning-contributor,repeat-contributor,valued-contributor,admired-contributor,star-contributor,distinguished-contributor]'
           thresholds: '[0,3,6,13,25,50]'
           badge-type: 'achievement'
-          ignore-usernames: '[RomainMuller,rix0rrr,MrArnoldPalmer,iliapolo,otaviomacedo,madeline-k,kaizencc,comcalvi,corymhall,peterwoodworth,ryparker,TheRealAmazonKendra,vinayak-kukreja,Naumel,mrgrain,pahud,cgarvis,kellertk,HBobertz,sumupitchayan,pattasai,SankyRed,udaypant,colifran,khushail,aws-cdk-automation,dependabot[bot],mergify[bot]]'
+          ignore-usernames: '[RomainMuller,rix0rrr,MrArnoldPalmer,iliapolo,otaviomacedo,madeline-k,kaizencc,comcalvi,corymhall,peterwoodworth,ryparker,TheRealAmazonKendra,vinayak-kukreja,Naumel,mrgrain,pahud,cgarvis,kellertk,HBobertz,sumupitchayan,pattasai,SankyRed,udaypant,colifran,khushail,scanlonp,aws-cdk-automation,dependabot[bot],mergify[bot]]'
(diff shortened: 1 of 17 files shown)


packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -1140,6 +1140,17 @@ export class Service extends cdk.Resource {
     this.secrets.push({ name: name, value: secret.arn });
   }
 
+  /**
+   * This method exposes the Instance Role after creating it if not set.
+   * @returns iam.IRole
+   */
+  public obtainInstanceRole(): iam.IRole {
```
corymhall, line 1147: What if we instead made `Service` implement `IGrantable`? Then we could just do
things like

```ts
bucket.grantRead(service);
```

corymhall (changes requested):

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -1052,6 +1053,7 @@ export class Service extends cdk.Resource {
     this.props = props;
 
     this.instanceRole = this.props.instanceRole;
+    this.grantPrincipal = this.instanceRole || new iam.UnknownPrincipal({ resource: this });
```
corymhall, line 1056: I think we should just go ahead and create the role if the user doesn't provide it. I'm not sure why we try to only generate one if needed.

corymhall (changes requested):
Can you also add an `addToRolePolicy` method? I'm also going to change this to a `feat`. Adding the auto creation of the role should cause the integ test to update and we'll need a README update.

lpizzinidev:
Exemption request.
I've updated the related integration tests.

mrgrain (approved):

lpizzinidev:
> Can you also add an `addToRolePolicy` method? I'm also going to change this to a `feat`. Adding the auto creation of the role should cause the integ test to update and we'll need a README update.

@corymhall I've implemented the requested changes.

corymhall (approved):

PR #26134, opened 2023-06-27
chore(ec2): replace latestAmazonLinux with latestAmazonLinux2 used inside BastionHostLinux 

clueleaf:
Synthesizing `BastionHostLinux` generates the following warning:

```
[WARNING] aws-cdk-lib.aws_ec2.MachineImage#latestAmazonLinux is deprecated.
  use MachineImage.latestAmazonLinux2 instead
  This API will be removed in the next major release.
```

This PR replaces `latestAmazonLinux` used inside `BastionHostLinux` with `latestAmazonLinux2` to remove the warning.

----


corymhall (changes requested):
@clueleaf thanks for creating this PR! It looks like changing the machine image results in a breaking change (the instance will be replaced) so this will have to be behind a feature flag.

If you are up for that then we should also make `machineImage` a required property and not default it for the user. Since we can't make it required from a type perspective we'll have to do it with a check inside the construct.

KouKinyo:
@corymhall 
Thanks for your review.
I'm not sure if we should call this a breaking change, because the document is already clarifying `the instance may be replaced on every deployment` when we do not specify a `machineImage`.

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/bastion-host.ts#L66-L67

clueleaf:
@corymhall 
Thanks for your review.
I'm not sure if we should call this a breaking change, because the document is already clarifying that `the instance may be replaced on every deployment` when we do not specify a `machineImage`.

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/bastion-host.ts#L66-L67

corymhall (dismissed):
> I'm not sure if we should call this a breaking change, because the document is already clarifying that the instance may be replaced on every deployment when we do not specify a machineImage.

Good point!

clueleaf:
@corymhall hi, will this PR get merged?

corymhall (approved):

PR #26144, opened 2023-06-28
feat(core): acknowledge warnings

corymhall:
This PR adds the ability to acknowledge annotation warning messages. 

The main motivation behind this is to allow people to set the `--strict` mode to fail synthesis on warnings. Currently it is all or nothing, you have to get rid of _all_ warnings to use `--strict`. With this feature users will be able to `acknowledge` warnings saying that they are aware, but it does not apply to them.

Since we want all warnings to now have an id this will deprecate the `addWarning` method and adds a new `addWarningV2` method.

Since the acknowledgements and warnings are written as metadata, it is possible to enhance this in the future to report on warnings and acknowledgements.

----

### DIFF
--- a/.github/workflows/update-contributors.yml
+++ b/.github/workflows/update-contributors.yml
@@ -22,7 +22,7 @@ jobs:
           # Git commit details
           branch: automation/update-contributors
           author: aws-cdk-automation <aws-cdk-automation@users.noreply.github.com>
-          commit-message: chore: update Contributors File
+          commit-message: 'chore: update Contributors File'
           # Pull Request details
           title: 'chore: update Contributors File'
           labels: auto-approve


allowed-breaking-changes.txt
```diff
@@ -164,3 +164,7 @@ removed:aws-cdk-lib.triggers.TriggerProps.timeout
 change-return-type:aws-cdk-lib.aws_ec2.SecurityGroup.determineRuleScope
 # broken only in non-JS/TS, where that was not previously usable
 strengthened:aws-cdk-lib.aws_s3_deployment.BucketDeploymentProps
+
+# weakened added additional types to a union type, meaning all new props are optional
```
corymhall, line 168: Need a second opinion on this

packages/@aws-cdk/aws-gamelift-alpha/lib/fleet-base.ts
```diff
@@ -639,10 +639,10 @@ export abstract class FleetBase extends cdk.Resource implements IFleet {
   }
 
   protected warnVpcPeeringAuthorizations(scope: Construct): void {
-    cdk.Annotations.of(scope).addWarning([
+    cdk.Annotations.of(scope).addWarningV2('Gamelift:Fleet:AutorizeVpcPeering', [
```
rix0rrr, line 642: If we're going to schematize this, let's set a standard? This one uses a 3-part warning, the one below uses a 2-part warning.

If we take a cue from the feature flags, which this feels a bit similar to, maybe the flag format is something like `@aws-cdk/aws-service:someCamelCaseTerm` ?

packages/@aws-cdk/aws-servicecatalogappregistry-alpha/test/application-associator.test.ts
```diff
@@ -237,7 +237,7 @@ describe('Scope based Associations with Application with Cross Region/Account',
     const crossRegionStack = new cdk.Stack(app, 'crossRegionStack', {
       env: { account: 'account', region: 'region' },
     });
-    Annotations.fromStack(crossRegionStack).hasWarning('*', 'Environment agnostic stack determined, AppRegistry association might not work as expected in case you deploy cross-region or cross-account stack.');
+    Annotations.fromStack(crossRegionStack).hasWarning('*', 'SCAppRegistry:EnvironmentAgnosticStack: Environment agnostic stack determined, AppRegistry association might not work as expected in case you deploy cross-region or cross-account stack.');
```
rix0rrr, line 240: Bikeshed:

```suggestion
    Annotations.fromStack(crossRegionStack).hasWarning('*', '[SCAppRegistry:EnvironmentAgnosticStack] Environment agnostic stack determined, AppRegistry association might not work as expected in case you deploy cross-region or cross-account stack.');
```

The `[flag] What the hell` feels more natural to me for some reason... ?

packages/aws-cdk-lib/assertions/test/annotations.test.ts
```diff
@@ -209,7 +209,7 @@ class MyAspect implements IAspect {
   };
 
   protected warn(node: IConstruct, message: string): void {
-    Annotations.of(node).addWarning(message);
+    Annotations.of(node).addWarningV2(node.node.id, message);
```
rix0rrr, line 212: I know this is only for tests, but a node ID as warning doesn't make sense to me. You could have put a dummy string here.

packages/aws-cdk-lib/cloud-assembly-schema/lib/cloud-assembly/metadata-schema.ts
```diff
@@ -231,10 +255,36 @@ export type LogicalIdMetadataEntry = string;
  */
 export type StackTagsMetadataEntry = Tag[];
 
+/**
+ * @see ArtifactMetadataEntryType.ACKNOWLEDGE
+ */
+export interface AcknowledgementMetadataEntry {
```
rix0rrr, line 261: Why do acknowledgements need to make it into the Cloud Assembly? Can't we choose to not emit acknowledged warnings?

packages/aws-cdk-lib/core/lib/annotations.ts
```diff
@@ -24,14 +24,59 @@ export class Annotations {
...
+   * declare const construct: Construct;
+   * Annotations.of(construct).addWarningV2('Library:Construct:ThisIsAWarning', 'Some message explaining the warning');
+   *
+   * @param id the unique identifier for the warning. This can be used to acknowledge the warning
    * @param message The warning message.
    */
+  public addWarningV2(id: string, message: string) {
+    const warning = {
+      id: id,
+      scope: this.scope.node.path,
```
rix0rrr, line 65: Metadata messages are already attached to a scope, so this information will be doubled?

packages/aws-cdk-lib/core/test/annotations.test.ts
```diff
@@ -79,18 +79,87 @@ describe('annotations', () => {
...
+    Annotations.of(c1).acknowledgeWarning('MESSAGE2', 'I Ack this');
+
+    // THEN
+    const assembly = app.synth();
+    let acknoledgement: any = {};
+    for (const s of Object.values(assembly.manifest.artifacts ?? {})) {
+      for (const [_path, md] of Object.entries(s.metadata ?? {})) {
+        for (const x of md) {
+          if (x.type === 'aws:cdk:acknowledge') {
+            acknoledgement.message = x.data as string;
```
rix0rrr, line 151: Mutations in the `THEN` part of a test?

packages/aws-cdk-lib/core/test/app.test.ts
```diff
@@ -78,8 +78,8 @@ describe('app', () => {
       '/stack1/s1c1': [{ type: 'aws:cdk:logicalId', data: 's1c1' }],
       '/stack1/s1c2':
         [{ type: 'aws:cdk:logicalId', data: 's1c2' },
-          { type: 'aws:cdk:warning', data: 'warning1' },
-          { type: 'aws:cdk:warning', data: 'warning2' }],
+          { type: 'aws:cdk:warning', data: { id: 'warning1', message: 'warning1', scope: 'stack1/s1c2' } },
```
rix0rrr, line 81: This changes the type associated with `aws:cdk:warning`. I don't think that's safe if we consider external processors.

Either we now have a new type `aws:cdk:warningv2`, or we find a way to add the new information additively into the old schema. 

But if we can keep warning filtering entirely in the CDK app, there's no need to change this format at all. We can turn the message of a `warningv2` into:

```
{ type: 'aws:cdk:warning', data: '[warning1] warning message here' }
```

And it'll be backwards compatible.

packages/aws-cdk-lib/cx-api/lib/cloud-artifact.ts
```diff
@@ -110,12 +111,36 @@ export class CloudArtifact {
...
+    // messageId: scopes[]
+    const acks = new Map<string, string[]>();
 
+    // collect any acknowledgements
+    for (const [_id, metadata] of Object.entries(this.manifest.metadata || { })) {
+      for (const entry of metadata) {
+        if (entry.type === cxschema.ArtifactMetadataEntryType.ACKNOWLEDGE) {
+          const data = (entry.data as cxschema.AcknowledgementMetadataEntry);
+          acks.set(data.id, data.scopes);
+        }
```
rix0rrr, line 123: I think all of this work should probably happen at synth time in the CDK app.

packages/aws-cdk-lib/core/test/app.test.ts
rix0rrr: Or honestly, a better message would probably be:

```
warning message here [warning1]
```

rix0rrr (changes requested):
Thanks for this, this will be a great improvement!

My overall concerns are:

- I'd like to do the filtering inside the CDK app and not change the cloud assembly format.
- Quite personal, but I don't like the format of the suppression flags. I'd personally prefer `camelCase` or `kebab-case`, and not too many components to it.

The last one can be argued though, there is no accounting for taste I guess. But maybe put it up to the team with a couple of alternatives?

packages/@aws-cdk/aws-servicecatalogappregistry-alpha/test/application-associator.test.ts
rix0rrr: Or `What the hell [flag]`, or perhaps `What the hell [acknowledge=flag]`

packages/@aws-cdk/aws-gamelift-alpha/lib/fleet-base.ts
corymhall: yeah I like this format better `@aws-cdk/aws-service:someCamelCaseTerm`. I'll update.

packages/@aws-cdk/aws-servicecatalogappregistry-alpha/test/application-associator.test.ts
corymhall: Good point, I kind of like `What the hell [ack: flag]` because it makes it more clear to the user what they need to provide to ack.

packages/aws-cdk-lib/cx-api/lib/cloud-artifact.ts
corymhall: I started going down that path and ended up with this solution for a couple of reasons, the main one being that we create a new instance of `Annotations` every time we create a message. We currently use the metadata as the storage mechanism. In order to do it at synth time we would need to implement a new storage mechanism. This probably wouldn't be too bad (probably just another construct we getOrCreate).

I can explore that.

packages/aws-cdk-lib/core/test/annotations.test.ts
corymhall: 🤦‍♂️

rix0rrr: Just asking, if you need to do this work it probably wants to go into a reusable function that's in the core library?

rix0rrr:
> We can't set context after constructs have been added.

Honestly, I've seen this restriction do more harm than good. 

I'm not opposed to removing it.

packages/aws-cdk-lib/core/lib/adr/acknowledge-warnings.md
```diff
@@ -0,0 +1,181 @@
...
+   functionality doesn't make much sense outside of this context (can't find
+   where anybody has asked for it). It also may require updating the metadata
+   types since currently there is no unique identifier for a given metadata
+   entry (other than the message).
+
+## Consequences
+
+With the recommended solution the only major consequence is that it requires
+updating the `metadata-schema`, but we can do this in a non-breaking way
+(addition of new types). The alternatives may also require changes to the schema
```
rix0rrr, line 180: It changes the schema, and interpreting the schema becomes more complex. I would really really like to avoid having an add list and a scrap list in the same piece of data, that the client hen has to evaluate.

Adding this as a feature to the core seems preferable to me.

corymhall: @rix0rrr Then do you think the context + remove metadata option is the one to go with?

rix0rrr (approved):

Issue #26148, opened 2023-06-28
RestApi: wrong description used for deployments


woltsu:
### Describe the bug

When creating an instance of a `RestApi` construct, by default it has a `Deployment` attached to it. The description of the `Deployment` can be configured with an option given to the `RestApi`, like so:

```tsx
const api = new RestApi(this, 'Api', {
	restApiName: 'MyApi',
	deployOptions: {
		description: 'Deployment description'
	}
});
```

However, looking at the [source code](https://github.com/aws/aws-cdk/blob/72eb1e957afb9d1573445999eddd38b5c345fa7a/packages/aws-cdk-lib/aws-apigateway/lib/restapi.ts#L592C25-L592C25), we see that instead of using the `props.deployOptions.description`, it uses `props.description`.

### Expected Behavior

It should use the description provided in `deployOptions`

### Current Behavior

It uses the description of the `RestApi`

### Reproduction Steps

```tsx
const api = new RestApi(this, 'Api', {
	restApiName: 'MyApi',
        description: 'My api description',
	deployOptions: {
		description: 'Deployment description'
	}
});
```

The created deployment's description will be `My api description` instead of `Deployment description`

### CDK CLI Version

2.85.0

### Node.js Version

16

### OS

Mac

### Language

Typescript

pahud:
Yes, we should get it fixed. Thanks for your report.

woltsu:
Fixed in https://github.com/aws/aws-cdk/pull/26149 😊

Issue #20934, opened 2022-06-30
(api-gateway): allow configuration of deployment description


rix0rrr:
### Describe the feature

Currently, the above is not possible as the text is hardcoded here https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-apigateway/lib/restapi.ts#L557

### Use Case

Customer has a pipeline that deploys an api gateway holding multiple resources. They would like to be able to set the description (check the attached screenshot) per deployment. From inside their pipeline, they could get the commit hash / commit message, timestamp, custom text, and other git-related metadata that they would like to set as description. 

### Proposed Solution

Make description configurable

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

-

### Environment details (OS name and version, etc.)

-

josephedward:
@peterwoodworth  I started working on this (https://github.com/aws/aws-cdk/pull/21207), but I wasn't sure where to edit a test - `_configureDeployment` and `RestApiBaseProps` do not seem to be present in the /test folder in aws-cdk/packages/@aws-cdk/aws-apigateway ?

PR #26149, opened 2023-06-28
fix(apigateway): deployOptions.description is not used for deployment's description

woltsu:
When creating an instance of a `RestApi` construct, by default it has a `Deployment` attached to it. The description of the `Deployment` can be configured with an option given to the `RestApi`, like so:

```tsx
const api = new RestApi(this, 'Api', {
	restApiName: 'MyApi',
	deployOptions: {
		description: 'Deployment description'
	}
});
```

However, looking at the [source code](https://github.com/aws/aws-cdk/blob/72eb1e957afb9d1573445999eddd38b5c345fa7a/packages/aws-cdk-lib/aws-apigateway/lib/restapi.ts#L592C25-L592C25), we see that instead of using the `props.deployOptions.description`, it uses `props.description`.

Fixes https://github.com/aws/aws-cdk/issues/26148

----


packages/aws-cdk-lib/aws-apigateway/lib/restapi.ts
```diff
@@ -589,7 +589,7 @@ export abstract class RestApiBase extends Resource implements IRestApi {
     if (deploy) {
 
       this._latestDeployment = new Deployment(this, 'Deployment', {
-        description: props.description? props.description :'Automatically created by the RestApi construct',
+        description: props.deployOptions?.description ?? 'Automatically created by the RestApi construct',
```
yamatatsu, line 592: That make sense.
I found the issue for this imprementation. And it say this imprementation is to be able to set the description per deployment.
https://github.com/aws/aws-cdk/issues/20934

I think this PR make the imprementation close to customer thinking.

yamatatsu (approved):
LGTM

yamatatsu (approved):
Reapprovement for changing PR label.

packages/aws-cdk-lib/aws-apigateway/lib/restapi.ts
rix0rrr: This is good, but please consider backwards compatibility. This will change the description of all deployments that don't have an explicit description today. This bug has been in here for ~4 years and people have so far been fine with the current behavior.

A safer implementation seems to me:

```suggestion
  description: props.deployOptions?.description ?? props.description ?? 'Automatically created by the RestApi construct',
```

rix0rrr (dismissed):

rix0rrr (approved):

rix0rrr (dismissed):

PR #26155, opened 2023-06-28
fix(efs): FileSystem deployment fails if vpcSubnets is changed

### LINKED ISSUE #25099: (aws-efs): EFS mount target will fail on subnet change
### Describe the bug

When removing an EFS Mount point from a CDK deployment (say by defining fewer subnets than what are previously deployed), CDK will remove the correct mount point but will use the old unique Id for the deleted one if that subnet is in the middle of the subnet list. This will cause CloudFormation to fail as it tries to make a new Mount Point for the same EFS File System in the same AZ, which it can't do.

### Expected Behavior

If I have a CDK construct
```typescript
const fileSystem = new FileSystem(this, "EfsFileSystem", {
    vpc,
    vpcSubnets: {
        subnets: [subnetA, subnetB, subnetC]
    }
})
```

This will synth 3 Mount points, eg.

```yaml
EfsFileSystemEfsMountTarget01:
  Properties:
    SubnetId: subnetA
EfsFileSystemEfsMountTarget02:
  Properties:
    SubnetId: subnetB
EfsFileSystemEfsMountTarget03:
  Properties:
    SubnetId: subnetC
```

When I remove one of the subnets, i.e. remove subnet B
```typescript
const fileSystem = new FileSystem(this, "EfsFileSystem", {
    vpc,
    vpcSubnets: {
        subnets: [subnetA, subnetC]
    }
})
```

CDK synth will create 2 mount points with correct Ids
```yaml
EfsFileSystemEfsMountTarget01:
  Properties:
    SubnetId: subnetA
EfsFileSystemEfsMountTarget03:
  Properties:
    SubnetId: subnetC
```

### Current Behavior

If I have a CDK construct
```typescript
const fileSystem = new FileSystem(this, "EfsFileSystem", {
    vpc,
    vpcSubnets: {
        subnets: [subnetA, subnetB, subnetC]
    }
})
```

This will synth 3 Mount points, eg.

```yaml
EfsFileSystemEfsMountTarget01:
  Properties:
    SubnetId: subnetA
EfsFileSystemEfsMountTarget02:
  Properties:
    SubnetId: subnetB
EfsFileSystemEfsMountTarget03:
  Properties:
    SubnetId: subnetC
```

When I remove one of the subnets, i.e. remove subnet B
```typescript
const fileSystem = new FileSystem(this, "EfsFileSystem", {
    vpc,
    vpcSubnets: {
        subnets: [subnetA, subnetC]
    }
})
```

CDK synth will create 2 mount points with an incorrect Id
```yaml
EfsFileSystemEfsMountTarget01:
  Properties:
    SubnetId: subnetA
EfsFileSystemEfsMountTarget02:
  Properties:
    SubnetId: subnetC
```

> Note the subnet names and unique Ids have changed for one of the mount targets.

Which is technically correct, but now CloudFormation will fail with this error:

```
mount target already exists in this AZ
```

This is because since the uniqueid of the mount target in subnetC has changed it's trying to create a new one.

### Reproduction Steps

Create a CDK construct with an EFS attached to 3 subnets.

```typescript
const fileSystem = new FileSystem(this, "EfsFileSystem", {
    vpc,
    vpcSubnets: {
        subnets: [subnetA, subnetB, subnetC]
    }
})
```

now delete subnetB.

### Possible Solution

CDK needs to keep track of the subnets and mount target unique IDs somehow

_No response_

### CDK CLI Version

2.67.0

### Framework Version

_No response_

### Node.js Version

18.12.1

### OS

Windows

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

ahmetsoykan:
Changing Logical IDs for EfsMountTarget

While using the forEach, index will be added as a suffix to logical id of EfsMountTarget 
this is causing an error when the subnets array in the props has changed.

Closes #25099

----

### DIFF
--- a/packages/@aws-cdk/aws-lambda-python-alpha/lib/Dockerfile
+++ b/packages/@aws-cdk/aws-lambda-python-alpha/lib/Dockerfile
@@ -6,6 +6,7 @@ FROM $IMAGE
 ARG PIP_INDEX_URL
 ARG PIP_EXTRA_INDEX_URL
 ARG HTTPS_PROXY
+ARG POETRY_VERSION=1.5.1
 
 # Add virtualenv path
 ENV PATH="/usr/app/venv/bin:$PATH"
@@ -17,7 +18,7 @@ ENV PIP_CACHE_DIR=/tmp/pip-cache
 ENV POETRY_CACHE_DIR=/tmp/poetry-cache
 
 RUN \
-# create a new virtualenv for python to use 
+# create a new virtualenv for python to use
 # so that it isn't using root
     python -m venv /usr/app/venv && \
 # Create a new location for the pip cache
@@ -31,7 +32,7 @@ RUN \
 # Ensure all users can write to poetry cache
     chmod -R 777 /tmp/poetry-cache && \
 # pipenv 2022.4.8 is the last version with Python 3.6 support
-    pip install pipenv==2022.4.8 poetry && \
+    pip install pipenv==2022.4.8 poetry==$POETRY_VERSION && \
 # Ensure no temporary files remain in the caches
     rm -rf /tmp/pip-cache/* /tmp/poetry-cache/*
 
--- a/packages/aws-cdk/lib/api/deployments.ts
+++ b/packages/aws-cdk/lib/api/deployments.ts
@@ -6,7 +6,7 @@ import { ISDK } from './aws-auth/sdk';
 import { CredentialsOptions, SdkForEnvironment, SdkProvider } from './aws-auth/sdk-provider';
 import { deployStack, DeployStackResult, destroyStack, makeBodyParameterAndUpload, DeploymentMethod } from './deploy-stack';
 import { HotswapMode } from './hotswap/common';
-import { loadCurrentTemplateWithNestedStacks, loadCurrentTemplate } from './nested-stack-helpers';
+import { loadCurrentTemplateWithNestedStacks, loadCurrentTemplate, flattenNestedStackNames, TemplateWithNestedStackCount } from './nested-stack-helpers';
 import { ToolkitInfo } from './toolkit-info';
 import { CloudFormationStack, Template, ResourcesToImport, ResourceIdentifierSummaries } from './util/cloudformation';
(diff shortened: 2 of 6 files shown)


packages/aws-cdk-lib/aws-efs/test/efs-file-system.test.ts
```diff
@@ -559,4 +559,62 @@ test('imported file system can not add statements to file system policy', () =>
   expect(importedFileSystemResult).toStrictEqual({
     statementAdded: false,
   });
+});
+
+test('mountTargetOrderInsensitiveLogicalId flag is true', () => {
```
lpizzinidev, line 564: Can you update the test to use custom subnets to check that the subnet ID is used as a suffix?

For example:
```
new FileSystem(customVpc, 'EfsFileSystem', {
  vpc: customVpc,
  vpcSubnets: { subnets: [ec2.Subnet.fromSubnetId(stack, 'SubnetExample', 'subnet-1')] },
});
```

lpizzinidev (changes requested):
Thanks for opening a new PR! 👍
Looks good overall, just a small change on the test.

packages/aws-cdk-lib/aws-efs/test/efs-file-system.test.ts
ahmetsoykan: @lpizzinidev hello, when checking subnetID from created VPC, SubnetID attr tokenised and hard to correlate with template output in this test. 

I think with this PR, we're actually not passing SubnetID attribute to MountTarget's logical IDs, we're passing ${subnet.node.id} which i believe sort of concatenation like `${subnetgroupname}Subnet${index}`. So i'm a bit unsure passing subnet selection to FileSystem Contruct. Did you want me to do this because of some future changes that might have happened in VPC definitions in this repository cause test failure?

Isn't it a good method that checking generated logical ids of mount target from resulting different cfn templates regarding to this feature flag?

lpizzinidev (approved):

rix0rrr (approved):

PR #26173, opened 2023-06-29
chore(app-staging-synthesizer): tool to test maximum ecr repository allowance

kaizencc:
And that maximum allowance right now is 20 (it's actually 21 but we are rounding down for safety).

```bash
repos: 2, size: 24820 bytes
repos: 3, size: 26189 bytes
repos: 4, size: 27558 bytes
repos: 5, size: 28927 bytes
repos: 6, size: 30296 bytes
repos: 7, size: 31665 bytes
repos: 8, size: 33034 bytes
repos: 9, size: 34403 bytes
repos: 10, size: 35772 bytes
repos: 11, size: 37148 bytes
repos: 12, size: 38524 bytes
repos: 13, size: 39900 bytes
repos: 14, size: 41276 bytes
repos: 15, size: 42652 bytes
repos: 16, size: 44028 bytes
repos: 17, size: 45404 bytes
repos: 18, size: 46780 bytes
repos: 19, size: 48156 bytes
repos: 20, size: 49532 bytes
repos: 21, size: 50908 bytes
```

Consistently, each additional repo adds 1369 bytes to the template. Before adding the inline functions, each additional repo was roughly 1200 bytes.

Current rough breakdown of size in the integ test (1 s3 bucket and 2 ecr repos):

25KB

~ 2KB for node version mapping
~ 6KB for auto delete images (4KB is the inline function)
~ 6KB for auto delete objects (4KB is the inline function)

Based on these calculations, it is semi reasonable that the repo allowance has gone from 38 to 21. 

----


rix0rrr:
One final thing that just occurred to me: whatever the number is at any one point, we can never lower it. Can you add a test that creates 14 different images and synths, to check that we can successfully add at least 14 images?

Name the test and/or put a comment next to it so that it's very obvious why it's testing that, so that a person in the future doesn't go "well let me lower that number to 12".

PR #26196, opened 2023-07-02
feat(scheduler): ScheduleGroup

filletofish:
This PR contains implementation of ScheduleGroup. 

A Schedule is the main resource in Amazon EventBridge Scheduler, this PR adds ScheduleGroup which can be used to group Schedules and on which Schedule depends.

Every AWS account comes with a default group for schedules. Customers can also create a custom groups to organise schedules that share a common purpose or belong to the same environment. 

Schedule has a property `group` that determines what group is the schedule associated with.

To be able to test adding schedules to the group I have added property `group` to private class `Schedule` and used `Lazy` functionality to be able to update `group` of the schedule dynamically. 

----


packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
```diff
@@ -0,0 +1,363 @@
...
+  metricSentToDLQTrunacted(props?: cloudwatch.MetricOptions): cloudwatch.Metric;
+
+  /**
+   * Grant the indicated permissions on this group to the given principal
+   */
+  grant(grantee: iam.IGrantable, ...actions: string[]): iam.Grant;
+  /**
+   * Grant list and get schedule permissions for schedules in this group to the given principal
+   */
+  grantReadSchedules(identity: iam.IGrantable): iam.Grant;
```
filletofish, line 115: Compared to RFC I have updated the method name to be more precise that the method grants permissions to manage schedules in the group, not the permissions for the group itself. 

There are IAM permissions for the schedules and for the groups.

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
```diff
@@ -0,0 +1,363 @@
...
+  arnForScheduleInGroup(scheduleName: string): string {
+    return Arn.format({
+      region: this.env.region,
+      account: this.env.account,
+      partition: Aws.PARTITION,
+      service: 'scheduler',
+      resource: 'schedule',
+      resourceName: this.groupName + '/' + scheduleName,
+    });
+  }
```
filletofish, line 261: Not sure if this is right, will check with @Jacco  how to test it better without implemented integration tests. 

Also, should we make this method private?

Jacco: Lets add a good integration in the next PR.

Jacco: Think it can be private.

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
```diff
@@ -36,19 +37,31 @@ export interface ScheduleProps {
...
   constructor(scope: Construct, id: string, props: ScheduleProps) {
     super(scope, id);
 
+    this.group = props.group;
+
     new CfnSchedule(this, 'Resource', {
       flexibleTimeWindow: { mode: 'OFF' },
       scheduleExpression: props.schedule.expressionString,
       scheduleExpressionTimezone: props.schedule.timeZone?.timezoneName,
+      groupName: Lazy.string({ produce: () => this.group?.groupName }),
```
Jacco, line 64: Why does this one need to be lazy?

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
```diff
@@ -0,0 +1,363 @@
...
+      name: this.physicalName,
+    });
+
+    this.groupArn = this.getResourceArnAttribute(group.attrArn, {
+      service: 'scheduler',
+      resource: 'schedule-group',
+      resourceName: this.physicalName,
+    });
+    this.groupName = this.physicalName;
+  }
```
Jacco, line 362: I think a call to applyRemovalPolicy is missing here? This means the removalPolicy from the props will not be emitted to CloudFormation.

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
```diff
@@ -36,19 +37,31 @@ export interface ScheduleProps {
...
 /**
  * An EventBridge Schedule
  */
 export class Schedule extends Resource implements ISchedule {
+  group?: IGroup;
+
   constructor(scope: Construct, id: string, props: ScheduleProps) {
     super(scope, id);
 
+    this.group = props.group;
```
Jacco, line 58: Not sure if we should do:

this.group = props.group ?? Group.fromGroupName(this, "DefaultGroup");

filletofish:
```
Group.fromDefaultGroup(this, "DefaultGroup")
``` 

I think it shouldn't create a new resource, as we just refer to existing one by ARN? 

Why would we add a default group here? So customers can refer to it?

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
filletofish: > Why does this one need to be lazy? 

Because group can be assigned dynamically later in the code after schedule is created. 

```
const schedule1  = new Schedule(...);
const group = new Group(...);

group.addSchedules(schedule1);
```

Jacco: Good point :-) Totally missed that.

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
Jacco: Yeah, for:

```ts
const schedule = new Schedule(this, 'MySchedule', {
   /* anything but no group */
});
schedule.group.grantReadSchedules(role);
```

filletofish: Hm, interesting. I am not sure if group should be explicitly set by customer to `default` or not. 

Curious what @kaizencc and @comcalvi think about it? Do you have similar examples in the CDK library already?

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
```diff
@@ -0,0 +1,372 @@
...
+  /**
+   * The arn of the schedule group
+   *
+   * @attribute
+   */
+  abstract readonly groupArn: string;
+
+  addSchedules(...schedules: ISchedule[]): void {
+    schedules.forEach(schedule => {
+      // Temporary using ISchedule as parameter for this method,
```
kaizencc, line 148: info: why temporarily using ISchedule?

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
```diff
@@ -0,0 +1,372 @@
...
+      }
+    });
+  }
+
+  /**
+   * Return the given named metric for this group schedules
+   *
+   * @default - sum over 5 minutes
+   */
+  metric(metricName: string, props?: cloudwatch.MetricOptions): cloudwatch.Metric {
```
kaizencc, line 161: can we explicitly call out public APIs as `public metric`? Pretty much everywhere -- we prefer to be explicit.

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
```diff
@@ -0,0 +1,372 @@
...
+
+  /**
+   * Import a default schedule group.
+   *
+   * @param scope construct scope
+   * @param id construct id
+   */
+  static fromDefaultGroup(scope: Construct, id: string): IGroup {
+    return Group.fromGroupName(scope, id, 'default');
+  }
```
kaizencc, line 335: I think this is a bit of a misnomer. IIRC we are not importing a default schedule group, we are referencing the default schedule group that we know will be created alongside the schedule?

If so, I suggest naming this function `defaultGroup` so it can be used as `Group.defaultGroup`. Now that I think about it, does this method belong here at all?

It feels like the use case is to attach a default group onto a scheduler if no group is provided:

```ts
this.group = props.group ?? Group.fromDefaultGroup(this, "DefaultGroup");
```

Is there a case to expose this as a static property at all? Or should it just be an internal method for `Schedule` to be called if no group is provided?

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
kaizencc: To answer the first question, calling `fromDefaultGroup` will not create a new resource so we are in the clear there.

I have a bit of an issue here, which is a scenario like this:

```ts
const s = new Schedule();
const dg = s.group // this would be the default group
const g = new Group();
g.addSchedules(s);

const ng = s.group;

expect(dg).toEqual(ng); // but this assertion will fail, which is unexpected
```

an additional question: can a schedule be a part of multiple groups? yes right?

```ts
const s = new Schedule();
const g1 = new Group();
g1.addSchedules(s);
const g2 = new Group();
g2.addSchedules(s);

// what will be the value for this?
s.group;
```

maybe the solution here is to expose a groups property and have the defaultGroup be an item in the list

kaizencc (changes requested):
Thanks @filletofish and @jacco for the great start. have a few api questions but nothing major

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
filletofish: I think there is a confusion. Schedule can be placed only in one group.

> expect(dg).toEqual(ng); // but this assertion will fail, which is unexpected

I think it is expected to fail. Why do you think it is a problem?

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
filletofish: Because `Schedule` is private now, but `ISchedule` is not. If I use private class `Schedule` as type for the parameter of a public method - the build will fail

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
kaizencc: > Schedule can be placed only in one group.

Ok. Do we enforce that in the cdk? If so, I don't see a unit test for it.


> I think it is expected to fail. Why do you think it is a problem?

The problem is that `s.group` equals different things over the course of the cdk app. At one point it is the default group, at another it is the given group. We should try to avoid that. Maybe we can lazily evaluate `group` so that we know whether or not the group is the default one or provided later:

```ts
// when group.addSchedules(s); is called, we populate s._group

this.group = Lazy.produce(() => this._group ?? Group.fromDefaultGroup(this, "DefaultGroup"));
```

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
kaizencc: I mean that we _should_ be using `ISchedule` here

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
filletofish: > Ok. Do we enforce that in the cdk? If so, I don't see a unit test for it.

hm, this behaviour is enforced by CloudFormation.  But we could check that after two updates of the groups, the schedule has the latest group. 

> The problem is that s.group equals different things over the course of the cdk app 

Sorry, I am not following, why would the fact that the variable `group` is different through out app execution is a problem?

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
filletofish: > Now that I think about it, does this method belong here at all?

Yeah, I think there are the following use-cases of using the default group: 

1. Get metrics
2. Grant permissions (to the schedules inside group and to the group itself)
3. Explicitly add schedules to the default group 

Do not mind calling `Group.defaultGroup`. I just thought that if we pass a `scope` to the method, it should have the word `from`, because we will be creating a CDK resource (but not a CFN resource)

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
filletofish: > I mean that we should be using ISchedule here

Hm, good point. Then we would need to make a mutable property `group` on interface `ISchedule`.  I originally thought that we could import schedules from ARN in the future (e.g. from another stack / account), so that's why it's better to explicitly point to `Schedule` as a class.

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
kaizencc: 1) If we know that a schedule can be placed only in one group, we should error out at synth time if we try to add a schedule to two groups. Should not wait for CFN behavior to kick in.

2) Because it is confusing. Someone may use the initial `group` to get metrics, or pass it to somewhere else in their cdk app, or assert its value, or whatever. Not knowing that they will override the `group` later in the CDK app. It's hard to reason about what is happening when a variable changes throughout app execution. Furthermore, I argue that it is _incorrect_ for `group` to be `defaultGroup` if later we add the schedule to a specific group.

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
kaizencc: Importing a schedule from ARN should return an `ISchedule` class. We can't mutate imports so they cannot be created as a `Schedule`.

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
kaizencc: I think the `from` keyword implies that we are grabbing a resource "from" somewhere that already exists in your aws account. But I'm not married to the semantics, I don't think its important either way.

Sounds like there is a use case beyond what I see in the implementation. Can we add a section to the README that explains how to access the defaultGroup then?

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
Jacco: Maybe - to avoid confusion - we should not have the addSchedules.

The only way to set a group then is through the constructor. Which probably also makes the Lazy unnecessary.

filletofish: @kaizencc, thank you! Your arguments make sense to me. I originally thought that mutable variables should not be an issue for CDK as we use an imperative language, but CDK classes are declarative. Plus in general, immutable is less confusing than mutable. 

I like Jacco's suggestion. I will remove method `addSchedules`. I like that there will be only a single way to put `Schedule` in a `Group` (through `Schedule` properties).

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
filletofish: > I think the from keyword implies that we are grabbing a resource "from" somewhere that already exists in your aws account. But I'm not married to the semantics, I don't think its important either way.

Exactly. That's why we used `fromDefaultGroup` as default group already exists in our AWS accounts (it doesn't need to be created). 

> Can we add a section to the README that explains how to access the defaultGroup then?

Will do, good call out.

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
kaizencc: We can always worry about adding in `addSchedules` later, if people want

packages/@aws-cdk/aws-scheduler-alpha/lib/group.ts
filletofish: Thanks! I have actually missed that.

> @default RemovalPolicy.RETAIN

Are we sure that we want the default to be `RETAIN`?

packages/@aws-cdk/aws-scheduler-alpha/README.md
```diff
@@ -97,7 +98,31 @@ const oneTimeSchedule = new Schedule(this, 'Schedule', {
...
 ### Grouping Schedules
 
-TODO: Group is not yet implemented. See section in [L2 Event Bridge Scheduler RFC](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0474-event-bridge-scheduler-l2.md)
+Your AWS account comes with a default scheduler group. You can access default group in CDK with:
+
+```text
+const defaultGroup = Group.fromDefaultGroup(this, "DefaultGroup");
+```
+
+If not specified a schedule is added to the default group. However, uou can also add the schedule to a custom scheduling group managed by you:
```
kaizencc, line 107:
```suggestion
If not specified a schedule is added to the default group. However, you can also add the schedule to a custom scheduling group managed by you:
```

packages/@aws-cdk/aws-scheduler-alpha/README.md
```diff
@@ -97,7 +98,31 @@ const oneTimeSchedule = new Schedule(this, 'Schedule', {
 
 ### Grouping Schedules
 
-TODO: Group is not yet implemented. See section in [L2 Event Bridge Scheduler RFC](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0474-event-bridge-scheduler-l2.md)
+Your AWS account comes with a default scheduler group. You can access default group in CDK with:
+
+```text
```
kaizencc, line 103: these examples can be in `ts` now? Since they should be fully implemented right. Then you will also have to double check with `yarn rosetta:extract --strict` or else the PR build may fail :)

packages/@aws-cdk/aws-scheduler-alpha/README.md
```diff
@@ -97,7 +98,31 @@ const oneTimeSchedule = new Schedule(this, 'Schedule', {
...
+const target = new targets.LambdaInvoke(props.func, {
+    input: ScheduleTargetInput.fromObject({
+        "payload": "useful",
+    }),
+});
+
+new Schedule(this, 'Schedule', {
+    scheduleExpression: ScheduleExpression.rate(Duration.minutes(10)),
+    target,
+    group
```
kaizencc, line 123:
```suggestion
    group,
```

packages/@aws-cdk/aws-scheduler-alpha/README.md
```diff
@@ -164,4 +189,26 @@ TODO: Not yet implemented. See section in [L2 Event Bridge Scheduler RFC](https:
...
+    metric: group.metricAllErrors(),
+    threshold: 0
+});
+
+// Or use default group
+const defaultGroup = Group.fromDefaultGroup(this, "DefaultGroup");
+new Alarm(this, 'DefaultGroupErrorAlarm', {
+    metric: defaultGroup.metricAllErrors(),
+    threshold: 0
+});
```
kaizencc, line 209:
```suggestion
const group = new Group(this, "Group", {
    groupName: "MyGroup",
});

new Alarm(this, 'MyGroupErrorAlarm', {
    metric: group.metricAllErrors(),
    threshold: 0,
});

// Or use default group
const defaultGroup = Group.fromDefaultGroup(this, "DefaultGroup");
new Alarm(this, 'DefaultGroupErrorAlarm', {
    metric: defaultGroup.metricAllErrors(),
    threshold: 0,
});
```

packages/@aws-cdk/aws-scheduler-alpha/lib/private/schedule.ts
```diff
@@ -36,19 +37,31 @@ export interface ScheduleProps {
...
+   * @deafult - By default a schedule will be associated with the `default` group.
+   */
+  readonly group?: IGroup;
 }
 
 /**
  * An EventBridge Schedule
  */
 export class Schedule extends Resource implements ISchedule {
+  group?: IGroup;
```
kaizencc, line 53:
```suggestion
  public readonly group?: IGroup;
```

packages/@aws-cdk/aws-scheduler-alpha/README.md
filletofish: Updated now, good catch

kaizencc (approved):

kaizencc (approved):

PR #26241, opened 2023-07-05
feat(glue): validate maxCapacity, workerCount, and workerType

go-to-k:
This PR adds the following validations, allows errors to be output in the synth phase instead of the deploy phase.

- The case that maxCapacity and workerType (and workerCount) are specified.
  - In this [doc](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-jobs-job.html#aws-glue-api-jobs-job-Job), `Do not set MaxCapacity if using WorkerType and NumberOfWorkers.`
- The case for maxCapacity with GlueVersion 2.0 or later.
  - In this [doc](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-jobs-job.html#aws-glue-api-jobs-job-Job), `For Glue version 2.0 or later jobs, you cannot specify a Maximum capacity`
- The case that only either workerType or workerCount is specified.
  - Then an error occurs in CloudFormation, `Please set both Worker Type and Number of Workers.`

----


packages/@aws-cdk/aws-glue-alpha/lib/job.ts
```diff
@@ -719,6 +719,16 @@ export class Job extends JobBase {
       }
     }
 
+    if (props.maxCapacity !== undefined && (props.workerType || props.workerCount !== undefined)) {
```
colifran, line 722: Looking for some clarification on this - the documentation says that `maxCapacity` cannot be used when setting `workerType` and `workerCount`. Why are we only checking for `workerType` or `workerCount` instead of `workerType` and `workerCount`?

colifran (commented):
@go-to-k this looks great! I just have one small question.

packages/@aws-cdk/aws-glue-alpha/lib/job.ts
go-to-k: @colifran 

Thanks for your review!

I was really worried.

If "and" is used, in the case where `maxCapacity` is specified but only one of `workerType` or `workerCount` is specified, it will break through this branch and throw the error `Both workerType and workerCount must be set`.
I did this because I thought that the essential error, the collision between `maxCapacity` and `workerType `, should be fixed first.

But it's certainly possible that there could be some confusion due to the deviation from the document, so I'll fix it!

packages/@aws-cdk/aws-glue-alpha/lib/job.ts
```diff
@@ -719,6 +719,16 @@ export class Job extends JobBase {
       }
     }
 
+    if (props.maxCapacity !== undefined && (props.workerType && props.workerCount !== undefined)) {
```
colifran, line 722: Sorry for not thinking of this earlier, but I have one more thing I think we should add. There is a small chance that `maxCapacity` and `workerCount` could be `Tokens`. Could we wrap this validation in an if statement that first checks if both `maxCapacity` and `workerCount` are resolved. As an example:

```ts
if (!Token.isUnresolved(props.maxCapacity) && !Token.isUnresolved(props.workerCount)) {
  // your code here
}
```

We should probably also include an associated unit test for this as well. I hope this makes sense. Let me know if you have any questions.

colifran (changes requested):
Your change looks good. There is one other minor piece I didn't think about earlier that we should add.

packages/@aws-cdk/aws-glue-alpha/lib/job.ts
go-to-k: @colifran 

> There is a small chance that maxCapacity and workerCount could be Tokens.

I see. However, I assume that `props. maxCapacity` and `props. workerCount` are input values from the user, not values generated from the resource, but when do they become Unresolved? Depending on that, I will consider new unit tests.

colifran: @go-to-k I realize it's a small chance, but it's better to add the check for it now since we're adding new code rather than having to go back and fix a bug. Basically, this would only occur if someone were to provide those as `Lazy` values. As an example:

```ts
{
  workerCount: cdk.Lazy.number({ produce: () => 5 }),
  maxCapacity: cdk.Lazy.number({ produce: () => 5 }),
}
```

mrgrain: Or when using a `Parameter` it can also happen.

go-to-k: @colifran 

Thanks, I understand.

In the following code, when `workerCount` and `maxCapacity` are `Token`, it does not validate.

```ts
if (!Token.isUnresolved(props.maxCapacity) && !Token.isUnresolved(props.workerCount)) {
  // your code here
}
```

However, even if the value is a `Token` and `Unresolved`, the fact that you are specifying that combination of parameters is itself a problem (i.e., specifying both `workerCount` and `maxCapacity`, regardless of the value, is itself a problem).

Therefore, shouldn't they be validated whether they are tokens or not?

I think we can check if the parameters are undefined or not, even if encoded numbers as `Token`.

colifran: @go-to-k good catch. I agree with this. We're checking for just the existence of the properties and not actually validating the value of the properties. That said, I think this is good to go. Thanks for your effort on this!

colifran (dismissed):

colifran (approved):

Issue #26075, opened 2023-06-21
core: logical ID validation regex is inconsistant with cloudformation specification


ashwin300:
### Describe the bug

The [CloudFormation documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html) indicates that logical ID's must be only alphanumeric, yet the regex in [the logical ID validation file](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/private/logical-id.ts#L72) is /^[A-Za-z][A-Za-z0-9]{1,254}$/. This causes logical IDs such as 030ffab42b5f448fb39750B452 to fail even though they are valid according to the CloudFormation standard.

### Expected Behavior

Logical IDs such as 030ffab42b5f448fb39750B452 to be considered valid to the CDK when synthesizing a stack

### Current Behavior

Logical IDs such as 030ffab42b5f448fb39750B452 are not considered valid when synthesizing a stack. 

Error: Resolution error: Logical ID must adhere to the regular expression: /^[A-Za-z][A-Za-z0-9]{1,254}$/, got '030ffab42b5f448fb39750B452'.

### Reproduction Steps

1. initialize sample cdk app
2. in the stack, add the following code: 
```
import * as cdk from 'aws-cdk-lib';
import { Construct } from 'constructs';
import * as sqs from 'aws-cdk-lib/aws-sqs';

export class CdkErrorTestStack extends cdk.Stack {
  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
    super(scope, id, props);

    // The code that defines your stack goes here

    // example resource
    const queue = new sqs.Queue(this, '030ffab42b5f448fb39750B452', {
      visibilityTimeout: cdk.Duration.seconds(300)
    });
  }
}

```

run cdk synth

### Possible Solution

Change the logical ID validation regex to fit CloudFormation documentation, or change the documentation to reflect the regex validation. 

### CDK CLI Version

2.85.0 (build 4e0d726)

### Node.js Version

v14.21.3

### OS

Linux

### Language

Typescript

peterwoodworth:
Thanks for reporting this, I think it's likely that CloudFormation used to have this constraint, but does not anymore. We'll see what we can do about this

peterwoodworth:
I just tried deploying through the console and CLI with a number at the beginning, and I'm getting this error:

```
[...]

PR #26251, opened 2023-07-05
fix(core): updates logical id validation regex

ashwin300:
Motivation
The [CloudFormation documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html) indicates that logical ID's must be only alphanumeric, yet the regex in [the logical ID validation file](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/private/logical-id.ts#L72) is /^[A-Za-z][A-Za-z0-9]{1,254}$/. This causes logical IDs such as 030ffab42b5f448fb39750B452 to fail even though they are valid according to the CloudFormation standard.

Expected Behavior
Logical IDs such as 030ffab42b5f448fb39750B452 to be considered valid to the CDK when synthesizing a stack

Current Behavior
Logical IDs such as 030ffab42b5f448fb39750B452 are not considered valid when synthesizing a stack.

Error: Resolution error: Logical ID must adhere to the regular expression: /^[A-Za-z][A-Za-z0-9]{1,254}$/, got '030ffab42b5f448fb39750B452'.

Fix
Removes the constraint from the logical ID regex that denotes that the first character of a logical ID must be a letter.

Closes https://github.com/aws/aws-cdk/issues/26075.

----

### DIFF
--- a/packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
+++ b/packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
@@ -17,6 +17,10 @@ Flags come in three types:
 
 | Flag | Summary | Since | Type |
 | ----- | ----- | ----- | ----- |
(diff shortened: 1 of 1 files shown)


mrgrain:
@ashwin300 Could you add a simple integration test that uses a number as logical id?

PR #26254, opened 2023-07-06
fix(apigateway): update imported RestAPIs type to grant access to RestAPIBase methods

### LINKED ISSUE #26247: (apigateway): Imported RestAPIs do not have access to its methods due to casting
### Describe the bug

Imported RestAPIs are returned by CDK as `IRestAPI`. However in the source code, the imports extend `RestAPIBase` which features additional helpful methods, such as the `metric` methods. The `IRestAPI` interface does not come with methods such as the `metric` methods, which prevents methods from being called without casting the imported RestAPI as a `RestAPIBase`.

### Expected Behavior

I expect to be able to use methods that are okay to use on an imported RestAPI.

### Current Behavior

I cannot use some methods that I should be able to

### Reproduction Steps

You can successfully call the method using the following code. Note that removing the typecasting causes this to fail

```ts
export class TestingStack extends cdk.Stack {
  public restApi: apigw.RestApiBase;
    super(scope, id, props);

    this.restApi = apigw.RestApi.fromRestApiId(this,"MyGatewayApi","restApiId") as apigw.RestApiBase;
    
    const api = new cw.Alarm(this, `api-error-alarm`, {
      alarmName: 'api-error-alarm',
      threshold: 5,
      evaluationPeriods: 1,
      metric: this.restApi.metricServerError()
    });
  }
}
```

### Possible Solution

Probably put these methods on `IRestAPI` interface, but I didn't dive too deep into this so that could have other side effects.

_No response_

### CDK CLI Version

latest

### Framework Version

_No response_

### Node.js Version

16

### OS

mac

### Language

Typescript

### Language Version

_No response_

### Other information

Sourced from [this discussion](https://github.com/aws/aws-cdk/discussions/26234)

lpizzinidev:
`fromRestApiId` and `fromRestApiAttributes` were returning the imported RestAPI as `IRestApi`, preventing methods implemented by `RestAPIBase` from being called without type casting.
This fix solves the problem by changing the return type of the two functions.

Closes #26247.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-apigateway/lib/restapi.ts
+++ b/packages/aws-cdk-lib/aws-apigateway/lib/restapi.ts
@@ -735,7 +735,7 @@ export class RestApi extends RestApiBase {
   /**
    * Import an existing RestApi.
    */
-  public static fromRestApiId(scope: Construct, id: string, restApiId: string): IRestApi {
+  public static fromRestApiId(scope: Construct, id: string, restApiId: string): RestApiBase {
     class Import extends RestApiBase {
       public readonly restApiId = restApiId;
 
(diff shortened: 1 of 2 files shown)


lpizzinidev:
Exemption Request.
Let me know if integration tests or other changes are needed.

mrgrain:
Hey thanks for the PR, unfortunately we can't go forward with this PR like this. Here's why:


`*Base` types are not considered "public" and should not be used as return or property types.
I appreciate that in the past we might have not been consistent with that. 

You are welcome to open a new PR with a different approach. As this is a `p2` issue, we cannot guarantee to have time to work with you on an alternative solution. Please reach out to the wider community for support.

PR #26265, opened 2023-07-06
fix(cli): specifying --quiet does not suppress asset building and publishing logs

fwang:
We're currently maintaining a patched version of CDK to use with [SST](https://docs.sst.dev/) because this flag isn't fully respected. When the `quiet` flag is specified, the `docker build` and `docker push` shell commands still print to the terminal. This PR propagates the `quiet` flag down to the `cdk-assets`'s `container-images` handler.

Would appreciate this functionality - would greatly simplify things on our end

----

### DIFF
--- a/packages/aws-cdk/lib/util/asset-publishing.ts
+++ b/packages/aws-cdk/lib/util/asset-publishing.ts
@@ -54,6 +54,7 @@ export async function publishAssets(
     publishInParallel: options.parallel ?? true,
     buildAssets: options.buildAssets ?? true,
     publishAssets: true,
+    quiet: options.quiet,
   });
   await publisher.publish();
   if (publisher.hasFailures) {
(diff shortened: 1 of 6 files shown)


mrgrain:
Looks good! Can you sort out the build failure and add/extend a test to cover this new use case?

There's an existing test but I'm not even sure it does test `--quiet` properly. 🙈 
 https://github.com/aws/aws-cdk/blob/a523e9fc81021897285b656c49c1e71d888a38f7/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts#L725

mrgrain:
@mergifyio update

mrgrain:
Closed in favor of https://github.com/aws/aws-cdk/pull/26493

PR #26271, opened 2023-07-06
fix(integ-tests-alpha): incorrect sdk client resolution

MrArnoldPalmer:
Fixes some issues with finding and calling v3 sdk clients.

Exports some utilities from `aws-cdk-lib.AwsCustomResource` for finding and calling `@aws-sdk/client-*` modules to avoid duplicating this logic.

----

### DIFF
--- a/packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/lambda-handler/sdk.ts
+++ b/packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/lambda-handler/sdk.ts
@@ -57,6 +57,7 @@ function getServiceClient(sdkPkg: V3SdkPkg): any {
     const ServiceClient = findV3ClientConstructor(sdkPkg.pkg);
     return new ServiceClient({});
   } catch (e) {
+    console.error(e);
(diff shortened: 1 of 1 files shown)


packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/lambda-handler/sdk.ts
```diff
@@ -29,22 +30,61 @@ export function flatten(object: object): { [key: string]: any } {
...
   try {
-    // eslint-disable-next-line @typescript-eslint/no-require-imports
-    const pkg = require(clientPackageName);
-    return new pkg[service]({});
+    const ServiceClient = findV3ClientConstructor(sdkPkg.pkg);
+    return new ServiceClient({});
   } catch (e) {
-    // Just import a known client module
-    throw Error(`Service ${service} client package with name '${clientPackageName}' does not exist.`);
+    throw Error(`No client constructor found within package: ${sdkPkg.packageName}`);
```
corymhall, line 60: Can we also log the error to make it easier to troubleshoot?

corymhall (approved):
Just one suggestion

PR #26285, opened 2023-07-07
feat(cloudwatch): dashboard variables

### LINKED ISSUE #26200: cloudwatch: support dashboard variables
### Describe the feature

CloudWatch recently released the new dashboard variables feature which allows parameterizing dashboards with variables to reduce repetition/duplication of dashboards
- https://aws.amazon.com/about-aws/whats-new/2023/06/amazon-cloudwatch-dashboard-variables/
- https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_dashboard_variables.html
- https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/CloudWatch-Dashboard-Body-Structure.html#CloudWatch-Dashboard-Properties-Variables-Structure

### Use Case

This feature would enable retiring near identical/duplicate dashboards which only differ in some parameters and enable consolidating them into one single parameterized dashboard

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.84.0

macOS

Stacy-D:
This change add the support for dashboard variables in CDK https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_dashboard_variables.html. It allows to reduce the number of repeated  CloudWatch dashboards by unifying them into one managed with variables.

Closes #26200

----


Stacy-D:
Example dashboard with a search variable created as a result of added integ test:

Screenshot 2023-07-07 at 17 02 02

humanzz:
Hello reviewer,
Ana and I had reached out to the cloudwatch team to update the documentation describing the dashboard json to include dashboard variables to help you with reviewing these changes. The documentation update had just been released at https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/CloudWatch-Dashboard-Body-Structure.html#CloudWatch-Dashboard-Properties-Variables-Structure

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -704,3 +704,56 @@ const dashboard = new cw.Dashboard(this, 'Dash', {
 ```
 
 Here, the dashboard would show the metrics for the last 7 days.
+
+### Dashboard variables
+
+Dashboard variables are a convenient way to create flexible dashboards which allow to quickly display different content depending
```
kaizencc, line 710:
```suggestion
Dashboard variables are a convenient way to create flexible dashboards that display different content depending
```

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -704,3 +704,56 @@ const dashboard = new cw.Dashboard(this, 'Dash', {
...
+on the value of an input field within a dashboard. They allow to create a dashboard on which it's possible to quickly switch between
+different Lambda functions, Amazon EC2 instances etc.
+
+You can learn more about Dashboard variables in the [Amazon Cloudwatch User Guide](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_dashboard_variables.html)
+
+There are two types of dashboard variables available: a property variable and a pattern variable.
+- Property variable allows to change any JSON property in the JSON source of a dashboard like `region`. It can also change dimension name for a metric.
+- Pattern variable allows to use a regular expression pattern to change all of a JSON property or only part of it.
```
kaizencc, line 718:
```suggestion
- Property variables can change any JSON property in the JSON source of a dashboard like `region`. It can also change the dimension name for a metric.
- Pattern variables use a regular expression pattern to change all or part of a JSON property.
```

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -704,3 +704,56 @@ const dashboard = new cw.Dashboard(this, 'Dash', {
 ```
 
 Here, the dashboard would show the metrics for the last 7 days.
+
+### Dashboard variables
+
+Dashboard variables are a convenient way to create flexible dashboards which allow to quickly display different content depending
+on the value of an input field within a dashboard. They allow to create a dashboard on which it's possible to quickly switch between
+different Lambda functions, Amazon EC2 instances etc.
```
kaizencc, line 712:
```suggestion
different Lambda functions, Amazon EC2 instances, etc.
```

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -704,3 +704,56 @@ const dashboard = new cw.Dashboard(this, 'Dash', {
...
+
+You can learn more about Dashboard variables in the [Amazon Cloudwatch User Guide](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_dashboard_variables.html)
+
+There are two types of dashboard variables available: a property variable and a pattern variable.
+- Property variable allows to change any JSON property in the JSON source of a dashboard like `region`. It can also change dimension name for a metric.
+- Pattern variable allows to use a regular expression pattern to change all of a JSON property or only part of it.
+
+The **property variable** to change the `region` property can be added to the dashboard as follows:
```
kaizencc, line 720:
```suggestion
A use case of a **property variable** is a dashboard with the ability to toggle the `region` property to see the same dashboard in different regions:
```

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -704,3 +704,56 @@ const dashboard = new cw.Dashboard(this, 'Dash', {
...
+    id: 'region2',
+    label: 'RegionPattern',
+    defaultValue: 'us-east-1',
+    visible: true,
+  })],
+});
+```
+
+You can add a variable after object instantiation with the method
+`addVariable()`.
```
kaizencc, line 759:
```suggestion
`dashboard.addVariable()`.
```

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -704,3 +704,56 @@ const dashboard = new cw.Dashboard(this, 'Dash', {
...
+    id: 'region',
+    label: 'Region',
+    defaultValue: 'us-east-1',
+    visible: true,
+    values: [{ label: 'IAD', value: 'us-east-1' }, { label: 'DUB', value: 'us-west-2' }],
+  })],
+});
+```
+
+This example shows how to change `region` everywhere, assuming the current dashboard is showing region `us-east-1` already, by using **pattern variable**
```
kaizencc, line 740: I don't really understand what is going on in this example and/or the use case. I'm not sure why `value` and `defaultValue` is `us-east-1`

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
kaizencc, line 1: The format of this file is a bit wonky. Can you go back and make sure that
- the docstrings are indented correctly
- there are no extraneous new lines like line 1, 3, 21, 37, 45, 100, 104, 119, 146, 181

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,182 @@
...
+export abstract class DashboardVariable implements IVariable {
+  private readonly baseProps: IDashboardVariable;
+
+  protected constructor(props: IDashboardVariable) {
+    this.baseProps = props;
+  }
+
+  toJson(): any {
+    return {
+      [this.baseProps.type]: this.baseProps.value,
```
kaizencc, line 91: Can you explain this line for me? I'm not sure what `value` is supposed to do and I don't see this as part of the [API](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/CloudWatch-Dashboard-Body-Structure.html#CloudWatch-Dashboard-Properties-Variables-Structure)

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,182 @@
...
+     */
+  readonly searchExpression: string;
+
+  /**
+     * Optional dimension name from the search
+     */
+  readonly populateFrom?: string;
+}
+
+export class SearchDashboardVariable extends DashboardVariable {
```
kaizencc, line 165: I like this! But this should be part of the README as well.

packages/aws-cdk-lib/aws-cloudwatch/test/dashboard.test.ts
```diff
@@ -230,6 +238,80 @@ describe('Dashboard', () => {
     const template = Annotations.fromStack(stack);
     template.hasWarning('/MyStack/MyDashboard', Match.stringLikeRegexp("Math expression 'oops' references unknown identifiers"));
   });
+
+  test('has different variables', () => {
```
kaizencc, line 242: lets split this unit test up into different tests that all test one specific use case.

packages/@aws-cdk-testing/framework-integ/test/aws-cloudwatch/test/integ.dashboard-variables.ts
```diff
@@ -0,0 +1,44 @@
...
+          metricName: 'BucketSizeBytes',
+          label: '[BucketName: ${PROP(\'Dim.BucketName\')}] BucketSizeBytes',
+          statistic: Stats.MAXIMUM,
+          dimensionsMap: { StorageType: 'StandardStorage', BucketName: 'my-bucket' },
+        }),
+      ],
+    });
+
+    // The dashboard variable which changes BucketName property on the dashboard
+    dashboard.addVariable(new SearchDashboardVariable({
```
kaizencc, line 25: can we integ test ValueDashboardVariable as well please. Can be in same integ test.

packages/aws-cdk-lib/aws-cloudwatch/lib/dashboard.ts
```diff
@@ -77,6 +78,13 @@ export interface DashboardProps {
    * @default - No widgets
    */
   readonly widgets?: IWidget[][]
+
+  /**
+   * Initial list of dashboard variables
```
kaizencc, line 83:
```suggestion
   * A list of dashboard variables
   *
   * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_dashboard_variables.html#cloudwatch_dashboard_variables_types
```

kaizencc (changes requested):
Hi @Stacy-D. Thanks for this PR -- it is a great start! Just a few nits here and there and sorry that I did not like the phrase `allow to` :)

humanzz:
Hey @kaizencc 
It's great that this overall looks good to you!
@Stacy-D is taking this week off and is likely unable to address things during that time. I'll try to address them in her absence so expect changes from me soon :)

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
Stacy-D: This line will resolve to either:
```
{
property: 'BucketName'
}
```
 if value specified is 'BucketName'. In this case the inputs from property variable will result into changing of BucketName property in the Dashboard json.

if variable type is pattern, then the value field corresponds to the pattern which has to be replaced in the dashboard. Say we want to replace all occurrences of word 'Placeholder' in the dashboard with another value from the variable. Then it's possible to do if variable json has:

```
{
pattern: "Placeholder"
}
```

From the API doc: 
for the property variable, here the `value` field is the `region`
```
"variables": [
     {
        "type": "property", 
        "property": "region",
        "inputType": "input",
        "id": "region",
        "label": "Region",
        "defaultValue": "us-east-1",
        "visible": true
    } 
],
```

for the pattern variable, here the `value` field is the `us-east-1`
```
"variables": [
    {
        "type": "pattern",
        "pattern": "us-east-1",
        "inputType": "input",
        "id": "region",
        "label": "Region",
        "defaultValue": "us-east-1",
        "visible": true
    }
],
```



So overall, the `value` field controls what exactly will change in the dashboard: some string, some property, dimension name and so on

packages/aws-cdk-lib/aws-cloudwatch/README.md
Stacy-D: `defaultValue` is the one to show on the dashboard from the beginning. The `value` corresponds to the part of the Dashboard json which will be replaced as a result of using this variable. so if we look at the dashboard json the `value` part will help to generate ` "pattern": "us-east-1",` part

The API doc has the same example:

```
"variables": [
    {
        "type": "pattern",
        "pattern": "us-east-1",
        "inputType": "input",
        "id": "region",
        "label": "Region",
        "defaultValue": "us-east-1",
        "visible": true
    }
], 
```

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
Stacy-D: Should be reformatted now

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
Stacy-D: Added a readme entry for the search variable

Stacy-D:
Updated result of the integration test with both search and value variables

Screenshot 2023-07-08 at 11 40 58

packages/aws-cdk-lib/aws-cloudwatch/test/dashboard.test.ts
Stacy-D: The tests are now all split

packages/@aws-cdk-testing/framework-integ/test/aws-cloudwatch/test/integ.dashboard-variables.ts
Stacy-D: Added a ValueDashboardVariable into integ test

Stacy-D:
Hey @kaizencc,
Thanks for your review! I addressed the requested changes and responded to your questions. So I'm now looking forward a second review 😄 

All the follow ups, should those be needed, will be addressed by @humanzz while I'm off.

humanzz:
@kaizencc just a question, re the enums introduced in this PR, in previous PRs we sometimes got the feedback to prefer classes with static members to kinda future proof against introductions of new types e.g. https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-lambda/lib/architecture.ts.
Any thoughts on this?

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   * Unique id
+   */
+  readonly id: string;
+
+  /**
+   * Optional label in the toolbar
+   */
+  readonly label?: string;
```
kaizencc, line 64: for this and the optional properties below: optional props should have an `@default` tag.

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,171 @@
+export enum VariableInputType {
```
kaizencc, line 1: Assuming this and the `enum` below are what you're talking about @humanzz. For the decision between enums and enum-like classes, I consider two things:

- Is it possible/likely for additional values to be added later?
- If an additional value was to be added, would it involve changes elsewhere in the CDK?

 I can't think of an additional `VariableInputType` that would come in the future since these three UI options seem pretty standard to me. 

For `VariableType`, if a new option came in, it likely would involve changing additional things around the CDK to support whatever new type it is, since that feels like it would be a pretty significant change. Basically, it doesn't feel like it would be as simple as adding in a new `VariableType` and calling it a day. It feels like the new `VariableType` would be a new use case unlocked that would have ramifications on other properties in CDK land as well.

If any of these assumptions are off base let me know! I'm definitely not married to `enums`, but that's why I didn't make a fuss.

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Optional dimension name from the search
+   */
+  readonly populateFrom?: string;
+}
+
+export class SearchDashboardVariable extends DashboardVariable {
+  private readonly props: ISearchDashboardVariable;
+
+  constructor(props: ISearchDashboardVariable) {
```
kaizencc, line 157: From the API docs, `search` and `populateFrom` are required  "if inputType is select or radio and you are not specifying values."

Just want to confirm that `SearchDashboardVariable` will still work with inputType `input` right?

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
kaizencc: Thanks for the detailed reply @Stacy-D! I really appreciate it.

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,171 @@
...
+    };
+  }
+
+}
+
+export interface IVariableValue {
+  /**
+   * Optional label for the selected item
+   *
+   * @default - value
```
kaizencc, line 102: What does `value` mean here? It's kind of an oversaturated word... if you mean the value you specify in the 'value' prop, then lets be specific :)

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * The variable populated from the metric search
+ */
+export interface ISearchDashboardVariable extends IDashboardVariable {
+  /**
+   * Search expression
+   */
+  readonly searchExpression: string;
+  /**
+   * Optional dimension name from the search
```
kaizencc, line 149: needs a `@default`. Not sure why this wouldn't throw an error somewhere in our PR build. Will have to take a look later.

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  toJson(): any {
+    const base = super.toJson();
+    return {
+      ...base,
+      search: this.props.searchExpression,
+      populateFrom: this.props.populateFrom,
+    };
+  }
+
```
kaizencc, line 170:
```suggestion
```

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,171 @@
...
+      [this.baseProps.type]: this.baseProps.value,
+      type: this.baseProps.type,
+      inputType: this.baseProps.inputType,
+      id: this.baseProps.id,
+      defaultValue: this.baseProps.defaultValue,
+      visible: this.baseProps.visible,
+      label: this.baseProps.label,
+    };
+  }
+
```
kaizencc, line 95:
```suggestion
```

kaizencc (dismissed):
This pretty much looks good to go to me @Stacy-D @humanzz! Just a few more minor comments.

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
humanzz: Thanks for this explanation. It makes sense to me :)

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
humanzz: So, I went to try a different combo of parameters manually, and what would essentially happen if `input` inputType and we have a search variable, is that it will just be a normal `input`. However, on saving one is faced with the nice error
```
The dashboard body is invalid, there are 2 validation errors: [ { "dataPath": "/variables/0/search", "message": "Should NOT be valid" }, { "dataPath": "/variables/0/populateFrom", "message": "Should NOT be valid" } ]
```

So, to me there are 2 things to do here
1. `SearchDashboardVariable` should validate that `inputType` is not `input`
2. `ISearchDashboardVariable.populateFrom` should not be optional

I think `ValueDashboardVariable` might need some similar validations as well

humanzz:
@kaizencc made a change at https://github.com/aws/aws-cdk/pull/26285/commits/6e6bdb576992a58da6ae065588222303d559d4b5 where
- I addressed the comments
- made some renames
- added some validations + tests

Hopefully this makes things a bit clearer for the users.

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * A pattern variable is one that changes a regex pattern across the dashboard JSON
+   */
+  PATTERN = 'pattern',
+}
+
+/**
+ * A single dashboard variable
+ */
+export interface IVariable {
```
kaizencc, line 33: I didn't notice before, but why are we naming this `IVariable`? the `IXxx` naming convention has meaning in cdk land. It usually serves as a way for users to specify internal or external resources (i.e. `fromVariableArn` or something). That doesn't seem to be the case here.

>Therefore, when constructs are referenced anywhere in the API (e.g. in properties or methods of other resources or higher-level constructs), the resource interface (IFoo) should be used over concrete resource classes (Foo). This will allow users to supply either internal or external resources [awslint:ref-via-interface].

[Design Guidelines](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#construct-interface)

packages/aws-cdk-lib/aws-cloudwatch/test/dashboard.test.ts
```diff
@@ -230,6 +241,178 @@ describe('Dashboard', () => {
...
+  });
+
+  test('dashboard has property/select search variable', () => {
+    // GIVEN
+    const stack = new Stack();
+    const dashboard = new Dashboard(stack, 'Dashboard');
+
+    // WHEN
+    dashboard.addVariable(new SearchDashboardVariable({
+      defaultValue: '__FIRST',
```
kaizencc, line 316: I just realized that `__FIRST` is a special value. Can we make that into a constant so that it can be referenced like:

```ts
defaultValue: `DefaultValue.FIRST`
```

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * Options for {@link SearchDashboardVariable}
+ */
+export interface SearchDashboardVariableOptions extends DashboardVariableOptions {
+  /**
+   * A search expression that specifies a namespace, dimension name and a metric name.
+   *
+   * For example `{AWS/EC2,InstanceId} MetricName=\"CPUUtilization\"`
+   */
+  readonly searchExpression: string;
```
kaizencc, line 162: Have you given any thought to strongly typing this expression? How unique can this string be? The examples in documentation all reference a namespace, dimension, and metricName in some format. Maybe with stronger types we can build this expression out for the user and not have to have them worry about typos.

I'm thinking something like:

```ts
class searchExpression {
  public static FROM_COMPONENTS(
    namespace: string,
    dimension: string,
    metricName: string,
  ) {
    return new searchExpression(`{${namespace},${dimension}} MetricName=\"${metricName\"`);
  }

  // users can still supply raw expressions if they want
  public constructor(public readonly expression: string) {}
}
```

WDYT?

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -723,3 +723,78 @@ const dashboard = new cw.Dashboard(this, 'Dash', {
...
+    id: 'region2',
+    label: 'RegionPattern',
+    defaultValue: 'us-east-1',
+    visible: true,
+  })],
+});
+```
+
+The following example generates a Lambda function variable, with a radio button for each function. Functions are discovered by a metric query search.
+The `defaultValue` with `__FIRST` indicates that the  default value will be the first value returned from the search.
```
kaizencc, line 778:
```suggestion
The `defaultValue` with `__FIRST` indicates that the default value will be the first value returned from the search.
```

kaizencc (changes requested):
@humanzz Looks awesome! 3 last minute comments. Otherwise I'm happy with this :).

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
humanzz: Yeah. I tried to rename it in an earlier commit to `Variable` but I got jsii build errors.

I even tried to make it `DashboardVariable` and rename `DashboardVariable` to `DashboardVariableBase` but I got another error that since the interface has a behavior, it needs to follow the `Ixx` convention.

Any thoughts/guidance here?

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
humanzz: So, I'd be a bit hesitant here, without further investigation, abt how metrics with multiple dimensions work - if they're supported or not.
If you think it's needed, I can give that a try and see if that's supported, in which case the search expression would need to allow for number of dimensions >= 1

packages/aws-cdk-lib/aws-cloudwatch/test/dashboard.test.ts
humanzz: so just a class with a static constant?

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
kaizencc: Well, I don't think its the end of the world for it to be `IVariable`. If jsii says so, lets just keep it that way.

packages/aws-cdk-lib/aws-cloudwatch/test/dashboard.test.ts
kaizencc: WDYT? Maybe its unnecessary.

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
kaizencc: Great! I think if possible, a class would be a better customer experience than a raw string.

humanzz: I blame you for the more code - now it needs more reviewing 😆 

https://github.com/aws/aws-cdk/pull/26285/commits/d8cc14a9494c9b907248b2437124a7b8557b512a

I ended up refactoring how search values are modeled!

packages/aws-cdk-lib/aws-cloudwatch/test/dashboard.test.ts
humanzz: Well, I made it take a bit more responsibility in https://github.com/aws/aws-cdk/pull/26285/commits/d8cc14a9494c9b907248b2437124a7b8557b512a, let me know what you think

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
humanzz: In fact, with this new change, I'm now wondering if we really need a separate `ValueDashboardVariable` and `SearchDashboardVariable`. The difference can really be just about the `values` prop. Something like `Values.fromSearch` and `Values.fromStatic` (or any better name).

Worth pursuing?

humanzz: I went ahead and did that. I believe it looks better.
Let me know what you think.
I also believe docs/README/error messages might need a good look to ensure they make sense for users.

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,231 @@
...
+export abstract class Values {
+  /**
+   * Create values from the components of search expression
+   *
+   * @param namespace the namespace to be used in the search expression
+   * @param dimensions the list of dimensions to be used in the search expression
+   * @param metricName the metric name to be used in the search expression
+   * @param populateFrom the dimension name, that the search expression retrieves, whose values will be used to populate the values to choose from
+   */
+  public static fromSearchComponents(namespace: string, dimensions: string[], metricName: string, populateFrom: string): Values {
```
kaizencc, line 66:
```ts
export interface SearchComponentProps {
  readonly namespace: string,
  readonly dimensions: string[],
  readonly metricName: string,
  readonly populateFrom: string,
}

export class Values {
  public static fromSearchComponents(props: SearchComponentProps): Values {
  }
}
```

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,231 @@
...
+      populateFrom: this.populateFrom,
+    };
+  }
+}
+
+/**
+ * Default value for use in {@link DashboardVariableOptions}
+ */
+export class DefaultValue {
+
```
kaizencc, line 135:
```suggestion
```

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -723,3 +723,77 @@ const dashboard = new cw.Dashboard(this, 'Dash', {
...
+
+const dashboard = new cw.Dashboard(this, 'Dash', {
+  defaultInterval: Duration.days(7),
+  variables: [new cw.SearchDashboardVariable({
+    id: 'functionName',
+    type: cw.VariableType.PATTERN,
+    label: 'Function',
+    inputType: cw.VariableInputType.RADIO,
+    value: 'originalFuncNameInDashboard',
+    values: cw.Values.fromSearchComponents('AWS/Lambda', ['FunctionName'], 'Duration', 'FunctionName'), // equivalent to cw.Values.fromSearch('{AWS/Lambda,FunctionName} MetricName=\"Duration\"', 'FunctionName'),
```
kaizencc, line 792:
```suggestion

    // equivalent to cw.Values.fromSearch('{AWS/Lambda,FunctionName} MetricName=\"Duration\"', 'FunctionName'),
    values: cw.Values.fromSearchComponents('AWS/Lambda', ['FunctionName'], 'Duration', 'FunctionName'),
```

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,231 @@
...
+   * @default - true
+   */
+  readonly visible?: boolean;
+}
+
+/**
+ * Dashboard Variable
+ */
+export class DashboardVariable implements IVariable {
+
```
kaizencc, line 209:
```suggestion
```

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,231 @@
...
+  public constructor(private readonly options: DashboardVariableOptions) {
+    if (options.inputType != VariableInputType.INPUT && !options.values) {
```
kaizencc, line 211:
```suggestion
    if (options.inputType !== VariableInputType.INPUT && !options.values) {
```

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,231 @@
...
+      throw new Error(`Variable with inputType (${options.inputType}) requires values to be set`);
+    }
+    if (options.inputType == VariableInputType.INPUT && options.values) {
+      throw new Error('Unsupported inputType INPUT. Please choose either SELECT or RADIO');
```
kaizencc, line 215:
```suggestion
      throw new Error('inputType INPUT cannot be combined with values. Please choose either SELECT or RADIO or remove \'values\' from the configuration.');
```

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -723,3 +723,77 @@ const dashboard = new cw.Dashboard(this, 'Dash', {
...
+    value: 'us-east-1',
+    defaultValue: cw.DefaultValue.of('us-east-1'),
+    visible: true,
+  })],
+});
+```
+
+The following example generates a Lambda function variable, with a radio button for each function. Functions are discovered by a metric query search.
+The `values` with `cw.Values.fromSearchComponents` indicates that the values will be populated from `FunctionName` values retrieved from the search expression `{AWS/Lambda,FunctionName} MetricName=\"Duration\"`.
+The `defaultValue` with `cw.DefaultValue.FIRST` indicates that the  default value will be the first value returned from the search.
```
kaizencc, line 779:
```suggestion
The `defaultValue` with `cw.DefaultValue.FIRST` indicates that the default value will be the first value returned from the search.
```

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,231 @@
...
+   * @default - no values
+   */
+  readonly values?: Values;
+
+  /**
+   * Optional default value
+   *
+   * @default - no default value is set
+   */
+  readonly defaultValue?: DefaultValue;
```
kaizencc, line 195: I feel like the word value is oversaturated in this PR... would it make more sense to name this `default`? Or do you want to keep `value`, `values`, and `defaultValue`?

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
```diff
@@ -0,0 +1,231 @@
...
+   */
+  public static readonly FIRST = new DefaultValue('__FIRST');
+
+  /**
+   * Create a default value
+   * @param value the value to be used as default
+   */
+  public static of(value: any) {
+    return new DefaultValue(value);
+  }
```
kaizencc, line 147: I don't love `of`. Reason being that we're not creating a default value 'of' something. We're just referencing a value. Some options:

```ts
DefaultValue.string('us-east-1')
DefaultValue.value('us-east-1') // contributes to overkill of the word value though
```

I don't really care. I just dont think `of` is right.

kaizencc (changes requested):
Okay! I think this is the last review. We're almost there. Thanks for bearing with me @humanzz

humanzz:
> Okay! I think this is the last review. We're almost there. Thanks for bearing with me @humanzz

no worries at all. I think we're making good progress here.

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
humanzz: In this one, I kinda feel `defaultValue` might be the clearest!

packages/aws-cdk-lib/aws-cloudwatch/lib/variable.ts
humanzz: I honestly can't think of a better option than `of` - except maybe the overkill `value` you suggested :)
`string` doesn't align well with the fact that it's actually `any`, with the documentation saying `Type: String, Number, or Boolean, depending on the type value for this variable`

I think I'll go with `value`

kaizencc (approved):
Sweet @humanzz! How wonderful to get multiple rounds of review bashed out in a day :). Really appreciate the fast turnaround time!

humanzz:
Awesome @kaizencc!
Thanks for the quick and actionable reviews!
Let's hope we catch the next version train 🚄 ✨

PR #26297, opened 2023-07-08
feat(cli): add diff message on the number of stacks with differences

### LINKED ISSUE #10417: [cli] cdk diff difficult to use programmatically
We're integrating CDK into a build pipeline and finding it difficult to differentiate between these possibilities:

1. CDK diff detected no changes
1. CDK diff detected changes
1. CDK diff failed (wrong credentials, etc.)

When running `cdk diff --fail`, we can check the return code to differentiate case 1 vs. 2/3, however we can't differentiate between diff detecting changes and failing for some other type of error (they both return exitCode 1).

Additionally, it appears that we can't rely on whether output was printed to stdout or stderr - from a quick read through the code it appears diffs are printed to stderr along with errors as well.

Is there a supported mechanism to call the `cdk diff` command and differentiate the options above?

### Use Case

We're integrating CDK into a build pipeline and would like to first determine if there are any differences before requiring an approval step for deployment.

### Proposed Solution

A few options might be possible:
* Return different exit codes with `cdk diff --fail` to differentiate "there are differences" from "there was an error calculating differences".
* Use stdout for output of differences and stderr just for error reporting.

### Other

---

krzysztof-slowinski:
Reported issue with `diff` is that it treats the fail return status for cases when there are actual diffs, making it hard to know what happened with pipeline automations.

The proposed solution adds a logging statement with a similar format that is used for deploy (but here the total time is reported) specifying how many stacks have differences, as presented below. As a result, it will be possible to check in pipelines for this logging statement to correctly detect situation when there are no actual changes, when there are changes, and when there are failures, since on failures this statement will not be present:

Case of no changes:

Case of changes in 5 stacks:
✨  Number of stacks with differences: 5

Closes #10417.

----

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/bootstrapping.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/bootstrapping.integtest.ts
@@ -255,6 +255,17 @@ integTest('can use the custom permissions boundary to bootstrap', withoutBootstr
   expect(template).toContain('permission-boundary-name');
 }));
 
+integTest('can use the custom permissions boundary (with slashes) to bootstrap', withoutBootstrap(async (fixture) => {
+  let template = await fixture.cdkBootstrapModern({
+    // toolkitStackName doesn't matter for this particular invocation
+    toolkitStackName: fixture.bootstrapStackName,
+    showTemplate: true,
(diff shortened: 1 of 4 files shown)


mrgrain:
> CLI code has changed. A maintainer must run the code through the testing pipeline

8b2cb7db-8c9a-4002-9636-1d3613dac38f

mrgrain (approved):

PR #26303, opened 2023-07-10
fix(servicecatalog): product stack asset bucket with forced encryption

### LINKED ISSUE #26302: aws-servicecatalog: ProductStack fails with assetBucket that requires encryption
### Describe the bug

Clients account setup strictly required that every bucket is encrypted. It also forces Bucket Policy to deny any unencrypted PutObject actions. When suppling ProductStack with such a Bucket as assetBucket the deployment fails on Custom Resource Lambda not being able to synch asset files into assetBucket (Access Denied).

### Expected Behavior

I would expect that assets will be encrypted when the asset bucket has encryption enabled.

### Current Behavior

Under the hood ProductStack uses BucketDeployment construct without encryption which results in AccessDenied. I does not considers Bucket encryption settings.

### Reproduction Steps

Create a ProductStack that includes a Lambda (to use assets). Provide as assetBucket a Bucket with SSE_KMS encryption enabled and BucketPolicy configured to deny any unencrypted PutObjects.

### Possible Solution

I've created a PR with a possible solution in which user would provide additional property which would turn on respectful options on BucketDeployment to enable encryption. This could also be implicitly learnt from assetBucket configuration but I feel it better to leave this decision to the user.

_No response_

### CDK CLI Version

2.87.0

### Framework Version

_No response_

### Node.js Version

18.16.1

### OS

MacOS

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

wolanlu:
Clients account setup strictly required that every bucket is encrypted. It also forces Bucket Policy to deny any unencrypted PutObject actions. When suppling ProductStack with such a Bucket as assetBucket the deployment fails on Custom Resource Lambda not being able to synch asset files into assetBucket (Access Denied).

Expected Behavior
I would expect that assets will be encrypted when the asset bucket has encryption enabled.

Current Behavior
Under the hood ProductStack uses BucketDeployment construct without encryption which results in AccessDenied. I does not considers Bucket encryption settings.

Possible Solution
This PR is a possible solution in which user would provide additional property `serverSideEncryption` along with encryption key `serverSideEncryptionAwsKmsKeyId` which would turn on respectful options on BucketDeployment to enable encryption. This could also be implicitly learnt from assetBucket configuration but I feel it better to leave this decision to the user - but I am open to change implementation if you thing this is better approach.

Closes #26302.

----


mrgrain (dismissed):

wolanlu:
@mergifyio refresh

Issue #22468, opened 2022-10-12
(secretsmanager): Partial ARN used in policies when Secret passed across environments


gshpychka:
### Describe the bug

When a `Secret` is created and then passed to a cross-env stack (provided the secret has a specified physical name), its internal `arnForPolicies` is its partial ARN without the 6-character suffix. This is wrong, as the generated policies from e.g. `grantRead` cannot use the partial ARN as is - they have to append `-??????` to it.

### Expected Behavior

When doing `Secret.grantRead` in a cross-environment stack, the resulting CF template contains the partial ARN of the secret followed by `-??????`.

### Current Behavior

The partial ARN is used as is.

### Reproduction Steps

Here's a simple Python app:

```python
import os
import aws_cdk as cdk
from aws_cdk import aws_iam as iam, aws_secretsmanager as sm
​
app = cdk.App()
​
root_env = cdk.Environment(
    account=os.environ["CDK_DEFAULT_ACCOUNT"], region="us-east-1"
)
cross_region_env = cdk.Environment(
    account=os.environ["CDK_DEFAULT_ACCOUNT"], region="us-west-2"
)
​
producer_stack = cdk.Stack(app, "producer", env=root_env)
consumer_stack = cdk.Stack(app, "consumer", env=cross_region_env)
​
secret = sm.Secret(producer_stack, "secret", secret_name="MySecret")
role = iam.Role(consumer_stack, "role", assumed_by=iam.ServicePrincipal("foo"))
​
secret.grant_read(role)
​
app.synth()
```

Relevant excerpt of the synthed template of the consumer:

```json
  "roleDefaultPolicy7C980EBA": {
   "Type": "AWS::IAM::Policy",
   "Properties": {
    "PolicyDocument": {
     "Statement": [
      {
       "Action": [
        "secretsmanager:DescribeSecret",
        "secretsmanager:GetSecretValue"
       ],
       "Effect": "Allow",
       "Resource": {
        "Fn::Join": [
         "",
         [
          "arn:",
          {
           "Ref": "AWS::Partition"
          },
          ":secretsmanager:us-east-1:[MY_ACCOUNT_ID_REDACTED]:secret:MySecret"
         ]
        ]
       }
      }
     ],
     "Version": "2012-10-17"
    },
    "PolicyName": "roleDefaultPolicy7C980EBA",
```

### Additional Information/Context

In addition, the secret's `secretFullArn` prop evaluates to the partial ARN at synth time (in the consumer stack, that is).

[...]

PR #26308, opened 2023-07-10
fix(secretsmanager): `arnForPolicies` evaluates to the partial ARN if accessed from a cross-env stack

gshpychka:
Currently, `Secret.secretFullArn` returns the partial ARN if the secret is referenced between cross-env stacks.
An obvious practical implication is that `grant*` methods will produce an incorrect ARN for the IAM policies, since a full ARN is required for cross-environment access.

This PR partially fixes the issue - I reimplemented `arnForPolicies` to be lazily evaluated. It checks if the value is being accessed across environments and adds `-??????` to the ARN if it is.

Now, this does not solve the underlying issue of `secretFullArn` returning the partial ARN. While it should return undefined, we have to check how the prop is accessed (same environment or cross-env) before we know whether to return the ARN or `undefined`. If we use a `Lazy` here, it still cannot return `undefined` (only `any` as the closest thing).

So I don't think the underlying cause could be solved currently, that's why I opted for this partial fix that would address the most practical consequence of the bug.
 
This is a partial fix for https://github.com/aws/aws-cdk/issues/22468.

----


gshpychka:
Not sure how to go about an integration test - I've written one but then realized that the AWS account number has to be provided (and included in the snapshots) - what should I do in this case?

gshpychka:
Hm, the integ tests failed because
```
      {
       "Action": [
        "secretsmanager:DescribeSecret",
        "secretsmanager:GetSecretValue"
       ],
       "Effect": "Allow",
       "Principal": {
        "AWS": {
         "Fn::Join": [
          "",
          [
           "arn:",
           {
            "Ref": "AWS::Partition"
           },
           ":iam::",
           {
            "Ref": "AWS::AccountId"
           },
           ":root"
          ]
         ]
        },
        "Service": "ecs-tasks.amazonaws.com"
       },
       "Resource": {
        "Ref": "testsecretAttachment19AD251F"
       }
      }
```

changed to 
```
      {
       "Action": [
        "secretsmanager:DescribeSecret",
        "secretsmanager:GetSecretValue"
       ],
       "Effect": "Allow",
       "Principal": {
        "AWS": {
         "Fn::Join": [
          "",
          [
           "arn:",
           {
            "Ref": "AWS::Partition"
           },
           ":iam::",
           {
            "Ref": "AWS::AccountId"
           },
           ":root"
          ]
         ]
        }
       },
       "Resource": {
        "Ref": "testsecretAttachment19AD251F"
       }
      },
      {
       "Action": [
        "secretsmanager:DescribeSecret",
        "secretsmanager:GetSecretValue"
       ],
       "Effect": "Allow",
       "Principal": {
        "Service": "ecs-tasks.amazonaws.com"
       },
       "Resource": {
        "Ref": "testsecretAttachment19AD251F"
       }
      }
```

Any idea why this could be?

packages/aws-cdk-lib/aws-secretsmanager/lib/secret.ts
```diff
@@ -450,7 +450,18 @@ abstract class SecretBase extends Resource implements ISecret {
...
+        const consumingStack = Stack.of(context.scope);
+        if (this.stack.account !== consumingStack.account ||
+          (this.stack.region !== consumingStack.region &&
+            !consumingStack._crossRegionReferences) || !this.secretFullArn) {
+          return `${this.secretArn}-??????`;
+        } else {
+          return this.secretFullArn;
+        }
+      },
+    });
```
mrgrain, line 464: think this is the right approach to solve this 👍🏻

mrgrain:
> testsecretAttachment19AD251F

mrgrain:
(Sorry I accidently closed this)

mrgrain:
> Hm, the integ tests failed because

Which test is this?

> Any idea why this could be?

I suspect the policy deduplication mechanism (search for `mergeStatements`) doesn't think these are combinable anymore. Probably because it is now a lazy token object which are not going to be equal to each other.

gshpychka:
> Which test is this?

It is [aws-rds/test/integ.serverless-cluster-secret-rotation-custom-names](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/framework-integ/test/aws-rds/test/integ.serverless-cluster-secret-rotation-custom-names.ts) and [aws-rds/test/integ.serverless-cluster-secret-rotation
](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/framework-integ/test/aws-rds/test/integ.serverless-cluster-secret-rotation.ts)

This would make it a breaing change, though.

Regarding an integration test for the fix itself - I'm not really comfortable with making it exempt, but not sure how a env-specific integ test can possibly be implemented.

mrgrain:
> > Which test is this?
> 
> It is [aws-rds/test/integ.serverless-cluster-secret-rotation-custom-names](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/framework-integ/test/aws-rds/test/integ.serverless-cluster-secret-rotation-custom-names.ts) and [aws-rds/test/integ.serverless-cluster-secret-rotation ](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/framework-integ/test/aws-rds/test/integ.serverless-cluster-secret-rotation.ts)
> 
> This would make it a breaing change, though.
> 
> Regarding an integration test for the fix itself - I'm not really comfortable with making it exempt, but not sure how a env-specific integ test can possibly be implemented.

Back to square one then.

> Currently, Secret.secretFullArn returns the partial ARN if the secret is referenced between cross-env stacks.

Have you figured out _why_ this happens?

gshpychka:
> Have you figured out _why_ this happens?

I believe it's because the default implementation of `secretFullArn` defined in `BaseSecret` returns `this.secretArn` and that doesn't change when a secret is used across envs. It seems that this is only overridden in `.fromSecret*` methods.

gshpychka:
> > Have you figured out _why_ this happens?
> 
> I believe it's because the default implementation of `secretFullArn` defined in `BaseSecret` returns `this.secretArn` and that doesn't change when a secret is used across envs. It seems that this is only overridden in `.fromSecret*` methods.And `secretArn` [uses `Resource.getResourceArnAttribute`](https://github.com/aws/aws-cdk/blob/4f23619ac773e002c48a6bcf51bfb5d890bd2afc/packages/aws-cdk-lib/aws-secretsmanager/lib/secret.ts#L655C4-L655C4) which in across-env case [returns a formatted ARN ](https://github.com/aws/aws-cdk/blob/43013d04f3a2637b799b9a57916751c3d04b9a2f/packages/aws-cdk-lib/core/lib/resource.ts#L296) and `ref` otherwise.

We could change `secretFullArn` to use a custom function instead of `Resource.getResourceArnAttribute` and return something else in case of cross-env access, but what do we return?

mrgrain:
> We could change secretFullArn to use a custom function instead of Resource.getResourceArnAttribute and return something else in case of cross-env access, but what do we return?

Hm, so I guess that makes `secretFullArn` broken whenever we have a cross stack reference. CR would be one way, could we also deal with this special case using import/exports or a SSM parameter? For both we should have some code doing something similar already.

gshpychka:
> Hm, so I guess that makes `secretFullArn` broken whenever we have a cross stack reference.

No, it works fine with cross-stack references if in the same environment (using Cfn imports under the hood). The relevant logic for cross-stack references is implemented elsewhere.

mrgrain:
> No, it works fine with cross-stack references if in the same environment (using Cfn imports under the hood). The relevant logic for cross-stack references is implemented elsewhere.

Yeah, but could we use the same/similar logic for cross-env?

gshpychka:
> > No, it works fine with cross-stack references if in the same environment (using Cfn imports under the hood). The relevant logic for cross-stack references is implemented elsewhere.
> 
> Yeah, but could we use the same/similar logic for cross-env?

Well, Cfn imports don't work cross-env. The cross-env logic is just "resolve the references at synth time".

What do you think we should be focusing on (I'm getting a bit lost in the discussion)?
Do you think the IAM policy merging could be acceptable? Because whatever we do, seems that there's no way around this.

gshpychka:
> > 
> 
> Yeah, but could we use the same/similar logic for cross-env?

To make sure we're on the same page - detecting that it's a cross-env reference is not the issue - that's easy. Returning a correct `arnForPolicies` in this case is easy as well - this PR does that.

We can have `secretFullArn` return a Lazy as well - and then we can implement the reference with a CR. 

But the issue with merging in IAM policies would still be there in both of these cases, as far as I understand.

mrgrain:
> > > 
> > 
> > 
> > Yeah, but could we use the same/similar logic for cross-env?
> 
> To make sure we're on the same page - detecting that it's a cross-env reference is not the issue - that's easy. Returning a correct `arnForPolicies` in this case is easy as well - this PR does that.
> 
> We can have `secretFullArn` return a Lazy as well - and then we can implement the reference with a CR.
> 
> But the issue with merging in IAM policies would still be there in both of these cases, as far as I understand.

I'm with you now. Having `arnForPolicies` a Lazy is probably okay then. I'll need to set aside a bit to look into the merging code if there is a way around it.

mrgrain:
@gshpychka I can't figure out how to push to your branch. But here's the change

https://github.com/aws/aws-cdk/compare/pr/26308?expand=1

mrgrain (approved):

rv2673:
@gshpychka @mrgrain This unfortunately break my app. 
However secrets Constructs created through fromSecretCompleteArn, that are used cross region, break completely with this change, since    it is treating them as a partialArn. 

I believe this is because the condition `!this.secretFullArn ` is used in or condition and is never reached since the cross region condition is short circuiting before secretFullArn is actually checked. So even if full arn is filled it is treated as partialArn.

So either the if block should only be entered if fullArn is undefined and cross-region or arnForPolicies should be overriden in class returned fromSecretAttributes.

This is a nice fix for secrets created(not imported) cross-region which would reduce some complexity in my apps since it has this case too.

mrgrain:
@rv2673 Thanks for letting us know! Would you mind creating a new issue for this and include a code sample for the reproduction. Thank you.

rv2673:
> @rv2673 Thanks for letting us know! Would you mind creating a new issue for this and include a code sample for the reproduction. Thank you.

Done https://github.com/aws/aws-cdk/issues/26811

PR #26311, opened 2023-07-10
fix(servicecatalog): support nested stacks in product stacks

### LINKED ISSUE #24317: (aws_servicecatalog): nested stack in product stack failed in `cdk synth`
### Describe the bug

When nested stack in product stack for AWS Service Catalog, `Cannot find asset` error for nested stack's template file  occurs in `cdk synth`.

### Expected Behavior

`cdk synth` should generate nested stack's template in product stack.

### Current Behavior

```
Error: Cannot find asset at /path/to/project/root/cdk.out/ServicecatalogSampleStackSampleProductStackSampleNestedStack62A00127.nested.template.json
    at new AssetStaging (/path/to/project/root/node_modules/aws-cdk-lib/core/lib/asset-staging.js:1:1298)
    at new Asset (/path/to/project/root/node_modules/aws-cdk-lib/aws-s3-assets/lib/asset.js:1:736)
    at Object.bind (/path/to/project/root/node_modules/aws-cdk-lib/aws-s3-deployment/lib/source.js:1:1200)
    at /path/to/project/root/node_modules/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.js:1:2942
    at Array.map (<anonymous>)
    at new BucketDeployment (/path/to/project/root/node_modules/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.js:1:2923)
    at ProductStackSynthesizer.addFileAsset (/path/to/project/root/node_modules/aws-cdk-lib/aws-servicecatalog/lib/private/product-stack-synthesizer.js:1:944)
    at SampleNestedStack._prepareTemplateAsset (/path/to/project/root/node_modules/aws-cdk-lib/core/lib/nested-stack.js:1:2902)
    at defineNestedStackAsset (/path/to/project/root/node_modules/aws-cdk-lib/core/lib/private/prepare-app.js:1:845)
    at Object.prepareApp (/path/to/project/root/node_modules/aws-cdk-lib/core/lib/private/prepare-app.js:1:695)

```

### 1. Use case:

If you want to write your product stack in CDK and use nested stack for better modularity.

### 2. Code Replication:

```typescript
#!/usr/bin/env node
import {
  aws_servicecatalog as sc,
  aws_s3 as s3,
  NestedStack,
  Stack,
} from 'aws-cdk-lib'

const app = new cdk.App();

export class ServiceCatalogSampleStack extends cdk.Stack {
  constructor(scope: Construct, id: string) {
    super(scope, id);


    new sc.CloudFormationProduct(this, 'SampleProduct', {
      productName: 'Sample Product',
      owner: 'owner',
      productVersions: [
        {
          productVersionName: 'v1',
          cloudFormationTemplate: sc.CloudFormationTemplate.fromProductStack(new SampleProductStack(this, 'SampleProductStack', {
            assetBucket: new s3.Bucket(this, 'AssetBucket', {
              bucketName: 'asset-bucket-for-service-catalog-' + Stack.of(this).account
            }),
          })),
        },
      ],
    });
  }
}

export class SampleNestedStack extends NestedStack {
  constructor(scope: Construct, id: string) {
    super(scope, id);

    new s3.Bucket(this, 'SampleBucketInNestedStack');
  }
}

export class SampleProductStack extends sc.ProductStack {
  constructor(scope: Construct, id: string, props: sc.ProductStackProps) {
    super(scope, id, props);

    new SampleNestedStack(this, 'SampleNestedStack');
  }
}
new ServiceCatalogSampleStack(app, 'ServicecatalogSampleStack');
```

### Possible Solution

Create asset for nested stack's template in product stack.

_No response_

### CDK CLI Version

2.66.0 (build c96c17d)

### Framework Version

_No response_

### Node.js Version

v18.7.0

### OS

macOS Ventura 13.2.1

### Language

Typescript

### Language Version

Version 4.9.5

### Other information

_No response_

colifran:
This PR fixes a bug that prevented a nested stack from being used within a product stack. The logic in the `addFileAsset` method as part of the `ProductStackSynthesizer` was updated to forward all assets to the parent stack and uses the `FileAssetLocation` returned from the parent stack synthesizer to create a bucket `Source` from the  `bucketName` and `objectKey`.

Closes #24317

----


packages/aws-cdk-lib/aws-servicecatalog/lib/private/product-stack-synthesizer.ts
```diff
@@ -10,45 +9,42 @@ import { ProductStack } from '../product-stack';
  * Interoperates with the StackSynthesizer of the parent stack.
  */
 export class ProductStackSynthesizer extends cdk.StackSynthesizer {
-  private readonly assetBucket?: IBucket;
   private bucketDeployment?: BucketDeployment;
+  private parentAssetBucket?: IBucket;
 
-  constructor(assetBucket?: IBucket) {
+  constructor(private readonly parentDeployment: cdk.IStackSynthesizer, private readonly assetBucket?: IBucket) {
```
mrgrain, line 15: This okay in principle, but we tend to prefer props objects.
Because I have just approved an other PR that also adds stuff to this: https://github.com/aws/aws-cdk/pull/26303/files#diff-26862b8f0be4955057aed48a0d3dfabefaa27d96e95d3c5c99a3a85cc10ce770R18

Can you rebase your PR and then turn this whole thing into a prop object please?

mrgrain, line 15:
```suggestion
  constructor(private readonly parentStack: cdk.Stack, private readonly assetBucket?: IBucket) {
```
I know this is only an internal class, but I wonder if passing the parent stack is a bit cleaner as an API. Passing a synthesizer feels like we are implying that any synthesizer could be passed in here.

packages/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.ts
```diff
@@ -377,9 +377,9 @@ export class BucketDeployment extends Construct {
       serviceToken: handler.functionArn,
       resourceType: 'Custom::CDKBucketDeployment',
       properties: {
-        SourceBucketNames: cdk.Lazy.list({ produce: () => this.sources.map(source => source.bucket.bucketName) }),
-        SourceObjectKeys: cdk.Lazy.list({ produce: () => this.sources.map(source => source.zipObjectKey) }),
-        SourceMarkers: cdk.Lazy.any({
+        SourceBucketNames: cdk.Lazy.uncachedList({ produce: () => this.sources.map(source => source.bucket.bucketName) }),
+        SourceObjectKeys: cdk.Lazy.uncachedList({ produce: () => this.sources.map(source => source.zipObjectKey) }),
+        SourceMarkers: cdk.Lazy.uncachedAny({
```
mrgrain, line 382: I'm sure this is needed.... but why?

mrgrain (changes requested):

packages/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.ts
colifran: Good question.

TL;DR - this is needed to ensure the nested stack template file is added as a `Source` to the `BucketDeployment`.

Take this code block as an example:

```ts
class SampleNestedStack extends cdk.NestedStack {
      constructor(scope: Construct, id: string) {
        super(scope, id);

        new lambda.Function(this, 'HelloHandler1', {
          runtime: lambda.Runtime.PYTHON_3_9,
          code: lambda.Code.fromAsset(path.join(__dirname, 'assets')),
          handler: 'index.handler',
        });
        
        new lambda.Function(this, 'HelloHandler2', {
          runtime: lambda.Runtime.PYTHON_3_9,
          code: lambda.Code.fromAsset(path.join(__dirname, 'assetsv2')),
          handler: 'index.handler',
        });
      }
    }
```

Note: I'm not a `Lazy` expert, but the process is something like this. I had some help from @rix0rrr to figure this part out.

First, the `Code` assets from the two lambda `Functions` will be added as `Sources` to the `BucketDeployment`. Since these are evaluated with `Lazy` they are added to a cache before evaluation is carried out. At the point when the `NestedStack` is ready to add its own template, the `Lazy` logic has already assumed all assets have been added to the cache. This results in the template file for `NestedStack` being left out when `SourceBucketNames`, `SourceObjectKeys`, and `SourceMarkers` is evaluated. `Lazy.uncachedList` is basically saying don't only rely on the cache, there may be more items to evaluate.

packages/aws-cdk-lib/aws-servicecatalog/lib/private/product-stack-synthesizer.ts
colifran: I've updated the code to pass the `parentStack` as part of the newly added `ProductStackSynthesizerProps`

packages/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.ts
rix0rrr: Btw, probably not necessary for the markers. Doesn’t really harm, but maybe put it back? And add a comment saying that its uncached because the contents may change for nestedstacks in a productstacj

mrgrain (approved):

PR #26318, opened 2023-07-11
feat(aws-cdk-lib): use new L1 codegen

mrgrain:
Replaces the existing `cfn2ts` code generator with the new `spec2cdk` generator based on `@cdklabs/typewriter` and the separately published service spec database `@aws-cdk/aws-service-spec`.

Other neccesary changes are:

- Lowered coverage threshold for branches to 35% since the newly generated code is more expressive. Arguably we should not include generated files in the coverage calculation, but that's a bigger change. 
- Asset hashes in `cfn-include` test changed due to props now being ordered alphabetic. I've manually verified that the templates are otherwise identical.
- Removing `port` property from `neptune-alpha` as this has been removed upstream and is not functional according to the service team.
- Switched `synthetics-alpha` tests to use `testDeprecated()` for anything using the `Canary`. This is due to the removal of the `deleteLambdaResourcesOnCanaryDeletion` prop in the CFN spec. The new codegen marks removed props as deprecated. We will replace this feature with a custom resource soon.

----

### DIFF
--- a/packages/@aws-cdk/app-staging-synthesizer-alpha/README.md
+++ b/packages/@aws-cdk/app-staging-synthesizer-alpha/README.md
@@ -249,6 +249,22 @@ const app = new App({
 });
 ```
 
+### Auto Delete Staging Assets on Deletion
+
+By default, the staging resources will be cleaned up on stack deletion. That means that the
+S3 Bucket and ECR Repositories are set to `RemovalPolicy.DESTROY` and have `autoDeleteObjects`
+or `autoDeleteImages` turned on. This creates custom resources under the hood to facilitate
+cleanup. To turn this off, specify `autoDeleteStagingAssets: false`.
+
+```ts
+const app = new App({
+  defaultStackSynthesizer: AppStagingSynthesizer.defaultResources({
+    appId: 'my-app-id',
+    autoDeleteStagingAssets: false,
+  }),
+});
+```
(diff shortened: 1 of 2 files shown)


mrgrain:
Exempting from linter rules since passing all existing tests is proof this works.
No README change required either since this is an internal tool.

kaizencc (approved):

bissli82:
@mrgrain any good reason why removing the port support is a necessity?

We use it and want to be able to customize neptune port, IMHO changing such functionality without backward compatibility damages Netpune credibility and functionality.

mrgrain:
Hey @bissli82 Thanks for reaching out! Looks like Neptune now offers a replacement `DBPort` for this. 🎉 
LMK if you want to contribute this back in, otherwise I'll try to get to it this week.

Please also note that `aws-cdk/aws-neptune-alpha` is currently an experimental module does not guarantee backwards compatibility. However this call-out is entirely reasonable of you!

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-neptune-dbcluster.html#cfn-neptune-dbcluster-port

image

bissli82:
Thanks!
We use alpha since we don't want to work with L1 constructs, and are heavily invested in CDK.
Should we expect dbport available in alpha as well?

PR #26321, opened 2023-07-11
chore(app-staging-synthesizer): document and test `autoDeleteStagingAssets: false`

kaizencc:
The motivation here is that I went to add this option and realized that it already existed. However, setting the property to `false` wasn't tested or documented. This PR does both of those things.

----


mrgrain (approved):

mrgrain:
@kaizencc failing build

PR #26349, opened 2023-07-13
fix(stepfunctions-tasks): specify tags in BatchSubmitJob properties

### LINKED ISSUE #26336: aws-stepfunctions-tasks: Tag Batch Jobs
### Describe the feature

Currently there's no way to tag a Batch job with `BatchSubmitJob` construct properties.
But looks like it's supported using AWS Console.

### Use Case

Categorize / manage resources used, here Batch jobs.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.74.0 (build 70e2a33)

MacOS

lpizzinidev:
This change adds the possibility to specify the [`tags`](https://docs.aws.amazon.com/batch/latest/APIReference/API_SubmitJob.html#:~:text=Required%3A%20No-,tags,-The%20tags%20that) property in the `BatchSubmitJob` construct.

Closes #26336.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
+++ b/packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
@@ -1,7 +1,7 @@
 import { Construct } from 'constructs';
 import { ISecurityGroup, SubnetSelection } from '../../../aws-ec2';
 import { FargateService, FargateTaskDefinition } from '../../../aws-ecs';
-import { FeatureFlags } from '../../../core';
+import { FeatureFlags, Token } from '../../../core';
 import * as cxapi from '../../../cx-api';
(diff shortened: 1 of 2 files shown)


packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/batch/submit-job.ts
```diff
@@ -337,4 +346,24 @@ export class BatchSubmitJob extends sfn.TaskStateBase {
       ResourceRequirements: resources.length ? resources : undefined,
     };
   }
+
+  /**
+   * Validate the specified tags according to
```
daschaa, line 351: I think the function is very well written and self explanatory. Do we really need the docstring? What happens if the URL will change in the future?

daschaa (approved):
This looks really good! I only left one comment regarding the docstring for a function.

mrgrain (approved):

PR #26353, opened 2023-07-13
docs(core): docs for fingerprint are incorrect

go-to-k:
This PR fixes a document for `fingerprint` in `core` module, because the caller's comments did not modify it.

----


go-to-k:
The original doc is [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/fs/fingerprint.ts#L28-L39).

```ts
/**
 * Produces fingerprint based on the contents of a single file or an entire directory tree.
 *
 * Line endings are converted from CRLF to LF.
 *
 * The fingerprint will also include:
 * 1. An extra string if defined in `options.extra`.
 * 2. The symlink follow mode value.
 *
 * @param fileOrDirectory The directory or file to fingerprint
 * @param options Fingerprinting options
 */
export function fingerprint(fileOrDirectory: string, options: FingerprintOptions = { }) {
```

packages/aws-cdk-lib/core/lib/fs/index.ts
```diff
@@ -26,10 +26,11 @@ export class FileSystem {
...
+   * Line endings are converted from CRLF to LF.
+   *
    * The fingerprint will also include:
    * 1. An extra string if defined in `options.extra`.
-   * 2. The set of exclude patterns, if defined in `options.exclude`
-   * 3. The symlink follow mode value.
+   * 2. The symlink follow mode value.
    *
    * @param fileOrDirectory The directory or file to fingerprint
    * @param options Fingerprinting options
```
go-to-k, line 36: The original doc is [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/fs/fingerprint.ts#L28-L39).

```ts
/**
 * Produces fingerprint based on the contents of a single file or an entire directory tree.
 *
 * Line endings are converted from CRLF to LF.
 *
 * The fingerprint will also include:
 * 1. An extra string if defined in `options.extra`.
 * 2. The symlink follow mode value.
 *
 * @param fileOrDirectory The directory or file to fingerprint
 * @param options Fingerprinting options
 */
export function fingerprint(fileOrDirectory: string, options: FingerprintOptions = { }) {
```

mrgrain (approved):

PR #26365, opened 2023-07-14
feat(core): add include property specifying particular files for assets

### LINKED ISSUE #26107: aws_lambda: support specifying included files for Code.from_asset
### Describe the feature

Add a property to `Code.from_assets` to indicate which particular files within the path should be included. If this property is present, only files matching the pattern will be considered. It is an error to specify this property if the path does not describe a directory.

### Use Case

We have a bunch of Go lambda functions in one repository. We can easily compile multiple of them in parallel with `go build -o $DIR my-module/cmd/...`. However, this places all the binaries together in one directory. We would like to be able to reference this directory in `Code.from_assets` and filter it to only the particular binary name that goes with the Lambda function in question.

Currently we have to manually arrange for each binary to be in its own folder, which is really annoying.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.85.0

Alpine 3.18

go-to-k:
This PR adds an `include` property including particular files for assets such as `lambda.Code.fromAsset`.

- Add an `include` parameter to `interface FileOptions` with `exclude`.
- Check for matching `include` along with `ignores` (actually `exclude`) in `copyDirectory`, where the actual file search is done.

Closes #26107.

----

### DIFF
--- a/packages/aws-cdk-lib/core/test/fs/fs-copy.test.ts
+++ b/packages/aws-cdk-lib/core/test/fs/fs-copy.test.ts
@@ -222,7 +222,7 @@ describe('fs copy', () => {
     ]);
   });
 
-  test('takes priority if both include and exclude are specified for the same file', () => {
+  test('exclude takes priority if both include and exclude are specified for the same file', () => {
     // GIVEN
     const outdir = fs.mkdtempSync(path.join(os.tmpdir(), 'copy-tests'));
 


packages/aws-cdk-lib/core/lib/fs/options.ts
```diff
@@ -73,6 +73,14 @@ interface FileOptions {
...
    */
   readonly ignoreMode?: IgnoreMode;
+
+  /**
+   * File paths matching the patterns will be included, based on simple glob patterns.
+   * Has no effect on Assets bundled using the `bundling` property.
+   *
+   * @default - everything is included
+   */
+  readonly include?: string[];
```
rittneje, line 83: I think it would help to add some more detail here.
1. How does this interact with `exclude`? That is, if a file matches something in `include` and `exclude`, what happens? My assumption is that `exclude` takes priority but it would be nice to mention that explicitly.
2. How does this interact with symlinks when using a `SymlinkFollowMode` other than `NEVER`? That is, if I include the symlink but not its target, what happens?

It would also be good to add test cases for these points.

go-to-k: @rittneje 

Thanks for review! I added the docs and tests.

1. `exclude` takes priority if both `include` and `exclude` are matched.
    - https://github.com/aws/aws-cdk/pull/26365/commits/5dedfcb1c8825051705353276c555b6ba02a6e56 
2. If matching the symlink but not its target, it is included even if `SymlinkFollowMode` is `NEVER` or not. This is based on the same checking the `exclude` patterns.
    - https://github.com/aws/aws-cdk/pull/26365/commits/bf41fa98defe0031267f4b585019a9483c71ff59

tmokmss:
Can't we just use `exclude` with a negation pattern to get the behavior?

```ts
include: ['index.py'],
// this should be equivalent to the following
exclude: ['*', !'index.py']
```

I am not entirely convinced that it is a good idea to have both include and exclude properties, because it is not clear what the priority is between include and exclude, and how to handle files that do not match both include and exclude, which will confuse users.

rittneje:
@tmokmss Negative exclusions aren't even documented. If that's going to be the solution (which I don't like because it's more confusing to have to think in terms of double negation), then the docs still should be amended.

tmokmss:
@rittneje 
It is deliberately supported even in glob patterns, so I don't see much reason not to use it :)

https://github.com/aws/aws-cdk/blob/eea223b52f4445e6084b1fa1fa15a3a78f83fa18/packages/aws-cdk-lib/core/lib/fs/ignore.ts#L122-L124

I think using a negation in ignore files (like .gitignore or .dockerignore) is a common pattern, but yeah, we might need a clearer doc on this.

go-to-k:
@tmokmss Hmmm, I was confused too, but you may be right! Then I would close this PR.

go-to-k:
As for the document, I'll consider it in another PR.

go-to-k:
I submitted the [PR](https://github.com/aws/aws-cdk/pull/26473) for the docs of `exclude` property using a negation.

PR #26368, opened 2023-07-14
feat(core): allow user to specify --platform

### LINKED ISSUE #25759: (core): Allow passing `platform` via `BundlingOptions`
### Describe the feature

It would be nice to be able to enforce the `platform` that any arbitrary docker bundling command uses via the [`BundlingOptions` property](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.BundlingOptions.html).

Currently, this is not possible when using an existing Docker image via the `BundlingOptions`. You can specify this when _building an image_, but not when using an image as a builder: https://github.com/aws/aws-cdk/blob/784cd0e3303ac3fd0d5270836fa9d3657ecf4e90/packages/aws-cdk-lib/aws-ecr-assets/lib/image-asset.ts#L247-L252

### Use Case

We make use of the [S3 Deployment `BucketDeployment` construct](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3_deployment-readme.html) to package a Virtual Environment dependency zipfile for EMR Serverless ([docs](https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/using-python-libraries.html)).

The process to produce the asset looks like this:

```python
pyspark_env_asset = Source.asset(
    emr_job_root,
    asset_hash=FileSystem.fingerprint(
        str(Path(emr_job_root) / "requirements.txt")
    ),
    bundling=BundlingOptions(
        image=DockerImage.from_registry(
            f"REDACTED.dkr.ecr.us-west-2.amazonaws.com/data-services:emr-serverless-base-python-3-9-16"
        ),
        environment={
            "PIP_INDEX_URL": f"https://aws:{extract_code_artifact_token()}@REDACTED.d.codeartifact.us-west-2.amazonaws.com/pypi/REDACTED/simple/",
        },
        user="root",
        command=[
            "/bin/bash",
            "-c",
            " && ".join(
                [
                    "yum install zip -y",
                    "python3.9 -m venv venv --clear --copies",
                    "source venv/bin/activate",
                    "cp -r /usr/local/lib/python3.9/* ./venv/lib/python3.9/",
                    "python3.9 -m pip install --upgrade pip",
                    "pip install -r requirements.txt",
                    # package venv to archive.
                    # **Note** that you have to supply --python-prefix option
                    # to make sure python starts with the path where your
                    # copied libraries are present.
                    # Copying the python binary to the "environment" directory.
                    "venv-pack --quiet -f -o /tmp/pyspark_env.tar.gz --python-prefix /home/hadoop/environment",
                    # CDK assets are a bit weird - we need to zip up the tarball to make sure our desired name is retained in the
                    # final S3 Bucket.
                    "cd /tmp",
                    "zip /asset-output/venv-output.zip ./pyspark_env.tar.gz",
                ]
            ),
        ],
    ),
)
```

The issue is that our `emr-serverless-base-python-3-9-16` is a [Multi-Platform image](https://docs.docker.com/build/building/multi-platform/). The EMR serverless runtime environment is `linux/amd64`, and our developer machines use `linux/arm64` (Apple Silicon) processors. When we run the build from our local machines, Docker will automatically select the `linux/arm64` variant. Therefore, as it stands, building the image from our local machines produces incompatible `venv`s.

### Proposed Solution

Expose a `platform` option in the [`BundlingOptions` property](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.BundlingOptions.html). When specified, the underlying `docker run` should use the `platform` flag as specified in the `BundlingOptions.platform` property.

### Other Information

For now, I can work around this by using the [`DockerImage.from_build` function](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.DockerImage.html#static-fromwbrbuildpath-options) which [accepts a `platform` parameter](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.DockerBuildOptions.html#platform).

So the `image` specification changes to look like this:

```python
bundling=BundlingOptions(
    image=DockerImage.from_build(
        str(Path(__file__).parent / "emr_serverless_docker"),
        platform="linux/amd64"
    ),
	# everything else from the snippet above remains the same
)
```

The `emr_serverless_docker` folder contains one file (`Dockerfile`):

```dockerfile
FROM --platform=linux/amd64 REDACTED.dkr.ecr.us-west-2.amazonaws.com/data-services:emr-serverless-base-python-3-9-16
```

(I'm not 100% sure I need to have the `--platform` specified here again, but I did it just to be safe).

This works and ensures that the `linux/amd64` version is always selected no matter the host p

### Acknowledgements

### CDK version used

2.80.0

MacOS

maafk:
Allowing user to provide a `platform` property when bundling docker assets ([ref](https://docs.docker.com/build/building/multi-platform/)).

Currently, this is not possible when using an existing Docker image via the [BundlingOptions](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.BundlingOptions.html). You can specify this when building a  [DockerImageAsset](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecr_assets.DockerImageAsset.html#platform), but not when using an image as a builder:

Closes #25759.

----


maafk:
Clarification Request - Looking for existing integration tests that I can use as an example for this use case.

I figured I could create within `aws-lambda` (using the `BundlingOptions` with the `platform` option, but I'm not seeing any example integ tests within `packages/aws-cdk-lib/aws-lambda/test` that I can steal from.

Any advice on where I can steal good integ test code from?

mrgrain (approved):

PR #26371, opened 2023-07-14
fix(ec2): flowLogName property for FlowLog construct does not work

### LINKED ISSUE #26370: ec2: flowLogName property in FlowLog construct does not work
### Describe the bug

FlowLog L2 construct has flowLogName property.
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.FlowLog.html

This is only assigned to physicalName here.
https://github.com/aws/aws-cdk/blob/9fca7909797e71788b65bd1bf7bd2c0c85a75159/packages/aws-cdk-lib/aws-ec2/lib/vpc-flow-logs.ts#L722-L724

However, we can't define physical name for `AWS::EC2::FlowLog` resource. The pysical name for VPC Flow Logs is always automatically generated id.

So, currently flowLogName property is meaningless property.

### Expected Behavior

In VPC console, a tag value for Name key is shown in Name column as below. 

flowlog-console

So, FlowLog L2 construct should add Name tag like VPC or Instance L2 construct.
https://github.com/aws/aws-cdk/blob/9fca7909797e71788b65bd1bf7bd2c0c85a75159/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts#L1450
https://github.com/aws/aws-cdk/blob/9fca7909797e71788b65bd1bf7bd2c0c85a75159/packages/aws-cdk-lib/aws-ec2/lib/instance.ts#L362

### Current Behavior

The value of flowLogName property is never referred in generated CloudFormation template.

### Reproduction Steps

We can reproduce it with the following snippet.

```typescript
    const vpc = ec2.Vpc.fromLookup(this, 'Vpc', {
      isDefault: true,
    });

    new ec2.FlowLog(this, 'FlowLog', {
      flowLogName: 'CustomFlowLogName',
      resourceType: ec2.FlowLogResourceType.fromVpc(vpc),
    });
```

### Possible Solution

Adding Name tag with the value of flowLogName property.

_No response_

### CDK CLI Version

v2.87.0

### Framework Version

_No response_

### Node.js Version

v18.16.0

### OS

Amazon Linux 2

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

tam0ri:
Currently, flowLogName property for FlowLog L2 is never referred in generated CloudFormation template. This PR solves the issue by adding Name tag with the value of flowLogName property.

Closes #26370

----


tam0ri:
I'll do integ test and add snapshots later.

watany-dev:
Are you saying that originally this resource did not have a property to attach a first place identifier, but L2Constructs had an interface?　The method of assigning a Name tag makes sense, but I thought that as a user of CDK constructs, I should be able to determine in advance whether it is a standard Nameid or tag support. Otherwise, users may use this property and Nametag at the same time.

tam0ri:
@watany-dev Thank you for your comment. The physical ID (identifier) for `AWS::EC2::FlowLog` resource is flow log id. It's automatically generated by EC2 [CreateFlowLogs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFlowLogs.html) API, and users can't define this id. Also, VPC flow log itself does not have a property like `name` in the API model. So, we can't specify the resource name in neither EC2 API and CloudFormation.

From the above technical background, currently flowLogName property for FlowLog construct is never used in generated CloudFormation template. This fix will utilize this property as the tag value for this resource. If Name tag is attached to flow log, VPC console shows the value as Name. Does it makes sense?

watany-dev:
1. disable `flowLogName` specification following CFn
2. treat it as a Tag like this commit, and emphasize that it should be expressed as a Nametag so that users understand it.
I personally think 2 is the way to go, but I'm waiting to see what the owners of CDK think.

lpizzinidev (approved):
Looks good to me 👍
Thanks for the contribution!

rix0rrr (approved):

rix0rrr:
Thanks!

PR #26375, opened 2023-07-14
fix(eks): cluster-resource-handler fails to verify delete operations

### LINKED ISSUE #26325: ‼️ NOTICE custom-resources: various custom resources may fail to deploy / destroy
### What is the issue?

In https://github.com/aws/aws-cdk/pull/26212, we upgraded our NodeJS runtime to Node18, which meant all our custom resources now needed to operate on AWS SDK for JavaScript v3. There were a few places that we missed:

- [eks: cluster-resource-handler onDelete fails for fargate and cluster events](https://github.com/aws/aws-cdk/issues/26325) ([Fix](https://github.com/aws/aws-cdk/pull/26283))
- s3: autoDeleteObjects fails when bucket doesn't exist on resource deletion ([Fix](https://github.com/aws/aws-cdk/pull/26433))
- [core: Custom::CrossRegionExportWriter fails](https://github.com/aws/aws-cdk/issues/26340) ([Fix](https://github.com/aws/aws-cdk/pull/26434)) ([Fix](https://github.com/aws/aws-cdk/pull/26392))
- [integ-test-alpha: assertions are not working](https://github.com/aws/aws-cdk/issues/26359) ([Fix](https://github.com/aws/aws-cdk/pull/26400))
- custom-resource: ignoreErrorCodesMatching broken on AwsCustomResource with Node.js 18/AWS SDK v3 ([Fix](https://github.com/aws/aws-cdk/pull/26430))
- [redshift-alpha: Custom resource unable to find AWS SDK v3](https://github.com/aws/aws-cdk/issues/26397) ([Fix](https://github.com/aws/aws-cdk/pull/26406))
- redshift-alpha: redshift cluster-reboot fails ([Fix](https://github.com/aws/aws-cdk/pull/26392))

### Who is affected?

Users of aws-cdk-lib version 2.87.0 

### How do I resolve this?

Upgrade to a version higher than 2.87.0

### Workaround

No workaround

------------------------------------

*Original posting*

### Describe the bug

When running the integration tests for aws-eks or aws-stepfunctions-tasks where the cluster-resource-handler is invoked will result in a failure when onDelete is called.  This is because the key `code` which is [caught](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-eks/lib/cluster-resource-handler/cluster.ts#L91) during the exception no longer exists.  Fargate's [handler](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-eks/lib/cluster-resource-handler/fargate.ts#L110) is similarly affected.

### Expected Behavior

When calling the integration tests I expected the clusters to successfully create, update, and delete themselves.

### Current Behavior

The final step of deleting the cluster fails with:

```
2023-07-10T18:59:56.441Z    fdc661e6-24f7-4d12-ac99-51342071842d    ERROR   Invoke Error    {
    "errorType": "ResourceNotFoundException",
    "errorMessage": "No cluster found for name: integrationtesteksclusterE5C0ED98-41454fc08d0746558ff42bc9a701230c.",
    "name": "ResourceNotFoundException",
    "$fault": "client",
    "$metadata": {
        "httpStatusCode": 404,
        "requestId": "75681069-c732-4896-98f7-ce3fb2f8e777",
        "attempts": 1,
        "totalRetryDelay": 0
    },
    "clusterName": "integrationtesteksclusterE5C0ED98-41454fc08d0746558ff42bc9a701230c",
    "nodegroupName": null,
    "fargateProfileName": null,
    "addonName": null,
    "stack": [
        "ResourceNotFoundException: No cluster found for name: integrationtesteksclusterE5C0ED98-41454fc08d0746558ff42bc9a701230c.",
        "    at deserializeAws_restJson1ResourceNotFoundExceptionResponse (/var/runtime/node_modules/@aws-sdk/client-eks/dist-cjs/protocols/Aws_restJson1.js:2586:23)",
        "    at deserializeAws_restJson1DescribeClusterCommandError (/var/runtime/node_modules/@aws-sdk/client-eks/dist-cjs/protocols/Aws_restJson1.js:1492:25)",
        "    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)",
        "    at async /var/runtime/node_modules/@aws-sdk/middleware-serde/dist-cjs/deserializerMiddleware.js:7:24",
        "    at async /var/runtime/node_modules/@aws-sdk/middleware-signing/dist-cjs/middleware.js:13:20",
        "    at async StandardRetryStrategy.retry (/var/runtime/node_modules/@aws-sdk/middleware-retry/dist-cjs/StandardRetryStrategy.js:51:46)",
        "    at async /var/runtime/node_modules/@aws-sdk/middleware-logger/dist-cjs/loggerMiddleware.js:6:22",
        "    at async ClusterResourceHandler.isDeleteComplete (/var/task/cluster.js:69:26)"
    ]
}
```

The cluster itself is deleting, but our evaluation of the result is failing, and thus is being treated as a failure.

### Reproduction Steps

Check out the most recent build of `aws-cdk` and run any eks test which includes a fargate profile (e.g. integ.eks-cluster-ipv6)

### Possible Solution

We can change the current evaluation to use `$e.metadata.httpResponseCode === 404` instead of a string evaluation against the message.

There's a bunch of other stuff that's broken in the eks tests, especially with the helm chart for the kubernetes-dashboard.  I've been working on a fix for the better part of 3 days and still haven't hit the bottom of the breakage.

This is affecting three tests in aws-stepfunctions-tasks.

I believe these failures are related to #26212 but I haven't had the time to identify the exact changes.  The upgrade from aws-sdk-js v2 to v3 would have ideally triggered a re-run of all of the integration tests which use the SDK, but I don't believe the resource trees can see that difference.

### CDK CLI Version

0.0.0 (build c38e784) (npx cdk)
2.42.0 (build 7d8ef0b) (local install)

### Framework Version

_No response_

### Node.js Version

v18.16.0

### OS

MacOS 13.4

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

kishiel:
## Motivation

The recent upgrade to the underlying aws-sdk-js dependency in the custom resource handler caused deletion events to fail.  This was due to the response structure differing between sdk versions, and the new response lacked the property the handler was previously targeting.

This change leverages the `httpResponseCode` to follow the same logic as before.

Some additional changes that come with this fix:

* Nginx dashboard's latest update breaks the cluster tests, and as a result I've removed them
* Nginx ingress controller was also removed as it is affected by a separate new defect either in EKS or VPC (I can't tell) which causes NLBs created by the EKS cluster to survive stack deletion.  This appears to be a transient issue, but it even appeared in the simplistic `hello-k8s` chart.

There's a broader question I have about the use of external Helm charts in these snapshot tests which do not target specific versions--this causes the snapshot tests, which are supposed to be deterministic, to become indeterministic.  Has the EKS team considered owning a few simple helm chart repos which provide us with stable versions (and thus deterministic tests)?  

I spent nearly 5 days fighting with the snapshot tests and found that a few tests (eks-cluster-test, bottlerocket, and cluster-imported test) would never use new assets and instead would deploy using the old.  If I synthesized the test using `npx cdk -a test/integ.testname.js` and deployed that it would use the new assets.  The only way I got past this was to completely delete the existing snapshot directory.  I couldn't find references to this behavior anywhere, but I wanted to call it out in case I deleted something that was not replaced with the snapshot update.

## Remaining work

The unit tests for the cluster and fargate handler were not updated.  I took a stab at them but the interface mocks that we have are no longer accurate, and I couldn't get them to mimic this change.

Fixes #26325

----


mrgrain:
This is great @kishiel thanks for looking into this. Can you check why the unit tests fail?

mrgrain:
Thanks again @kishiel Gonna close this as the team is looking into this with priority now.
Reference: https://github.com/aws/aws-cdk/pull/26283

PR #26382, opened 2023-07-16
fix(cli): hotswap doesn't update SSM parameter environment variables properly

### LINKED ISSUE #25387: Hotswap doesn't update SSM parameter environment variables properly
### Describe the bug

My lambda function is like the following.
`new NodejsFunction(this, "some-id", { environment: { VAR1: StringParameter.valueForStringParameter(this, "/au/parameter/cool-parameter") } })`

If I deploy the function before and apply hotswap, the deployment works fine. However, if I add a new environment variable to the function and use hotswap to deploy it. The value of the variable VAR1 is resolved to the path (i.e. `/au/parameter/cool-parameter`) instead of the true value in the parameter store.

### Expected Behavior

CDK should be able to resolve the right SSM parameter value.

### Current Behavior

The value of the variable VAR1 is resolved to the path (i.e. `/au/parameter/cool-parameter`) instead of the true value in the parameter store.

### Reproduction Steps

```ts
new NodejsFunction(this, "some-id", { environment: { VAR1: StringParameter.valueForStringParameter(this, "/au/parameter/cool-parameter") } })
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.69.0

### Framework Version

_No response_

### Node.js Version

16

### OS

macOS Ventura 13.2.1

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

### LINKED ISSUE #25483: CDK Hotswap: Firelens Log Router Method Incompatible With Hotswap
### Describe the bug

When CDK hotswapping a task definition that uses the function `obtainDefaultFluentBitECRImage` found in [firelens-log-router.ts](https://github.com/aws/aws-cdk/blob/1f4c4fafe22bdd80c43e80e8a968428f3aa827e1/packages/aws-cdk-lib/aws-ecs/lib/firelens-log-router.ts#L155), the use of an image URI which is grabbed from an SSM repository leads to failed deployments. Upon hotswapping, the API call made to ECS simply uses the SSM parameter name (`/aws/service/aws-for-fluent-bit:latest`) instead of the value of the SSM parameter, as happens in CDK. This makes the function incompatible with CDK hotswap.

### Expected Behavior

CDK Hotswap uses a resolvable image URI when making ECS API calls

### Current Behavior

CDK Hotswap fails with the error:

```
Task is stopping
InternalError: failed to create container model: failed to normalize image reference "/aws/service/aws-for-fluent-bit/2.23.4". Launch a new task to retry.
```

### Reproduction Steps

Using a log router which gets a firelens container from `obtainDefaultFluentBitECRImage` leads to this issue. After deploying a stack, attempt to run a hotswap deployment.

### Possible Solution

CDK Hotswap should either resolve ssm parameters in these scenarios such that the image URI is resolved in time for the image field in a task definition to be usable by ECS, or `obtainDefaultFluentBitECRImage` should change its implementation to be compatible with hotswap. Otherwise, CDK hotswap perhaps could continue to use correct fields from the original/previous deploy when the fields have no diff.

_No response_

### CDK CLI Version

2.78

### Framework Version

_No response_

### Node.js Version

16

### OS

Mac OSX

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

tmokmss:
Closes #25387
Closes #25483  

The actual value for a CfnParameter backed by a SSM parameter is returned via `ResolvedValue` (and only for SSM parameters, [doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Parameter.html#API_Parameter_Contents)). We use the field from DescirbeStacks API response to populate `stackParams`, instead of `ParameterValue`, which is just a parameter's name. 

As far as I checked it is not a breaking change. The `parameter` field is not a public API, and internally the values of SSM parameters are only used for hotswap.

----

### DIFF
--- a/packages/@aws-cdk/integ-runner/README.md
+++ b/packages/@aws-cdk/integ-runner/README.md
@@ -25,7 +25,7 @@ publishing this tool so that it can be used by the community and we would love t
 on use cases that the tool should support, or issues that prevent the tool from being used in your
 library.
 
-This tool is meant to be used with the [integ-tests](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/integ-tests) library.
+This tool is meant to be used with the [integ-tests](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/integ-tests-alpha) library.
 
 ## Usage
 
--- a/packages/@aws-cdk/integ-runner/lib/runner/integ-test-runner.ts
+++ b/packages/@aws-cdk/integ-runner/lib/runner/integ-test-runner.ts
@@ -77,8 +77,8 @@ export class IntegTestRunner extends IntegRunner {
     // test then point the user to the new `IntegTest` construct
     if (!this.hasSnapshot() && this.isLegacyTest) {
       throw new Error(`${this.testName} is a new test. Please use the IntegTest construct ` +
-       'to configure the test\n' +
-        'https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/integ-tests',
+        'to configure the test\n' +
+        'https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/integ-tests-alpha',
       );
     }
   }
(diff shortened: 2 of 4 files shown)


tmokmss:
Exemption Request: This change should be covered with cli integ tests.

mrgrain (approved):

Issue #4391, opened 2019-10-07
support routing policy setting for ARecord class


geekybaiyi:
Please support routing policy parameters for ARecord in CDK.

Not all parameters are supported in CDK Route53 module currently. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html

any plan to support those in the future?

---

This is a :rocket: Feature Request

NGL321:
Hi @geekybaiyi,

Thank you for submitting the request! From what I could tell in the [docs](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-route53.RecordSetProps.html), it looks like all the parameters supported in Cloudformation are supported in the CDK.

Could you tell me which parameters you are missing?

geekybaiyi:
how about these ?
"Failover" , "GeoLocation", "MultiValueAnswer", "Region", "SetIdentifier", "Weight"

nmussy:
Indeed, they are missing from the L2 `ResourdSet` construct. You can access them right now via the [`CfnRecordSet`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-route53.CfnRecordSet.html) resource, for example:

```ts
const zone = HostedZone.fromLookup(this, 'zone', {domainName: 'example.com'});

const record = new ARecord(this, 'A', {
    zone,
    target: RecordTarget.fromIpAddresses('1.1.1.1'),
    recordName: 'www',
});

const recordSet = (record.node.defaultChild as CfnRecordSet);
recordSet.failover = 'PRIMARY';
```

geekybaiyi:
I noticed an PR is created associated with this issue. 
any idea when it would be merged?  thx!

itaykat:
How can one reach CfnRecordSet from an ApplicationLoadBalancedFargateService object?

ryanvaloris:
The [CfnRecordSet](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-route53.CfnRecordSet.html) seems to be missing the Latency routing policy as well.  Is there a timeline for when that will be added?

jdnz:
@ryanvaloris you can enable latency routing by setting the `region` and `setIdentifier` of the [CfnRecordSet](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-route53.CfnRecordSet.html), e.g.
```
[...]

PR #26383, opened 2023-07-16
feat(route53): support geolocation routing

### LINKED ISSUE #9478: [route53] Support for geolocation routing
Currently the only way to generate a geolocation record appears to be by utilizing the `Cfn` resources. It would be great to have higher level constructs for defining these

### Other

---

jumic:
Add support for [geolocation routing](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy-geo.html) in Route53 RecordSets.

This PR adds attribute `geoLocation` to `RecordSetOptions` and new class `GeoLocation`. This enables developers to use geolocation routing.

The new feature can be used like this (more examples in README):
```ts
new route53.ARecord(this, 'ARecordGeoLocationContinent', {
  zone: myZone,
  target: route53.RecordTarget.fromIpAddresses('1.2.3.0', '5.6.7.0'),
  geoLocation: route53.GeoLocation.continent('EU'), // Europe
});
```

Closes #9478.

----


packages/aws-cdk-lib/aws-route53/lib/record-set.ts
```diff
@@ -270,6 +276,12 @@ export class RecordSet extends Resource implements IRecordSet {
       aliasTarget: props.target.aliasTarget && props.target.aliasTarget.bind(this, props.zone),
       ttl,
       comment: props.comment,
+      geoLocation: props.geoLocation ? {
+        continentCode: props.geoLocation.continentCode,
+        countryCode: props.geoLocation.countryCode,
+        subdivisionCode: props.geoLocation.subdivisionCode,
+      } : undefined,
+      setIdentifier: props.geoLocation ? this.configureSetIdentifer(props.geoLocation) : undefined,
```
jumic, line 284: Before merging this PR: I will check feature request #4391 to see if the logic for `setIdentifer` conflicts with other features that rely on the same field. I will give you feedback in the next few days. Please also feel free to think about it.

packages/aws-cdk-lib/aws-route53/lib/geo-location.ts
```diff
@@ -0,0 +1,80 @@
...
+export class GeoLocation {
+  /**
+   * Geolocation resource record based on continent code.
+   * @param continentCode Code of the continent. Valid codes are AF (Africa), AN (Antarctica), AS (Asia),
+   * EU (Europe), OC (Oceania), NA (North America), SA (South America).
+   * @returns Continent-based geolocation record
+   */
+  public static continent(continentCode: string) {
+    return new GeoLocation(continentCode, undefined, undefined);
+  }
```
kylelaker, line 13: Since this has a very small list of acceptable values (and new continents are unlikely), is there value in this being an enum instead?

```ts
export enum Continent {
  AFRICA = "AF",
  ANTARCTICA = "AN",
  ASIA = "AS",
  EUROPE = "EU",
  OCEANIA = "OC",
  NORTH_AMERICA = "NA",
  SOUTH_AMERICA = "SA"
}
```

That way users can do `route53.GeoLocation.continent(route53.Continent.ASIA)` instead of needing to know the code (and perhaps have autocomplete/suggestions from an IDE/tool)?

It might be a bit overly verbose to do that though, especially since enums aren't really the easiest option for the remaining values.

packages/aws-cdk-lib/aws-route53/lib/geo-location.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Geolocation resource record based on subdivision code (e.g. state of the United States).
+   * @param subdivisionCode Code of the subdivision (e.g. state of the United States)
+   * @param countryCode Country code (ISO 3166-1-alpha-2) of this record, by default US (United States).
+   * @returns Subdivision-based geolocation record
+   */
+  public static subdivision(subdivisionCode: string, countryCode: string = 'US') {
+    return new GeoLocation(undefined, countryCode, subdivisionCode);
+  }
```
kylelaker, line 33: CloudFormation (and the Route 53 API) _forbid_ specifying `subdivision` when the country code is value other than US. At least for now, would it better to add a check for that to catch it at synthesis time (though maybe in the constructor instead of right here)?


```ts
route53.GeoLocation.subdivision('XX', 'CA'); // -> "Subdivisions are only supported when the countryCode is 'US'"
```

> For geolocation resource record sets, the two-letter code for a state of the United States. Route 53 doesn't support any other values for `SubdivisionCode`. 
> If you specify `subdivisioncode`, you must also specify `US` for `CountryCode`.
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset-geolocation-1.html#cfn-route53-recordset-geolocation-subdivisioncode

packages/aws-cdk-lib/aws-route53/lib/geo-location.ts
```diff
@@ -0,0 +1,80 @@
...
+   */
+  public static continent(continentCode: string) {
+    return new GeoLocation(continentCode, undefined, undefined);
+  }
+
+  /**
+   * Geolocation resource record based on country code.
+   * @param countryCode Two-letter, uppercase country code for the country.
+   * See ISO 3166-1-alpha-2 code on the *International Organization for Standardization* website
+   * @returns Country-based geolocation record
```
kylelaker, line 19: Let's link to the specific docs (both from AWS and the ones AWS links to)?

```suggestion
   * @see https://docs.aws.amazon.com/Route53/latest/APIReference/API_GeoLocation.html#Route53-Type-GeoLocation-CountryCode
   * @see https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
   * @returns Country-based geolocation record
```

packages/aws-cdk-lib/aws-route53/lib/geo-location.ts
```diff
@@ -0,0 +1,80 @@
...
+   */
+  public static country(countryCode: string) {
+    return new GeoLocation(undefined, countryCode, undefined);
+  }
+
+  /**
+   * Geolocation resource record based on subdivision code (e.g. state of the United States).
+   * @param subdivisionCode Code of the subdivision (e.g. state of the United States)
+   * @param countryCode Country code (ISO 3166-1-alpha-2) of this record, by default US (United States).
```
kylelaker, line 28:
```suggestion
   * @param countryCode Country code (ISO 3166-1-alpha-2) of this record, by default US (United States).
   * @see https://pe.usps.com/text/pub28/28apb.htm
   * @see https://docs.aws.amazon.com/Route53/latest/APIReference/API_GeoLocation.html#Route53-Type-GeoLocation-SubdivisionCode
```

kylelaker (commented):
This is really exciting -- it looks like this is the first complex routing supported in the L2 RecordSet construct!

I'm personally really interested in using these APIs and I have a question on the API for continents, some small docs suggestions, and a question about maybe trying to catch a failure case early (during synthesis).

packages/aws-cdk-lib/aws-route53/lib/geo-location.ts
jumic: Route53 also supports `subdivisioncode` for `CountryCode` US (Ukraine). However, I can't find it in the documentation, too. It can be maintained using AWS console and CloudFormation. AWS console:
image

I added an integration test for a subdivision in Ukraine which can be deployed successfully.
```ts
new ARecord(stack, 'GeoLocationSubDividionUA', {
  zone: privateZone,
  target: RecordTarget.fromIpAddresses('1.2.3.4', '5.6.7.4'),
  geoLocation: GeoLocation.subdivision('30', 'UA'), // Ukraine, Kyiv
});
```

Which field order do you prefer?
Option 1: SubDivision, Country (country can be optional for US).
```ts
GeoLocation.subdivision('WA', 'US')
GeoLocation.subdivision('30', 'UA')
```

Option 2: Country, SubDivision (country 
```ts
GeoLocation.subdivision('US', 'WA')
GeoLocation.subdivision('UA', '30')
```

I think I would prefer option 2 because the generic information (country) has to be specified first.

packages/aws-cdk-lib/aws-route53/lib/geo-location.ts
jumic: Yes, this makes sense. The enum is easier to use. It's not consistent anymore because country uses a string and continent uses an enum. However, I like the autocompletion.

packages/aws-cdk-lib/aws-route53/lib/geo-location.ts
jumic: I added validation for country code in subdivision, only UA (Ukraine) and US (United States) is allowed now.

packages/aws-cdk-lib/aws-route53/lib/geo-location.ts
```diff
@@ -0,0 +1,114 @@
...
+   * Default (wildcard) routing record if no specific geolocation record is found.
+   * @returns Wildcard routing record
+   */
+  public static default() {
+    return new GeoLocation(undefined, '*', undefined);
+  }
+
+  private static COUNTRY_REGEX = /(^[A-Z]{2}|^\*{1})$/;
+  private static COUNTRY_FOR_SUBDIVISION_REGEX = /\b(?:UA|US)\b/;
+  private static SUBDEVISION_REGEX = /^[A-Z0-9]{1,3}$/;
```
colifran, line 47: Nit: Can we fix the spelling here to `SUBDIVISION_REGEX` instead of `SUBDEVISION_REGEX`

packages/aws-cdk-lib/aws-route53/lib/geo-location.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (!GeoLocation.COUNTRY_REGEX.test(country)) {
+      // eslint-disable-next-line max-len
+      throw new Error(`Invalid country format for country: ${country}, country should be two-letter and uppercase country ISO 3166-1-alpha-2 code`);
+    }
+  }
+
+  private static validateCountryForSubdivision(country: string) {
+    if (!GeoLocation.COUNTRY_FOR_SUBDIVISION_REGEX.test(country)) {
+      // eslint-disable-next-line max-len
+      throw new Error(`Invalid country for subdivions geolocation: ${country}, only UA (Ukraine) and US (United states) are supported`);
```
colifran, line 59: Nit: Can we correct the spelling of `subdivision` in the error message?

packages/aws-cdk-lib/aws-route53/lib/record-set.ts
colifran: Was this looked into or are we still awaiting feedback?

colifran (changes requested):
@jumic this looks great! I left a few very minor comments.

packages/aws-cdk-lib/aws-route53/lib/geo-location.ts
jumic: @colifran Can you also give a feedback to this questions? Should I implement option 2 (as proposed)? Or do you prefer option 1? Thanks.

colifran: @jumic My preference would be option 1. Since the static method is `subdivision`, I think it makes more sense for the first field to be the `subdivision` to use followed by the country that the `subdivision` is for.

jumic: OK, then we will use option 1. In this case, no further changes are necessary.

colifran (approved):
@jumic Looks great! Thanks for your work and contribution.

PR #26396, opened 2023-07-18
feat: support max-buffer-size for AWSLogs driver

PettitWesley:
When mode is set to `non-blocking`, logs are stored in a buffer. The setting `max-buffer-size` controls the size of this buffer. Being able to set this buffer is very important, the larger the buffer, the less likely there is to be log loss.

Recently I performed benchmarking of `non-blocking` mode, and found that the default buffer size is not sufficient to prevent log loss: https://github.com/moby/moby/issues/45999

We're planning to run an education campaign to ensure ECS customers understand the risk with the default `blocking` mode, and the risk with `non-blocking` mode with the default buffer size. Therefore, we need folks to be able to control this setting in their CDK.

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/README.md
+++ b/packages/@aws-cdk-testing/framework-integ/README.md
@@ -1,0 +1,40 @@
+# AWS CDK Framework Integration Test Suite
+
+Run the full test suite with `yarn integ`.
+
+See `integ-runner` package or `yarn integ --help` for detailed instructions.
+
+## Common Errors
+
+### Error: For this test you must provide your own HostedZoneId/HostedZoneName/DomainName
+
+Some test cases require a publicly available domain name attached to a Amazon Route 53 Hosted Zone to work.
+These test cases need to add DNS records that are then retrieved via the public internet infrastructure. This can be a subdomain to an existing domain, but it must have a Hosted Zone configured and nameservers delegated to it.
+
+If you haven't got one ready, see the [Route 53 guide](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/CreatingHostedZone.html) to manually create a hosted zone.
+
+AWS CDK core team members, please check our team internal docs for guidance on how to configure domains for testing.
+
+#### How to correctly run these tests
+
+**A) Each test needs exclusive access to the Hosted Zone.**\
+You can only run one test at a time.
+
+**B) Must be run with `--disable-update-workflow`**\
+The checked-in snapshot uses dummy values that will not deploy.
+
+- Go to your Hosted Zone and write down the values for `HostedZoneId`, `HostedZoneName` and `DomainName`.
+   All values must related to the **same** Hosted Zone.
+- In your terminal run the following commands:
+
+```console
+export HOSTED_ZONE_ID=your_hosted_zone_id
+export HOSTED_ZONE_NAME=your_hosted_zone_name
+export DOMAIN_NAME=your_domain_name
+```
+
+- Finally, in the same terminal run your specific test case with the **update workflow disabled**. For example:
+
+```console
+yarn integ --disable-update-workflow aws-certificatemanager/test/integ.certificate-name
+```
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-certificatemanager/test/integ.certificate-name.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-certificatemanager/test/integ.certificate-name.ts
@@ -9,11 +9,11 @@ import { Certificate, CertificateValidation } from 'aws-cdk-lib/aws-certificatem
  *
 */
 const hostedZoneId = process.env.CDK_INTEG_HOSTED_ZONE_ID ?? process.env.HOSTED_ZONE_ID;
-if (!hostedZoneId) throw new Error('For this test you must provide your own HostedZoneId as an env var "HOSTED_ZONE_ID"');
+if (!hostedZoneId) throw new Error('For this test you must provide your own HostedZoneId as an env var "HOSTED_ZONE_ID". See framework-integ/README.md for details.');
 const hostedZoneName = process.env.CDK_INTEG_HOSTED_ZONE_NAME ?? process.env.HOSTED_ZONE_NAME;
-if (!hostedZoneName) throw new Error('For this test you must provide your own HostedZoneName as an env var "HOSTED_ZONE_NAME"');
+if (!hostedZoneName) throw new Error('For this test you must provide your own HostedZoneName as an env var "HOSTED_ZONE_NAME". See framework-integ/README.md for details.');
 const domainName = process.env.CDK_INTEG_DOMAIN_NAME ?? process.env.DOMAIN_NAME;
-if (!domainName) throw new Error('For this test you must provide your own Domain Name as an env var "DOMAIN_NAME"');
+if (!domainName) throw new Error('For this test you must provide your own DomainName as an env var "DOMAIN_NAME". See framework-integ/README.md for details.');
 
 const app = new App();
 const stack = new Stack(app, 'integ-certificate-name');
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-certificatemanager/test/integ.dns-validated-certificate.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-certificatemanager/test/integ.dns-validated-certificate.ts
@@ -20,11 +20,11 @@ import { DnsValidatedCertificate, CertificateValidation } from 'aws-cdk-lib/aws-
  */
 
 const hostedZoneId = process.env.CDK_INTEG_HOSTED_ZONE_ID ?? process.env.HOSTED_ZONE_ID;
-if (!hostedZoneId) throw new Error('For this test you must provide your own HostedZoneId as an env var "HOSTED_ZONE_ID"');
+if (!hostedZoneId) throw new Error('For this test you must provide your own HostedZoneId as an env var "HOSTED_ZONE_ID". See framework-integ/README.md for details.');
 const hostedZoneName = process.env.CDK_INTEG_HOSTED_ZONE_NAME ?? process.env.HOSTED_ZONE_NAME;
-if (!hostedZoneName) throw new Error('For this test you must provide your own HostedZoneName as an env var "HOSTED_ZONE_NAME"');
+if (!hostedZoneName) throw new Error('For this test you must provide your own HostedZoneName as an env var "HOSTED_ZONE_NAME". See framework-integ/README.md for details.');
 const domainName = process.env.CDK_INTEG_DOMAIN_NAME ?? process.env.DOMAIN_NAME;
-if (!domainName) throw new Error('For this test you must provide your own Domain Name as an env var "DOMAIN_NAME"');
+if (!domainName) throw new Error('For this test you must provide your own DomainName as an env var "DOMAIN_NAME". See framework-integ/README.md for details.');
 
 const app = new App();
 const stack = new Stack(app, 'integ-dns-validated-certificate');
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-cloudfront/test/integ.cloudfront-cross-region-cert.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-cloudfront/test/integ.cloudfront-cross-region-cert.ts
@@ -7,11 +7,11 @@ import * as cloudfront from 'aws-cdk-lib/aws-cloudfront';
 
 const account = process.env.CDK_INTEG_ACCOUNT ?? process.env.CDK_DEFAULT_ACCOUNT;
 const hostedZoneId = process.env.CDK_INTEG_HOSTED_ZONE_ID ?? process.env.HOSTED_ZONE_ID;
-if (!hostedZoneId) throw new Error('For this test you must provide your own HostedZoneId as an env var "HOSTED_ZONE_ID"');
+if (!hostedZoneId) throw new Error('For this test you must provide your own HostedZoneId as an env var "HOSTED_ZONE_ID". See framework-integ/README.md for details.');
 const hostedZoneName = process.env.CDK_INTEG_HOSTED_ZONE_NAME ?? process.env.HOSTED_ZONE_NAME;
-if (!hostedZoneName) throw new Error('For this test you must provide your own HostedZoneName as an env var "HOSTED_ZONE_NAME"');
+if (!hostedZoneName) throw new Error('For this test you must provide your own HostedZoneName as an env var "HOSTED_ZONE_NAME". See framework-integ/README.md for details.');
 const domainName = process.env.CDK_INTEG_DOMAIN_NAME ?? process.env.DOMAIN_NAME;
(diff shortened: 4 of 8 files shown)


packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -82,6 +82,16 @@ export interface AwsLogDriverProps {
    * @default - AwsLogDriverMode.BLOCKING
    */
   readonly mode?: AwsLogDriverMode;
+
+  /**
+   * When AwsLogDriverMode.NON_BLOCKING is configured, this parameter
+   * controls the size of the non-blocking buffer used to temporarily
+   * store messages. This parameter is not valid with 
+   * AwsLogDriverMode.BLOCKING.
```
PettitWesley, line 90: Since this is not valid with blocking mode, should I add some sort of check? How does that work? Can someone help me with how to do that- does it go in the `bind` method and what do I make it do/throw/return on failure?

PettitWesley: Something like? =>

```
if (this.props.maxBufferSize != "" && this.props.mode != AwsLogDriverMode. NON_BLOCKING) {
    throw new Error('max-buffer-size requires non-blocking mode')
}
```

iamhopaul123: Yes i was about to ask for the same thing. It will be great if we can perform a check in L2.

Maybe
```ts
if (this.props.maxBufferSize != "" && this.props.mode === AwsLogDriverMode.BLOCKING) { throw new Error('max-buffer-size is not valid with blocking mode') }
```
which is more related to "This parameter is not valid with AwsLogDriverMode.BLOCKING.", unless you want to change the comment msg to "this parameter is only valid with non-blocking mode".

PettitWesley:
Integ test results showing that the buffer size was set in the task def:
Screen Shot 2023-07-19 at 4 27 41 PM
a-479f-8815-aeab1e1e1f80">

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -116,6 +126,10 @@ export class AwsLogDriver extends LogDriver {
       retention: this.props.logRetention || Infinity,
     });
 
+    if (this.props.maxBufferSize != "" && this.props.mode === AwsLogDriverMode.BLOCKING) { 
+      throw new Error('max-buffer-size is not valid with blocking mode')
```
iamhopaul123, line 130:
```suggestion
      throw new Error('maxBufferSize is not valid with blocking mode') 
```

iamhopaul123 (approved):
LGTM overall! Can we also add a unit test for the new error scenario?

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
PettitWesley: Sure

PettitWesley:
Added unit test

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -106,6 +116,10 @@ export class AwsLogDriver extends LogDriver {
     if (props.logGroup && props.logRetention) {
       throw new Error('Cannot specify both `logGroup` and `logRetentionDays`.');
     }
+
+    if (this.props.maxBufferSize != '' && this.props.mode === AwsLogDriverMode.BLOCKING) {
```
colifran, line 120: Since we're still in the constructor, we can just use `props` here, no need to use `this`. Additionally, let's use strict equality, i.e., `!==`, when checking for an empty string. So:
`if (props.maxBufferSize !== '' && props.mode === AwsLogDriverMode.BLOCKING)`

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -82,6 +82,16 @@ export interface AwsLogDriverProps {
...
+
+  /**
+   * When AwsLogDriverMode.NON_BLOCKING is configured, this parameter
+   * controls the size of the non-blocking buffer used to temporarily
+   * store messages. This parameter is not valid with
+   * AwsLogDriverMode.BLOCKING.
+   *
+   * @default - No default
+   */
+  readonly maxBufferSize?: string;
```
colifran, line 94: I think a better way of handling this is make `maxBufferSize` to be a `number` type. This will provide a better user experience as it isn't extremely clear upfront that the user should provide something of the form  1m, 5m, 10m, etc. We can then do the work to append 'm' in the `bind` method.

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -82,6 +82,16 @@ export interface AwsLogDriverProps {
...
    */
   readonly mode?: AwsLogDriverMode;
+
+  /**
+   * When AwsLogDriverMode.NON_BLOCKING is configured, this parameter
+   * controls the size of the non-blocking buffer used to temporarily
+   * store messages. This parameter is not valid with
+   * AwsLogDriverMode.BLOCKING.
+   *
+   * @default - No default
```
colifran, line 92: I think our default here needs to be `1m` since this is the documented default for `maxBufferSize`. See linked documentation - https://docs.aws.amazon.com/AmazonECS/latest/userguide/using_awslogs.html

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -106,6 +116,10 @@ export class AwsLogDriver extends LogDriver {
     if (props.logGroup && props.logRetention) {
       throw new Error('Cannot specify both `logGroup` and `logRetentionDays`.');
     }
+
+    if (props.maxBufferSize !== '' && props.mode === AwsLogDriverMode.BLOCKING) {
+      throw new Error('maxBufferSize is not valid with blocking mode');
```
colifran, line 121: Nit: Could we make this consistent with the other error message? Something like:

```ts
Cannot specify `maxBufferSize` when the driver mode is blocking
```

colifran (dismissed):
@PettitWesley This looks pretty good so far. I have a few more changes I think we should make to improve the code.

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -106,6 +116,10 @@ export class AwsLogDriver extends LogDriver {
     if (props.logGroup && props.logRetention) {
       throw new Error('Cannot specify both `logGroup` and `logRetentionDays`.');
     }
+
+    if (props.maxBufferSize !== '' && props.mode === AwsLogDriverMode.BLOCKING) {
```
colifran, line 120: I think we need to also check if `props.mode` is undefined here. Since `mode` is optional it might not be provided which will cause the default mode of blocking. So maybe it would be cleaner for us to validate based on `maxBufferSize` being provided and `mode` not being `NON_BLOCKING`. Note: I'm basing this off of us making `maxBufferSize` a `number` instead of a `string`

```ts
if (props.maxBufferSize !== undefined && props.mode !== AwsLogDriverMode.NON_BLOCKING) {
  // code
}
```

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
PettitWesley: @colifran The problem here is that while "number + m" is the most likely user config, it's not the only thing that Docker supports. The string arg to max-buffer-size can have multiple units:
> format: <number>[<unit>], where unit = b, k, m or g)

So this is valid and works:

```
docker run -it --log-opt mode=non-blocking --log-opt max-buffer-size=4k alpine
```

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
PettitWesley: Updated

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
PettitWesley: Updated

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
PettitWesley: But wait, I think we want our default to be undefined since the blocking mode is default mode which is incompatible with this option.  Does the `@default` key word here only impact the docs or the CDK code itself? When a user does not specify this option, we do not want it to be set.

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
colifran: @PettitWesley Ahh I see. In that case, it seems like this property might be a good candidate to utilize an enum-like class. I think we want to help direct the user input and tighten up what a user can provide as an option. As an example,

```ts
export enum BufferSizeType {
  MB = 'MB',
  KB = 'KB',
  GB = 'GB',
}

export class BufferSize {
  public static megabyte(size: number) {
    return new BufferSize(BufferSizeType.MB, size);
  }

  public static kilobyte(size: number) {
    return new BufferSize(BufferSizeType.KB, size);
  }

  public static gigabyte(size: number) {
    return new BufferSize(BufferSizeType.GB, size);
  }
  
  //  more buffer size options

  private constructor(public readonly type: string, public readonly size: number) {}
}
```

What do you think?

PettitWesley: What does the user input? A string or do they create one of these structs?

PettitWesley: I think I'd rather not create a new struct just for this field, how about instead I add better validations in the constructor that require that this is a number + some unit character? That would serve the same use case.

PettitWesley: Docker code:
- https://pkg.go.dev/github.com/docker/go-units#RAMInBytes
- https://github.com/docker/go-units/blob/master/size.go#L36

The code always processes either a string or an int, and uses functions there to convert back and forth.

PettitWesley: Should this be common validation code that code in a centralized location and not just this log driver?

PettitWesley: Something like: `IsValidDockerMemoryString()` which I call in the constructor to validate that the user inputted string is valid.

PettitWesley: As discussed on slack, I will use the existing class:
 https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.Size.html. I think we could just use this as the type for maxBufferSize . I think we could convert to bytes in the bind method. It would looks something like this:
let bufferSize: string | undefined = undefined;
if (props.mode === AwsLogsDriver.NON_BLOCKING) {
  bufferSize = (props.maxBufferSize ?? cdk.Size.mebibytes(1)).toBytes({ rounding: cdk.SizeRoundingBehavior.FLOOR }) + 'b';
}

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -116,6 +131,11 @@ export class AwsLogDriver extends LogDriver {
       retention: this.props.logRetention || Infinity,
     });
 
+    var maxBufferSize = '';
```
colifran, line 134: This needs to be `undefined` in the event that `props.maxBufferSize` isn't provided. Otherwise, `removeEmpty` on the log driver options won't remove `max-buffer-size`. The build failure is coming from other integ tests being impacted since with `maxBufferSize` being an empty string it'll be set on everything using this even when `props.maxBufferSize`. Isn't provided. Change this to:

`let maxBufferSize: string | undefined = undefined;`

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -106,6 +117,10 @@ export class AwsLogDriver extends LogDriver {
     if (props.logGroup && props.logRetention) {
       throw new Error('Cannot specify both `logGroup` and `logRetentionDays`.');
     }
+
+    if (props.maxBufferSize !== undefined && props.mode !== AwsLogDriverMode.NON_BLOCKING) {
```
colifran, line 121: No that we're using `Size` as the type we can just use:

`if (props.maxBufferSize && props.mode !== AwsLogDriverMode.NON_BLOCKING`

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -116,6 +131,11 @@ export class AwsLogDriver extends LogDriver {
       retention: this.props.logRetention || Infinity,
     });
 
+    var maxBufferSize = undefined;
```
colifran, line 134: We really shouldn't use var here. Honestly we can probably just simplify this whole statement with a ternary operator:

```ts
const maxBufferSize = this.props.maxBufferSize
  ? `${this.props.maxBufferSize.toBytes({ rounding: SizeRoundingBehavior.FLOOR })}b`
  : undefined;
```

packages/aws-cdk-lib/aws-ecs/test/aws-log-driver.test.ts
```diff
@@ -146,6 +148,20 @@ describe('aws log driver', () => {
 
   });
 
+  test('throws error when specifying maxBufferSize and blocking mode', () => {
```
colifran, line 151: Can you add one more unit test to validate that this will throw when maxBufferSize is used and no blocking mode is specified that way we're validating that we can't specify maxBufferSize with the default driver mode

colifran (changes requested):
A few more minor comments and then this will be good to go.

colifran (approved):
@PettitWesley Looks great. Thanks for your work on this!

colifran (approved):

colifran (dismissed):

colifran (dismissed):

colifran (approved):

PR #26403, opened 2023-07-18
feat(opensearchservice): L2 properties for offPeakWindowOptions and softwareUpdateOptions

### LINKED ISSUE #26388: opensearch: High Level Constructs For OpenSearchOffPeakWindowOptions and Software Update Options Feature
### Describe the feature

The OpenSearch Team has recently launched OffPeak and Software Update OptionsFeature for OpenSearch Domain. [OffPeakWindowOptionsLink](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/off-peak.html). 
[SoftwareUpdateOptionsLink](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/service-software.html)
According to the CDK docs currently there is no high level CDK construct for these features.
[CDK docs link for OffPeakWindowOptionsProperty](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_opensearchservice.CfnDomain.OffPeakWindowOptionsProperty.html)
[CDK Docs link for  OffPeakWindow](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_opensearchservice.CfnDomain.OffPeakWindowProperty.html)
[CDK Docs link for SoftwareUpdateOptions](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_opensearchservice.CfnDomain.SoftwareUpdateOptionsProperty.html)
[CFN docs link for OffPeakWindow](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-opensearchservice-domain-offpeakwindow.html)
[CFN docs link for OffPeakWindowOptions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-opensearchservice-domain-offpeakwindowoptions.html)
[CFN docs link for  SoftwareUpdateOptions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-opensearchservice-domain-softwareupdateoptions.html)

### Use Case

The general recommendation is to use high level constructs and due to lack of High level construct for this feature, the CDK template needs to be migrated to use CFN constructs if there is a need to Create OffPeak window for OpenSearch Domain . This serves as a hinderance to the adoption of the  OffPeak Window and Software Update Options feature for AWS OpenSearch Domains.

### Proposed Solution

Add the OffPeakWindowOptions to the OpenSearch Domain Props(https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_opensearchservice.Domain.html) .
Configuring OffPeakWindowOptions on OpenSearchDomain created using CDK high level constructs should look like
const domain = new Domain(this, 'Domain', {
version: EngineVersion.OPENSEARCH_1_0,

offPeakWindowOptions: {
   enabled: true (possible values: [true,false])
   offPeakWindow: {
    windowStartTime: {
   hours: 4 [possible values from 0 to 60]
  minutes: 4 [possible values from 0 to 60]
 }
 }
},
softwareUpdateOptions : {
 autoSoftwareUpdateEnabled: true (possible values: [true,false])
}
});

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.84.0

macOS Ventura 13.4

lpizzinidev:
The [`OffPeakWindowOptions`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_opensearchservice.CfnDomain.OffPeakWindowOptionsProperty.html) and [`SoftwareUpdateOptions`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_opensearchservice.CfnDomain.SoftwareUpdateOptionsProperty.html) are supported by OpenSearch Domain, but not by the CDK high-level construct.

This change adds the corresponding properties to the `Domain` construct:

```ts
const domain = new Domain(this, 'Domain', {
  version: EngineVersion.OPENSEARCH_1_3,
  offPeakWindowEnabled: true, // can be omitted if offPeakWindowStart is set
  offPeakWindowStart: {
    hours: 20,
    minutes: 0,
  },
  enableAutoSoftwareUpdate: true,
});
```

Closes #26388.

----


packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
```diff
@@ -511,6 +557,17 @@ export interface DomainProps {
    * @default - no custom domain endpoint will be configured
    */
   readonly customEndpoint?: CustomEndpointOptions;
+
+  /**
+   * Options for a domain's off-peak window, during which OpenSearch Service can perform mandatory
+   * configuration changes on the domain.
+   */
+  readonly offPeakWindowOptions?: OffPeakWindowOptions;
```
kaizencc, line 565: specifically the default for this prop may be confusing. looks like the default should be that it is configured by default and cannot be turned off, but still, what will the default window be?

> Off-peak windows were introduced on February 16, 2023. All domains created before this date have the off-peak window disabled by default. You must manually enable and configure the off-peak window for these domains. All domains created after this date will have the off-peak window enabled by default. You can't disable the off-peak window for a domain after it's enabled.

https://docs.aws.amazon.com/opensearch-service/latest/developerguide/off-peak.html

kaizencc (changes requested):
Looks pretty good @lpizzinidev! Just a couple comments about defaults

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
lpizzinidev: I defaulted it to `00:00 UTC`

> If you don't specify a custom window start time, it defaults to 00:00 UTC.

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
```diff
@@ -328,6 +328,65 @@ export interface CustomEndpointOptions {
...
+   *
+   * @default - default window start time is 00:00 UTC
+   */
+  readonly offPeakWindow?: OffPeakWindow;
+}
+
+/**
+ * Options for configuring service software updates for a domain.
+ */
+export interface SoftwareUpdateOptions {
```
kaizencc, line 381: i dont think we need an interface for this, lets default to a flatter API. Think we should just expose a boolean `enableAutoSoftwareUpdate`

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
```diff
@@ -328,6 +328,65 @@ export interface CustomEndpointOptions {
...
+   * @default - true
+   */
+  readonly enabled?: boolean;
+
+  /**
+   * Off-peak window settings for the domain.
+   *
+   * @default - default window start time is 00:00 UTC
+   */
+  readonly offPeakWindow?: OffPeakWindow;
```
kaizencc, line 375: I don't think we need this to be an interface either. We should be able to get away with just `offPeakWindow?: OffPeakWindow`. If its specified, users probably want it to be enabled as well. In general I like to default to a single prop rather than specifying interfaces.

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
```diff
@@ -1558,6 +1632,23 @@ export class Domain extends DomainBase implements IDomain, ec2.IConnectable {
       }
     }
 
+    let offPeakWindowOptions: OffPeakWindowOptions = {
+      enabled: props.offPeakWindowOptions?.enabled ?? true,
```
kaizencc, line 1636: > All domains created before this date have the off-peak window disabled by default. 

So we can't just default to true here, as that will represent a behavior change to existing domains. Rather, without setting `offPeakWindow` (i.e. `offPeakWindow: undefined`) what we have to do is document the service-level behavior (enabled post 2/16/23, disabled prior). Then, we just pass `undefined` into the L1 and let the service handle the default.

kaizencc (changes requested):
Thanks @lpizzinidev, a couple more comments

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
```diff
@@ -511,6 +538,29 @@ export interface DomainProps {
...
+   * configuration changes on the domain.
+   *
+   * Off-peak windows were introduced on February 16, 2023.
+   * All domains created before this date have the off-peak window disabled by default.
+   * You must manually enable and configure the off-peak window for these domains.
+   * All domains created after this date will have the off-peak window enabled by default.
+   * You can't disable the off-peak window for a domain after it's enabled.
+   *
+   * @see https://docs.aws.amazon.com/it_it/AWSCloudFormation/latest/UserGuide/aws-properties-opensearchservice-domain-offpeakwindow.html
+   * @default - no off-peak window will be configured
```
kaizencc, line 553:
```suggestion
   * @default - disabled for domains created before February 16, 2023. enabled for domains created after.
```

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
```diff
@@ -1632,6 +1684,18 @@ export class Domain extends DomainBase implements IDomain, ec2.IConnectable {
...
       advancedOptions: props.advancedOptions,
+      offPeakWindowOptions: props.offPeakWindow ? {
+        enabled: true,
+        offPeakWindow: {
+          windowStartTime: props.offPeakWindow.windowStartTime ?? {
+            hours: 22,
+            minutes: 0,
+          },
+        },
+      } : undefined,
```
kaizencc, line 1695: so the reason why I'm not in love with this API yet is because to specify offPeakWindow with the given defaults we have to do:

```ts
offPeakWindow: {},
```

That's not very intuitive. Which now makes me understand why `enabled` was a prop in the first place :). Still, though, this potential experience to get the default window isn't great either:

```ts
offPeakWindow: {
  enabled: true,
},
```

It feels weird to supply an object with just a boolean prop. So I suggest we do the following:

```ts
offPeakWindowEnabled: true,
// if people want to customize the window
// offPeakWindowStart: {
//  hours: 22,
//  minutes: 0,
// },
```

I think having two separate properties on the base construct is the way to go and what we've generally done in the past for similar situations.

Of course, the default for `offPeakWindowEnabled` is `true if offPeakWindowStart is set, false otherwise`.

kaizencc (dismissed):
Apologies @lpizzinidev I feel like I've been giving you half-baked reviews on this PR. But this last little redesign should be enough I think. Thanks for the quick turnarounds.

kaizencc (dismissed):
Thanks @lpizzinidev! Just added one last thing that allows `offPeakWindowEnabled` to be omitted if you specify `offPeakWindowStart`. Thanks for the contribution!

kaizencc (approved):

PR #26404, opened 2023-07-18
fix(cli): cannot hotswap ECS task definitions containing certain intrinsics

### LINKED ISSUE #25563: cli: allow hotswap of ECS task definitions that contain tokens

tmokmss:
## Reason for this change
Currently an ECS task definition cannot be hotswapped in many cases, for example when it contains references to certain AWS resources as environment variables. This PR removes the limitation and make hotswaps possible in much more various situations.

Specifically, if there is any token that is not part of the following, we cannot hotswap a task definition:

1. Ref functions, which will be resolved as physical resource ids
2. GetAtt functions for some resources and attributes (manually supported one by one, [code](https://github.com/aws/aws-cdk/blob/5ccc56975c323ea19fd0917def51184e13f440d9/packages/aws-cdk/lib/api/evaluate-cloudformation-template.ts#L352))
3. several simple CFn functions (join, split, select, sub)
4. parameters like AWS::AccountId, Region, Partition, or UrlSuffix

Although it is not supported much, using `GetAtt` function in a task definition is very common (imagine you reference other resource's property as an environment variable). This PR allows to hotswap a task definition even if it contains these tokens.

## Solution
To hotswap a task definition, we need to construct a task definition to call `registerTaskDefinition` API. For this, we have to [evaluate](https://github.com/aws/aws-cdk/blob/5ccc56975c323ea19fd0917def51184e13f440d9/packages/aws-cdk/lib/api/evaluate-cloudformation-template.ts#L134) CloudFormation template locally to resolve all the intrinsics in the template. However, some intrinsics such as `Fn::GetAtt` is not fully supported by CDK CLI because we have to manually support them for each AWS resource type. 

The reason why some task definitions are unhotswappable is that there are such intrinsics in the template and the CDK fails to evaluate it locally. So the basic idea to overcome this limitation in this PR is that we don't try to evaluate it locally, but we fetch the latest task definition already deployed and get the required values from it.

Here's how we can implement the idea.

### How we determine if changes to a task definition can be hotswapped
In the hotswap process, we have to decide whether the change can be hotswapped or not. Now we can hotswap the task definition if

1. there are only changes in `ContainerDefinitions` field, and all the fields in the task definition can be evaluated locally. (original behavior) OR,
2. there are only changes in `ContainerDefinitions` field, and all the **updated** field can be evaluated locally (added in this PR).

The first condition can actually be included in the second condition, but for now I keep it as-is to avoid the possibility of breaking the existing behavior.

If the second condition is true, we fetch the latest task definition from AWS account, override the updated fields, and register a new task definition to update a service. By this way, we don't have to evaluate tokens in unchanged fields locally, allowing to use hotswap in much more situations.

### How we compare the old and new task definition
Here is an example task definition:

```json
{
    "ContainerDefinitions": [
        {
            "Environment": [
                {
                    "Name": "VPC_ID",
                    "Value": {
                        "Fn::GetAtt": [
                            "Vpc8378EB38",
                            "CidrBlock"
                        ]
                    }
                }
            ],
            "Essential": true,
            "Image": "nginx:stable",
            "Name": "EcsApp"
        }
    ],
    "Cpu": "256",
    "Family": "myteststackTask289798EC",
    "Memory": "512",
    "NetworkMode": "awsvpc",
    "RequiresCompatibilities": [
        "FARGATE"
    ],
    "TaskRoleArn": {
        "Fn::GetAtt": [
            "TaskTaskRoleE98524A1",
            "Arn"
        ]
    }
}
```

We compare the old and new task definition in the following steps:

1. Check if there are only changes in `ContainerDefinitions` field. If not, we cannot hotswap.
2. Try `evaluateCfnExpression` on the containerDefinitons. If it can be evaluated, proceed to hotswap. If not, proceed to step 3.
3. Check if the length of `ContainerDefinitions` is the same. If not, we cannot hotswap.
4. For each container definition, deep-compare each key (e.g. `Environment`, `Image`, `Name`, etc)
5. For each key, if there is any diff in the corresponding value, try `evaluateCfnExpression` on the value. If the evaluation fails, we cannot hotswap.
6. After checking all the keys and there is no field that cannot be hotswapped, proceed to hotswap.

Imagine if there is a change only in `Image` field (container image tag)  but `Environment` field contains unsupported intrinsics (e.g. `"Fn::GetAtt": ["Vpc8378EB38", "CidrBlock"]`). In the previous CDK CLI we cannot hotswap it due to an evaluation error. We can now hotswap it because we don't have to evaluate the `Environment` field when it has no diffs.

Closes #25563

----


kaizencc:
Running the cli integ tests right now, just so I can exempt you from the linter

tmokmss:
@kaizencc Thank you very much! 

By the way, how about removing the cli integ test check from the PR linter? I believe the PR linter is for a contributor to validate if the PR is ready for review, but as for the cli integ check, there is nothing they can do to make the linter pass, which is a bit frustrating. I think the need to run the cli integ test can be expressed with another label.

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -80,6 +85,23 @@ export async function isHotswappableEcsServiceChange(
             },
           },
         });
+
+        if (taskDefinitionResource.hotswappableWithMerge) {
+          // get the latest task definition of the family
+          const describeTaskDefinitionResponse = await sdk
+            .ecs()
+            .describeTaskDefinition({
```
comcalvi, line 93: No other hotswappable resource does a `describe` on the service. What is missing from `EvaluateCfnTemplate` that forces us to do this? We already have these prior definitions from the previous form of the CFN template.

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -167,13 +189,47 @@ export async function isHotswappableEcsServiceChange(
   return ret;
 }
 
+function mergeTaskDefinitions(patch: {containerDefinitions: {[key:string]: any}[]}, target: AWS.ECS.DescribeTaskDefinitionResponse) {
+  const src = patch.containerDefinitions;
+  const dst = target.taskDefinition?.containerDefinitions;
+  if (dst === undefined) return;
```
comcalvi, line 195:
```suggestion
  if (dst === undefined) {
    return;
  }
```

comcalvi (commented):
Thanks for opening this PR and writing a detailed explanation of the motivation! I have a few questions on this:

> However, some intrinsics such as Fn::GetAtt is not fully supported by CDK CLI because we have to manually support them for each AWS resource type.

I'm curious what you mean by this. `EvaluteCfnTemplate` doesn't explicitly check any of these resources, it instead looks up every arg to `GetAtt` in the current deployed template.

> The reason why some task definitions are unhotswappable is that there are such intrinsics in the template and the CDK fails to evaluate it locally. So the basic idea to overcome this limitation in this PR is that we don't try to evaluate it locally, but we fetch the latest task definition already deployed and get the required values from it.

So instead of going to CFN to get the values, which is what `EvaluteCfnTemplate` does, we go to the ECS service directly. Why? Is there something that `EvaluateCfnTemplate` can't handle?

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
tmokmss: Hi @comcalvi, thanks for looking at this!

> I'm curious what you mean by this. EvaluteCfnTemplate doesn't explicitly check any of these resources, it instead looks up every arg to GetAtt in the current deployed template.

We can't know the actual values for a GetAtt just by looking up the template. For example, given this string: `"Fn::GetAtt": ["SecurityGroupC96ED6A7", "GroupId"]`, how do we know that it should resolve to the security group id? In the current implementation, we have a manually created mapping table for the `GetAtt` functions, and  it can only evaluate one of the resource attributes defined here:

https://github.com/aws/aws-cdk/blob/5ccc56975c323ea19fd0917def51184e13f440d9/packages/aws-cdk/lib/api/evaluate-cloudformation-template.ts#L352-L365 

If we want to evaluate all attributes, we have to manually define these mappings one by one, for all resources and attributes, which is hardly a realistic solution. Since this problem makes ECS hotswap almost unusable for many real use cases, I believe it's worth optimizing the hotswap logic specifically for ECS.

> So instead of going to CFN to get the values, which is what EvaluteCfnTemplate does, we go to the ECS service directly. Why? Is there something that EvaluateCfnTemplate can't handle?

Yes, `EvaluateCfnTemplate` can't handle certain intrinsics, such as `GetAtt` for the attributes that don't have the mappings described above, or `Fn::ImportValue`, etc. An easy way to get the actual values for the attributes is to query the ECS service, which is done by this PR.

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
tmokmss: OK! Shouldn't we enforce this in eslintrc?

https://github.com/aws/aws-cdk/blob/1807f5754885e4b1b1c8d12ca7a1cc7efab9ef2c/tools/%40aws-cdk/cdk-build-tools/config/eslintrc.js#L65

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -167,13 +189,49 @@ export async function isHotswappableEcsServiceChange(
...
   readonly serviceArn: string;
 }
 
 async function prepareTaskDefinitionChange(
-  evaluateCfnTemplate: EvaluateCloudFormationTemplate, logicalId: string, change: HotswappableChangeCandidate,
-) {
+  evaluateCfnTemplate: EvaluateCloudFormationTemplate,
+  logicalId: string,
+  change: HotswappableChangeCandidate,
+): Promise<undefined | { taskDefinition: any; family: string; hotswappableWithMerge: boolean }> {
```
rix0rrr, line 234: Can you define a named type for this, and add comments to the fields?

Also, `any`s scare me. Can we type the taskDefinition a bit more thoroughly?

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -45,14 +46,18 @@ export async function isHotswappableEcsServiceChange(
   const namesOfHotswappableChanges = Object.keys(classifiedChanges.hotswappableProps);
   if (namesOfHotswappableChanges.length > 0) {
     const taskDefinitionResource = await prepareTaskDefinitionChange(evaluateCfnTemplate, logicalId, change);
+    if (taskDefinitionResource === undefined) {
+      reportNonHotswappableChange(ret, change, undefined, 'Found unsupported changes to the task definition', false);
```
rix0rrr, line 50: Would be nice if we can report the property path that we found unsupported changes in.

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -195,11 +253,93 @@ async function prepareTaskDefinitionChange(
...
+  if (oldDefinition.length !== newDefinition.length) {
+    // one or more containers are added or removed
+    return false;
+  }
+  for (const i in oldDefinition) {
+    result.push({});
+    const prev = oldDefinition[i];
+    const next = newDefinition[i];
+
+    if (!deepCompareObject(Object.keys(prev), Object.keys(next))) {
```
rix0rrr, line 305: I think this assumes that the keys are in the same order. Is that necessarily always true?

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -167,13 +189,49 @@ export async function isHotswappableEcsServiceChange(
...
+  const dst = target.taskDefinition?.containerDefinitions;
+  if (dst === undefined) {
+    return;
+  }
+  // schema: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html
+  for (const i in src) {
+    for (const key of Object.keys(src[i])) {
+      (dst[i] as any)[key] = src[i][key];
+    }
+  }
```
rix0rrr, line 203: Doesn't this need to be recursive?

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -195,11 +253,93 @@ async function prepareTaskDefinitionChange(
...
+    const prev = oldDefinition[i];
+    const next = newDefinition[i];
+
+    if (!deepCompareObject(Object.keys(prev), Object.keys(next))) {
+      // one or more fields are added or removed
+      return false;
+    }
+    for (const key of Object.keys(prev)) {
+      // compare two properties first
+      if (deepCompareObject(prev[key], next[key])) {
```
rix0rrr, line 311: Doesn't this need to be recursive? Or we're purposely not supporting the case where a simple env variable got added?  (That's fine, but I'd like that called out in comments)

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -167,13 +189,49 @@ export async function isHotswappableEcsServiceChange(
   return ret;
 }
 
+function mergeTaskDefinitions(patch: {containerDefinitions: {[key:string]: any}[]}, target: AWS.ECS.DescribeTaskDefinitionResponse) {
```
rix0rrr, line 192: It would be slightly nicer (imo) to have this return a new object, instead of mutating the existing one in-place.

rix0rrr (changes requested):
I like your thoughts and strategy! 

I understand that you're hesitant to change code paths that work, but alternatively: now we have 2 different code paths and that makes it harder to test for correctness.

So I'd rather collapse the two code paths into one, and always use your suggestion: download the existing TaskDefinition and apply changes to it. 

In fact, it's such a good idea, could you try writing the patcher code with an eye to using the same pattern for all hotswappable resources, and put it in a separate file?

Maybe something like:

```ts
function hotswappableProperties(template: EvaluateCloudFormationTemplate, change: HotswappableChangeCandidate): HotswappablePropertyUpdates {
  // ...
}

function applyPropertyUpdates(template: EvaluateCloudFormationTemplate, target: unknown) {
  // ...
}

interface HotswappablePropertyUpdates {
  readonly updates: Record<string, unknown>;
  readonly unevaluatableUpdates: string[];
}
```

The lack of typing makes me quite uneasy, but I'm not exactly sure what to do about that yet...

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
tmokmss: Since this compares the old and new CFn templates, I think yes. Otherwise we would have unnecessary diff even on usual CFn deployment.

https://github.com/aws/aws-cdk/blob/64f2c20310ba09673c37d3d6d8700f730b957bf0/packages/aws-cdk/lib/api/hotswap/ecs-services.ts#L265-L268

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
tmokmss: I added a named type.

I agree that `any` should be avoided, but actually it seems difficult in this case, because we are using a returned value from `evaluateCfnExpression` and its signature forces `any` as the returning type.

https://github.com/aws/aws-cdk/blob/64f2c20310ba09673c37d3d6d8700f730b957bf0/packages/aws-cdk/lib/api/hotswap/ecs-services.ts#L260-L263

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
tmokmss: Now I deep-copy the `target` argument. `target = JSON.parse(JSON.stringify(target));`

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
tmokmss: No, because we don't also recurse here: https://github.com/aws/aws-cdk/pull/26404#discussion_r1283073969

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
tmokmss: No, I'm afraid it overly complicates the code, while not adding much value to the actual hotswap experience. I added a comment: `We don't recurse properties to simplify the update logic. It should still cover most hotswap use cases.`

We might revisit this decision when someone actually needed the deep comparison :)

tmokmss: On second thought, I changed it to recurse anyway to make the logic more reusable.

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -79,7 +127,14 @@ export async function isHotswappableEcsServiceChange(
...
             },
           },
-        });
+        } as const;
+        // We first uppercase the task definition to properly merge it with the one from CloudFormation template.
+        const upperCasedTaskDef = transformObjectKeys(target.taskDefinition, upperCaseFirstCharacter, excludeFromTransform);
+        // merge evaluatable diff from CloudFormation template.
+        const updatedTaskDef = applyPropertyUpdates(changes.updates, upperCasedTaskDef);
+        // lowercase the merged task definition to use it in AWS SDK.
+        const lowercasedTaskDef = transformObjectKeys(updatedTaskDef, lowerCaseFirstCharacter, excludeFromTransform);
```
rix0rrr, line 136: Wouldn't it be shorter to transform the updates to lowercase?

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
```diff
@@ -117,6 +172,12 @@ export async function isHotswappableEcsServiceChange(
           }),
         );
 
+        if (taskDefRevArn !== undefined) {
+          // Make the task definition registered by hotswap INACTIVE immediately.
+          // This is to prevent describeTaskDefinition API from returning a drifted task definition.
```
rix0rrr, line 177: This I don't get, but I like that the comment is here! Can you try explaining when/why this matters?

If the problem is that when you do this method of hotswap multiple times the updates wouldn't apply anymore after the first hotswap, could we circumvent that by using `DescribeTemplateResources` to get the ARN of the TaskDefinition that we based the updates off of?

rix0rrr (commented):
Yes, exactly what I meant! ❤️ it!

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
tmokmss: I think no. There are certain keys that should not be transformed in a task definition, as specified in `excludeFromTransform`. If we transform the updates, it requires additional logic to determine if it's okay to transform the key or not because the schema is different in those extracted updates and we cannot reuse the `excludeFromTransform`.

packages/aws-cdk/lib/api/hotswap/ecs-services.ts
tmokmss: > If the problem is that when you do this method of hotswap multiple times the updates wouldn't apply anymore after the first hotswap

Yes, that is exactly the problem I want to solve. Since we are taking diff between the old and new template, a task definition to apply patches must be the one deployed by the old template.

And I agree with that using `DescribeTemplateResources` is more reliable. (Thanks, I didn't come up with that!) I'll fix this soon :)

tmokmss:
@rix0rrr I added/fixed unit tests and now it's ready for review :) Hope you can check this, thanks.

rix0rrr:
I like this a lot, thanks! Ready to ship it, but final check: did you do manual testing on the latest revision with all the requested changes in it?

Cheers!

tmokmss:
@rix0rrr Thanks :) Yes, I confirmed that I can successfully hotswap this CDK code: https://github.com/aws/aws-cdk/pull/26404#issuecomment-1640432823

rix0rrr (dismissed):

tmokmss:
@rix0rrr mergify was [failing](https://github.com/aws/aws-cdk/pull/26404/checks?check_run_id=15956994986) for some reason:

```
Mergify doesn't have permission to update

For security reasons, Mergify can't update this pull request. Try updating locally.
GitHub response: refusing to allow a GitHub App to create or update workflow .github/workflows/close-stale-prs.yml without workflows permission
err-code: 5E909
```

I guess merging with main manually will fix this issue. Can you approve it again? Thanks 🙏

Issue #26397, opened 2023-07-18
@aws-cdk/aws-redshift-alpha: Custom resource unable to find AWS SDK v3


dhruvsharma78:
### Describe the bug

The [DatabaseQuery](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-redshift-alpha/lib/private/database-query.ts) handler in @aws-cdk/aws-redshift-alpha uses AWS SDK v3 but is set to run in the Node14 runtime. This leads to a crash at runtime where the lambda cannot find SDK v3.

### Expected Behavior

The lambda should execute without error

### Current Behavior

In https://github.com/aws/aws-cdk/pull/25895, all alpha modules were updated to use AWS SDK v3, but a handler defined in the aws-redshift-alpha module (https://github.com/aws/aws-cdk/blob/6c75581ae2b9537fa9d1d724b837fe81ae22d345/packages/%40aws-cdk/aws-redshift-alpha/lib/private/database-query.ts) uses Node 16. This leads to this error:

```
{
    "errorType": "Runtime.ImportModuleError",
    "errorMessage": "Error: Cannot find module '@aws-sdk/client-redshift-data'\nRequire stack:\n- /var/task/redshift-data.js\n- /var/task/privileges.js\n- /var/task/index.js\n- /var/runtime/index.mjs",
    "stack": [
        "Runtime.ImportModuleError: Error: Cannot find module '@aws-sdk/client-redshift-data'",
        "Require stack:",
        "- /var/task/redshift-data.js",
        "- /var/task/privileges.js",
        "- /var/task/index.js",
        "- /var/runtime/index.mjs",
        "    at _loadUserApp (file:///var/runtime/index.mjs:997:17)",
        "    at async Object.UserFunction.js.module.exports.load (file:///var/runtime/index.mjs:1032:21)",
        "    at async start (file:///var/runtime/index.mjs:1195:23)",
        "    at async file:///var/runtime/index.mjs:1201:1"
    ]
}
```
### Possible Solution

* In https://github.com/aws/aws-cdk/pull/26212, all custom resources were migrated to use NodeJS v18, and also for all custom resources to use AWS SDK v3. Would it also be possible to upgrade [this](https://github.com/aws/aws-cdk/blob/6c75581ae2b9537fa9d1d724b837fe81ae22d345/packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/aws-custom-resource.ts) to Node18?
[...]

PR #26406, opened 2023-07-18
fix(redshift-alpha): incorrect CR runtime version

MrArnoldPalmer:
Fixes the database-query custom resource in redshift alpha to correctly use Node18 as the runtime version since it now uses sdk v3.

Fixes: #26397

----


manmartgarc:
Has this been relased in any version? I cannot the closed issue number on any release?

PR #26443, opened 2023-07-20
feat(stepfunctions): add stateMachineRevisionId property to StateMachine

### LINKED ISSUE #26440: (stepfunctions): Expose StateMachineRevisionId attribute from StateMachine L2 construct
### Describe the feature

[`AWS::StepFunctions::StateMachine`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-stepfunctions-statemachine.html) has a [`StateMachineRevisionId`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions.CfnStateMachine.html#attrstatemachinerevisionid) read-only attribute which changes when the state machine is updated and produces a new revision. It can be used with [`CfnStateMachineVersion`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions.CfnStateMachineVersion.html) to automatically create a new state machine version when there is a new revision.

[StateMachine](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions.StateMachine.html) should expose this property directly for convenience. 

The current workaround is read the value from the L1. For example:
```javascript
// StateMachine
const stateMachine = new sfn.StateMachine(this, 'myStateMachine', {
  definition: new sfn.Pass(this, 'DefineRepositories', { result: sfn.Result.fromString("version 2") })
});

// StateMachineRevisionId is not exposed by the L2 Construct yet,
// so we need to reach down to the L1 to read the attribute
const cfnStateMachine = stateMachine.node.defaultChild as sfn.CfnStateMachine;
const revisionId = cfnStateMachine.attrStateMachineRevisionId

// Version that always points to the latest revision by using `stateMachineRevisionId`
const version = new sfn.CfnStateMachineVersion(this, 'Version', {
  stateMachineArn: stateMachine.stateMachineArn,
  stateMachineRevisionId: revisionId,
});
```

### Use Case

StateMachineRevisionId is not accessible without using the L1. See above

### Proposed Solution

Add a readonly property `stateMachineRevisionId` to `StateMachine`'s properties. It can be assigned a value after the constructor creates the CfnStateMachine instance here: https://github.com/aws/aws-cdk/blob/efb0a4c785bb794e3ee2e7444b29e4a00d7d28f1/packages/aws-cdk-lib/aws-stepfunctions/lib/state-machine.ts#L434C11-L442

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.87.0

MacOS 12.6.5

wong-a:
Expose `stateMachineRevisionId` as a readonly property to StateMachine whose value is a reference to the `StateMachineRevisionId` attribute of the underlying CloudFormation resource.

Closes #26440

----


packages/aws-cdk-lib/aws-stepfunctions/lib/state-machine.ts
```diff
@@ -407,6 +407,12 @@ export class StateMachine extends StateMachineBase {
    */
   public readonly stateMachineType: StateMachineType;
 
+  /**
+   * Identifier for the state machine revision
```
mrgrain, line 411:
```suggestion
   * Identifier for the state machine revision, which is an immutable, read-only snapshot of a state machine’s definition and configuration.
```

mrgrain (dismissed):

evscott (approved):
Nice!

mrgrain (approved):

PR #26450, opened 2023-07-20
docs(triggers): add consideration about invocationType for the behavior of trigger failures

### LINKED ISSUE #26341: aws-cdk-lib/triggers: Confusing/missing explanation on trigger failures
### Describe the issue

From the docs about trigger failures:

> If the trigger handler fails (e.g. an exception is raised), the CloudFormation deployment will fail, as if a resource failed to provision.

However, when the `invocationType` in the trigger's props is `triggers.InvocationType.EVENT` a deployment will succeed regardless of if an error is thrown in the handler body. The quoted statement is true if `triggers.InvocationType.REQUEST_RESPONSE` is used however. Since the example code snippet of a trigger uses `triggers.InvocationType.EVENT` it gives the impression that the deployment will fail if an exception is raised in the handler for the event. Perhaps this is an actual bug in this case, but if it works as intended then it would be good to be explicit about the effect of the different `invocationType`s in the documentation.

### Links

[aws-cdk-lib](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.triggers-readme.html#trigger-failures)

tam0ri:
TriggerCustomResourceProvider takes only the status code of Invoke API call into account.
https://github.com/aws/aws-cdk/blob/7a6f953fe5a4d7e0ba5833f06596b132c95e0709/packages/aws-cdk-lib/triggers/lib/lambda/index.ts#L69-L73

If invocationType is `EVENT`, Lambda function is invoked asynchronously. In that case, if Lambda function is invoked successfully, the trigger will success regardless of the result for the function execution. I add this consideration into README.

Closes #26341

----


packages/aws-cdk-lib/triggers/README.md
```diff
@@ -57,6 +57,8 @@ deployment will fail, as if a resource failed to provision. This makes it easy
 to implement "self tests" via triggers by simply making a set of assertions on
 some provisioned infrastructure.
 
+Note that this behavior is only applied when invocationType is `REQUEST_RESPONSE`. When invocationType is `EVENT`, Lambda function is invoked asynchronously. In that case, if Lambda function is invoked successfully, the trigger will success regardless of the result for the function execution.
```
kaizencc, line 60:
```suggestion
Note that this behavior is only applied when invocationType is `REQUEST_RESPONSE`. When invocationType is `EVENT`, Lambda function is invoked asynchronously.
In that case, if Lambda function is invoked successfully, the trigger will success regardless of the result for the function execution.
```

daschaa (approved):

kaizencc (approved):

PR #26466, opened 2023-07-21
fix(core): policy validation trace incorrect for larger constructs

corymhall:
Ran into an issue where the construct trace was incorrect in larger projects, specifically where there are constructs that contain multiple constructs.

To get the construct trace tree we first construct a new tree that only contains the path to the terminal construct (the one with the trace). We drop all the other constructs in the tree that don't relate. For example, if we had a tree like:

```
->App
-->MyStage
--->MyStack1
---->MyConstruct
----->Resource
--->MyStack2
---->MyConstruct
----->Resource
--->MyStack3
---->MyConstruct
----->Resource
```

And we want to get a new tree for
`/App/MyStage/MyStack2/MyConstruct/Resource` it should look like this:

```
->App
-->MyStage
--->MyStack2
---->MyConstruct
----->Resource
```

We weren't correctly generating the new tree correctly and would always end up with the tree being the first item in the list.

I've updated one of the tests, and also tested this on a more complex application.

----


packages/aws-cdk-lib/core/lib/private/tree-metadata.ts
```diff
@@ -109,21 +111,37 @@ export class TreeMetadata extends Construct {
...
+    function renderTreeWithSingleChild(currentNode: Mutable<Node>, currentNodeChild: Mutable<Node>) {
+      currentNode.children = {
+        [currentNodeChild.id]: currentNodeChild,
       };
+      if (currentNode.parent) {
+        currentNode.parent = renderTreeWithSingleChild(currentNode.parent, currentNode);
+      }
+      // we just reset the parent, but the currentNodeChild still has the original
+      // parent. Replace that as well
+      currentNode.children[currentNodeChild.id].parent = currentNode;
```
otaviomacedo, line 131: I don't understand this. Isn't the parent of the children of `x` already `x` by definition?

otaviomacedo (approved):

packages/aws-cdk-lib/core/lib/private/tree-metadata.ts
corymhall: yeah you are right. My thinking was left over from a previous solution that wasn't mutating the object.

Updated 👍

PR #26475, opened 2023-07-23
chore(lambda): update default node runtime for all partitions

### LINKED ISSUE #26461: (lambda): (CDK lambdas still using node16)
### Describe the bug

Since https://github.com/aws/aws-cdk/releases/tag/v2.87.0 CDK lambda runtime should be using node 18 runtime.

We are seeing some runtimes still using node 16 e.g. LogRetention lambda. (there may be more)

https://github.com/aws/aws-cdk/blob/v2.87.0/packages/aws-cdk-lib/aws-logs/lib/log-retention.ts

image

it seems to be defaulting to node_16 if a regional Fact is not set – where is this set? I couldn’t see anything in release notes relating to this. is this something AWS configures  so that node 18 is used? Or do we have to do something?

### Expected Behavior

node 18 runtime should be used in *all* CDK generated lambdas 

### Current Behavior

Node 16 is being used in some CDK generated lambdas

### Reproduction Steps

this is autogenerated by CDK

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.87.0

### Framework Version

2.87.0

### Node.js Version

16

### OS

Ubuntu standard:6.0

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

lpizzinidev:
Node18 is the default runtime environment for lambdas in the [default](https://aws.amazon.com/it/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/), [China](https://www.amazonaws.cn/en/new/2023/amazon-lambda-adds-support-for-node-js-18-in-amazon-web-services-china-regions/), and [GovCloud](https://aws.amazon.com/it/about-aws/whats-new/2023/04/aws-lambda-node-js-18-govcloud-regions/) regions.

This fix updates the runtime default versions for constructs that were still using older versions, in particular:

- `LogRetention`
- `AwsCustomResource`
- `CustomResourceProvider`

Also, it updates the `DefaultCrNodeVersionMap` in the fact tables to reflect the updated values.

Closes #26461.

----


lpizzinidev:
Clarification Request.
Is an integration test required in this case?

kaizencc (dismissed):

kaizencc (commented):
received a tip about a custom resource potentially not working with node18. so i'm gonna run it to check

kaizencc:
Hmm I got this error on delete when running log-retention integ. Might not be ready for node18 yet but I haven't had time to dive in and figure out whats necessary to fix

```
{
    "errorType": "Runtime.ImportModuleError",
    "errorMessage": "Error: Cannot find module 'aws-sdk'\nRequire stack:\n- /var/task/index.js\n- /var/runtime/index.mjs",
    "stack": [
        "Runtime.ImportModuleError: Error: Cannot find module 'aws-sdk'",
        "Require stack:",
        "- /var/task/index.js",
        "- /var/runtime/index.mjs",
        "    at _loadUserApp (file:///var/runtime/index.mjs:997:17)",
        "    at async UserFunction.js.module.exports.load (file:///var/runtime/index.mjs:1029:21)",
        "    at async start (file:///var/runtime/index.mjs:1192:23)",
        "    at async file:///var/runtime/index.mjs:1198:1"
    ]
}
111

**Comment by rix0rrr** [comment]
I am surprised that this would still be necessary. I thought we just did a big campaign upgrading aws-sdkv2 -> v3 everywhere, which should have included this.

File: packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts [hunk 1]
```diff
@@ -123,9 +123,9 @@ export const PARTITION_MAP: { [region: string]: Region } = {
...
 
 export const CR_DEFAULT_RUNTIME_MAP: Record<Partition, string> = {
-  [Partition.Default]: 'nodejs16.x',
-  [Partition.Cn]: 'nodejs16.x',
-  [Partition.UsGov]: 'nodejs16.x',
+  [Partition.Default]: 'nodejs18.x',
+  [Partition.Cn]: 'nodejs18.x',
+  [Partition.UsGov]: 'nodejs18.x',
   [Partition.UsIso]: 'nodejs14.x',
   [Partition.UsIsoB]: 'nodejs14.x',
```
rix0rrr, line 130: Since we rely on the AWS SDK that Lambda builds into the runtime, either all regions have to be Node18 (SDK v3) or < Node 18 (SDK v2).

We can't have a mix of 14 and 18 using one code base.

lpizzinidev: Thanks for the review!
So moving `UsIso` and `UsIsoB` to Node 18 would solve [this](https://github.com/aws/aws-cdk/pull/26475#issuecomment-1650492079) runtime error?

mrgrain: @lpizzinidev I'm working on the runtime issue. I think we will also take over this PR and extend it to what needs to change.

lpizzinidev: @mrgrain 👍 Let me know if you need me to apply further changes

mrgrain: Thanks @lpizzinidev  If you wanted to update the iso regions to node18 as well and then your PR should be ready to go once https://github.com/aws/aws-cdk/pull/26609 is merged. 🚀

rix0rrr (approved):

PR #26498, opened 2023-07-25
feat(logs): configure custom subscription filter name

### LINKED ISSUE #26485: (logs): Configure subscription filter name on l2
### Describe the feature

Would like to set the subscription filter name as a prop on the L2 https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_logs.SubscriptionFilter.html

### Use Case

To set a specific name without requiring escape hatches

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

latest

mac

tam0ri:
Currently, we can't set the subscription filter name as a prop for L2 SubscriptionFilter construct. This PR introduces the new prop `filterName`. This will let us set a specific name without requiring escape hatches.

Closes #26485

----


packages/aws-cdk-lib/aws-logs/lib/log-group.ts
```diff
@@ -524,6 +524,11 @@ export interface SubscriptionFilterOptions {
    * Log events matching this pattern will be sent to the destination.
    */
   readonly filterPattern: IFilterPattern;
+
+  /**
+   * The name of the subscription filter.
+   */
+  readonly filterName?: string;
```
kaizencc, line 531: need a default here

packages/aws-cdk-lib/aws-logs/lib/subscription-filter.ts
```diff
@@ -62,6 +64,7 @@ export class SubscriptionFilter extends Resource {
       destinationArn: destProps.arn,
       roleArn: destProps.role && destProps.role.roleArn,
       filterPattern: props.filterPattern.logPatternString,
+      filterName: this.physicalName,
```
kaizencc, line 67: wait, won't this change all subscription filters even if they don't supply `props.filterName`? Why is there a need to propagate the `filterName` to `physicalName` this way

kaizencc (dismissed):
Couple comments @tam0ri. Thanks for taking this on!

packages/aws-cdk-lib/aws-logs/lib/subscription-filter.ts
tam0ri: Thank you for your comment!

---

> won't this change all subscription filters even if they don't supply props.filterName? 

In my understanding, if we don't specify `filterName` prop introduced by this PR, `this.physicalName` will be undefined and then `filterName` for L1 CfnSubscriptionFilter construct will be also undefined. As a result, FilterName porperty for [AWS::Logs::SubscriptionFilter](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-subscriptionfilter.html) resource in generted CFn template will be omitted. I already tested with the sample app in my env. 

---

> Why is there a need to propagate the filterName to physicalName this way

I align this with the similar constructs such as LogGroup, LogStream or MetricFilter as below.

https://github.com/aws/aws-cdk/blob/51b20f9d07f66ddf677cea62c75c13f20b066293/packages/aws-cdk-lib/aws-logs/lib/log-group.ts#L466-L469
https://github.com/aws/aws-cdk/blob/9a3399f11c18cb8c2f5a92dd3a6773681aa77e39/packages/aws-cdk-lib/aws-logs/lib/log-stream.ts#L67-L70
https://github.com/aws/aws-cdk/blob/9a3399f11c18cb8c2f5a92dd3a6773681aa77e39/packages/aws-cdk-lib/aws-logs/lib/metric-filter.ts#L25-L28

From CFn perspective, if we specify `FilterName` property, this is used as physical id. Shouldn't we propagate the filterName to physicalName?

packages/aws-cdk-lib/aws-logs/lib/log-group.ts
tam0ri: Thank you for your comment! Sorry, I don't understand your intention. What do you mean `need a default`? As I described in the below conversation, I think it's better that filterName is optional because this can still let CDK/CFn generate physical id (filter name) automatically when this prop isn't specified.

kaizencc: For optional properties, we supply a default value in the documentation so our users know what it means when they don't supply anything:

```ts
/**
 * The name of the subscription filter.
 *
 * @default - a generated name is created for you
 */
readonly filterName?: string;
```

packages/aws-cdk-lib/aws-logs/lib/subscription-filter.ts
kaizencc: We can't, because that would be a breaking change for users who don't specify `filterName`. The behavior that existed before needs to be maintained, unfortunately.

In practice, if we really want to change the behavior we hide it behind a feature flag. But this ins't important enough to go down that route, imo.

kaizencc: Oh wait you're saying `this.physicalName` will be undefined if no `filterName` is given... hmm maybe thats the case but I would be surprised.

kaizencc: Nevermind, you are right :)

packages/aws-cdk-lib/aws-logs/lib/log-group.ts
tam0ri: Makes sense! I aligned with logGroupName and logStreamName.

https://github.com/aws/aws-cdk/blob/e53e8dcede7b2ffc84f8f5151eff9ace3b005684/packages/aws-cdk-lib/aws-logs/lib/log-group.ts#L381-L386
https://github.com/aws/aws-cdk/blob/e53e8dcede7b2ffc84f8f5151eff9ace3b005684/packages/aws-cdk-lib/aws-logs/lib/log-group.ts#L502-L509

kaizencc (approved):

Issue #26484, opened 2023-07-24
(aws-batch-alpha): Construct Props RuntimePlatform


gmarchand:
### Describe the feature

The property [RuntimePlatform](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-containerproperties.html#cfn-batch-jobdefinition-containerproperties-runtimeplatform) is not present in the [interface IEcsContainerDefinition](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-batch-alpha.IEcsContainerDefinition.html)

### Use Case

I would want to use AWS Batch with Fargate on ARM64 as describe in this news : https://aws.amazon.com/about-aws/whats-new/2023/07/aws-batch-fargate-linux-arm64-windows-x86-containers-cli-sdk/

>To do that customers will need to create ARM64 or Windows job definition using RegisterJobDefinition API of AWS Batch. ARM64-based and Windows-based containers are supported in all Fargate on-demand compute environments.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.88.0

### Environment details (OS name and version, etc.)

All OS

pahud:
Makes sense to me. We could add [RuntimePlatform](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-containerproperties.html#cfn-batch-jobdefinition-containerproperties-runtimeplatform) in https://github.com/aws/aws-cdk/blob/f22bd4e2b1914b42450ffa061d27009039469b2b/packages/%40aws-cdk/aws-batch-alpha/lib/ecs-container-definition.ts#L361

khushail:
working on it.

gmarchand:
Hello @khushail @pahud the bot @aws-cdk-automation closed your PR last week https://github.com/aws/aws-cdk/pull/26506#issuecomment-1722070067
And the release https://github.com/aws/aws-cdk/releases/tag/v2.96.0 has stabilized the aws-batch module

Do you think your code is still up to date ?

khushail:
Hi @gmarchand , I need to fix some comments, will do it this week hopefully and update the PR. Thanks for checking.

gmarchand:
Hello, Any update on this one ?

khushail:
[...]

PR #26506, opened 2023-07-26
feat(batch-alpha): add RuntimePlatform prop

khushail:
Closes #[26484](https://github.com/aws/aws-cdk/issues/26484).

----

### DIFF
--- a/packages/@aws-cdk/aws-batch-alpha/lib/ecs-container-definition.ts
+++ b/packages/@aws-cdk/aws-batch-alpha/lib/ecs-container-definition.ts
@@ -667,6 +667,10 @@ abstract class EcsContainerDefinitionBase extends Construct implements IEcsConta
       logConfiguration: this.logDriverConfig,
       readonlyRootFilesystem: this.readonlyRootFilesystem,
       resourceRequirements: this._renderResourceRequirements(),
+      runtimePlatform: {
+        cpuArchitecture: this.runtimePlatform?.cpuArchitecture?.toString(),
+        operatingSystemFamily: this.runtimePlatform?.operatingSystemFamily?.toString(),
+      },
       secrets: this.secrets ? Object.entries(this.secrets).map(([name, secret]) => {
         secret.grantRead(this.executionRole);
 


packages/@aws-cdk/aws-batch-alpha/lib/ecs-container-definition.ts
```diff
@@ -618,6 +638,10 @@ abstract class EcsContainerDefinitionBase extends Construct implements IEcsConta
     this.user = props.user;
     this.volumes = props.volumes ?? [];
 
+    if (props.runtimePlatform) {
+      this.runtimePlatform = props.runtimePlatform;
+    }
+
```
pahud, line 644: I think you should modify the  _renderContainerDefinition() function to add the new `runtimePlatform` property like

```
runtimePlatform: this.runtimePlatform
```

So it will synth a new container definition for [ContainerProperties](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-containerproperties.html)

pahud (changes requested):

pahud:
OK looks like the RuntimePlatform support was just added into cloudformation in Aug 2023
image

AWS CDK at this moment uses CFN spec `130.1.0` which does not include this property yet.
https://github.com/aws/aws-cdk/blob/ed9b53789a0c7de44ffdc20a047110ca444d44bd/packages/%40aws-cdk/cfnspec/spec-source/specification/000_cfn/000_official/000_AWS_Batch.json#L2

cfnspec 130.1.0 was PR merged in July 14
https://github.com/aws/aws-cdk/pull/26362

The latest CFN spec is 135.0.0 which DOES have `runtimePlatform` property.
https://d1uauaxba7bl26.cloudfront.net/latest/gzip/CloudFormationResourceSpecification.json

We will need to wait another PR to support 135.0.0 or above before we are allowed to continue this PR.

pahud:
I was wrong. CDK now uses awscdk-service-spec as the source of the truth for the L1 and the latest definition can be found here:
https://github.com/cdklabs/awscdk-service-spec/blob/main/sources/CloudFormationSchema/us-east-1/aws-batch-jobdefinition.json

And I can confirm `ContainerProperties.RuntimePlatform` is already in the definition so I think it should be available in the L1 props.

```
    "ContainerProperties" : {
      "type" : "object",
      "additionalProperties" : false,
      "properties" : {
        "User" : {
          "type" : "string"
        },
        "Secrets" : {
          "type" : "array",
          "uniqueItems" : false,
          "items" : {
            "$ref" : "#/definitions/Secret"
          }
        },
        "Memory" : {
          "type" : "integer"
        },
        "Privileged" : {
          "type" : "boolean"
        },
        "LinuxParameters" : {
          "$ref" : "#/definitions/LinuxParameters"
        },
        "FargatePlatformConfiguration" : {
          "$ref" : "#/definitions/FargatePlatformConfiguration"
        },
        "JobRoleArn" : {
          "type" : "string"
        },
        "ReadonlyRootFilesystem" : {
          "type" : "boolean"
        },
        "Vcpus" : {
          "type" : "integer"
        },
        "Image" : {
          "type" : "string"
        },
        "ResourceRequirements" : {
          "type" : "array",
          "uniqueItems" : false,
          "items" : {
            "$ref" : "#/definitions/ResourceRequirement"
          }
        },
        "LogConfiguration" : {
          "$ref" : "#/definitions/LogConfiguration"
        },
        "MountPoints" : {
          "type" : "array",
          "uniqueItems" : false,
          "items" : {
            "$ref" : "#/definitions/MountPoints"
          }
        },
        "ExecutionRoleArn" : {
          "type" : "string"
        },
        "RuntimePlatform" : {
          "$ref" : "#/definitions/RuntimePlatform"
        },
```

packages/@aws-cdk/aws-batch-alpha/lib/ecs-container-definition.ts
```diff
@@ -429,6 +430,15 @@ export interface IEcsContainerDefinition extends IConstruct {
    */
   readonly readonlyRootFilesystem?: boolean;
 
+  /**
+   * The operating system that your task definitions are running on.
+   *
+   * A runtimePlatform is supported only for tasks using the Fargate launch type.
+   *
+   * @default - Undefined.
```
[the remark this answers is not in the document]
rix0rrr: Does this property need to be on the interface? `IEcsContainerDefinition` ?

It only needs to be here if downstream consumers of this class need to make decisions based on `runtimePlatform`. If so, it also needs to be added to `EcsContainerDefinition.fromContainerDefinitionAttributes()`.

packages/@aws-cdk/aws-batch-alpha/lib/ecs-container-definition.ts
```diff
@@ -546,6 +556,15 @@ export interface EcsContainerDefinitionProps {
...
+   */
+  readonly runtimePlatform?: RuntimePlatform;
```
rix0rrr, line 566: Instead of having this subproperty, how about we inline the 2 fields from `RuntimePlatform` here? It will be easier to work with for users.

[Prefer flat properties](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#flat), CDK Design Guidelines, section *Props*.

rix0rrr:
Also this is not a `chore`, `feat` or `fix` are better choices.

gmarchand:
Hello @pahud, Sorry I can not help you on this one, I don't have the skills to contribute. But the need is always here.

mrgrain:
@comcalvi FYI

PR #26528, opened 2023-07-27
refactor(lambda): missing Lambda VPC and security validation

### LINKED ISSUE #26508: aws-lambda: VpcConfig is not configured for Lambda Function without specifying Vpc prop
### Describe the bug

Specifying `securityGroups` without specifying `vpc` prop in Lambda Function leads to `VpcConfig` not added to the generated CloudFormation Template.

### Expected Behavior

VpcConfig is added to Lambda Function:
```json
  "functionF19B1A04": {
   "Type": "AWS::Lambda::Function",
   "Properties": {
    "Code": {
     "ZipFile": "..."
    },
    "Handler": "index.handler",
    "Role": {
     "Fn::GetAtt": [
      "functionServiceRoleEF216095",
      "Arn"
     ]
    },
    "Runtime": "nodejs18.x",
    "VpcConfig": { // VpcConfig is added here
        ...
    }
   },
   "DependsOn": [
    ...
   ]
```

### Current Behavior

VpcConfig is not in Lambda Function

```json
  "functionF19B1A04": {
   "Type": "AWS::Lambda::Function",
   "Properties": {
    "Code": {
     "ZipFile": "..."
    },
    "Handler": "index.handler",
    "Role": {
     "Fn::GetAtt": [
      "functionServiceRoleEF216095",
      "Arn"
     ]
    },
    "Runtime": "nodejs18.x"
   },
   "DependsOn": [
    "functionServiceRoleEF216095"
   ],
   "Metadata": {
    "aws:cdk:path": "TestStack/function/Resource"
   }
  },
```

### Reproduction Steps

The following code does not add `VpcConfig` to CloudFormation template for Lambda function:

```typescript
import * as ec2 from 'aws-cdk-lib/aws-ec2'
import * as lambda from 'aws-cdk-lib/aws-lambda';

export class TestStack extends cdk.Stack {
    super(scope, id, props);

    const vpc = new ec2.Vpc(this, "vpc");
    const sg = new ec2.SecurityGroup(this, "sg", {
      vpc
    });

    const fn = new lambda.Function(this, 'function', {
      handler: 'index.handler',
      code: lambda.Code.fromInline("..."),
      securityGroups:[sg],
    });
  }
}
```

However, by adding `vpc` to `props`, now `VpcConfig` is added to the CloudFormation template

```typescript
    const fn = new lambda.Function(this, 'function', {
      handler: 'index.handler',
      code: lambda.Code.fromInline("..."),
      securityGroups:[sg],
      vpc, // Add vpc to prop
    });
```

### Possible Solution

`securityGroup` (no s) is deprecated and CDK is not throwing error if `securityGroups` is specified and `vpc` is not specified

https://github.com/aws/aws-cdk/blob/39a1d6bf5b5d87d73ec10de1a7a74b4cc4d1ac66/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L1225-L1227

_No response_

### CDK CLI Version

2.87.0

### Framework Version

_No response_

### Node.js Version

18

### OS

Window

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

lpizzinidev:
Specifying the `securityGroups` property requires a `vpc`.

This fix adds validation for the case when a `vpc` is not specified, but `securityGroups` is.
It also adds validation for the case when both `securityGroups` and `allowAllOutbound` are specified (`allowAllOutbound` should be configured in the SGs).

**Question for the reviewers**
How should we handle the case of an empty list in `securityGroups`? (eg `securityGroups: []`)

Closes #26508.

----


lpizzinidev:
Exemption Request.
The unit tests should already cover all the validation cases for the related function.

rix0rrr (approved):

rix0rrr (approved):

mbhaalalhhaexchange:
When **we are using the Security group without VPC  for @aws-cdk/aws-lambda package it was working well** but **when we have migrate it to the aws-cdk-lib/aws-lambda** package **we are getting following error message**,
**Cannot configure 'securityGroups' without configuring a VPC**

Kindly suggest the possible solution for it.

lpizzinidev:
Hi @mbhaalalhhaexchange ✋
You should provide a `vpc` for your function if you want to specify the `securityGroups` parameter ([validation](https://github.com/aws/aws-cdk/blob/10ed1948beb0f83c1b978da9c0a656aa01a382cb/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L1382)).

mbhaalalhhaexchange:
Hi @lpizzinidev , It was working fine with aws cdk v1 library (@aws-cdk/aws-lambda). So, why it is changed it in aws cdk v2 library ?

lpizzinidev:
@mbhaalalhhaexchange 
Can you please provide info about the v1 version that you were using and a snippet of the code where you initialize your functions?

mbhaalalhhaexchange:
Hi @lpizzinidev , 
Below are the code snippet that we are using,

`    super(scope, id, {
      ...props,
      vpc: undefined,
      code: resolve(__dirname, '..', 'dist', 'src'),
      lambdas: lambdas.map((lambda) => ({
        ...lambda,
        environment: {
          ...(lambda.environment || {}),
          DB_SECRET_ARN: props.dbSecretArn,
          LOGGING_LEVEL: 'debug',
          SQS_CALL_INFO_QUEUE: props.sqsCallInfoQueue.queueUrl
        },
        securityGroups: [props.securityGroup],
        policyStatements: [
          ...(lambda.policyStatements || []),
          new PolicyStatement({
            actions: ['secretsmanager:GetSecretValue'],
            resources: [props.dbSecretArn]
          }),
          new PolicyStatement({
            actions: ['kms:Decrypt'],
            resources: [props.dbSecretKmsKeyArn]
          }),
          new PolicyStatement({
            actions: ['cloudwatch:PutMetricData'],
            resources: ['*']
          }),
          new PolicyStatement({
            actions: ['sqs:SendMessage'],
            resources: [props.sqsCallInfoQueue.queueArn]
          }),
          new PolicyStatement({
            actions: ['connect:StartOutboundVoiceContact'],
            resources: ['*']
          })
        ],
        memorySize: lambda.memorySize || 512,
        retryAttempts: 0,
        lambdaTimeout: lambda?.lambdaTimeout || 8,
        runtime: Runtime.NODEJS_18_X
      })),
      layer: {
        folderPath: resolve(__dirname, '..', 'dist', 'lambdaLayer'),
        compatibleRuntimes: [Runtime.NODEJS_18_X]
      }

    });`
    
    **FYI** - This **code snippet was working very well with AWS CDK V1 version for @aws-cdk/aws-lambda** package.
    **Same code snippet is not working** and get this VPC related error while u**sing AWS CDK V2 version for aws-cdk-lib/aws-lambda** package.

lpizzinidev:
@mbhaalalhhaexchange 
`securityGroups` are [Only used if 'vpc' is supplied](https://github.com/aws/aws-cdk/blob/10ed1948beb0f83c1b978da9c0a656aa01a382cb/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L249).
This PR only adds validation for it, your code was working, but, since `vpc: undefined` the `securityGroups` you specified were ignored.

You can either:
1. Specify a `vpc` for your function (will update the deployment as the function will be placed in a custom VPC)
2. Remove `securityGroups: props.securityGroups` (won't change the deployment as the attribute was already ignored)

mbhaalalhhaexchange:
Hi @lpizzinidev , It is understandable this changes is introduced in AWS CDK V2 right ? 
For AWS-CDK V1 it was optional to pass VPC even if security groups are specified because we never got this type of error with same code snippet with CDK C1 library.

lpizzinidev:
@mbhaalalhhaexchange 

> It is understandable this changes is introduced in AWS CDK V2 right ?

Yep.

> For AWS-CDK V1 it was optional to pass VPC even if security groups are specified

Both parameters are still optional, we just added validation as specifying `securityGroups` or `securityGroup` (deprecated) without `vpc` ignores the parameters.

If you don't need to change your implementation, you can remove the `securityGroups: props.securityGroups` property and it will work as before (feel free to follow up if you encounter any errors after the change).

mbhaalalhhaexchange:
Hi @lpizzinidev , We can understand that but we want to make sure like same code should be work with AWS CDK V2  as well but due to this validation , also need to modify code and i am feeling worry about for future overheads.

mbhaalalhhaexchange:
After doing some research , some articles are claiming that AWS Team has introduced this change due to security concerns but here I do not see any security concerns.

If we do not pass to VPC and only want to pass security groups then it should allow. This condition is forcing us to do required changes in our code.

What's your thoughts on this ?

mbhaalalhhaexchange:
Hi @lpizzinidev ,  What's your thoughts on this ?

lpizzinidev:
@mbhaalalhhaexchange 
Passing security groups to a Lambda outside a VPC has no effect. That's why the validation was added.
Your current implementation with v1 is passing the property, but that value is not used as `vpc: undefined`.
If you want to manage the security groups of your function you'll need to associate it with a VPC (whether you stay on v1 or upgrade to v2).
If you want to keep the current implementation and don't manage the security groups you'll need to avoid passing the `securityGroups` property (which was silently ignored in v1).

mbhaalalhhaexchange:
Hi @lpizzinidev , It is understood but it is strange that AWS has silently ignored in v1.

PR #26529, opened 2023-07-27
fix(glue): Make targetTable optional

moomindani:
This PR makes `targetTable` parameter optional when defining Glue Data Quality Rulesets.

----


moomindani:
Exemption Request: This is a minor fix to make one parameter optional. No need to add integration test.

daschaa (changes requested):
Even if this is a minor change, I would prefer to have an integration test for this change. Is there a strong reason to not do an integration test for this?

packages/@aws-cdk/aws-glue-alpha/lib/data-quality-ruleset.ts
```diff
@@ -75,7 +75,7 @@ export interface DataQualityRulesetProps {
    * The target table of the ruleset
    * @attribute
```
lpizzinidev, line 76: The `@default` value or behavior for the optional variable should be specified.

lpizzinidev (changes requested):
I agree with @daschaa that an integration test should be provided to verify that a `DataQualityRuleset` without a `targetTable` deploys correctly.
You should be able to use [this](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-glue-alpha/test/integ.data-quality-ruleset.ts) test as a blueprint.

moomindani:
Sure I will do that.

moomindani:
This is waiting for one dependency being solved. Will work on it when it becomes actionable.

PR #26531, opened 2023-07-27
chore: update aws service spec in a separate job

mrgrain:
Currently this is just part of the `Yarn upgrade` workflow. 
We want a separate workflow so the updates are visible in the CHANGELOG.

This should also reduce potential failures in the `Yarn upgrade` workflow and vice-versa, so they don't block each other.

Tested o fork: https://github.com/mrgrain/aws-cdk/actions/runs/5680204361

----


.github/workflows/yarn-upgrade.yml
```diff
@@ -50,19 +50,20 @@ jobs:
...
+        # Upgrade special cases:
+        # - Various `@types/*` packages need to be pinned to specific versions due to breaking changes in minor upgrades https://github.com/DefinitelyTyped/DefinitelyTyped/issues/64266
+        # - `typescript` is not semantically versioned, so we only upgrade the patch version
+        # - `constructs` because we need to stay in control of the minimum compatible version
+        # - `aws-sdk-mock` because of breaking changes in type exports https://github.com/dwyl/aws-sdk-mock/pull/260. We are not respecting `@ts-ignore`.
+        # - `@aws-cdk/aws-service-spec` and `@aws-cdk/service-spec-types` have their own update workflow
         run: |-
           # Upgrade dependencies at repository root
           ncu --upgrade --filter=typescript --target=patch
-          ncu --upgrade --reject=@types/node,@types/prettier,constructs,typescript,lerna --target=minor
```
mrgrain, line 61: Special case notice was:

> We special-case lerna because we have a patch on it that stops applying if Lerna upgrades. Remove this once https://github.com/lerna/lerna/pull/2874 releases.

The linked PR has now been published and we have already upgraded lerna manually beyond it

kaizencc (approved):

PR #26535, opened 2023-07-27
chore(prlint): trigger pr linter from pr reviews

kaizencc:
This ends up being a lot of work. Essentially, the `pull_request_review` trigger runs the action on the merge branch which does not have the necessary secrets. This workflow has been inspired from github [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow) on what to do in this situation.

The PR linter takes the commit sha and pr number from the `pull_request_target` workflow and assumes that those values are available. So we will take those values from `pull_request_review`, upload them as artifacts and then download them when the pr linter action is triggered by the workflow run.

The code in `index.ts` hasn't been tested, but should work.

----


.github/workflows/pr-linter-trigger.yml
```diff
@@ -0,0 +1,23 @@
+name: Pr Linter Trigger
```
mrgrain, line 1:
```suggestion
name: PRLinter Trigger
```

.github/workflows/pr-linter.yml
```diff
@@ -25,8 +68,8 @@ jobs:
       statuses: read
       issues: read
     runs-on: ubuntu-latest
+    needs: download-if-workflow-run
```
mrgrain, line 71: Doesn't this mean `download-if-workflow-run` always has to run and complete?

.github/workflows/pr-linter.yml
```diff
@@ -11,12 +11,55 @@ on:
       - opened
       - synchronize
       - reopened
-  pull_request_review:
+  workflow_run:
+    workflows: [PR Linter Trigger]
```
kaizencc, line 15:
```suggestion
    workflows: [PRLinter Trigger]
```

mrgrain (commented):
@kaizencc Can you add the README entry?

.github/workflows/pr-linter-trigger.yml
```diff
@@ -0,0 +1,23 @@
+name: PRLinter Trigger
```
mrgrain, line 1:
```suggestion
name: PR Linter Trigger
```
FFS!

.github/workflows/pr-linter.yml
kaizencc: yes but as far as i can tell, in order to get an output from a previous step you must depend on that step

kaizencc: so right now if its not a workflow run i skip all individual steps of `download-if-workflow-run` (but do not skip the job)

.github/workflows/pr-linter.yml
```diff
@@ -11,12 +11,55 @@ on:
       - opened
       - synchronize
       - reopened
-  pull_request_review:
+  workflow_run:
+    workflows: [PRLinter Trigger]
```
kaizencc, line 15:
```suggestion
    workflows: [PR Linter Trigger]
```

mrgrain (approved):

PR #26543, opened 2023-07-27
feat(core): Fn.findInMap supports default value

### LINKED ISSUE #26125: core: Fn.findInMap Should Support Default Value
### Describe the feature

Cloudformation recently [added support](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-findinmap-enhancements.html) for a default value in the `FindInMap` intrinsic function. AWS CDK should also support a default value.

### Use Case

We use the CDK to synthesize cloudformation templates for deployment in multiple environments. Thus we are not able to use language features to replicate this functionality. We need to be able to use a default value when a parameter value does not exist in a mapping.

### Proposed Solution

Add a `Default` parameter to the [find_in_map method](https://docs.aws.amazon.com/cdk/api/v1/python/aws_cdk.core/Fn.html#aws_cdk.core.Fn.find_in_map) that would be used if no match is found in the mapping based on the provided 1st/2nd level keys.

### Other Information

_No response_

### Acknowledgements

### CDK version used

latest

AWS Linux 2023, Ubuntu 22.04

scanlonp:
Cloudformation recently [added support](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-findinmap-enhancements.html) for a default value in the FindInMap intrinsic function. This adds that support.

Closes #26125.

----


mrgrain (changes requested):
This one should have an integration test!

Also if you read the announcement carefully, you'll notice it mentions the Language Extension Transform. What do you make of that in regards to this PR?

packages/aws-cdk-lib/core/lib/cfn-mapping.ts
```diff
@@ -148,6 +148,9 @@ class CfnMappingEmbedder implements IResolvable {
     private readonly defaultValue?: string) { }
 
   public resolve(context: IResolveContext): string {
+    if (this.defaultValue !== undefined) {
+      Stack.of(context.scope).addTransform('AWS::LanguageExtensions'); // this has a push command, what if its pushed multiple times?
```
mrgrain, line 152: How about how about you write a unit test for it 😃

packages/aws-cdk-lib/core/README.md
```diff
@@ -1058,6 +1058,28 @@ declare const regionTable: CfnMapping;
 regionTable.findInMap(Aws.REGION, 'regionName');
 ```
 
+An optional default value can also be passed to `findInMap`. If either key is not found in the map and the mapping is lazy, `findInMap` will return the default value. If the mapping is not lazy or either key is an unresolved token, the call to `findInMap` will return a token that resolves to `{ "Fn::FindInMap": [ "MapName", "TopLevelKey", "SecondLevelKey", { "DefaultValue": "DefaultValue" } ] }`. Note that the `AWS::LanguageExtentions` transform is added to enable the default value functionality.
```
scanlonp, line 1061: Is the last line necessary?

packages/aws-cdk-lib/core/lib/cfn-fn.ts
```diff
@@ -500,9 +502,31 @@ class FnFindInMap extends FnBase {
...
+    this.mapName = mapName;
+    this.topLevelKey = topLevelKey;
+    this.secondLevelKey = secondLevelKey;
+    this.defaultValue = defaultValue;
+  }
+
+  public resolve(context: IResolveContext): any {
+    if (this.defaultValue !== undefined) {
+      Stack.of(context.scope).addTransform('AWS::LanguageExtensions');
+    }
```
scanlonp, line 524: This could also be located in the `resolve()` function of the `CfnMappingEmbedder` class. It would remove the need for assigning the 4 variables in this class. However, if it is there, using `Fn.findInMap` would not work with the default value. Only `CfnMapping.findInMap` could use the functionality.

packages/aws-cdk-lib/core/lib/cfn-fn.ts
```diff
@@ -233,10 +233,12 @@ export class Fn {
   /**
    * The intrinsic function ``Fn::FindInMap`` returns the value corresponding to
    * keys in a two-level map that is declared in the Mappings section.
+   * Prefer to use `CfnMapping.findInMap`.
+   * Warning: do not use with lazy maps.
```
scanlonp, line 237: Is this how we like to convey warnings / best practices?

packages/aws-cdk-lib/core/lib/cfn-fn.ts
```diff
@@ -500,9 +502,31 @@ class FnFindInMap extends FnBase {
    * @param mapName The logical name of a mapping declared in the Mappings section that contains the keys and values.
    * @param topLevelKey The top-level key name. Its value is a list of key-value pairs.
    * @param secondLevelKey The second-level key name, which is set to one of the keys from the list assigned to TopLevelKey.
+   * @param defaultValue The value of the default value returned if either the key is not found in the map
```
scanlonp, line 505: Does this need to indicate that the parameter is optional in some way?

scanlonp (commented):
I had a couple of lingering questions about comment guidelines and where code should live. I feel confident with the PR as is, but wanted to get feedback on those points in particular.

packages/aws-cdk-lib/core/lib/cfn-fn.ts
rix0rrr: 👍 this is the better location

packages/aws-cdk-lib/core/lib/cfn-fn.ts
```diff
@@ -500,9 +502,31 @@ class FnFindInMap extends FnBase {
...
+  public resolve(context: IResolveContext): any {
+    if (this.defaultValue !== undefined) {
+      Stack.of(context.scope).addTransform('AWS::LanguageExtensions');
+    }
+    return { 'Fn::FindInMap': [this.mapName, this.topLevelKey, this.secondLevelKey, this.defaultValue !== undefined ? { DefaultValue: this.defaultValue } : undefined] };
+  }
+
+  public toString() {
+    return Token.asString(this);
   }
```
rix0rrr, line 530: 💅 If you don't add this function, don't you get the inherited `toString()` function which should do the same thing?

packages/aws-cdk-lib/core/lib/cfn-mapping.ts
```diff
@@ -63,30 +63,41 @@ export class CfnMapping extends CfnRefElement {
   }
 
   /**
-   * @returns A reference to a value in the map based on the two keys.
+   * @returns A reference to a value in the map based on the two keys. If mapping is lazy, the value from the map or default value is returned instead of the reference.
```
rix0rrr, line 66: I'm pretty sure "lazy" means: only render the actual mapping to the template if there is at least one value that actually needs to be looked up from it at deploy time.

Conversely: if no value is ever read from the map, or all values can be resolved at synth time, then there is no need to render the map.

The line of documentation you added here doesn't help me understand that :).

packages/aws-cdk-lib/core/lib/cfn-mapping.ts
```diff
@@ -63,30 +63,41 @@ export class CfnMapping extends CfnRefElement {
...
    */
-  public findInMap(key1: string, key2: string): string {
+  public findInMap(key1: string, key2: string, defaultValue?: string): string {
```
rix0rrr, line 68: Ohmigod. Can this method be simplified if we reorganize it?

rix0rrr (approved):
Nice! I've got some detail notes, but conditionally approved. Do with my comments as you will and feel free to remove the `pr/do-not-merge` label yourself when you are feel you have addressed them to the degree that you want to 😉.

One final thing: in the title, you're describing the feature using the Python casing `Fn.find_in_map`. In TypeScript the function is called `Fn.findInMap`, and we usually use that casing to describe changes.

packages/aws-cdk-lib/core/lib/cfn-mapping.ts
scanlonp: Felt that it wasn't accurate to say it returns a reference is sometimes it just passes a value. Also could help debug if the short circuit was not documented. Can add a line about the mapping behavior too

packages/aws-cdk-lib/core/lib/cfn-mapping.ts
scanlonp: Would be glad to, but not seeing a clear way. Think we ship it and someone enterprising can tidy it up later.

packages/aws-cdk-lib/core/README.md
kaizencc, line 1061:
```suggestion
An optional default value can also be passed to `findInMap`. If either key is not found in the map and the mapping is lazy, `findInMap` will return the default value.
If the mapping is not lazy or either key is an unresolved token, the call to `findInMap` will return a token that resolves to
`{ "Fn::FindInMap": [ "MapName", "TopLevelKey", "SecondLevelKey", { "DefaultValue": "DefaultValue" } ] }`.
Note that the `AWS::LanguageExtentions` transform is added to enable the default value functionality.
```

packages/aws-cdk-lib/core/lib/cfn-fn.ts
kaizencc: I think you should explain succinctly why you should not use with lazy maps

packages/aws-cdk-lib/core/lib/cfn-fn.ts
kaizencc: no i dont think so

kaizencc (approved):
Have a few minor nits of my own lmao. Feel free to remove `do-not-merge` when you address the one comment

packages/aws-cdk-lib/core/README.md
scanlonp: done

PR #26546, opened 2023-07-28
feat(stepfunctions-tasks): specify step functions alias or version ARN in StartExecution

### LINKED ISSUE #26532: aws-stepfunctions-tasks: support revision id in StepFunctionsStartExecution
### Describe the feature

Step Functions now supports Versions and Aliases.

- Learn more: https://aws.amazon.com/about-aws/whats-new/2023/06/aws-step-functions-versions-aliases/
- Developer Guide: https://docs.aws.amazon.com/step-functions/latest/dg/concepts-cd-aliasing-versioning.html
- AWS Compute Blog post: https://aws.amazon.com/blogs/compute/deploying-state-machines-incrementally-with-versions-and-aliases-in-aws-step-functions/

I would like to have the `tasks.StepFunctionsStartExecution` link to a specific state machine revision ID.

### Use Case

The use case is where there is a parent and child nested step function. If a deploy occurs that updates the nested step function, this may break long running executions of the parent. By linking the parent step function to a revision of the child step function, we can prevent many such breakages.

### Proposed Solution

Currently tasks.StepFunctionsStartExecution accepts a `StateMachine`. I would instead like to pass the `stateMachineRevisionId` which was added here: https://github.com/aws/aws-cdk/pull/26443

### Other Information

_No response_

### Acknowledgements

### CDK version used

Latest

Amazon Linux 2

lpizzinidev:
The `stateMachineArn` property allows specifying the ARN of a [state machine alias or version](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-cd-aliasing-versioning.html) into the `StepFunctionsStartExecution` construct.

This allows to start the execution of a specific state machine alias or version.

Closes #26532.

----


packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/stepfunctions/start-execution.ts
```diff
@@ -119,7 +131,7 @@ export class StepFunctionsStartExecution extends sfn.TaskStateBase {
     const policyStatements = [
       new iam.PolicyStatement({
         actions: ['states:StartExecution'],
-        resources: [this.props.stateMachine.stateMachineArn],
+        resources: [this.stateMachineArn],
```
wong-a, line 134: StartExecution only has resource-level permissions on the unqualified state machine ARN. If `this.stateMachineArn` is a version or alias ARN, we need to strip the qualifier `:<versionId>` or `<:aliasName>` at the end

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/stepfunctions/start-execution.ts
```diff
@@ -42,6 +42,15 @@ export interface StepFunctionsStartExecutionProps extends sfn.TaskStateBaseProps
    * @default - false
    */
   readonly associateWithParent?: boolean;
+
+  /**
+   * The Step Functions state machine ARN to start the execution on.
+   * This allows to specify the ARN of a specific state machine version or alias.
```
wong-a, line 48:
```suggestion
   * This allows you to specify a state machine version or alias ARN.
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/stepfunctions/start-execution.ts
```diff
@@ -42,6 +42,15 @@ export interface StepFunctionsStartExecutionProps extends sfn.TaskStateBaseProps
...
+   * If specified, it overrides the ARN of the stateMachine.
+   *
+   * @default - The ARN of stateMachine is used
```
wong-a, line 51: Not sure what the convention is for referencing other properties. Perhaps specify `stateMachine` is from `StepFunctionsStartExecutionProps`?

```suggestion
   * @default - The ARN of StepFunctionsStartExecutionProps.stateMachine is used
```

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/stepfunctions/integ.start-execution-version.js.snapshot/integ-sfn-start-execution-alias.template.json
```diff
@@ -0,0 +1,86 @@
+{
+ "Resources": {
+  "ChildRole1E3E0EF5": {
```
wong-a, line 3: Shouldn't this have some permissions?

wong-a (changes requested):
Thanks for contributing this! A few small comments for permissions and docs. The approach looks good to me

kaizencc (commented):
@lpizzinidev I chnaged this to a feat which means that there should be readme change introducing this option too :)

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/stepfunctions/start-execution.ts
```diff
@@ -42,6 +42,15 @@ export interface StepFunctionsStartExecutionProps extends sfn.TaskStateBaseProps
...
   readonly associateWithParent?: boolean;
+
+  /**
+   * The Step Functions state machine ARN to start the execution on.
+   * This allows you to specify a state machine version or alias ARN.
+   * If specified, it overrides the ARN of the stateMachine.
+   *
+   * @default - The ARN of StepFunctionsStartExecutionProps.stateMachine is used
+   */
+  readonly stateMachineArn?: string;
```
rix0rrr, line 53: Why do this, instead writing:

```ts
new StartExecution({
  stateMachine: StateMachine.fromStateMachineArn(this, 'SM', 'arn:aws:...'),
});
```

Like we do for every other construct?

rix0rrr (changes requested):

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/stepfunctions/start-execution.ts
wong-a: I guess that works from looking at the implementation. However, StateMachineVersion and StateMachineAlias are different constructs. The ARN of an alias (e.g. `arn:aws:....:stateMachine:MyStateMachineName:MyAlias`) is not a state machine ARN

https://github.com/aws/aws-cdk/blob/27b720135e30d58732997a5a5e172daaa9453381/packages/aws-cdk-lib/aws-stepfunctions/lib/state-machine.ts#L166-L168

If that's the preferred approach, it needs to be documented.

indrora:
@lpizzinidev There are outstanding comments that need to be addressed on this PR. Can you address them or should this PR be closed?

lpizzinidev:
@indrora 
I'll need to run some tests to verify that 
```
StateMachine.fromStateMachineArn(..., 'my-arn').stateMachineArn == 'my-arn'
```
holds for ARNs of versions or aliases.
If that is the case, this PR will resolve to a doc update to specify [this](https://github.com/aws/aws-cdk/pull/26546#discussion_r1301546834) behavior.
Otherwise, a code change will be required to allow those ARNs to be passed in some way (probably not in the current way, as [discussed](https://github.com/aws/aws-cdk/pull/26546#discussion_r1310279150)).
I'll try to dig into it this week.

PR #24308, opened 2023-02-23
fix(redshift): tables were dropped on table name change

Rizxcviii:
Upon a refactor of the `UserTablePrivilieges` now using `tableId` to record changes, this fix will essentially allow table names to be changed.

Closes #24246.

BREAKING CHANGE: Further updates of the Redshift table will fail for existing tables, if the table name is changed. Therefore, changing the table name for existing Redshift tables have been disabled.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

Issue #24246, opened 2023-02-21
aws-redshift: table is dropped on table name change


Rizxcviii:
### Describe the bug

When a table name changes, the entire table is dropped (if the removal policy is `DESTROY`). Amazon Redshift allows the table name to be altered, but the table is being dropped, even though the original table construct is still intact (the resource ID does not change).

### Expected Behavior

Table names should change

### Current Behavior

A new redshift table is created, with the old table being dropped. 
Provided a `DESTROY` removal policy, the table, and it's respective data will be deleted.

### Reproduction Steps

```python
redshift.Table(
    self,
    id="ConsistentTableId",
    table_name="old_table_name",
    removal_policy=cdk.RemovalPolicy.DESTROY,
    # other attributes needed for redshift table
)
```

We then change the redshift table name, this will drop the redshift table, and create a new table. 

```python
redshift.Table(
    self,
    id="ConsistentTableId",
    table_name="new_table_name",
    removal_policy=cdk.RemovalPolicy.DESTROY,
    # other attributes needed for redshift table
)
```

### Possible Solution

Persist the redshift table when table names change, given the use of the redshift data API, we could use `ALTER` statements to change the name of the table. Redundant tables that are no longer used in the Stack can still be "removed" by specifying a new id/removing the existing table construct in the code.

### Additional Information/Context

Obviously not all developers would add a `DESTROY` removal policy, but this was not the point of this issue. The point of the issue is the table not persisting through a table name change, but a new table being created. I am simply using a `DESTROY` removal policy for the worst case scenario, as an example.

Not quite sure if this is intended behaviour. However from my personal development experience, I would say I was surprised that tables were being dropped/being replaced in the stack.

The `cdk diff` command shows the following changes
```
[~] Custom::RedshiftDatabaseQuery RedshiftClusterStack/ExampleTable/Resource/Resource ExampleTable808D9241 
 └─ [~] tableName
     └─ [~] .prefix:
         ├─ [-] example_test
         └─ [+] example
```
The above shows to myself that we're just changing the table name, not creating a new construct

### CDK CLI Version

2.65.0

### Framework Version

2.65.0

### Node.js Version

v16.16.0

### OS

Amazon Linux 2

### Language

Python

### Language Version

3.7.16

### Other information

This might be a feature request... If the original behaviour is intended behaviour.

pahud:
I think you are right.

[...]

PR #26558, opened 2023-07-28
fix(redshift): UserTablePriviliges to track changes using table IDs'

Rizxcviii:
To support solving #24246, there needs to first be a refactor of the UserTablePriviliges to instead record the table id. The reason being that new privileges would be granted/revoked on old tables that now have new names, since a change in table name caused an UPDATE action to be triggered. However the issue remains where, since the table has a new name, the grant/revoke action will be called on an invalid table name (i.e. the old table name).
We now use the table id to track tables, therefore preventing UPDATE events to be triggered.

This blocks #24308

This was originally PR #24874, however that had closed. @kaizencc requested changes, that I had added in here.

----


Rizxcviii:
@kaizencc if you could review this please, I took consideration with your previous changes requested on the old PR

Rizxcviii:
Exemption Request: According to @kaizencc this is a fix, rather than a refactor. However, what should be noted is that the integration test is running, just a change to an integ file is not necessary. The reason being that the change is an internal change with an internal interface, not accessible for public use.

packages/aws-cdk-lib/aws-ec2/lib/machine-image/index.ts
kaizencc, line 1: why these changes to ec2? seems irrelevant

packages/@aws-cdk/aws-redshift-alpha/test/database-query-provider/privileges.test.ts
```diff
@@ -147,9 +149,27 @@ describe('update', () => {
     }));
   });
 
-  test('does not replace when privileges change', async () => {
```
kaizencc, line 150: maybe im reading the diff wrong but seems like the old test got replaced. if thats necessary, can you let me know why? if not, i think we should keep the old test and add in your new ones as separate tests.

packages/@aws-cdk/aws-redshift-alpha/test/integ.database.js.snapshot/tree.json
kaizencc, line 1: whats going on with the changes to `integ.database.js.snapshot` without any changes to `integ.database.js`? Think this needs to get reverted (our integ tests are eventually consistent so you may have run the test to see if it was still working and got the diff). But im confused as to why the assets have changed

packages/@aws-cdk/aws-redshift-alpha/test/database-query-provider/privileges.test.ts
```diff
@@ -162,7 +182,46 @@ describe('update', () => {
       Sql: `REVOKE INSERT, SELECT ON ${tableName} FROM ${username}`,
     }));
     expect(mockExecuteStatement).toHaveBeenCalledWith(expect.objectContaining({
-      Sql: `GRANT DROP ON ${newTableName} TO ${username}`,
+      Sql: `GRANT DROP ON ${tableName} TO ${username}`,
     }));
   });
-});
+
+  test('does not replace when table id is changed', async () => {
```
kaizencc, line 189: shouldn't we want replacement when table id changes?

kaizencc (changes requested):
hi @Rizxcviii! A few comments to get me to better understand the purpose of this PR. thanks for sticking with it!

packages/aws-cdk-lib/aws-ec2/lib/machine-image/index.ts
Rizxcviii: It was causing an error upon running a build. For some reason, changing the filename worked.

packages/@aws-cdk/aws-redshift-alpha/test/database-query-provider/privileges.test.ts
Rizxcviii: The old test was removed, as it was more generic. The new tests are more fined tuned, specific to table names, ids and actions. The old test was used whenever the privilege changed. The test itself was probably generic since the function itself was very basic. However, the new functionality is more complex, so more fined tuned tests are required I think.

packages/aws-cdk-lib/aws-ec2/lib/machine-image/index.ts
kaizencc: this shouldn't be the case. lets revert and see what the error is

Rizxcviii: oh it worked, not sure as to why it was causing issues beforehand

packages/@aws-cdk/aws-redshift-alpha/test/database-query-provider/privileges.test.ts
Rizxcviii: When the table id is changed, the privilege will change to append a new one onto the new table. However the `UserTablePrivilege` construct will not need replacing. Therefore, this test description makes sense from that perspective, which is the notation the other tests are using.

Rizxcviii:
@kaizencc could you review this please?

Rizxcviii:
Bump

PR #7453, opened 2020-04-20
fix(dynamodb): grantXxx() does not grant in replication regions

jogold:
### Commit Message
fix(dynamodb): grantXxx() does not grant in replication regions

Add all replication regions when granting permissions.

Closes #7362

### End Commit Message

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #26563, opened 2023-07-28
feat(dynamodb): global table L2

### LINKED ISSUE #16118: (aws-dynamodb): L2 Construct for GlobalTable

colifran:
This PR is for the implementation of the `GlobalTable` L2.

Please reference the following RFC: https://github.com/aws/aws-cdk-rfcs/blob/master/text/0510-dynamodb-global-table.md

Closes #16118 

----

### DIFF
--- a/packages/aws-cdk/lib/util/content-hash.ts
+++ b/packages/aws-cdk/lib/util/content-hash.ts
@@ -3,3 +3,42 @@ import * as crypto from 'crypto';
 export function contentHash(data: string | Buffer | DataView) {
   return crypto.createHash('sha256').update(data).digest('hex');
 }
+
+/**
+ * A stably sorted hash of an arbitrary JS object
+ */
+export function contentHashAny(value: unknown) {
+  const ret = crypto.createHash('sha256');
+  recurse(value);
+  return ret.digest('hex');
+
+  function recurse(x: unknown) {
+    if (typeof x === 'string') {
+      ret.update(x);
+      return;
+    }
+
+    if (Array.isArray(x)) {
+      ret.update('[');
+      for (const e of x) {
+        recurse(e);
+        ret.update('||');
+      }
+      ret.update(']');
+      return;
+    }
+
+    if (x && typeof x === 'object') {
+      ret.update('{');
+      for (const key of Object.keys(x).sort()) {
+        ret.update(key);
+        ret.update(':');
+        recurse((x as any)[key]);
+      }
+      ret.update('}');
+      return;
+    }
+
+    ret.update(`${x}${typeof x}`); // typeof to make sure hash('123') !== hash(123)
+  }
+}
--- a/packages/aws-cdk/lib/util/work-graph-builder.ts
+++ b/packages/aws-cdk/lib/util/work-graph-builder.ts
@@ -1,5 +1,6 @@
 import * as cxapi from '@aws-cdk/cx-api';
 import { AssetManifest, IManifestEntry } from 'cdk-assets';
+import { contentHashAny } from './content-hash';
 import { WorkGraph } from './work-graph';
 import { DeploymentState, AssetBuildNode, WorkNode } from './work-graph-types';
 
@@ -26,7 +27,7 @@ export class WorkGraphBuilder {
     this.graph.addNodes({
       type: 'stack',
       id: `${this.idPrefix}${artifact.id}`,
-      dependencies: new Set(this.getDepIds(onlyStacks(artifact.dependencies))),
+      dependencies: new Set(this.stackArtifactIds(onlyStacks(artifact.dependencies))),
       stack: artifact,
       deploymentState: DeploymentState.PENDING,
       priority: WorkGraphBuilder.PRIORITIES.stack,
@@ -37,27 +38,26 @@ export class WorkGraphBuilder {
    * Oof, see this parameter list
    */
   // eslint-disable-next-line max-len
-  private addAsset(parentStack: cxapi.CloudFormationStackArtifact, assetArtifact: cxapi.AssetManifestArtifact, assetManifest: AssetManifest, asset: IManifestEntry) {
+  private addAsset(parentStack: cxapi.CloudFormationStackArtifact, assetManifestArtifact: cxapi.AssetManifestArtifact, assetManifest: AssetManifest, asset: IManifestEntry) {
     // Just the artifact identifier
     const assetId = asset.id.assetId;
-    // Unique per destination where the artifact needs to go
-    const assetDestinationId = `${asset.id}`;
 
-    const buildId = `${this.idPrefix}${assetId}-build`;
-    const publishNodeId = `${this.idPrefix}${assetDestinationId}-publish`;
+    const buildId = `build-${assetId}-${contentHashAny([assetId, asset.genericSource]).substring(0, 10)}`;
+    const publishId = `publish-${assetId}-${contentHashAny([assetId, asset.genericDestination]).substring(0, 10)}`;
 
     // Build node only gets added once because they are all the same
     if (!this.graph.tryGetNode(buildId)) {
       const node: AssetBuildNode = {
         type: 'asset-build',
         id: buildId,
+        note: assetId,
         dependencies: new Set([
-          ...this.getDepIds(assetArtifact.dependencies),
+          ...this.stackArtifactIds(assetManifestArtifact.dependencies),
           // If we disable prebuild, then assets inherit (stack) dependencies from their parent stack
-          ...!this.prebuildAssets ? this.getDepIds(onlyStacks(parentStack.dependencies)) : [],
+          ...!this.prebuildAssets ? this.stackArtifactIds(onlyStacks(parentStack.dependencies)) : [],
         ]),
-        parentStack,
-        assetManifestArtifact: assetArtifact,
+        parentStack: parentStack,
+        assetManifestArtifact,
         assetManifest,
         asset,
         deploymentState: DeploymentState.PENDING,
@@ -66,16 +66,17 @@ export class WorkGraphBuilder {
       this.graph.addNodes(node);
     }
 
-    const publishNode = this.graph.tryGetNode(publishNodeId);
+    const publishNode = this.graph.tryGetNode(publishId);
     if (!publishNode) {
       this.graph.addNodes({
         type: 'asset-publish',
-        id: publishNodeId,
+        id: publishId,
+        note: `${asset.id}`,
         dependencies: new Set([
           buildId,
         ]),
         parentStack,
-        assetManifestArtifact: assetArtifact,
+        assetManifestArtifact,
         assetManifest,
         asset,
         deploymentState: DeploymentState.PENDING,
@@ -83,7 +84,7 @@ export class WorkGraphBuilder {
       });
     }
 
-    for (const inheritedDep of this.getDepIds(onlyStacks(parentStack.dependencies))) {
+    for (const inheritedDep of this.stackArtifactIds(onlyStacks(parentStack.dependencies))) {
       // The asset publish step also depends on the stacks that the parent depends on.
       // This is purely cosmetic: if we don't do this, the progress printing of asset publishing
       // is going to interfere with the progress bar of the stack deployment. We could remove this
@@ -91,11 +92,11 @@ export class WorkGraphBuilder {
       // Note: this may introduce a cycle if one of the parent's dependencies is another stack that
       // depends on this asset. To workaround this we remove these cycles once all nodes have
       // been added to the graph.
-      this.graph.addDependency(publishNodeId, inheritedDep);
+      this.graph.addDependency(publishId, inheritedDep);
     }
 
     // This will work whether the stack node has been added yet or not
-    this.graph.addDependency(`${this.idPrefix}${parentStack.id}`, publishNodeId);
+    this.graph.addDependency(`${this.idPrefix}${parentStack.id}`, publishId);
   }
 
   public build(artifacts: cxapi.CloudArtifact[]): WorkGraph {
@@ -131,17 +132,15 @@ export class WorkGraphBuilder {
     return this.graph;
   }
 
-  private getDepIds(deps: cxapi.CloudArtifact[]): string[] {
-    const ids = [];
-    for (const artifact of deps) {
-      if (cxapi.AssetManifestArtifact.isAssetManifestArtifact(artifact)) {
-        // Depend on only the publish step. The publish step will depend on the build step on its own.
-        ids.push(`${this.idPrefix}${artifact.id}-publish`);
-      } else {
-        ids.push(`${this.idPrefix}${artifact.id}`);
-      }
+  private stackArtifactIds(deps: cxapi.CloudArtifact[]): string[] {
+    return deps.flatMap((d) => cxapi.CloudFormationStackArtifact.isCloudFormationStackArtifact(d) ? [this.stackArtifactId(d)] : []);
+  }
+
+  private stackArtifactId(artifact: cxapi.CloudArtifact): string {
+    if (!cxapi.CloudFormationStackArtifact.isCloudFormationStackArtifact(artifact)) {
+      throw new Error(`Can only call this on CloudFormationStackArtifact, got: ${artifact.constructor.name}`);
     }
-    return ids;
+    return `${this.idPrefix}${artifact.id}`;
   }
 
   /**
--- a/packages/aws-cdk/lib/util/work-graph-types.ts
+++ b/packages/aws-cdk/lib/util/work-graph-types.ts
@@ -16,6 +16,8 @@ export interface WorkNodeCommon {
   readonly id: string;
   readonly dependencies: Set<string>;
   deploymentState: DeploymentState;
+  /** Some readable information to attach to the id, which may be unreadable */
+  readonly note?: string;
 }
 
 export interface StackNode extends WorkNodeCommon {
--- a/packages/aws-cdk/lib/util/work-graph.ts
+++ b/packages/aws-cdk/lib/util/work-graph.ts
@@ -217,19 +217,16 @@ export class WorkGraph {
     function renderNode(id: string, node: WorkNode): string[] {
       const ret = [];
       if (node.deploymentState === DeploymentState.COMPLETED) {
-        ret.push(`  "${simplifyId(id)}" [style=filled,fillcolor=yellow];`);
+        ret.push(`  ${gv(id, { style: 'filled', fillcolor: 'yellow', comment: node.note })};`);
       } else {
-        ret.push(`  "${simplifyId(id)}";`);
+        ret.push(`  ${gv(id, { comment: node.note })};`);
       }
       for (const dep of node.dependencies) {
-        ret.push(`  "${simplifyId(id)}" -> "${simplifyId(dep)}";`);
+        ret.push(`  ${gv(id)} -> ${gv(dep)};`);
       }
       return ret;
     }
 
-    function simplifyId(id: string) {
-      return id.replace(/([0-9a-f]{6})[0-9a-f]{6,}/g, '$1');
-    }
   }
 
   /**
@@ -391,4 +388,14 @@ function sum(xs: number[]) {
 
 function retainOnly<A>(xs: A[], pred: (x: A) => boolean) {
   xs.splice(0, xs.length, ...xs.filter(pred));
+}
+
+function gv(id: string, attrs?: Record<string, string | undefined>) {
+  const attrString = Object.entries(attrs ?? {}).flatMap(([k, v]) => v !== undefined ? [`${k}="${v}"`] : []).join(',');
+
+  return attrString ? `"${simplifyId(id)}" [${attrString}]` : `"${simplifyId(id)}"`;
+}
+
+function simplifyId(id: string) {
+  return id.replace(/([0-9a-f]{6})[0-9a-f]{6,}/g, '$1');
 }
--- a/packages/aws-cdk/test/work-graph-builder.test.ts
+++ b/packages/aws-cdk/test/work-graph-builder.test.ts
@@ -3,6 +3,8 @@ import * as path from 'path';
 import * as cxschema from '@aws-cdk/cloud-assembly-schema';
 import * as cxapi from '@aws-cdk/cx-api';
 import { CloudAssemblyBuilder } from '@aws-cdk/cx-api';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { expect } from '@jest/globals';
 import { WorkGraph } from '../lib/util/work-graph';
 import { WorkGraphBuilder } from '../lib/util/work-graph-builder';
 import { AssetBuildNode, AssetPublishNode, StackNode, WorkNode } from '../lib/util/work-graph-types';
@@ -16,6 +18,25 @@ afterEach(() => {
   rootBuilder.delete();
 });
 
+function superset<A>(xs: A[]): Set<A> {
+  const ret = new Set(xs);
+  (ret as any).isSuperset = true;
+  return ret;
+}
+
+expect.addEqualityTesters([
+  function(exp: unknown, act: unknown): boolean | undefined {
+    if (exp instanceof Set && isIterable(act)) {
+      if ((exp as any).isSuperset) {
+        const actSet = new Set(act);
+        return Array.from(exp as any).every((x) => actSet.has(x));
+      }
+      return this.equals(Array.from(exp).sort(), Array.from(act).sort());
+    }
+    return undefined;
+  },
+]);
+
 describe('with some stacks and assets', () => {
   let assembly: cxapi.CloudAssembly;
   beforeEach(() => {
@@ -28,34 +49,34 @@ describe('with some stacks and assets', () => {
 
     expect(assertableNode(graph.node('stack2'))).toEqual(expect.objectContaining({
       type: 'stack',
-      dependencies: expect.arrayContaining(['F1:D1-publish']),
-    } as StackNode));
+      dependencies: superset(['publish-F1-add54bdbcb']),
+    } as Partial<StackNode>));
   });
 
   test('asset publishing step depends on asset building step', () => {
     const graph = new WorkGraphBuilder(true).build(assembly.artifacts);
 
-    expect(graph.node('F1:D1-publish')).toEqual(expect.objectContaining({
+    expect(graph.node('publish-F1-add54bdbcb')).toEqual(expect.objectContaining({
       type: 'asset-publish',
-      dependencies: new Set(['F1-build']),
-    } as Partial<AssetPublishNode>));
+      dependencies: superset(['build-F1-a533139934']),
+    } satisfies Partial<AssetPublishNode>));
   });
 
   test('with prebuild off, asset building inherits dependencies from their parent stack', () => {
     const graph = new WorkGraphBuilder(false).build(assembly.artifacts);
 
-    expect(graph.node('F1-build')).toEqual(expect.objectContaining({
+    expect(graph.node('build-F1-a533139934')).toEqual(expect.objectContaining({
       type: 'asset-build',
-      dependencies: new Set(['stack0', 'stack1']),
+      dependencies: superset(['stack0', 'stack1']),
     } as Partial<AssetBuildNode>));
   });
 
   test('with prebuild on, assets only have their own dependencies', () => {
     const graph = new WorkGraphBuilder(true).build(assembly.artifacts);
 
-    expect(graph.node('F1-build')).toEqual(expect.objectContaining({
+    expect(graph.node('build-F1-a533139934')).toEqual(expect.objectContaining({
       type: 'asset-build',
-      dependencies: new Set(['stack0']),
+      dependencies: superset(['stack0']),
     } as Partial<AssetBuildNode>));
   });
 });
@@ -84,13 +105,16 @@ test('can handle nested assemblies', async () => {
 
   let workDone = 0;
   const graph = new WorkGraphBuilder(true).build(assembly.artifacts);
+
   await graph.doParallel(10, {
     deployStack: async () => { workDone += 1; },
     buildAsset: async () => { },
     publishAsset: async () => { workDone += 1; },
   });
 
-  expect(workDone).toEqual(8);
+  // The asset is shared between parent assembly and nested assembly, but the stacks will be deployed
+  // 3 stacks + 1 asset + 3 stacks (1 reused asset)
+  expect(workDone).toEqual(7);
 });
 
 test('dependencies on unselected artifacts are silently ignored', async () => {
@@ -143,8 +167,8 @@ describe('tests that use assets', () => {
     const traversal = await traverseAndRecord(graph);
 
     expect(traversal).toEqual([
-      'work-graph-builder.test.js-build',
-      'work-graph-builder.test.js:D1-publish',
+      expect.stringMatching(/^build-work-graph-builder.test.js-.*$/),
+      expect.stringMatching(/^publish-work-graph-builder.test.js-.*$/),
       'StackA',
       'StackB',
     ]);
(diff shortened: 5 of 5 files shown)


packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,993 @@
...
+   *
+   * @default - a single replica table will be created in the region the global table is
+   * deployed to
+   */
+  readonly replicas?: ReplicaTableProps[];
+
+  /**
+   * Global secondary indexes to define on all replica tables in the global table.
+   *
+   * Note: Global secondary indexes can be configured on a per-replica basis. Tables only
```
rix0rrr, line 276: What does "configured" mean? Entire GSI definition, or just some aspects of it?

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,993 @@
...
+  readonly tableArn?: string;
+  readonly tableName?: string;
+  readonly tableId?: string;
+  readonly tableStreamArn?: string;
+}
+
+/**
+ * The base class for a global table
+ */
+abstract class GlobalTableBase extends Resource implements IGlobalTable {
```
rix0rrr, line 327: Looks like this class ain't doing much. But I suppose this is where a lot of the `ITable` implementation will go in the future, yeah?

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,993 @@
...
+    return globalSecondaryIndexes;
+  }
+
+  private get replicaTables() {
+    const replicaTables: CfnGlobalTable.ReplicaSpecificationProperty[] = [];
+
+    if (!this._replicaTables.has(this.region)) {
+      const replicaTable = this.configureReplicaTable({ region: this.region });
+      replicaTables.push(replicaTable);
+    }
```
rix0rrr, line 506: Can't we make it impossible to define a replica with the current region?

What's the use case for allowing this?

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,993 @@
...
+    return {
+      readCapacityAutoScalingSettings: {
+        minCapacity: readCapacity.minCapacity,
+        maxCapacity: readCapacity.maxCapacity,
+        targetTrackingScalingPolicyConfiguration: {
+          targetValue: readCapacity.targetUtilizationPercent,
+        },
+      },
+    };
+  }
```
rix0rrr, line 731: Same for these, can we not defer to the implementations?

Make sure they return `any` in the declaration, but they can use TypeScript `satisfies` so you get type checking:

```ts
abstract class Capacity {
  public static onDemand(): Capacity {
    return new class extends Capacity {
      public renderCapacity(): any { return undefined; } /* Note return type declaration, is necessary for jsii */
    }();
  }

  public static autoScaled(/* ... */): Capacity {
    return new class extends Capacity {
      public renderCapacity(): any {
       return {
        readCapacityAutoScalingSettings: { minCapacity: /* ... */ },
       } satisifies CfnGlobalTable.ReadProvisionedThroughputSettingsProperty;
    }
  }

  abstract public renderCapacity(): any;
}
```

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,993 @@
...
+  /**
+   * Configure table encryption using a customer managed key.
+   *
+   * Note: This will set the encryption type to CUSTOMER_MANAGED.
+   *
+   * @param tableKey the KMS key used for the replica table in the deployment region
+   * @param replicaKeyArns KMS key ARNS for all replica tables except the replica table in the
+   * deployment region
+   */
+  public static customerManagedKey(tableKey: IKey, replicaKeyArns: { [region: string]: string } = {}) {
```
rix0rrr, line 966: If you think we're ever going to need additional options here, better to stick the replica Arns in a struct.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
colifran: No, this is the default for a Global Table. From the CloudFormation documentation - "The list must contain at least one element, the region where the stack defining the global table is deployed. For example, if you define your table in a stack deployed to us-east-1, you must have an entry in Replicas with the region us-east-1. You cannot remove the replica in the stack region" (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-globaltable.html#cfn-dynamodb-globaltable-replicas)

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
colifran: Good point. Probably better wording would be something like "Read capacity and contributor insights can be configured on a per-replica basis"

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,993 @@
...
+   *
+   * @default RemovalPolicy.RETAIN
+   */
+  readonly removalPolicy?: RemovalPolicy;
+
+  /**
+   * The billing mode and the associated read and write capacity settings for all replica
+   * tables in the global table.
+   *
+   * Note: Read capacity is configurable on a per-replica basis.
```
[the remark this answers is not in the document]
colifran: Correct and good catch. I will update the wording.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
colifran: I think I misunderstood, but you're right. I updated this.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,993 @@
...
+  public abstract readonly tableArn: string;
+  public abstract readonly tableName: string;
+  public abstract readonly tableId?: string;
+  public abstract readonly tableStreamArn?: string;
+}
+
+/**
+ * A global table
+ */
+export class GlobalTable extends GlobalTableBase {
```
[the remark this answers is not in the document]
colifran: Updated!

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,993 @@
...
+    }
+
+    if (this.encryption && this.encryption.type === TableEncryption.CUSTOMER_MANAGED) {
+      const regionInReplicaKeyArns = this.encryption.replicaKeyArns.hasOwnProperty(region);
+      if (!regionInReplicaKeyArns) {
+        throw new Error(`You must specify a KMS key ARN for each replica table when encryption type is ${TableEncryption.CUSTOMER_MANAGED}`);
+      }
+    }
+
+    return { kmsMasterKeyId: this.encryption.replicaKeyArns[region] };
```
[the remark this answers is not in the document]
colifran: Updated this similar to changes made to `Capacity` and `Billing`. Thanks for pointing this out. It makes the implementation a lot cleaner.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
colifran: Correct. ITable implementation has been added.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
```diff
@@ -0,0 +1,488 @@
...
+   * @param actions the set of actions to allow (i.e., 'dynamodb:PutItem', 'dynamodb:GetItem', etc.)
+   */
+  public grant(grantee: IGrantable, ...actions: string[]): Grant {
+    return Grant.addToPrincipal({
+      grantee,
+      actions,
+      resourceArns: [
+        this.tableArn,
+        this.hasIndex ? `${this.tableArn}/index/*` : Aws.NO_VALUE,
+        ...this.replicaArns,
```
rix0rrr, line 82: When we do `grantXxx` on a GlobalTable, do we want to grant to all the replicas at the same time? I was thinking we grant only to the primary replica, just like a normal Table

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
```diff
@@ -0,0 +1,488 @@
...
+        ...this.replicaArns.map(arn => this.hasIndex ? `${arn}/index/*` : Aws.NO_VALUE),
```
rix0rrr, line 83: You don't need `Aws.NO_VALUE` here -- you already know `hasIndex` is either true or false, so you can just splat in the strings, or not:

```suggestion
        ...(this.hasIndex ? this.replicaArns.map(arn => `${arn}/index/*`) : [])
```

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
```diff
@@ -0,0 +1,488 @@
...
+import * as perms from './perms';
+import { Operation, SystemErrorsForOperationsMetricOptions, OperationsMetricOptions, ITable } from './shared';
+import { IMetric, MathExpression, Metric, MetricOptions, MetricProps } from '../../aws-cloudwatch';
+import { Grant, IGrantable } from '../../aws-iam';
+import { IKey } from '../../aws-kms';
+import { Aws, Resource } from '../../core';
+
+export interface IGlobalTable extends ITable {
+  /**
+   * The ID of the Replica Table in the deployment region of the Global Table.
```
rix0rrr, line 11: Wut?

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
```diff
@@ -0,0 +1,831 @@
...
+      public readonly tableStreamArn?: string;
+      public readonly encryptionKey?: IKey;
+
+      protected readonly region: string;
+      protected readonly hasIndex = (attrs.grantIndexPermissions ?? false) ||
+        (attrs.globalIndexes ?? []).length > 0 ||
+        (attrs.localIndexes ?? []).length > 0;
+
+      public constructor(tableArn: string, tableName: string, tableId?: string, tableStreamArn?: string) {
+        super(scope, id);
```
rix0rrr, line 345: Don't forget to pass `envFromArn` to the parent constructor.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
```diff
@@ -0,0 +1,488 @@
...
+        region: this.region,
+        ...props,
+      });
+    }
+
+    return new Metric({
+      account: this.stack.account,
+      ...props,
+    });
+  }
```
rix0rrr, line 487: Looks like the defaults are the same in each case, so this can be simplified?

```suggestion
  private configureMetric(props: MetricProps) {
    return new Metric({
      ...props,
      region: props?.region ?? this.region,
      account: props?.account ?? this.stack.account,
    });
  }
```

(I'm doing `...props` before instead of after, because if someone passes `({ region: undefined })` then the `undefined` will overwrite the value we just defaulted again. [Try it out](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAZwA4BsCGUBCBTYOAJzwAo5UpkAuRAJTwmIBMAeZKImMAcwBpE4ANZg4AdzAA+AJSIA3gChEiJmGRx0eAHTo4PUouXKtJilT5KjhOLQDk12xeUBfaQG4FzhQtCRYCFAxsAEFgKDwickoaekYWdk5ufkEwEXEpWUMVBHVNHT0DS2VrOwcnIxMtM2Ry1w8vBTQsXAJiMjlEACNMIjtuoltEOsagqFDwyI7+vp7B4ZHm-EISA0QSlOYCbjxmAWnEW3659wWQsIjV9fBN4G3drp6ZgaH3RAB6N8RWAFpvxDgKMggA)

rix0rrr (commented):
Starting to look good! I have some questions around permissions, granting on all replicas and all keys seems a little dangerous to me

packages/aws-cdk-lib/aws-dynamodb/lib/capacity.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/**
+ * Options used to configure autoscaled capacity.
+ */
+export interface AutoscaledCapacityOptions {
+  /**
+   * The minimum allowable capacity.
+   */
+  readonly minCapacity: number;
```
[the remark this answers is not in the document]
colifran: I think a better default here would be 1. The console uses 1 as a pre-populated `minCapacity` and 10 as a pre-populated `maxCapacity`. Do you think we should do the same for both?

packages/aws-cdk-lib/aws-dynamodb/lib/capacity.ts
```diff
@@ -0,0 +1,119 @@
...
+   */
+  readonly maxCapacity: number;
+
+  /**
+   * The ratio of consumed capacity units to provisioned capacity units.
+   *
+   * Note: Target utilization percent cannot be less than 20 and cannot be greater
+   * than 90.
+   *
+   * @default 70
```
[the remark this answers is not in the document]
colifran: The console uses 70 as a pre-populated value for target utilization so I assumed it would be a good choice for a default here.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
colifran: I went back and forth with this. This is the current behavior that is used on `Table`. So if you're providing replicas on a `Table` the grants will apply to all replicas and their indexes. I was thinking from the user perspective that might be what is expected since we already do it with a `Table`. My thinking was that you could use `replica` to retrieve the specific replica table in a region and then `grantXxx` there would apply to just the individual replica.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
colifran: I can change the wording here if it's confusing, but this is how all `GlobalTable` attributes work. The attribute be it `TableArn`, `TableStreamArn`, `TableId`, etc. will be for the `Replica` in the region that the stack containing the `GlobalTable`is deployed to. CloudFormation's exact wording is - "The TableId returned is that of the replica in the region the stack is deployed to". Would it be better to follow that verbiage?

packages/aws-cdk-lib/aws-dynamodb/lib/billing.ts
```diff
@@ -0,0 +1,69 @@
...
+        return props.readCapacity._renderReadCapacity();
+      }
+
+      public _renderWriteCapacity() {
+        return props.writeCapacity._renderWriteCapacity();
+      }
+    }) (BillingMode.PROVISIONED);
+  }
+
+  private constructor (public readonly mode: string) {}
```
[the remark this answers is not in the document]
colifran: Yes it should. Updated.

packages/aws-cdk-lib/aws-dynamodb/lib/capacity.ts
```diff
@@ -0,0 +1,119 @@
...
+   */
+  readonly targetUtilizationPercent?: number
+}
+
+/**
+ * Represents the amount of read and write operations supported by a DynamoDB table.
+ */
+export abstract class Capacity {
+  /**
+   * Provisioned throughput capacity is configured with fixed capacity units.
```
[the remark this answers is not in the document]
colifran: Updated.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
colifran: Way simpler! Updated.

packages/aws-cdk-lib/aws-dynamodb/lib/capacity.ts
```diff
@@ -0,0 +1,119 @@
...
+      }
+
+      public _renderWriteCapacity() {
+        return {
+          writeCapacityAutoScalingSettings: this.renderAutoscaledCapacity(),
+        } satisfies CfnGlobalTable.WriteProvisionedThroughputSettingsProperty;
+      }
+
+      private renderAutoscaledCapacity() {
+        if (options.minCapacity > options.maxCapacity) {
```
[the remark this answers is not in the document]
colifran: I've moved this into the constructor.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
colifran: After doing some more thinking on this, I think the approach we should proceed with (as you mentioned) is that `grantXxx` should only apply to the primary replica. This issue (https://github.com/aws/aws-cdk/pull/7453) was what led to `grantXxx` for a `Table` applying to all replicas. However, `Table` offered no functionality allowing a user to get a specific replica from their `Table` and it appears `grantXxx` applying to all replicas was the only solution. With `GlobalTable` we now offer a method named `replica` that will give the user the ability to `grant` on any of their replicas. I've made the appropriate changes for this. Also, this aligns the behavior of `metricXxx` with `grantXxx` which will give a better experience, more consistent experience.

packages/aws-cdk-lib/aws-dynamodb/lib/capacity.ts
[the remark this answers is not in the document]
colifran: Updated with `minCapacity` as an optional property with the default as 1

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
colifran: Thanks for pointing this out. This was removed as part of the update to how `grantXxx` is being handled.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table.ts
colifran: I'm not sure I understand. What is `envFromArn`?

packages/aws-cdk-lib/aws-dynamodb/lib/capacity.ts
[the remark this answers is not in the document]
rix0rrr: Let's not default `maxCapacity`. `minCapacity` has an obvious value, but `max` doesn't.

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
rix0rrr: "The ID of the table in the main deployment region" ?

packages/aws-cdk-lib/aws-dynamodb/adr/index.md
rix0rrr, line 1: ❤️

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
```diff
@@ -0,0 +1,461 @@
...
+   */
+  private combinedGrant(grantee: IGrantable, options: { keyActions?: string[], tableActions?: string[], streamActions?: string[] }) {
+    if (options.keyActions && this.encryptionKey) {
+      this.encryptionKey.grant(grantee, ...options.keyActions);
+    }
+
+    if (options.tableActions) {
+      const resources = [
+        this.tableArn,
+        this.hasIndex ? `${this.tableArn}/index/*` : Aws.NO_VALUE,
```
rix0rrr, line 428: `Aws.NO_VALUE` here still.

rix0rrr (commented):
I like it. Just the README and a couple of integ tests and we should be good!

packages/aws-cdk-lib/aws-dynamodb/lib/global-table-base.ts
colifran: I'll update the wording for all the attributes.

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
 # Amazon DynamoDB Construct Library
 
 
+Here is a minimal deployable DynamoDB `GlobalTable` definition:
```
rix0rrr, line 4: Don't we need a stability header at the top?

I would also like a block quote   style admonition at the top explaining that there are 2 Table constructs, and that `GlobalTable` is the new preferred one, even for single-region tables.

And maybe a link to the original README that explains `Table` (can be copied to a separate file)

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
...
+```
+
+The above `GlobalTable` definition will result in the provisioning of a single table in `us-west-2` with properties that match what was defined on the `GlobalTable`.
+
+Further reading:
+https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GlobalTables.html
+
+## Keys
+
+When a `GlobalTable` is defined, you must define it's schema using the `partitionKey` (required) and `sortKey` (optional) properties.
```
rix0rrr, line 30:
```suggestion
When a `GlobalTable` is defined, you must define its schema using the `partitionKey` (required) and `sortKey` (optional) properties.
```

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
...
+  contributorInsights: true,
+});
+```
+
+Further reading:
+https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/contributorinsights_HowItWorks.html
+
+## Deletion Protection
+
+`deletionProtection` determines if your `GlobalTable` is protected from deletion. When enabled, the `GlobalTable` cannot be deleted by any user or process.
```
rix0rrr, line 55: This should tie into a short discussion about `RemovalPolicy` as well, and how you should delete a table that is deletionProtected.

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
...
+      region: 'us-east-2',
+      kinesisStream: stream2,
+    },
+  ],
+});
+```
+
+## Replicas
+
+A `GlobalTable` can be configured with replica tables. To do this, the `GlobalTable` must be defined in a region non-agnostic `Stack`. Additionally, the main deployment region must not be given as a replica because this is created by default with the `GlobalTable`. The following is a minimal `GlobalTable` definition with replicas defined in `us-east-1` and `us-east-2`:
```
rix0rrr, line 120: The `additionally` confuses me more than it helps.

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
...
+When a `GlobalTable` is defined, you must define it's schema using the `partitionKey` (required) and `sortKey` (optional) properties.
+
+```ts
+const globalTable = new dynamodb.GlobalTable(this, 'GlobalTable', {
+  partitionKey: { name: 'pk', type: dynamodb.AttributeType.STRING },
+  sortKey: { name: 'sk', type: dynamodb.AttributeType.NUMBER },
+});
+```
+
+## Contributor Insights
```
rix0rrr, line 39: In terms of paragraph ordering, I think things like billing mode/capacity/gsis/replicas are all much more interesting than contributor insights and deletion protection, so should come first.

Yes, probably these got added in the past by random contributors just looking to satisfy their "must have edited the README" requirement without putting much thought into anything, because in general developers hate writing documentation. They love having it, but *someone else* should be providing them with it! 

Unfortunately, since you're touching this now, that someone else is you today 😅

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
...
+    { region: 'us-east-2' },
+  ],
+});
+
+const replicaTable = globalTable.replica('us-east-2');
+```
+
+Note: You can create a new `GlobalTable` with as many replicas as needed as long as there is only one replica per region. After table creation you can add or remove replicas, but you can only add or remove a single replica in each update.
+
+## Capacity
```
rix0rrr, line 208: Billing and capacity should be combined into one section, and billing should be introduced first. 

The tree goes:

* OnDemand
* Provisioned
  * Fixed
  * AutoScaled

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
...
+import * as cdk from 'aws-cdk-lib';
+import * as kms from 'aws-cdk-lib/aws-kms';
+
+const app = new cdk.App();
+const stack = new cdk.Stack(app, 'Stack', { env: { region: 'us-west-2' } });
+
+const tableKey = new kms.Key(stack, 'Key');
+const replicaKeyArns = {
+  'us-east-1': 'arn:aws:kms:us-east-1:123456789012:key/g24efbna-az9b-42ro-m3bp-cq249l94fca6',
+  'us-east-2': 'arn:aws:kms:us-east-2:123456789012:key/g24efbna-az9b-42ro-m3bp-cq249l94fca6',
```
rix0rrr, line 322: For realism's sake this should probably be a different ID.

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
...
+});
+
+```
+
+Note: When encryption is configured with customer managed keys, you must have a key already created in each replica region.
+
+Further reading:
+https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-mgmt
+
+## Global Secondary Indexes
```
rix0rrr, line 341: Can we have an intermediary header that just contains and explains indexing?

```
## Indexing

DDB can automatically update indexes for you. There are 2 types of indexes:

* Global indexes: they do XXX
* Local indexes: they do YYY

### Global secondary indexes

To do this ...

### Local secondary indexes

To do that ...

```

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
...
+    },
+    {
+      region: 'us-east-2',
+      tags: [{ key: 'fizz', value: 'buzz' }],
+    },
+  ],
+});
+```
+
+## Importing Existing Global Tables
```
rix0rrr, line 541: We don't say "importing" anymore. Change this into "referencing" (and also below)

rix0rrr: This can go quite far below.

rix0rrr (changes requested):
Sweet! Can you improve the structure of the README a bit while you're in here anyway? 

Other than that, I think we can ship it.

packages/aws-cdk-lib/aws-dynamodb/adr/index.md
```diff
@@ -0,0 +1,28 @@
...
+2. Grants for a Global Table will only apply to the primary Replica Table.
+
+### Decision
+
+After considering the two choices for implementing grants for a Global Table, it was decided that a grant should only apply to the Global Table's primary Replica Table.
+
+### Considerations
+
+We have decided to implement grants for Global Tables such that they only apply to the primary Replica Table for the following reasons:<br>
+1. It is only possible to implement metrics for the primary Replica Table of a Global Table. Implementing grants to apply to all Replica Tables would result in a confusing user experience.
```
mrgrain, line 22: I don't understand this. The first sentence is a fact (which i can accept). The second sentence should explain why and how it is confusing.

And in general why are `metrics` relevant here? It's the first time used in the ADR.

packages/aws-cdk-lib/aws-dynamodb/adr/index.md
```diff
@@ -0,0 +1,28 @@
...
+2. Applying grants to all Replica Tables and all associated customer-managed KMS keys is counter to the principal of least privilege. Permissions may be given for keys and Replica Tables that a user is not intending permissions to be given for.
+3. The Global Table API offers users a replica method that can be used to work with an individual instance of a Replica Table. This can be used to apply grants on a per-replica basis if needed, but this decision should be made by the user.
+
+### Consequences
+
+Some users may expect grants to apply to all Replica Tables (see: https://github.com/aws/aws-cdk/issues/7362). However, unlike the Table L2 that will precede the Global Table L2, the Global Table API offers users the replica method as a way to work with individual Replica Tables. Additionally, the Global Table API design can support the future implementation of a method allowing a user to retrieve a list of all Replica Tables. A user could then iterate over each Replica Table and apply grants to all Replica Tables individually. This design decision supports both use cases while upholding the principle of least privilege.
```
mrgrain, line 28:
```suggestion
Some users may expect grants to apply to all Replica Tables (see: https://github.com/aws/aws-cdk/issues/7362). However, unlike the Table L2 that is replaced by the Global Table L2, the Global Table API offers users the replica method as a way to work with individual Replica Tables. Additionally, the Global Table API design can support the future implementation of a method allowing a user to retrieve a list of all Replica Tables. A user could then iterate over each Replica Table and apply grants to all Replica Tables individually. This design decision supports both use cases while upholding the principle of least privilege.
```

mrgrain (dismissed):
Nice! Just got a minor thing on the ADR.

packages/aws-cdk-lib/aws-dynamodb/README.md
colifran: Thanks. updated.

packages/aws-cdk-lib/aws-dynamodb/README.md
colifran: Removed `additionally`

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
...
+    },
+    {
+      region: 'us-east-2',
+      contributorInsights: false,
+    },
+  ],
+});
+```
+
+You can retrieve a single replica from a `GlobalTable` using the `replica` method:
```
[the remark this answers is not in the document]
colifran: Updated.

packages/aws-cdk-lib/aws-dynamodb/README.md
colifran: Updated.

packages/aws-cdk-lib/aws-dynamodb/README.md
colifran: Updated.

packages/aws-cdk-lib/aws-dynamodb/README.md
colifran: Good call. I've updated the `README` accordingly

packages/aws-cdk-lib/aws-dynamodb/README.md
colifran: Updated.

packages/aws-cdk-lib/aws-dynamodb/README.md
colifran: Happy to be that someone else! I've reordered the paragraphs

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,6 +1,704 @@
...
+  tableClass: dynamodb.TableClass.STANDARD_INFREQUENT_ACCESS,
+});
+```
+
+Further reading:
+https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.TableClasses.html
+
+## Kinesis Stream
+
+A Kinesis Data Stream can be configured on a `GlobalTable` to capture item-level changes. The Kinesis Data Stream configured on a `GlobalTable` will only apply to the table in the primary deployment region and will not be inherited by any replica tables. Replica specific Kinesis Data Streams should be configured on a per-replica basis.
```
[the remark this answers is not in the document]
colifran: Should they? The stream needs to exist in the region that the replica will live in otherwise CloudFormation will throw an error. Doesn't this mean that the Kinesis stream for any replicas needs to already exist in that region and then be referenced in the user's app? As far as I can tell there isn't a way for us to create a region specific stream for the user.

packages/aws-cdk-lib/aws-dynamodb/adr/index.md
colifran: Updated.

packages/aws-cdk-lib/aws-dynamodb/adr/index.md
colifran: Yeah reading this back is confusing. `Metrics` are relevant because both `metrics` and `grants` are defined on `ITable`. As a user it would be confusing to use `metrics` and have them only work on the primary replica table, but then `grants` propagate to all replica tables. I don't think that is a good user experience so I wanted to make sure the implementation for `metrics` and `grants` are aligned. I've updated the wording. Let me know what you think.

packages/aws-cdk-lib/aws-dynamodb/README.md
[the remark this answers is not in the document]
colifran: Unless you're referring to DynamoDB streams in which case yes, they will automatically be added to any replicas. I think this section could follow the same format at `Indexing`:

```
## Streams

### DynamoDB Streams

### Kinesis Streams

```

packages/aws-cdk-lib/aws-dynamodb/README.md
colifran: I've added in content discussing `RemovalPolicy` and I've added content discussing how `RemovalPolicy` and `deletionProtection` behave when both are configured.

packages/aws-cdk-lib/aws-dynamodb/README.md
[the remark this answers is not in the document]
colifran: Stream section has been updated.

packages/aws-cdk-lib/aws-dynamodb/README.md
colifran: Updated.

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -1,240 +1,748 @@
...
+```
+
+The above `GlobalTable` definition will result in the provisioning of a single table in `us-west-2` with properties that match what was defined on the `GlobalTable`.
+
+Further reading:
+https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GlobalTables.html
+
+## Replicas
+
+A `GlobalTable` can be configured with replica tables. To do this, the `GlobalTable` must be defined in a region non-agnostic `Stack`. The main deployment region must not be given as a replica because this is created by default with the `GlobalTable`. The following is a minimal `GlobalTable` definition with `replicas` defined in `us-east-1` and `us-east-2`:
```
rix0rrr, line 33: Minor minor: negatives are annoying to parse. Try to avoid them.

> region non-agnostic Stack

⇒

> Stack with a defined region

rix0rrr: Also, make sure in the prose there will be 3 copies of the table, one in the primary region and 2 replicas.

rix0rrr (approved):
Got the tiiiiniest note, and after that merge at your own leisure!

packages/aws-cdk-lib/aws-dynamodb/README.md
colifran: Thanks. Updated!

PR #26569, opened 2023-07-29
fix(synthetics): updated handler validation

### LINKED ISSUE #26540: synthetics: cdk requires handler to end in .handler when synthetics do not
### Describe the bug

 From the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-synthetics-canary-code.html#cfn-synthetics-canary-code-handler):

> The entry point to use for the source code when running the canary. For canaries that use the syn-python-selenium-1.0 runtime or a syn-nodejs.puppeteer runtime earlier than syn-nodejs.puppeteer-3.4, the handler must be specified as fileName.handler. For syn-python-selenium-1.1, syn-nodejs.puppeteer-3.4, and later runtimes, the handler can be specified as fileName.functionName , or you can specify a folder where canary scripts reside as folder/fileName.functionName .

Currently CDK enforces that the `handler` is named filename.handler for all runtimes, which is too strict. 

### Expected Behavior

CDK enforces what Synthetics enforces

### Current Behavior

CDK enforces legacy rules.

### Reproduction Steps

n/a

### Possible Solution

Removing the synth-time check that enforces this rule.

_No response_

### CDK CLI Version

latest

### Framework Version

_No response_

### Node.js Version

latest

### OS

latest

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

lpizzinidev:
This fix updates handler validation based on Synthetic rules for [Node](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_Canaries_WritingCanary_Nodejs.html#:~:text=and%20files.-,Handler%20name,-Be%20sure%20to) and [Python](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_Canaries_WritingCanary_Python.html#:~:text=Packaging%20your%20canary%20files) runtimes.

Closes #26540.

----


lpizzinidev:
Exemption Request.
All validation cases should be covered by the unit tests.
Let me know if integration tests are needed.

packages/@aws-cdk/aws-synthetics-alpha/lib/code.ts
```diff
@@ -164,12 +164,7 @@ export class InlineCode extends Code {
     }
   }
 
-  public bind(_scope: Construct, handler: string, _family: RuntimeFamily): CodeConfig {
-
-    if (handler !== 'index.handler') {
-      throw new Error(`The handler for inline code must be "index.handler" (got "${handler}")`);
-    }
```
kaizencc, line 171: I'm concerned about removing this. Have you tested this to make sure that it works? IIRC lambda will bundle inline code into an `index.js|py` file, which means that it should at least always start with `index.`

packages/@aws-cdk/aws-synthetics-alpha/lib/canary.ts
```diff
@@ -435,6 +420,31 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
     };
   }
 
+  private validateHandler(handler: string, runtime: Runtime) {
```
kaizencc, line 423: link to the cfn docs that describe what a handler must look like?

packages/@aws-cdk/aws-synthetics-alpha/lib/canary.ts
```diff
@@ -435,6 +420,31 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
+    if (
+      runtime === Runtime.SYNTHETICS_PYTHON_SELENIUM_1_0 ||
+      runtime === Runtime.SYNTHETICS_NODEJS_PUPPETEER_3_0 ||
+      runtime === Runtime.SYNTHETICS_NODEJS_PUPPETEER_3_1 ||
+      runtime === Runtime.SYNTHETICS_NODEJS_PUPPETEER_3_2 ||
+      runtime === Runtime.SYNTHETICS_NODEJS_PUPPETEER_3_3
+    ) {
```
kaizencc, line 430:
```suggestion
    const oldRuntimes = [/*put those runtimes here */];
    if (oldRuntimes.includes(runtime)) {
```

packages/@aws-cdk/aws-synthetics-alpha/lib/canary.ts
```diff
@@ -435,6 +420,31 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
+      if (!handler.match(/^[0-9A-Za-z_\\-]+\.handler*$/)) {
+        throw new Error(`Canary Handler must be specified as \'fileName.handler\' for legacy runtimes, received ${handler}`);
+      }
+    } else if (runtime === Runtime.SYNTHETICS_NODEJS_PUPPETEER_3_4) {
+      if (!handler.match(/^[0-9A-Za-z_\\-]+\.[A-Za-z_][A-Za-z0-9_]*$/)) {
+        throw new Error(`Canary Handler must be specified either as \'fileName.handler\' or \'fileName.functionName\' for the \'syn-nodejs-puppeteer-3.4'\ runtime, received ${handler}`);
```
kaizencc, line 436: where did you get the info that 3.4 has it's own specific rule? From the cfn docs it seems this isn't the case:

> For syn-python-selenium-1.1, syn-nodejs.puppeteer-3.4, and later runtimes, the handler can be specified as fileName.functionName , or you can specify a folder where canary scripts reside as folder/fileName.functionName .

packages/@aws-cdk/aws-synthetics-alpha/test/canary.test.ts
```diff
@@ -703,3 +703,81 @@ testDeprecated('Role policy generated as expected', () => {
...
+        test: synthetics.Test.custom({
+          handler: 'folder/fileName.functionName',
+          code: synthetics.Code.fromInline('/* Synthetics handler code'),
+        }),
+        runtime: synthetics.Runtime.SYNTHETICS_NODEJS_PUPPETEER_3_4,
+      });
+    }).toThrow(/Canary Handler must be specified either as 'fileName.handler' or 'fileName.functionName' for the 'syn-nodejs-puppeteer-3.4' runtime/);
+  });
+
+  testDeprecated('recent runtimes', () => {
```
kaizencc, line 758: can we also unit test that `fileName.functionName` and `folder/fileName.functionName` passes the regex for recent runtimes

kaizencc (changes requested):
Thanks for taking this on @lpizzinidev. One more thing: I think we should add a basic integ test that tests the `folder/fileName.functionName` scenario with a recent runtime just to make sure that the docs aren't wrong :) because sometimes they are.

PR #26575, opened 2023-07-30
feat(scheduler): base target methods and lambda invoke target

filletofish:
This PR contains implementation for Schedule Targets:

1. Creates a separate module for targets
2. Support imported resources, but not cross account, cross region resources as we discussed in RFC. The unit tests should cover 4 cases (target and role within the same stack, target is imported, role is imported, target and role are imported), 
3. I have moved out class `Schedule` from private package to depend on it in `schedule-targets` unit tests. 

----

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3868,7 +3868,7 @@
     expect "^28.0.0"
     pretty-format "^28.0.0"
 
-"@types/jest@^29.5.4":
+"@types/jest@^29.5.3", "@types/jest@^29.5.4":
   version "29.5.4"
   resolved "https://registry.npmjs.org/@types/jest/-/jest-29.5.4.tgz#9d0a16edaa009a71e6a71a999acd582514dab566"
   integrity sha512-PhglGmhWeD46FYOVLt3X7TiWjzwuVGW9wG/4qocPevXMjCmrIc5b6db9WjeGE4QYVpUAWMDv3v0IiBwObY289A==


packages/@aws-cdk/aws-scheduler-alpha/test/group.test.ts
```diff
@@ -101,18 +102,18 @@ describe('Schedule Group', () => {
     const schedule1 = new Schedule(stack, 'MyScheduleDummy1', {
       schedule: expr,
       group: group,
-      target: new targets.LambdaInvoke({
+      target: new targets.LambdaInvoke(func, {
```
filletofish, line 105: Putting target resource before properties. I think it makes more sense than other way around.

packages/@aws-cdk/aws-scheduler-alpha/test/target.test.ts
```diff
@@ -0,0 +1,53 @@
...
+describe('schedule target', () => {
+  let stack: Stack;
+  let func: lambda.IFunction;
+  const expr = ScheduleExpression.at(new Date(Date.UTC(1969, 10, 20, 0, 0, 0)));
+
+  beforeEach(() => {
+    const app = new App();
+    stack = new Stack(app, 'Stack', { env: { region: 'us-east-1', account: '123456789012' } });
+    func = lambda.Function.fromFunctionArn(stack, 'Function', 'arn:aws:lambda:us-east-1:123456789012:function/somefunc');
+  });
```
filletofish, line 17: Created role and stack in the same account. Otherwise `lambda.Function.grantInvoke` fails as it assumes that function is imported (from a different account).

Also, `grantInvoke` method creates permission on Lambda function itself (updates resource policy) if the role is in the different account.

packages/@aws-cdk/aws-scheduler-alpha/test/group.test.ts
Jacco: Agreed :-)

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
```diff
@@ -0,0 +1,231 @@
...
+import { Construct, IConstruct } from 'constructs';
+import { ScheduleTargetInput } from './input';
+import { ISchedule } from './schedule';
+
+export interface IScheduleTarget {
+  bind(_schedule: ISchedule): ScheduleTargetConfig;
+}
+
+export interface ScheduleTargetBaseProps {
+  readonly role?: iam.IRole;
```
Jacco, line 15: All properties need some documentation

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
```diff
@@ -0,0 +1,231 @@
...
+   *
+   * @default 185
+   */
+  readonly retryAttempts?: number;
+}
+
+/**
+ * Properties for an Schedule Target
+ */
+export interface ScheduleTargetConfig {
```
Jacco, line 50: While you removal makes the code a bit easier, the use when specifying paramaters for a specific target becomes a bit harder because codecompletion will show you a lot of options to choose from while only one of the parameter properties should be filled in any specific case.

I personally don't like it. I think it is fine for CDK internal code to be complicated if it enhanced the use by users of the library.

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * Obtain the Role for the EventBridge Schedule
+ *
+ * If a role already exists, it will be returned. This ensures that if multiple
+ * events have the same target, they will share a role.
+ * @internal
+ */
+export function singletonEventRole(scope: IConstruct): iam.IRole {
+  const id = 'SchedulesRole';
```
Jacco, line 121: In the code we prepared this had a lot more beef.

1. You use one role named 'SchedulesRole' for all targets in the current stack. The more targets there are the less "least privilege' this will get. Personally I think one role per target makes more sense then one role per stack.
2. The code we started with had Stack.of(schedule) and then stack.node.tryFindChild(id) in stead of scope.node.tryFindChild. If you find the role for a target it is fine.

Jacco: 3. Why do we not also narrow down of the use of the role to "aws:SourceArn" "aws:SourceAccount"

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
filletofish: Did you maybe confuse it  with `ScheduleTargetBaseProps`? The interface that the customers will use is `ScheduleTargetBaseProps`, while the interface in discussion - `ScheduleTargetConfig` is used as output of `bind` operation. I think the documentation is misleading and I should update it.

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
filletofish: I originally preferred this implementation inspired by `events.target` (https://github.com/aws/aws-cdk/blob/e479bd4353aefa5e48189d2c71f6067489afe141/packages/aws-cdk-lib/aws-events-targets/lib/util.ts#L63-L81) as it seemed simpler to me. However, I agree that the least privilege policy is a good point. Now I understand where the complexity is coming from.

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
Jacco: I see you point now. I guess I just have to play around with it a bit :-)

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
```diff
@@ -0,0 +1,277 @@
...
+function singletonScheduleRole(schedule: ISchedule, targetArn: string): iam.IRole {
+  const stack = Stack.of(schedule);
+  const arn = Token.isUnresolved(targetArn) ? stack.resolve(targetArn).toString() : targetArn;
+  const hash = md5hash(arn).slice(0, 6);
+  const id = 'SchedulerRoleForTarget-' + hash;
+  const existing = stack.node.tryFindChild(id) as iam.IRole;
+
+  const principal = new iam.PrincipalWithConditions(new iam.ServicePrincipal('scheduler.amazonaws.com'), {
+    StringEquals: {
+      'aws:SourceAccount': schedule.env.account,
```
filletofish, line 154: @Jacco I have removed condition for checking sourceArn as it created a circular dependency between schedule -> target -> role -> schedule. 

In your original prototype it was implemented as: 
```
 const scheduleArn = Arn.format({
    service: 'scheduler',
    resourceName: `${schedule.group.groupName}/${schedule.scheduleName}`,
    resource: 'schedule'
  }, Stack.of(schedule));
  const principal = new iam.PrincipalWithConditions(new iam.ServicePrincipal('scheduler.amazonaws.com'), {
    'StringEquals': {
        "aws:SourceArn": scheduleArn,
        "aws:SourceAccount": schedule.env.account,
    }
  })
```

It didn't create a dependency cycle in your code, as now the `scheduleName` is a reference to the CFN resource instead of just a string:
```
    this.scheduleName = this.getResourceNameAttribute(resource.ref);
```

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
```diff
@@ -0,0 +1,277 @@
...
+  // in the account where the schedule is created.
+  if (sameEnvDimension(schedule.env.account, queue.env.account)) {
+    const policyStatementId = `AllowSchedule${Names.nodeUniqueId(schedule.node)}`;
+
+    queue.addToResourcePolicy(new iam.PolicyStatement({
+      sid: policyStatementId,
+      principals: [new iam.ServicePrincipal('events.amazonaws.com')],
+      effect: iam.Effect.ALLOW,
+      actions: ['sqs:SendMessage'],
+      resources: [queue.queueArn],
```
filletofish, line 193: @Jacco  similarly, there was an option to create a condition for `sourceArn` here. But it would create a dependency cycle between target and schedule.

packages/@aws-cdk/aws-scheduler-alpha/test/group.test.ts
```diff
@@ -15,9 +16,18 @@ describe('Schedule Group', () => {
...
+    role = new iam.Role(stack, 'Role', {
+      roleName: 'someRole',
+      assumedBy: new iam.AccountRootPrincipal(),
+    });
+    func = new lambda.Function(stack, 'MyLambda', {
+      code: new lambda.InlineCode('foo'),
+      handler: 'index.handler',
+      runtime: lambda.Runtime.NODEJS_14_X,
+      tracing: lambda.Tracing.PASS_THROUGH,
+    });
```
filletofish, line 30: Had to specify environment here, otherwise schedule creation failed with an error that target and schedule should be in the same region / account. That's because we consider target and schedule to be in different accounts if one of their accounts is unresolved. 

See `sameEnvDimension` implementation. 

Let me know if that's a problem

packages/@aws-cdk/aws-scheduler-alpha/test/input.test.ts
```diff
@@ -12,18 +14,22 @@ describe('schedule target input', () => {
...
-    role = iam.Role.fromRoleArn(stack, 'Role', 'arn:aws:iam::123456789012:role/johndoe');
-    func = lambda.Function.fromFunctionArn(stack, 'Function', 'arn:aws:lambda:us-east-1:123456789012:function/somefunc');
+    const app = new App();
+    stack = new Stack(app);
+    func = new lambda.Function(stack, 'MyLambda', {
+      code: new lambda.InlineCode('foo'),
+      handler: 'index.handler',
+      runtime: lambda.Runtime.NODEJS_14_X,
+      tracing: lambda.Tracing.PASS_THROUGH,
+    });
```
filletofish, line 24: See https://github.com/aws/aws-cdk/pull/26575/files#r1291166569

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
```diff
@@ -0,0 +1,277 @@
...
+export function bindBaseTargetConfig(props: ScheduleTargetBaseProps) {
+  let { deadLetterQueue } = props;
+
+  return {
+    deadLetterConfig: deadLetterQueue ? { arn: deadLetterQueue?.queueArn } : undefined,
+    retryPolicy: renderRetryPolicy(props.maxEventAge, props.retryAttempts),
+  };
+}
+
+const roleCache: {[key: string]: iam.Role} = {};
```
kaizencc, line 136: why a global prop? Can you identify what the problem is that requires a global prop and if there are any other options? We can discuss decision making in this thread.

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
```diff
@@ -0,0 +1,277 @@
...
+
+/**
+ * Obtain the Role for the EventBridge Scheduler event
+ *
+ * If a role already exists, it will be returned. This ensures that if multiple
+ * events have the same target, they will share a role.
+ * @internal
```
kaizencc, line 143: no need for `@internal` since this isn't exported

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
```diff
@@ -0,0 +1,277 @@
...
+   * The templated target type for the Amazon SQS SendMessage API Operation
+   */
+  readonly sqsParameters?: CfnSchedule.SqsParametersProperty;
+}
+
+/**
+ * Bind props to base schedule target config.
+ * @internal
+ */
+export function bindBaseTargetConfig(props: ScheduleTargetBaseProps) {
```
kaizencc, line 127: why are all of these exported functions, instead of belonging to a construct?

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
```diff
@@ -0,0 +1,277 @@
...
+    Annotations.of(schedule).addWarning(`Cannot add a resource policy to your dead letter queue associated with schedule ${schedule.scheduleName} because the queue is in a different account. You must add the resource policy manually to the dead letter queue in account ${queue.env.account}.`);
+  }
+}
+
+/**
+ * Whether two string probably contain the same environment dimension (region or account)
+ *
+ * Used to compare either accounts or regions, and also returns true if both
+ * are unresolved (in which case both are expted to be "current region" or "current account").
+ * @internal
```
kaizencc, line 205: same, no `@internal`

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+    maxAttempts = maximumRetryAttempts;
+  }
+  return {
+    maximumEventAgeInSeconds: maxAge,
+    maximumRetryAttempts: maxAttempts,
+  };
+}
+
+export namespace targets {
```
kaizencc, line 240: i thought we decided on putting these in `aws-scheduler-targets-alpha`?

packages/@aws-cdk/aws-scheduler-alpha/test/input.test.ts
```diff
@@ -75,12 +80,11 @@ describe('schedule target input', () => {
   test('create an input from object with a ref', () => {
     new Schedule(stack, 'MyScheduleDummy', {
       schedule: expr,
-      target: new targets.LambdaInvoke({
-        role,
+      target: new targets.LambdaInvoke(func, {
```
kaizencc, line 83: just confirming these diffs are not breaking changes because `targets.LambdaInvoke` was never exposed.

kaizencc (commented):
A few minor comments mostly to get more information. But my main comment is that I would like to see the README first, as that is usually the entrypoint to the PR as a reviewer.

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
filletofish: Yeah, you are right. I just wanted to get the constructs right for this commit to get feedback. Forgot to mention in the review notes. Will put the lambda targets in the separate package and the readme with the next commit.

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
filletofish: good catch, I do not think we need it. It was used in the prototype code to find the previously created role and update the assumed role policy. We could just get the previously created IAM roles with `stack.node.tryFindChild(id) as iam.Role` and that's it. I will check that in the new commit (and add a unit test).

packages/@aws-cdk/aws-scheduler-alpha/lib/target.ts
filletofish: Moved now to abstract class

packages/@aws-cdk/aws-scheduler-alpha/lib/schedule.ts
```diff
@@ -1,6 +1,100 @@
...
+   *
+   * @deafult - By default a schedule will be associated with the `default` group.
+   */
+  readonly group?: IGroup;
+}
+
+/**
+ * An EventBridge Schedule
+ */
+export class Schedule extends Resource implements ISchedule {
```
filletofish, line 59: @kaizencc I have moved `Schedule` class out of private zone. Do we need to add a note to readme or anything?

packages/@aws-cdk/aws-scheduler-alpha/rosetta/default.ts-fixture
```diff
@@ -6,8 +6,10 @@ import * as iam from 'aws-cdk-lib/aws-iam';
 import * as kms from 'aws-cdk-lib/aws-kms';
 import * as sqs from 'aws-cdk-lib/aws-sqs';
 import * as cloudwatch from 'aws-cdk-lib/aws-cloudwatch';
+// This import will fail as aws-scheduler-alpha does not depend on aws-scheduler-targets-alpha to avoid cyclic dependency 
+// import * as targets from '@aws-cdk/aws-scheduler-targets-alpha';
```
filletofish, line 10: Calling this out explicitly in comments too @kaizencc. Do you have any ideas for how to handle that?

kaizencc: in package.json we can add 

```json
  "jsiiRosetta": {
    "exampleDependencies": {
      "@aws-cdk/aws-schduler-targets-alpha": "^2.92.0",
    }
  },
```

packages/@aws-cdk/aws-scheduler-alpha/README.md
```diff
@@ -156,7 +161,30 @@ const input = ScheduleTargetInput.fromText(text);
...
 
-TODO: Not yet implemented. See section in [L2 Event Bridge Scheduler RFC](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0474-event-bridge-scheduler-l2.md)
+An execution role is an IAM role that EventBridge Scheduler assumes in order to interact with other AWS services on your behalf.
+
+The classes for templated schedule targets automatically create an IAM role with all the minimum necessary
+permissions to interact with the templated target. If you wish you may specify your own IAM role, then the templated targets
+will grant minimal required permissions. For example: for invoking Lambda function target `LambdaInvoke` will grant
+execution IAM role permission to `lambda:InvokeFunction`.
+
+```text
```
kaizencc, line 171: this should be able to be a ts test once we add

```json
"jsiiRosetta": {
    "exampleDependencies": {
      "@aws-cdk/aws-schduler-targets-alpha": "^2.92.0",
    }
  },
```

packages/@aws-cdk/aws-scheduler-alpha/lib/schedule.ts
```diff
@@ -1,6 +1,100 @@
...
+  public readonly group?: IGroup;
```
kaizencc, line 60: docstrings for these public props

kaizencc (commented):
@filletofish I'm ready to approve this, just some minor stuff surrounding docs. 

Next up **has** to be adding integ tests where appropriate. We won't accept any more additions until the integ tests are up and running.

packages/@aws-cdk/aws-scheduler-alpha/rosetta/default.ts-fixture
filletofish: Thanks! That worked.

packages/@aws-cdk/aws-scheduler-alpha/lib/schedule.ts
filletofish: Added now

packages/@aws-cdk/aws-scheduler-targets-alpha/test/target.test.ts
```diff
@@ -0,0 +1,515 @@
...
+            ],
+          },
+        ],
+      },
+      Roles: ['someRole'],
+    });
+  });
+
+  test('throws when lambda function is imported from different account', () => {
+    const importedFunc = lambda.Function.fromFunctionArn(stack, 'ImportedFunction', 'arn:aws:lambda:us-east-1:111122223333:function/somefunc');
```
kaizencc, line 323: TIL as well, but take a look at this file [here](https://github.com/aws/aws-cdk/blob/apigateway-sagemaker-integration/.gitallowed)

tl;dr: cross env tests need to use `234567890123`

kaizencc (dismissed):

kaizencc:
```
@aws-cdk/aws-lambda-python-alpha: ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'OpenSSL 1.0.2k-fips  26 Jan 2017'. See: https://github.com/urllib3/urllib3/issues/2168
```

wtf?

kaizencc:
@filletofish still another error related to this PR it seems

kaizencc (approved):

PR #26580, opened 2023-07-31
feat(synthetics): enable auto delete lambdas via custom resource

### LINKED ISSUE #18448: (aws-synthetics): add cleanup lambda
### Description

Currently, when you delete a synthetic it leaves behind a massive set of detrius. This includes CW resources, buckets, lambdas, and other resources. This is a limitation of the Synthetics API where it will not explicitly delete resources left behind.

### Use Case

To fully cleanup resources upon stack deletion.

### Proposed Solution

Similar to S3, provision an additional lambda that fires upon Canary deletion and removes all the associated Canary resources.

### Other information

_No response_

### Acknowledge

kaizencc:
Synthetics [used](https://aws.amazon.com/about-aws/whats-new/2022/05/amazon-cloudwatch-synthetics-support-canary-resources-deletion/) to have a property `deleteLambdaResourceOnCanaryDeletion` that has since been deprecated and erased from cloudformation docs. Although this property still works today synthetics makes no promises that this is supported in the future.

Here in CDK land, this PR serves as a replacement to the `deleteLambdaResourceOnCanaryDeletion` property (called `enableAutoDeleteLambdas` on the L2 Canary) by implementing a custom resource similar to what we have in S3 and ECR.

**This PR deprecates `enableAutoDeleteLambdas` in favor of `cleanup: cleanup.LAMBDA`, an enum that achieves the same thing but via custom resource**

Closes #18448

----

### DIFF
--- a/packages/@aws-cdk/custom-resource-handlers/test/aws-synthetics-alpha/auto-delete-underlying-resources-handler.test.ts
+++ b/packages/@aws-cdk/custom-resource-handlers/test/aws-synthetics-alpha/auto-delete-underlying-resources-handler.test.ts
@@ -140,7 +140,7 @@ test('deletes lambda when the name changes on update event', async () => {
   mockSyntheticsClient.send.mockReturnValue({
     Canary: {
       Tags: {
-        'aws-cdk:auto-delete-lambda': 'true',
+        'aws-cdk:auto-delete-underlying-resources': 'true',
       },
       Id: id,
     },


packages/@aws-cdk/aws-synthetics-alpha/lib/canary.ts
```diff
@@ -301,12 +307,49 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
       code: this.createCode(props),
       runConfig: this.createRunConfig(props),
       vpcConfig: this.createVpcConfig(props),
-      deleteLambdaResourcesOnCanaryDeletion: props.enableAutoDeleteLambdas,
```
kaizencc, line 304: should we rename `enableAutoDeleteLambdas` to `autoDeleteLambda` (singular)?

kaizencc (commented):
Struggling whether to call this a breaking change. The behavior should stay the same... the implementation will be wildly different. I'm tempted to call it a breaking change just to make sure people's attention are called to this PR.

Also: this PR is missing proper testing

rix0rrr:
> Struggling whether to call this a breaking change. The behavior should stay the same... the implementation will be wildly different. I'm tempted to call it a breaking change just to make sure people's attention are called to this PR.

I think the deprecation of one parameter in favor of another is the right implementation, and makes this not a breaking change.

packages/@aws-cdk/aws-synthetics-alpha/README.md
```diff
@@ -102,27 +102,35 @@ const schedule = synthetics.Schedule.cron({
...
 
-const canary = new synthetics.Canary(stack, 'Canary', {
+```ts
+const canary = new synthetics.Canary(this, 'Canary', {
   test: synthetics.Test.custom({
     handler: 'index.handler',
     code: synthetics.Code.fromInline('/* Synthetics handler code'),
   }),
-  enableAutoDeleteLambdas: true,
+  autoDeleteLambda: true,
```
rix0rrr, line 127: Why did it use to be plural but is singular now? Is this on purpose?

packages/@aws-cdk/aws-synthetics-alpha/README.md
```diff
@@ -102,27 +102,35 @@ const schedule = synthetics.Schedule.cron({
...
   runtime: synthetics.Runtime.SYNTHETICS_NODEJS_PUPPETEER_4_0,
 });
 ```
 
-Synthetic Canaries create additional resources under the hood beyond Lambda functions. Setting `enableAutoDeleteLambdas: true` will take care of
-cleaning up Lambda functions on deletion, but you still have to manually delete other resources like S3 buckets and CloudWatch logs.
+Setting `autoDeleteLambda: true` will take care of cleaning up Lambda functions on deletion,
+but you still have to manually delete other resources like S3 buckets and CloudWatch logs.
```
rix0rrr, line 133: Ufff. Is that a good experience? I don't care so much about the log group, but the S3 bucket seems like a pain?

Alternative solution:

```ts
cleanup: Cleanup.CLEANUP_LAMBDA
```

Which we can extend in the future with `Cleanup.CLEANUP_ALL` ?

packages/@aws-cdk/custom-resource-handlers/lib/aws-synthetics-alpha/auto-delete-lambda-handler/index.ts
```diff
@@ -0,0 +1,77 @@
...
+    // Canary or Lambda doesn't exist. Exiting.
+  }
+}
+
+/**
+ * The canary will only be tagged for deletion if it's being deleted in the same
+ * deployment as this Custom Resource.
+ *
+ * If the Custom Resource is ever deleted before the repository, it must be because
+ * `autoDeleteImages` has been switched to false, in which case the tag would have
```
rix0rrr, line 71: You forgot to uncopy/paste something 😛

packages/@aws-cdk/custom-resource-handlers/lib/aws-synthetics-alpha/auto-delete-lambda-handler/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  const newCanaryName = updateEvent.ResourceProperties?.CanaryName;
+  const canaryNameHasChanged = (newCanaryName && oldCanaryName)
+    && (newCanaryName !== oldCanaryName);
+
+  // If the name of the canary has changed, CloudFormation will delete the canary
+  // and create a new one with the new name. So the old lambda will be isolated
+  // and we should try to delete the old lambdae. */
+  if (canaryNameHasChanged) {
+    return onDelete(oldCanaryName);
+  }
```
rix0rrr, line 36: I think a simpler implementation would be to return a different `PhysicalResourceId` (based off the CanaryName), then CFN will send us a `DELETE` event at the end of the UPDATE operation. (Is this current implementation rollbackable? Do all cleanup handlers suffer from this?)

packages/@aws-cdk/aws-synthetics-alpha/lib/canary.ts
```diff
@@ -195,9 +200,17 @@ export interface CanaryProps {
    * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-synthetics-canary.html#cfn-synthetics-canary-deletelambdaresourcesoncanarydeletion
    *
    * @default false
+   * @deprecated use autoDeleteLambda instead
```
rix0rrr, line 203: Maybe explain why this is in the property body. Something like "this feature has been deprecated by the service team; instead, use `autoDeleteLambda` which will use a Custom Resource to achieve the same effect"

?

packages/@aws-cdk/aws-synthetics-alpha/lib/canary.ts
```diff
@@ -195,9 +200,17 @@ export interface CanaryProps {
...
    */
   readonly enableAutoDeleteLambdas?: boolean;
 
+  /**
+   * Whether or not to delete the lambda resource when the canary is deleted
+   *
```
rix0rrr, line 209: Note that this will use a Custom Resource. People care about that.

rix0rrr (changes requested):
Mostly a set of questions, biggest one being the rollbackability of these cleanup handlers. Consider the following scenario:

* UPDATE target resource (replacement, creates a new resource)
* UPDATE custom resource (old -> new)
* (...stuff happens...)
* ERROR, triggers a rollback
* UPDATE custom resource (new -> old)
* DELETE target resource (deletes the new resource, remembers the existing one)

Won't we have deleted the old Lambda at that point? Since you probably copy/pasted this, it might be that all cleanup handlers suffer from this problem, but let's make sure we fix it here and then also fix it for the other ones after.

packages/@aws-cdk/aws-synthetics-alpha/README.md
kaizencc: Theres only one lambda to be deleted, so I think singular makes more sense

packages/@aws-cdk/aws-synthetics-alpha/lib/canary.ts
```diff
@@ -10,6 +10,11 @@ import { Runtime } from './runtime';
 import { Schedule } from './schedule';
 import { CloudWatchSyntheticsMetrics } from 'aws-cdk-lib/aws-synthetics/lib/synthetics-canned-metrics.generated';
 import { CfnCanary } from 'aws-cdk-lib/aws-synthetics';
+import { CustomResource, CustomResourceProvider, CustomResourceProviderRuntime } from 'aws-cdk-lib/core';
+import * as path from 'path';
+
+const AUTO_DELETE_UNDERLYING_RESOURCES_RESOURCE_TYPE = 'Custom::SyntheticsAutoDeleteUnderlyingResources';
```
kaizencc, line 16: i'm sure you're going to hate how long/awkward this name is but I don't have a better one

packages/@aws-cdk/custom-resource-handlers/README.md
```diff
@@ -4,6 +4,7 @@ This package contains the following custom resource handlers:
 
 - aws-s3/auto-delete-objects-handler
 - aws-ecr/auto-delete-images-handler
+- aws-synthetics-alpha/auto-delete-lambda-handler
```
kaizencc, line 7:
```suggestion
- aws-synthetics-alpha/auto-delete-underlying-resources-handler
```

packages/@aws-cdk/aws-synthetics-alpha/lib/canary.ts
rix0rrr: Fortunately it's private so I don't have to care 😉 .

I might have gone for something with "cleanup" in the name, that is sorta a synonym to "auto delete underlying resources".

packages/@aws-cdk/custom-resource-handlers/lib/aws-synthetics-alpha/auto-delete-underlying-resources-handler/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    case 'Update':
+      return onUpdate(event);
+    case 'Delete':
+      return onDelete(event.ResourceProperties?.CanaryName);
+  }
+};
+
+async function onUpdate(event: AWSLambda.CloudFormationCustomResourceEvent) {
+  const updateEvent = event as AWSLambda.CloudFormationCustomResourceUpdateEvent;
+  const oldCanaryName = updateEvent.OldResourceProperties?.CanaryName;
```
rix0rrr, line 27: This might be correct, but it feels more natural to me to get this from `PhysicalResourceId`.

And then the rest of this method is also quite trivial, because you can always just return `newCanaryName` (if it's the same, that's fine!)

rix0rrr (approved):
Conditionally shipped; I think the `onUpdate` handler can be simplified a bit more.

PR #26592, opened 2023-08-01
chore(codebuild): improve the doc for subnetSelection

pahud:
If `vpc` is specified with `subnetSelection` undefined, according to this:

https://github.com/aws/aws-cdk/blob/d5c64cba9aa8af8e92af4893657d144ef8e4f873/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts#L655-L660

CDK will look for `PRIVATE_WITH_EGRESS`, `PRIVATE_ISOLATED`, and `PUBLIC` in order. If customer does not have `PRIVATE_WITH_EGRESS` subnets, they will need to have vpc endpoints if they need to access AWS services such as AWS Secrets Manager or Amazon ECR.

This PR improves the doc to clarify.

Closes #<issue number here>.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-eks/lib/cluster-resource-handler/cluster.ts
+++ b/packages/aws-cdk-lib/aws-eks/lib/cluster-resource-handler/cluster.ts
@@ -1,9 +1,7 @@
 /* eslint-disable no-console */
 
 // eslint-disable-next-line import/no-extraneous-dependencies
-import { ResourceNotFoundException } from '@aws-sdk/client-eks';
-// eslint-disable-next-line import/no-extraneous-dependencies
-import * as aws from 'aws-sdk';
+import * as EKS from '@aws-sdk/client-eks';
 import { EksClient, ResourceEvent, ResourceHandler } from './common';
 import { compareLoggingProps } from './compareLogging';
 import { IsCompleteResponse, OnEventResponse } from '../../../custom-resources/lib/provider-framework/types';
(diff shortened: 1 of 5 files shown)


packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -616,9 +616,10 @@ export interface CommonProjectProps {
   /**
    * Where to place the network interfaces within the VPC.
    *
-   * Only used if 'vpc' is supplied.
+   * Only used if 'vpc' is supplied. If you have no PRIVATE_WITH_EGRESS subnets in the specified `vpc`, leaving this
+   * undefined will require VPC endpoints to access AWS services.
    *
-   * @default - All private subnets.
+   * @default - private subnets if any else public subnets
```
kaizencc, line 622: this doesn't make sense to me. if any what? if any subnets exist?

pahud: How about "private subnets if available else public subnets"

rix0rrr (changes requested):
See comment above

pahud:
@rix0rrr 

According to [Best practices for VPCs](https://docs.aws.amazon.com/codebuild/latest/userguide/vpc-support.html),

>You need a NAT gateway or NAT instance to use CodeBuild with your VPC so that CodeBuild can reach public endpoints (for example, to run CLI commands when running builds). You cannot use the internet gateway instead of a NAT gateway or a NAT instance because CodeBuild does not support assigning Elastic IP addresses to the network interfaces that it creates, and auto-assigning a public IP address is not supported by Amazon EC2 for any network interfaces created outside of Amazon EC2 instance launches.

I believe if user selects public subnets, codebuild will not be able to access the public endpoints. So the option would be either PRIVATE_WITH_EGRESS or enable vpc endpoints.

I have modified the suggested changed. Let me know if it works for you.

rix0rrr (approved):
Good find on the CodeBuild exception, I didn't know that. Thanks!

PR #26595, opened 2023-08-01
feat(core): Add RetainExceptOnCreate removal policy

### LINKED ISSUE #26570: aws-cdk-lib: add RetainExceptOnCreate to RemovalPolicy enum
### Describe the feature

CloudFormation recently added a new DeletionPolicy `RetainExceptOnCreate`. Please add an equivalent enum value to `RemovalPolicy`.

Note: There is no corresponding UpdateRemovalPolicy for this. Presumably the new enum value should translate to an UpdateRemovalPolicy of `Retain`.

### Use Case

We don't want to retain newly created resources since they are essentially empty. Generally this comes up where the deployment fails for some reason and CloudFormation rolls back. But then we cannot try again without manual intervention since it will try to create the same resoruce again and fail.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.89.0

Alpine 3.18

pahud:
Add the `RetainExceptOnCreate` RemovalPolicy support.

https://aws.amazon.com/about-aws/whats-new/2023/07/aws-cloudformation-deletion-policies-dev-test-cycle/
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html#aws-attribute-deletionpolicy-options

Closes #26570 

----

### DIFF
--- a/packages/aws-cdk-lib/core/lib/removal-policy.ts
+++ b/packages/aws-cdk-lib/core/lib/removal-policy.ts
@@ -48,6 +48,16 @@ export enum RemovalPolicy {
    * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html#aws-attribute-deletionpolicy-options
    */
   SNAPSHOT = 'snapshot',
+
+  /**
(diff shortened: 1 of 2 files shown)


pahud:
Exemption Request

PR #26597, opened 2023-08-02
fix(RecordSet): recordName with a token do not suffix with zone name

### LINKED ISSUE #26572: (module name): (short issue description)
### Describe the bug

`RecordSet`'s `recordName` normalization via `determineFullyQualifiedDomainName` fails to work when passing in a CfnParameter or imported value.

### Expected Behavior

Since user is expected to pass in FQDN and the helper method is intended to suffix it if missed, the default should be to leave the value as is if input can't be determined and assume user supplied the correct FQDN in parameter/imported value.

### Current Behavior

The logic in determineFullyQualifiedDomainName suffixes zone name if it cannot read the parameter/imported value resulting in incorrect value.

### Reproduction Steps

```
determineFullyQualifiedDomainName(
    'dataspray.io',
    {zoneName: 'dataspray.io.'});
=> CORRECT: dataspray.io.
determineFullyQualifiedDomainName(
    CfnParameter.valueAsString()...,
    {zoneName: 'dataspray.io.'});
=> INCORRECT: dataspray.io.dataspray.io
determineFullyQualifiedDomainName(
    Fn.importValue()...,
    {zoneName: 'dataspray.io.'});
=> INCORRECT: dataspray.io.dataspray.io
```

### Possible Solution

Workaround is to suffix with "." to make it a FQDN and bypass suffixing:
```
determineFullyQualifiedDomainName(
    Fn.join('', [Fn.importValue()..., '.'],
    {zoneName: 'dataspray.io.'});
=> CORRECT: dataspray.io
```

_No response_

### CDK CLI Version

2.77.0

### Framework Version

_No response_

### Node.js Version

18

### OS

osx 14

### Language

Typescript

### Language Version

4.9.5

### Other information

_No response_

matusfaro:
Closes #26572.

### Context
`RecordSet` attempts to be smart about the input value of `recordName`. Typically, for a zone `example.com`, a FQDN is required for the `recordName` such as `test.example.com.`. However, RecordSet's `determineFullyQualifiedDomainName` method attempts to fix user input in case they forgot to add the zone name turning `test` into `test.example.com`. This works great in the usual cases, except for parameterized input such as CfnCondition or imported value from another stack. The `determineFullyQualifiedDomainName` thinks the input does not end in the zone name and tacks it on. When `test.example.com.` is passed in via CfnParameter, the resulting record becomes `test.example.com.example.com`.

### Solution

If the `recordName` (a.k.a. `providedName` inside `determineFullyQualifiedDomainName`) is found to contain a token, leave the input as is and do not suffix it with zone name. This is because the input should have been a FQDN in the first place. So we assume the token includes the correct FQDN as required since we are not able to confirm ourselves.

----


matusfaro:
@rix0rrr Friendly ping. Which option do you prefer?

In summary, yes it will break existing customers, but:
- I assume minimal customers impacted; easy fix for them.
- The ones that are relying on it are already breaking the stated documentation
- AFAIK, there is no other way to fix this without breaking existing customers.

PR #26602, opened 2023-08-02
feat(core): new RemovalPolicy.RETAIN_EXCEPT_ON_CREATE to only retain resources that have been successfully created

mrgrain:
Adds support for the RetainExceptOnCreate DeletionPolicy.
When `applyToUpdateReplacePolicy` is set, this uses the 'Retain' UpdateReplacePolicy.

https://aws.amazon.com/about-aws/whats-new/2023/07/aws-cloudformation-deletion-policies-dev-test-cycle/
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html#aws-attribute-deletionpolicy-options

Closes https://github.com/aws/aws-cdk/issues/26570

Replaces #26595

----


mrgrain:
This is documented in code.
Can't integration test this was it would require a sequenced deployment, rollback and then a check.

packages/aws-cdk-lib/core/lib/cfn-resource.ts
```diff
@@ -120,14 +120,22 @@ export class CfnResource extends CfnRefElement {
...
         deletionPolicy = CfnDeletionPolicy.DELETE;
+        updateReplacePolicy = CfnDeletionPolicy.DELETE;
         break;
 
       case RemovalPolicy.RETAIN:
         deletionPolicy = CfnDeletionPolicy.RETAIN;
+        updateReplacePolicy = CfnDeletionPolicy.RETAIN;
+        break;
+
+      case RemovalPolicy.RETAIN_EXISTING:
```
rix0rrr, line 136: How does this compile? Shouldn't it be `RemovalPolicy.RETAIN_EXCEPT_ON_CREATE` as per below?

packages/aws-cdk-lib/core/lib/removal-policy.ts
```diff
@@ -48,6 +48,19 @@ export enum RemovalPolicy {
...
+   * except for the stack operation that initially created the resource.
+   * When `applyToUpdateReplacePolicy` is set, this uses the 'Retain' UpdateReplacePolicy.
+   *
+   * If the stack operation that created the resource is rolled back, it will delete the resource.
+   * For all other stack operations, such as stack deletion, it retains the resource and its contents.
+   * The result is that new, empty, and unused resources are deleted, while in-use resources and their data are retained.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html#aws-attribute-deletionpolicy-options
+   */
+  RETAIN_EXCEPT_ON_CREATE= 'retain-except-on-create',
```
rix0rrr, line 63: All in all I think I'd like `RETAIN_EXISTING` or `RETAIN_AFTER_CREATE` better, but this will do

packages/aws-cdk-lib/core/lib/cfn-resource.ts
mrgrain: It doesn't, the build fails 😅

packages/aws-cdk-lib/core/lib/removal-policy.ts
```diff
@@ -48,6 +48,21 @@ export enum RemovalPolicy {
    * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html#aws-attribute-deletionpolicy-options
    */
   SNAPSHOT = 'snapshot',
+
+  /**
+   * Resource will be retained when they are requested to be deleted
+   * or need to be replaced due to an update request.
+   * Resource are not retained, if the creation is rolled back.
```
madeline-k, line 55: Would this be too verbose? I'm on the fence. I think it's good to clarify that the resource can still be deleted if a delete request comes from something besides CFN. 

```suggestion
   * Resource will be retained when they are requested to be deleted during a stack delete request
   * or need to be replaced due to a stack update request.
   * Resource are not retained, if the creation is rolled back.
```

madeline-k (approved):
Adding my approval, but I've added the do-not-merge label in case others want to weigh in.

PR #26605, opened 2023-08-02
fix: removes the requirement to have a domain name when using SSL certificates

nsmith-amplience:
This PR allows for domain names to be optional when creating a Cloudfront Distribution that has a certificate attached to it. As Cloudfront Distributions allow for domains to be transferred with the `associate-alias` command there are scenarios where having an SSL certificate attached without a domain are valid, and attaching the SSL certificate in the CDK creation phase minimises the amount of stack drift.

----


nsmith-amplience:
Exemption Request

This PR is rolling back a previous change rather than implementing new functionality, it doesn't seem to meet the criteria in `When are Integration Tests` required.

kaizencc (changes requested):
Hi @nsmith-amplience, before we can look at this PR the build needs to succeed. The current issue is:

```
aws-cdk-lib: aws-cloudfront/test/distribution.test.ts:514:1 - error TS1128: Declaration or statement expected.
aws-cdk-lib: 514 });
aws-cdk-lib:     ~
aws-cdk-lib: aws-cloudfront/test/distribution.test.ts:514:2 - error TS1128: Declaration or statement expected.
aws-cdk-lib: 514 });
```

PR #26608, opened 2023-08-02
fix(ecr): error handling compatible with aws sdk v3

tmokmss:
Just fixing this in the same way as #26433

----


tmokmss:
Exemption Request
I think unit testing is enough for this change

MrArnoldPalmer:
Looks like the app-staging-synthesizer-alpha package has a dependent integ snapshot that needs to be updated as well, otherwise I think this looks fine. Thank you for the catch @tmokmss, we made sure we caught all the error.code checks but error.name was available on only some errors in sdk v2 I'm assuming?

tmokmss:
@MrArnoldPalmer 
I misunderstood this would not work, but it seems it does 🤦 

> This change is backward compatible—all the existing error handling code will still work. You can still assert the error by the name value. https://aws.amazon.com/blogs/developer/service-error-handling-modular-aws-sdk-js/

I guess `instanceof` is preferable way to handle errors in sdk v3 though.

MrArnoldPalmer (approved):

PR #26609, opened 2023-08-02
fix(logs): LogRetention Custom Resource does not use Node18

mrgrain:
Migrate the LogRetention Custom Resource to AWS SDK v3.

Part of a fix for #26461 

----


mrgrain:
Added exemptions as existing tests provide coverage.

rix0rrr (approved):

packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts
```diff
@@ -206,19 +174,41 @@ export async function handler(event: AWSLambda.CloudFormationCustomResourceEvent
...
+
+  return async (block: () => Promise<void>) => {
+    do {
+      try {
+        return await block();
+      } catch (error: any) {
+        if (error instanceof Logs.OperationAbortedException || error.name === 'OperationAbortedException') {
+          if (maxRetries > 0) {
+            maxRetries--;
+            await new Promise(resolve => setTimeout(resolve, delay));
```
rix0rrr, line 203: Can we  make this exponential delay?

rix0rrr (changes requested):

packages/aws-cdk-lib/aws-logs/lib/log-retention-provider/index.ts
rix0rrr:
```
delay *= 2;
```

mrgrain: Going with 10s as a max now. This creates this distribution for the default of 10 rounds, for a total of ~20s.
```
[
    60,    70,   90,
   130,   210,  370,
   690,  1330, 2610,
  5170, 10000
]
20730ms
```

rix0rrr (approved):

PR #26621, opened 2023-08-03
docs(custom-resource): add example for AwsApiCall

ymhiroki:
There weren't any examples for `AwsApiCall` in the document other than the Overview.
This commit adds examples to the document of  `AwsApiCall` and it helps users understand the usage.

----


lpizzinidev (approved):

rix0rrr (changes requested):
Hey, thanks for this! 

I was confused about how a user chooses between v2 and v3 on the custom resource, but I see from the README that it's done by registering a special fact. That information is missing here, which might be a bit confusing.

But I think we might change something about this construct anyway, so let's hold off on this change a little until we've settled on how it works.

rix0rrr:
Okay yeah, the v3 documentation is out of date. See the following PR: https://github.com/aws/aws-cdk/pull/26631

ymhiroki:
@rix0rrr Thanks for your comments!
I updated the PR to use only SDKv3. How about it?

rix0rrr:
> I updated the PR to use only SDKv3. How about it?

I meant the other way around. It should all be only SDKv2-style strings.

rix0rrr (dismissed):
I meant the other way around. It should all be only SDKv2-style strings.

ymhiroki:
I see. I fixed it.

rix0rrr (commented):
Thanks!

rix0rrr (approved):

PR #26636, opened 2023-08-04
fix(apigateway): duplicate methodResponses if the same array is reused between addMethod calls

### LINKED ISSUE #26586: apigateway: Multiple StepFunctionsIntegrations with options lead to "Property MethodResponses contains duplicate values"
### Describe the bug

Using the `StepFunctionsIntegration.startExecution()` integration with API Gateway with cors options (see reproduction steps) works, as long as only one method has the integration with options.

If a second method uses the same options you will get the error
`The stack named TestStack failed to deploy: UPDATE_ROLLBACK_COMPLETE: Property MethodResponses contains duplicate values., Property MethodResponses contains duplicate values.` and in the cdk.out it generates multiple `ResponseModels`.

### Expected Behavior

Multiple methods with the `StepFunctionsIntegration.startExecution()` creates methods with individual options.

### Current Behavior

```
❌ Deployment failed: Error: The stack named TestStack failed to deploy: UPDATE_ROLLBACK_COMPLETE: Property MethodResponses contains duplicate values., Property MethodResponses contains duplicate values.
    at FullCloudFormationDeployment.monitorDeployment (/opt/homebrew/Cellar/aws-cdk/2.89.0/libexec/lib/node_modules/aws-cdk/lib/index.js:426:10236)
    at async Object.deployStack2 [as deployStack] (/opt/homebrew/Cellar/aws-cdk/2.89.0/libexec/lib/node_modules/aws-cdk/lib/index.js:429:153307)
    at async /opt/homebrew/Cellar/aws-cdk/2.89.0/libexec/lib/node_modules/aws-cdk/lib/index.js:429:136985

The stack named TestStack failed to deploy: UPDATE_ROLLBACK_COMPLETE: Property MethodResponses contains duplicate values., Property MethodResponses contains duplicate values.
```

### Reproduction Steps

- create a new CDK project `dk init app --language typescript`
- insert the following Code in the Stack
- deploy with `cdk deploy --require-approval never`

```
        const methodOptions = {
            methodResponses: [
                {
                    statusCode: '200',
                    responseParameters: {
                        'method.response.header.Access-Control-Allow-Origin': true,
                    }
                },
            ]
        }

        const integrationOptions = {
            integrationResponses: [
                {
                    responseParameters: {
                        'method.response.header.Access-Control-Allow-Origin': "'*'",
                    },
                    statusCode: '200'
                }
            ]
        }
        const api = new RestApi(this, 'api');
        const machine = new StateMachine(this, 'machine', {
            stateMachineType: StateMachineType.EXPRESS,
            definitionBody: DefinitionBody.fromChainable(new Pass(this, 'Pass'))
        })
        // uncomment the following line and comment out the line after that to make it running - but without the cors options
        // api.root.addMethod('POST', StepFunctionsIntegration.startExecution(machine))
        api.root.addMethod('POST', StepFunctionsIntegration.startExecution(machine, integrationOptions), methodOptions)
        api.root.addMethod('GET', StepFunctionsIntegration.startExecution(machine, integrationOptions), methodOptions)
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.89.0 (build 2ad6683)

### Framework Version

_No response_

### Node.js Version

v20.5.0

### OS

macOS 13.5

### Language

Typescript

### Language Version

TypeScript (5.1.6)

### Other information

_No response_

lpizzinidev:
Adding a new method to an API with `addMethod` and passing `methodOptions.methodResponses` was generating duplicate entries using `StepFunctionsIntegration.startExecution` as integration:

For example:
```
const integ = apigw.StepFunctionsIntegration.startExecution(stateMachine, integrationOptions);
api.root.addMethod('GET', integ, methodOptions);
api.root.addMethod('POST', integ, methodOptions);
```

Would generate (fails on deployment):
```
 "MethodResponses": [
     {
      "ResponseParameters": {
       "method.response.header.Access-Control-Allow-Origin": true
      },
      "StatusCode": "200"
     },
     {
      "ResponseModels": {
       "application/json": "Empty"
      },
      "StatusCode": "200"
     },
     {
      "ResponseModels": {
       "application/json": "Error"
      },
      "StatusCode": "400"
     },
     {
      "ResponseModels": {
       "application/json": "Error"
      },
      "StatusCode": "500"
     },
     {
      "ResponseModels": {
       "application/json": "Empty"
      },
      "StatusCode": "200"
     },
     {
      "ResponseModels": {
       "application/json": "Error"
      },
      "StatusCode": "400"
     },
     {
      "ResponseModels": {
       "application/json": "Error"
      },
      "StatusCode": "500"
     }
    ],
```

With this fix, it will keep only the specified `methodResponses`.

Also, the `integrationResponses` option in `StepFunctionsIntegration.startExecution` was not used by the corresponding `Integration`.
This fix will allow to use the specified option value.

Closes #26586.

----


rix0rrr (commented):
What if `methodOptions.methodResponses` is different between the two calls? Won't ignoring it the second time do something unexpected? 

Wouldn't it be better to disallow the second call, and force you to add both methods in a single call?

lpizzinidev:
@rix0rrr 
What if instead of falling back to `undefined` we default to `METHOD_RESPONSES` in [`renderMethodResponses`](https://github.com/aws/aws-cdk/blob/dabf868ed81235174b59d4990157b777ed23be64/packages/aws-cdk-lib/aws-apigateway/lib/method.ts#L315)?
This would centralize the handling of the variable in the `Method` class and remove the need for populating the array on [binding](https://github.com/aws/aws-cdk/blob/dabf868ed81235174b59d4990157b777ed23be64/packages/aws-cdk-lib/aws-apigateway/lib/integrations/stepfunctions.ts#L160-L162) (which to me looks like the source of the problem).
And [this](https://github.com/aws/aws-cdk/pull/19486) should still work.

rix0rrr:
I didn't understand why this would be happening because `addMethod('GET', ...)` and `addMethod('POST', ...)` should be adding 2 different Methods, each with their own MethodResponses.

So I added more validation and as far as I can tell, we also have duplicate method responses even with a single integration?

I have pushed my changes to this branch so you can see. Am I misunderstanding the API Gateway contract, or something else screwy going on?

rix0rrr:
For example, I'm now seeing this error in a bunch of places:

```
[aws-cdk-aws-apigateway-stepfunctions-startexecution/my-rest-api/Default/GET] 3 methodResponses for code 200: {"statusCode":"200","responseParameters":{"method.response.header.Access-Control-Allow-Origin":true}},
{"statusCode":"200","responseModels":{"application/json":{"modelId":"Empty"}}},
{"statusCode":"200","responseModels":{"application/json":{"modelId":"Empty"}}}
```

Is this expected?

rix0rrr:
Uh oh. It seems like the CloudFormation should have been rejecting this:

```
[
{"statusCode":"200","responseParameters":{"method.response.header.Access-Control-Allow-Origin":true}},
{"statusCode":"200","responseModels":{"application/json":{"modelId":"Empty"}}},
]
```

But didn't. But it does throw an error on duplicate entries in the array, because it's trying to convert them to a set and throws an error if that conversion is lossy.

rix0rrr:
After looking into it a bit more, something entirely different is happening.

In the code example, the `methodOptions` argument is reused for both `StartExecution`s. Then inside the `Method` constructor, we assign the array from the options to become the private member:

https://github.com/aws/aws-cdk/blob/e593229e095b5800a053a0c51903a4f27ccf6228/packages/aws-cdk-lib/aws-apigateway/lib/method.ts#L206

Since arrays are reference values, **this leads to object sharing**! Both the GET and POST method refer to the exact same array, and so the integration being executed twice adds the same elements to the same array, twice.

Had the code been written like this:

```ts
api.root.addMethod('POST', integration, {
  methodResponses: [ ... ],
});
api.root.addMethod('GET', integration, {
  methodResponses: [ ... ],
});
```

Both methods would have used a different `methodResponses` array, and the duplicate adding of identical elements would not have occurred.

------

That's the immediate problem. But there is another problem, that I alluded to before. If there are multiple `methodResponse`s for the same `statusCode`, CloudFormation will apply them one by one and you'll end up with one of them, arbitrarily. CloudFormation only detects problems when the elements in the array are *exactly* the same, which happened to happen here.

So already doing the following:

```ts
api.root.addMethod('Get', StepFunctionsIntegration.startExecution(machine, integrationOptions), {
  methodResponses: [
    statusCode: '200',
    /* anything here */
  ]
});
```

Becomes a problem, because:

* You are adding a `methodResponse` with statusCode = 200
* The integration is also adding a `methodResponse` with statusCode = 200

You now have multiple methodResponses with the same status code, and after deployment your API will end up with one of them, but not both.

So the original use case of passing `methodResponses` in the method to add a header to all responses isn't even supported.

lpizzinidev:
@rix0rrr 

There are 2 problems here:

1. The [`addMethod`](https://github.com/aws/aws-cdk/blob/e593229e095b5800a053a0c51903a4f27ccf6228/packages/aws-cdk-lib/aws-apigateway/lib/resource.ts#L186-L188) function passes the same reference of `options` to successive calls if declared as a variable.
For example, considering this scenario:
```
    const methodOptions = {
      methodResponses: [
        {
          statusCode: '200',
          responseParameters: {
            'method.response.header.Access-Control-Allow-Origin': true,
          },
        },
      ],
    };

    const integ = apigw.StepFunctionsIntegration.startExecution(stateMachine, integrationOptions);
    api.root.addMethod('GET', integ, methodOptions);
    api.root.addMethod('POST', integ, methodOptions);
```
The [`methodResponses`](https://github.com/aws/aws-cdk/blob/e593229e095b5800a053a0c51903a4f27ccf6228/packages/aws-cdk-lib/aws-apigateway/lib/method.ts#L206) of both `Method`s will reference the same array, causing the duplicates that you mentioned [here](https://github.com/aws/aws-cdk/pull/26636#issuecomment-1673496652).

2. [This](https://github.com/aws/aws-cdk/blob/e593229e095b5800a053a0c51903a4f27ccf6228/packages/aws-cdk-lib/aws-apigateway/lib/integrations/stepfunctions.ts#L160-L162) doesn't take into consideration methods passed via properties, basically overriding in any case, causing the duplicates that you mentioned [here](https://github.com/aws/aws-cdk/pull/26636#issuecomment-1673523993)

I pushed a tentative solution that should solve both issues.
Let me know if further changes are needed and if [this](https://github.com/aws/aws-cdk/pull/26636/files#diff-60836377094d8b17cd38042d1b77087976d89fa4e085cabfb320a09344e2740fR246) validation should be kept (I will add unit tests if so).

PS 
I saw [this](https://github.com/aws/aws-cdk/pull/26636#issuecomment-1674470218) after pushing the last changes and writing the comment.😄

rix0rrr:
I'm going to split this into 2 PRs:

* One for the false sharing
* One for the response merging (only applies to the stepfunctions integration, because that's the only integration that calls `addMethodResponse`).

After lunch 😉

rix0rrr (dismissed):

rix0rrr (dismissed):

rix0rrr (approved):

PR #26648, opened 2023-08-06
fix(apigateway): allowedOrigins are incorrectly interpreted as regexes

### LINKED ISSUE #26623: aws-apigateway: Allowed origins for CORS preflight are treated as regular expressions
### Describe the bug

The CDK constructs for API Gateway allow supplying multiple trusted origins to be used in the auto-generated OPTIONS method handlers for CORS preflight responses ([CorsOptions interface](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.CorsOptions.html)).

The CORS specification only allows for a _single_ origin in the response headers, so what CDK does in this case is generate a response template that modifies the response based on the origin that the client is requesting. It checks each of the supplied origins (beyond the first one) to see if it matches the Origin request header, and if it does, it sets the Access-Control-Allow-Origin response header to the origin that was supplied in the request.

However, this check treats the configured origin as a regular expression ([the code in question](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-apigateway/lib/resource.ts#L323)). Therefore, if `https://www.company.com` has been configured as an allowed origin, an attacker could register the domain `https://wwwxcompany.com`. A crafted request from this origin will match the regular expression and the preflight request succeeds.

It's possible that this is intended behavior, however the documentation doesn't mention anything to that effect. It would also be inconsistent as it does not apply to the first specified origin, only additional ones.

### Expected Behavior

Given an API Gateway resource configured with the CORS preflight origins `"https://www.firstorigin.com", "https://www.secondorigin.com"`:

An OPTIONS request to the resource with a request header of `Origin: https://wwwXsecondorigin.com` should lead to a response header similar to `Access-Control-Allow-Origin: https://www.firstorigin.com`, i.e. not allowing the invalid origin from the request.

### Current Behavior

The OPTIONS request leads to a response header of `Access-Control-Allow-Origin: https://wwwXsecondorigin.com`, signaling to the requesting browser that the requested origin from is valid.

### Reproduction Steps

Create a POC stack such as:
```typescript
import * as apigateway from 'aws-cdk-lib/aws-apigateway';

export class CorsPocStack extends cdk.Stack {
    super(scope, id, props);

    new apigateway.RestApi(this, 'api', {
      defaultCorsPreflightOptions: {
        allowOrigins: ['https://www.firstorigin.com', 'https://www.secondorigin.com'],
      }
    })
      .root
      .addResource('hello')
      .addMethod('GET', new apigateway.MockIntegration({}));
  }
}
```

Deploy the stack.

Call the CORS preflight method, e.g. using httpie:
`http OPTIONS https://xxx.execute-api.xxx.amazonaws.com/prod/hello "Origin: https://wwwXsecondorigin.com"`

Examine the response and see that it includes the header: `Access-Control-Allow-Origin: https://wwwXsecondorigin.com`

### Possible Solution

Change `$origin.matches("${o}")` to `$origin == "${o}"`

_No response_

### CDK CLI Version

2.89.0 (build 2ad6683)

### Framework Version

_No response_

### Node.js Version

v20.2.0

### OS

macOS

### Language

Typescript

### Language Version

TypeScript (5.1.6)

### Other information

_No response_

lpizzinidev:
Allowed origins for CORS preflight were treated like regular expressions in the checking condition.

For example, with this spec:
```
new apigateway.RestApi(this, 'api', {
      defaultCorsPreflightOptions: {
        allowOrigins: ['https://www.firstorigin.com', 'https://www.secondorigin.com'],
      }
    })
      .root
      .addResource('hello')
      .addMethod('GET', new apigateway.MockIntegration({}));
```

Calling:
```
http OPTIONS https://xxx.execute-api.xxx.amazonaws.com/prod/hello "Origin: https://wwwXsecondorigin.com"
```

The response header would include the header (allowing the invalid origin from the request):
```
Access-Control-Allow-Origin: https://wwwXsecondorigin.com
```

This fix solves the issue.

Closes #26623.

----


rix0rrr (changes requested):
Good call, but this change will break everyone who is aware that the field accepts a regex and relies on it.

Better to add a new field, maybe called `allowedOriginsLiteral` (or something) and document that the current one accepts regexes.

rix0rrr:
> Note that the first origin in the array is not treated as a regex, only additional ones starting from the second item.

This confused me for a bit, but I now see that the code is modal. It's not between the first and other elements, but it's between a singleton array and an array with multiple elements. Those get treated differently, and in the multiple elements case all elements get treated as regexes.

Given that, I'm inclined to agree with you.

rix0rrr (dismissed):

rix0rrr (approved):

PR #26652, opened 2023-08-07
feat(cdk): `cdk diff --quiet` suppresses progress messages

### LINKED ISSUE #26526: CLI: Add an option for not showing diffs for stacks with no changes
### Describe the feature

I recommend adding an option to the `cdk diff` command, which suppresses the display of

```
my-super-stackname:
There were no differences
```

messages completely.

If this feature is enabled, cdk will not even print the names of those stacks.

### Use Case

I maintain a rather big CDK repository with over 700 stacks. When our developers change  something, the CI/CD pipeline runs `cdk diff` as a first step. The developer and a colleague (4-eyes principle) is then expected to review what consequences their code change would have, in addition to reviewing the code-diff. This is to prevent unexpected side-effects bringing down our production systems.

With the current behavior the diff is very long which makes this process time-consuming, which leads to people not looking through the diff thoroughly.

In Addition, sometimes important changes get lost in the vast amount of text.

### Proposed Solution

A flag should be introduced which allows to enable or disable printing of those "There were no differences" messages.
It should ideally be configurable via CLI option and via `cdk.json` and `~/.cdk.json`.

To remain backwards compatible, the default setting should still print those messages.

### Other Information

A few years back I wrote a wrapper around `cdk` in python3 called `pcdk` where I added support for parallel deployments (something which later was included directly into `cdk`) and one of the features is a command line option called `--supress-unchanged-stacks` which does exactly what I described.

Similar to `cdk` my script collects the output of the `diff` run for a single stack and prints it to the terminal in one go, as soon as the `diff` run is complete. If the flag is set though and the output consists of "There were no differences", the print function is not called and the output is thus never printed to screen.

If this were implemented directly in cdk, it will certainly be a cleaner solution as the code would have access to internal metrics, like the number of actual changes and such.

Unfortunately my companies policy doesn't allow me to share the source of  `pcdk` publicly.

### Acknowledgements

### CDK version used

2.88.0 (build 5d497f9)

Ubuntu 23.04

MatasNed:
Hello maintainers

This is my first PR, so all feedback is welcome. This change adds support for the `--quiet` flag within cdk diff command, for users that would not like the following messages 'Stack <name>' and 'There were no differences' to be silenced for stacks which do not have *any* identified differences. The request was made by @nomike in the #26526, I simply followed the very well detailed description.

The change follows similar structure to existing command and how they are passed, I have also added to the README as per the contributing guidelines.

Before submitting I have.
- Built the packages & linted them
- Ran unit tests for the entire module
- Ran all integration tests in the suite from `@aws-cdk-testing/cli-integ/bin/run-suite -a cli-integ-tests` 
- E2E tested by launching my small stack in account and executing diffs on it from my build.

Closes #26526

----


MatasNed:
`Clarification Request` the aws-cdk CLI does not seem to mention or run snapshot tests as the configuration of the tests are different for the CLI. Reference to [README](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/cli-integ)

MatasNed:
> 
> 	❌ Features must contain a change to an integration test file and the resulting snapshot.
> 	❌ CLI code has changed. A maintainer must run the code through the testing pipeline (git fetch origin pull/26652/head && git push -f origin FETCH_HEAD:test-main-pipeline), then add the 'pr-linter/cli-integ-tested' label when the pipeline succeeds.
> 
> <b>PRs must pass status checks before we can provide a meaningful review.</b>
> 

Exemption Request due to CLI tests not having snapshots as far as I can find

rix0rrr (approved):
Thanks for this! Looks well-done!

PR #26665, opened 2023-08-08
chore: remove DEFAULT_CR_NODE_VERSION regional map 

MrArnoldPalmer:
Removes the map of regions -> available lambda nodejs runtime versions
that previously was used for all custom resources vended in the aws-cdk.

----


rix0rrr:
I guess we maybe don't need this map anymore, but wouldn't it be helpful to have a single place where we write down our default CR version? So that when Node 20 is released, we don't have to hunt and peck through our code to go and replace 18 -> 20 in 5820 different places, but can do it in one place?

MrArnoldPalmer:
Helpful yes but I think the mechanism should be different to allow migrating resources one by one in case of runtime compatibility issues like we saw with 16 -> 18 and the sdks. Not exactly sure what that looks like yet.

rix0rrr (approved):

PR #26675, opened 2023-08-09
feat(stepfunctions-tasks): FastFile mode for SageMaker Training Job

### LINKED ISSUE #26653: (stepfunctions-tasks): InputMode for SageMaker does not support FastFile
### Describe the feature

* [InputMode enum](https://github.com/aws/aws-cdk/blob/v2.90.0/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/base-types.ts#L458)
* [SageMaker API param](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_Channel.html#sagemaker-Type-Channel-InputMode)

### Use Case

The feature can lead to performance gains when your training job works on a subset of the data that lives under the prefix, with little implementation effort.

### Proposed Solution

Add 'FastFile' to the enum

### Other Information

I haven't checked whether the SFN integration with the SageMaker API call supports this or throws an error, but other API fields are fully supported now so I don't see why not.

### Acknowledgements

### CDK version used

2.90.0

Mac OS

tmyoda:
According to the AWS documentation, the TrainingInputMode for a SageMaker Training Job can be one of the following: `Pipe | File | FastFile`

https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_Channel.html#sagemaker-Type-Channel-InputMode

https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_AlgorithmSpecification.html#API_AlgorithmSpecification_Contents

I have just added `FastFile` below to align with the official documentation.
https://github.com/aws/aws-cdk/blob/v2.90.0/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/base-types.ts#L458

Closes #26653.

----


lpizzinidev (changes requested):
Looks good overall.

I think that since we are here we should add validation for the `algorithmName` parameter in the `SageMakerCreateTrainingJob` [constructor](https://github.com/aws/aws-cdk/blob/ec61b09f6f3f49ace109ec150064fb948635eee1/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts#L143)  following [this](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_AlgorithmSpecification.html#:~:text=Required%3A%20Yes-,AlgorithmName,-The%20name%20of) spec, and the related unit tests.

tmyoda:
Thank you for the review.
I will add validation for `algorithmName ` in this PR.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts
```diff
@@ -163,6 +163,14 @@ export class SageMakerCreateTrainingJob extends sfn.TaskStateBase implements iam
       throw new Error('Must define either an algorithm name or training image URI in the algorithm specification');
     }
 
+    // validate the algorithmName if the algorithmName is defined
```
lpizzinidev, line 166: We need to add validation also for the case where [both `algorithmName` and `trainingImage` are specified](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_AlgorithmSpecification.html#:~:text=Note%20that%20the%20AlgorithmName%20parameter%20is%20mutually%20exclusive%20with%20the%20TrainingImage%20parameter.%20If%20you%20specify%20a%20value%20for%20the%20AlgorithmName%20parameter%2C%20you%20can%27t%20specify%20a%20value%20for%20TrainingImage%2C%20and%20vice%20versa.):
```
if (props.algorithmSpecification.algorithmName && props.algorithmSpecification.trainingImage) {
      throw ... 
}

File: packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts [hunk 2]
```diff
@@ -163,6 +163,14 @@ export class SageMakerCreateTrainingJob extends sfn.TaskStateBase implements iam
...
+    if (props.algorithmSpecification.algorithmName) {
```
lpizzinidev, line 167: I think we should create a private `validateAlgorithmName` function to do the validation which:
1. Returns if `algorithmName` is `undefined`
2. Throws if the length of `algorithmName` is [not between 1 and 170](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_AlgorithmSpecification.html#:~:text=Length%20Constraints%3A%20Minimum%20length%20of%201.%20Maximum%20length%20of%20170.)
3. Throws if `algorithmName` doesn't satisfy the regex pattern

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts
```diff
@@ -163,6 +163,14 @@ export class SageMakerCreateTrainingJob extends sfn.TaskStateBase implements iam
...
+      const regex = /^(arn:aws[a-z\-]*:sagemaker:[a-z0-9\-]*:[0-9]{12}:[a-z\-]*\/)?([a-zA-Z0-9]([a-zA-Z0-9-]){0,62})(?<!-)$/;
+      if (!regex.test(props.algorithmSpecification.algorithmName)) {
+        throw new Error(`Value '${props.algorithmSpecification.algorithmName}' at 'algorithmName' must satisfy regular expression pattern: ${regex.source}`);
```
lpizzinidev, line 170: A message like `Expected algorithm name to match pattern ${regex}, but got ${value}` would be more inline with the conventions used in the rest of the project.

lpizzinidev (changes requested):
Thanks for the follow-up work 💪
I left you a couple of comments on things that I think should be fixed

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts
tmyoda: Oh, it's true. I will add it

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts
tmyoda: I just missed 2. Thank you for pointing it out.
I will implement it as you suggested.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts
tmyoda: OK, got it.
I will change the message to `Expected algorithm name to match pattern ${regex}, but got ${value}`

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/base-types.ts
```diff
@@ -464,7 +464,12 @@ export enum InputMode {
   /**
    * File mode.
    */
-  FILE = 'File'
+  FILE = 'File',
+
+  /**
+   * FastFile mode.
+   */
+  FASTFILE = 'FastFile'
```
lpizzinidev, line 472: Should be renamed `FAST_FILE` to follow the naming conventions in the rest of the file.

lpizzinidev (changes requested):
Looks great, thank you 👍
I left you a comment on the enum value name (sorry, I missed that in the previous review).
If you could fix that then this will be good to go for me.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/base-types.ts
tmyoda: OK, will change it

lpizzinidev (approved):

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/sagemaker/create-training-job.test.ts
```diff
@@ -408,3 +408,146 @@ test('Cannot create a SageMaker train task with both algorithm name and image na
...
+    ],
+    outputDataConfig: {
+      s3OutputLocation: tasks.S3Location.fromBucket(s3.Bucket.fromBucketName(stack, 'Bucket', 'mybucket'), 'myoutputpath/'),
+    },
+  }))
+    .toThrowError(/Algorithm name length must be between 1 and 170, but got 171/);
+});
+
+test('Cannot create a SageMaker train task with incorrect algorithmName', () => {
+
```
kaizencc, line 531:
```suggestion
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts
```diff
@@ -324,6 +332,21 @@ export class SageMakerCreateTrainingJob extends sfn.TaskStateBase implements iam
       : {};
   }
 
+  private validateAlgorithmName(algorithmName?: string): void {
```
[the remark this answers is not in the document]
tmyoda: Alright, will create another PR for the algorithm name validation

tmyoda:
I plan to create another PR for algorithm name validation once it's merged

kaizencc (approved):

kaizencc (approved):

PR #26676, opened 2023-08-09
fix(triggers): executed on update even when executeOnHandlerChange is false

### LINKED ISSUE #25939: Triggers: executed on update even `executeOnHandlerChange: false`
### Describe the bug

Even if I set `executeOnHandlerChange: false`, the function gets executed on every deployment. This happens regardless of whether I just changed `executeOnHandlerChange` to `false` or if I first deployed with this flag, and _then_ changed my function.

It doesn't seems like this property is used anywhere within CDK.

### Expected Behavior

The handler is invoked just once - when Stack is originally created. Alternatively, just remove the option so that it doesn't confuse.

### Current Behavior

The handler is executed every time I run `cdk deploy`

### Reproduction Steps

Here's the CDK script:

```typescript
const app = new App();
const mainStack = new Stack(app, 'Stas-HelloWorld', {env});
let scriptDir = 'Lambda-HelloWorld';
new Trigger(mainStack, 'InitDbTriggerForLambda', {
    handler: new Function(mainStack, scriptDir, {
        runtime: Runtime.NODEJS_16_X,
        code: Code.fromAsset(scriptDir),
        handler: 'lambda.handler',
        timeout: Duration.seconds(120),
        logRetention: RetentionDays.ONE_DAY
    }),
    executeAfter: [mainStack],
    executeOnHandlerChange: false
});
```

Originally the function did nothing, just declared the handler. Then I changed it so that it throws an error to be sure that it's actually invoked (this made the deployment fail):

```javascript
 exports.handler = async function(event) {
    throw new Error("ooops");
 }
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.82.0

### Framework Version

_No response_

### Node.js Version

v18.16.0

### OS

MacOS

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

tam0ri:
Currently, trigger is re-executed every time when the handler function code or configuration changes even if executeOnHandlerChange is set to false. This is because executeOnHandlerChange prop is never used in the current implementation. 

This PR solves the issue by referring executeOnHandlerChange prop and skipping function invocation when executeOnHandlerChange is set to false.

Closes #25939

----


tam0ri:
Exemption Request

rix0rrr (approved):

PR #26696, opened 2023-08-10
feat(glue): Job construct does not honor SparkUIProps S3 prefix when granting S3 access

### LINKED ISSUE #19862: glue: Job construct does not honor SparkUIProps S3 prefix when granting S3 access
### Describe the bug

glue.Job accepts a `SparkUIProps` object as an argument. The job then in turn grants some S3 permissions for the `bucket` attribute of this object. Unfortunately, it does not take the `prefix` attribute of this object into account for the grant. See: https://github.com/aws/aws-cdk/blob/9487b39d6da4ab7efd884c3a8379b03e317786b9/packages/%40aws-cdk/aws-glue/lib/job.ts#L726-L732

### Expected Behavior

I expect `glue.Job` to include the `prefix` attribute of the `SparkUIProps` object when granting S3 permissions

### Current Behavior

`glue.Job` grants read-write access to the entire `bucket` attribute of the `SparkUIProps` object passed to it

### Reproduction Steps

```python
from typing import cast

from aws_cdk import aws_glue as glue
from aws_cdk import aws_s3 as s3
from aws_cdk import core
from aws_cdk.assertions import Template, Match

stack = core.Stack()
code_bucket = s3.Bucket(stack, "CodeBucket")
ui_bucket = s3.Bucket(stack, "UIBucket")

spark_ui_prefix = "/foo/bar/baz"

job = glue.Job(stack, "Job",
               spark_ui=glue.SparkUIProps(enabled=True, bucket=ui_bucket, prefix=spark_ui_prefix),
               executable=glue.JobExecutable.python_etl(
                   glue_version=cast(glue.GlueVersion, glue.GlueVersion.V3_0),
                   python_version=glue.PythonVersion.THREE,
                   script=glue.Code.from_bucket(bucket=code_bucket, key="script.py")
               )

               )

template = Template.from_stack(stack)
template.has_resource_properties("AWS::IAM::Policy",
                                 Match.object_like(
                                     {
                                         "PolicyDocument": {
                                             "Statement": [
                                                 {
                                                     "Action": [
                                                         "s3:GetObject*",
                                                         "s3:GetBucket*",
                                                         "s3:List*",
                                                         "s3:DeleteObject*",
                                                         "s3:PutObject*",
                                                         "s3:Abort*"
                                                     ],
                                                     "Effect": "Allow",
                                                     "Resource": [
                                                         {
                                                             "Fn::GetAtt": [
                                                                 "UIBucketB980636D",
                                                                 "Arn"
                                                             ]
                                                         },
                                                         {
                                                             "Fn::Join": [
                                                                 "",
                                                                 [
                                                                     {
                                                                         "Fn::GetAtt": [
                                                                             "UIBucketB980636D",
                                                                             "Arn"
                                                                         ]
                                                                     },
                                                                     f"{spark_ui_prefix}*"
                                                                 ]
                                                             ]
                                                         }
                                                     ]
                                                 },
                                                 Match.any_value(),
                                             ]
                                         }
                                     }
                                 )
                                 )
```

### Possible Solution

Change `bucket.grantReadWrite(role);` to `bucket.grantReadWrite(role, props.prefix);`

_No response_

### CDK CLI Version

1.148.0 (build 69a50f1)

### Framework Version

_No response_

### Node.js Version

v14.17.6

### OS

OSX

### Language

Python

### Language Version

_No response_

### Other information

_No response_

scanlonp:
The `SparkUIProps.prefix` attribute of `glue.Job` now has a consistent format, is validated, and the bucket ReadWrite role is only given permissions to the folder of the bucket specified by the `prefix` if provided.

Adds a suggested format for the prefix: `/foo/bar` and throws and error if the prefix does not start with a `/` or ends without a `/`.

This may be a breaking change for users who configured their prefix in a style different than this enforces. However, I believe it is the correct standardization going forward.

Closes #19862.

----


scanlonp:
I did not include an integ test since Spark did not write to s3 even when enabled (for a trivial script). Non trivial scripts usually failed since we did not have the set up for them, or Spark did not write anything. 

I verified that the set up in the console is correct, but could not get anything to be written by Spark through manual testing. 

Since this is an alpha module (and this issue has gotten very little attention), I think it is fine to wait to add the integ test until a user either says this does not work or provides a script that shows Spark writing logs. There is already an integ test that has Spark enabled, so there is almost no value add for an additional test with the prefix set up, since it is currently functionally the same.

(If we feel strongly that we need one, I have one written and can throw it in!)

rix0rrr (approved):

PR #26718, opened 2023-08-11
feat(apigateway): mutate methodresponses added by integrations

rix0rrr:
The `StepFunctionsIntegration` integration can produce `200`, `400` or `500` status codes.

To make its usage more convenient, it also adds default methodResponses for each of those statuses; if it didn't do this, the user would have to define the 3 method responses every time it uses the integration. However, the problem comes if the user wants to modify the default method responses. This is currently not possible, as `addMethodResponse` will just add to the array, adding multiple entries for the same statusCode, of which CloudFormation will deploy one nondeterministically and the rest is lost.

-----

This PR address this problem by extending the feature set of `addMethodResponse`. It shouldn't currently be called multiple times for the same status code. We add that feature, and define that it merges the new method response fields with any existing method responses that share the same status code. 

This makes it trivial to do things like add headers to existing responses.

Sister PR to #26720.

Relates to #26586 and #26636.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-apigateway/lib/method.ts
+++ b/packages/aws-cdk-lib/aws-apigateway/lib/method.ts
@@ -278,9 +278,23 @@ export class Method extends Resource {
 
   /**
    * Add a method response to this method
+   *
+   * If a method response for the same status code already exists, the `responseModels`
+   * and `responseParameters` maps will be merged.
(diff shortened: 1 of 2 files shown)


rix0rrr:
No one seems to need this change. Closing until we get interest on it.

PR #26725, opened 2023-08-11
feat(cdk): enable hotswap for appsync function and resolver code

onlybakam:
Update the hotswap feature for AppSync resolvers and functions to include the `code` attribute. Currently hotswap only works for mapping templates. With the introduction of JavaScript support in AppSync, developers use the `code` attribute to set their JavaScript code.

----


packages/aws-cdk-lib/aws-appsync/lib/appsync-function.ts
```diff
@@ -161,7 +161,8 @@ export class AppsyncFunction extends Resource implements IAppsyncFunction {
       runtime: props.runtime?.toProperties(),
       codeS3Location: code?.s3Location,
       code: code?.inlineCode,
-      functionVersion: '2018-05-29',
+      // function Version should not be set when Code is provided
```
comcalvi, line 164: I'll give this a more thorough look on Monday, but could explain in this commend why the version should not be set? Does this publish a new version implicitly and we wind up using `$LATEST` if the version is `undefined` here?

onlybakam: `functionVersion` is a setting only used when the function runtime is `VTL` (the AppSync default), when the runtime is set to `APPSYNC_JS` and `code` is specified, `functionVersion` should not be used (it is ignored).

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -3,7 +3,9 @@ import { ISDK } from '../aws-auth';
 import { EvaluateCloudFormationTemplate } from '../evaluate-cloudformation-template';
 
 export async function isHotswappableAppSyncChange(
-  logicalId: string, change: HotswappableChangeCandidate, evaluateCfnTemplate: EvaluateCloudFormationTemplate,
+  logicalId: string,
+  change: HotswappableChangeCandidate,
+  evaluateCfnTemplate: EvaluateCloudFormationTemplate,
```
comcalvi, line 8: Can we keep these on one line? That's the style in the rest of hotswap code

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -47,23 +49,35 @@ export async function isHotswappableAppSyncChange(
...
 
+        // copy the old properties
         const sdkProperties: { [name: string]: any } = {
           ...change.oldValue.Properties,
-          requestMappingTemplate: change.newValue.Properties?.RequestMappingTemplate,
-          responseMappingTemplate: change.newValue.Properties?.ResponseMappingTemplate,
         };
+
+        // if the code is set, make sure the `functionVersion` is not set
+        // otherwhise, this is a mapping template change
```
comcalvi, line 58:
```suggestion
        // otherwise, this is a mapping template change
```

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -47,23 +49,35 @@ export async function isHotswappableAppSyncChange(
...
-            functionId: functionId!,
-          }).promise();
+          const { functionId } = functions?.find((fn) => fn.name === physicalName) ?? {};
+          await sdk
+            .appsync()
+            .updateFunction({
+              ...sdkRequestObject,
+              functionId: functionId!,
+            })
+            .promise();
```
comcalvi, line 80: can we keep the old style here?

comcalvi (changes requested):
This looks pretty good overall, but it needs unit tests. Please add tests that verify:

1. The `requestMappingTemplate` and `responseMappingTemplate` are not passed to the `updateFunction` SDK call when `Code` is specified. 
2. The `code` is not passed to `updateFunction` when either `requestMappingTemplate` / `responseMappingTemplate` are specified
3. The `code` is passed to `updateFunction` when specified.

packages/aws-cdk-lib/aws-appsync/lib/appsync-function.ts
comcalvi: Please add that information to the comment on line 164, to make it clear why this is the way it is

onlybakam: addressed in latest commit

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
onlybakam: addressed in last commit

PR #26756, opened 2023-08-14
fix: aws-sdk still used in EKS custom resources

MrArnoldPalmer:
Removes usage of aws-sdk in eks custom resources. The remaining usage was only type references that appear to be forward compatible but this cleans up the code and makes it possible to remove aws-sdk as a dev dependency to aws-cdk-lib once the rout53 cross account zone delegation handler is updated.

----


MrArnoldPalmer:
Still running integ tests so not ready for review yet, but I need CI to show me where snapshots are still failing etc.

rix0rrr (approved):

PR #23196, opened 2022-12-01
feat(lambda): support for Lambda SnapStart

DeerajTheepshi:
Fixes #23153 

----

### All Submissions:

* [x] Have you followed the guidelines in our [Contributing guide?](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md)

### Adding new Construct Runtime Dependencies:

* [ ] This PR adds new construct runtime dependencies following the process described [here](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md/#adding-construct-runtime-dependencies)

### New Features

* [x] Have you added the new feature to an [integration test](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md)?
	* [x] Did you use `yarn integ` to deploy the infrastructure and generate the snapshot (i.e. `yarn integ` without `--dry-run`)?

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #26761, opened 2023-08-15
feat(lambda): L2 constructs for SnapStart

### LINKED ISSUE #23153: Lambda: Support SnapStart
### Describe the feature

https://techcrunch.com/2022/11/28/aws-makes-lambda-cold-start-latency-a-thing-of-the-past-with-snapstart/

Currently SnapStart is only supported with existing lambdas, and currently there is no mention in the AWS Docs of this feature being supported.

### Use Case

As an internal team to Amazon, we want to use SnapStart to eliminate coldstart on our Java lambdas.

### Proposed Solution

Put a boolean into the Lambda construct to enable SnapStart.

### Other Information

_No response_

### Acknowledgements

### CDK version used

Any

Any

roger-zhangg:
Closes #23153

- Support SnapStart on L2 constructs by setting `snapStart: lambda.SnapStartConfig.ON_PUBLISHED_VERSIONS`.
This value will be converted to `SnapStart: {ApplyOn: 'PublishedVersions',}` in template.
- Supports checking limitation stated in [SnapStart](https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html) except for Provisioned Concurrency which is at version level and can't be checked with function level props.
- This PR is heavily inspired by [PR](https://github.com/aws/aws-cdk/pull/23196) from @DeerajTheepshi and [discussion](https://github.com/aws/aws-cdk/issues/23153#issuecomment-1407779812) in the issue, Thank you
----


roger-zhangg:
added a `handler-snapstart.zip` in integration test for the original `handler.zip` seems corrupted. Can we confirm if this make sense, then I'll replace the handler.zip with this new zip

roger-zhangg:
Got some strange 503 error in codebuild

----
aws-cdk: [2K[1Gaws-cdk: [2m$ cdk-build[22m
cdk-assets: [2K[1Gcdk-assets: [1myarn run v1.22.19[22m
cdk-assets: [2K[1G[2m$ cdk-test[22m
aws-cdk: curl: (22) The requested URL returned error: 503 
aws-cdk: Error: ./generate.sh exited with error code 22
aws-cdk: Build failed.!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-----

Doesn't have any problem with yarn build locally. Can anyone help to take a look?

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -1283,6 +1293,40 @@ Environment variables can be marked for removal when used in Lambda@Edge by sett
...
+    if (!props.snapStart) {
+      return undefined;
+    }
+
+    // SnapStart does not support arm64 architecture, Amazon Elastic File System (Amazon EFS), or ephemeral storage greater than 512 MB.
+    // SnapStart doesn't support provisioned concurrency either, but that's configured at the version level,
+    // so it can't be checked at function set up time
+    // SnapStart supports the Java 11 and Java 17 (java11 and java17) managed runtimes.
+    // See https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html
+    if (props.snapStart != SnapStartConfig.ON_PUBLISHED_VERSIONS) {
```
valerena, line 1307: Instead of checking the entire object, you should check if `props.snapStart.ApplyOn  != 'PublishedVersions'`

This compares the full objects, and customers should be able to configure this as 
`snapStart: new SnapStartConfig('PublishedVersions');`
if they want instead of always having to use the `SnapStartConfig.ON_PUBLISHED_VERSIONS`.

packages/aws-cdk-lib/aws-lambda/lib/snapstart-config.ts
```diff
@@ -0,0 +1,14 @@
...
+  public static readonly ON_PUBLISHED_VERSIONS= new SnapStartConfig('PublishedVersions');
+
+  readonly snapStartConfig: CfnFunction.SnapStartProperty;
+
+  protected constructor(public readonly applyValue: string) {
+    this.snapStartConfig = {
+      applyOn: applyValue,
+    };
+  }
+}
```
colifran, line 14: What about implementing this using an abstract class and let the class return the implementation? One of the methods could be `render` which could then return the snap start configuration after also performing validation. Here is how I think this could look:

```ts
export abstract class SnapStart {
  public static readonly ON_PUBLISHED_VERSIONS = SnapStart.applyOn('PublishedVersions');

  private applyOn(value: string): SnapStart {
    return new (class extends SnapStart {
      public _render(props: FunctionProps) {
        this.validateSnapStart(props);
        return { 
          applyOn: value
        } satisifies CfnFunction.SnapStartProperty;
      }

      private validateSnapStart(props: FunctionProps) {
         // move validation code here
      }
    }) ();
  }

  /*
  * @internal
  */
  public abstract _render(props: FunctionProps): any;
}
```

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -832,6 +841,7 @@ export class Function extends FunctionBase {
       codeSigningConfigArn: props.codeSigningConfig?.codeSigningConfigArn,
       architectures: this._architecture ? [this._architecture.name] : undefined,
       runtimeManagementConfig: props.runtimeManagementMode?.runtimeManagementConfig,
+      snapStart: this.configureSnapStart(props),
```
colifran, line 844: Given my below comment, we could format this as follows:

```ts
snapStart: props.snapStart ? snapStart._render(props) : undefined,
```

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3220,32 +3328,6 @@ test('function using a reserved environment variable', () => {
   })).toThrow(/AWS_REGION environment variable is reserved/);
 });
 
-test('set SnapStart to desired value', () => {
```
colifran, line 3223: Why are we removing this test?

roger-zhangg: Sorry for the confusion, I just moved this test with the new added snapstart test into the snapstart section.

packages/aws-cdk-lib/aws-lambda/lib/function.ts
roger-zhangg: Thanks, wil check this out

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -1283,6 +1293,40 @@ Environment variables can be marked for removal when used in Lambda@Edge by sett
...
+
+    if (props.filesystem) {
+      throw new Error('SnapStart is currently not supported using EFS');
+    }
+
+    if (props.ephemeralStorageSize && props.ephemeralStorageSize?.toMebibytes() > 512) {
+      throw new Error('SnapStart is currently not supported using more than 512 MiB Ephemeral Storage');
+    }
+
+    return props.snapStart.snapStartConfig;
```
colifran, line 1327: Given my below comment we could the have this be - `return props.snapStart._render()`

colifran (changes requested):
@roger-zhangg this is looking pretty good so far! See my comments and let me know if you have any questions.

packages/aws-cdk-lib/aws-lambda/lib/function.ts
colifran: Just as a heads up I think this is fine as is! It looks like we're using configure all over so we could just call `render` from inside `configureSnapStart`.

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
colifran: Ahh I see it now. Thanks for pointing that out!

packages/aws-cdk-lib/aws-lambda/lib/snapstart-config.ts
roger-zhangg: Hey Colin, Is there a example of this abstract class in the repo? I'm still pretty new to this approach so it would be nice to have a reference. Thank you!

colifran: Awesome! That should be a good reference. The various lambda extensions also use a similar approach - parameters and secrets extension, insights layer, adot layer, etc. Let me know if you have any questions.

roger-zhangg: Just pushed the change, Please check it out ,thanks!

packages/aws-cdk-lib/aws-lambda/lib/function.ts
roger-zhangg: For we moved to use abstract class/render we can't check the value inside `applyOn` So currently I'll stick with this approach. Thanks

roger-zhangg:
@colifran I've updated per our discussion, please take a look when you have time, thanks!

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -902,6 +902,23 @@ const fn = new lambda.Function(this, 'MyFunction', {
...
+SnapStart is currently supported only on Java 11/Java 17 runtime. SnapStart does not support provisioned concurrency, the arm64 architecture, Amazon Elastic File System (Amazon EFS), or ephemeral storage greater than 512 MB. After you enable Lambda SnapStart for a particular Lambda function, publishing a new version of the function will trigger an optimization process.
+
+See [the AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html) to learn more about AWS Lambda SnapStart
+
+```ts
+const fn = new lambda.Function(this, 'MyFunction', {
+  code: lambda.Code.fromAsset(path.join(__dirname, 'handler.zip')),
+  runtime: lambda.Runtime.JAVA_11,
+  handler: 'example.Handler::handleRequest',
+  snapStart: lambda.SnapStartConfig.ON_PUBLISHED_VERSIONS,
```
colifran, line 916: @roger-zhangg in case you didn't catch this from the build logs, it looks like the build is failing because you changed the class to `SnapStartConf`, but didn't update it here.

roger-zhangg: Thanks for the catch, I'll update it

roger-zhangg: Also I wonder is there a local equivalent to this cloud build? So I won't miss next time

colifran: No that I'm aware of. If I have a failing build I always use the build logs we generate to understand why it's failing. In this case, at the bottom of the build logs it pointed to the `README` so I knew to check there.

colifran (dismissed):
@roger-zhangg sorry to be doing this now, but there was some additional context I wasn't aware of regarding the first go at implementing this. Specifically, we were against adding `SnapStart` as a property on `FunctionOptions`. The reason for this is that from the perspective of a user, they would see this and want to use it not necessarily knowing all the other requirements such as runtime, architecture, storage size restrictions, etc. The direction we wanted to move this in was to implement `SnapStart` as a method on the `Function`construct. This could then be used as: `lambdaFunction.enableSnapStart()`. This method would then be responsible for performing the necessary validation and enabling `SnapStart` on the function. Again sorry for recognizing this late. @TheRealAmazonKendra can provide some more context if necessary as they were involved on the previous PR. In the meantime, I'll also think about how this method would work. Let me know if you have any questions.

roger-zhangg:
Thanks Colin, I think I got what you meant. I'll check out the code first and get to you if I have any problem

valerena:
> @roger-zhangg sorry to be doing this now, but there was some additional context I wasn't aware of regarding the first go at implementing this. Specifically, we were against adding `SnapStart` as a property on `FunctionOptions`. The reason for this is that from the perspective of a user, they would see this and want to use it not necessarily knowing all the other requirements such as runtime, architecture, storage size restrictions, etc. The direction we wanted to move this in was to implement `SnapStart` as a method on the `Function`construct. This could then be used as: `lambdaFunction.enableSnapStart()`. This method would then be responsible for performing the necessary validation and enabling `SnapStart` on the function. Again sorry for recognizing this late. @TheRealAmazonKendra can provide some more context if necessary as they were involved on the previous PR. In the meantime, I'll also think about how this method would work. Let me know if you have any questions.

What's the difference in practice with having "a method" instead of having the validation as part of the constructor? We took inspiration from the existing behavior for the VPC Configuration [`vpcConfig` field](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L822), which also has a lot of restrictions and validation.

valerena:
My main concern about that is: how do customers know what is a property and what is a method? (of course, they will have to look at the documentation). Is there any other property that exists directly in a Lambda function via the API that in CDK is not a field in the constructor? Just so we don't add this as an "exception" from all the other fields that are directly supported.

roger-zhangg:
Hi @colifran , do you have a chance to look into @valerena comments? Thanks!

colifran:
@roger-zhangg @valerena sorry for the delay. I will get back to you on this today.

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -1283,6 +1293,40 @@ Environment variables can be marked for removal when used in Lambda@Edge by sett
...
+    // SnapStart does not support arm64 architecture, Amazon Elastic File System (Amazon EFS), or ephemeral storage greater than 512 MB.
+    // SnapStart doesn't support provisioned concurrency either, but that's configured at the version level,
+    // so it can't be checked at function set up time
+    // SnapStart supports the Java 11 and Java 17 (java11 and java17) managed runtimes.
+    // See https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html
+    if (props.snapStart != SnapStartConf.ON_PUBLISHED_VERSIONS) {
+      throw new Error('SnapStart currently only support published versions');
+    }
+
+    if (props.runtime != Runtime.JAVA_11 && props.runtime != Runtime.JAVA_17 ) {
```
colifran, line 1311: @roger-zhangg looks like we're still waiting on some feedback, but pending that the only other change I would make here is to update our `Runtime` class to have `supportsSnapStart` as a `LambdaRuntimeProp`. Then we can add `supportsSnapStart` as true to any relevant `Runtime` when it gets added in the future:

https://github.com/aws/aws-cdk/blob/5fef403825110205d472bae90b63866c850efed6/packages/aws-cdk-lib/aws-lambda/lib/runtime.ts#L3-L26

We could then change this check to `if (!props.runtime.supportsSnapStart) { //... }`

mrgrain:
I think a prop with validation is fine.
We should however add a warning that informs users about the need to create a version to get any benefit from it.
See: https://github.com/aws/aws-cdk/pull/26144

roger-zhangg:
Hey @colifran @mrgrain Thank you all for your comments, I've updated the code per your suggestions. Please help to take another look when you are available. Thanks!

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -1292,6 +1302,41 @@ Environment variables can be marked for removal when used in Lambda@Edge by sett
...
+    if (!props.snapStart) {
+      return undefined;
+    }
+
+    // SnapStart does not support arm64 architecture, Amazon Elastic File System (Amazon EFS), or ephemeral storage greater than 512 MB.
+    // SnapStart doesn't support provisioned concurrency either, but that's configured at the version level,
+    // so it can't be checked at function set up time
+    // SnapStart supports the Java 11 and Java 17 (java11 and java17) managed runtimes.
+    // See https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html
+    if (props.snapStart != SnapStartConf.ON_PUBLISHED_VERSIONS) {
```
colifran, line 1316: Last thing and then we can approve this. Can we remove this? The only way a user can use `SnapStartConf` is as `SnapStartConf.ON_PUBLISHED_VERSIONS`

colifran (dismissed):
@roger-zhangg Looks good!

colifran (approved):

mriccia:
It is great to have SnapStart support with CDK, thank you for this feature.

I have a question about this feature implementation. Looking through the PR it seems that there is no check verifying that Lambda Versioning is enabled. 

> You can use SnapStart only on [published function versions](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html#configuration-versions-config) and [aliases](https://docs.aws.amazon.com/lambda/latest/dg/configuration-aliases.html) that point to versions. You can't use SnapStart on a function's unpublished version ($LATEST). [Source](https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html)

IMO it would make sense to either print out a warning, or even raise an error if SnapStart is configured without Versions being enabled. Would it be possible to consider this improvement?

mriccia:
That's great @valerena, yes this warning was what I was looking for (apologies for not spotting it in the code!).

PR #26767, opened 2023-08-15
fix(assertions): stack overflow while parsing template

### LINKED ISSUE #26766: assertions: Maximum call stack size exceeded when parsing template
### Describe the bug

Calling `Template.fromStack(stack)` on some (CDK) stacks results in a (JavaScript) stack overflow

### Expected Behavior

Either it synthesizes a template, or prints a helpful error

### Current Behavior

```
RangeError: Maximum call stack size exceeded
    at Set.values (<anonymous>)
    at recurse (/Users/chrisr/dev/cdk-test/node_modules/aws-cdk-lib/assertions/lib/private/cyclic.js:3:2297)
    at recurse (/Users/chrisr/dev/cdk-test/node_modules/aws-cdk-lib/assertions/lib/private/cyclic.js:3:2361)
    at recurse (/Users/chrisr/dev/cdk-test/node_modules/aws-cdk-lib/assertions/lib/private/cyclic.js:3:2361)
    at recurse (/Users/chrisr/dev/cdk-test/node_modules/aws-cdk-lib/assertions/lib/private/cyclic.js:3:2361)
    at recurse (/Users/chrisr/dev/cdk-test/node_modules/aws-cdk-lib/assertions/lib/private/cyclic.js:3:2361)
    at recurse (/Users/chrisr/dev/cdk-test/node_modules/aws-cdk-lib/assertions/lib/private/cyclic.js:3:2361)
    at recurse (/Users/chrisr/dev/cdk-test/node_modules/aws-cdk-lib/assertions/lib/private/cyclic.js:3:2361)
    at recurse (/Users/chrisr/dev/cdk-test/node_modules/aws-cdk-lib/assertions/lib/private/cyclic.js:3:2361)
    at recurse (/Users/chrisr/dev/cdk-test/node_modules/aws-cdk-lib/assertions/lib/private/cyclic.js:3:2361)
```

### Reproduction Steps

```ts
interface MyTriggerProps extends NodejsFunctionProps {
  executeAfter?: IConstruct[];
  executeBefore?: IConstruct[];
}

class MyTrigger extends Construct {
  public readonly fn: NodejsFunction;
  constructor(scope: Construct, id: string, props: MyTriggerProps) {
    super(scope, id);

    this.fn = new NodejsFunction(this, "Function", {
      entry: props.entry,
      handler: props.handler,
      functionName: props.functionName,
    });

    const trigger = new Trigger(this, "Trigger", {
      handler: this.fn,
    });
    trigger.executeAfter(...(props.executeAfter ?? []));
    trigger.executeBefore(...(props.executeBefore ?? []));
  }
}

export class CdkTestStack extends Stack {
    super(scope, id, props);

    const triggerHello = new MyTrigger(this, "TriggerHello", {
      handler: "handler",
      entry: join(__dirname, "lmb/hello.ts"),
      functionName: "hello",
    });

    const triggerGoodbye = new MyTrigger(this, "TriggerGoodbye", {
      handler: "handler",
      entry: join(__dirname, "lmb/goodbye.ts"),
      functionName: "goodbye",
    });

    const triggerMiddle = new MyTrigger(this, "TriggerMiddle", {
      handler: "handler",
      entry: join(__dirname, "lmb/middle.ts"),
      functionName: "middle",
      executeAfter: [triggerHello.fn],
      executeBefore: [triggerGoodbye.fn],
    });
  }
}

const app = new cdk.App();
let stack = new CdkTestStack(app, 'CdkTestStack');
const template = Template.fromStack(stack);
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.92.0

### Framework Version

_No response_

### Node.js Version

18.14.1

### OS

macOS 13.4

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

Chriscbr:
Closes #26766

The function `findCycle` tries to find a cycle by using a depth-first search (DFS). The DFS is implemented recursively in the recurse function. For each node, it tries to find a path that eventually leads back to the start of the path. If such a path is found, a cycle exists, and the nodes forming this cycle are returned.

One of the bugs in the current implementation is that it only checks whether the current dependency `dep` is equal to the first node of the current path `path[0]`. This means it will only detect a cycle if the cycle includes the first node of the search, which might not always be the case.

To fix this, the function should check whether the current dependency `dep` is already somewhere in the current path `path`. If it is, then a cycle has been found.

----


Chriscbr:
exemption request please 🙂

packages/aws-cdk-lib/assertions/test/template.test.ts
```diff
@@ -1345,6 +1345,29 @@ describe('Template', () => {
     }).toThrow(/dependency cycle/);
   });
 
+  test('throws when given a more complex template with cyclic dependencies', () => {
```
kaizencc, line 1348: While i can see how the change in `cyclic.ts` makes the `recurse` function more efficient, this test isn't going to show that one way or another right? the test will succeed with the old code as well?

I'm prepared to just approve as is anyway, because this looks low risk and is hard to test, but want to make sure i have the whole picture first.

kaizencc (commented):
Hi @Chriscbr :) Can you provide a bit more in the PR description explaining how this is the root cause for the stack overflow issue?

packages/aws-cdk-lib/assertions/test/template.test.ts
Chriscbr: I initially ran the test with the old code (unchanged) and it fails for me because of a stackoverflow error. 
Screenshot 2023-08-17 at 5 46 06 PM

I think the call stack in the test case I added ends up looking like this (without the fix applied):

```
findCycle(deps)
recurse("Res1", ["Res1"])
recurse("Res2", ["Res1", "Res2"])
recurse("Res3", ["Res1", "Res2", "Res3"])
recurse("Res2", ["Res1", "Res2", "Res3", "Res2"])
recurse("Res3", ["Res1", "Res2", "Res3", "Res2", "Res3"])
...
```

Chriscbr:
@kaizencc I updated the description 🙏

packages/aws-cdk-lib/assertions/test/template.test.ts
kaizencc: Ah i see

kaizencc (approved):
Thanks @Chriscbr

Issue #26771, opened 2023-08-16
eks: ProxyAgent is not a constructor


trondhindenes:
### Describe the bug

The newest cdk library (2.92.0) produces this error. Downgrading to 2.90.0 solves it. We create clusters where we specify the `kubectl_layer` parameter. 

### Expected Behavior

it shouldn't fail

### Current Behavior

I'm getting this error:
```
11:02:05 AM | CREATE_FAILED        | Custom::AWSCDK-EKS-Cluster            | clusterC5B25D0D
Received response status [FAILED] from custom resource. Message returned: ProxyAgent is not a constructor
```

### Reproduction Steps

See descr

### CDK CLI Version

2.89.0

### Framework Version

2.92.0

### Node.js Version

v18.17.0

### OS

Ubuntu 

### Language

Python

### Language Version

Python 3.8

pahud:
Can you share the smallest code snippet to reproduce this error?

juweeks:
this is also breaking my eks deployment on v2.92. reverted back to v2.91 and seems to be fixed

cfn error:
```
Received response status [FAILED] from custom resource. Message returned: ProxyAgent is not a constructor

Logs: /aws/lambda/xxxxOnEventHandler42BEBAE0-jXvWcu93ZFF4

at Object.<anonymous> (/var/task/index.js:1:417)
at Module._compile (node:internal/modules/cjs/loader:1256:14)
at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
at Module.load (node:internal/modules/cjs/loader:1119:32)
at Module._load (node:internal/modules/cjs/loader:960:12)
at Module.require (node:internal/modules/cjs/loader:1143:19)
at require (node:internal/modules/cjs/helpers:110:18)
at _tryRequireFile (file:///var/runtime/index.mjs:912:37)
at _tryRequire (file:///var/runtime/index.mjs:962:25)
at _loadUserApp (file:///var/runtime/index.mjs:991:22) (RequestId: 09418d76-9c40-46da-9cfe-1d056ff8e2ad)


 ❌  xxxstack failed: Error: The stack named xxxstack failed to deploy: UPDATE_ROLLBACK_COMPLETE: Received response status [FAILED] from custom resource. Message returned: ProxyAgent is not a constructor

Logs: /aws/lambda/xxxxOnEventHandler42BEBAE0-jXvWcu93ZFF4

    at Object.<anonymous> (/var/task/index.js:1:417)
[...]

PR #26782, opened 2023-08-16
fix(eks): undefined constructor in custom resource

MrArnoldPalmer:
When we [upgraded the proxy-agent v6](https://github.com/aws/aws-cdk/pull/26722) we did not update how we were importing from the package. Where [in 5.0 commonjs exports were used](https://github.com/TooTallNate/node-proxy-agent/blob/5.0.0/index.js#L25) , in [6.0 a named esm export is used](https://github.com/TooTallNate/proxy-agents/blob/proxy-agent%406.1.0/packages/proxy-agent/src/index.ts#L71).

Updated to use a named import statement instead of a default require. Typescript also confirms the old import was an error now since we use `import` instead of `require`.

Fixes: #26771

----


MrArnoldPalmer:
I ran the `eks-cluster` integ test using `--force` because it doesn't have `diffAssets` enabled. I confirmed that this test was failing with the expected error before. I did not --force the run of all of the other eks integ tests because they take literally forever.

We should diffAssets on all of these tests but it results in needing to re-run them on a cadence that is currently intolerable without automation doing it for us.

kaizencc (approved):

PR #26784, opened 2023-08-17
docs(iot): fix CloudWatch Metric example

rix0rrr:
This example gets replicated to the CloudWatch Metrics documentation, where it is wrong.

----


kaizencc (approved):
Double checked to make sure that `dimensions` exists but is deprecated, which is why rosetta did not catch this

Issue #26785, opened 2023-08-17
(cli-lib-alpha): Missing hotswap option in the deploy command


amine-mf:
### Describe the feature

I'm missing the `hotswap` deployment option in the CDK CLI that is mandatory for our use case.

### Use Case

We use a lambda to deploy state machines, without the `hotswap` it is way too slow for our use.
We are currently using an older version of CDK and trying to upgrade. So we are adapting our code to use `cli-lib-alpha` instead of using `SdkProvider` and `CloudFormationDeployments`.

### Proposed Solution

#26786

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.92.0

### Environment details (OS name and version, etc.)

macOS Ventura

peterwoodworth:
Thanks for the request and PR 🙂

fab-mindflow:
Our team has intensively used internal API to run CDK in lambdas. We are in the process to move to `cli-lib` because we need to stick with latest CDK version. Hence the work by @amine-mf because this is blocking for us. We will continue to provide feedback as needed on this module as we migrate. Do you think this can be merged pretty soon in upcoming release so that we don't fork meanwhile?

amine-mf:
@peterwoodworth A gentle reminder. Thanks a lot. (Sorry to insist, but we really need it)

amine-mf:
I have no idea on your release process, is there a way to know when is the next release?
🙏

mrgrain:
> @mrgrain I have no idea on your release process, is there a way to know when is the next release? 🙏

We release weekly, usually Wednesdays. But various things can affect this.
For this week we seem to be on track.

mrgrain:
I spoke too soon. Looks like it will be Friday this week. @amine-mf

PR #26786, opened 2023-08-17
feat(cli-lib-alpha): support hotswap deployments

amine-mf:
Closes aws/aws-cdk#26785.

Exemption Request: The `cli-lib-alpha` README does not currently specify the exhaustive list of arguments, which looks like a choice, thus I did not add the new argument. Currently, no integration tests were implemented for this lib, only unit tests which I did update/add.

----


ajhool:
I recently submitted an issue related to the bootstrap roles and Hotswap in the CDK CLI (linked above, #26978). I'm not sure whether CDK CLI and CDK Lib use the same approach to assuming CDK Boostrap IAM roles, but if they do then I think that issue also affects CDK Lib. 

TLDR: The CDK bootstrap roles don't have necessary permissions for Hotswap to work, so for hotswap you would need to create a custom IAM role that has all required permissions (like `UpdateLambda`-type permissions). The #26978 is a request for the hotswap team to create a CDK role that includes the necessary permissions so that users don't need to manage custom IAM users / roles.

PS. First time seeing that CDK Lib is in alpha and I'm really excited!

amine-mf:
> I recently submitted an issue related to the bootstrap roles and Hotswap in the CDK CLI (linked above, #26978). I'm not sure whether CDK CLI and CDK Lib use the same approach to assuming CDK Boostrap IAM roles, but if they do then I think that issue also affects CDK Lib.
> 
> TLDR: The CDK bootstrap roles don't have necessary permissions for Hotswap to work, so for hotswap you would need to create a custom IAM role that has all required permissions (like `UpdateLambda`-type permissions). The #26978 is a request for the hotswap team to create a CDK role that includes the necessary permissions so that users don't need to manage custom IAM users / roles.
> 
> PS. First time seeing that CDK Lib is in alpha and I'm really excited!

We never faced the issue before, we run the hotswap as admins as we do it only on our dev sandboxes (talking about `lambdas`). Using the `cli-lib-alpha`, we hotswap state machines, and we did not face any issue related to access rights (this runs inside a lambda with a specific role).
I think this `lib` and the `cli` do share code and the problem is probably more global but for me it is out of this PR's scope. Especially that it is very urgent for our team to have it merged.

amine-mf:
Can anybody add pr-linter/exempt-integ-test and pr-linter/exempt-readme labels as per the Exemption Request on the first comment? (This will give a more accurate state of the PR and remove the change request status).
Thanks

mrgrain:
@amine-mf Apologies. I have been busy this week. Should be able to get to it next week.

mrgrain (approved):

PR #26800, opened 2023-08-18
fix(ecr): AutoDeleteImages custom resource description contains "undefined"

go-to-k:
The repository name is embedded in the `description` of `CustomResourceProvider` for the ECR AutoDeleteImages, but the repository name becomes to `undefined` because the `CustomResourceProvider` is created before the repository name is set.

```ts
  private enableAutoDeleteImages() {
    const firstTime = Stack.of(this).node.tryFindChild(`${AUTO_DELETE_IMAGES_RESOURCE_TYPE}CustomResourceProvider`) === undefined;
    const provider = CustomResourceProvider.getOrCreateProvider(this, AUTO_DELETE_IMAGES_RESOURCE_TYPE, {
      codeDirectory: path.join(__dirname, '..', '..', 'custom-resource-handlers', 'dist', 'aws-ecr', 'auto-delete-images-handler'),
      useCfnResponseWrapper: false,
      runtime: CustomResourceProviderRuntime.NODEJS_18_X,
      description: `Lambda function for auto-deleting images in ${this.repositoryName} repository.`,
    });
```

So I moved the `enableAutoDeleteImages` calls after following codes that set `repositoryName` in the `Repository` constructor.

```ts
this.repositoryName = this.getResourceNameAttribute(resource.ref);
```

----


tmokmss (commented):
If we just move this assignment before calling `enableAutoDeleteImages`, will that resolve the problem?

https://github.com/aws/aws-cdk/blob/bc86d050295fc0d4904e4123f120d706689ca952/packages/aws-cdk-lib/aws-ecr/lib/repository.ts#L722

go-to-k:
@tmokmss 

> If we just move this assignment before calling enableAutoDeleteImages, will that resolve the problem?

In that case, `${this.repositoryName}` is just `Token[XX]` instead of `undefined`, which may not make much sense.

I considered that as well, since it is more natural for the source code, but I did not change it out of respect for the intent of the original code.

go-to-k:
@tmokmss

Sorry, I was misunderstood.

The `getResourceNameAttribute` uses `mimicReference`. So we should move the assignment before calling `enableAutoDeleteImages`.

I will change.

tmokmss (approved):
Thanks, LGTM. Make sure you change PR title and description as well :)

go-to-k:
Thank you, I changed!

rix0rrr (dismissed):

rix0rrr (approved):

PR #26808, opened 2023-08-18
feat(cloudfront): Add RealtimeLogConfig to Distribution

jamiepmullan:
Pull request to implement RealtimeLogConfig on CloudFront Distribution.
As well as being able to use `RealtimeLogConfigArn` on `Distribution`, I also added `RealtimeLogConfig` as a Construct so that the Cfn is abstracted too.

The reason this is needed so that we can add realtime log config to a `Distribution` without reverting to using `CfnDistribution`.

Closes #<issue number here>.

----


packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+export interface KinesisStreamConfig {
+  /**
+   * The Amazon Resource Name (ARN) of an AWS Identity and Access Management (IAM) role that CloudFront can use to send real-time log data to your Kinesis data stream.
+   */
+  readonly roleArn: string;
+  /**
+   * The Amazon Resource Name (ARN) of the Kinesis data stream where you are sending real-time log data.
+   */
+  readonly streamArn: string;
```
colifran, line 23: These both should just be the respective interfaces. For example, `readonly role: IRole` and `readonly kinesisStream: IStream`

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,96 @@
...
+export interface EndPoint {
+  /**
+   * Contains information about the Amazon Kinesis data stream where you are sending real-time log data.
+   */
+  readonly kinesisStreamConfig: KinesisStreamConfig,
+  /**
+   * The type of data stream where you are sending real-time log data.
+   */
+  readonly streamType: DataStreamType;
+}
```
colifran, line 38: I have a few thoughts about this:
1. If the only `streamType` is kinesis, why do we need to have that as a property? Couldn't we just set that for the user rather than having them provide it? I guess my counter thought is will there be more `streamTypes` in the future
2. We recommend not nesting properties in our design guidelines. To me, it doesn't seem like the `kinesisStreamConfig` property is necessary here. Why not just just move the properties from the `kinesisStreamConfig` directly into this interface?

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    const resource: CfnRealtimeLogConfig = new CfnRealtimeLogConfig(this, 'Resource', {
+      endPoints: props.endPoints,
+      fields: props.fields,
+      name: this.physicalName,
+      samplingRate: props.samplingRate,
+    });
+
+    this.realtimeLogConfigName = resource.ref;
+    this.realtimeLogConfigArn = resource.attrArn;
```
colifran, line 94: This should be:
```ts
this.realtimeLogConfigName = this.getResourceArnAttribute(resource.attrArn, {
  service: 'cloudfront',
  resource: 'realtime-log-config',
  resourceName: this.physicalName,
});
```

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    const resource: CfnRealtimeLogConfig = new CfnRealtimeLogConfig(this, 'Resource', {
+      endPoints: props.endPoints,
+      fields: props.fields,
+      name: this.physicalName,
+      samplingRate: props.samplingRate,
+    });
+
+    this.realtimeLogConfigName = resource.ref;
```
colifran, line 93: This should be:
```ts
this.realtimeLogConfigName = this.getResourceNameAttribute(resource.ref);
```

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+export interface IRealtimeLogConfig extends IResource {
+  /**
+   * The name of the realtime log config.
+   */
+  readonly realtimeLogConfigName: string;
+  /**
+   * The arn of the realtime log config.
+   */
+  readonly realtimeLogConfigArn: string;
```
colifran, line 51: Since these are attributes, we should also be using the `@attribute` tag. So,
```ts
**
* The name of the realtime log config.
*
* @attribute
*/
readonly realtimeLogConfigName: string;
```

colifran (dismissed):
@jamiepmullan Thanks for getting this started! I left some initial comments for you to review.

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: > I have a few thoughts about this:
> 
> 1. If the only `streamType` is kinesis, why do we need to have that as a property? Couldn't we just set that for the user rather than having them provide it? I guess my counter thought is will there be more `streamTypes` in the future

Yeah - I wasn't sure if we were going to introduce new types in the future - so wanted to keep it expandable incase

> 2. We recommend not nesting properties in our design guidelines. To me, it doesn't seem like the `kinesisStreamConfig` property is necessary here. Why not just just move the properties from the `kinesisStreamConfig` directly into this interface?

That makes sense to move it inside the interface to adhere to design guidelines

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,96 @@
...
+  constructor(scope: Construct, id: string, props: RealtimeLogConfigProps) {
+    super(scope, id, {
+      physicalName: props.name,
+    });
+
+    if ((props.samplingRate < 1 || props.samplingRate > 100)) {
+      throw new Error(`Sampling rate must be between 1 and 100 (inclusive), received ${props.samplingRate}`);
+    }
+
+    const resource: CfnRealtimeLogConfig = new CfnRealtimeLogConfig(this, 'Resource', {
```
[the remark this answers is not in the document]
jamiepmullan: Yeah - it will - my bad, was from first draft when I was getting stuff working! Will remove now

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: Changed!

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: changed!

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: that makes sense

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
colifran: Per our slack discussion - it sounds like other stream types will be added so we want to make sure we can expand the implementation when that happens. An enum-like stream config class might be a good approach for this:

```ts
abstract class StreamConfig {
  public static fromKinesisStream(stream: IStream, role: IRole): StreamConfig {
    return new (class extends StreamConfig {
      public _renderStreamConfig() {
        return {
          kinesisStreamConfig: {
            roleArn: role.roleArn,
            streamArn: stream.streamArn,
          },
          streamType: 'Kinesis',
        }
      }
    });
  }
 
  // New streams could just be added here when support is added for them

  private constructor() {}
  
  /*
  * @internal
  */
  public _renderStreamConfig(): any;
  ```

File: packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts [hunk 7]
```diff
@@ -0,0 +1,106 @@
...
+import { CfnRealtimeLogConfig } from './cloudfront.generated';
+import * as iam from '../../aws-iam';
+import * as kinesis from '../../aws-kinesis';
+import { IResource, Resource } from '../../core';
+
+/**
+ * Represents the endpoints available for targetting within a realtime log config resource
+ */
+export abstract class Endpoint {
+  public static fromKinesisStream(stream: kinesis.IStream, role: iam.IRole): Endpoint {
```
jogold, line 11: * The role could be made optional and created by the construct
* The IAM permissions should be automatically configured by the construct (https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/real-time-logs.html#understand-real-time-log-config-iam-role)

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+/**
+ * Properties for defining a RealtimeLogConfig resource.
+ */
+export interface RealtimeLogConfigProps {
+  /**
+   * The unique name of this real-time log configuration.
+   */
+  readonly name: string;
```
jogold, line 56: Can be made optional by using `Names.uniqueId()` as default

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * A list of fields that are included in each real-time log record.
+   */
+  readonly fields: string[];
+  /**
+   * The sampling rate for this real-time log configuration.
+   */
+  readonly samplingRate: number;
```
jogold, line 64: Should this default to 100?

packages/aws-cdk-lib/aws-cloudfront/test/distribution.test.ts
```diff
@@ -1146,3 +1149,54 @@ test('grants createInvalidation', () => {
...
+
+  new Distribution(stack, 'MyDist', {
+    defaultBehavior: {
+      origin: defaultOrigin(),
+      realtimeLogConfig: realTimeConfig,
+    },
+  });
+
+  Template.fromStack(stack).hasResourceProperties('AWS::CloudFront::Distribution', {
+    DistributionConfig: {
```
jogold, line 1180: We can use `Match.objectLike` here to only check the relevant part of the test (`RealtimeLogConfigArn`)

packages/aws-cdk-lib/aws-cloudfront/test/distribution.test.ts
```diff
@@ -1146,3 +1149,54 @@ test('grants createInvalidation', () => {
...
+      Origins: [{
+        DomainName: 'www.example.com',
+        Id: 'StackMyDistOrigin1D6D5E535',
+        CustomOriginConfig: {
+          OriginProtocolPolicy: 'https-only',
+        },
+      }],
+    },
+  });
+});
```
jogold, line 1202: Also add a test to check the correct the setup of the IAM permissions when implemented.

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: I did it this way as to auto-generate we would have to alter it to also include KMS key (for encrypted streams)? i.e.
```
  public static fromKinesisStream(stream: kinesis.IStream, role?: iam.IRole, kmsKey?: kms.Key): Endpoint {
```

jamiepmullan: I'll take a look thanks!

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: As a user needs to specify the fields they want to capture - I think it makes sense to keep this configurable too (with no default)

packages/aws-cdk-lib/aws-cloudfront/test/distribution.test.ts
jamiepmullan: added IAM tests

packages/aws-cdk-lib/aws-cloudfront/test/distribution.test.ts
jamiepmullan: Fixed 👍

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: Changed to make `name` optional

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+    });
+  }
+
+  private constructor() {}
+
+  /**
+  * @internal
+  */
+  public abstract _renderEndpoint(): any;
```
jogold, line 66: How about this:

```ts
  public abstract _renderEndpoint(scope: Construct): any;
```

which let's you do `endpoint._renderEndpoint(this)` when creating the `CfnRealtimeLogConfig` (the role is created in the scope of the realtime log config) and then you don't have to pass a scope when using `fromKinesisStream()`

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,144 @@
...
+ * Represents the endpoints available for targetting within a realtime log config resource
+ */
+export abstract class Endpoint {
+  /**
+   * Configure a Kinesis Stream Endpoint for Realtime Log Config
+   *
+   * @default - a role will be created and used across your endpoints
+   */
+  public static fromKinesisStream(scope: Construct, stream: kinesis.IStream, role?: iam.IRole): Endpoint {
+    const cloudfrontRole = role || singletonKinesisRole(scope);
```
jogold, line 33:
```suggestion
    const cloudfrontRole = role ?? singletonKinesisRole(scope);
```

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    if (!role) {
```
jogold, line 35: We can setup the permissions even if a role is passed

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jogold: ok

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: Makes sense - will change!

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: Good spot - thanks!

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: Sure - will remove the conditional check

jamiepmullan:
@jogold thanks for the reviews! I really appreciate it

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    if ((props.samplingRate < 1 || props.samplingRate > 100)) {
+      throw new Error(`Sampling rate must be between 1 and 100 (inclusive), received ${props.samplingRate}`);
+    }
+
+    const resource = new CfnRealtimeLogConfig(this, 'Resource', {
+      endPoints: props.endPoints.map(endpoint => {
+        return endpoint._renderEndpoint(this);
+      }),
+      fields: props.fields,
```
colifran, line 131: Have we considered users that may want to add `fields` or `endpoints` after creating the `RealtimeLogConfig`? This could be done with something like `addEndpoint` or `addField`. Note that these two properties would then need to be evaluated with `Lazy`

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,144 @@
...
+   */
+  readonly endPoints: Endpoint[];
+}
+
+/**
+ * A Realtime Log Config configuration
+ *
+ * @resource AWS::CloudFront::RealtimeLogConfig
+ */
+export class RealtimeLogConfig extends Resource implements IRealtimeLogConfig {
```
colifran, line 114: I think this class needs the static methods that allow a user to import an existing `RealtimeLogConfig`. We generally have 3 - `fromRealtimeLogConfigName`, `fromRealtimeLogConfigArn`, `fromRealtimeLogConfigAttributes` - for example - https://github.com/aws/aws-cdk/blob/27b720135e30d58732997a5a5e172daaa9453381/packages/aws-cdk-lib/aws-iam/lib/instance-profile.ts#L113-L157

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: I can't see it being used afterward `RealtimeLogConfig` - is it something we could come back to as a feature if required/asked for?

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,144 @@
...
+  public readonly realtimeLogConfigName: string;
+  public readonly realtimeLogConfigArn: string;
+
+  constructor(scope: Construct, id: string, props: RealtimeLogConfigProps) {
+    super(scope, id, {
+      physicalName: props.realtimeLogConfigName ?? Lazy.string({ produce: () => Names.uniqueId(this) }),
```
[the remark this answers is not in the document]
jamiepmullan: Will revert this change...

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,144 @@
...
+    assumedBy: new iam.ServicePrincipal('cloudfront.amazonaws.com'),
+  });
+
+  return role;
+}
+
+/**
+ * Represents the endpoints available for targetting within a realtime log config resource
+ */
+export abstract class Endpoint {
```
colifran, line 26: At this point, this class probably needs to be moved into its own file. We can call it `endpoint.ts`. With that, we should also have unit tests for this.

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
```diff
@@ -0,0 +1,144 @@
+import { Construct } from 'constructs';
+import { CfnRealtimeLogConfig } from './cloudfront.generated';
+import * as iam from '../../aws-iam';
+import * as kinesis from '../../aws-kinesis';
+import { IResource, Lazy, Names, PhysicalName, Resource } from '../../core';
+
+/**
+ * @internal
+ */
+function singletonKinesisRole(scope: Construct): iam.IRole {
```
colifran, line 10: Could we just move this into the `Endpoint` class and make it a private method?

colifran (changes requested):
@jamiepmullan Looking good! I added some comments. Thanks for your continued work on this :)

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: Moved - i moved some tests over too - to just get some coverage on the `render` part of this class

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
jamiepmullan: Added them

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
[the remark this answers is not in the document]
jogold: You need `Lazy` here otherwise you cannot call `this` because `super()` has not been called yet...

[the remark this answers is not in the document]
colifran: Ahh yeah you're right!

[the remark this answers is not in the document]
jamiepmullan: in that case - do we want `Names.uniqueId` or `Names.uniqueResourceName`?

packages/aws-cdk-lib/aws-cloudfront/lib/realtime-log-config.ts
colifran: Yeah that's fair.

colifran (dismissed):
@jamiepmullan Great work! Thanks for your contribution.

colifran (approved):

PR #26813, opened 2023-08-18
fix(secretsmanager): fix cross-region policy arn for imported secrets

### LINKED ISSUE #26811: secretsmanager: Cross region policy reference imported Secret has wildcard suffix on completeSecretArn
### Describe the bug

Pre 2.89.0 we could import a secret from a complete secret arn in one stack and reference this secret from another stack in a different region to include it in a policy/role through grantRead on the secret construct. 

Since 2.89.0 the arn in the policy it treats the compleSecretArn as a partial arn adding -?????? which makes the policy invalid and not allowing access to the secret as intended.

### Expected Behavior

If importing secret from a complete arn the arn used in policies should be complete arn and not include wildcard for matching a suffix. Which allows intended action by the policy.

### Current Behavior

When referencing import secret construct from a cross-region stack the arn used in policies has wildcard suffix matcher appended which makes the referenced policy invalid for allowing intended action.

### Reproduction Steps

- Create cdk app with following stack
- Create stack in one region which includes a Secret construct imported from a complete secret arn
- Create stack in other region which includes Role which is granted access to the secret with grantRead on the secret construct
- synthesis stack, and see additional role suffix.

Reproduction repository: 
https://github.com/rv2673/aws-cdk-cross-region-secret-import-reproduction

### Possible Solution

- Override `arnForPolicies` in the class returned from fromSecretAttributes method on Secret construct.
- Fix conditions in the _arnForPolicies Lazy.string implementation function

_No response_

### CDK CLI Version

2.92.0

### Framework Version

_No response_

### Node.js Version

18

### OS

ubuntu

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

rv2673:
Pre 2.89.0 we could import a secret from a complete secret arn in one stack and reference this secret from another stack in a different region to include it in a policy/role through grantRead on the secret construct.

Since 2.89.0 the arn in the policy it treats the compledSecretArn as a partial arn adding -?????? which makes the policy invalid and not allowing access to the secret as intended.

This PR fixes that by overriding arnForPolicies for imported secrets to either return provided complete arn or partial arn with suffix.

Fixes #26811.

----


packages/aws-cdk-lib/aws-secretsmanager/test/secret.test.ts
```diff
@@ -1390,3 +1390,59 @@ test('cross-environment grant with direct object reference', () => {
...
+          'secretsmanager:GetSecretValue',
+          'secretsmanager:DescribeSecret',
+        ],
+        Effect: 'Allow',
+        Resource: `${secretPartialArn}-??????`,
+      }],
+    },
+  });
+
+});
```
mrgrain, line 1448:
```suggestion
});

```

mrgrain (approved):

mrgrain (approved):

PR #26821, opened 2023-08-21
fix(core): logical IDs are not allowed to start with digits

### LINKED ISSUE #26075: core: logical ID validation regex is inconsistant with cloudformation specification

kirkwat:
The resource logical ID validation regex (`VALID_LOGICALID_REGEX`) is updated to be consistent with the [CloudFormation documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html) which says that logical IDs are alphanumeric.

Closes #26075.

----


packages/aws-cdk-lib/core/test/logical-id.test.ts
```diff
@@ -266,6 +286,19 @@ function generateString(chars: number) {
...
+function generateAlphaNumericString(chars: number) {
+  let s = '';
+  for (let i = 0; i < chars; ++i) {
+    s += randomAlphaNumeric();
+  }
+  return s;
+
+  function randomAlphaNumeric() {
+    const alphaNumericChars = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789';
+    return alphaNumericChars.charAt(Math.floor(Math.random() * 62));
```
lpizzinidev, line 298: Better to go with `Math.random() * alphaNumericChars.length`

packages/aws-cdk-lib/core/test/logical-id.test.ts
```diff
@@ -266,6 +286,19 @@ function generateString(chars: number) {
   }
 }
 
+function generateAlphaNumericString(chars: number) {
```
lpizzinidev, line 289: We should use `generateAlphaNumericString` also [here](https://github.com/aws/aws-cdk/blob/26dcc1e11a07d93681145049aa06d80a7d2114b9/packages/aws-cdk-lib/core/test/logical-id.test.ts#L171-L176).

lpizzinidev (changes requested):
Thanks for the contribution! 💪
Looks good overall, just a couple of minor changes on the tests should be done in my opinion.

rix0rrr (approved):

rix0rrr (approved):

Issue #26804, opened 2023-08-18
aws_appconfig_alpha: delete fail on stack destroy


ran-isenberg:
### Describe the bug

I'm using the new L2 constructs, sample code below.
Deploying works just fine.
However, on destroy i get the following error:
`11:24:04 AM | DELETE_FAILED        | AWS::AppConfig::ConfigurationProfile       | rootappconfigOrder...ionProfile9C631413
Cannot delete configuration profile bvdne2p because there are still hosted configuration versions existing under it. (Service: AmazonAppConfig; Status Code: 400; Error Code:
BadRequestException; Request ID: d1d449de-9138-4a5e-8461-05f925d07bae; Proxy: null)
`

Python code:
```python
from pathlib import Path

import aws_cdk.aws_appconfig_alpha as appconfig
from aws_cdk import Duration
from constructs import Construct

from cdk.my_service.configuration.schema import FeatureFlagsConfiguration

DEFAULT_DEPLOYMENT_STRATEGY = 'AppConfig.AllAtOnce'

CUSTOM_ZERO_TIME_STRATEGY = 'zero'


class ConfigurationStore(Construct):

    def __init__(self, scope: Construct, id_: str, environment: str, service_name: str, configuration_name: str) -> None:
        """
        This construct should be deployed in a different repo and have its own pipeline so updates can be decoupled from
        running the service pipeline and without redeploying the service lambdas.

        Args:
            scope (Construct): The scope in which to define this construct.
            id_ (str): The scoped construct ID. Must be unique amongst siblings. If the ID includes a path separator (``/``), then it will be
                        replaced by double dash ``--``.
            environment (str): environment name. Used for loading the corresponding JSON file to upload under
                               'configuration/json/{environment}_configuration.json'
            service_name (str): application name.
            configuration_name (str): configuration name
        """
        super().__init__(scope, id_)

        configuration_str = 'bla bla freeform conf'
        self.app_name = f'{id_}{service_name}'
        self.config_app = appconfig.Application(
            self,
            id=self.app_name,
            name=self.app_name,
        )

        self.config_env = appconfig.Environment(
            self,
            id=f'{id_}env',
            application=self.config_app,
            name=environment,
        )

        # zero minutes, zero bake, 100 growth all at once
        self.config_dep_strategy = appconfig.DeploymentStrategy(
            self,
            f'{id_}{CUSTOM_ZERO_TIME_STRATEGY}',
[...]

PR #26829, opened 2023-08-21
fix(appconfig): apply retain removal policy only to hosted configurat…

chenjane-dev:
…ion version update replace

Addressing https://github.com/aws/aws-cdk/issues/26804 to add retain removal policy only on update replace and not on delete.

> I'm using the new L2 constructs, sample code below.
Deploying works just fine.
However, on destroy i get the following error:
11:24:04 AM | DELETE_FAILED        | AWS::AppConfig::ConfigurationProfile       | rootappconfigOrder...ionProfile9C631413 Cannot delete configuration profile bvdne2p because there are still hosted configuration versions existing under it. (Service: AmazonAppConfig; Status Code: 400; Error Code: BadRequestException; Request ID: d1d449de-9138-4a5e-8461-05f925d07bae; Proxy: null) 

This fixes the problem, since before we are using `applyRemovalPolicy(RemovalPolicy.RETAIN)` on the hosted config version, which sets the removal to retain on both delete and update. We wanted to set this to retain so hosted config versions are not deleted if updated. However, if the stack is being deleted, then we don't want to retain the hosted config version and that is why we are fixing it to only set the removal policy to retain on an update replace.

> In addition, I see reference in the original thread to "I have a CloudFormation template that destroys fine". The current change does not seem to move us closer to destroying properly?

This is in reference to the user using our L1 constructs, where the hosted config version deletes correctly.

----


rix0rrr (changes requested):
Please make sure that your PR body describes the problem the PR is solving, and the design approach and alternatives considered. Explain why the PR solves the problem. A link to an issue is helpful, but does not replace an explanation of your thought process (See [Contributing Guide, Pull Requests](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request))

----

In addition, I see reference in the original thread to "I have a CloudFormation template that destroys fine". The current change does not seem to move us closer to destroying properly?

rix0rrr:
Why do we have RETAIN on these resources in the first place? Do we know?

Are we sure using a different policy for stack updates and stack deletes is the right solution? It feels to me that either we need to retain them in both cases, for some reason, or we need to retain them in neither case.

Here's my motivation for questioning this solution: if we had the following sequence of events:

## 1.  Deploy using HostedConfiguration and HostedConfigurationVersion

The service state will look like this:

```
                                                                                           
                 State of service                     State tracked by CloudFormation      
                                                                                           
                                                                                           
┌────────────────────────┐                               ┌────────────────────────────────┐
│                        │                               │                                │
│  HostedConfiguration   │◀──────────────────────────────│      HostedConfiguration       │
│                        │                               │                                │
└────────────────────────┘                               └────────────────────────────────┘
             │                                                                             
             │    ┌─────────────────┐                    ┌────────────────────────────────┐
             │    │                 │                    │                                │
             └────│   HCVersion 1   │◀───────────────────│   HostedConfigurationVersion   │
                  │                 │                    │                                │
                  └─────────────────┘                    └────────────────────────────────┘
```


## 2. Update HostedConfigurationVersion causing a replacement

The old Version will be RETAINed, because its `UpdateReplacePolicy` will be set to RETAIN.

The service will end up 2 versions, of which CloudFormation will be tracking the latest.

```
                                                                                           
                 State of service                     State tracked by CloudFormation      
                                                                                           
                                                                                           
┌────────────────────────┐                               ┌────────────────────────────────┐
│                        │                               │                                │
│  HostedConfiguration   │◀──────────────────────────────│      HostedConfiguration       │
│                        │                               │                                │
└────────────────────────┘                               └────────────────────────────────┘
             │                                                                             
             │    ┌─────────────────┐                                                      
             │    │                 │                                                      
             ├────│   HCVersion 1   │                                                      
             │    │                 │                                                      
             │    └─────────────────┘                                                      
             │                                                                             
             │                                                                             
             │    ┌─────────────────┐                    ┌────────────────────────────────┐
             │    │                 │                    │                                │
             └────│   HCVersion 2   │◀───────────────────│   HostedConfigurationVersion   │
                  │                 │                    │                                │
                  └─────────────────┘                    └────────────────────────────────┘
```

## 3. Delete stack

CloudFormation will delete the resources it knows about, in reverse dependency order.

Because CloudFormation doesn't know about `HCVersion 1` anymore, it will not delete it, and so deleting the `HostedConfiguration` will fail.


```
                                                                                           
                 State of service                     State tracked by CloudFormation      
                                                                                           
                                                                                           
┌────────────────────────┐                               ┌────────────────────────────────┐
│                        │                               │                                │
│  HostedConfiguration   │◀───2. DELETE !!! FAIL !!!─────│      HostedConfiguration       │
│                        │                               │                                │
└────────────────────────┘                               └────────────────────────────────┘
             │                                                                             
             │    ┌─────────────────┐                                                      
             │    │                 │                                                      
             ├────│   HCVersion 1   │                                                      
             │    │                 │                                                      
             │    └─────────────────┘                                                      
             │                                                                             
             │                                                                             
             │    ┌─────────────────┐                    ┌────────────────────────────────┐
             │    │                 │                    │                                │
             └────│   HCVersion 2   │◀───── 1. DELETE ───│   HostedConfigurationVersion   │
                  │                 │                    │                                │
                  └─────────────────┘                    └────────────────────────────────┘
```

chenjane-dev:
> Why do we have RETAIN on these resources in the first place? Do we know?
> 
> Are we sure using a different policy for stack updates and stack deletes is the right solution? It feels to me that either we need to retain them in both cases, for some reason, or we need to retain them in neither case.
> 
> Your current change will fix the following sequence of events:
> 
>     1. Deploy
> 
>     2. Destroy
> 
> 
> But I'm pretty sure it will fail in the following sequence of events:
> 
>     1. Deploy
> 
>     2. Update causing a replacement
> 
>     3. Destroy
> 
> 
> Here's my motivation for questioning this solution: if we had the following sequence of events:
> ## 1. Deploy using HostedConfiguration and HostedConfigurationVersion
> 
> The service state will look like this:
> 
> ```
>                                                                                            
>                  State of service                     State tracked by CloudFormation      
>                                                                                            
>                                                                                            
> ┌────────────────────────┐                               ┌────────────────────────────────┐
> │                        │                               │                                │
> │  HostedConfiguration   │◀──────────────────────────────│      HostedConfiguration       │
> │                        │                               │                                │
> └────────────────────────┘                               └────────────────────────────────┘
>              │                                                                             
>              │    ┌─────────────────┐                    ┌────────────────────────────────┐
>              │    │                 │                    │                                │
>              └────│   HCVersion 1   │◀───────────────────│   HostedConfigurationVersion   │
>                   │                 │                    │                                │
>                   └─────────────────┘                    └────────────────────────────────┘
> ```
> 
> ## 2. Update HostedConfigurationVersion causing a replacement
> 
> The old Version will be RETAINed, because its `UpdateReplacePolicy` will be set to RETAIN.
> 
> The service will end up 2 versions, of which CloudFormation will be tracking the latest.
> 
> ```
>                                                                                            
>                  State of service                     State tracked by CloudFormation      
>                                                                                            
>                                                                                            
> ┌────────────────────────┐                               ┌────────────────────────────────┐
> │                        │                               │                                │
> │  HostedConfiguration   │◀──────────────────────────────│      HostedConfiguration       │
> │                        │                               │                                │
> └────────────────────────┘                               └────────────────────────────────┘
>              │                                                                             
>              │    ┌─────────────────┐                                                      
>              │    │                 │                                                      
>              ├────│   HCVersion 1   │                                                      
>              │    │                 │                                                      
>              │    └─────────────────┘                                                      
>              │                                                                             
>              │                                                                             
>              │    ┌─────────────────┐                    ┌────────────────────────────────┐
>              │    │                 │                    │                                │
>              └────│   HCVersion 2   │◀───────────────────│   HostedConfigurationVersion   │
>                   │                 │                    │                                │
>                   └─────────────────┘                    └────────────────────────────────┘
> ```
> 
> ## 3. Delete stack
> 
> CloudFormation will delete the resources it knows about, in reverse dependency order.
> 
> Because CloudFormation doesn't know about `HCVersion 1` anymore, it will not delete it, and so deleting the `HostedConfiguration` will fail.
> 
> ```
>                                                                                            
>                  State of service                     State tracked by CloudFormation      
>                                                                                            
>                                                                                            
> ┌────────────────────────┐                               ┌────────────────────────────────┐
> │                        │                               │                                │
> │  HostedConfiguration   │◀───2. DELETE !!! FAIL !!!─────│      HostedConfiguration       │
> │                        │                               │                                │
> └────────────────────────┘                               └────────────────────────────────┘
>              │                                                                             
>              │    ┌─────────────────┐                                                      
>              │    │                 │                                                      
>              ├────│   HCVersion 1   │                                                      
>              │    │                 │                                                      
>              │    └─────────────────┘                                                      
>              │                                                                             
>              │                                                                             
>              │    ┌─────────────────┐                    ┌────────────────────────────────┐
>              │    │                 │                    │                                │
>              └────│   HCVersion 2   │◀───── 1. DELETE ───│   HostedConfigurationVersion   │
>                   │                 │                    │                                │
>                   └─────────────────┘                    └────────────────────────────────┘
> ```

The reason this is set to retain currently is because this deals with config versions. When a customer creates a new config version, we want to retain any old, existing versions, in the case that the customer later wants to deploy/access an old version. We have had customers ask for this when using our L1 constructs.

Having them as separate removal policies makes sense because in the case of removal, the customer is deleting the entire stack and hence all AppConfig resources in the stack. If the config profile is getting deleted, then it doesn't make sense for the hosted config version to be retained since it only exists under the scope of a config profile. Also, since we have nested resources, keeping the hosted config version set to retain on delete will cause failures when trying to delete the application and config profile in the stack (as mentioned in the issue linked above), which does not make for the best customer experience, as they are trying to delete the entire stack (including all AppConfig resources). Only on update does the retain policy make sense for the user because of the reasoning in the first paragraph.

rix0rrr:
> we want to retain any old, existing versions, in the case that the customer later wants to deploy/access an old version

I see how that makes sense.

Do you understand my example of how RETAIN-on-replace, DELETE-on-delete is going to lead to the exact same problem you're trying to fix though, just one step later?

CloudFormation can only keep track of as many Hosted Config Versions as there are `AWS::AppConfig::HostedConfigVersion` resources in the template. In the current design there is only 1 of those, so CloudFormation can only keep track of 1 HCV in your service. It will therefore only call `DeleteHostedConfigVersion` at most once come delete time.

If you have triggered any in the mean time, those other HCVs will have "leaked", and will cause an error when `DeletedHostedConfig` is called.

To make the `DeleteHostedConfig` call succeed, one of the following must be true:

- `DeleteHostedConfig` must automatically delete all Versions (I'm pretty sure this is not true, otherwise we wouldn't have seen the reported error).
- HostedConfigVersion replacements can and will never happen. That means there will always be at most one Version, and that's the one that will be deleted (I'm not sure whether or not this is true, I'm guessing it's not).
- We must have as many `AWS::AppConfig::HostedConfigVersion` resources in the CloudFormation template as there are versions in the HostedConfig. That's the only way to make CloudFormation clean them up correctly. This could be possible, but may need a redesign of the construct, and might become awkward to use (I haven't looked at the API or the usage patterns yet).
- We need some out-of-band way to clean up all the HostedConfigVersions come delete time. Probably a custom resource, the same way we clean out an S3 bucket before calling `DeleteBucket`.

chenjane-dev:
Ah, yes I see what you're saying. I will get back to you on this.

rix0rrr:
Turning this into a Draft for now to signal that it's not in a state that needs review.

keenangraham:
> > we want to retain any old, existing versions, in the case that the customer later wants to deploy/access an old version
> 
> I see how that makes sense.
> 
> Do you understand my concern of how RETAIN-on-replace, DELETE-on-delete is going to lead to the exact same problem you're trying to fix though, just one step later?
> 
> CloudFormation can only keep track of as many Hosted Config Versions as there are `AWS::AppConfig::HostedConfigVersion` resources in the template. In the current design there is only 1 of those, so CloudFormation can only keep track of 1 HCV in your service. It will therefore only call `DeleteHostedConfigVersion` at most once come delete time.
> 
> If you have triggered any replacements in the mean time, those other HCVs will have "leaked", and will cause an error when `DeletedHostedConfig` is called.
> 
> To make the `DeleteHostedConfig` call succeed, one of the following must be true:
> 
> * `DeleteHostedConfig` must automatically delete all Versions (I'm pretty sure this is not true, otherwise we wouldn't have seen the reported error).
> * HostedConfigVersion replacements can and will never happen. That means there will always be at most one Version, and that's the one that will be deleted (I'm not sure whether or not this is true, I'm guessing it's not).
> * We must have as many `AWS::AppConfig::HostedConfigVersion` resources in the CloudFormation template as there are versions in the HostedConfig. That's the only way to make CloudFormation clean them up correctly. This could be possible, but may need a redesign of the construct, and might become awkward to use (I haven't looked at the API or the usage patterns yet).
> * We need some out-of-band way to clean up all the HostedConfigVersions come delete time. Probably a custom resource, the same way we clean out an S3 bucket before calling `DeleteBucket`.


> Ah, yes I see what you're saying. I will get back to you on this.

Did this get resolved? The PR says merged but it doesn't look like any AppConfig code added.

Seems like need a way to clean out all HostedConfigVersions when deleting a stack (like S3 bucket and files).

A little confused how the underlying CloudFormation is supposed to work here, since it only tracks and tries to delete the version it deployed, even though the whole point of AppConfig is to make quick deployments in the console by flipping a toggle (without need for CF deployment)? But if you make any changes in the console, the CF stack will fail to delete because of untracked versions.

PR #26831, opened 2023-08-21
chore(region-info): sort tables and remove region name comments

SankyRed:
**Code Changes:**

- Refactoring the fact-tables.ts file to make the entire file to be set alphabetically.
- Helps while adding new region, looking up values for already set regions and avoid duplication.

Closes #26826.

----

### DIFF
--- a/packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
+++ b/packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
@@ -122,14 +122,6 @@ export const PARTITION_MAP: { [region: string]: Region } = {
   'us-isob-': { partition: Partition.UsIsoB, domainSuffix: 'sc2s.sgov.gov' },
 };
 
-export const CR_DEFAULT_RUNTIME_MAP: Record<Partition, string> = {
-  [Partition.Default]: 'nodejs18.x',
-  [Partition.Cn]: 'nodejs18.x',
-  [Partition.UsGov]: 'nodejs18.x',
-  [Partition.UsIso]: 'nodejs18.x',
-  [Partition.UsIsoB]: 'nodejs18.x',
-};
-
 // https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-logging-bucket-permissions
 // https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-access-logs.html#attach-bucket-policy
 // Any not listed regions use the service principal "logdelivery.elasticloadbalancing.amazonaws.com"


packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
```diff
@@ -39,36 +70,67 @@ export const AWS_CDK_METADATA = new Set([
  * @see https://docs.aws.amazon.com/general/latest/gr/s3.html#s3_region
  */
 export const ROUTE_53_BUCKET_WEBSITE_ZONE_IDS: { [region: string]: string } = {
+  // Africa (Cape Town)
```
rix0rrr, line 73: I'm sure this was all a great lot of effort to put all the comments in here... but why? 

In all honestly, I don't see the value this adds.

rix0rrr (changes requested):
Please explain to me what changed and why, and why it's now more "consistent" ?

packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
SankyRed: When looked up the fact-tables file, it looked like some fields were setup alphabetically while for some contributions from open community added comments specifying the regions. So, wanted to make it consistent throughout which will help when we want to add a new region or find the missing values for any region when looked up.

mrgrain: Alphabetization seems definitely like a good idea. 👍🏻

SankyRed: yes, added comments for regions because `CLOUDWATCH_LAMBDA_INSIGHTS_ARNS` section in the fact-tables file has changes specifying regions so added those to make it look clean. But, will get those out as discussed.

packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
```diff
@@ -896,7 +696,7 @@ const ADOT_LAMBDA_LAYER_JAVA_SDK_ARNS: { [version: string]: { [arch: string]: {
       'eu-west-1': 'arn:aws:lambda:eu-west-1:901920570463:layer:aws-otel-java-wrapper-amd64-ver-1-24-0:1',
       'eu-west-2': 'arn:aws:lambda:eu-west-2:901920570463:layer:aws-otel-java-wrapper-amd64-ver-1-24-0:1',
       'eu-west-3': 'arn:aws:lambda:eu-west-3:901920570463:layer:aws-otel-java-wrapper-amd64-ver-1-24-0:1',
-      'ma-central-1': 'arn:aws:lambda:me-central-1:901920570463:layer:aws-otel-java-wrapper-amd64-ver-1-24-0:1',
+      'me-central-1': 'arn:aws:lambda:me-central-1:901920570463:layer:aws-otel-java-wrapper-amd64-ver-1-24-0:1',
```
rix0rrr, line 699: Nice!

packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
```diff
@@ -696,106 +538,64 @@ export const FIREHOSE_CIDR_BLOCKS: { [region: string]: string } = {
...
+      'us-east-1': 'arn:aws:lambda:us-east-1:177933569100:layer:AWS-Parameters-and-Secrets-Lambda-Extension-Arm64:4',
+      'us-east-2': 'arn:aws:lambda:us-east-2:590474943231:layer:AWS-Parameters-and-Secrets-Lambda-Extension-Arm64:4',
+      'us-west-1': 'arn:aws:lambda:us-west-1:997803712105:layer:AWS-Parameters-and-Secrets-Lambda-Extension-Arm64:1',
+      'us-west-2': 'arn:aws:lambda:us-west-2:345057560386:layer:AWS-Parameters-and-Secrets-Lambda-Extension-Arm64:4',
     },
   },
 };
 
 const ADOT_LAMBDA_LAYER_JAVA_SDK_ARNS: { [version: string]: { [arch: string]: { [region: string]: string } } } = {
-  '1.28.1': {
```
rix0rrr, line 757: I see we're removing a bunch of ADOT Lambda layer ARNs.

Is that on purpose?

rix0rrr (changes requested):

packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
SankyRed: Good catch, got removed during refactor. Added them back. Thanks @rix0rrr

packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
```diff
@@ -122,6 +122,14 @@ export const PARTITION_MAP: { [region: string]: Region } = {
   'us-isob-': { partition: Partition.UsIsoB, domainSuffix: 'sc2s.sgov.gov' },
 };
 
+export const CR_DEFAULT_RUNTIME_MAP: Record<Partition, string> = {
```
rix0rrr, line 125: I'm pretty sure this just got removed, and this PR is adding it back in.

rix0rrr (changes requested):

packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
SankyRed: Compared the entire file and did the clean up.

rix0rrr (approved):

PR #26852, opened 2023-08-23
refactor(eks): exposed chart, repository, version and chartAsset props for HelmChart construct

### LINKED ISSUE #26678: aws-eks: Add properties from IHelmChartOptions to HelmChart construct
### Describe the feature

I would like the helm chart properties (at least: chart, repository, version) to be available in the HelmChart construct after instantiation.

### Use Case

I was looking at writing an aspect that I wanted to use to inspect all helm charts added in our CDK stacks

### Proposed Solution

Add and populate the properties in the HelmChart 
```
export class HelmChart extends Construct {
  public abstract readonly chart: string;
  public abstract readonly repository: string;
  public abstract readonly version: string;
  public abstract readonly chartAsset: Asset;
  ....
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.90.0

Windows

davidcosc:
> This change was made to improve convenience when working with the HelmChart construct.
> For example this allows for simpler creation of aspects for inspecting helm charts as mentioned in #26678 .

Closes #26678.

----


packages/aws-cdk-lib/aws-eks/lib/helm-chart.ts
```diff
@@ -108,10 +108,21 @@ export class HelmChart extends Construct {
    * The CloudFormation resource type.
    */
   public static readonly RESOURCE_TYPE = 'Custom::AWSCDK-EKS-HelmChart';
+  public readonly chart: string | undefined;
+  public readonly repository: string | undefined;
+  public readonly version: string | undefined;
+  public readonly chartAsset: Asset | undefined;
```
lpizzinidev, line 114: I think it's better to match the [HelmChartOptions](https://github.com/aws/aws-cdk/blob/7af20babac091dfa39bbeb625069799c0f002a76/packages/aws-cdk-lib/aws-eks/lib/helm-chart.ts#L11) interface declaration a change these to:
```
public readonly chart?: string;
...
```

lpizzinidev (changes requested):
Thanks for the contribution!
Looks good overall to me, I left you a comment for a small change on variable declaration.

packages/aws-cdk-lib/aws-eks/lib/helm-chart.ts
davidcosc: Will do

davidcosc:
Thanks for the fast review! I added another commit matching the type definitions to the ones used by the HelmChartOptions.

colifran (dismissed):
@davidcosc This looks good. Thanks for your contribution!

colifran (approved):

PR #26854, opened 2023-08-23
chore(aws-cdk-lib): increase `jest` test speed, remove warnings

rix0rrr:
Two changes to the `jest` tests:

- We are currently running the tests directly from `.ts` source files using `ts-jest`. Disable type checking (by setting `isolatedModules: true`) to improve the speed. This brings down running 4 test suites from an arbitrary service (`aws-servicecatalog`) down from `79s` to `19s` on my machine.
- Change the "coverage ignore files" a little to get rid of the "Note: The code generator has deoptimised the styling" errors produced by jest when it wants to read the coverage of the generated source files.

----


mrgrain (approved):

rix0rrr:
This would be great but it doesn't actually work. It will always prefer `.js` files over `.ts` files, and there's no way to convince it otherwise.

Given that we do in-place compilation, this is going to mess with our workflow too much.

PR #26858, opened 2023-08-23
fix(logs): LogRetention resources fail with rate exceeded errors

### LINKED ISSUE #26837: CDK deploy: Lambda LogRetention resources fail with rate exceeded errors
### Describe the bug

Deploying multiple stacks with a lot of Lambda functions, results in a lot of CloudWatch Log groups being created by CDK. CDK provides retry options through the [logRetentionRetryOptions](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.LogRetentionRetryOptions.html) property of a Lambda Function construct to prevent rate exceeded errors from happening. 

Until `CDK 2.89.0` this worked fine. In `CDK 2.90.0`, the retry implementation changed in [this PR](https://github.com/aws/aws-cdk/pull/26609/files#diff-d55347d6fbeca6f583a04b79dab153dc73d1756203544571136caa03f4e014fdR187) and the rate exceeded errors occur again.

### Expected Behavior

The retry mechanism should just function properly, same way it did in older, `CDK 2.89` and lower, versions. Stacks should deploy without issues.

### Current Behavior

Stacks cannot be deployed with `CDK 2.90.0` and up as they error with a `Rate exceeded` error:
```
LogRetentionRateLimitStack3 |  57/183 | 11:34:35 AM | CREATE_FAILED        | Custom::LogRetention        | hello_11/LogRetention (hello11LogRetention0EC20DD0) Received response status [FAILED] from custom resource. Message returned: Rate exceeded (RequestId: b84c54c4-3053-487a-82b9-48671904d05a)
```

### Reproduction Steps

I created a small test project to reproduce the issue: https://github.com/jaapvanblaaderen/log-retention-rate-limit. With this simple setup, the issue can be easily reproduced when deploying the stacks in parallel as described in the readme of the repo. I even increased the [amount of retries](https://github.com/jaapvanblaaderen/log-retention-rate-limit/blob/master/lib/log-retention-rate-limit-stack.ts#L24) to `50` but then it still fails.

Testing it against `2.89.0` (which works fine) is possible by checking out the [cdk-289 tag](https://github.com/jaapvanblaaderen/log-retention-rate-limit/releases/tag/cdk-289).

### Possible Solution

Revert the retry mechanism as implemented in `CDK 2.89.0` or troubleshoot why the new retry mechanism doesn't work properly.

I'm actually a bit surprised and disappointed that this issue popped up again. I [fixed this issue myself in CDK](https://github.com/aws/aws-cdk/issues/8257) 3 years ago and now we have this regression on it which causes it to fail.

### CDK CLI Version

2.90.0

### Framework Version

_No response_

### Node.js Version

18

### OS

OSX

### Language

Typescript

### Language Version

_No response_

### Other information

Related issues:
- https://github.com/aws/aws-cdk/issues/8257
- https://github.com/aws/aws-cdk/issues/24485

mrgrain:
The LogRetention Custom Resource used to be able to handle server-side throttling, when a lot of requests to the CloudWatch Logs service are made at the same time.
Handling of this error case got lost during the migration to SDK v3.

If we have (read: a lot) `LogRetention` Custom Resources in a _single_ Stack, CloudFormation apparently applies some internal breaks to the amount of parallelism. For example it appears that resources are batched in smaller groups that need to be completed before the next group is provisioned. And within the groups there appears to be a ever so slight delay between individual resources. Together this is enough to avoid rate limiting in most circumstances.

**Therefore, in practice this issues only occurs when multiple stacks are deployed in parallel.**

To test this scenario, I have added support for `integ-runner` to deploy all stacks of a test case concurrently.
Support for arbitrary command args already existed, but needed to explicitly include the `concurrency` option.

I then create an integration test that deploys 3 stacks à 25 LogRetention resources.
This triggers the error cases described in #26837.

The fix itself is twofold:
- Pass the `maxRetries` prop value to the SDK client to increase the number of attempts of the SDK internal throttling handling. But also enforce a minimum for these retries since they might catch additional retryable failures that our custom outer loop does not account for.
- Explicitly catch `ThrottlingException` errors in the outer retry loop.

Closes #26837

----


packages/aws-cdk-lib/cloud-assembly-schema/lib/integ-tests/commands/deploy.ts
```diff
@@ -94,4 +94,11 @@ export interface DeployOptions extends DefaultCdkOptions {
    * @default false
    */
   readonly ci?: boolean;
+
+  /**
+   * Deploy multiple stacks in parallel
+   *
+   * @default 1
+   */
+  readonly concurrency?: number;
```
rix0rrr, line 103: Is this for a different change?

rix0rrr (approved):

PR #26870, opened 2023-08-24
docs(stepfunctions): multi-state example for map state

### LINKED ISSUE #26707: aws-cdk-lib.stepfunctions: Add Non-trivial Examples to Map Documentation
### Describe the issue

The documentation for the Map constructor currently points to using a single Pass state, however this is far from a real-world scenario. (Without intending to be rude, I'd say this is useless.) Recently I was bitten by a few bugs and lost a few hours thanks to the way I arranged the tasks within Map state.

Consider adding an example or two which show how to add more than one state to a Map, and how to proceed from a Map to the outside. And also how to use the `next` methods in such cases.

### Links

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions-readme.html#map

anshikam:
Closes #26707.

----


packages/aws-cdk-lib/aws-stepfunctions/README.md
```diff
@@ -467,8 +467,27 @@ execute the same steps for multiple entries of an array in the state input.
...
+const choice = new sfn.Choice(this, 'Choice');
+const condition1 = sfn.Condition.stringEquals('$.item.status', 'SUCCESS');
+const step1 = new sfn.Pass(this, 'Step1');
+const step2 = new sfn.Pass(this, 'Step2');
+const finish = new sfn.Pass(this, 'Finish');
+
+const definition = choice
+    .when(condition1, step1)
+    .otherwise(step2)
+    .afterwards())
```
kaizencc, line 487:
```suggestion
    .afterwards()
```

kaizencc (commented):
I think this will help fix the build error. But you can also check locally with `yarn rosetta --strict`. 

When the build succeeds feel free to move this out of draft mode. Thanks for your contribution!

packages/aws-cdk-lib/aws-stepfunctions/README.md
anshikam: Thanks so much @kaizencc for feedback.
I am running into  `JS heap out of memory` issue when I run  `yarn rosetta:extract --strict` (hence having to publish to this draft PR after I make changes and note the build failures). 

Are there any known workarounds? I tried increasing mem size using `NODE_OPTIONS="--max-old-space-size=819` but to no avail

anshikam: Also posted my query on cdk.dev slack for ref https://cdk-dev.slack.com/archives/C041656G340/p1692825834382109

kaizencc: I am not sure, but it is definitely a memory intensive process that may take more memory than `yarn build`. Can you create an issue for it? That's probably the best thing to do here. Thanks for your contribution anyway!

kaizencc (approved):

packages/aws-cdk-lib/aws-stepfunctions/README.md
anshikam: Thanks @kaizencc for approving, will do!

PR #26873, opened 2023-08-24
chore(region-info): Add AppConfig Lambda Layer ARNs

bdgmb2:
This PR adds the AWS AppConfig Lambda Layer ARNs (found [here in the AWS Docs](https://docs.aws.amazon.com/appconfig/latest/userguide/appconfig-integration-lambda-extensions-versions.html)) to the region information. 

Previously this information was only available in the AppConfig Alpha library - I don't think this is the right place to store this information, since the AppConfig Lambda Layer is conceptually different from AppConfig CDK constructs and can be included without using any other constructs.

----


zeyad001:
Hey there. Can this be relooked at. I think the snapshot tests are failing.

PR #26877, opened 2023-08-24
feat(stepfunctions-tasks): `algorithmName` validation for `SageMakerCreateTrainingJob`

tmyoda:
Referencing PR #26675, I have added validation for the `algorithmName` parameter in `SageMakerCreateTrainingJob`.
However, it was suggested that changes for validation should be separated.  So, I have created this PR.

Docs for `algorithmName`:
https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_AlgorithmSpecification.html#API_AlgorithmSpecification_Contents

Exemption Request:
This change does not alter the behavior.
I believe the unit test `create-training-job.test.ts` that I have added is sufficient to test this change.

----


tmyoda:
Exemption Request:
This change does not alter the behavior.
I believe the unit test `create-training-job.test.ts` that I have added is sufficient to test this change.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts
```diff
@@ -324,6 +332,21 @@ export class SageMakerCreateTrainingJob extends sfn.TaskStateBase implements iam
       : {};
   }
 
+  private validateAlgorithmName(algorithmName?: string): void {
+    if (algorithmName === undefined) {
+      return;
+    }
```
rix0rrr, line 338: This also needs check for `Token.isUnresolved()`.

rix0rrr (changes requested):
Almost there!

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts
tmyoda: I added check for `Token.isUnresolved()`

rix0rrr (dismissed):

rix0rrr (approved):

PR #26880, opened 2023-08-25
fix(core): add RetainExceptOnCreate to ParseDeletionPolicy method

### LINKED ISSUE #26863: cfn-parse.ts: parseDeletionPolicy is missing CreateExceptOnDelete
### Describe the bug

CFN recently added support for a new [deletion policy called RetainExceptOnCreate](https://aws.amazon.com/about-aws/whats-new/2023/07/aws-cloudformation-deletion-policies-dev-test-cycle/). This policy was added to the removal-policy enum in [e0d5ca](https://github.com/aws/aws-cdk/pull/26602/commits/e0d5cad96d982c531bb19788939cede36c133e5c), however it is missing from the [switch/case block in `parseDeletionPolicy`](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/helpers-internal/cfn-parse.ts#L468), which causes errors when this removal policy is used.

EDIT: It looks like this code may only be executed when using `CfnInclude` to create a stack from an existing CloudFormation template. Updating the rest of this issue to be more clear about that.

### Expected Behavior

You should be able to use `CfnInclude` to import an existing cloudformation template with valid usage of `DeletionPolicy: RetainExceptOnCreate` without any errors

### Current Behavior

If you define a stack resource with `DeletionPolicy: RetainExceptOnCreate`, then use `CfnInclude` to import that stack you'll get this error:

```
Error: Unrecognized DeletionPolicy 'RetainExceptOnCreate'
```

### Reproduction Steps

* Create Cloudformation template with a stack resource with `DeletionPolicy: RetainExceptOnCreate`
* Import that template as a cdk stack using [`CfnInclude`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.cloudformation_include.CfnInclude.html)

### Possible Solution

* Add a case for `RetainExceptOnCreate`
* It looks like[ cfn-parse uses this same parsing function for both](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/helpers-internal/cfn-parse.ts#L356) `deletionPolicy` and `updateReplacePolicy`. It looks like the new `RetainExceptOnCreate` policy is not a valid `updateReplacePolicy`. There will need to be two separate parsing functions.

_No response_

### CDK CLI Version

2.91.0 

### Framework Version

_No response_

### Node.js Version

v14.21.3

### OS

AL2

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

dineshkatamneni:

CFN recently added support for a new [deletion policy called RetainExceptOnCreate](https://aws.amazon.com/about-aws/whats-new/2023/07/aws-cloudformation-deletion-policies-dev-test-cycle/). This policy was added to the removal-policy enum in [e0d5ca](https://github.com/aws/aws-cdk/pull/26602/commits/e0d5cad96d982c531bb19788939cede36c133e5c), however it is missing from the [switch/case block in parseDeletionPolicy](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/helpers-internal/cfn-parse.ts#L468), which causes errors when this removal policy is used.

Closes #26863.

----


dineshkatamneni:
Exemption Request: it's a small change, and the function doesn't have any test cases before.

rix0rrr (approved):

PR #26883, opened 2023-08-25
feat(rds): support certificate autority certificate

### LINKED ISSUE #26865: (RDS): Support RDS certificate authority
### Latest Update from CDK Team:

Still missing support for reader/writers on cluster. See:
https://github.com/aws/aws-cdk/issues/26865#issuecomment-1716153176

### Describe the feature

Since cloudformation supports changing the rds certificate authority CDK should do so as well. aws-cloudformation/cloudformation-coverage-roadmap#211.

### Use Case

My current CA is rds-ca-2019 and I need to change it. All our AWS resources are scripted as IaC. Therfore, I do not want to change that via the AWS Console but use CDK instead.

### Proposed Solution

Add a property supporting the CA

### Other Information

To contribute a PR for this I probably would need some guidance since I this would be my first contribution to the CDK repository

### Acknowledgements

### CDK version used

2.91.0

macOS 13.5.1

lpizzinidev:
Exposes the `caCertificate` property for an RDS instance to allow specifying a custom CA identifier using the `CaCertificate` class.

Usage:
```
new DatabaseInstance(stack, 'Instance', {
      engine:DatabaseInstanceEngine.mysql({ version: rds.MysqlEngineVersion.VER_8_0_30 }),
      instanceType: ec2.InstanceType.of(ec2.InstanceClass.BURSTABLE3, ec2.InstanceSize.SMALL),
      vpc,
      caCertificate: CaCertificate.RDS_CA_RSA2048_G1,
});
```

Closes #26865.

----

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1139,6 +1139,7 @@ Most build issues can be solved by doing a full clean rebuild:
 
 ```shell
 $ git clean -fqdx .
+$ yarn install
 $ yarn build
 ```
 


packages/aws-cdk-lib/aws-rds/lib/instance.ts
```diff
@@ -713,6 +738,20 @@ export interface DatabaseInstanceNewProps {
...
+   * Specifying or updating this property triggers a reboot.
+   *
+   * For RDS DB engines:
+   * @see https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html
+   * For Aurora DB engines:
+   * @see https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.SSL-certificate-rotation.html
+   *
+   * @default - rds-ca-2019
+   */
+  readonly caCertificateIdentifier?: CertificateIdentifier;
```
mrgrain, line 754:
```suggestion
  readonly caCertificate?: CertificateIdentifier;
```

Can you drop the `Identifier` everywhere please. It's not really relevant to the user intent.
Also in the docstring, e.g. `The the CA certificate used for this DB instance.`

packages/aws-cdk-lib/aws-rds/lib/instance.ts
```diff
@@ -347,6 +347,31 @@ export enum NetworkType {
   DUAL = 'DUAL'
 }
 
+/**
+ * The CA identifier of the DB instance
+ */
+export enum CertificateIdentifier {
```
mrgrain, line 353:
```suggestion
export enum CaCertificate {
```

mrgrain, line 353: Can you turn the enum into an enum like class instead, with the ability to create a CaCertificate from a custom string: `new CaCertificate('future-identifier')`

```ts
class CaCertificate {
  static public RDS_CA_2019() { return new CaCertificate('rds-ca-2019') }
  public constructor(identifier: string) { }
}
```

mrgrain (changes requested):
Nice one, thank you. Got just a couple requests.

robertd:
@mrgrain @lpizzinidev I'm worming up to enum-like class approach. Would this approach be better than getters & setters? In the code we'd use `CaCertificate.RDS_CA_2019` compared to `CaCertificate.rdsCaRsa2019()` which feels more like invoking an enum. Thoughts?

```ts
export class CaCertificate {
  /**
   * rds-ca-2019 certificate authority
   */
  public static readonly RDS_CA_2019 = CaCertificate.of('rds-ca-2019');

  /**
   * rds-ca-ecc384-g1 certificate authority
   */
  public static readonly RDS_CA_ECC384_G1 = CaCertificate.of('rds-ca-ecc384-g1');

  /**
   * rds-ca-rsa2048-g1 certificate authority
   */
  public static readonly RDS_CA_RDS2048_G1 = CaCertificate.of('rds-ca-rsa2048-g1');

  /**
   * rds-ca-rsa4096-g1 certificate authority
   */
  public static readonly RDS_CA_RDS4096_G1 = CaCertificate.of('rds-ca-rsa4096-g1');

  /**
   * Custom certificate ID
   * @param version custom certificate authority ID
   */
  public static of(version: string) { return new CaCertificate(version); }

  /**
   *
   * @param version certificate authority ID
   */
  private constructor(public readonly version: string) { }
}
```

mrgrain:
@robertd Thanks for the review! Yes this is exactly what I had in mind. Apologies for not conveying this in my review.

I think in general it's the better pattern because it enables us to easily extend features in future. E.g. what if (theoretically)  you could upload your own CA Certificate in future. It's more strongly typed and unlike enums can be extended on by userland code. In the wider TS community, there's some talk at the moment about how enums are bad in general.

So yes, we probably should use enum-like classes everywhere instead.

In practice, it might more relevant for some situations than others. I picked it up here, because historically with certificates there is ALWAYS going to be a later one.

robertd:
@mrgrain Yep, makes total sense in this use case.  If @lpizzinidev can change enum like class to use `CaCertificate.RDS_CA_2019` instead of `CaCertificate.rdsCaRsa2019()` (and update stuff in other places accordingly) then I think we’re good to ship this. 👍 


> ```ts
> export class CaCertificate {
>   /**
>    * rds-ca-2019 certificate authority
>    */
>   public static readonly RDS_CA_2019 = CaCertificate.of('rds-ca-2019');
> 
>   /**
>    * rds-ca-ecc384-g1 certificate authority
>    */
>   public static readonly RDS_CA_ECC384_G1 = CaCertificate.of('rds-ca-ecc384-g1');
> 
>   /**
>    * rds-ca-rsa2048-g1 certificate authority
>    */
>   public static readonly RDS_CA_RDS2048_G1 = CaCertificate.of('rds-ca-rsa2048-g1');
> 
>   /**
>    * rds-ca-rsa4096-g1 certificate authority
>    */
>   public static readonly RDS_CA_RDS4096_G1 = CaCertificate.of('rds-ca-rsa4096-g1');
> 
>   /**
>    * Custom certificate identifier
>    * @param custom certificate authority identifier
>    */
>   public static of(identifier: string) { return new CaCertificate(identifier); }
> 
>   /**
>    *
>    * @param certificate authority identifier
>    */
>   private constructor(public readonly identifier: string) { }
> }
> ```

robertd:
@lpizzinidev is on the top of the game with this PR with updating it to the latest changes we discussed. LGTM. 👍

packages/aws-cdk-lib/aws-rds/lib/instance.ts
```diff
@@ -347,6 +347,45 @@ export enum NetworkType {
   DUAL = 'DUAL'
 }
 
+/**
+ * The CA certificate used for this DB instance.
```
robertd, line 351: One minor nitpick… add `@see: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html` in jsdocs section.

lpizzinidev:
@robertd Thanks for the feedback.
I missed the blueprint used for enum-like classes in the rest of the project.
Now the code is much cleaner 👍

packages/aws-cdk-lib/aws-rds/lib/instance.ts
```diff
@@ -713,6 +754,20 @@ export interface DatabaseInstanceNewProps {
...
+   * The identifier of the CA certificate for this DB instance.
+   *
+   * Specifying or updating this property triggers a reboot.
+   *
+   * For RDS DB engines:
+   * @see https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html
+   * For Aurora DB engines:
+   * @see https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.SSL-certificate-rotation.html
+   *
+   * @default - CaCertificate.RDS_CA_2019
```
mrgrain, line 768: This default is unexpected and seems like a potentially breaking change to me. Can you explain or remove it?

packages/aws-cdk-lib/aws-rds/lib/instance.ts
```diff
@@ -865,6 +920,7 @@ abstract class DatabaseInstanceNew extends DatabaseInstanceBase implements IData
       domain: this.domainId,
       domainIamRoleName: this.domainRole?.roleName,
       networkType: props.networkType,
+      caCertificateIdentifier: props.caCertificate ? props.caCertificate.identifier : CaCertificate.RDS_CA_2019.identifier,
```
mrgrain, line 923: Default question, see above.

```suggestion
      caCertificateIdentifier: props.caCertificate ? props.caCertificate.identifier : undefined,
```

Btw, we also could implement a `toString()` method on `CaCertificate`. This is works as well though.

mrgrain (changes requested):
One last question regarding the default.

packages/aws-cdk-lib/aws-rds/lib/instance.ts
lpizzinidev: Trying to create an instance from the console, it suggests `rds-ca-2019` as default (this is the reason for the `@default`).

Screenshot 2023-08-28 at 15 53 43

Anyway, looking at the CFN [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-cacertificateidentifier), it does not specify any default, so I guess I should remove it.

mrgrain: I see. It probably doesn't change much, but I think we want to end up in this case:
"If you don't select a certificate authority, RDS chooses one for you."

```suggestion
   * @default - Let RDS chooses a certificate authority for you
```

mrgrain (approved):

robertd (approved):
LGTM. Many thanks @lpizzinidev 🚀

mrgrain (approved):

StefanPrintezis:
static .of() usually refers to a known given, like an enum. Why would the constructor be private and then publish the of( method to be  a constructor wrapper, it's a bit inverse right now. If you want a custom CA i'd think is fine `new PrivateCa('')` , like InstanceType.

I think I'd still prefer to go with the enums for consistency. Feels more light weight than having the instances be instantiated in properties of a static class.


```
export declare enum CertificateIdentifier {
    /**
     * rds-ca-2019 certificate authority
     */
    RDS_CA_2019 = 'rds-ca-2019',
    /**
     * rds-ca-ecc384-g1 certificate authority
     */
    RDS_CA_ECC384_G1 = 'rds-ca-ecc384-g1',
    /**
     * rds-ca-rsa2048-g1 certificate authority
     */
    RDS_CA_RDS2048_G1 = 'rds-ca-rsa2048-g1',
    /**
     * rds-ca-rsa4096-g1 certificate authority
     */
    RDS_CA_RDS4096_G1 = 'rds-ca-rsa4096-g1',
}
/**
 * The CA certificate used for this DB instance.
 *
 * This class takes a literal string, good if you already
 * know the identifier of the type you want.
 */
export declare class CaCertificate {

    static of(identifier: CertificateIdentifier): CaCertificate {
        return new CaCertificate(identifier);
    }
    constructor(private readonly identifier: string) {};

    /**
     * Return the instance type as a string.
     */
    toString(): string {
        return this.identifier;
    }
}

```

mrgrain:
> static .of() usually refers to a known given, like an enum. Why would the constructor be private and then publish the of( method to be a constructor wrapper, it's a bit inverse right now. If you want a custom CA i'd think: `new PrivateCa('custom-ca')`
> 
> I think I'd still prefer to go with the enums for consistency. Feels more light weight than having the instances be instantiated in properties of a static class.

But using it would still be typed to `CaCertificate`, so as a user you will have to do:

```ts
new CaCertificate(CertificateIdentifier. RDS_CA_2019)
```

and instantiate anyway.

`static of(identifier: string)` is the pattern we use right now everywhere else.

mrgrain:
@mergifyio requeue

baguilarq:
do we need another implementation to support this for `DatabaseCluster`, isn't it?

PR #26896, opened 2023-08-26
feat(s3-deployment): specify lambda execution role in DeployTimeSubstitutedFile

### LINKED ISSUE #26879: DeployTimeSubstitutedFile: should allow passing IAM role
### Describe the feature

DeployTimeSubstitutedFile is an extension of BucketDeployment that allows users to upload individual files and specify to make substitutions in the file. 

Our CDK assets bucket is encrypted and we need to grant the `DeployTimeSubstitutedFile` access to our KMS key. In the past we have used `BucketDeployment` and passed a role via the `BucketDeployment` props with right set of permissions.

Reference: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3_deployment.BucketDeployment.html#role 

### Use Case

Fetch assets from encrypted KMS S3 Buckets

### Proposed Solution

Expose the role as parameter within `DeployTimeSubstitutedFileProps` and pass it down to `BucketDeployment` parent

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.93.0

Mac

lpizzinidev:
Exposes the `role` property in `DeployTimeSubstitutedFile`.
The role will be passed to the generated `BucketDeployment` lambda function.

Closes #26879.

----


mrgrain (approved):
Nice!

PR #26898, opened 2023-08-26
fix(ec2): networkAclName property for NetworkAcl does not work

### LINKED ISSUE #26897: ec2: networkAclName property in NetworkAcl construct does not work
### Describe the bug

NetworkAcl L2 construct has networkAclName property.
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.NetworkAcl.html

This is only assigned to physicalName here.
https://github.com/aws/aws-cdk/blob/724bd01b8870888ed45783fc952b71cd89408c4e/packages/aws-cdk-lib/aws-ec2/lib/network-acl.ts#L115-L117

However, we can't define physical name for `AWS::EC2::NetworkAcl` resource. The physical name for Network ACL is always automatically generated id.

So, currently networkAclName property is meaningless property. This issue is similar to #26370.

### Expected Behavior

In VPC console, a tag value for Name key is shown in Name column as below.
vpc-console-acl-name

So, NetworkAcl L2 construct should add Name tag like Vpc or FlowLog L2 construct.
https://github.com/aws/aws-cdk/blob/724bd01b8870888ed45783fc952b71cd89408c4e/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts#L1457
https://github.com/aws/aws-cdk/blob/724bd01b8870888ed45783fc952b71cd89408c4e/packages/aws-cdk-lib/aws-ec2/lib/vpc-flow-logs.ts#L736

### Current Behavior

The value of networkAclName property is never referred in generated CloudFormation template.

### Reproduction Steps

We can reproduce it with the following snippet.

```typescript
    const vpc = ec2.Vpc.fromLookup(this, 'Vpc', {
      isDefault: true,
    });

    new ec2.NetworkAcl(stack, 'MyNetworkAcl', {
      vpc: vpc,
      networkAclName: 'CustomNetworkAclName',
  });
```

### Possible Solution

Adding Name tag with the value of networkAclName property.

_No response_

### CDK CLI Version

v2.93.0

### Framework Version

_No response_

### Node.js Version

v18.17.1

### OS

Amazon Linux 2

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

tam0ri:
Currently, networkAclName property for NetworkAcl L2 is never referred in generated CloudFormation template. This PR solves the issue by adding Name tag with the value of networkAclName property. This fix is similar to #26371

Closes #26897

----

### DIFF
--- a/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts
+++ b/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts
@@ -4,7 +4,7 @@ import { renderEnvironment, renderTags } from './private/utils';
 import * as ec2 from '../../../aws-ec2';
 import * as iam from '../../../aws-iam';
 import * as sfn from '../../../aws-stepfunctions';
-import { Duration, Lazy, Size, Stack } from '../../../core';
+import { Duration, Lazy, Size, Stack, Token } from '../../../core';
 import { integrationResourceArn, validatePatternSupported } from '../private/task-utils';
 
 /**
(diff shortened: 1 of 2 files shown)


kylelaker (approved):
Thanks for the fix and the link to the similar PR!

rix0rrr (approved):

rix0rrr:
Thank you!

rix0rrr (approved):

PR #26910, opened 2023-08-28
fix(cli): asset not uploaded with different synthesizer configs

### LINKED ISSUE #25927: DefaultStackSynthesizer: Using a StackSynthesizer and Sharing Assets will only Publish One Asset and Not the Other
### Describe the bug

If a Synthesizer is used and the an Asset is shared between two Stacks on the same App, one of the Stacks will have a published asset to reference while the other Stack will not.

### Expected Behavior

If a Synthesizer is used and the prefixes are different, the Assets should be able accessible to both Stacks.

### Current Behavior

One Stack will be able to access the asset, while the other will return a `Resource handler returned message: "Error occurred while GetObject. S3 Error Code: NoSuchKey. S3 Error Message: The specified key does not exist. ` due to a missing asset.

### Reproduction Steps

1. Define a `bin` or `app.py` file that creates two stacks
2. Define a `DefaultStackSynthesizer`
3. Define a `Lambda Function` resource with a `logRetention` prop

### Possible Solution

Since this is unique to version 2.8X.X, the Asset removal might need to be reviewed

_No response_

### CDK CLI Version

2.81.0

### Framework Version

_No response_

### Node.js Version

18.16.0

### OS

macOS 12.6.6

### Language

Typescript, Python, .NET, Java, Go

### Language Version

5.0.4

### Other information

_No response_

rix0rrr:
If the same asset is used in 2 stacks that use different synthesizer configurations for publishing (for example, by using a different prefix) the asset will only be uploaded once instead of twice.

We used to make the assumption that it was okay to use the destination ID as token of uniqueness. This is true inside a single manifest, but does not hold when there is more than stack that each have a manifest: both may have the destination ID `current_account:current_region`, but have different parameters for each destination.

Instead, we calculate a content hash over the destination definition itself. That way, if the definitions are different we will create different nodes for each of them.

Fixes #25927.

----


packages/aws-cdk/lib/util/work-graph-builder.ts
```diff
@@ -37,27 +38,26 @@ export class WorkGraphBuilder {
...
-  private addAsset(parentStack: cxapi.CloudFormationStackArtifact, assetArtifact: cxapi.AssetManifestArtifact, assetManifest: AssetManifest, asset: IManifestEntry) {
+  private addAsset(parentStack: cxapi.CloudFormationStackArtifact, assetManifestArtifact: cxapi.AssetManifestArtifact, assetManifest: AssetManifest, asset: IManifestEntry) {
     // Just the artifact identifier
     const assetId = asset.id.assetId;
-    // Unique per destination where the artifact needs to go
-    const assetDestinationId = `${asset.id}`;
 
-    const buildId = `${this.idPrefix}${assetId}-build`;
-    const publishNodeId = `${this.idPrefix}${assetDestinationId}-publish`;
+    const buildId = `build-${assetId}-${contentHashAny([assetId, asset.genericSource]).substring(0, 10)}`;
```
iliapolo, line 45: - Why no longer consider `this.idPrefix`? was it not needed here in the first place? 
- Why substringing? 
- Would this trigger a re upload of basically all assets everywhere? thus doubling the amount of assets in the user's account?

rix0rrr: > Why no longer consider `this.idPrefix`? was it not needed here in the first place?

It was an attempt to prevent false sharing between assets in subassemblies and top level assemblies. Because of the hash, this is now no longer necessary (and in fact hampers legitimate sharing).

> Why substringing?

Just to keep the id somewhat reasonable.

> Would this trigger a re upload of basically all assets everywhere? thus doubling the amount of assets in the user's account?

No effect there.

iliapolo (approved):

PR #26928, opened 2023-08-30
docs(redshift): add details on cluster resizing

### LINKED ISSUE #21103: (redshift): document the resizing operation
### Describe the issue

We should add a section to the README that explains how to resize a cluster using CloudFormation. There are two methods, `elastic` and `classic` which both have implications. There are also different scenarios where you would want to use one vs the other.

Useful docs:
https://aws.amazon.com/premiumsupport/knowledge-center/redshift-troubleshoot-elastic-resize/
https://docs.aws.amazon.com/redshift/latest/mgmt/managing-cluster-operations.html#rs-resize-tutorial

### Links

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-redshift-alpha-readme.html

msambol:
I felt it best to keep most of the details in the linked documentation, but I believe this provides a good overview.

Closes #21103.

----


packages/@aws-cdk/aws-redshift-alpha/README.md
```diff
@@ -543,3 +543,19 @@ const cluster = new Cluster(this, 'Redshift', {
 });
 cluster.addIamRole(role);
 ```
+
+## Resizing
+
+As your data warehousing needs change, it's possible to resize your Redshift cluster, but it's important to understand certain caveats.
+Most importantly, if the cluster was deployed via CDK, the cluster should be resized via CDK so it's registered in the AWS CloudFormation template.
+This ensures the `NumberOfNodes` parameter gets updated so the correct type of resize is done. There are two types of resize operations:
```
msambol, line 551: CFN has this as `NumberOfNodes`, CDK has it as `numberOfNodes`. I felt it best to be `NumberOfNodes` to represent the CFN template.

rix0rrr: I think this whole sentence can be removed. Just noting that the resize has to be done via CloudFormation is a good reminder, no need to go into detail. You are saying "certain caveats". Are there more than one?

As for your casing example: as a rule, I would say that CDK documentation should stay in CDK-land, and not presume CloudFormation knowledge. I know this is not entirely achievable always, but that should be our goal. The point is moot though, again, I would remove the sentence.

packages/@aws-cdk/aws-redshift-alpha/README.md
```diff
@@ -543,3 +543,19 @@ const cluster = new Cluster(this, 'Redshift', {
...
+
+* Elastic resize - You can add nodes to or remove nodes from the cluster. Also, you can change the node type, such as from DS2 nodes to RA3 nodes.
+AWS recommmends elastic resize as the first option, as it's a fast operation and typically completes in minutes. Elastic resize is appropriate when you:
+  * Add or reduce nodes in your existing cluster, but don't change the node type.
+  * Change the node type of your cluster.
+
+* Classic resize - Number of nodes, node type, or both, can be changed. This operation takes longer to complete, but is useful when the resize operation
+doesn't meet the criteria of an elastic resize. See [AWS Knowledge Center](https://repost.aws/knowledge-center/redshift-troubleshoot-elastic-resize) for further conditions that require classic resizing.
```
rix0rrr, line 559: It feels a bit like teasing to say "see this other place for the conditions". Can you summarize? Or is it by exclusion, everything that's not an elastic resize is a classic resize?

Also, do people have to do anything to their code to pick one or the other resize? Please explain if they do or do not.

rix0rrr (dismissed):

rix0rrr (approved):
👨‍🍳👌

PR #26930, opened 2023-08-30
fix(firehose): remove unused role during DeliveryStream creation

### LINKED ISSUE #26927: (kinesisfirehose): DeliveryStream creates useless role
### Describe the bug

The DeliveryStream construct always creates a role https://github.com/aws/aws-cdk/blob/724bd01b8870888ed45783fc952b71cd89408c4e/packages/%40aws-cdk/aws-kinesisfirehose-alpha/lib/delivery-stream.ts#L325

However, this role ends up with no permissions, and no reason to exist if no source stream or encryption key are used:

```json
  "DeliveryStreamServiceRole0CF4E414": {
   "Type": "AWS::IAM::Role",
   "Properties": {
    "AssumeRolePolicyDocument": {
     "Statement": [
      {
       "Action": "sts:AssumeRole",
       "Effect": "Allow",
       "Principal": {
        "Service": "firehose.amazonaws.com"
       }
      }
     ],
     "Version": "2012-10-17"
    }
   },
   "Metadata": {
    "aws:cdk:path": "EipFailTagsStack/DeliveryStream/Service Role/Resource"
   }
  },
```

### Expected Behavior

I expect no role to be created

### Current Behavior

An unnecessary role is created

### Reproduction Steps

Create a DeliveryStream without an `encryptionKey`, `encryption`, or `sourceStream` prop.

```ts
    new firehose.DeliveryStream(this, 'DeliveryStream', { 
      destinations: [new destinations.S3Bucket(new s3.Bucket(this, 'Bucket'))]
    });
```

### Possible Solution

In this case, will need to check for appropriate props before creating the role

_No response_

### CDK CLI Version

current

### Framework Version

_No response_

### Node.js Version

16

### OS

mac

### Language

Typescript

### Language Version

_No response_

### Other information

Can workaround for now with the following escape hatch:
```ts
stream.node.tryRemoveChild('Service Role')
```

msambol:
When a DeliveryStream is created without `sourceStream` or `encryptionKey`,
an extra role is being created that is unused. This PR removes creation of that role. 

I also learned that the role created for `encryptionKey` is used "indirectly" for a grant 
put on the KMS key...interesting.

Closes #26927.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
+++ b/packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
@@ -99,10 +99,10 @@ export interface ParameterOptions {
   readonly parameterName?: string;
 
   /**
-   * Indicates of the parameter name is a simple name (i.e. does not include "/"
+   * Indicates if the parameter name is a simple name (i.e. does not include "/"
    * separators).
    *
-   * This is only required only if `parameterName` is a token, which means we
+   * This is required only if `parameterName` is a token, which means we
    * are unable to detect if the name is simple or "path-like" for the purpose
    * of rendering SSM parameter ARNs.
    *
@@ -168,7 +168,7 @@ abstract class ParameterBase extends Resource implements IParameter {
   /**
    * The encryption key that is used to encrypt this parameter.
    *
-   * * @default - default master key
+   * @default - default master key
    */
   public readonly encryptionKey?: kms.IKey;
 
@@ -337,10 +337,10 @@ export interface CommonStringParameterAttributes {
   readonly parameterName: string;
 
   /**
-   * Indicates of the parameter name is a simple name (i.e. does not include "/"
+   * Indicates if the parameter name is a simple name (i.e. does not include "/"
    * separators).
    *
-   * This is only required only if `parameterName` is a token, which means we
+   * This is required only if `parameterName` is a token, which means we
    * are unable to detect if the name is simple or "path-like" for the purpose
    * of rendering SSM parameter ARNs.
    *
--- a/tools/@aws-cdk/spec2cdk/package.json
+++ b/tools/@aws-cdk/spec2cdk/package.json
@@ -32,8 +32,8 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@aws-cdk/aws-service-spec": "^0.0.10",
-    "@aws-cdk/service-spec-types": "^0.0.10",
+    "@aws-cdk/aws-service-spec": "^0.0.11",
+    "@aws-cdk/service-spec-types": "^0.0.11",
     "@cdklabs/tskb": "^0.0.1",
     "@cdklabs/typewriter": "^0.0.1",
     "camelcase": "^6",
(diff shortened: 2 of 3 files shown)


mrgrain (changes requested):
Hmm, this is kind of changing the contract for `DeliveryStream.grantPrincipal`. Yes technically it is still a principle, but now it might be unusable. (Although, `grantPrincipal` 🤨  why isn't this just called `role` - but that's a different story).

Can you change `this.grantPrincipal` to use a getter that will return the role if we already have one and only create the unused role when the property is accessed? That should give us the best of both worlds. What do you think?

msambol:
> Contributor

@mrgrain Thanks for the review! I'm totally up for making that change but wanted to point out one line of code: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-autoscaling/lib/auto-scaling-group.ts#L1274. There seems to be some precedent for this? I'm good either way, let me know what you'd prefer.

mrgrain:
> @mrgrain Thanks for the review! I'm totally up for making that change but wanted to point out one line of code: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-autoscaling/lib/auto-scaling-group.ts#L1274. There seems to be some precedent for this? I'm good either way, let me know what you'd prefer.

Yes there's precedent and if it were a new construct I'd probably lean that way. But because it is a change, I'd like us to try and stay with the previously established contract.

packages/@aws-cdk/aws-kinesisfirehose-alpha/lib/delivery-stream.ts
```diff
@@ -311,21 +311,34 @@ export class DeliveryStream extends DeliveryStreamBase {
...
+    this._role = props.role;
+
     if (props.destinations.length !== 1) {
       throw new Error(`Only one destination is allowed per delivery stream, given ${props.destinations.length}`);
     }
 
-    const role = props.role ?? new iam.Role(this, 'Service Role', {
-      assumedBy: new iam.ServicePrincipal('firehose.amazonaws.com'),
-    });
-    this.grantPrincipal = role;
```
msambol, line 328: I don't think I need this line anymore?

msambol:
This is expected as I'm removing the unused role:
```bash
Stack: aws-cdk-firehose-delivery-stream-s3-all-properties - Resource: DeliveryStreamServiceRole964EEBCC - Impact: WILL_DESTROY
```

packages/@aws-cdk/aws-kinesisfirehose-alpha/lib/delivery-stream.ts
mrgrain: I agree!

mrgrain (approved):

msambol:
@mrgrain Good to :shipit: ? Test failing because it's deleting a resource..

mrgrain:
Good to ship, but we'll need the build passing. Can you run that integration test and update the snapshot?

msambol:
@mrgrain I missed those other ones—apologies!

mrgrain (approved):

Issue #26839, opened 2023-08-22
(awslint): modules inside aws-cdk-lib are not linted anymore


jogold:
### Describe the bug

Modules inside `aws-cdk-lib` are not `awslint`ed anymore. Only `core` is linted.

### Expected Behavior

All code should be linted.

### Current Behavior

Modules exported with `as` are not linted (e.g. `export * as aws_s3 from './aws-s3'`)

### Reproduction Steps

Remove the docstring of a public API element and run `yarn build`, it should fail but passes.

### CDK CLI Version

2.92.0

### Node.js Version

All

### OS

All

### Language

Typescript

PR #26942, opened 2023-08-30
fix(awslint): include types from submodules during linting (WIP)

mikewrighton:
Addresses #26839 

The initial change in this draft modifies some of the linters within the `awslint` tool so that they include types from submodules, when reading from the JSII assembly. The results in a large number of lint errors since it causes L1 constructs to be included incorrectly - this is happening because 1) L1 constructs are submodules in the assembly, and 2) the `CoreTypes.isCfnResource()` function no longer works as it relies on the legacy core module which no longer exists.

Follow-up work:
1. Fix `CoreTypes` and any other code within awslint which references `@aws-cdk/core`
2. Fix [this broken code](https://github.com/aws/aws-cdk/blob/b6f7f390e663cc69b618c43af52ec771ddfc2f41/packages/awslint/lib/rules/cfn-resource.ts#L28-L30) (which becomes active after #1) which relies on a docstring pattern to generate the CFN resource name. Using a tag like  `@cloudformationResource` (need re-adding) is probably a better approach.
3. Any other fixes needed on the linter code paths which are now active.
----


packages/awslint/lib/rules/docs.ts
```diff
@@ -53,7 +53,9 @@ docsLinter.add({
     if (e.ctx.kind !== 'interface-property') { return; }
     if (!e.ctx.containingType.isDataType()) { return; }
     // this rule does not apply to L1 constructs
-    if (CoreTypes.isCfnType(e.ctx.containingType)) { return; }
+    if (CoreTypes.isCfnType(e.ctx.containingType)) {
+      return;
```
mikewrighton, line 57: Changed for debugging purposes - can revert.

Issue #21804, opened 2022-08-29
(lambda): Add property for log removal policy of Lambda function log groups


tmokmss:
### Describe the feature

#21113 introduced `removalPolicy` for `LogRetention` custom resource to allow us to delete log groups inside a stack. However, because we don't have a corresponding property in [Lambda Function construct](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.Function.html#logretention), it seems we still cannot remove a log group for a Lambda function automatically when we delete a stack. (Sorry if I'm missing something)

### Use Case

Automatically remove log groups for lambda functions inside a stack when we delete it. 

### Proposed Solution

Add a property e.g. `logRetentionRemovalPolicy?` here: 
https://github.com/aws/aws-cdk/blob/478b9967f4a5814d36472ecee5fc7157e3c2ad74/packages/%40aws-cdk/aws-lambda/lib/function.ts#L294

The property will be only valid when `logRetention` is set. There might be a better API for this but at least it should work and won't introduce any breaking change :(

### Other Information

An aspect like below will not work either:

```ts
class SetLogGroupRemovalPolicy implements IAspect {
  public visit(node: CfnResource): void {
    if (node.cfnResourceType == 'Custom::LogRetention') {
      node.addOverride('Properties.RemovalPolicy', 'destroy');
    }
  }
}
```

because we still need to configure IAM policy to allow the lambda to delete the log group, which is set here.

https://github.com/aws/aws-cdk/blob/478b9967f4a5814d36472ecee5fc7157e3c2ad74/packages/%40aws-cdk/aws-logs/lib/log-retention.ts#L162-L183

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.39.0

### Environment details (OS name and version, etc.)

macOS

peterwoodworth:
Thanks for submitting this, we accept contributions! Check out our [contributing guide](https://github.com/aws/aws-cdk/blob/master/CONTRIBUTING.md) if you're interested - there's a low chance the team will be able to address this soon but we'd be happy to review a PR 🙂 I think we would want to be able to set this new property even if `logRetention` is not set

tmokmss:
Hi thanks! I agree with that so I'll submit a PR implementing 2nd option.

rittneje:
[...]

PR #21820, opened 2022-08-30
feat(lambda): allow to delete log group of a lambda function when it is removed

tmokmss:
**No more orphaned log groups after your trial and error!**

image

closes #21804

Now we can set `autoDeleteLogGroup: true` to delete an associated Lambda log automatically.

```ts
new lambda.Function(stack, 'AutoDeleteLogGroup', {
  code: new lambda.InlineCode('exports.handler = (event) => console.log(JSON.stringify(event));'),
  handler: 'index.handler',
  runtime: lambda.Runtime.NODEJS_14_X,
  autoDeleteLogGroup: true,
});
```

### Consideration on property removal

If a user remove `autoDeleteLogGroup` property as below, the logRetention resource will be removed from a synthesized template, and it will result in the removal of log group itself. This implicit behavior should be clearly warned in the doc.

```diff
new lambda.Function(stack, 'AutoDeleteLogGroup', {
  code: new lambda.InlineCode('exports.handler = (event) => console.log(JSON.stringify(event));'),
  handler: 'index.handler',
  runtime: lambda.Runtime.NODEJS_14_X,
-  autoDeleteLogGroup: true,
});
```

There was an issue regarding s3 Bucket's `autoDeleteObjects` feature: #16603. We might need to implement a similar approach. It might be difficult since we are not directly managing LogGroup in CFn though.

----

### All Submissions:

* [X] Have you followed the guidelines in our [Contributing guide?](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md)

### Adding new Unconventional Dependencies:

* [ ] This PR adds new unconventional dependencies following the process described [here](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md/#adding-new-unconventional-dependencies)

### New Features

* [X] Have you added the new feature to an [integration test](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md)?
	* [X] Did you use `yarn integ` to deploy the infrastructure and generate the snapshot (i.e. `yarn integ` without `--dry-run`)?

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #26947, opened 2023-08-30
feat(logs): generic service managed log group to replace log retention

mrgrain:
A general purpose implementation for a LogGroup that is usually managed by a service directly.

Supports all features a LogGroup supports.
It looks like a LogGroup, swims like a LogGroup and quacks like a LogGroup.

Replaces https://github.com/aws/aws-cdk/pull/26049 to close https://github.com/aws/aws-cdk/issues/17533
Replaces https://github.com/aws/aws-cdk/pull/21820 to close https://github.com/aws/aws-cdk/issues/21804

Currently missing features, might end up dropping them intentionally
- [ ] logGroupRegion
- [ ] role for custom resource
- [ ] retry config
- [ ] Replace all uses of LogRetention

----

### DIFF
--- a/packages/aws-cdk-lib/aws-logs/lib/service-managed-log-group.ts
+++ b/packages/aws-cdk-lib/aws-logs/lib/service-managed-log-group.ts
@@ -1,0 +1,272 @@
+import * as path from 'path';
+import { Construct, IConstruct } from 'constructs';
+import { BaseLogGroupProps, ILogGroup } from './log-group';
+import { LogGroupBase } from './log-group-base';
+import { validateLogGroupRetention } from './private/util';
+import * as iam from '../../aws-iam';
+import * as s3_assets from '../../aws-s3-assets';
+import * as cdk from '../../core';
+
+const SERVICE_MANAGED_LOG_GROUP_TYPE = 'Custom::ServiceManagedLogGroup';
+const SERVICE_MANAGED_LOG_GROUP_TAG = 'aws-cdk:service-managed-log-group';
+
+/**
+ * Properties for ServiceManagedLogGroup
+ */
+export interface ServiceManagedLogGroupProps extends BaseLogGroupProps {}
+
+/**
+ * Options for binding a ServiceManagedLogGroup to its parent
+ */
+export interface ServiceManagedLogGroupBindOptions {
+  /**
+   * The arn of the log group
+   */
+  readonly logGroupArn: string;
+
+  /**
+   * The resource owning the log group.
+   */
+  readonly parent: IConstruct;
+
+  /**
+   * Configuration for tagging the parent resource
+   */
+  readonly tagging: ServiceManagedLogGroupTaggingConfig;
+}
+
+/**
+ * Tagging config for retrieving tags from the owning resource
+ */
+export interface ServiceManagedLogGroupTaggingConfig {
+  /**
+   * The service managing the log group
+   */
+  readonly service: string;
+  /**
+   * The API action used to retrieve tags
+   */
+  readonly action: string;
+  /**
+   * The request field in dot notation to pass the resource name
+   * @default "Resource"
+   */
+  readonly requestField?: string;
+  /**
+   * The response field in dot notation that will contain the list of tags
+   * @default "Tags"
+   */
+  readonly responseField?: string;
+  /**
+   * Additional permissions given to the custom resource function to query tags
+   * @default "[]"
+   */
+  readonly permissions?: string[];
+}
+
+/**
+ * A CloudWatch Log Group that is created and managed by another service
+ *
+ * With this construct, the Log Group configuration can be changed anyway.
+ * You must call `bind()` on the ServiceManagedLogGroup, to connect it with the parent resource.
+ *
+ * @resource AWS::Logs::LogGroup
+ */
+export class ServiceManagedLogGroup extends LogGroupBase implements ILogGroup {
+  public readonly logGroupArn: string;
+  public readonly logGroupName: string;
+
+  /**
+   * Tags for the LogGroup.
+   */
+  public readonly tags: cdk.TagManager = new cdk.TagManager(cdk.TagType.KEY_VALUE, 'AWS::Logs::LogGroup');
+
+  private provider: ServiceManagedLogGroupFunction;
+  private props: BaseLogGroupProps;
+  private _logGroupArn?: string;
+
+  constructor(scope: Construct, id: string, props: ServiceManagedLogGroupProps = {}) {
+    super(scope, id);
+
+    this.props = {
+      ...props,
+      retention: validateLogGroupRetention(props.retention),
+    };
+
+    this.logGroupArn = cdk.Lazy.string({
+      produce: () => this._logGroupArn + ':*',
+    });
+    this.logGroupName = cdk.Lazy.string({
+      produce: () => cdk.Stack.of(this).splitArn(this._logGroupArn!, cdk.ArnFormat.COLON_RESOURCE_NAME).resourceName,
+    });
+
+    // Custom resource provider
+    this.provider = this.ensureSingletonProviderFunction();
+  }
+
+  /**
+   * Helper method to ensure that only one instance of the provider function resources is in the stack.
+   * Mimicking the behavior of @aws-cdk/aws-lambda's SingletonFunction to prevent circular dependencies.
+   */
+  private ensureSingletonProviderFunction() {
+    const functionLogicalId = 'ServiceManagedLogGroup' + 'f0360f7393ea41069d5f706d30f37fa7';
+    const existing = cdk.Stack.of(this).node.tryFindChild(functionLogicalId);
+    if (existing) {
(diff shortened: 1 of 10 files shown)


humanzz:
Hi @mrgrain,
I came across this a while back through the linked issues a while back but arrived here again as I was checking the latest commits and coming across https://github.com/aws/aws-cdk/pull/28737 and the note abt moving from `LogRetention` to `LogGroup` but needing to use a new log group name.

The change in here seems like a great follow up to actually ease the pain of such migration - as we've accumulated data in the lambda's log group that we wouldn't want to just drop, and this seems like a good option to continue using the same name we currently have.

Is the plan for this change to still happen?

mrgrain:
Hey @humanzz thanks for reaching out. I'm not sure how this PR would ease any migration pain. Could you elaborate on this? To me it feels like this PR would almost contribute to the opposite and reduce pressure to migrate.

Also, you should not lose any log data from the old log groups. But new data would go into the new log group. Depending on how long you typically go back in your logs and your data retention setting, this might be a quite limited transition period. For example if you have a 14days log retention policy, after 14 days you will only have a single log group again.

On a related note if you are storing logs for a really long time, you might want to consider moving long term storage away from CloudWatch Logs, as there are cheaper options available.

humanzz:
So, here's some context and thoughts
- I understand that the new `logGroup` lambda prop can allow one to specify a new log group, and with being careful abt log retention's removal policy, I can easily end up with 2 log groups, the old original one, and the new easily fully managed one via `logs.LogGroup`. I'm not super keen on this option, as there are features in Lambda/Insights/etc. that I guess will simply leverage the lambda declared log group to click through (think from Lambda Console, Lambda Insights Console, etc.). From time to time I do queries over these logs, to identify some historical trends, etc. I'm not saying it's impossible to do with 2 log groups, I'm just saying it becomes less convenient. We keep our logs for 18 months in general based on recommendations from ORRs/Security Reviews. Multiply that by working on a dozen lambda functions, and having 2 log groups for each function to keep that history becomes even more "less inconvenient"
- I've noticed that in your implementation here `class ServiceManagedLogGroup extends LogGroupBase implements ILogGroup`, while [FunctionOptions.logGroup](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L486) is an `ILogGroup`. To me, this suggested that `ServiceManagedLogGroup` can be used as a value for `logGroup` if someone wants to customize the service managed log group, or a concrete `LogGroup`. Using the `ServiceManagedLogGroup` can therefore help us address the `logRetention` deprecation, while still keeping the same log group we have.
- `ServiceManagedLogGroup` does seem like a good idea, as it's not just about Lambda - a service that manages its log group by default - but can be used in other L2 constructs for services that manage their own log groups - and like my point above, can really have a smooth path/consistent experience for when such services might introduces a similar support of specifying the log group to use

humanzz:
Hey @mrgrain, just wondering if u got a chance to checkout my earlier comment and if u have any thoughts abt it?

polothy:
Likely unrelated'ish to this PR, but wanted to just say that LoggingConfig for Lambda isn't available in GovCloud yet. Please don't rush to change all the lambdas that CDK core defines to use the new LoggingConfig because it'll break in GovCloud.  Maybe other partitions as well.

mrgrain:
> So, here's some context and thoughts
> 
> * I understand that the new `logGroup` lambda prop can allow one to specify a new log group, and with being careful abt log retention's removal policy, I can easily end up with 2 log groups, the old original one, and the new easily fully managed one via `logs.LogGroup`. I'm not super keen on this option, as there are features in Lambda/Insights/etc. that I guess will simply leverage the lambda declared log group to click through (think from Lambda Console, Lambda Insights Console, etc.). From time to time I do queries over these logs, to identify some historical trends, etc. I'm not saying it's impossible to do with 2 log groups, I'm just saying it becomes less convenient. We keep our logs for 18 months in general based on recommendations from ORRs/Security Reviews. Multiply that by working on a dozen lambda functions, and having 2 log groups for each function to keep that history becomes even more "less inconvenient"

That's very understandable. But this also implies you will never be able to migrate due to this inconvenience. 

> * I've noticed that in your implementation here `class ServiceManagedLogGroup extends LogGroupBase implements ILogGroup`, while [FunctionOptions.logGroup](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L486) is an `ILogGroup`. To me, this suggested that `ServiceManagedLogGroup` can be used as a value for `logGroup` if someone wants to customize the service managed log group, or a concrete `LogGroup`. Using the `ServiceManagedLogGroup` can therefore help us address the `logRetention` deprecation, while still keeping the same log group we have.

If completed that would have worked. Maybe. There would have still be a circular dependency. 

> * `ServiceManagedLogGroup` does seem like a good idea, as it's not just about Lambda - a service that manages its log group by default - but can be used in other L2 constructs for services that manage their own log groups - and like my point above, can really have a smooth path/consistent experience for when such services might introduces a similar support of specifying the log group to use

We will definitely keep considering this given that other services don't have the flexibility lambda offers. Please open a new issue for the specific services you are interested in so we can prioritize this appropriately. 

---

### Re Migration

Since all your well-named log groups already exists, you could important them into your stack using the existing name.
But I'm not sure if that is much less work.

---

Fundamentally the issue with this comes down to `Function name` being part of `Log group name`. This creates a circular dependency that can only be resolved in of two ways:
- Not have the `Function name` in `Log group name`
- Allow to pass-through all features that a log group supports via the Lambda function

The Lambda team decided to go with the first option. While I personally agree with this choice anyway, it also very much means that the CDK will have to go with this choice. You are rightfully asking us to maintain the second option as a mitigation to complex migration. That's a fair ask, but I can't tell you what the likelihood is that we will do this. In general we are reluctant to create duplication with native CFN features.

However none of that answers your very valid concerns about migration. So we'll have to think more about this.

mrgrain:
> Likely unrelated'ish to this PR, but wanted to just say that LoggingConfig for Lambda isn't available in GovCloud yet. Please don't rush to change all the lambdas that CDK core defines to use the new LoggingConfig because it'll break in GovCloud. Maybe other partitions as well.
> 
> EDIT: I tried this last Friday to fix the CDK deprecations and this is the error I got:
> 
> ```
> Resource handler returned message: "LoggingConfig is not supported in us-gov-west-1. Remove LoggingConfig value from your request and try again
> ```

Well 💩 That's one is on me.

mrgrain:
@polothy I'm trying to get confirmation if LoggingConfig is supposed to be available in GovtCloud or not. If you can and don't mind, I'd also recommend you put a support ticket in. I have so far found the announcement that states "all commercial regions" and a blog post that states "all regions". 😐

packages/aws-cdk-lib/aws-logs/lib/service-managed-log-group.ts
```diff
@@ -0,0 +1,272 @@
...
+        ServiceToken: this.provider.functionArn,
+        LogGroupName: this.logGroupName,
+        LogGroupRegion: cdk.Lazy.string({
+          produce: () => cdk.Stack.of(this).splitArn(this._logGroupArn!, cdk.ArnFormat.COLON_RESOURCE_NAME).region,
+        }),
+        DataProtectionPolicy: this.props.dataProtectionPolicy?._bind(this),
+        KmsKeyId: this.props.encryptionKey?.keyArn,
+        RetentionInDays: this.props.retention,
+        Tags: this.tags.renderedTags,
+        Tagging: options.tagging,
```
polothy, line 138: Is it possible to get tags from the CloudFormation stack? We stopped using tags in our templates and moved to tagging stacks due to some resources not allowing tag updates and/or differences with GovCloud.

mrgrain: That's one reason why this PR isn't finished. 😓

polothy:
> @polothy I'm trying to get confirmation if LoggingConfig is supposed to be available in GovtCloud or not. If you can and don't mind, I'd also recommend you put a support ticket in. I have so far found the announcement that states "all commercial regions" and a blog post that states "all regions". 😐

I'm working with our AWS TAM to check into it and add our +1 to it going to GovCloud.

polothy:
> Well 💩 That's one is on me. Can you please open a new issue @polothy

Created https://github.com/aws/aws-cdk/issues/28919

mrgrain:
> > @polothy I'm trying to get confirmation if LoggingConfig is supposed to be available in GovtCloud or not. If you can and don't mind, I'd also recommend you put a support ticket in. I have so far found the announcement that states "all commercial regions" and a blog post that states "all regions". 😐
> 
> I'm working with our AWS TAM to check into it and add our +1 to it going to GovCloud.

I've got confirmation that LoggingConfig is not available in GovCloud yet. Your TAM will be your best path to get a date.

mrgrain:
Reverting the deprecation: https://github.com/aws/aws-cdk/pull/28934

PR #26950, opened 2023-08-30
feat(eks): allow associating external oidc provider with cluster

Advill:
Implement `associateOpenIdConnectProvider` method in `Cluster` resource.

Implement test for connecting an OIDC resource after creating a cluster

A project I'm working on has fairly strict requirements from the managing IT org around IAM resources. Long story short, they'd prefer that we spin up our clusters and let them add OIDC providers to them. Currently there is no way to short-circuit the lazy OIDC creation logic or to provide a cluster with an OIDC provider manually.

This PR simply grabs an OIDC provider by it's ARN and slots it in the internal OIDC variable.

The added test could probably use some work. I'm not familiar enough with jest or the CDK's `Template` construct to work out how to check that a certain resource type is _not_ provisioned in a synthesis, but I've tested what I can.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-eks/README.md
+++ b/packages/aws-cdk-lib/aws-eks/README.md
@@ -1116,6 +1116,9 @@ const cluster = eks.Cluster.fromClusterAttributes(this, 'MyCluster', {
   kubectlRoleArn: 'arn:aws:iam::123456:role/service-role/k8sservicerole',
 });
 
+// you can also associate a provider created outside the CDK with an existing cluster
+const provider3 = cluster.associateOpenIdConnectProvider('arn:aws:iam::123456:oidc-provider/oidc.eks.eu-west-1.amazonaws.com/id/AB123456ABC')
+
 const serviceAccount = cluster.addServiceAccount('MyServiceAccount');
 
 const bucket = new s3.Bucket(this, 'Bucket');


mrgrain (changes requested):
Hi, can you describe the use case in more detail? Specifically how are you making sure that this "external" OIDC provider hast the correct url?

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
```diff
@@ -1869,6 +1869,20 @@ export class Cluster extends ClusterBase {
     });
   }
 
+  /**
+   * Manually associate an `OpenIdConnectProvider` resource to this cluster. Skips the lazy
+   * allocation of an OpenIdConnectProvider.
+   *
+   * @param arn the arn of the OIDC provider
+   */
+  public associateOpenIdConnectProvider(arn: string): iam.IOpenIdConnectProvider {
```
mrgrain, line 1878: This would need to accept a `IOpenIdConnectProvider`. not an ARN.

Advill: For sure, I was pretty new to the CDK and my exact use case when I first made this PR. Still finishing up some more important work but once I finish I'm going to clean this up and finish off the parts the PR is missing.

Advill:
> Hi, can you describe the use case in more detail? Specifically how are you making sure that this "external" OIDC provider hast the correct url?

The use case is certainly an edge one, but is pretty accurately described in the above post. I could add some checks to the imported OIDC provider to try to ensure that it's configured correctly but I'll need to get more familiar with the creation process for OIDCs for that and it may take some back and forth on the request.

In impementing this, hopefully at least a few other users with specific use cases like mine would be able to jump past the OIDC creation requirement of automatically adding `ALBControllers` and `ServiceAccounts` using provided functions, if it seems like I'm alone in this requirement though it's fair if this PR is denied.

mrgrain:
> > Hi, can you describe the use case in more detail? Specifically how are you making sure that this "external" OIDC provider hast the correct url?
> 
> The use case is certainly an edge one, but is pretty accurately described in the above post. I could add some checks to the imported OIDC provider to try to ensure that it's configured correctly but I'll need to get more familiar with the creation process for OIDCs for that and it may take some back and forth on the request.
> 
> In impementing this, hopefully at least a few other users with specific use cases like mine would be able to jump past the OIDC creation requirement of automatically adding `ALBControllers` and `ServiceAccounts` using provided functions, if it seems like I'm alone in this requirement though it's fair if this PR is denied.

Thanks your response. I'm happy to move forward with this, but we will need to use the interface for it.

PR #26955, opened 2023-08-31
fix(redshift): UserTablePriviliges to track changes using table IDs

Rizxcviii:
To support solving https://github.com/aws/aws-cdk/issues/24246, there needs to first be a refactor of the UserTablePriviliges to instead record the table id. The reason being that new privileges would be granted/revoked on old tables that now have new names, since a change in table name caused an UPDATE action to be triggered. However the issue remains where, since the table has a new name, the grant/revoke action will be called on an invalid table name (i.e. the old table name).
We now use the table id to track tables, therefore preventing UPDATE events to be triggered.

This blocks https://github.com/aws/aws-cdk/pull/24308

This was originally PR https://github.com/aws/aws-cdk/pull/24874, however that had closed. @kaizencc requested changes, that I had added in here.

This was originally PR #26558, however that had closed. @kaizencc requested changes, that I have already implemented previously. However, he did not review them.

BREAKING CHANGE: the behavior of redshift tables has changed. UPDATE action will not be triggered on new table names and instead be triggered on table id changes.

----


Rizxcviii:
Exemption Request: Already stated, an integration test already takes place internally with the internal interface.

kaizencc:
@Rizxcviii sorry for the delay. I just saw these pings -- have been busy on other stuff past few weeks. I will review this tomorrow.

packages/@aws-cdk/aws-redshift-alpha/test/database-query-provider/privileges.test.ts
kaizencc, line 1: On your unit tests here, can you write some inline comments about why the assertions demonstrate what the test is trying to test? I agree with the names of the unit tests, but the assertions aren't immediately obvious to me what their purpose is.

If you can do that successfully, _maybe_ I will agree that we don't need an integ tests here. Feel free to further explain your reasoning as well :)

kaizencc (changes requested):
@Rizxcviii I'm fine with your code at this point. Just want to make sure that the tests fully explain what we're testing and then we're good to go

packages/@aws-cdk/aws-redshift-alpha/test/database-query-provider/privileges.test.ts
Rizxcviii: This has been added.

Rizxcviii:
> @Rizxcviii I'm fine with your code at this point. Just want to make sure that the tests fully explain what we're testing and then we're good to go
> 
> I also updated your PR description to describe the breaking change that i believe is happening here. Feel free to modify that comment how you see fit.

I agree with the breaking change, it would make sense to add it in here :)

kaizencc (approved):
Thanks @Rizxcviii. Sorry for how long it took to get back to you again.

Rizxcviii:
@kaizencc for some reason mergify is giving the error
> Mergify doesn't have permission to update

Rizxcviii:
@mergify refresh

Rizxcviii:
@mergify refresh

PR #26968, opened 2023-09-01
feat(stepfunction-tasks): allow string in `outputLocation` parameter

razin99:
# Issue currently faced
I tried to do the following:
```ts
new AthenaStartQueryExecution(
  stack,
  'Start Athena Query',
  {
    queryString: sfn.JsonPath.stringAt('$.queryString'),
    queryExecutionContext: {
      databaseName: 'database',
    },
    resultConfiguration: {
      encryptionConfiguration: {
        encryptionOption: EncryptionOption.S3_MANAGED,
      },
     outputLocation: {
        bucketName: 'query-results-bucket',
        objectKey: sfn.JsonPath.stringAt('$.folder'),
    },
  });
```
But was unable to due to an error message:
```
Field references must be the entire string, cannot concatenate them (found 's3://$[Token]/$[Token]')
```

# Solution
Allow `outputLocation` to receive string. Then the new parameter can also be the following:
```
outputLocation: sfn.JsonPath.format(
  's3://query-results-bucket/folder/{}',
  sfn.JsonPath.format('$.folder'),
)
```
----

### DIFF
--- a/packages/aws-cdk-lib/aws-stepfunctions-tasks/README.md
+++ b/packages/aws-cdk-lib/aws-stepfunctions-tasks/README.md
@@ -258,6 +258,15 @@ const startQueryExecutionJob = new tasks.AthenaStartQueryExecution(this, 'Start 
   },
 });
 ```
+Alternatively, we could also use the following for `outputLocation`
+```
+resultConfiguration: {
+  outputLocation: sfn.JsonPath.format(
+    's3://query-results-bucket/folder/{}',
+    sfn.JsonPath.format('$.folder'),
+  )
+}
+```
 
 ### GetQueryExecution
 


packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
```diff
@@ -228,7 +238,7 @@ export interface ResultConfiguration {
    *
    * @default - Query Result Location set in Athena settings for this workgroup
   */
-  readonly outputLocation?: s3.Location;
+  readonly outputLocation?: s3.Location | string;
```
mrgrain, line 241: Unfortunately using a union type here will make any other language but typescript very unusable. They just turn into an `any` equivalent.

I see you have encountered an issue and I appreciate your drive to just go an fix it! In this case however, would you mind open an issue first? This gives our triage and support team a chance to have a look at it, possibly come with a workaround and at least inform the solution better.

mrgrain (changes requested):
I see you have encountered an issue and I appreciate your drive to just go an fix it! In this case however, would you mind open an issue first? This gives our triage and support team a chance to have a look at it, possibly come with a workaround and at least inform the solution better.

PR #26969, opened 2023-09-01
feat(cloudwatch): Widgets can define `start` and `end` times, including relative values

### LINKED ISSUE #26945: (aws-cloudwatch): Ability to set time frames for individual graphs in a dashboard
### Describe the feature

Currently, we can set a time frame for an entire dashboard, or set individual time frames for graphs in the CloudWatch dashboard console manually, but the functionality to do so inside CDK doesn't exist. 

Setting these time frames manually in the CloudWatch console is not an option for me, as any changes I make there are overwritten with the next code deployment. 

### Use Case

I'd like to display graphs with different time frames in the same dashboard. For example, I'd like to display the same metric over a period of 7 days, and also display it over 8 hours so I get a more complete look at the metric and its history. 

### Proposed Solution

Just as we have the option to specify start and end times for a [dashboard](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudwatch.Dashboard.html), I'd like that option added for a graph. The start/end times that a graph specifies should override the start/end of the dashboard that the graph sits in. This way, we can specify a default time range for graphs in a dashboard, but individual graphs can still override that when needed.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.79.1

macOS 13.3.1

go-to-k:
This PR adds `start` and `end` properties to Widgets.

These properties can be used to specify the time range for each graph widget independently from those of the dashboard.
The parameters can be specified at `GraphWidget`, `GaugeWidget`, and `SingleValueWidget`.

Closes #26945.

----


packages/aws-cdk-lib/aws-cloudwatch/lib/graph.ts
```diff
@@ -204,6 +204,27 @@ export interface GaugeWidgetProps extends MetricWidgetProps {
...
+   * The start of the time range to use for each widget independently from those of the dashboard.
+   * You can specify start without specifying end to specify a relative time range that ends with the current time.
+   * In this case, the value of start must begin with -P, and you can use M, H, D, W and M as abbreviations for
+   * minutes, hours, days, weeks and months. For example, -PT8H shows the last 8 hours and -P3M shows the last three months.
+   * You can also use start along with an end field, to specify an absolute time range.
+   * When specifying an absolute time range, use the ISO 8601 format. For example, 2018-12-17T06:00:00.000Z.
+   *
+   * @default When the dashboard loads, the start time will be the default time range.
+   */
+  readonly start?: string;
```
kaizencc, line 218: hmm i don't love the idea that we're typing these values as `string` (both for `start` and `end`).

can you either:

a) point me to a place elsewhere in aws-cdk-lib where we already type something similar as a string. 
b) help me design a better type system that can then generate the format that cloudwatch wants

go-to-k: @kaizencc 

I can certainly see that.

a. The `start` and `end` are already used in Dashboard.
https://github.com/aws/aws-cdk/blob/21077901758d1b95a23b09465511f4d3ef8d2728/packages/aws-cdk-lib/aws-cloudwatch/lib/dashboard.ts#L44-L63

b. For example, I came up with a way to create a static method for each type.

```ts
export class StartTimeRange {
  public static fromMinutes(start: number): StartTimeRange {
    if (start <= 0) {
      throw new Error('start must be greater than 0');
    }
    return new StartTimeRange(`-PT${start}M`);
  }

  public static fromHours(start: number): StartTimeRange {
    if (start <= 0) {
      throw new Error('start must be greater than 0');
    }
    return new StartTimeRange(`-PT${start}H`);
  }

  public static fromDays(start: number): StartTimeRange {
    if (start <= 0) {
      throw new Error('start must be greater than 0');
    }
    return new StartTimeRange(`-P${start}D`);
  }

  public static fromWeeks(start: number): StartTimeRange {
    if (start <= 0) {
      throw new Error('start must be greater than 0');
    }
    return new StartTimeRange(`-P${start}W`);
  }

  public static fromMonths(start: number): StartTimeRange {
    if (start <= 0) {
      throw new Error('start must be greater than 0');
    }
    return new StartTimeRange(`-P${start}M`);
  }

  public static fromTimestamp(start: Date): StartTimeRange {
    return new StartTimeRange(start.toISOString());
  }

  private constructor(public readonly start: string) { }
}
```

go-to-k: b-2. Or I thought of this too. By creating `StartTimeRange` interface and separating into unit and date, the combination of end and start can be validated. (e.g. It must be specified `date` in `start` if `end` is specified.)

```ts
interface StartTimeRangeType {
  unit?: string;
  date?: string;
}

export class StartTimeRange {
  public static fromMinutes(unit: number): StartTimeRange {
    if (unit <= 0) {
      throw new Error('start must be greater than 0');
    }
    return new StartTimeRange({
      unit: `-PT${unit}M`,
    });
  }

  public static fromHours(unit: number): StartTimeRange {
    if (unit <= 0) {
      throw new Error('start must be greater than 0');
    }
    return new StartTimeRange({
      unit: `-PT${unit}H`,
    });
  }

  public static fromDays(unit: number): StartTimeRange {
    if (unit <= 0) {
      throw new Error('start must be greater than 0');
    }
    return new StartTimeRange({
      unit: `-P${unit}D`,
    });
  }

  public static fromWeeks(unit: number): StartTimeRange {
    if (unit <= 0) {
      throw new Error('start must be greater than 0');
    }
    return new StartTimeRange({
      unit: `-P${unit}W`,
    });
  }

  public static fromMonths(unit: number): StartTimeRange {
    if (unit <= 0) {
      throw new Error('start must be greater than 0');
    }
    return new StartTimeRange({
      unit: `-P${unit}M`,
    });
  }

  public static fromDate(start: Date): StartTimeRange {
    return new StartTimeRange({
      date: start.toISOString(),
    });
  }

  private constructor(public readonly startTimeRangeType: StartTimeRangeType) {}
}

export class EndTimeRange {
  public static fromDate(end: Date): EndTimeRange {
    return new EndTimeRange(end.toISOString());
  }

  private constructor(public readonly end: string) { }
}

export class Dashboard extends Resource {
  // ...
  // ...
  constructor(scope: Construct, id: string, props: DashboardProps = {}) {
    // ...
    // ...
    if (this.props.end !== undefined && this.props.start.date === undefined) {
      throw new Error('It must be specified `date` in `start` if `end` is specified.')
```

mrgrain: Had a chat with @kaizencc 

Given that we already use a `string` elsewhere, and that it might be quite an effort to provide a complete implementation of this rather complex standard, using `string` will be fine.

Users can always fallback to native language features and provide something like `new Date().toISOString()`

mrgrain (approved):

kaizencc (approved):

mrgrain (approved):

PR #26992, opened 2023-09-03
feat(core): add setter for termination protection on stack

### LINKED ISSUE #14463: (core): add setter for termination protection
Please add a setter for the termination protection flag to the `Stack` class. https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Stack.html#terminationprotection

### Use Case

We would like to be able to automatically enable termination protection to all of our stacks via an `Aspect`. However, since there is no setter for this property (it can only be set in the constructor), there's no way for that to work.

---

### LINKED ISSUE #21304: (aws-cdk-lib): (Change terminationProtection on existing Stack instances)
### Describe the feature

Currently, the only way to enable termination protection on a CDK Stack is in the props passed to the Stack's constructor, e.g.:

```ts
const myStack = new MyStack(app, "MyStack", {
    terminationProtection: true,
    // ...
});
```

However, there is no way to set termination protection on an _existing_ Stack instance. E.g., in the code above, if I then write `myStack.terminationProtection = false`, then I get the following compiler error:

> Cannot assign to 'terminationProtection' because it is a read-only property.

I request that this property be writable.

### Use Case

I'm trying to define an Aspect to enable termination protection on all of my production stacks, so that I don't accidentally forget to enable it in the CDK app. I was hoping to define this Aspect as follows, but instead I get the aforementioned compiler error.

```ts
export class ProdTerminationProtectionAspect implements IAspect {

    constructor(private readonly isProduction: boolean) { }

    public visit(construct: IConstruct): void {
        if (!this.isProduction || !Stack.isStack(construct))
            return;

        construct.terminationProtection = this.isProduction;    // ERROR
    }
}

```

### Proposed Solution

_No response_

### Other Information

Alternatively, you could take this as a feature request to add an "auto-termination-protection-enabling" Aspect to the CDK itself. 🙂

### Acknowledgements

### CDK version used

2.33.0 (build 859272d)

Debian buster (technically in a WSL devcontainer on Windows 11 Pro)

msambol:
Closes #14463.
Closes #21304.

----


packages/@aws-cdk-testing/framework-integ/test/core/test/integ.stack.js.snapshot/manifest.json
```diff
@@ -0,0 +1,107 @@
...
+        "requiresBootstrapStackVersion": 6,
+        "bootstrapStackVersionSsmParameter": "/cdk-bootstrap/hnb659fds/version"
+      }
+    },
+    "Stack": {
+      "type": "aws:cloudformation:stack",
+      "environment": "aws://unknown-account/unknown-region",
+      "properties": {
+        "templateFile": "Stack.template.json",
+        "terminationProtection": true,
```
msambol, line 17: set to true. stack was initialized with false but then set to true with setter.

msambol:
@mrgrain Can I bother you for a review here ?

colifran (approved):
@msambol Looks good!

mrgrain (approved):

mrgrain:
@mergify update

mrgrain:
@mergifyio requeue

PR #27000, opened 2023-09-04
fix(custom-resources): empty Lambda response payload causes deployment failure

### LINKED ISSUE #26429: (custom-resources): Provider function is throwing "Error: return values from user-handlers must be JSON objects. got: ""
### Describe the bug

Hello team, 

I am seeing the error `Error: return values from user-handlers must be JSON objects. got: ""` recently while using the `Provider` construct of `aws-cdk-lib/custom-resources`. Checking the cloudwatch logs of the request, I see that this is the return value sent back from the oneventhandler lambda fn. 

```
{
    $metadata: {
      httpStatusCode: 200,
      requestId: 'xxxx',
      attempts: 1,
      totalRetryDelay: 0
    },
    ExecutedVersion: '$LATEST',
    Payload: {},
    StatusCode: 200
  }
```

I see that there was a change recently in [parseJsonPayload](https://github.com/aws/aws-cdk/blob/399b6bbdbfbf6440145e6d78e18898aae6261702/packages/aws-cdk-lib/custom-resources/lib/provider-framework/runtime/framework.ts#L149C10-L149C26) function from `const text = payload.toString();` to `let text = new TextDecoder().decode(Buffer.from(payload));`.

### Expected Behavior

Provider fn to return empty values/success

### Current Behavior

Provider function is throwing "Error: return values from user-handlers must be JSON objects. got: ""

### Reproduction Steps

Not sure of the reproduction steps here. Although I do not see any return from the event lambda. 
This works fine if the return logged is "" like 
```
{
    $metadata: {
      httpStatusCode: 200,
      requestId: 'xxxx',
      attempts: 1,
      totalRetryDelay: 0
    },
    ExecutedVersion: '$LATEST',
    Payload: "",
    StatusCode: 200
  }
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.87.0

### Framework Version

_No response_

### Node.js Version

18.16

### OS

mac

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

rix0rrr:
The payload response of a Lambda used to be a `string`, and could occasionally be `""`, which we detected and special-case parsed to an empty object. The Payload should never be empty, and will only be that under exceptional circumstances which we haven't been able to pin down yet, but we shouldn't fail in any case.

In SDKv3, the payload response of a Lambda changed to type `Uint8Array`, but a `Uint8Array(0)` doesn't check as *falsey*, so we'd decode it to `""` and then the `JSON.parse()` of that would fail.

First decode, then check the string for emptyness.

Fixes #26429.

----


iliapolo (commented):
I think a test is warranted...if not an integ that at least a unit test. though i suspect integ would actually be easier to write :)

mrgrain (approved):

rix0rrr:
> though i suspect integ would actually be easier to write :)

In fact I don't know how to trigger the empty Payload response condition. It seems to be an unusual case ☹️

packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/lambda-handler/utils.ts
```diff
@@ -1,11 +1,12 @@
...
+  // can be cast into a buffer and then decoded.
+  const text = new TextDecoder().decode(Buffer.from(payload ?? ''));
+  if (!text) { return { }; }
   try {
-    const buffer = Buffer.from(payload);
-    return JSON.parse(new TextDecoder().decode(buffer));
+    return JSON.parse(text);
   } catch {
-    return payload;
+    throw new Error(`return values from user-handlers must be JSON objects. got: "${text}"`);
```
mrgrain, line 9: Can a Lambda Handler not return text? Or is it always a JSON parsable string, i.e. `"foobar"`?

rix0rrr: Looks to be always a JSON representation of a JavaScript object.

rix0rrr: I.e. a plain text is returned as `"Hi I'm a plain string"`.

mrgrain: I'm still concerned we are going to fail on valid values or return unexpected values.
In the assertion handler `parseJsonPayload` is naively called on a potentially existing `payload` property of the response object. That's different than the provider framework where we know we are dealing with a lambda response.

In hindsight, maybe we should do the assertions handler on a different PR.

rix0rrr: I agree that the code is making too many assumptions, and might be subtly broken.

I would posit that it is subtly broken already, and that nothing we're doing is making it worse than what it currently is.

But sure, I'll revert!

mrgrain (approved):

analytically:
This seems to break the release. 
screentshot-2023-10-04 at 16 26 58@2x

PR #27001, opened 2023-09-04
feat(cloud9-alpha): add support for `federated-user` and `assumed-role` for Cloud9 environment ownership

markusz:
Currently, the cloud9-alpha module only supports two IAM entities as the owners of a Cloud9 environment

- Account Root
- User

However, in many environments, access to an AWS account is gained via Federation. To use Cloud9 via the CDK in such environments, workarounds like the following one where required:

```
const cloud9 = new Ec2Environment(this, 'Cloud9', {..});

const cfnC9 = cloud9.node.findChild('Resource') as CfnEnvironmentEC2;
cfnC9.ownerArn = 'arn:sts:..;
```

This merge request adds support for assumed roles and federated users to be owners of C9 environments directly in the CDK construct.

----

### DIFF
--- a/packages/@aws-cdk/aws-cloud9-alpha/test/integ.owner.js.snapshot/manifest.json
+++ b/packages/@aws-cdk/aws-cloud9-alpha/test/integ.owner.js.snapshot/manifest.json
@@ -14,6 +14,7 @@
       "environment": "aws://unknown-account/unknown-region",
       "properties": {
         "templateFile": "cloud9-owner-integ.template.json",
+        "terminationProtection": false,
         "validateOnSynth": false,
         "assumeRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-deploy-role-${AWS::AccountId}-${AWS::Region}",
         "cloudFormationExecutionRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-cfn-exec-role-${AWS::AccountId}-${AWS::Region}",
@@ -168,10 +169,7 @@
         "/cloud9-owner-integ/C9Env/Resource": [
           {
             "type": "aws:cdk:logicalId",
-            "data": "C9EnvF05FC3BE",
-            "trace": [
-              "!!DESTRUCTIVE_CHANGES: WILL_REPLACE"
-            ]
+            "data": "C9EnvF05FC3BE"
           }
         ],
         "/cloud9-owner-integ/URL": [
@@ -214,6 +212,7 @@
       "environment": "aws://unknown-account/unknown-region",
       "properties": {
         "templateFile": "OwnerIntegDefaultTestDeployAssertEE359F09.template.json",
+        "terminationProtection": false,
         "validateOnSynth": false,
         "assumeRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-deploy-role-${AWS::AccountId}-${AWS::Region}",
         "cloudFormationExecutionRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-cfn-exec-role-${AWS::AccountId}-${AWS::Region}",


packages/@aws-cdk/aws-cloud9-alpha/README.md
```diff
@@ -105,7 +105,17 @@ Every Cloud9 Environment has an **owner**. An owner has full control over the en
...
-`Owner` is a user that owns a Cloud9 environment . `Owner` has their own access permissions, resources. And we can specify an `Owner`in an Ec2 environment which could be of two types, 1. AccountRoot and 2. Iam User. It allows AWS to determine who has permissions to manage the environment, either an IAM user or the account root user (but using the account root user is not recommended, see [environment sharing best practices](https://docs.aws.amazon.com/cloud9/latest/user-guide/share-environment.html#share-environment-best-practices)).
+`Owner` is an IAM entity that owns a Cloud9 environment . `Owner` has their own access permissions, resources. You can specify an `Owner`in an Ec2 environment which could be of the following types:
+
+1. Account Root
+2. IAM User
+3. IAM Federated User
+4. IAM Assumed Role
+
+The ARN of the owner must satisfy the following regular expression: `^arn:(aws|aws-cn|aws-us-gov|aws-iso|aws-iso-b):(iam|sts)::\d+:(root|(user\/[\w+=/:,.@-]{1,64}|federated-user\/[\w+=/:,.@-]{2,32}|assumed-role\/[\w+=:,.@-]{1,64}\/[\w+=,.@-]{1,64}))$
+`
```
lpizzinidev, line 116:
```suggestion
The ARN of the owner must satisfy the following regular expression: `^arn:(aws|aws-cn|aws-us-gov|aws-iso|aws-iso-b):(iam|sts)::\d+:(root|(user\/[\w+=/:,.@-]{1,64}|federated-user\/[\w+=/:,.@-]{2,32}|assumed-role\/[\w+=:,.@-]{1,64}\/[\w+=,.@-]{1,64}))$`
```

packages/@aws-cdk/aws-cloud9-alpha/test/cloud9.environment.test.ts
```diff
@@ -5,6 +5,7 @@ import * as iam from 'aws-cdk-lib/aws-iam';
 import * as cdk from 'aws-cdk-lib';
 import * as cloud9 from '../lib';
 import { ConnectionType, ImageId, Owner } from '../lib';
+import { CfnUser } from 'aws-cdk-lib/aws-iam';
```
lpizzinidev, line 8:
```suggestion
```
Already imported at line 4

packages/@aws-cdk/aws-cloud9-alpha/test/cloud9.environment.test.ts
```diff
@@ -124,14 +125,43 @@ test('environment owner can be an IAM user', () => {
     imageId: cloud9.ImageId.AMAZON_LINUX_2,
     owner: Owner.user(user),
   });
+
+  const userLogicalId = stack.getLogicalId(user.node.defaultChild as CfnUser);
   // THEN
```
lpizzinidev, line 130:
```suggestion
  // THEN
  const userLogicalId = stack.getLogicalId(user.node.defaultChild as iam.CfnUser);
```
After removing the redundant `import` statement.

packages/@aws-cdk/aws-cloud9-alpha/README.md
```diff
@@ -135,9 +144,35 @@ new cloud9.Ec2Environment(this, 'C9Env', {
 })
 ```
 
+To specify a specific IAM Federated User as the environment owner, use `Owner.federatedUser(accountId, userName)`. The user should have the `AWSCloud9Administrator` managed policy
```
lpizzinidev, line 147:
```suggestion
To specify a specific IAM Federated User as the environment owner, use `Owner.federatedUser(accountId, userName)`. 
The user should have the `AWSCloud9User` (preferred) or `AWSCloud9Administrator` managed policy attached.
```
From [docs](https://docs.aws.amazon.com/cloud9/latest/user-guide/share-environment.html#share-environment-member-roles). Should be updated also in the rest of the README.

packages/@aws-cdk/aws-cloud9-alpha/test/integ.owner.ts
```diff
@@ -0,0 +1,35 @@
+import * as ec2 from 'aws-cdk-lib/aws-ec2';
+import { Stack, App, StackProps, CfnOutput } from 'aws-cdk-lib';
+import * as integ from '@aws-cdk/integ-tests-alpha';
+import { Construct } from 'constructs';
+import * as cloud9 from '../lib';
+import { ImageId, Owner } from '../lib';
```
lpizzinidev, line 6:
```suggestion
```
Already imported at line 5

packages/@aws-cdk/aws-cloud9-alpha/test/integ.owner.ts
```diff
@@ -0,0 +1,35 @@
...
+      restrictDefaultSecurityGroup: false,
+      maxAzs: 2,
+      natGateways: 1,
+    });
+
+    // create a cloud9 ec2 environment in a new VPC
+    const c9env = new cloud9.Ec2Environment(this, 'C9Env', {
+      vpc,
+      imageId: ImageId.AMAZON_LINUX_2,
+      owner: Owner.assumedRole('807336301584', 'Admin/test'),
```
lpizzinidev, line 22:
```suggestion
      imageId: cloud9.ImageId.AMAZON_LINUX_2,
      owner: cloud9.Owner.assumedRole('807336301584', 'Admin/test'),
```
After removing the redundant `import` statement.

lpizzinidev (changes requested):
Thanks! 👍
Looks good overall, just some cleanup and documentation adjustments.
And sorry for the late review.

packages/@aws-cdk/aws-cloud9-alpha/README.md
markusz: Thanks. Resolved in [91d8984](https://github.com/aws/aws-cdk/pull/27001/commits/91d898486feaf9d87bd86bde8b2fb69fe9eae601)

markusz:
> Thanks! 👍 Looks good overall, just some cleanup and documentation adjustments. And sorry for the late review.

Thanks @lpizzinidev for the review. 

I just added the proposed changes.

lpizzinidev (approved):

packages/@aws-cdk/aws-cloud9-alpha/README.md
```diff
@@ -105,7 +105,16 @@ Every Cloud9 Environment has an **owner**. An owner has full control over the en
 
 By default, the owner will be the identity that creates the Environment, which is most likely your CloudFormation Execution Role when the Environment is created using CloudFormation. Provider a value for the `owner` property to assign a different owner, either a specific IAM User or the AWS Account Root User.
 
-`Owner` is a user that owns a Cloud9 environment . `Owner` has their own access permissions, resources. And we can specify an `Owner`in an Ec2 environment which could be of two types, 1. AccountRoot and 2. Iam User. It allows AWS to determine who has permissions to manage the environment, either an IAM user or the account root user (but using the account root user is not recommended, see [environment sharing best practices](https://docs.aws.amazon.com/cloud9/latest/user-guide/share-environment.html#share-environment-best-practices)).
+`Owner` is an IAM entity that owns a Cloud9 environment . `Owner` has their own access permissions, resources. You can specify an `Owner`in an Ec2 environment which could be of the following types:
```
vinayak-kukreja, line 108: NIT:

```suggestion
`Owner` is an IAM entity that owns a Cloud9 environment. `Owner` has their own access permissions, and resources. You can specify an `Owner`in an EC2 environment which could be of the following types:
```

packages/@aws-cdk/aws-cloud9-alpha/README.md
```diff
@@ -105,7 +105,16 @@ Every Cloud9 Environment has an **owner**. An owner has full control over the en
...
+
+1. Account Root
+2. IAM User
+3. IAM Federated User
+4. IAM Assumed Role
+
+The ARN of the owner must satisfy the following regular expression: `^arn:(aws|aws-cn|aws-us-gov|aws-iso|aws-iso-b):(iam|sts)::\d+:(root|(user\/[\w+=/:,.@-]{1,64}|federated-user\/[\w+=/:,.@-]{2,32}|assumed-role\/[\w+=:,.@-]{1,64}\/[\w+=,.@-]{1,64}))$`
```
vinayak-kukreja, line 115: I think this should be part of docs for the property itself and not the Readme

packages/@aws-cdk/aws-cloud9-alpha/lib/environment.ts
```diff
@@ -257,6 +257,26 @@ export class Owner {
     return { ownerArn: user.userArn };
   }
 
+  /**
+   * Make an IAM assumed role the environment owner
+   *
+   * @param accountId The AccountId of the target account
```
vinayak-kukreja, line 263:
```suggestion
   * @param accountId The account id of the target account
```

packages/@aws-cdk/aws-cloud9-alpha/test/integ.owner.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+const app = new App();
+
+new integ.IntegTest(app, 'OwnerInteg', {
+  testCases: [new Cloud9Env(app, 'cloud9-owner-integ')],
+});
+
+app.synth({ force: true });
```
vinayak-kukreja, line 34: Why do we need this?

packages/@aws-cdk/aws-cloud9-alpha/lib/environment.ts
```diff
@@ -257,6 +257,26 @@ export class Owner {
...
+   * @param roleName The name of the assumed role
+   */
+  public static assumedRole(accountId: string, roleName: string): Owner {
```
vinayak-kukreja, line 266: Can this be named `assumeRole`?

vinayak-kukreja, line 266: Why not user [IRole](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_iam.Role.html)?

```suggestion
  public static assumedRole(accountId: string, roleName: string): Owner {
```

packages/@aws-cdk/aws-cloud9-alpha/lib/environment.ts
```diff
@@ -257,6 +257,26 @@ export class Owner {
...
+  }
+
+  /**
+   * Make an IAM federated user the environment owner
+   *
+   * @param accountId The AccountId of the target account
+   * @param userName The name of the federated user
+   */
+  public static federatedUser(accountId: string, userName: string): Owner {
+    return { ownerArn: `arn:${cdk.Aws.PARTITION}:sts::${accountId}:federated-user/${userName}` };
```
vinayak-kukreja, line 277: Could you please tell me where did you find this `arn pattern`?

```
arn:${cdk.Aws.PARTITION}:sts::${accountId}:federated-user/${userName}
```

vinayak-kukreja (commented):
Hey, this is great. Thank you for opening a PR with us.

I have some queries. Please let me know if any of it does not make sense.

packages/@aws-cdk/aws-cloud9-alpha/lib/environment.ts
markusz: The pattern is defined [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html) and matches the constraint documented in the [OwnerArn property](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloud9-environmentec2.html#cfn-cloud9-environmentec2-ownerarn) in Cloud9s Cfn Resource Definiton

packages/@aws-cdk/aws-cloud9-alpha/lib/environment.ts
markusz: I did use `IRole` first (see [this commit](https://github.com/aws/aws-cdk/pull/27001/commits/7848cc36c246ed2a2806bcf4deea1fe6b2929f07)) but this creates ARNs that are not expected by Cloud9.

Cloud9 only allows the user to be of type `root`, `user`, `federated-user` and `assumed-role` (see [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloud9-environmentec2.html#cfn-cloud9-environmentec2-ownerarn)).

The `IRole` interface is to broad for this and would allow developers to use any Role as the Cloud9 owner. This would not work due to above C9 constraints and ultimately lead to a failed development.

I would prefer to use a CDK construct here, but I don't think there is a CDK IAM construct that always resolves to `arn:*/assumed-role/*`.

Please let me know if there is one

markusz: This maps to an IAM identifier called assume**d**-role, so I think this is the best name for the function

packages/@aws-cdk/aws-cloud9-alpha/test/integ.owner.ts
markusz: I assume you refer to `{ force: true }`. This is a left-over from debugging the tests, we don't need it. Removed in 170f23fc

If you refer to the entire LoC: It's the only way I know how to programmatically synthesize CDK templates

packages/@aws-cdk/aws-cloud9-alpha/README.md
markusz: I considered this, but I feel that this makes the docs very verbose while adding little value for the user

By adding the two functions `assumedRole(..)` and `federatedUser(..)` I abstract that very complexity away from the users. They now only need to know about the account id and the user / role name they want to use. For both the Regex adds no value since they need to refer to existing users / roles that are inherently well-formatted

I even considered leaving the regex out entirely, but ultimately I think it's worth to mention it once

markusz:
@vinayak-kukreja Thank you for the review.

I incorporated some changes and argued to keep my implementation for others.

Happy to hear your thoughts

packages/@aws-cdk/aws-cloud9-alpha/test/integ.owner.ts
```diff
@@ -0,0 +1,34 @@
...
+      restrictDefaultSecurityGroup: false,
+      maxAzs: 2,
+      natGateways: 1,
+    });
+
+    // create a cloud9 ec2 environment in a new VPC
+    const c9env = new cloud9.Ec2Environment(this, 'C9Env', {
+      vpc,
+      imageId: cloud9.ImageId.AMAZON_LINUX_2,
+      owner: cloud9.Owner.assumedRole('807336301584', 'Admin/test'),
```
vinayak-kukreja, line 21: Hey, just to confirm, this is not your AWS account id or a valid one?

vinayak-kukreja:
> @vinayak-kukreja Thank you for the review.
> 
> I incorporated some changes and argued to keep my implementation for others.
> 
> Happy to hear your thoughts

Hey, @markusz thank you for addressing feedback. Just want to know about the account id mentioned in the test.

packages/@aws-cdk/aws-cloud9-alpha/test/integ.owner.ts
mikewrighton: Seems to be somewhat of a convention to use `123456789012` for tests?

markusz: It is an actual `accountId` and an actual user in that account. If I understood integration tests in the CDK correctly, they synth, deploy and destroy and actual stack in an actual environment to verify correctness.

In my case, I need to refer to existing IAM entities in the account that I run the integration tests in and I don't see any other way.

I could avoid having to specify the account id by using `Stack.of(this).account`, but that would require a breaking change to `environment.ts:Owner` as the functions are static and not members of the `Ec2Environment` construct. This would also make assumed-roles that are defined in other accounts impossible (although I don't think it's this is possible)

Please advise if you see another way to test this

vinayak-kukreja: But these integ tests would fail if the resource is deleted from the account or the permissions no longer allow operations on the account :)
Also, it is not recommended usually to hardcode account ids in code base. 

----

What you can do here is,
1. Create a role in the integ test and that should be a dependency for provisioning your C9 Environment.
2. Use the [environment](https://docs.aws.amazon.com/cdk/v2/guide/environments.html) from the CLI which has the account and the region. For instance,
```
new MyDevStack(app, 'dev', { 
  env: { 
    region: process.env.CDK_DEFAULT_REGION 
}});
````

Let me know if this makes sense.

vinayak-kukreja: Also, would `accountId` ever be different from the user's account id?

markusz:
@vinayak-kukreja I read up again on how integration tests work and I think I found a good solution that makes the test self-contained and requires no hardcoded values.

Can you have a look at the changes in https://github.com/aws/aws-cdk/pull/27001/commits/c4589cc964a12c059d9eaa700487e0b348753b10

Thanks

vinayak-kukreja:
Hey, thank you for making the changes. We will take a look at the PR soon.

vinayak-kukreja (approved):
Hey, thank you for your contribution. Apologies for the delay. 🙏

PR #27002, opened 2023-09-04
fix(events-targets): AwsApi is still using Node 16

### LINKED ISSUE #26998: lambda-events AwsApi needs updating to Node18 and SDKv3
### Describe the bug

The handler function in `AwsApi` is using Node16 and SDKv2.
It should use Node18 and SDKv3.

### Expected Behavior

It is using node 18

### Current Behavior

Using node 16

### Reproduction Steps

N/A

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.94.0

### Framework Version

2.94.0

### Node.js Version

Any

### OS

Any

### Language

Typescript, Python, .NET, Java, Go

### Language Version

_No response_

### Other information

_No response_

mrgrain:
Migrate the `aws-api-handler` code to use SDKv3 and move it into `custom-resource-handlers`.
Updates the Construct config to use the new code and Node18.
To verify functionality, the existing integration test has been extended to include a new rule and target that can be asserted on.

Includes a config change to the `custom-resource-handlers` package to run test from TS.

Closes #26998

----

### DIFF
--- a/packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
+++ b/packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
@@ -756,7 +756,7 @@ interface AuthenticationConfiguration {
    * The Amazon Resource Name (ARN) of the IAM role that grants the App Runner service access to a
    * source repository. It's required for ECR image repositories (but not for ECR Public repositories).
    *
-   * @defult - no access role.
+   * @default - no access role.
    */
   readonly accessRoleArn?: string;
 


packages/@aws-cdk/sdk-v2-to-v3-adapter/lib/sdk-v3-helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+import { getV3ClientPackageName } from './get-v3-client-package-name';
+
+interface SdkV3Package {
+  service: string;
+  packageName: string;
+  packageVersion: string;
+  pkg: object;
+}
+
+export function loadV3ClientPackage(service: string): SdkV3Package {
```
rix0rrr, line 11: Is this code fresh or coming from somewhere else? Is the `AwsCustomResource` using the same code?

mrgrain: Copied over from Integ-Test assertions. I plan a follow up PR to unify this, but wanted to avoid updating all the integ test snapshots in this PR.

 `AwsCustomResource` uses a more complicated loading mechanism since it allows users to install a newer version of the SDK. It also does initialize Clients differently, e.g. allowing a different region and assumedRole to be specified.

After getting this to Node18, I still wonder if we should deprecate/replace the current implementation of `AwsApi` with one that uses the same code as AwsCustomResource.

packages/@aws-cdk/custom-resource-handlers/lib/aws-events-targets/aws-api-handler/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  console.log(`AWS SDK V3: ${awsSdk.packageName}@${awsSdk.packageVersion}`);
+
+  const client = getV3Client(awsSdk, { apiVersion: event.apiVersion });
+  const Command = getV3Command(awsSdk, event.action);
+  const commandInput = event.parameters ?? {};
+
+  try {
+    const response = await client.send(new Command(commandInput));
+    delete response.$metadata;
+    console.log('Response: %j', response);
```
MrArnoldPalmer, line 26: I get that this handler is for triggering effects and doesn't return a value, but I still feel like there might be a sharp edge here wrt Uint8Array types. Someone could use this to trigger Kinesis today passing a string payload and that would break right? We just pass parameters as aws-events passes them I suppose, and I'm not even sure how we could place a marker or something for special case deserialization.

MrArnoldPalmer (approved):
Approved with DNM because this seems like something others may want to give a peek as well.

packages/@aws-cdk/custom-resource-handlers/lib/aws-events-targets/aws-api-handler/index.ts
mrgrain: Let me double check this,. But I'm quite sure the default logging algorithm does work for every object without fail. Unlike JSON.stringify() which often throws for unknown data.

iliapolo (approved):

packages/@aws-cdk/custom-resource-handlers/lib/aws-events-targets/aws-api-handler/index.ts
mrgrain: Checked this. It logs the payload as bytes, `{"0":123,"1":34,"2":101,"3":114,"4":114,"5":111,"6":114,"7":84,"8":121, ...` and doesn't fail. It's not great, but I think it's okay.

As for input: The function is called from EvenBridge (and the API directly) which is JSON by definition. Any other invocations would be outside the expected usage pattern and transformation into a string would be the users responsibility.

rix0rrr: I think Mitch' concern is the opposite?

If the SDKv2 API used to accept a `string` and pass it on, but SDKv3 API now accepts a `Uint8Array` for that same field (but not a `string`), existing use cases are now broken.

mrgrain: Ah I see. Yes you are right about this. 

> It would seem like the only way to do this properly is to know exactly what fields are marked as "blob"s in the API

Yes and that seems unreasonable for us to do.

rix0rrr: That seems unreasonable, yes. But right now this is breaking certain use cases.

Issue #26798, opened 2023-08-18
(integ-tests-alpha): awsApiCall failing for Kinesis putRecord - cannot properly pass Data


postsa:
### Describe the bug

When I pass an expected `Uint8Array` to the params of `Kinesis.putRecord` I receive the failure message 

```
Received response status [FAILED] from custom resource. Message returned: Cannot read properties of undefined (reading 'byteLength')
```

### Expected Behavior

Put record should succeed with the input expected by the [PutRecordCommandInput](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-kinesis/Interface/PutRecordCommandInput/) interface

### Current Behavior

Here is the relevant log from the assertion lambda:
```
2023-08-18T00:01:41.346Z	5ad0ff87-aaa8-4784-9458-8f1de4e21e51	INFO	SDK request to Kinesis.putRecord with parameters 
{
    "StreamName": "KinesisEventProcessingIntegrationStack",
    "PartitionKey": "testKey",
    "Data": {
        "0": 123,
        "1": 34,
        "2": 110,
        "3": 97,
        "4": 109,
        "5": 101,
        "6": 34,
        "7": 58,
        "8": 34,
        "9": 119,
        "10": 111,
        "11": 114,
        "12": 108,
        "13": 100,
        "14": 34,
        "15": 125
    }
}
```

I think this is because the parameters are [encoded](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/sdk.ts#L95) with [JSON.stringify](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/providers/provider.ts#L232) which turns a Uint8Array into a JSON

```
> data
Uint8Array(16) [
  123,  34, 110,  97, 109,
  101,  34,  58,  34, 119,
  111, 114, 108, 100,  34,
  125
]
> JSON.stringify(data)
'{"0":123,"1":34,"2":110,"3":97,"4":109,"5":101,"6":34,"7":58,"8":34,"9":119,"10":111,"11":114,"12":108,"13":100,"14":34,"15":125}'
```

### Reproduction Steps

Create a test stack with a kinesis stream and put a record on it with `assertions.awsApiCall`

 ```
  const kinesisStream: IStream = new Stream(stack, "TestKinesisStream");
 
[...]

PR #27009, opened 2023-09-05
fix(integ-tests): Uint8Arrays are not decoded properly 

otaviomacedo:
AWS SDK v3 has changed some data types to Uint8Array. If a value of this type was provided by the user, it would not be decoded properly back to a Uint8Array before making the call.

Recognize when there is an encoded Uint8Array and decode it back properly.

Closes https://github.com/aws/aws-cdk/issues/26798.

----


packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/lambda-handler/utils.ts
```diff
@@ -12,11 +12,34 @@ export function parseJsonPayload(payload: any): any {
...
     }
   }));
 }
+
+function decodeValue(value: string): any {
+  const parsed = JSON.parse(value);
+
+  if (parsed != null && !Array.isArray(parsed) && typeof parsed === 'object') {
+    const entries = Object.entries(parsed);
+    if (isByteArray(entries)) {
```
iliapolo, line 29: But what if the payload is a dictionary that *looks* like a stringified bytearray, but is actually supposed to be a normal dictionary? 

I agree it might be a little edgy, but can we think of a different solution that considers it? Like adding a marker property when we encode, so we can use it when we decode? 

@mrgrain remember this? can you think of another way?

rix0rrr: I don't even understand how we end up here. Where is that `Uint8Array` coming from, and how did it end up `JSON.stringify()`ed here?

iliapolo: It is being passed directly by the user.

```ts
const kinesisStream: IStream = new Stream(stack, "TestKinesisStream");

 const buf = Buffer.from(JSON.stringify({ name: "world" }));
 const data = new Uint8Array(
   buf.buffer,
   buf.byteOffset,
   buf.byteLength / Uint8Array.BYTES_PER_ELEMENT
 );

 testCase.assertions.awsApiCall("Kinesis", "putRecord", {
   StreamName: kinesisStream.streamName,
   Data: data,
   PartitionKey: "testKey",
 });
```

mrgrain: Yeah, there is a reason why untyped arguments that go over the network should really be JSON strings 🙈 
I wonder if this is even a SDKv3 bug. I know we changed the handling here, but this might have not worked before (but I don't know).

I'll guess we have to put a marker on it if we want to support it. Also should probably base64 encode it during transit.

rix0rrr: Yeah a `Uint8Array` cannot be expected to be sent over the wire just like that.

Either:
- We reject the value, but allow the use case by making the equality testing a bit smarter in the assertions library -- perhaps we can test the equality of a Uint8Array to a string (assuming utf-8 encoding), or an array of numbers.
- Either that, or we marshal the `Uint8Array` explicitly over the wire with a type tag, and reconstruct it on the other end.

Honestly, I'm feeling most for the first solution. Morally, what we are trying to do is test equality of the buffer to a certain string. The fact that we need to encode/decode that string/byte array could be an implementation detail that that the assertions module just takes care of.

rix0rrr (changes requested):
See my last comment

packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/lambda-handler/types.ts
```diff
@@ -150,6 +150,14 @@ export interface AwsApiCallRequest {
...
    */
   readonly outputPaths?: string[];
+
+  /**
+   * Types that can't be properly serialized with JSON.stringify()
+   * are marked here. For example, an entry "foo" -> "ArrayBufferView"
+   * signals to the decoder that "foo" should be handled separately
+   * and treated as an array of bytes.
+   */
+  readonly specialTypes?: Record<string, string>;
```
rix0rrr, line 160: This solution of marking certain parameters as having a different type cannot be applied recursively (or at least not conveniently).

Can we not do an object with a type tag, that we can apply recursively?

```ts
encode({
  string: 'string',
  obj: { field: 'value' },
  weirdarray: new Uint8Array([1, 2, 3]),
})
// => 
{
  string: 'string',
  obj: { field: 'value' },
  weirdarray: {
    '$type': 'ArrayBufferView',
    elements: [1, 2, 3],
  },
}
```

rix0rrr (approved):

PR #27010, opened 2023-09-05
feat(stepfunctions): add comment ability for when condition

### LINKED ISSUE #27005: aws-stepfunctions: comment field appears not supported
### Describe the feature

When editing a state machine via the AWS Console, I could provide a comment for a choice rule. It suggest my state language will look like this:

```json
    "Verification needed?": {
      "Type": "Choice",
      "Choices": [
        {
          "Variable": "$.verification_needed",
          "IsBoolean": true,
          "Next": "Set status to pending verification",
          "Comment": "yes"
        }
      ],
      "Default": "Wait for approval"
    },
```

This "yes" comment gets nicely displayed in the graph. However, I could not see a way to generate this comment field withi CDK. Is that correct?

Are there any plans to introduce that? Really great for more complex flows.

Unfortunately this comment is only visible in the editor, not in the "Definition" tab.

### Use Case

Improve output of state machine graph.

### Proposed Solution

Add a property or optional input to `sfn.Condition.isBoolean()`.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.91.0

Mac x86_64, Darwin kernel 22.6.0

msambol:
Currently you can comment on a `Choice` state. 
This adds the ability to comment on a when condition.

Screenshot 2023-09-05 at 7 47 35 AM

Screenshot 2023-09-05 at 7 54 29 AM

Closes #27005.

----


packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
```diff
@@ -318,16 +318,16 @@ export abstract class State extends Construct implements IChainable {
   /**
    * Add a choice branch to this state
    */
-  protected addChoice(condition: Condition, next: State) {
-    this.choices.push({ condition, next });
+  protected addChoice(condition: Condition, next: State, comment?: string) {
+    this.choices.push({ condition, next, comment });
```
kaizencc, line 322: Ideally this function would be 

```ts
protected addChoice(choice: ChoiceTransition) {
  this.choices.push(choice);
}
```

but I guess we can't change the function signature now

kaizencc: I wonder if we can do this to help make this function signature stable if we ever add more props to `ChoiceTransition`:

```ts
interface ChoiceTransition extends ChoiceTransitionOptions {
  condition;
  next;
}

interface ChoiceTransitionOptions {
  comment?;
}

protected addChoice(condition: Condition, next: State, options: ChoiceTransitionOptions) {}
```

This has 2 benefits:

- using the function would look like

```ts
choice.when(sfn.Condition.stringEquals('$.color', 'BLUE'), handleBlueItem, {
  comment: 'blue item comment',
});
```

which imo makes it clear that we're providing a comment.

- if we add an additional prop to `ChoiceTransition` in the future, like `default` or something, we can easily add it to `ChoiceTransitionOptions`

kaizencc (dismissed):
Thanks for contributing this @msambol. Just a minor nit on how the function looks

msambol:
@kaizencc I updated per your feedback—thanks!

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
```diff
@@ -491,13 +491,26 @@ interface ChoiceTransition {
   condition: Condition;
 }
 
+/**
+ * Options for Choice Transition
+ */
+export interface IChoiceTransitionOptions {
```
kaizencc, line 497: @msambol what was the reason for naming this `IChoiceTransitionOptions`? I assume you ran into some linter rule that yelled at you? Want to find out because I don't think it makes sense here, and we should exempt this from that rule if thats the case

msambol: @kaizencc Yep, I got:

```
aws-cdk-lib:     The declaring class is introduced here
aws-cdk-lib: aws-stepfunctions/lib/states/state.ts:503:3 - error JSII3008: The "comment" property of struct "aws-cdk-lib.aws_stepfunctions.ChoiceTransitionOptions" must be "readonly". Rename "aws-cdk-lib.aws_stepfunctions.ChoiceTransitionOptions" to "IChoiceTransitionOptions" if it is meant to be a behavioral interface.
```

msambol: @kaizencc From what I'm reading there isn't a way to silence JSII errors? Hmm..

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
```diff
@@ -491,13 +491,26 @@ interface ChoiceTransition {
...
+  /**
+   * An optional description for the choice transition
+   *
+   * @default No comment
+   */
+  comment?: string;
```
jogold, line 503: does this fix the JSII error maybe?

```suggestion
  readonly comment?: string;
```

msambol: @jogold It did. Good shout, Jonathan. I was following the pattern of `state` and `condition` but there is no reason this can't be `readonly`.

jogold (approved):

msambol:
@kaizencc Mind taking another look here?

PR #27012, opened 2023-09-05
feat(backup): schedule expression timezone for backup plans

jogold:
Add timezone support for schedule expressions in backup plans.

See https://aws.amazon.com/about-aws/whats-new/2023/08/aws-backup-local-time-zone-selections/

----

### DIFF
--- a/packages/aws-cdk-lib/aws-backup/README.md
+++ b/packages/aws-cdk-lib/aws-backup/README.md
@@ -71,6 +71,7 @@ plan.addRule(new backup.BackupPlanRule({
     hour: '3',
     minute: '30',
   }),
+  scheduleExpressionTimezone: 'Europe/Brussels', // Optional, defaults to UTC
   moveToColdStorageAfter: Duration.days(30),
 }));
 ```


packages/aws-cdk-lib/aws-backup/lib/rule.ts
```diff
@@ -44,6 +44,17 @@ export interface BackupPlanRuleProps {
...
+  /**
+   * The name of the time zone for the schedule expression, following the IANA time zone standard.
+   *
+   * @example America/Los_Angeles
+   *
+   * @see https://www.iana.org/time-zones
+   *
+   * @default - UTC
+   */
+  readonly scheduleExpressionTimezone?: string;
```
kaizencc, line 56: we export a `TimeZone` prop out of core.

packages/aws-cdk-lib/aws-backup/lib/rule.ts
```diff
@@ -213,6 +224,10 @@ export class BackupPlanRule {
       throw new Error('`scheduleExpression` must be of type `cron`');
     }
 
+    if (props.scheduleExpressionTimezone && !props.scheduleExpression) {
```
kaizencc, line 227: there are some `scheduleExpression`s that don't make sense with a timezone, and this check ignores that.

kaizencc (changes requested):
Hi @jogold, I think we should implement this similar to what we do in `scheduler-alpha`:

https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/aws-scheduler-alpha/lib/schedule-expression.ts

Of course, I still don't understand why we don't just invest time in getting the schedule class into core and I know you've tried exactly that before.

jogold:
> Hi @jogold, I think we should implement this similar to what we do in `scheduler-alpha`:
> 
> [`main`/packages/@aws-cdk/aws-scheduler-alpha/lib/schedule-expression.ts](https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/aws-scheduler-alpha/lib/schedule-expression.ts?rgh-link-date=2023-09-05T13%3A49%3A15Z)
> 
> Of course, I still don't understand why we don't just invest time in getting the schedule class into core and I know you've tried exactly that before.

OK, this means deprecating `scheduleExpression` in favor of `schedule`? `cronExpression`? other name? and you don't want to import the `ScheduleExpression` class of `aws-scheduler-alpha` in here?

packages/aws-cdk-lib/aws-backup/lib/rule.ts
jogold: missed that one

jogold:
> that `scheduleExpression` only accepts CRON expressions so I wonder what happens if you supply `events.Schedule.rate('1 minute')`.

https://github.com/aws/aws-cdk/blob/d69c51a2e10864e69d4610b6df30d40a9bfec3ac/packages/aws-cdk-lib/aws-backup/lib/rule.ts#L212-L214

> I hate this and want to fix it once and for all. What do you think @jogold? If you agree, I can spend some time trying to get the first part right and that should kick start the other changes.

Yeah, it seems to be the right thing to do.

kaizencc:
Oh my that's a horrible interface in backup. What we're talking about seems like it will be an improvement across the board. Give me till the end of the week to stand up a PR to put abstract schedule in core?

jogold:
Closing in favor of #27105

PR #27014, opened 2023-09-05
fix(lambda-nodejs): incorrect SDK v2 warning generated

### LINKED ISSUE #26966: aws-lambda-nodejs/lib/bundlingts.ts: Missing condition when checking for v2 runtime
### Describe the bug

Hi, 

I have been receiving weird warnings from aws cdk about nodejs runtime since upgrading to v2.93. 

When creating new NodejsFunction, I make sure to pass `runtime: Runtime.NODEJS_16_X` as one of the props. However, I still get this warning: "If you are relying on AWS SDK v2 to be present in the Lambda environment already, please explicitly configure a NodeJS runtime of Node 16 or lower". 

My company requires all cdk deployment to be executed with strict-mode so we can't ignore this warning.

### Expected Behavior

No warning raised by cdk because I am specifying Nodejs 16 or lower as instructed

### Current Behavior

A warning is raised whenever I create a NodejsFunction with `runtime: Runtime.NODEJS_16_X` 

### Reproduction Steps

Run cdk synth to create following resource:
```
const lambda = new NodejsFunction(this, 'id', {
      entry: join(__dirname, "..", "lambdas", "lambda.ts"),
      handler: "handler",
      timeout: Duration.minutes(3),
      functionName: lambdaFnName,
      runtime: Runtime.NODEJS_16_X,
      memorySize: 512,
      depsLockFilePath: join(__dirname, "..", "..", "..", "..", "pnpm-lock.yaml"),
      role: lambdaExecutionRole,
      logRetention: RetentionDays.TWO_WEEKS,
    });
```

### Possible Solution

I believe this is the code section that cause the warning: 

https://github.com/aws/aws-cdk/blob/d6b71adb41cb2cccf30e2301f872806e12ad4a87/packages/aws-cdk-lib/aws-lambda-nodejs/lib/bundling.ts#L127C5-L143C1

```
// Modules to externalize when using a constant known version of the runtime.
    // Mark aws-sdk as external by default (available in the runtime)
    const isV2Runtime = isSdkV2Runtime(props.runtime);
    const versionedExternals = isV2Runtime ? ['aws-sdk'] : ['@aws-sdk/*'];
    // Don't automatically externalize any dependencies when using a `latest` runtime which may
    // update versions in the future.
    const defaultExternals = props.runtime?.isVariable ? [] : versionedExternals;
    const externals = props.externalModules ?? defaultExternals;

    // Warn users if they are trying to rely on global versions of the SDK that aren't available in
    // their environment.
    if (isV2Runtime && externals.some((pkgName) => pkgName.startsWith('@aws-sdk/'))) {
      cdk.Annotations.of(scope).addWarningV2('@aws-cdk/aws-lambda-nodejs:sdkV3NotInRuntime', 'If you are relying on AWS SDK v3 to be present in the Lambda environment already, please explicitly configure a NodeJS runtime of Node 18 or higher.');
    } else if (externals.includes('aws-sdk')) {
      cdk.Annotations.of(scope).addWarningV2('@aws-cdk/aws-lambda-nodejs:sdkV2NotInRuntime', 'If you are relying on AWS SDK v2 to be present in the Lambda environment already, please explicitly configure a NodeJS runtime of Node 16 or lower.');
    }

```
If someone pass `runtime: Runtime.NODEJS_16_X` and no externalModules provided like my example above
 -> isV2Runtime is True -> versionedExternals =  ['aws-sdk'] -> defaultExternals = ['aws-sdk'] -> externals = ['aws-sdk']
-> the condition `if (externals.includes('aws-sdk'))` on line 140 is true 
-> warning message displayed

**Possible fix**: adding more condition to line 140: if (!isV2Runtime && externals.includes('aws-sdk'))

_No response_

### CDK CLI Version

2.93.0

### Framework Version

_No response_

### Node.js Version

16

### OS

Ubuntu

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

plumdog:
Fixes to remove an incorrectly generated warning when using Node JS runtime <= 16 with `NodejsFunction`

Closes #26966 

----

### DIFF
--- a/packages/aws-cdk-lib/aws-lambda-nodejs/test/bundling.test.ts
+++ b/packages/aws-cdk-lib/aws-lambda-nodejs/test/bundling.test.ts
@@ -882,6 +882,24 @@ test('bundling with <= Node16 warns when sdk v3 is external', () => {
   );
 });
 
+test('bundling with <= Node16 does not warn with default externalModules', () => {
+  const myStack = new Stack(app, 'MyTestStack2');
+  Bundling.bundle(myStack, {
+    entry,
+    projectRoot,
+    depsLockFilePath,
+    runtime: Runtime.NODEJS_16_X,
+    architecture: Architecture.X86_64,
+  });
+
+  Annotations.fromStack(myStack).hasNoWarning('*',
+    'If you are relying on AWS SDK v3 to be present in the Lambda environment already, please explicitly configure a NodeJS runtime of Node 18 or higher. [ack: @aws-cdk/aws-lambda-nodejs:sdkV3NotInRuntime]',
+  );
+  Annotations.fromStack(myStack).hasNoWarning('*',
+    'If you are relying on AWS SDK v2 to be present in the Lambda environment already, please explicitly configure a NodeJS runtime of Node 16 or lower. [ack: @aws-cdk/aws-lambda-nodejs:sdkV2NotInRuntime]',
+  );
+});
+
 test('bundling with >= Node18 warns when sdk v3 is external', () => {
   Bundling.bundle(stack, {
     entry,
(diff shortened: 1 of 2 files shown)


packages/aws-cdk-lib/aws-lambda-nodejs/test/bundling.test.ts
```diff
@@ -882,6 +882,23 @@ test('bundling with <= Node16 warns when sdk v3 is external', () => {
...
+    runtime: Runtime.NODEJS_16_X,
+    architecture: Architecture.X86_64,
+  });
+
+  Annotations.fromStack(stack).hasNoWarning('*',
+    'If you are relying on AWS SDK v3 to be present in the Lambda environment already, please explicitly configure a NodeJS runtime of Node 18 or higher. [ack: @aws-cdk/aws-lambda-nodejs:sdkV3NotInRuntime]',
+  );
+  Annotations.fromStack(stack).hasNoWarning('*',
+    'If you are relying on AWS SDK v2 to be present in the Lambda environment already, please explicitly configure a NodeJS runtime of Node 16 or lower. [ack: @aws-cdk/aws-lambda-nodejs:sdkV2NotInRuntime]',
+  );
```
plumdog, line 899: Check for "does not produce a warning" this way seems very brittle, but I couldn't see a better way from looking at https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.assertions.Annotations.html.

plumdog:
Exemption Request

This fix only changes what warnings are generated, so no changes to integration tests or snapshots required.

MrArnoldPalmer (approved):
Was literally just working on this with the exact same solution and tests. LGTM! We should indeed move all the other unit tests to use their own unique `stack` as well.

filipeandre:
I'm not sure how can I disable this warning ? I'm using SDK v3 and I don't know how to disable the warning :\

PR #27026, opened 2023-09-06
chore(ec2): add verified permissions vpc interface endpoint

clueleaf:
Add Verified Permissions VPC Endpoint to `InterfaceVpcEndpointAwsService` so we can create one by using class property.

see: https://docs.aws.amazon.com/verifiedpermissions/latest/userguide/vpc-interface-endpoints.html

----


lpizzinidev (approved):
Thanks 👍

rix0rrr (approved):

PR #27029, opened 2023-09-06
feat(apprunner): add HealthCheckConfiguration property in Service

### LINKED ISSUE #26972: (apprunner-alpha): (Allow specify custom health endpoint)
### Describe the feature

By using the new apprunner alpha construct it is currently not possible to change the health configuration ourself.

This is the old apprunner health configuration: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apprunner.CfnService.HealthCheckConfigurationProperty.html

But in the apprunner alpha package https://docs.aws.amazon.com/cdk/api/v2/docs/aws-apprunner-alpha-readme.html there is now way to change the health endpoint that should be used.

### Use Case

I want to move a Spring Boot application from on-premise to apprunner. This already has an health endpoint in use called /actuator/health. I would like to use this for AppRunner instead of just the root path.

I want to use the alpha apprunner construct as it allows me to upload a docker asset as part of the CDK to make it easier to migrate to cloud for our existing applications.

### Proposed Solution

Allow health configuration to be set in the apprunner alpha construct

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.93.0

typescript

go-to-k:
This PR adds HealthCheckConfiguration property in Service construct.

Closes #26972.

----


packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -1074,6 +1147,10 @@ export class Service extends cdk.Resource implements iam.IGrantable {
       throw new Error('configurationValues cannot be provided if the ConfigurationSource is Repository');
     }
 
+    if (this.props.healthCheckConfiguration?.path && this.props.healthCheckConfiguration?.protocol !== HealthCheckProtocolType.HTTP) {
```
lpizzinidev, line 1150: Can you please also add validation for `HealthyThreshold`, `Interval`, `Timeout`, and `UnhealthyThreshold` min/max values according to [this](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apprunner-service-healthcheckconfiguration.html#cfn-apprunner-service-healthcheckconfiguration-healthythreshold) spec?

lpizzinidev (changes requested):
Thanks for the contribution!
Looks good overall, just a small addition on parameter validation.

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
go-to-k: @lpizzinidev 

Thanks for review! I missed!

I changed them, please see again.

https://github.com/aws/aws-cdk/pull/27029/commits/2a11dc4f2d510a68f8e9be333ef911890e21553c

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -1074,6 +1147,35 @@ export class Service extends cdk.Resource implements iam.IGrantable {
       throw new Error('configurationValues cannot be provided if the ConfigurationSource is Repository');
     }
 
+    if (this.props.healthCheckConfiguration?.path !== undefined) {
```
lpizzinidev, line 1150: The implementation is correct.
I'd create a separate function to handle the logic to keep the constructor cleaner, something like:
```
private validateHealthCheckConfiguration(healthCheckConfiguration?: HealthCheckConfiguration) {
    // Add guard to avoid using ? on every successive control
    if (!healthCheckConfiguration) return; 
    // Validate parameters...
}
```

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -1074,6 +1147,35 @@ export class Service extends cdk.Resource implements iam.IGrantable {
...
+      if (this.props.healthCheckConfiguration?.protocol !== HealthCheckProtocolType.HTTP) {
+        throw new Error('path is only applicable when you set Protocol to HTTP');
+      }
+      if (this.props.healthCheckConfiguration?.path.length === 0) {
+        throw new Error('path length must be greater than 0');
+      }
+    }
+    if (this.props.healthCheckConfiguration?.healthyThreshold !== undefined) {
+      if (this.props.healthCheckConfiguration?.healthyThreshold < 1 || this.props.healthCheckConfiguration?.healthyThreshold > 20) {
+        throw new Error('healthyThreshold must be greater than or equal to 1 and less than or equal to 20');
```
lpizzinidev, line 1160: We should add the received value in the error message as a feedback to the user and make the message more concise:
```
throw new Error(`healthyThreshold must be between 1 and 20, got ${healthCheckConfiguration.healthyThreshold}`);
```

packages/@aws-cdk/aws-apprunner-alpha/test/service.test.ts
```diff
@@ -1350,4 +1350,544 @@ test('addToRolePolicy', () => {
...
+        path: '',
+        protocol: apprunner.HealthCheckProtocolType.HTTP,
+        timeout: cdk.Duration.seconds(2),
+        unhealthyThreshold: 5,
+      },
+    });
+  }).toThrow('path length must be greater than 0');
+});
+
+test('OK if healthyThreshold is 1 in healthCheckConfiguration', () => {
```
lpizzinidev, line 1455: I think that we should only leave the unit tests for the failing cases of validation, we are already checking that the implementation is working in the `Service has healthCheckConfiguration` test.
Adding test cases for checking limit values is a bit redundant in my opinion.

lpizzinidev (changes requested):
Thanks for the fast implementation 💪
Looks good overall!
Just a couple of changes on code cleanup and message formatting.

lpizzinidev (changes requested):
Thanks! 
One last change on the formatting of the durations in the validation message

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -1259,4 +1337,48 @@ export class Service extends cdk.Resource implements iam.IGrantable {
...
+      }
+    }
+    if (healthCheckConfiguration.unhealthyThreshold !== undefined) {
+      if (healthCheckConfiguration.unhealthyThreshold < 1 || healthCheckConfiguration.unhealthyThreshold > 20) {
+        throw new Error(`unhealthyThreshold must be between 1 and 20, got ${healthCheckConfiguration.unhealthyThreshold}`);
+      }
+    }
+    if (healthCheckConfiguration.interval !== undefined) {
+      if (healthCheckConfiguration.interval.toSeconds() < 1 || healthCheckConfiguration.interval.toSeconds() > 20) {
+        throw new Error(`interval must be between 1 and 20 seconds, got ${healthCheckConfiguration.interval}`);
```
lpizzinidev, line 1375: `interval` and `timeout` should be converted with `toSeconds()`

go-to-k: OK, I changed!

https://github.com/aws/aws-cdk/pull/27029/commits/629b13113dd5282b369b196417b7c3e443f7f44f

lpizzinidev (approved):

packages/@aws-cdk/aws-apprunner-alpha/README.md
```diff
@@ -215,3 +215,24 @@ const service = new apprunner.Service(stack, 'Service', {
...
+  }),
+  healthCheckConfiguration: {
+    healthyThreshold: 5,
+    interval: Duration.seconds(10),
+    path: '/',
+    protocol: apprunner.HealthCheckProtocolType.HTTP,
+    timeout: Duration.seconds(10),
+    unhealthyThreshold: 10,
+  },
+});
```
mrgrain, line 237: I'd like us to use an enum like class for this, because HTTP and TCP health checks have different props:
```suggestion
new apprunner.Service(this, 'Service', {
  source: apprunner.Source.fromEcrPublic({
    imageConfiguration: { port: 8000 },
    imageIdentifier: 'public.ecr.aws/aws-containers/hello-app-runner:latest',
  }),
  healthCheck: apprunner.HealthCheck.http({
    healthyThreshold: 5,
    interval: Duration.seconds(10),
    path: '/',
    timeout: Duration.seconds(10),
    unhealthyThreshold: 10,
  }),
});
```

Have a look at `appmesh.HealthCheck`

mrgrain (changes requested):

packages/@aws-cdk/aws-apprunner-alpha/README.md
go-to-k: @mrgrain 

I changed to use the `HealthCheck` class like `appmesh.HealthCheck`.

https://github.com/aws/aws-cdk/pull/27029/commits/d0cd25506a66ab4c62547827fbe7410a40891263

rix0rrr (approved):

PR #27031, opened 2023-09-06
fix(eks): use Python 3.10 runtime for kubectl-provider

### LINKED ISSUE #26994: aws-eks: do not use Python 3.7 Lambda runtime for the kubectl provider
### Describe the bug

AWS emailed me about the deprecation of the Python 3.7 Lambda runtime. Inspecting resources, found that - even where I'm on latest CDK - still have one using 3.7 due to the kubectl provider: https://github.com/aws/aws-cdk/blob/v2.94.0/packages/aws-cdk-lib/aws-eks/lib/kubectl-provider.ts#L138

(A quick review of https://github.com/search?q=repo%3Aaws%2Faws-cdk+PYTHON_3_7+language%3ATypeScript+path%3A%2F%5Epackages%5C%2Faws-cdk-lib%5C%2F%2F&type=code suggests this is the only remaining such Lambda, hence creating an issue for this Lambda specifically, rather than a general one.)

### Expected Behavior

For the CDK not to use Python 3.7 as it has an impending EOL.

### Current Behavior

Uses Python 3.7.

### Reproduction Steps

Deploy an EKS cluster with latest CDK (2.94.0), see it creates a Lambda for the kubectl provider handler that uses Python 3.7.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.94.0

### Framework Version

_No response_

### Node.js Version

n/a

### OS

n/a

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

plumdog:
Use Python 3.10 runtime for kubectl-provider. The previously used version is now EOL and will be deprecated soon in AWS Lambda.

Note that while newer Python runtime versions are already available, 3.10 is the newest runtime available in all partitions (including GovCloud and CN).

Closes #26994.

----


packages/aws-cdk-lib/aws-eks/lib/kubectl-provider.ts
```diff
@@ -135,7 +135,7 @@ export class KubectlProvider extends NestedStack implements IKubectlProvider {
 
     const handler = new lambda.Function(this, 'Handler', {
       code: lambda.Code.fromAsset(path.join(__dirname, 'kubectl-handler')),
-      runtime: lambda.Runtime.PYTHON_3_7,
+      runtime: lambda.Runtime.PYTHON_3_11,
```
plumdog, line 138: @pahud re https://github.com/aws/aws-cdk/issues/26994#issuecomment-1706777375, I couldn't think of a reason not to pick the latest runtime.

I have checked that all the existing code is syntactically valid in Python 3.11.

There might be reasons, like regional availability, to pick a less recent version, but I'm unsure how to determine that.

plumdog:
Clarification Request

- Maybe this is not a feature? Regardless, I don't think it warrants a change to the README.
- It doesn't appear that the unit tests check for this
- I'm unable to run the integration tests.

plumdog:
I lieu of running the integration tests, I have modified a kubectl provider Lambda deployed by CDK and found that it was able to run and upgrade a helm chart for me, so I am confident it works.

plumdog:
Exemption Request

mrgrain:
Thanks @plumdog I'll pick this up soon. Apologies for the delay. We were operating with reduced capacity the last couple weeks.

mrgrain:
>Is Python 3.11 the right runtime? I couldn't find anything confirming which Lambda runtimes are available in which regions. The blog post https://aws.amazon.com/about-aws/whats-new/2023/07/aws-lambda-python-3-11/ says "The Python 3.11 runtime is available in all Regions where Lambda is available, except for China and GovCloud Regions." which was published 27th July 2023.

We have to use Python 3.10 for now. Once Python 3.11 is cleared for the remaining regions, we can switch over.

> I have not run the integration tests because - based on https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#integration-tests - I understand doing so will attempt to deploy some AWS resources. I don't know what the billing implications of this would be, but I don't have an AWS account where I'm happy to chance this.

I can do this.

mrgrain (approved):

mrgrain (approved):

PR #27034, opened 2023-09-06
fix(events-targets): AwsApi fails when SDKv3 Command expects Uint8Array

mrgrain:
In #27002 we have migrated the AwsApi Construct to SDKv3.
However some SDKv3 Commands expect a `Uint8Array` in places where SDKv2 previously would accept a `string`.
This is a problem because handler input is passed directly into the SDK calls and the input can only contain JSON data types.
To solves this, we keep a list of SDK actions and parameters that should be coerced to `Uint8Array`.

----


packages/@aws-cdk/sdk-v2-to-v3-adapter/lib/coerce-api-parameters.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export function coerceToUint8Array(obj: any, path: string[]): any {
+  const currentKey = path?.[0];
+  if (!currentKey) {
+    return;
+  }
+
+  // recurse into array
+  if (currentKey === '*' && Array.isArray(obj)) {
+    obj.map(e => coerceToUint8Array(e, path.slice(1)));
```
rix0rrr, line 34: This doesn't work if the array contains the coercible values directly (instead of the array containing objects that have the coercible values). 

You'll need to `return` the new map.

packages/@aws-cdk/sdk-v2-to-v3-adapter/lib/coerce-api-parameters.ts
```diff
@@ -0,0 +1,54 @@
...
+  const pathsToCoerce = UINT8ARRAY_PARAMETERS?.[service.toLowerCase()]?.[action.toLowerCase()] ?? [];
+  for (const path of pathsToCoerce) {
+    coerceToUint8Array(parameters, path.split('.'));
+  }
+  return parameters;
+}
+
+/**
+ * Given an object and a path, traverse to the leaf and coerce the value into a Uint8Array.
+ * This function mutates the original object.
```
rix0rrr, line 19: I think this implementation can be made simpler if we return the coerced values, but make no guarantee that we won't mutate the input objects.

So for primitives and arrays we return a fresh value, but we mutate objects in place (and then return it at the end).

packages/@aws-cdk/sdk-v2-to-v3-adapter/lib/coerce-api-parameters.ts
```diff
@@ -0,0 +1,54 @@
...
+  // recurse into array
+  if (currentKey === '*' && Array.isArray(obj)) {
+    obj.map(e => coerceToUint8Array(e, path.slice(1)));
+    return;
+  }
+
+  // Check if we expect the leaf to be an array
+  // We need to change this before traversing down, otherwise we can't mutate
+  if (path.length === 2 && path[1] === '*' && Array.isArray(obj[currentKey])) {
+    obj[currentKey] = obj[currentKey].map((e: any) => new TextEncoder().encode(e));
```
rix0rrr, line 41: Encoding is done in multiple code paths. It would be nicer if it's in one place at the top of the function, which indicates the end of the recursion:

```ts
export function coerceToUint8Array(obj: any, path: string[]): any {
  if (path.length === 0) {
    return coerceValueToUint8Array(obj);
  }
  
  if (path[0] === '*' && Array.isArray(obj)) {
    return obj.map((e) => coerceToUint8Array(e, path.slice(1));
  }

  if (obj && typeof obj === 'object') {
    if (path[0] in obj) {
      obj[path[0]] = coerceToUint8Array(obj[path[0]], path.slice(1));
    }
    return obj;
  }
  return obj; 
}

function coerceValueToUint8Array(x: unknown): Uint8Array {
  if (x instanceof Uint8Array) { 
    return x;
  }
  // FIXME: need to think about Buffer?
  return new TextEncoder().encode(e);
}
```

rix0rrr (approved):

PR #27036, opened 2023-09-06
feat(ecs): Support specifying revision of task definition

### LINKED ISSUE #26983: (aws-ecs): Support specifying latest revision of task family in service instead of specific revision
### Describe the feature

CloudFormation allows you to specify a task family without a revision when creating a service to have it use the latest active task definition. However, when passing a task definition to a service in the CDK, it always uses the ARN of the specific revision

### Use Case

I use the ECS deployment action in CodePipeline to automatically create a new task revision on deployment, as well as storing the image ID in an SSM parameter so that if I update the task definition CDK-side, it uses the correct (most recent) image ID. However, if the service needs to be redeployed without changing the task definition, the service will be redeployed with the last revision of the task that was deployed via CDK/CFN, which rolls back the image (which is unacceptable as it could cause critical application failures).

### Proposed Solution

Add a parameter to the service constructor called something like `taskRevision`, allowing it to specify a specific revision or `latest`, using that revision and the family from the task definition instead of the .

An alternative could be #7863 (where you create the task definition, construct an arn from that, import the resource, then pass that instead of the task def), but in this case we already have the task definition we want to use, so that seems to require more workarounds.

### Other Information

Yes, I understand the recommended approach is to have the pipeline redeploy the CDK stack with a parameter specifying the image as a means to keep the CDK app itself deterministic, however my position is that is substantial added complexity (namely, in the mental gymnastics needed for someone to understand it) for little concrete benefit - the end result is the same, and this approach means that there is truly no drift. Additionally the current behavior can cause critical application failures if an old image is accidentally deployed.

### Acknowledgements

### CDK version used

2.93.0

N/A

luxaritas:
If using CodePipeline EcsDeployAction without using the CODE_DEPLOY deployment controller, future deployments of an ECS service will revert the task definition to the task definition deployed by CloudFormation, even though the latest active revision created by the deploy action is the one that is intended to be used. This provides a way to specify the specific revision of a task definition that should be used, including the special value `latest` which uses the latest ACTIVE revision.

Closes #26983.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
+++ b/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
@@ -1241,11 +1241,6 @@ export function isExternalCompatible(compatibility: Compatibility): boolean {
  */
 export class TaskDefinitionRevision {
   /**
-   * The string representation of this revision
-   */
-  public readonly revision: string;
-
-  /**
    * The most recent revision of a task
    */
   public static readonly LATEST = new TaskDefinitionRevision('latest');
@@ -1259,6 +1254,11 @@ export class TaskDefinitionRevision {
     }
     return new TaskDefinitionRevision(revision.toString());
   }
+
+  /**
+   * The string representation of this revision
+   */
+  public readonly revision: string;
 
   private constructor(revision: string) {
     this.revision = revision;


packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -643,6 +651,14 @@ export abstract class BaseService extends Resource
...
     }
 
+    if (props.taskDefinitionRevision) {
+      this.resource.taskDefinition = taskDefinition.family;
+      if (props.taskDefinitionRevision !== 'latest') {
+        this.resource.taskDefinition += `:${props.taskDefinitionRevision}`;
+      }
+      this.node.addDependency(taskDefinition);
+    }
+
```
lpizzinidev, line 661:
```suggestion
    } else if (props.taskDefinitionRevision) {
      if (props.taskDefinitionRevision !== 'latest' && props.taskDefinitionRevision < 1) {
        throw new Error(`taskDefinitionRevision must be 'latest' or a positive number, got ${props.taskDefinitionRevision}`);
      }
      this.resource.taskDefinition = taskDefinition.family + (props.taskDefinitionRevision !== 'latest' 
 `:${props.taskDefinitionRevision}` : '');
      this.node.addDependency(taskDefinition);
    }

```
Should the revision number be enforced only for non-`CODE_DEPLOY` types?
Also, I would add validation for the revision when numeric (and a relative unit test).

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -345,6 +345,14 @@ export interface BaseServiceOptions {
    * cannot make requests to other services via Service Connect.
    */
   readonly serviceConnectConfiguration?: ServiceConnectProps;
+
+  /**
+   * The revision of the service's task definition to use for tasks in this service
+   * or 'latest' to use the latest ACTIVE task revision
```
lpizzinidev, line 351:
```suggestion
   * Revision number for the task definition or `latest` to use the latest active task revision.
```

lpizzinidev (changes requested):
Thanks for the implementation (and sorry for the late review) 👍
Some minor adjustments are needed in my opinion.
Not sure about the behavior for `DeploymentControllerType.CODE_DEPLOY` so feel free to clarify that.
Also, are you sure that it is necessary to explicitly set `latest`?
Looks like the [ECS service](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html#cfn-ecs-service-taskdefinition) defaults to the latest active value already.

luxaritas:
Thanks for the response - some good points. With respect to the comments not mentioned inline:
> Not sure about the behavior for DeploymentControllerType.CODE_DEPLOY so feel free to clarify that.

See https://github.com/aws/aws-cdk/blob/c551a643fdf5e2e6e8a254089ee0eedd7350c32d/packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts#L638-L644

> Also, are you sure that it is necessary to explicitly set latest?
Looks like the [ECS service](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html#cfn-ecs-service-taskdefinition) defaults to the latest active value already.

Yes - and indeed it is already done in the code I linked above. I believe the reason why it is necessary is because in the concrete implementations of BaseService (FargateService/Ec2Service), the task _arn_ is passed as part of additionalProps, and the arn includes the revision

packages/aws-cdk-lib/aws-ecs/README.md
```diff
@@ -610,6 +610,37 @@ const service = new ecs.ExternalService(this, 'Service', {
...
+```ts
+declare const cluster: ecs.Cluster;
+declare const taskDefinition: ecs.TaskDefinition;
+
+const service = new ecs.ExternalService(this, 'Service', {
+  cluster,
+  taskDefinition,
+  desiredCount: 5,
+  taskDefinitionRevision: 'latest'
+});
```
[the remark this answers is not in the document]
luxaritas: To make sure I understand correctly, you think the description of the `latest` special case and why it would be used is unnecessary/too much information?

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
luxaritas: Hmmm... initially I didn't see any reason why the revision can't be made more specific than `latest` for `CODE_DEPLOY` - it's only as it is so that it's not _whatever the latest version is at the time of deployment_. However, the entire point of `CODE_DEPLOY` is that it's being deployed from code deploy, so if a new deployment is triggered it would override this anyways. I could maybe see a use case for if you want to temporarily pin and just not run the pipeline, but that seems rare and likely to cause more confusion than benefit. So you're probably right.

Good point on the additional validation

packages/aws-cdk-lib/aws-ecs/README.md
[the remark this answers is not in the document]
lpizzinidev: No, but I think that documentation needs to be clear and direct.
Feel free to add examples or use cases to my suggestion.

lpizzinidev:
@luxaritas 

> Yes - and indeed it is already done in the code I linked above. I believe the reason why it is necessary is because in the concrete implementations of BaseService (FargateService/Ec2Service), the task arn is passed as part of additionalProps, and the arn includes the revision

Ok, [got it](https://github.com/aws/aws-cdk/blob/0deed4dda79f360ab23535f90aa32c8c637adafa/packages/aws-cdk-lib/aws-ecs/lib/fargate/fargate-service.ts#L139).
Thanks for the follow-up and clarifications 👍

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
lpizzinidev: > the entire point of CODE_DEPLOY is that it's being deployed from code deploy, so if a new deployment is triggered it would override this anyways

If this is the case it's better to keep the `else if` then, otherwise `this.resource.taskDefinition` would get overridden if `taskDefinitionRevision` is numeric.

luxaritas: In this case, I'll also throw an error if a non-latest revision is specified and the deployment type is CODE_DEPLOY, to prevent confusion around explicitly asking for something different and it silently not doing that

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -345,6 +345,14 @@ export interface BaseServiceOptions {
...
+   *
+   * @default - Uses the revision of the passed task definition deployed by CloudFormation
+   */
+  readonly taskDefinitionRevision?: string;
```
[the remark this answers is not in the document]
luxaritas: Isn't this not allowed by CDK guidelines? https://github.com/aws/aws-cdk/blob/94ccec65f7a8515fc32a01100924fbec9efc72b3/docs/DESIGN_GUIDELINES.md?plain=1#L660-L675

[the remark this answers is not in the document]
luxaritas: Ive attempted to do what the guidelines suggest and use a class with static methods, though would appreciate a look over how I approached that

luxaritas:
Sorry for the delay in revisiting this. Made changes based on your comments. I've adjusted the way taskDefinitionRevision is typed per the comment thread above, let me know if it's still not right.

packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
```diff
@@ -1234,3 +1234,29 @@ export function isFargateCompatible(compatibility: Compatibility): boolean {
...
+  static revision(revision: number) {
+    return new TaskDefinitionRevision(revision.toString());
+  }
+
+  public readonly revisionId: string;
+
+  constructor(revision: string) {
+    this.revisionId = revision;
+  }
+}
```
lpizzinidev, line 1262:
```suggestion

/**
 * Represents a task definition revision.
 */
export class TaskDefinitionRevision {
  /**
   * The most recent revision of a task
   */
  public static readonly LATEST = new TaskDefinitionRevision('latest');

  /**
   * Specific revision of a task
   */
  public static of(revision: number) {
    if (revision < 1) {
      throw new Error(`A task definition revision must be 'latest' or a positive number, got ${revision}`);
    }
    return new TaskDefinitionRevision(revision.toString());
  }

  private constructor(public readonly revision: string) {}
}
```
Let's follow the enum-class pattern for this.
Thanks for the clarification on union types 👍
A unit test is needed on revision number validation.

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -635,12 +647,26 @@ export abstract class BaseService extends Resource
...
     if (props.deploymentController?.type === DeploymentControllerType.CODE_DEPLOY) {
       // Strip the revision ID from the service's task definition property to
       // prevent new task def revisions in the stack from triggering updates
       // to the stack's ECS service resource
       this.resource.taskDefinition = taskDefinition.family;
       this.node.addDependency(taskDefinition);
+    } else if (props.taskDefinitionRevision) {
+      this.resource.taskDefinition = taskDefinition.family;
+      if (props.taskDefinitionRevision.revisionId !== 'latest') {
+        this.resource.taskDefinition += `:${props.taskDefinitionRevision.revisionId}`;
```
lpizzinidev, line 667:
```suggestion
      if (props.taskDefinitionRevision !== TaskDefinitionRevision.LATEST) {
        this.resource.taskDefinition += `:${props.taskDefinitionRevision.revision}`;
```

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -635,12 +647,26 @@ export abstract class BaseService extends Resource
...
     }
 
+    if (
+      props.deploymentController?.type === DeploymentControllerType.CODE_DEPLOY
+      && props.taskDefinitionRevision
+      && props.taskDefinitionRevision.revisionId !== 'latest'
+    ) {
+      throw new Error('CODE_DEPLOY deploymentController cannot be used with a non-latest task definition revision');
+    }
+
```
lpizzinidev, line 657:
```suggestion
```
I don't think it's necessary.
If `props.deploymentController?.type === DeploymentControllerType.CODE_DEPLOY` we will run the first `if` statement and `taskDefinitionRevision` will just get ignored.

lpizzinidev (changes requested):
Thanks 👍
Some adjustments for the added data structure are needed.

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
luxaritas: The reason I did this is because if the user _specifically asked_ for revision 23, but they got `latest` instead, that seems like an unintended surprise/footgun

lpizzinidev: Thanks for clarifying, however, I think it's overkill to fail a deployment that would work.
Let's use `Annotations.addWarningV2` to provide some feedback and be more user-friendly.

luxaritas: Warning sounds completely fair - will do!

lpizzinidev (changes requested):
Thanks for the changes!
Some minor cleanup and it will be good to go for me.
Note https://github.com/aws/aws-cdk/pull/27036#discussion_r1390250403 from the previous review.

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -635,12 +647,26 @@ export abstract class BaseService extends Resource
       throw new Error('Deployment alarms requires the ECS deployment controller.');
     }
 
+    if (
+      props.deploymentController?.type === DeploymentControllerType.CODE_DEPLOY
+      && props.taskDefinitionRevision
+      && props.taskDefinitionRevision.revisionId !== 'latest'
```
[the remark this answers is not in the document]
luxaritas: 🤦Not sure how I missed that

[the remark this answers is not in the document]
luxaritas: Actually there is a reason for doing it the other way - someone could `new TaskDefinitionRevision('latest')`. However, we can do TaskDefinitionRevision.LATEST.revision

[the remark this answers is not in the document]
luxaritas: ...which is not a problem if the constructor is private like it should be 🤦. Fixed

lpizzinidev (approved):

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
paulhcsun: I disagree here, I feel like an error is reasonable even though the deployment would work because we are changing a setting that the user had set and not actually running the revision the user had specified.

paulhcsun:
Overall looks good to me, just left one comment regarding changing the error to a warning. Thanks for the contribution and thanks @lpizzinidev for reviewing!

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
luxaritas: That was my gut feeling here. I'll revert that commit unless I'm told otherwise

luxaritas:
> [Container] 2023/12/13 22:28:16.388276 Running command npm config set unsafe-perm true
npm ERR! `unsafe-perm` is not a valid npm option

Nothing I did, is this a situation where I just need to pull in the latest changes from main?

paulhcsun:
I'm seeing the following linting error in the build logs, but it's strange because a docstring is included.
```
aws-cdk-lib: error: [awslint:docs-public-apis:aws-cdk-lib.aws_ecs.TaskDefinitionRevision.revision] Public API element must have a docstring 
aws-cdk-lib: Error: /codebuild/output/src2582654880/src/github.com/aws/aws-cdk/tools/@aws-cdk/cdk-build-tools/bin/cdk-awslint exited with error code 1
aws-cdk-lib: Build failed.!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

luxaritas:
I think it's the public member declared in the constructor?

packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
```diff
@@ -1234,3 +1234,26 @@ export function isFargateCompatible(compatibility: Compatibility): boolean {
 export function isExternalCompatible(compatibility: Compatibility): boolean {
   return [Compatibility.EXTERNAL].includes(compatibility);
 }
+
+/**
+ * Represents revision of a task definition, either a specific numbered revision or
+ * the specia "latest" revision
```
paulhcsun, line 1240:
```suggestion
 * the `latest` revision
```

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -635,12 +647,26 @@ export abstract class BaseService extends Resource
       throw new Error('Deployment alarms requires the ECS deployment controller.');
     }
 
+    if (
+      props.deploymentController?.type === DeploymentControllerType.CODE_DEPLOY
+      && props.taskDefinitionRevision
+      && props.taskDefinitionRevision !== TaskDefinitionRevision.LATEST
+    ) {
+      throw new Error('CODE_DEPLOY deploymentController cannot be used with a non-latest task definition revision');
```
paulhcsun, line 655:
```suggestion
      throw new Error('CODE_DEPLOY deploymentController can only be used with the `latest` task definition revision');
```

paulhcsun (commented):
A couple minor wording changes.

paulhcsun:
> I think it's the public member declared in the constructor? I'm assuming that needs to be pulled out in order to add a docstring

That could be it, could you give that a try please?

paulhcsun (approved):
Great work @luxaritas, thanks for the contribution!

PR #27040, opened 2023-09-07
fix(rds): unable to create cluster with per-instance parameters (under feature flag)

### LINKED ISSUE #27035: aws_rds: unable to use per-instance parameters
### Describe the bug

When trying to use the `parameters` property during `ClusterInstance.provisioned`, it will try to create a ParameterGroup for you using the id `InstanceParameterGroup`.

If you try to have multiple instances (ie multiple readers) in a DatabaseCluster, this will throw error:
```
Error: There is already a Construct with name 'InstanceParameterGroup' in DatabaseCluster [Database]
```

Here's the specific line of code:
https://github.com/aws/aws-cdk/blob/v2.94.0/packages/aws-cdk-lib/aws-rds/lib/aurora-cluster-instance.ts#L457-L464

### Expected Behavior

I should be able to provide a simple parameters object and a parameter group to be created for each unique instance.

### Current Behavior

Throws an error

### Reproduction Steps

Create a `DatabaseCluster` with multiple readers with the `parameters: {}` property set:

```
ClusterInstance.provisioned(`Instance${id}`, {
         parameters: {}
    });
```

### Possible Solution

The ParameterGroup scope is the cluster, either the scope should be specific to the ClusterInstance, or the ParameterGroup should have a unique ID.

_No response_

### CDK CLI Version

2.94.0

### Framework Version

_No response_

### Node.js Version

18

### OS

mac osx latest

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

go-to-k:
This PR fixes that DatabaseCluster cannot create multiple readers with each parameters.

Closes #27035.

----


go-to-k:
Exemption Request because I think this bug can be covered in an unit test.

packages/aws-cdk-lib/aws-rds/lib/aurora-cluster-instance.ts
```diff
@@ -456,7 +456,7 @@ class AuroraClusterInstance extends Resource implements IAuroraClusterInstance {
 
     const instanceParameterGroup = props.parameterGroup ?? (
       props.parameters
-        ? new ParameterGroup(props.cluster, 'InstanceParameterGroup', {
+        ? new ParameterGroup(props.cluster, `${id}InstanceParameterGroup`, {
```
go-to-k, line 459: This code change will replace the logical name of a ParameterGroup that has already been created by specifying `props.parameters` so far.

If we want to prevent this, I think there is another way.

- Check if 'InstanceParameterGroup' has already been created in the stack.
  - e.g. `if (Stack.of(this).node.tryFindChild('InstanceParameterGroup') === undefined)`
- Create 'InstanceParameterGroup' only if it is.
- If not, create an `${id}InstanceParameterGroup`.

However, this is not intuitive, and unexpected resource substitutions will occur if the order in the `readers` array is changed. Also, if we create an 'InstanceParameterGroup' ParameterGroup elsewhere, there will be a conflict.

For these reasons, I stopped.

mrgrain (dismissed):
Thanks @go-to-k for this change!

I think we should just put this behind a feature flag. That makes it the default for new apps and opt-in for existing apps.

packages/aws-cdk-lib/aws-rds/lib/aurora-cluster-instance.ts
mrgrain, line 459: I reckon this would be better:
```suggestion
        ? new ParameterGroup(this, `InstanceParameterGroup`, {
```

go-to-k: OK, I changed!

go-to-k:
@mrgrain 

Thanks for your review!

> I think we should just put this behind a feature flag. That makes it the default for new apps and opt-in for existing apps.

I have never had a modification that required a feature flag. Is there anything I should do with this PR?

mrgrain:
> I have never had a modification that required a feature flag. Is there anything I should do with this PR?

Exciting! The required steps are described in the [contributing guide](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#feature-flags). LMK if you run into any issues.

go-to-k:
@mrgrain 

I tried to respond to the feature flag and an integ test. Please check if there is anything wrong.

rix0rrr (approved):

rix0rrr (approved):

PR #27043, opened 2023-09-07
chore: update SDKv3 blob types model

rix0rrr:
Add a script that will automatically extract the blob parameter type data from the AWS SDKv3.

`scripts/update-sdkv3-blobs-model.sh` will download the files and drive `scripts/update-sdkv3-blobs-model.ts`, which does the actual collection.

----

### DIFF
--- a/packages/aws-cdk/lib/api/bootstrap/deploy-bootstrap.ts
+++ b/packages/aws-cdk/lib/api/bootstrap/deploy-bootstrap.ts
@@ -7,6 +7,7 @@ import { BOOTSTRAP_VERSION_OUTPUT, BootstrapEnvironmentOptions, BOOTSTRAP_VERSIO
 import * as logging from '../../logging';
 import { Mode, SdkProvider, ISDK } from '../aws-auth';
 import { deployStack, DeployStackResult } from '../deploy-stack';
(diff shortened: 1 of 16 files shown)


packages/@aws-cdk/sdk-v2-to-v3-adapter/lib/parameter-types.ts
```diff
@@ -0,0 +1,561 @@
+// This file was generated from the aws-sdk-js-v3 at Thu Sep 07 2023 13:59:10 GMT+0200 (Central European Summer Time)
+export interface BlobTypeMapping {
```
mrgrain, line 2:
```suggestion
export interface TypeCoercionMap {
```

packages/@aws-cdk/sdk-v2-to-v3-adapter/lib/parameter-types.ts
```diff
@@ -0,0 +1,561 @@
...
+  [service: string]: {
+    [action: string]: string[]
+  }
+};
+export const blobTypes: BlobTypeMapping = {
```
mrgrain, line 7:
```suggestion
export const UINT8ARRAY_PARAMETERS: TypeCoercionMap = {
```

mrgrain (approved):

PR #27051, opened 2023-09-07
feat(stepfunctions): properties for ErrorPath, CausePath, MaxDelay, Jitter

kaizencc:
The Fail state supports JsonPath values in the `ErrorPath` and `CausePath` field to filter the state input and serve as the state error and cause.

The Retry block supports exponential backoff with Jitter and capped interval retry growth with MaxDelay. 

----

### DIFF
--- a/packages/aws-cdk-lib/aws-stepfunctions/lib/types.ts
+++ b/packages/aws-cdk-lib/aws-stepfunctions/lib/types.ts
@@ -139,14 +139,14 @@ export interface RetryProps {
   /**
    * Maximum limit on retry interval growth during exponential backoff.
    *
-   * @default No max delay
+   * @default - No max delay
    */
   readonly maxDelay?: Duration;
 
   /**
    * Introduces a randomization over the retry interval.
    *
-   * @default No jitter strategy
+   * @default - No jitter strategy
    */
   readonly jitterStrategy?: JitterType;
 


packages/aws-cdk-lib/aws-stepfunctions/lib/types.ts
```diff
@@ -121,6 +136,20 @@ export interface RetryProps {
...
+   * Maximum limit on retry interval growth during exponential backoff.
+   *
+   * @default No max delay
+   */
+  readonly maxDelay?: Duration;
+
+  /**
+   * Introduces a randomization over the retry interval.
+   *
+   * @default No jitter strategy
```
kaizencc, line 149:
```suggestion
   * @default - No jitter strategy
```

packages/aws-cdk-lib/aws-stepfunctions/lib/types.ts
```diff
@@ -121,6 +136,20 @@ export interface RetryProps {
    */
   readonly maxAttempts?: number;
 
+  /**
+   * Maximum limit on retry interval growth during exponential backoff.
+   *
+   * @default No max delay
```
kaizencc, line 142:
```suggestion
   * @default - No max delay
```

MrArnoldPalmer (approved):

PR #27059, opened 2023-09-08
feat(batch): Stabilize Batch

### LINKED ISSUE #27054: (@aws-cdk/aws-batch-alpha): (unable to update maxVcpus)
### Describe the bug

We are using Fargate compute environment and we are unable to update the max vcpus on CFN. 
We get the error 
```
The following parameters can\'t be updated for Fargate compute environment: allocationStrategy, bidPercentage, ec2Configuration, ec2KeyPair, imageId, instanceRole, instanceTypes, launchTemplate, placementGroup, tags, type, updatePolicy, updateToLatestImageVersion. (Service: Batch, Status Code: 400, Request ID: 5bf1284d-eefb-42c1-921a-41e6d0b44565)
```

### Expected Behavior

Should update the maxVcpus directly

### Current Behavior

Throws the error
```
The following parameters can\'t be updated for Fargate compute environment: allocationStrategy, bidPercentage, ec2Configuration, ec2KeyPair, imageId, instanceRole, instanceTypes, launchTemplate, placementGroup, tags, type, updatePolicy, updateToLatestImageVersion. (Service: Batch, Status Code: 400, Request ID: 5bf1284d-eefb-42c1-921a-41e6d0b44565)
```

### Reproduction Steps

1. Make a Compute environment with some vcpu setting something like:
```
 new FargateComputeEnvironment(this, batchComputeEnvironmentName, {
            enabled: true,
            maxvCpus: 512,
            vpc: props.vpc,
            computeEnvironmentName: batchComputeEnvironmentName,
        });
```

2. Deploy this to an AWS account
3. Change the maxVcpu value to be 2048 or something different.
4. Deploy again.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

V2

### Framework Version

_No response_

### Node.js Version

18.16.1

### OS

Linux

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

comcalvi:
graduates the alpha module.

All users of any `ManagedComputeEnvironment` who left `updateToLatestImageVersion` unspecified will see it default to `undefined` instead of `true`. *If you use `FargateComputeEnvironment`, this upgrade may cause deployment errors; destroy the Fargate CE and recreate it to resolve this, if encountered.* 

Related: #27054.

----


packages/aws-cdk-lib/aws-batch/jest.config.js
TheRealAmazonKendra, line 1: We don't want this file moved.

packages/aws-cdk-lib/aws-batch/rosetta/default.ts-fixture
TheRealAmazonKendra, line 1: This file shouldn't be within batch. It should be within the rosetta folder that then contains a folder for each module.

TheRealAmazonKendra (changes requested):
I didn't look at the specifics of the src or tests since these are just ported over but I do have a few comments about files that ended up in the wrong spot. Additionally. I think we should be calling this a feat instead of a chore and not include the breaking change in the text. batch didn't have anything in it so it can't be breaking.

packages/aws-cdk-lib/aws-batch/lib/managed-compute-environment.ts
```diff
@@ -88,7 +88,12 @@ export interface IManagedComputeEnvironment extends IComputeEnvironment, ec2.ICo
    *
    * If you specify a specific AMI, this property will be ignored.
    *
-   * @default true
+   * Note: the CDK will never set this value by default, `false` will set by CFN.
+   * This is to avoid a deployment failure that occurs when this value is set.
+   *
+   * @see https://github.com/aws/aws-cdk/issues/27054
+   *
+   * @default false
```
mrgrain, line 96: I might be missing something here, but the linked issue makes it sound like setting this prop to `true` can break the deployment. 

Can we fix the underlying issue before stabilization?

mrgrain (dismissed):

packages/aws-cdk-lib/aws-batch/lib/managed-compute-environment.ts
comcalvi: The underlying issue is in CFN, and this is the fix. The problem is that whenever a ComputeEnvironment is created, the CFN handler first calls CreateCE, and then UpdateCE. The issue with this is that CreateCE does not set certain properties (in this case, this prop). UpdateCE is given these ignored properties, but UpdateCE sees that this property has changed; and, as the error indicates, these properties are not allowed to be changed.

CDK's default value is causing this issue by passing it in as `true` by default. The same issue would occur if it was `false`. This only occurs with `Fargate`, but the service team suggested the value not be set by default anywhere, so that is what this does. 

The issue for this is really a needs-cfn, I'll update the issue.

comcalvi:
@TheRealAmazonKendra true but it's breaking from the alpha module

it is also breaking the alpha module to move all of this code out

TheRealAmazonKendra (approved):

PR #27060, opened 2023-09-08
fix(route53): timeouts due to delete-existing-record-set-handler's wait time

pistazie:
This PR updates the wait time of the `delete-existing-record-set-handler` from 60 to 1800 seconds.

Commit 399b6bbdbfbf6440145e6d78e18898aae6261702 upgraded the handler to use the AWS SDK V3 (was V2) and made the following change:

```
-  await route53.waitFor('resourceRecordSetsChanged', { Id: changeResourceRecordSets.ChangeInfo.Id }).promise();
+  await waitUntilResourceRecordSetsChanged({ client: route53, maxWaitTime: 60 }, { Id: changeResourceRecordSets?.ChangeInfo?.Id })
```

The V2 `route53.waitFor` waits for 1800 seconds (see [reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Route53.html#waitFor-property))
> `Waits for the resourceRecordSetsChanged state by periodically calling the underlying [Route53.getChange()](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Route53.html#getChange-property) operation every 30 seconds (at most 60 times).` 

The new explicit configuration was set to 60 seconds which is too short and caused timeouts and failure in our cloudformation stacks

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-route53/test/integ.delete-existing-record-set.js.snapshot/cdk-route53-integ-delete-existing-record-set.assets.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-route53/test/integ.delete-existing-record-set.js.snapshot/cdk-route53-integ-delete-existing-record-set.assets.json
@@ -1,20 +1,20 @@
 {
   "version": "34.0.0",
   "files": {
-    "b3a6a7ffec7a4baf8647b80c08ebb1b91986d1beedfead0945b9de4b5ac368b6": {
+    "078a7e499edd0fc4a43e07567eb3cf0a1cfe2057083da97527eec2d81905ebd4": {
       "source": {
-        "path": "asset.b3a6a7ffec7a4baf8647b80c08ebb1b91986d1beedfead0945b9de4b5ac368b6",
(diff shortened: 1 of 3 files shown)


kaizencc:
hi @pistazie I went to update the snapshot for you but it looks like I don't have permissions to update to the `pitch-io` fork. 

Instead, I've created #27068. If you want, you can update the snapshot on this PR before the other one is merged and we'll go with yours. Otherwise, I'll add you as a collaborator on the other PR and get it in asap.

kaizencc:
@pistazie

https://github.com/aws/aws-cdk/pull/27068#discussion_r1320010721

Are you down to move the timeout to `890`? the custom resource lambda times out after 15 minutes, which is the max.

kaizencc:
fyi I went and called out this exact scenario in #27072. Basically, its really tough for us to accept contributions without allowing commits to your branch, since not only does that mean I can't do it but neither can our automation that syncs with main. In short, to accept contributions we need contributors to develop on a fork in their own personal account (the issue here), and not on their fork's `main` branch.

Going to close this over the other one for this exact reason, but thank you for the contribution and highlighting a nasty bug in our custom resource.

PR #27065, opened 2023-09-08
fix(custom-resources): AwsCustomResource fails when SDKv3 expects Uint8Array

rix0rrr:
The SDKv3 has changed the input representation of `blob` types from `string | Buffer | Uint8Array` to `Uint8Array`. That means all places where customers were passing in `strings` now fail to pass the right kind of value.

Similar to #27034, but for `AwsCustomResource`.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-backup/lib/rule.ts
+++ b/packages/aws-cdk-lib/aws-backup/lib/rule.ts
@@ -10,7 +10,7 @@ export interface BackupPlanRuleProps {
    * The duration after a backup job is successfully started before it must be
    * completed or it is canceled by AWS Backup.
    *
-   * @default - 8 hours
+   * @default - 7 days
    */
   readonly completionWindow?: Duration;
 


packages/@aws-cdk/custom-resource-handlers/test/custom-resources/aws-custom-resource/aws-sdk-v3-handler.test.ts
```diff
@@ -709,3 +711,37 @@ test('invalid v2 service name throws explicit error', async () => {
...
+          KeyId: 'key-id',
+          Plaintext: 'dummy-data',
+        },
+      } satisfies AwsSdkCall),
+    },
+  }, {} as AWSLambda.Context);
+
+  expect(kmsMock).toHaveReceivedCommandWith(EncryptCommand, {
+    KeyId: 'key-id',
+    Plaintext: new Uint8Array([
```
MrArnoldPalmer, line 741: wait the field named `Plaintext` is Uint8Array? 😕

MrArnoldPalmer (approved):

packages/@aws-cdk/custom-resource-handlers/test/custom-resources/aws-custom-resource/aws-sdk-v3-handler.test.ts
mrgrain: Yes!

PR #27068, opened 2023-09-08
fix(route53): timeouts due to delete-existing-record-set-handler's wait time

kaizencc:
This PR updates the wait time of the `delete-existing-record-set-handler` from 60 to 1800 seconds.

Commit 399b6bbdbfbf6440145e6d78e18898aae6261702 upgraded the handler to use the AWS SDK V3 (was V2) and made the following change:

```
-  await route53.waitFor('resourceRecordSetsChanged', { Id: changeResourceRecordSets.ChangeInfo.Id }).promise();
+  await waitUntilResourceRecordSetsChanged({ client: route53, maxWaitTime: 60 }, { Id: changeResourceRecordSets?.ChangeInfo?.Id })
```

The V2 `route53.waitFor` waits for 1800 seconds (see [reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Route53.html#waitFor-property))
> `Waits for the resourceRecordSetsChanged state by periodically calling the underlying [Route53.getChange()](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Route53.html#getChange-property) operation every 30 seconds (at most 60 times).` 

The new explicit configuration was set to 60 seconds which is too short and caused timeouts and failure in our cloudformation stacks

Supersedes and closes #27060 

Co-authored-by: Tom Roshko [hi@pistazie.rocks](mailto:hi@pistazie.rocks)

----


packages/aws-cdk-lib/aws-route53/lib/delete-existing-record-set-handler/index.ts
```diff
@@ -46,7 +46,7 @@ export async function handler(event: AWSLambda.CloudFormationCustomResourceEvent
     },
   });
 
-  await waitUntilResourceRecordSetsChanged({ client: route53, maxWaitTime: 60 }, { Id: changeResourceRecordSets?.ChangeInfo?.Id });
+  await waitUntilResourceRecordSetsChanged({ client: route53, maxWaitTime: 1800 }, { Id: changeResourceRecordSets?.ChangeInfo?.Id });
```
rix0rrr, line 49: Should be max 900, better yet 890 (to give Lambda 10s to send a response), and should match the defined `Timeout` of the Lambda function itself.

rix0rrr (changes requested):

MrArnoldPalmer (approved):

PR #27088, opened 2023-09-10
chore(scheduler-targets): lambda invoke target integration test

filletofish:
Continuing @Jacco work from https://github.com/aws/aws-cdk/pull/26268 

Currently we were second-guessing if the implementations of ScheduleExpression and ScheduleTargetInput generate valid and correct CloudFormation.

This integration test verifies that lambda invoke target is invoked:
- A schedule is deployed that triggers a Lambda every minute after deployment.
- The function adds a Tag to itself Key=OutputValue Value=base64 of the JSON string made of the input event
- An assertion is made that checks for 10 minutes if the expected Tag appears

----


packages/@aws-cdk/aws-scheduler-targets-alpha/test/integ.schedule.ts
```diff
@@ -0,0 +1,90 @@
...
+const functionArn = stack.formatArn({
+  service: 'lambda',
+  resource: 'function',
+  resourceName: functionName,
+  arnFormat: cdk.ArnFormat.COLON_RESOURCE_NAME,
+});
+const payload = 'test';
+
+const func = new lambda.Function(stack, 'MyLambda', {
+  code: new lambda.InlineCode(`
```
kaizencc, line 29: make this a `lambda.Asset` please. Motivation is that it is very hard to identify and make maintenance updates to this code if its inline.

packages/@aws-cdk/aws-scheduler-targets-alpha/test/integ.schedule.ts
```diff
@@ -0,0 +1,90 @@
...
+        };
+        const command = new TagResourceCommand(input);
+        const response = await client.send(command);
+    };`,
+  ),
+  handler: 'index.handler',
+  runtime: lambda.Runtime.NODEJS_18_X,
+  functionName: functionName,
+});
+func.addEnvironment('FUNC_ARN', stack.resolve(functionArn));
```
kaizencc, line 52: why do we have to add this and why can't we use `func.functionArn`

packages/@aws-cdk/aws-scheduler-targets-alpha/test/integ.schedule.ts
```diff
@@ -0,0 +1,90 @@
...
+    role,
+  }),
+});
+
+const integ = new IntegTest(app, 'integtest-endpoint', {
+  testCases: [stack],
+  stackUpdateWorkflow: false, // this would cause the schedule to trigger with the old code
+});
+
+const invoke = integ.assertions.awsApiCall('Lambda', 'listTags', {
```
kaizencc, line 79: how does this demonstrate that the schedule was successful? can you document it here as well?

packages/@aws-cdk/aws-scheduler-targets-alpha/test/integ.schedule.ts
kaizencc: ^

packages/@aws-cdk/aws-scheduler-targets-alpha/test/integ.schedule.ts
```diff
@@ -0,0 +1,90 @@
+import * as scheduler from '@aws-cdk/aws-scheduler-alpha';
+import { ExpectedResult, IntegTest } from '@aws-cdk/integ-tests-alpha';
+import * as cdk from 'aws-cdk-lib';
+import * as iam from 'aws-cdk-lib/aws-iam';
+import * as lambda from 'aws-cdk-lib/aws-lambda';
+import { LambdaInvoke } from '../lib';
+
+/*
+ * Stack verification steps:
+ * The lambda MyLambda is triggered after 5 minutes which adds a Tag:
```
kaizencc, line 10: 5 minutes or 1 minute?

kaizencc (changes requested):
@filletofish a few comments but otherwise this should be good. thanks for writing this!

packages/@aws-cdk/aws-scheduler-targets-alpha/test/integ.schedule.ts
filletofish: Otherwise - it creates a cyclic dependency

packages/@aws-cdk/aws-scheduler-targets-alpha/test/integ.lambda-invoke.js.snapshot/aws-cdk-schedule.template.json
```diff
@@ -108,7 +108,7 @@
      "Mode": "OFF"
     },
     "ScheduleExpression": "rate(1 minute)",
-    "ScheduleExpressionTimezone": "Etc/UTC",
+    "State": "ENABLED",
```
filletofish, line 111: @kaizencc  that was the change in the template after your PR (https://github.com/aws/aws-cdk/pull/27105). I am not sure that we should pass the `State` property, by default it's `ENABLED` anyway. Are there any recommendations on passing default values?

filletofish:
@kaizencc could you please restart the build? it failed with seemingly transient error when attempting to download packages from NPM

packages/@aws-cdk/aws-scheduler-targets-alpha/test/integ.lambda-invoke.js.snapshot/aws-cdk-schedule.template.json
kaizencc: It's fine to pass in the default in this case

rix0rrr (approved):

mrgrain (approved):

mrgrain (approved):

PR #27099, opened 2023-09-11
feat(lambda-nodejs): add support bun

WinterYukky:
## Summary
Add support [bun](https://bun.sh/) to NodejsFunction.

## Why need this change?
Currently, `aws-lambda-nodejs`does not support bun, so when we synth a CDK project managed by bun, we get the following error message.
```
Error: Cannot find a package lock file (`pnpm-lock.yaml`, `yarn.lock` or `package-lock.json`). Please specify it with `depsLockFilePath`.
```

Bun works very fast, so having bun support would be very helpful.

----


WinterYukky:
Exemption Request
This PR doesn't need integration test. Because this PR only add build command and it can tests by unit-test.

mrgrain (dismissed):
Conceptually this is a bit weird. We are creating a NodeJsFunction, but using Bun as a package manager only. And also not as a bundler. Need some time to think about this and discuss with the team. 🤔 

But if we do this, we will definitely need a disclaimer and warning that Bun is only used as a package manager and no Bun runtime features are supported.

indrora:
Agreeing that we should have either a better abstraction somewhere (JSFunction -> [NodeJSfunction, BunJSFunction, etc]) or a better abstraction for the bundler itself (e.g. `{"bundler":"npm|bun|deno"}`)

mrgrain (changes requested):
Still needs doc updates

WinterYukky:
Thanks your reviewing @mrgrain @indrora .
Your feedback is correct and customers will find the behaviour strange if they use bun only as a package manager. I will try to provide an option to use bun as a bundler.
However, if you try to use bun as a runtime, you will be using a custom runtime or DockerImage runtime instead of Node.js, which is outside the responsibility of the lambda-nodejs module. Therefore, I would not do this.

Regarding the design, the Runtime is still Node.js, so creating a BunjsFunction-like construct could cause confusion for customers. Also, the current bundling options are built specifically for esbuild and it would be difficult to use `{"bundler": "npm|bun|deno"}`. Therefore, we provide `{"bundler": Bundler}` which deprecates the current `bundling` option and uses the new `Bundler` type.
The code for the same configuration as now would be as follows.
```ts
// current version
new NodejsFunction(this, 'CurrentFunction', {
  bundling: {
    esbuildVersion: '0.19.3'
  }
});

// same as current version
new NodejsFunction(this, 'NewFunction', {
  bundler: Bundler.esbuild({
    esbuildVersion: '0.19.3'
  })
});

new NodejsFunction(this, 'BunFunction', {
  bundler: Bundler.bun({
    bunVersion: '1.0.3'
  })
});
```

I'm also considering switching between looking for esbuild or bun by default depending on the lock file.

I will work on this tomorrow as it is too late today. Sorry for the delay in getting started!

mrgrain:
Thanks @WinterYukky I honestly think we probably want a completely separate, new construct that goes all in with bun:
- Use bun as package manager
- Use bun as bundler
- Use the [bun lambda runtime](https://github.com/oven-sh/bun/tree/main/packages/bun-lambda)

Should we even use Docker for this?
And because this is going to be so different from everything else, we also might have to put this into a separate repo in cdklabs.

WinterYukky:
@mrgrain This may indeed be true.
It certainly could be. I initially intended to use it only as a package manager, but I wonder if many people actually want to use it at runtime as well. If a lot of customers want to use it at Node.js runtime as well, I'll revisit this PR, but I'll try to build it in cdklabs first.

PR #27052, opened 2023-09-07
fix(applicationautoscaling): add timezone property to scalable target

jrn35:
Add `timeZone` property to ScalableTarget

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-applicationautoscaling-scalabletarget-scheduledaction.html#aws-properties-applicationautoscaling-scalabletarget-scheduledaction-properties

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #19197, opened 2022-03-01
chore(aws-events): add warning when minute is not defined

TheRealAmazonKendra:
Per issue #17881, the default value for CRON schedule configuration parameters are  which runs every minute/hour/day.

This change adds a warning to be emitted upon synthesis and deployment, if minute is empty, that the CRON job will run every minute.

Closes #17881.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #27105, opened 2023-09-11
feat(core): `Schedule` class

kaizencc:
My latest attempt at schedule unification across modules. 

The following modules use a version of schedule, which this PR aims to unify:

- aws-scheduler-alpha
- aws-events
- aws-application-autoscaling
- aws-autoscaling
- aws-synthetics-alpha
- aws-backup

The idea is to have a single source of truth, `core.Schedule` that is exposed and meant to be extended by modules that use a schedule. This is to avoid breaking changes -- every module that currently exports a schedule class continues to do so. Each module can customize their schedule class to its liking, for example, whether or not to support `schedule.at` or `cronOptions.timeZone`.

This PR will fix inconsistencies like:

  - `backup.scheduleExpression` depending on `events.Schedule`, which is semi-deprecated by the Events team (they want people to use the Schedule class in `aws-scheduler-alpha`). 
  - `aws-scheduler-alpha` depending on `events.Schedule` as well.
  - `backup.scheduleExpression` allowing `schedule.rate(duration)` to be specified (synth-time error) when we know that backup schedules only can be cron expressions.
  - having to implement the new `timeZone` property in all instances of schedules
  - avoids us from having to perform maintenance in multiple places like #19197
  - `timeZone` property existing directly on a construct when it only pertains to `cron` expressions. This is an anomaly because we typically do not want construct-level properties to only be impactful depending on other properties. [See superseded PRs]

Challenges:

  - subtle differences in expressions that are accepted. This is solved by `core.Schedule` only exposing `protected` APIs, which are then picked by the consuming modules to be exposed as `public`.
  - subtle difference in `cron` expressions accepted. I do some magic in `aws-autoscaling` to get the cron expression returned there to be as expected.

Supersedes #27052 and #27012

----


packages/aws-cdk-lib/aws-backup/lib/rule.ts
```diff
@@ -41,9 +42,17 @@ export interface BackupPlanRuleProps {
...
    *
    * @default - no schedule
+   * @deprecated use schedule prop instead
    */
   readonly scheduleExpression?: events.Schedule;
 
+  /**
+   * A CRON expression specifying when AWS Backup initiates a backup job.
+   *
+   * @default - no schedule
```
rix0rrr, line 52: The word "schedule" now has lost all meaning to me and has turned into an abstract garble of letters 🫣

packages/aws-cdk-lib/aws-backup/lib/plan.ts
```diff
@@ -187,7 +187,8 @@ export class BackupPlan extends Resource implements IBackupPlan {
         moveToColdStorageAfterDays: rule.props.moveToColdStorageAfter?.toDays(),
       },
       ruleName: rule.props.ruleName ?? `${this.node.id}Rule${this.rules.length}`,
-      scheduleExpression: rule.props.scheduleExpression?.expressionString,
+      scheduleExpression: rule.props.schedule?.expressionString ?? rule.props.scheduleExpression?.expressionString,
```
rix0rrr, line 190: Needs a runtime check to make sure users are not specifying both.

rix0rrr (approved):
Nice! I'm conditionally approving this to allow you to get on with things.

A potentially even nicer solution to get rid of the `Schedule` subclass proliferation (I didn't know it had gotten that bad) could be to just have the singular `core.Schedule` class, that holds on to its input and returns it unrendered, and everyone accepts that class as a parameter.

It's then each service's responsibility to render that data to a string as the service would accept it. 

In other words: holding+rendering is now the responsibility of each service's `Schedule` class, but we could make `core.Schedule` do the holding, and `autoscaling.ScalableTarget` do the rendering.

I'll leave it to you to decide.

packages/aws-cdk-lib/core/lib/helpers-internal/index.ts
```diff
@@ -3,4 +3,4 @@ export * from './cfn-parse';
 export { md5hash } from '../private/md5';
 export * from './customize-roles';
 export * from './string-specializer';
-export { constructInfoFromConstruct, constructInfoFromStack } from '../private/runtime-info';
\ No newline at end of file
+export { constructInfoFromConstruct, constructInfoFromStack } from '../private/runtime-info';
```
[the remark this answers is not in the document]
mrgrain: He's added a newline at the end of the file.

packages/aws-cdk-lib/core/lib/schedule.ts
```diff
@@ -0,0 +1,205 @@
...
+ * @see https://docs.aws.amazon.com/eventbridge/latest/userguide/scheduled-events.html
+ */
+export abstract class Schedule {
+  /**
+   * Construct a one-time schedule from a date.
+   *
+   * @param date The date and time to use. The millisecond part will be ignored.
+   * @param timeZone The time zone to use for interpreting the date. Default: - UTC
+   */
+  protected static protectedAt(date: Date, timeZone?: TimeZone): Schedule {
```
kaizencc, line 20: did some research and it seems like there's nothing inherently wrong about having a `protected static`, just that the need for such a thing is incredibly niche. I believe this is one such scenario where this makes sense, because we want to show construct authors the available APIs that make sense, but we don't want to expose them as is to consumers

packages/@aws-cdk/aws-scheduler-alpha/test/schedule-expression.test.ts
```diff
@@ -28,20 +28,26 @@ describe('schedule expression', () => {
...
+    // GIVEN
+    const cron = ScheduleExpression.cron({
+      minute: '0',
+      hour: '10',
+      timeZone: TimeZone.EUROPE_LONDON,
+    });
+    cron._bind(new Stack());
+
+    // eslint-disable-next-line no-console
+    console.log(cron.expressionString, cron.timeZone);
```
mrgrain, line 40:
```suggestion
```

kaizencc, line 40:
```suggestion
```

saudansari:
Having timezone in CRON was useful for ECS scheduled tasks as well. Ever since this change, that feature is broken. I'd appreciate if timezone can be added to ecs.scaleOnSchedule

PR #27112, opened 2023-09-12
fix(custom-resources): does not accept strings where numbers are expected, skips recursive types

otaviomacedo:
AWS SDK v3 strictly validates numeric types, as opposed to v2, which allowed, for example, "123" to be passed where an integer was expected.

This script adds all attributes of a numeric type ("byte", "short", "integer", "long", "bigInteger", "float", "double" and "bigDecimal") to a new type coercion map. It also compacts the map by trimming each part of the path to its unique prefix among all the paths in its operation.

We also change the data structure from a list of paths (forming a tree) to a graph/state machine. This is necessary in order to support self-recursive types, which can loop arbitrarily and therefore cannot be represented in fixed length prefix lists.

See https://github.com/aws/aws-sdk-js-v3/issues/5198

----


rix0rrr (changes requested):
Good stuff, but this PR is incomplete. Doesn't gzipping have better value than unique prefixing?

The consuming code still needs to be changed to account for the new map format.

And we need to make sure that the JSON file is correctly bundled by `esbuild`.

otaviomacedo:
> Good stuff, but this PR is incomplete. Doesn't gzipping have better value than unique prefixing?
> 
> The consuming code still needs to be changed to account for the new map format.
> 
> And we need to make sure that the JSON file is correctly bundled by `esbuild`.

I was planning to do all this in subsequent PRs, to make it easier for reviewers.

PR #27116, opened 2023-09-12
chore: fix minimatch imports

rix0rrr:
For technical reasons, the way we were importing `minimatch` works in the current version of TypeScript, but is actually illegal according to the ESM spec. This poses problems when running `aws-cdk-lib` through `esbuild`.

Update to a different import style that is correct in both.

----


mrgrain (approved):
This makes me sad, but okay.

PR #27025, opened 2023-09-06
fix(appsync): Source APIs are not auto-merged by default

frixaco:
Creating Merged API does not auto-merge Source API's because it's Execution Role doesn't have enough permissions.
Gives broader (`appsync:*`) set of permissions for Merged API Execution Role to fix the issue.

Closes #26986.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #27121, opened 2023-09-13
feat(appsync): Standalone L2 construct for SourceApiAssociation

### LINKED ISSUE #26986: `aws-appsync`: "Auto-merge failed" when creating Merged AppSync API with Source API merge type set `AUTO_MERGE`
### Describe the bug

Hi and thank you for your work on the library!

Merged API Execution Role does not automatically merge the Source API, even with `merge_type=appsync.MergeType.AUTO_MERGE`.
If I understood correctly, the policy added to this role only allows [`appsync:SourceGraphQL`](https://github.com/aws/aws-cdk/blob/408f1f663ee300a783e36a35b5f103a514d8c12e/packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts#L727C7-L727C7).

### Expected Behavior

Creating Merged AppSync API should, by default and when `merge_type` is specifically set as `AUTO_MERGE`, automatically merge Source AppSync APIs.

### Current Behavior

Default behaviour for Source APIs does not work. Manually setting `merge_type` to `AUTO_MERGE` also does not work.

### Reproduction Steps

```
endpoint1_lambda = lambda_.Function(
    self,
    id="endpoint1-lambda",
    function_name="endpoint1-lambda",
    runtime=lambda_.Runtime.PYTHON_3_11,
    handler="app.lambda_handler",
    code=lambda_.Code.from_asset(
        "./lambda_functions/endpoint1",
    ),
)

source_api_1 = appsync.GraphqlApi(
    self,
    id="source-api-1",
    name="source-api-1",
    definition=appsync.Definition.from_schema(
        appsync.SchemaFile.from_asset(
            "./graphql_schemas/sourceApi1.graphql"
        )
    ),
)
endpoint1_lambda_ds = appsync.LambdaDataSource(
    self,
    id="endpoint1-lambda-ds",
    lambda_function=endpoint1_lambda,
    api=source_api_1,
)

appsync.Resolver(
    self,
    id="endpoint1-resolver",
    api=source_api_1,
    type_name="Query",
    field_name="endpoint1",
    data_source=endpoint1_lambda_ds,
    code=appsync.Code.from_asset(
        "./asd_cdk/graphql_unit_resolvers/base.js",
    ),
    runtime=appsync.FunctionRuntime.JS_1_0_0,
)

merged_api = appsync.GraphqlApi(
    self,
    id="merged-api",
    name="merged-api",
    definition=appsync.Definition.from_source_apis(
        source_apis=[
            appsync.SourceApi(
                source_api=source_api_1,
                merge_type=appsync.MergeType.AUTO_MERGE, # this doesn't work
            )
        ],
    ),
)
```

### Possible Solution

Manually creating Execution Role fixes the issue:
```
...
merged_api = appsync.GraphqlApi(
    self,
    id="merged-api",
    name="merged-api",
    definition=appsync.Definition.from_source_apis(
        source_apis=[
            appsync.SourceApi(
                source_api=source_api_1,
                merge_type=appsync.MergeType.AUTO_MERGE,
            )
        ],
        merged_api_execution_role=iam.Role( # <=====
            self,
            id="merged-api-execution-role",
            assumed_by=iam.ServicePrincipal("appsync.amazonaws.com"),
            inline_policies={
                "appsync": iam.PolicyDocument(
                    statements=[
                        iam.PolicyStatement(
                            resources=["*"],
                            actions=["appsync:*"],
                        )
                    ]
                )
            },
        ),
    ),
)
```

Extending [allowed list of actions](https://github.com/aws/aws-cdk/blob/408f1f663ee300a783e36a35b5f103a514d8c12e/packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts#L729C9-L729C44) should fix the issue. `['appsync:*]` instead of `['appsync:SourceGraphQL']`?

### CDK CLI Version

2.94.0 (build 987c329)

### Framework Version

_No response_

### Node.js Version

v18.16.0

### OS

MacOS

### Language

Python

### Language Version

Python (3.11.5)

### Other information

CleanShot 2023-09-02 at 16 40 45@2x

ndejaco2:
As part of supporting AppSync Merged APIs, this change introduces a standalone SourceApiAssociation construct for declaring a source api association between a source API and a Merged API. 

Why do we need a standalone construct?

* There are two potential deployment models when dealing with separate stacks/pipelines between the source API and Merged API: 1. Push model where the source API owners manage the association in their stack 2. Pull model where the associations are managed in the Merged API stack. 
* Having a standalone construct gives developers more flexibility while still handling all the IAM permission handling in a single place. 
* Developers can continue to use the GraphQLApi construct and declare the source api configuration all within a single construct as before. But, if they want to have the source api association as a standalone object this change gives them flexibility

I also fixed two issues related to IAM:
1. The resource for appsync:SourceGraphQL needs both the source api arn and the source api arn + "/*" to get all top level fields.
2. The merged api execution role also needs appsync:StartSchemaMerge if the association is using AUTO_MERGE. The fix here is preferred over existing PR: #27025

Closes #26986 

----


ndejaco2:
Build is failing due to rename of certain types being a breaking change. Please advice on best way to rename / deprecate.

ndejaco2:
I went ahead and undid the renaming to not introduce a breaking change. I believe it would benefit from updating the naming, but its not a high priority worth dealing with in this PR.

ndejaco2:
@rix0rrr any chance you could review? On the service team, we have been cut a few tickets regarding the IAM permissions bugs as well as the construct for Merged APIs so I would like to get this one addressed.

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
```diff
@@ -702,32 +695,44 @@ export class GraphqlApi extends GraphqlApiBase {
...
-    this.definition.sourceApiOptions?.sourceApis.forEach(sourceApiOption => {
-      new CfnSourceApiAssociation(this, `${sourceApiOption.sourceApi.node.id}Association`, {
-        sourceApiIdentifier: sourceApiOption.sourceApi.apiId,
-        mergedApiIdentifier: this.api.attrApiId,
+    this.definition.sourceApiOptions?.sourceApis.forEach(sourceApiConfig => {
+      const mergeType = sourceApiConfig.mergeType ?? MergeType.AUTO_MERGE;
+      var sourceApiIdentifier = sourceApiConfig.sourceApi.apiId;
+      var mergedApiIdentifier = this.apiId;
+
+      if (FeatureFlags.of(this).isEnabled(cxapi.APPSYNC_ENABLE_USE_ARN_IDENTIFIER_SOURCE_API_ASSOCIATION)) {
```
MrArnoldPalmer, line 703: Can we just add a comment here talking about why this is behind a feature flag, IE resource replacement even though the service supports both IDs and ARNs, only ARNs being supported cross region and changes resulting in resource replacement etc.

packages/aws-cdk-lib/aws-appsync/lib/source-api-association.ts
```diff
@@ -0,0 +1,220 @@
...
+  }
+}
+
+/**
+* Adds an IAM permission to the Merged API execution role for GraphQL access on the source AppSync api.
+*
+* @param sourceApiAssociation The CfnSourceApiAssociation resource which to add a permission to access at runtime.
+* @param mergedApiExecutionRole The merged api execution role on which to add the permission.
+*/
+export function addSourceGraphQLPermission(sourceApiAssociation: CfnSourceApiAssociation, mergedApiExecutionRole: IRole) {
```
MrArnoldPalmer, line 200: Does this need to be a free floating function? Can it be a private method on `SourceApiAssociation`? Or even on `ISourceApiAssociation` since it operates on `CfnSourceApiAssociation` under the hood.

packages/aws-cdk-lib/aws-appsync/lib/source-api-association.ts
```diff
@@ -0,0 +1,220 @@
...
+  }));
+}
+
+/**
+* Adds an IAM permission to the Merged API execution role for automatically merging the source API metadata whenever
+* the source API is updated.
+* @param sourceApiAssociation The CfnSourceApiAssociation resource which to add permission to perform merge operations on.
+* @param mergedApiExecutionRole The merged api execution role on which to add the permission.
+*/
+export function addSourceApiAutoMergePermission(sourceApiAssociation: CfnSourceApiAssociation, mergedApiExecutionRole: IRole) {
```
MrArnoldPalmer, line 214: Same as above?

MrArnoldPalmer (dismissed):
Just some small stuff for now. Gonna ask for another set of eyes since this is new api surface area that we will need to guarantee backwards compatibility on.

packages/aws-cdk-lib/aws-appsync/lib/source-api-association.ts
ndejaco2: I am using it as free floating because unfortunately this logic is needed by GraphqlApi construct as well and I want it used in one place. Should I make it a static within SourceApiAssociation?

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
ndejaco2: Sure

packages/aws-cdk-lib/aws-appsync/lib/source-api-association.ts
ndejaco2: Discussed offline, keeping it free floating as its shared with GraphqlApi construct as well

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
```diff
@@ -702,32 +695,47 @@ export class GraphqlApi extends GraphqlApiBase {
...
         sourceApiAssociationConfig: {
-          mergeType: sourceApiOption.mergeType ?? MergeType.AUTO_MERGE,
+          mergeType: mergeType,
         },
+        description: sourceApiConfig.description,
       });
+
+      // Add permissions to merged api execution role, only if it was not passed in.
+      if (!this.definition.sourceApiOptions?.mergedApiExecutionRole) {
+        const executionRole = this.mergedApiExecutionRole as IRole;
```
rix0rrr, line 722: This is not the right way to allow that use case.

If people are passing in a role and they don't want it modified, they can pass [role.withoutPolicyUpdates()](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_iam.Role.html#withoutwbrpolicywbrupdatesoptions), or import an existing one with `Role.fromRoleName({mutable: false })`.

packages/aws-cdk-lib/aws-appsync/lib/source-api-association.ts
```diff
@@ -0,0 +1,220 @@
...
+/**
+ * AppSync SourceApiAssociation which associates an AppSync source API to an AppSync Merged API.
+ * The initial creation of the SourceApiAssociation merges the source API into the Merged API schema.
+ */
+export class SourceApiAssociation extends Resource implements ISourceApiAssociation {
+
+  /**
+   * Import Appsync Source Api Association from source API, merged api, and merge type.
+   */
+  public static fromSourceApiAssociationAttributes(scope: Construct, id: string, attrs: SourceApiAssociationAttributes): ISourceApiAssociation {
```
rix0rrr, line 112: I'm not necessarily against adding this function, but its use is very unlikely, right?

There isn't a single API that accepts an `ISourceApiAssociation` ?

packages/aws-cdk-lib/aws-appsync/README.md
```diff
@@ -391,21 +391,33 @@ secondApi.addNoneDataSource('SecondSourceDS', {
...
-        sourceApi: secondApi,
-        mergeType: appsync.MergeType.AUTO_MERGE,
-      },
     ],
   }),
 });
+
+
+// An additional way of declaring a source api association using the SourceApiAssociation construct. This is useful when a source api is deployed in a different stack from the Merged API.
+// The source api owner can associate it to the Merged API with this construct.
```
rix0rrr, line 408: Rather than updating the previous example, I'd rather you make this into a new use-case driven section with a fully fresh example. Perhaps named something like "merging APIs from different stacks".

rix0rrr (dismissed):

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
ndejaco2: The existing support works like this where it is not updating the role if it was passed in. Would it be a breaking change?

MrArnoldPalmer: it might be... since this is pre-existing we can look at changing this separately.

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
```diff
@@ -702,32 +695,44 @@ export class GraphqlApi extends GraphqlApiBase {
...
 
   private setupSourceApiAssociations() {
-    this.definition.sourceApiOptions?.sourceApis.forEach(sourceApiOption => {
-      new CfnSourceApiAssociation(this, `${sourceApiOption.sourceApi.node.id}Association`, {
-        sourceApiIdentifier: sourceApiOption.sourceApi.apiId,
-        mergedApiIdentifier: this.api.attrApiId,
+    this.definition.sourceApiOptions?.sourceApis.forEach(sourceApiConfig => {
+      const mergeType = sourceApiConfig.mergeType ?? MergeType.AUTO_MERGE;
+      var sourceApiIdentifier = sourceApiConfig.sourceApi.apiId;
+      var mergedApiIdentifier = this.apiId;
```
MrArnoldPalmer, line 701:
```suggestion
      let sourceApiIdentifier = sourceApiConfig.sourceApi.apiId;
      let mergedApiIdentifier = this.apiId;
```

MrArnoldPalmer (changes requested):

MrArnoldPalmer (approved):

MrArnoldPalmer (approved):

PR #27122, opened 2023-09-13
fix(integ-tests): use `transformToString` on API call response

### LINKED ISSUE #27114: IntegTest Assertions fails while logging response
See the integ test in
`aws-s3-deployment/test/integ.bucket-deployment-substitution.ts` for an example.

The assertions handler will attempt to log the response of the made API request. In this case it is S3.getObject which includes a Blob in the `Body` field.

When logging the response, the handler uses `JSON.stringify()` which fails with to the following error:

```
Converting circular structure to JSON
    --> starting at object with constructor 'TLSSocket'
    |     property 'parser' -> object with constructor 'HTTPParser'
    --- property 'socket' closes the circle (RequestId: 50c1b6cd-47d2-494f-baf1-d22646cd4e5f)
```

This might also conceal an issue with returning blob data from the ApiCall, but not sure yet.

colifran:
This PR fixes an issue where our assertions handler will fail when logging the response of an API call if the response body is a Blob. Specifically, the following error is thrown:

```
Converting circular structure to JSON
    --> starting at object with constructor 'TLSSocket'
    |     property 'parser' -> object with constructor 'HTTPParser'
    --- property 'socket' closes the circle (RequestId: 50c1b6cd-47d2-494f-baf1-d22646cd4e5f)
```

The fix for this issue was to call the `transformToString` method on the response body. This is mentioned in aws-sdk-js-v3 [`UPGRADE.md`](https://github.com/aws/aws-sdk-js-v3/blob/main/UPGRADING.md?plain=1#L573-L576) as the solution for `Lambda::Invoke`. Its unclear why `S3::GetObject` isn't mentioned as well, but it works for that too. 

However, instead of explicitly extracting the `Body` property from a response, we now recursively traverse the response and detect any such blob types that should be converted to strings - this should protect us against any other APIs that may exhibit this behavior. 

> `transformToString` is implemented as part of the [`SdkStreamMixin` interface](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-smithy-types/Interface/SdkStreamMixin/).

Included in this PR is a fix for `aws-s3-deployment/test/integ.bucket-deployment-substitution.ts` which was previously failing with the error shown above.

Co-authored by: @iliapolo
Co-authored by: @vinayak-kukreja 
Co-authored by: @scanlonp 

Closes #27114

----


packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/lambda-handler/sdk.ts
```diff
@@ -88,12 +88,14 @@ export class AwsApiCallHandler extends CustomResourceHandler<AwsApiCallRequest,
...
     console.log(`SDK request to ${sdkPkg.service}.${request.api} with parameters ${JSON.stringify(commandInput)}`);
-    const response = await client.send(new Command(commandInput));
+    let response = await client.send(new Command(commandInput));
 
     // Lambda Invoke returns the payload as a buffer
     // we need to serialize the buffer so we can assert on it
     if (response.Payload) {
       response.Payload = parseJsonPayload(response.Payload);
+    } else if (response.Body) {
+      response = await response.Body.transformToString('utf-8');
```
iliapolo, line 98:
```suggestion
      response.Body = await response.Body.transformToString('utf-8');
```

packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/lambda-handler/sdk.ts
```diff
@@ -88,12 +88,14 @@ export class AwsApiCallHandler extends CustomResourceHandler<AwsApiCallRequest,
     const commandInput = coerceApiParametersToUint8Array(request.service, request.api, parameters);
 
     console.log(`SDK request to ${sdkPkg.service}.${request.api} with parameters ${JSON.stringify(commandInput)}`);
-    const response = await client.send(new Command(commandInput));
+    let response = await client.send(new Command(commandInput));
```
iliapolo, line 91:
```suggestion
    const response = await client.send(new Command(commandInput));
```

iliapolo:
For anyone looking at this, i'll take over the PR now and address my own comments :)

packages/@aws-cdk-testing/framework-integ/test/aws-s3-deployment/test/integ.bucket-deployment-substitution.ts
```diff
@@ -37,11 +37,11 @@ class Test extends cdk.Stack {
 }
 
 const app = new cdk.App();
-new Test(app, 'my-stack');
 
 const testCase = new Test(app, 'test-s3-deploy-substitution');
 const integ = new IntegTest(app, 'deploy-time-substitution-integ-test', {
   testCases: [testCase],
+  diffAssets: true,
```
mrgrain, line 44: Is this needed? Will hashes change with dependency updates? If so, it we need to disable it.

mrgrain (commented):
Can you add unit tests for `coerceResponse`?

mrgrain (approved):

msambol:
@colifran I'm getting the following when running integration tests in `aws-lambda-python-alpha`. Still investigating but wondering if it's related to this PR.

```
2023-09-16T13:54:21.340Z	4a446e5c-4fc1-4c97-8cb7-e8316ee45075	INFO	TypeError: Cannot read properties of undefined (reading 'req')
    at AwsApiCallHandler.processEvent (/var/task/index.js:32693:26)
    at async AwsApiCallHandler.handle (/var/task/index.js:32360:26)
    at async Runtime.handler (/var/task/index.js:32753:20)
```

This is directly after the following log line:
```
2023-09-16T13:54:19.940Z	4a446e5c-4fc1-4c97-8cb7-e8316ee45075	INFO	SDK request to Lambda.invoke with parameters {
    "FunctionName": "integ-lambda-python-poetr-myhandlerpython38withhas-YlEnyP5nCKSl"
}
```

msambol:
@rix0rrr / @colifran I went back to `main` to make sure I wasn't missing something. I changed this [line](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-lambda-python-alpha/test/integ.function.poetry.ts#L24) to force a change in the integration test (changed `my_handler_inline_with_hashes` to `my_handler_inline_foo`) and got this: 

```
The stack named poetryDefaultTestDeployAssertE9C9CB8F failed to deploy: CREATE_FAILED (The following resource(s) failed to create: [LambdaInvoke631dde0680edf7d2f0eea8d9b9c06c75, LambdaInvokefbdb1fc975f2f372fd64335aa07b46b7, LambdaInvoke26ce5a20d1ffee210d056cd2aa005172, LambdaInvokeead1d1e2cc2d2a843a514ffdc015b2c4, LambdaInvoke26497bb1c12c14ed3deada8d7d76b39e, LambdaInvoke8cbaf98b5be6e4f6f81f5f10b5b8dc89, LambdaInvoke6fd2d93b0d07a2b4220fae7c91786512, LambdaInvoke9a0beb4ea6cc38db92e9ff664c085292]. ): Response object is too long., Response object is too long., Response object is too long., Response object is too long., Response object is too long., Response object is too long.
```

msambol:
@rix0rrr / @colifran One issue at a time... In this [line](https://github.com/aws/aws-cdk/blob/af998c8581d2600fa4aa8d5fc5137da84d0f3418/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/providers/lambda-handler/base.ts#L120), I logged the size of `responseBody` and it is 13611 bytes for tests in `integ.function.poetry.ts`. Maximum size is 4096 bytes (see [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html)).

mrgrain:
@msambol FYI you can pass `--force` in to always run the integration test.

It's possible this has broken one of the existing tests, 'cause you are absolutely right: They don't get automatically re-run for changes to `integ-tests-alpha`.

If the response body is too large, we will have to set output paths to limit the response.

msambol:
Happy to submit a PR with some guidance.

PR #27124, opened 2023-09-13
feat(cloudwatch): validate Dashboards with an `end` time must also have a `start` time

go-to-k:
This PR adds a validation for a `start` and an `end` in `Dashboard`.

It throws an error if you specify an `end` without a `start`.

While it is possible to deploy a dashboard with only an end time in CloudFormation, the setting will be ignored and the dashboard displayed with the default time period.

With this validation, we are stopping the faulty deployment early.
API Reference contains the following description and the user will not be aware of this situation.

If you specify a value for end, you must also specify a value for start.

https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/CloudWatch-Dashboard-Body-Structure.html

And each Widget in graph.ts also validates the same case.

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-cloudwatch/lib/graph.ts#L245-L247

----


packages/aws-cdk-lib/aws-cloudwatch/lib/dashboard.ts
```diff
@@ -133,7 +133,11 @@ export class Dashboard extends Resource {
     }
 
     if (props.start !== undefined && props.defaultInterval !== undefined) {
-      throw ('both properties defaultInterval and start cannot be set at once');
+      throw new Error('both properties defaultInterval and start cannot be set at once');
+    }
+
+    if (props.end !== undefined && props.start === undefined) {
+      throw new Error('You must also specify a start if you specify an end');
```
lpizzinidev, line 140: I think that a better error message could be `If you specify a value for end, you must also specify a value for start.` (taken from the [docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudwatch.Dashboard.html#:~:text=If%20you%20specify%20a%20value%20for%20end%2C%20you%20must%20also%20specify%20a%20value%20for%20start.))

lpizzinidev (changes requested):
Thanks 👍
Just a minor change on the message.

packages/aws-cdk-lib/aws-cloudwatch/lib/dashboard.ts
go-to-k: @lpizzinidev 

Thanks for your review. I changed!

https://github.com/aws/aws-cdk/pull/27124/commits/d3d0b2bb6413073e76a10837eecadbbc633680cf

lpizzinidev (approved):

mrgrain:
@go-to-k Is this failing in CFN if `end` is provided without `start`?

go-to-k:
@mrgrain 

> Is this failing in CFN if end is provided without start?

No, the deployment succeeds by CFn. Also manual setting without CFn in CloudWatch Dashboard console succeeds.

But that setting is ignored and the dashboard is displayed with the default time period.

I think should be stopped because the user will not be aware of this situation.

`If you specify a value for end, you must also specify a value for start.`

https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/CloudWatch-Dashboard-Body-Structure.html

mrgrain:
Thank you, save I fully agree. We need to record these explanations on the PR for future reference.

mrgrain (approved):

PR #5773, opened 2020-01-13
fix(cli): equals sign in a context value is dropped

jzj:
Used a regular expression to capture the context value after the first equals sign.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #27130, opened 2023-09-13
fix(cli): equals sign in a tag value is dropped

### LINKED ISSUE #21003: (cli): tags with an equals sign in the value only use the value of the string prior to the equals sign
### Describe the bug

If you need a tag with a key value pair such as

Key : MyKey
Value : MyValue=Something

The stack will have a key value pair that shows

Key: MyKey
Value: MyValue

### Expected Behavior

I would expect the value of the tag to be MyValue=Something

### Current Behavior

Doesn't read the full tag value

### Reproduction Steps

Try to create a stack with a tag that has an equals sign in the value

### Possible Solution

The code which parses tags in the command line will split when it sees an equals sign, and will only use the second split value as the tag value, when it should be everything after the first split

https://github.com/aws/aws-cdk/blob/a7e3eab8f96f896f4f72bf4b14a366e9953a86e7/packages/aws-cdk/lib/settings.ts#L327-L348

_No response_

### CDK CLI Version

.

### Framework Version

_No response_

### Node.js Version

.

### OS

.

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

snaksa:
When overriding tags from the CLI the tag key-value string is split by an equal sign and the left and right parts are used for key and value respectively. The documentation for the [String.prototype.split()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) states that the second parameter sets a limit of the returned values after execution:
> If provided, splits the string at each occurrence of the specified separator, but stops when limit entries have been placed in the array. Any leftover text is not included in the array at all.

So basically if we have `foo=bar=test` and we apply a split with a limit of 2 (`"foo=bar=test".split("=", 2)`) we will get the following array `["foo", "bar"]` instead of the expected one`["foo", "bar=test"]`. 

There has been the same problem with the context values and since it is the same problem the solution is also the same: #5773 

With this fix the `context` and `tags` get the same behaviour when parsing the passed values.

Closes #21003 

----


mrgrain (approved):
Code looks good. Just running the integ tests.

PR #27134, opened 2023-09-13
chore: update cdk migrate command to allow for stack name input and additional languages

TheRealAmazonKendra:
This change updates use of noctilucent to cdk-from-cfn and installs it from npm. It also adds some additional language support and stack renaming.

Not included in this PR:
1. Automated testing - this PR has been manually tested and automated tests will be added in a subsequent PR.
2. Go support - there are still some bugs to work out with go (in cdk-from-cfn) so implementation for it will be added after that's been worked out.
3. The option to compress the app to a zip file
4. The other TODOs listed in the `migrate` file. 
----

### DIFF
--- a/CHANGELOG.v2.alpha.md
+++ b/CHANGELOG.v2.alpha.md
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.96.1-alpha.0](https://github.com/aws/aws-cdk/compare/v2.96.0-alpha.0...v2.96.1-alpha.0) (2023-09-14)
+
 ## [2.96.0-alpha.0](https://github.com/aws/aws-cdk/compare/v2.95.1-alpha.0...v2.96.0-alpha.0) (2023-09-13)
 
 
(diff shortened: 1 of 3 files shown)


TheRealAmazonKendra:
Adding the do-not-merge so that this doesn't get approved before I'm ready for it to merge. Using this PR to run the integ tests against my changes.

TheRealAmazonKendra:
This PR is now ready to go.

packages/aws-cdk/lib/init.ts
```diff
@@ -247,10 +257,17 @@ export async function printAvailableTemplates(language?: string) {
   }
 }
 
-export async function initializeProject(template: InitTemplate, language: string, canUseNetwork: boolean, generateOnly: boolean, workDir: string) {
+async function initializeProject(
```
kaizencc, line 260: non binding comment: a lot of these functions have a ton of props, and given that its experimental why not change it so that it takes in a property bag `initializeProjectOptions` so that we can easily add properties to it in the future

kaizencc (approved):

packages/aws-cdk/lib/init.ts
TheRealAmazonKendra: Good call out! I'll make this change in a future PR.

PR #27140, opened 2023-09-14
fix(s3): RETAIN removal policy does not apply to an auto created bucket policy

### LINKED ISSUE #27125: aws_s3: Bucket with removal policy = retain lose resource policy on cdk destroy
### Describe the bug

Consider the following CDK app having a bucket that has `enforce_ssl=True` and `removal_policy=Retain`. On `cdk destroy` the bucket is retained, as expected, though the bucket resource policy is deleted.

```python
from aws_cdk import Stack, RemovalPolicy

from aws_cdk import aws_s3 as s3

class CdkS3BugStack(Stack):

    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
        super().__init__(scope, construct_id, **kwargs)

        s3.Bucket(
            self,
            "RetainBucketTest",
            enforce_ssl=True,
            removal_policy=RemovalPolicy.RETAIN
        )
```

### Expected Behavior

The bucket's resource policy is retained as well

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Deny",
            "Principal": {
                "AWS": "*"
            },
            "Action": "s3:*",
            "Resource": [
                "arn:aws:s3:::cdks3bugstack-retainbuckettestef48b7bf-ms4g1j3j8dy6",
                "arn:aws:s3:::cdks3bugstack-retainbuckettestef48b7bf-ms4g1j3j8dy6/*"
            ],
            "Condition": {
                "Bool": {
                    "aws:SecureTransport": "false"
                }
            }
        }
    ]
}
```

### Current Behavior

The bucket is retained, its resource policy is not.

### Reproduction Steps

Deploy the abovementioned app `cdk deploy` 
Check the bucket resource policy, should contain the policy to disable non-SSL transport
Destroy the app `cdk destroy`
Check the bucket is still there, the policy is gone

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.93.0 (build 724bd01)

### Framework Version

_No response_

### Node.js Version

v18.14.2

### OS

Fedora 36

### Language

Python

### Language Version

_No response_

### Other information

_No response_

go-to-k:
This PR fixes the bug that a RETAIN removal policy for the bucket does not apply to an auto created bucket policy.

Closes #27125.

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/pipelines/test/integ.newpipeline-with-codebuild-logging.js.snapshot/PipelineStack.template.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/pipelines/test/integ.newpipeline-with-codebuild-logging.js.snapshot/PipelineStack.template.json
@@ -122,7 +122,9 @@
      ],
      "Version": "2012-10-17"
     }
-   }
+   },
+   "UpdateReplacePolicy": "Retain",
+   "DeletionPolicy": "Retain"
   },
   "PipelineRoleB27FAA37": {
    "Type": "AWS::IAM::Role",
--- a/packages/@aws-cdk-testing/framework-integ/test/pipelines/test/integ.newpipeline-with-cross-account-keys.js.snapshot/PipelineStack.template.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/pipelines/test/integ.newpipeline-with-cross-account-keys.js.snapshot/PipelineStack.template.json
@@ -211,7 +211,9 @@
      ],
      "Version": "2012-10-17"
     }
-   }
+   },
+   "UpdateReplacePolicy": "Retain",
+   "DeletionPolicy": "Retain"
   },
   "PipelineRoleB27FAA37": {
    "Type": "AWS::IAM::Role",
--- a/packages/@aws-cdk-testing/framework-integ/test/pipelines/test/integ.newpipeline-with-vpc.js.snapshot/PipelineStack.template.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/pipelines/test/integ.newpipeline-with-vpc.js.snapshot/PipelineStack.template.json
@@ -513,7 +513,9 @@
      ],
      "Version": "2012-10-17"
     }
-   }
+   },
+   "UpdateReplacePolicy": "Retain",
+   "DeletionPolicy": "Retain"
   },
   "PipelineRoleB27FAA37": {
    "Type": "AWS::IAM::Role",
--- a/packages/@aws-cdk-testing/framework-integ/test/pipelines/test/integ.newpipeline.js.snapshot/PipelineStack.template.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/pipelines/test/integ.newpipeline.js.snapshot/PipelineStack.template.json
@@ -122,7 +122,9 @@
      ],
      "Version": "2012-10-17"
     }
-   }
+   },
+   "UpdateReplacePolicy": "Retain",
+   "DeletionPolicy": "Retain"
   },
   "PipelineRoleB27FAA37": {
    "Type": "AWS::IAM::Role",


packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -651,6 +653,9 @@ export abstract class BucketBase extends Resource implements IBucket {
   public addToResourcePolicy(permission: iam.PolicyStatement): iam.AddToResourcePolicyResult {
     if (!this.policy && this.autoCreatePolicy) {
       this.policy = new BucketPolicy(this, 'Policy', { bucket: this });
+      if (this.removalPolicy === RemovalPolicy.RETAIN || this.removalPolicy === RemovalPolicy.RETAIN_ON_UPDATE_OR_DELETE) {
```
go-to-k, line 656: Focusing only on resolving the issue in [this issue](https://github.com/aws/aws-cdk/issues/27125), it is only for `RETAIN` or `RETAIN_ON_UPDATE_OR_DELETE`.

- Since the default for `RemovalPolicy` is `DESTROY` in the S3 Bucket and BucketPolicies, there is little benefit to bother with `applyRemovalPolicy` even when `DESTROY` is used.
- If we did that, the scope of impact on existing implementations would be very large and we would need to change many more integ tests!

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -651,6 +653,9 @@ export abstract class BucketBase extends Resource implements IBucket {
...
+        this.policy.applyRemovalPolicy(this.removalPolicy);
+      }
```
go-to-k, line 658: Focusing only on resolving the issue in [this issue](https://github.com/aws/aws-cdk/issues/27125), it is only for `RETAIN` or `RETAIN_ON_UPDATE_OR_DELETE`.

- Since the default for `RemovalPolicy` is `DESTROY` in the S3 Bucket and BucketPolicies, there is little benefit to bother with `applyRemovalPolicy` even when `DESTROY` is used.
- If we did that, the scope of impact on existing implementations would be very large and we would need to change many more integ tests!

go-to-k, line 658: Focusing only on resolving [this issue](https://github.com/aws/aws-cdk/issues/27125), it is only for `RETAIN` or `RETAIN_ON_UPDATE_OR_DELETE`.

- Since the default for `RemovalPolicy` is `DESTROY` in the S3 Bucket and BucketPolicies, there is little benefit to bother with `applyRemovalPolicy` even when `DESTROY` is used.
- If we did that, the scope of impact on existing implementations would be very large and we would need to change many more integ tests!

Also, `applyRemovalPolicy` is done only for `BucketPolicy` automatically generated here, and `BucketPolicy` that is not automatically generated is excluded. Therefore, I put it in `if (!this.policy && this.autoCreatePolicy)`.

go-to-k, line 658: Focusing only on resolving [this issue](https://github.com/aws/aws-cdk/issues/27125), it is only for `RETAIN` or `RETAIN_ON_UPDATE_OR_DELETE`.

- Since the default for `RemovalPolicy` is `DESTROY` in the S3 Bucket and BucketPolicies, there is little benefit to bother with `applyRemovalPolicy` even when `DESTROY` is used.
- If we did that, the scope of impact on existing implementations would be very large and we would need to change many more integ tests!

Also, `applyRemovalPolicy` is done only for `BucketPolicy` automatically generated here, and `BucketPolicy` that is not automatically generated is excluded. Therefore, I put it in `if (!this.policy && this.autoCreatePolicy)`.

tmokmss (commented):
Thanks for working on this! I found this change can actually create another problem in the following scenario:

1. Deploy a bucket with a bucket policy with RETAIN removal policy

```ts
new Bucket(stack, 'Bucket', {
  enforceSSL: true,
  removalPolicy: cdk.RemovalPolicy.RETAIN,
});
```

2. Then remove a bucket policy and deploy

```diff
new Bucket(stack, 'Bucket', {
-  enforceSSL: true,
  removalPolicy: cdk.RemovalPolicy.RETAIN,
});
```


3. Finally re-create a bucket policy and deploy

```diff
new Bucket(stack, 'Bucket', {
+  enforceSSL: true,
  removalPolicy: cdk.RemovalPolicy.RETAIN,
});
```

We will get a CFn error on the final deployment : `The bucket policy already exists on bucket xxx`

Considering this, I think setting a removal policy to a bucket policy is not necessarily a good idea. What do you think?

edit) `enforceSSL` is just for example here. Any configuration that uses bucket policy, such as VPC flow log, ALB access logging, CloudTrail, etc will be affected.

go-to-k:
> But isn't that not specific to this case, but also to other cases of RETAIN of other resources?

I tried that other cases (set RETAIN, delete, and re-create) such as a `Bucket`, but it succeeded, so no duplicate errors occurred....

However, I still think we don't want the policy to disappear when the bucket is deleted. Is there a good way to do this?

go-to-k:
I had a misconception. In this case, the error is caused by the fact that Bucket can only have one bucket policy.

If this is the case, we may not be able to handle it in CDK layer because of the difficulty of updatable import of the resource that is the old policy, unless we remove the old policy with custom resource Lambda, etc.

> I think setting a removal policy to a bucket policy is not necessarily a good idea.

So I agree with you.

go-to-k:
Or I thought about creating a BucketPolicy from the beginning without enforceSSL, but I don't know if it is a good idea.

tmokmss:
I'm still not sure how common this issue is. Also, can there be any reverse use case such that a user wants to delete a bucket policy while retaining data in a bucket (e.g. to remove unnecessary permissions)? Then it would break those use cases.

If you want to proceed further I guess you can summarize available options and ask a core team member which way to go. IMHO an escape hatch is enough to workaround this issue though.`(bucket.findChild('Policy') as CfnResource).applyRemovalPolicy(RemovalPolicy.RETAIN)`

go-to-k:
> Also, can there be any reverse use case such that a user wants to delete a bucket policy while retaining data in a bucket (e.g. to remove unnecessary permissions)?

I don't think that case is very likely. Because if we wanted to leave the bucket alone and rewrite the policy to something else later, we could just manually overwrite the settings.

However, I can't see where this PR solution would probably go, so I'm going to close it once and for all. If driven by the need, it will be reopened then.

PR #27166, opened 2023-09-15
chore(init-templates): allow renaming in templates for cdk migrate apps

HBobertz:
New integ test for the new CDK migrate command. Template.txt is a sample cloudformation template with enough complexity to be a good enough sample.

Tests for Java and Csharp aren't there because they are broken right now. Add them back after we fix

----


TheRealAmazonKendra (changes requested):
I'm thinking this should actually be placed within the `cli-integ-tests` folder. It could even go in the `cli.integtest.ts` file because it is a CLI test.

HBobertz:
> I'm thinking this should actually be placed within the `cli-integ-tests` folder. It could even go in the `cli.integtest.ts` file because it is a CLI test.

I had planned to separate them because it seems any test that uses cdk init is separated from the other commands. I didn't know if this was due to a compatibility issue, runtime optimization, or if it was just arbitrary. But I had planned to keep that structure just in case. If that's not right, then I'm happy to put it in the cli-integ-tests suite because it would definitely simplify the process.

TheRealAmazonKendra:
Looking a bit more closely, I do think that we should put this in the `cli.integtest.ts` file and then reuse the setup already being used in there so that we can actually do a deployment of the synthesized app. I know we had discussed only testing as far as synthesis goes before, but since, in this case we know we're starting with good templates, I think we should actually deploy it and verify that deployment.

HBobertz:
> Looking a bit more closely, I do think that we should put this in the `cli.integtest.ts` file and then reuse the setup already being used in there so that we can actually do a deployment of the synthesized app. I know we had discussed only testing as far as synthesis goes before, but since, in this case we know we're starting with good templates, I think we should actually deploy it and verify that deployment.

will do

HBobertz:
Closing

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/template.txt
rix0rrr, line 1: `template.txt` 🫣

HBobertz: lol want it to be .json? or just a more descriptive name than "tempalte"

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -50,6 +50,31 @@ integTest('VPC Lookup', withDefaultFixture(async (fixture) => {
...
+  withTemporaryDirectory(
+    withPackages(async (context) => {
+
+      const tempShell = ShellHelper.fromContext(context);
+      await context.packages.makeCliAvailable();
+
+      const tempPath = path.resolve(context.integTestDir);
+      const inputFile = path.join(__dirname, 'template.txt');
+
+      await tempShell.shell([
```
TheRealAmazonKendra, line 64: This can't possibly pass. It's missing `--stack-name`.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -50,6 +50,31 @@ integTest('VPC Lookup', withDefaultFixture(async (fixture) => {
   await fixture.cdkDeploy('import-vpc', { modEnv: { ENABLE_VPC_TESTING: 'IMPORT' } });
 }));
 
+integTest(
```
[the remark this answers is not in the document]
TheRealAmazonKendra: Also, I'd move it below the tests that actually test synth and deploy.

TheRealAmazonKendra (changes requested):

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
HBobertz: Hm good point. It's actually kind of alarming because this did pass, and that likely means our other init tests aren't being tested currently.

TheRealAmazonKendra: Most of those tests have an `expect` statement of some sort. I see that not all do but I'm not totally familiar with their inner workings.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/template.txt
mrgrain: Both!

HBobertz: changed them

packages/@aws-cdk-testing/cli-integ/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@aws-cdk-testing/cli-integ",
   "description": "Integration tests for the AWS CDK CLI",
-  "version": "0.0.0",
+  "version": "2.96.2",
```
TheRealAmazonKendra, line 4: Anything we have listed as version `0.0.0` should not be changed (also in the instances below).

TheRealAmazonKendra (changes requested):

packages/aws-cdk/lib/commands/migrate.ts
```diff
@@ -79,7 +79,7 @@ export async function generateCdkApp(stackName: string, stack: string, language:
  */
 export function generateStack(template: string, stackName: string, language: string) {
   try {
-    const formattedStackName = `${camelCase(decamelize(stackName), { pascalCase: true })}Stack`;
+    const formattedStackName = `${camelCase(decamelize(stackName), { pascalCase: true })}` + language == 'csharp' ? '' : 'Stack';
```
TheRealAmazonKendra, line 82: This breaks all the csharp test. Can you elaborate on what you were trying to fix here?

TheRealAmazonKendra (changes requested):

packages/aws-cdk/lib/commands/migrate.ts
HBobertz: Typescript fun fact of the day. "+" has a higher priority than "?" so this evaluates in the incorrect order. Added some () to fix the PEMDAS here. This is only appending stack to the end if it is a language other than CSharp. This is because the CSharp standard is to name the namespace "appName" and then the stack "appNameStack". So since we want to maintain this then we need to make sure we are not appending Stack all the time to CSharp. If we didn't do this then we'd be naming things "myAppStack" and "myAppStackStack" which is a bit silly. Or in this case "migrateStackStack" and "migrateStackStackStack" which is even sillier.

All good now. Should be 5/5 once the cdk-from-cfn change is merged

```
Test Suites: 1 failed, 1 skipped, 1 of 2 total
Tests:       1 failed, 73 skipped, 4 passed, 78 total
Snapshots:   0 total
Time:        468.022 s
```

HBobertz:
I don't think our code build job is actually running the tests so I'll just paste this here (I checked the logs and couldn't find any logs for the tests)

Test Suites: 1 skipped, 1 passed, 1 of 2 total
Tests:       73 skipped, 5 passed, 78 total
Snapshots:   0 total
Time:        568.341 s

packages/aws-cdk/lib/init-templates/app/csharp/src/%name.PascalCased%/Program.template.cs
```diff
@@ -10,7 +10,7 @@ sealed class Program
         public static void Main(string[] args)
         {
             var app = new App();
-            new %name.PascalCased%Stack(app, "%stackname%", new StackProps
+            new %PascalStackName%(app, "%stackname%", new %PascalStackProps%
```
TheRealAmazonKendra, line 13: This shouldn't change `StackProps`

packages/aws-cdk/test/commands/migrate.test.ts
```diff
@@ -177,7 +177,7 @@ describe('Migrate Function Tests', () => {
     // Replaced stack file is referenced correctly in app file
     const app = fs.readFileSync(path.join(workDir, 'GoodCSharp', 'src', 'GoodCSharp', 'Program.cs'), 'utf8').split('\n');
     expect(app.map(line => line.match('namespace GoodCSharp')).filter(line => line).length).toEqual(1);
-    expect(app.map(line => line.match(/        new GoodCSharpStack\(app, "GoodCSharp", new StackProps/)).filter(line => line).length).toEqual(1);
+    expect(app.map(line => line.match(/        new GoodCSharp\(app, "GoodCSharp", new GoodCSharpProps/)).filter(line => line).length).toEqual(1);
```
TheRealAmazonKendra, line 180: I don't think this line should be changing here.

packages/aws-cdk/lib/init.ts
```diff
@@ -173,6 +173,8 @@ export class InitTemplate {
     }
     return template.replace(/%name%/g, project.name)
       .replace(/%stackname%/, project.stackName ?? '%name.PascalCased%Stack')
+      .replace(/%PascalStackName%/, project.stackName ? camelCase(project.stackName, { pascalCase: true }) : '%name.PascalCased%Stack')
+      .replace(/%PascalStackProps%/, project.stackName ? (camelCase(project.stackName, { pascalCase: true }) + 'Props') : 'StackProps')
```
TheRealAmazonKendra, line 177: Pretty sure we should only be changing the first line, not this second one.

packages/aws-cdk/test/commands/test-resources/stacks/S3Stack.cs
```diff
@@ -3,8 +3,12 @@
 using Constructs;
 using System.Collections.Generic;
 
-namespace Com.Acme.Test.Simple
+namespace GoodCSharpStack
```
TheRealAmazonKendra, line 6: I thought we talked about removing `Stack` here and just making this `GoodCSharp`

packages/@aws-cdk-testing/cli-integ/resources/templates/simpleDDB.yaml
```diff
@@ -0,0 +1,79 @@
+AWSTemplateFormatVersion: "2010-09-09"
```
TheRealAmazonKendra, line 1: Where is this being used?

packages/@aws-cdk-testing/cli-integ/resources/templates/autoscaling-template.yml
```diff
@@ -0,0 +1,677 @@
+AWSTemplateFormatVersion: 2010-09-09
```
TheRealAmazonKendra, line 1: Where is this being used?

packages/aws-cdk/lib/init-templates/app/csharp/src/%name.PascalCased%/Program.template.cs
TheRealAmazonKendra: Line 6 of this template is what we discussed changing, not line 13.

TheRealAmazonKendra (changes requested):

packages/aws-cdk/lib/init-templates/app/csharp/src/%name.PascalCased%/Program.template.cs
HBobertz: if the stack is expecting our own stack props then we can't use StackProps or it would give us

```
error CS1503: Argument 3: cannot convert from 'Amazon.CDK.StackProps' to 'GoodCSharp.GoodCSharpStackProps' [/Users/bobertzh/work/deleteme/temp/GoodCSharp/src/GoodCSharp/GoodCSharp.csproj]

```

HBobertz: I did realize though that we actually need more changes on 13 and on 6 since we aren't going to differentiate between stack name and app name. I'll push a change

packages/@aws-cdk-testing/cli-integ/resources/templates/simpleDDB.yaml
HBobertz: It's not as it doesn't pass for csharp. Had planned to have multiple integ test files for the various languages but these aren't being used until csharp passes generally

packages/@aws-cdk-testing/cli-integ/resources/templates/autoscaling-template.yml
HBobertz: It's not as it doesn't pass for csharp. Had planned to have multiple integ test files for the various languages but these aren't being used until csharp passes generally

packages/aws-cdk/test/commands/test-resources/stacks/S3Stack.cs
HBobertz: This was my suggestion. You suggested we don't do that and instead have all names GoodCSharpStack. Forgot to change this in the init template on line 6 as well

packages/aws-cdk/test/commands/migrate.test.ts
HBobertz: Yup I'll fix the init template and change this

HBobertz: There seems to have been a misunderstanding between us on how you wanted this to look. Lets sync offline on how we want the csharp applciation to look

TheRealAmazonKendra (changes requested):
Looks like a few comments haven't been addressed yet.

packages/aws-cdk/lib/init-templates/app/csharp/src/%name.PascalCased%/Program.template.cs
```diff
@@ -3,14 +3,14 @@
...
-namespace %name.PascalCased%
+namespace %PascalNameSpace%
 {
     sealed class Program
     {
         public static void Main(string[] args)
         {
             var app = new App();
-            new %name.PascalCased%Stack(app, "%stackname%", new StackProps
+            new %name.PascalCased%Stack(app, "%stackname%", new %PascalStackProps%
```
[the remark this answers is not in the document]
HBobertz: They should as it is our documented best practice for csharp: https://docs.aws.amazon.com/cdk/v2/guide/stack_how_to_create_multiple_stacks.html#cdk-how-to-create-multiple-stacks-extend-stackprops

packages/aws-cdk/lib/init.ts
```diff
@@ -173,6 +173,8 @@ export class InitTemplate {
     }
     return template.replace(/%name%/g, project.name)
       .replace(/%stackname%/, project.stackName ?? '%name.PascalCased%Stack')
+      .replace(/%PascalNameSpace%/, project.stackName ? camelCase(project.stackName + 'Stack', { pascalCase: true }) : '%name.PascalCased%')
+      .replace(/%PascalStackProps%/, project.stackName ? (camelCase(project.stackName, { pascalCase: true }) + 'StackProps') : 'StackProps')
```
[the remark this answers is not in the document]
HBobertz: Same comment as last review on this line

packages/@aws-cdk-testing/cli-integ/resources/templates/autoscaling-template.yml
HBobertz: I wanted to use them as tests for our languages so we had a more robust set of tests as the current one does not test much. But given not all languages can handle all templates just yet. I will remove for now and add again later

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -571,6 +571,52 @@ integTest('deploy with role', withDefaultFixture(async (fixture) => {
   }
 }));
 
+['typescript', 'python', 'csharp', 'java', 'go'].forEach(langChoice => {
```
TheRealAmazonKendra, line 574:
```suggestion
// TODO: Add Go Tests
['typescript', 'python', 'csharp', 'java'].forEach(langChoice => {
```

TheRealAmazonKendra:
@Mergifyio update

TheRealAmazonKendra (approved):

PR #27184, opened 2023-09-18
chore: add --compress flag to cdk migrate

TheRealAmazonKendra:
This also updates some of the aliases used because they clashed with other flags. We'll just remove these.

The archive file is shamelessly ripped off from the cdk-assets package.

----


TheRealAmazonKendra:
Note: using chore for all of the incremental changes. Once we're ready to have this in the changelog, I'll make a feat PR and references all the smaller changes I'm making ahead of that.

PR #27195, opened 2023-09-19
fix(cdk): Resolve cross stack and default parameters for hotswaps

Amplifiyer:
This PR solves two problems when doing hotswap deployments with nested stacks.

1. Adding capabilities to evaluate CFN parameters of `AWS::CloudFormation::Stack` type (i.e. a nested stack). In this PR, we are only resolving `Outputs` section of `AWS::CloudFormation::Stack` and it can be expanded to other fields in the future. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/quickref-cloudformation.html#w2ab1c17c23c19b5 for CFN documentation around using Outputs ref parameters
2. If a template has parameters with default values and they are not provided (a value) by the parent stack when invoking, then resolve these parameters using the default values in the template.

Partially helps https://github.com/aws/aws-cdk-cli/issues/874 and https://github.com/aws/aws-cdk-cli/issues/858

----


packages/aws-cdk/lib/api/evaluate-cloudformation-template.ts
```diff
@@ -74,22 +81,34 @@ export class EvaluateCloudFormationTemplate {
...
+      stackName,
       template: nestedTemplate,
-      parameters: nestedStackParameters,
+      parameters: evaluatedParams,
       account: this.account,
       region: this.region,
       partition: this.partition,
       urlSuffix: this.urlSuffix,
-      listStackResources: listNestedStackResources,
+      sdk: this.sdk,
```
comcalvi, line 110: why do we need this here?

packages/aws-cdk/lib/api/hotswap-deployments.ts
```diff
@@ -54,21 +55,21 @@ export async function tryHotswapDeployment(
   // create a new SDK using the CLI credentials, because the default one will not work for new-style synthesis -
   // it assumes the bootstrap deploy Role, which doesn't have permissions to update Lambda functions
   const sdk = (await sdkProvider.forEnvironment(resolvedEnv, Mode.ForWriting)).sdk;
-  // The current resources of the Stack.
-  // We need them to figure out the physical name of a resource in case it wasn't specified by the user.
-  // We fetch it lazily, to save a service call, in case all hotswapped resources have their physical names set.
-  const listStackResources = new LazyListStackResources(sdk, stackArtifact.stackName);
```
comcalvi, line 60: this has been moved elsewhere, could you explain why? I don't see why this is necessary.

comcalvi (commented):
I need to understand the addition of the `sdk` to an `EvaluateCloudFormationTemplate` before I can give this a better review.

comcalvi (approved):
Great work! This will merge soon, just need the integ tests to pass

TheRealAmazonKendra:
Apologies, I stopped the integ test progress on this because I thought it was a change I pushed. I've restarted it now.

PR #27197, opened 2023-09-19
feat(cdk): add AppSync GraphQLSchema and pipeline resolvers as hot swappable

Amplifiyer:
1. Add GraphQLSchema as another AppSync resource that can be hotswapped
2. For all AppSync resources, accept the change in S3 assets/files instead of just inline code as a candidate for hotswap
3. Make pipeline resolvers hotswappable by resolving the functionId of AppSync functions.

Closes aws/aws-cdk#2659, aws/aws-cdk-cli#872, aws/aws-cdk-cli#871.

----


packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -49,25 +50,76 @@ export async function isHotswappableAppSyncChange(
...
-            functionId: functionId!,
-          }).promise();
+          await simpleRetry(
+            () => sdk.appsync().updateFunction({ ...sdkRequestObject, functionId: functionId! }).promise(),
+            3,
+            'ConcurrentModificationException');
+        } else {
+          let schemaCreationResponse: GetSchemaCreationStatusResponse = await sdk.appsync().startSchemaCreation(sdkRequestObject).promise();
+          while (schemaCreationResponse.status && ['PROCESSING', 'DELETING'].some(status => status === schemaCreationResponse.status)) {
+            await new Promise(resolve => setTimeout(resolve, 1000)); // poll every second
```
comcalvi, line 90: can you make a little helper for this, similar to: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/util/archive.ts#L84

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -1,29 +1,30 @@
...
   const ret: ChangeHotswapResult = [];
-  if (isResolver && change.newValue.Properties?.Kind === 'PIPELINE') {
-    reportNonHotswappableChange(
-      ret,
-      change,
-      undefined,
-      'Pipeline resolvers cannot be hotswapped since they reference the FunctionId of the underlying functions, which cannot be resolved',
-    );
-    return ret;
-  }
```
comcalvi, line 24: is this no longer true?

comcalvi (changes requested):
looks really good, overall!

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
Amplifiyer: Yes, now that we can resolve the functionIds https://github.com/aws/aws-cdk/pull/27197/files#diff-ff9724092d7f1e6bf3141e51ddb00ecbc2148e16f141b0bd2074d6edc60c16b6R394-R398

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
Amplifiyer: Good idea, updated.

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -49,25 +50,76 @@ export async function isHotswappableAppSyncChange(
...
+          requestMappingTemplateS3Location: change.newValue.Properties?.RequestMappingTemplateS3Location,
           responseMappingTemplate: change.newValue.Properties?.ResponseMappingTemplate,
+          responseMappingTemplateS3Location: change.newValue.Properties?.ResponseMappingTemplateS3Location,
         };
         const evaluatedResourceProperties = await evaluateCfnTemplate.evaluateCfnExpression(sdkProperties);
         const sdkRequestObject = transformObjectKeys(evaluatedResourceProperties, lowerCaseFirstCharacter);
 
+        // resolve s3 location files as SDK doesn't take in s3 location but inline code
+        if (sdkRequestObject.requestMappingTemplateS3Location) {
+          sdkRequestObject.requestMappingTemplate = (await fetchFileFromS3(sdkRequestObject.requestMappingTemplateS3Location, sdk))?.toString('utf8');
```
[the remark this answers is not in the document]
alharris-at: Because Synth produces assets with a hash in the filename, I think the S3Location *will* change if a resolver or function's code is updated, and yeah we use L1 constructs here.

@comcalvi not urgent, but we might actually want to open the discussion to using S3Location instead of the inline approach for the L2 (when an Asset is provided), since that means customers will far more quickly hit Cfn template size limits (we see this happen for Amplify customers w/o inlining much of these assets, since with AppSync it's easy to produce a large number of resources quickly, and if we're inlining all code then it's going to expand very quickly).

alharris-at:
@Amplifiyer - looks like this doesn't support `code` props yet. Is that tracked in a separate PR, or is there a good reason to separate out? Seems like a the change will have significant overlap w/ the updates to S3Locatio, etc.

Amplifiyer:
> @Amplifiyer - looks like this doesn't support `code` props yet. Is that tracked in a separate PR, or is there a good reason to separate out? Seems like a the change will have significant overlap w/ the updates to S3Locatio, etc.

@alharris-at, Yes I intend to add the `Code` in a follow up PR. Can you expand on what the overlap you mentioned here?

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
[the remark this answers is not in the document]
comcalvi: ahh okay, so the s3 locations are managed through the CDK here. That makes sense, thanks for clarifying.

comcalvi (approved):

PR #27217, opened 2023-09-20
feat(aws-cdk-lib): reduce load time of JavaScript library

rix0rrr:
This change halves the load time of `require('aws-cdk-lib')`, from ~600ms to ~300ms on my machine. It also shaves ~8min off running all integ tests (again, on my machine).

It does this by making all loading of JavaScript modules lazy, so they are only loaded just before any symbols from them are actually accessed (instead of loading all modules at startup). This new automatic module replaces the previous manual effort to do something similar, in `lazy-index.ts`.

Note that this transformation (unfortunately) does not affect TypeScript compilation times: the TypeScript compiler is still going to read all `.d.ts` files. To improve interactive TypeScript performance, use `swc` or `ts-node` with type checking disabled. Fully type checking your application using `tsc` will take the amount of time it takes.

## What's the big idea?

Turning this:

```js
const x = require("x");
module.exports.banana = function() {
  return x.hello();
}
```

Into this:

```js
var x = () => { var tmp = require("x"); x = () => tmp; return tmp; };
module.exports.banana = function () {
    return x().hello();
};
```

----


tools/@aws-cdk/lazify/lib/index.ts
```diff
@@ -0,0 +1,220 @@
...
+            || ts.isMethodDeclaration(node.parent) // public ident() { ... }
+            || ts.isMethodSignature(node.parent) // interface X { ident(); }
+            || ts.isPropertyDeclaration(node.parent) // class X { ident: string }
+            || ts.isPropertySignature(node.parent) // interface X { ident: string }
+            || ts.isGetAccessor(node.parent) // class X { get ident() { ... } }
+            || ts.isGetAccessorDeclaration(node.parent) // interface X { get ident: string }
+            || ts.isSetAccessor(node.parent) // class X { set ident() { ... } }
+            || ts.isSetAccessorDeclaration(node.parent) // interface X { set ident: string }
+          );
+          // We should also ignore this identifier if it is shadowed
```
otaviomacedo, line 104: So why are we not doing it?

tools/@aws-cdk/lazify/lib/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      const visit: ts.Visitor = node => {
+        // If this is the statement, replace it with a function definition
+        if (node === stmt) {
+          return createVariable(factory, binding,
+            factory.createArrowFunction(undefined, undefined, [], undefined, undefined,
+              factory.createBlock([
+                // tmp = require(...)
+                createVariable(factory, 'tmp', factory.createCallExpression(factory.createIdentifier('require'), [], [factory.createStringLiteral(moduleName)])),
+
+                // <this_fn> = () => tmp
```
otaviomacedo, line 72: Consider commenting on why we need this. My naive expectation would be to have something like:

```ts
var x = () => { return require("x"); };
```

without any need for a temporary variable.

otaviomacedo (approved):

PR #27223, opened 2023-09-20
fix(batch): make `instanceType` optional on multinode jobs

### LINKED ISSUE #27185: aws-batch: MultiNodeJobDefinition Incorrectly Requires InstanceType
### Describe the bug

It seems that the Batch [MultiNodeJobDefinition CDK Construct](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_batch.MultiNodeJobDefinition.html) requires instance type to be defined. However, instanceType is not required by the Batch AWS SDK.

This seems to be more than a documentation issue, as I'm unable to create a MultiNodeJobDefinition in my CDK package without specifying instance type.

This constraint takes away from one of the primary advantages to having Batch manage our EC2 resources in that they optimally assign batch jobs to instance types based on availability/resource requirements.

### Expected Behavior

We shouldn't need to specify instance type. Batch should determine it based on job resource requirements.

### Current Behavior

CDK requires instance type to be required at job definition level.

### Reproduction Steps

Create a MultiNodeJobDefinition.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.96.2

### Framework Version

_No response_

### Node.js Version

18.x

### OS

linux

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

comcalvi:
Make `instanceType` optional. It used to be required. It will default to a batch optimal instance type.

The breaking change is for `MultiNodeJobDefinition.instanceType`, which is now optional. This is not truly breaking because everyone has this property explicitly set now, and will continue to have it set after this change. 

Closes #27185.

----


TheRealAmazonKendra:
Weird that setting something to optional is considered a breaking change.

allowed-breaking-changes.txt
```diff
@@ -188,3 +188,6 @@ removed:aws-cdk-lib.aws_cloudwatch.CfnAlarm.attrId
 removed:aws-cdk-lib.aws_ec2.CfnNetworkInterface.enablePrimaryIpv6
 removed:aws-cdk-lib.aws_ec2.CfnNetworkInterfaceAttachment.attrId
 removed:aws-cdk-lib.aws_ec2.CfnNetworkInterfaceProps.enablePrimaryIpv6
+
+# Make this optional so Batch can select the optimal instance type by default
+changed-type:aws-cdk-lib.aws_batch.MultiNodeJobDefinition.instanceType
```
rix0rrr, line 193: Why is it okay to make breaking changes to a stable API here? Especially since this  can be avoided?

rix0rrr:
On the question of making properties optional, this comes up way too often, so I made an explainer for it:

https://github.com/aws/jsii/pull/4274/files?short_path=f5d91d1#diff-f5d91d104bff5f6eba3495d9aa4ef6d0e2854a010d6c2754acdd98536afc3c05

comcalvi:
thanks for writeup @rix0rrr, that's helpful!

TheRealAmazonKendra (approved):

PR #27229, opened 2023-09-20
feat(appsync): Custom Resource to Perform Manual Merge of a Source API And Signal Cloudformation Based On Merge Operation Result

### LINKED ISSUE #27170: (aws-appsync): Custom Resource to Perform Manual Merge of a Source API And Signal Cloudformation Based On Merge Operation Result
### Describe the feature

AppSync recently launched Merged APIs: https://aws.amazon.com/blogs/mobile/introducing-merged-apis-on-aws-appsync/. 
CDK support has been added to support Merged APIs with the following PRs:

1. https://github.com/aws/aws-cdk/pull/26895
2. https://github.com/aws/aws-cdk/pull/27121

The main resource for Merged APIs is the CfnSourceApiAssociation which links a source API with a merged API. When creating this resource, either through the GraphqlApi construct or the standalone SourceApiAssociation construct, you need to specify a merge type. The default type in CDK is AUTO_MERGE. AUTO_MERGE ensures that any change that is made on a source API is propagated to the Merged API endpoint in the background by the AppSync service itself. However, since this happens outside of Cloudformation, if there is a failure due to a merge conflict, the Cloudformation stack update is not aware of it and will not rollback automatically. Developers may want to ensure that a merge succeeds prior to the Cloudformation stack update succeeding. 

### Use Case

This feature allows the source API owner to ensure that when merge conflicts are introduced, the Cloudformation stack update would fail and cause a rollback. Otherwise, if the merge is successful, the team knows that the updates were successfully applied to the merged API. 

### Proposed Solution

I believe we can support this use case with a custom resource in the appsync CDK library. The custom resource would have a lambda function that submits the merge of the source API manually and a Lambda function that handles polling to check when the merge operation has finished. When the operation has finished, it will signal Cloudformation to succeed or fail based on the status of the merge operation. 

Some key considerations:

- The custom resource needs to dependOn the source API and all children of the source API such as datasources, resolvers and functions. Otherwise, it might merge the API metadata prior to it being updated again. 
- Ideally, we would only invoke the custom resource when children of the source API have been updated / added / deleted. Otherwise, if there is no good way to determine whether something has changed and trigger merge of the source api, we might have to settle for allowing no-ops. If the community has some recommendations on the best way to implement this that would be great! 
- The Provider framework is ideal for this use case because it allows us to submit the merge and poll separately in a different lambda function.
- Ideally, we would want to have a single Lambda function per stack rather than creating a separate instance of the Lambda per source api association. While there would be a single Lambda function per stack, there would be a CustomResource per source api association. 
- I believe this could be functionality that is added within the ISourceApiAssociation interface and available through a function like addCustomSourceApiMergeOperation() so that it is an opt in feature for the user.

Below is a snippet of code for the POC I have built for this feature:

```
/**
   * Creates a custom resource which will submit a schema merge of the source api each time the stack is deployed.
   * The custom resource will wait for the merge to finish and fail the update if the merge operation has any failure.
   */
  public addCustomSourceApiMergeOperation() {

    // Use NODEJS_LATEST so that it bundles the most recent aws sdk which has the Merged APIs operations.
    const schemaMergeLambda = new NodejsFunction(this, 'MergeSourceApiSchemaLambda', {
      runtime: Runtime.NODEJS_LATEST,
      entry: path.join(__dirname, 'mergeSourceApiSchemaHandler', 'index.ts'),
      handler: 'onEvent',
      timeout: Duration.minutes(2),
    });

    schemaMergeLambda.addToRolePolicy(new PolicyStatement({
      effect: Effect.ALLOW,
      resources: [this.associationArn],
      actions: ['appsync:StartSchemaMerge'],
    }));

    const sourceApiStabilizationLambda = new NodejsFunction(this, 'PollSourceApiMergeLambda', {
      runtime: Runtime.NODEJS_LATEST,
      entry: path.join(__dirname, 'mergeSourceApiSchemaHandler', 'index.ts'),
      handler: 'isComplete',
      timeout: Duration.minutes(2),
    });

    sourceApiStabilizationLambda.addToRolePolicy(new PolicyStatement({
      effect: Effect.ALLOW,
      resources: [this.associationArn],
      actions: ['appsync:GetSourceApiAssociation'],
    }));

    const provider = new Provider(this, 'SourceApiSchemaMergeOperationProvider', {
      onEventHandler: schemaMergeLambda,
      isCompleteHandler: sourceApiStabilizationLambda,
    });

    // The goal is for this custom resource to run once per Cloudformation stack update.
    // We force this by generating a property that is a random uuid each time cdk is synthesized.
    // This will force creation of a new CustomResource rather than recognizing it as existing, which will trigger the merge.
    // Ideally, we could figure out a better way to avoid the potential no-ops. 
    const customResource = new CustomResource(this, 'SourceApiSchemaMergeOperationCustomResource', {
      serviceToken: provider.serviceToken,
      resourceType: 'Custom::AppSyncSourceApiMergeOperation',
      properties: {
        associationId: this.associationId,
        mergedApiIdentifier: this.mergedApi.arn,
        sourceApiIdentifier: this.sourceApi.arn,
        alwaysUpdate: randomUUID(),
      },
    });

    // If a reference to the source API exists,
    // add a dependency on all children of the source api in order to ensure that this resource is created at the end.
    if (this.sourceApi) {
      this.sourceApi.node.children.forEach((child) => {
        if (CfnResource.isCfnResource(child)) {
          customResource.node.addDependency(child);
        }

        if (Construct.isConstruct(child) && child.node.defaultChild && CfnResource.isCfnResource(child.node.defaultChild)) {
          customResource.node.addDependency(child.node.defaultChild);
        }
      });
    }
}
```

Implementation for the Lambda handlers:

```
// eslint-disable-next-line import/no-extraneous-dependencies
import {
  StartSchemaMergeCommand,
  AppSyncClient,
  SourceApiAssociationStatus,
  GetSourceApiAssociationCommand,
  GetSourceApiAssociationCommandInput,
} from '@aws-sdk/client-appsync';

const appSyncClient = new AppSyncClient();

type SchemaMergeResult = {
  associationId?: string,
  mergedApiIdentifier?: string,
  PhysicalResourceId: string,
  sourceApiAssociationStatus?: SourceApiAssociationStatus,
};

type IsCompleteResult = {
  IsComplete: boolean,
  Data?: Object
}

export async function onEvent(event: AWSLambda.CloudFormationCustomResourceEvent): Promise<SchemaMergeResult> {
  const params = {
    associationId: event.ResourceProperties.associationId,
    mergedApiIdentifier: event.ResourceProperties.mergedApiIdentifier,
  };

  switch (event.RequestType) {
    case 'Update':
    case 'Create':
      return performSchemaMerge(params);
    case 'Delete':
    default:
      return {
        ...params,
        PhysicalResourceId: params.associationId,
      };
  }
}

export async function isComplete(event: AWSLambda.CloudFormationCustomResourceEvent): Promise<IsCompleteResult> {
  const params = {
    associationId: event.ResourceProperties.associationId,
    mergedApiIdentifier: event.ResourceProperties.mergedApiIdentifier,
  };

  return getSchemaMergeStatus(params);
}

async function performSchemaMerge(params: any): Promise<SchemaMergeResult> {
  const command = new StartSchemaMergeCommand(params);

  try {
    const response = await appSyncClient.send(command);
    switch (response.sourceApiAssociationStatus) {
      case SourceApiAssociationStatus.MERGE_SCHEDULED:
      case SourceApiAssociationStatus.MERGE_IN_PROGRESS:
      case SourceApiAssociationStatus.MERGE_SUCCESS:
        break;
      default:
        throw new Error('Unexpected status after starting schema merge:' + response.sourceApiAssociationStatus);
    }

    return {
      ...params,
      PhysicalResourceId: params.associationId,
      sourceApiAssociationStatus: response.sourceApiAssociationStatus,
    };
  } catch (error: any) {

    // eslint-disable-next-line no-console
    console.error('An error occurred submitting the schema merge', error);
    throw error;
  }
}

async function getSchemaMergeStatus(params: GetSourceApiAssociationCommandInput): Promise<IsCompleteResult> {
  const command = new GetSourceApiAssociationCommand(params);
  var response;

  try {
    response = await appSyncClient.send(command);
  } catch (error: any) {
    // eslint-disable-next-line no-console
    console.error('Error starting the schema merge operation', error);
    throw error;
  }

  if (!response.sourceApiAssociation) {
    throw new Error(`SourceApiAssociation ${params.associationId} not found.`);
  }

  switch (response.sourceApiAssociation.sourceApiAssociationStatus) {
    case SourceApiAssociationStatus.MERGE_SCHEDULED:
    case SourceApiAssociationStatus.MERGE_IN_PROGRESS:
    case SourceApiAssociationStatus.DELETION_SCHEDULED:
    case SourceApiAssociationStatus.DELETION_IN_PROGRESS:
      return {
        IsComplete: false,
      };

    case SourceApiAssociationStatus.MERGE_SUCCESS:
      return {
        IsComplete: true,
        Data: {
          ...params,
          sourceApiAssociationStatus: response.sourceApiAssociation.sourceApiAssociationStatus,
          sourceApiAssociationStatusDetail: response.sourceApiAssociation.sourceApiAssociationStatusDetail,
          lastSuccessfulMergeDate: response.sourceApiAssociation.lastSuccessfulMergeDate?.toString(),
        },
      };

    case SourceApiAssociationStatus.MERGE_FAILED:
    case SourceApiAssociationStatus.DELETION_FAILED:
    case SourceApiAssociationStatus.AUTO_MERGE_SCHEDULE_FAILED:
      throw new Error(`Source API Association: ${response.sourceApiAssociation.associationArn} failed to merge with status: `
        + `${response.sourceApiAssociation.sourceApiAssociationStatus} and message: ${response.sourceApiAssociation.sourceApiAssociationStatusDetail}`);

    default:
      throw new Error(`Unexpected source api association status: ${response.sourceApiAssociation.sourceApiAssociationStatus}`);
  }
}
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

v2.96.2

macOS 13.4

ndejaco2:
This feature allows the source API owner to ensure that when merge conflicts are introduced, the Cloudformation stack update would fail and cause a rollback. Otherwise, if the merge is successful, the team knows that the updates were successfully applied to the merged API.

- The SourceApiAssociationMergeOperationProvider construct creates the Lambda handlers for submitting the merge operation and ensuring that it succeeded or failed.
- The SourceApiAssociationMergeOperation construct provides the ability to merge a source api to a Merged Api, invoking a merge within a Cloudformation custom
resource. If the merge operation fails with a conflict, the Cloudformation update will fail and rollback the changes to the source API in the stack. This constructs accepts a merge operation provider which allows the customer to customize the default SourceApiAssociationMergeOperationProvider class if they desire and ensures that we only require a single provider that can be used across multiple merge operations. 

Closes #27170.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-appsync/lib/source-api-association-merge.ts
+++ b/packages/aws-cdk-lib/aws-appsync/lib/source-api-association-merge.ts
@@ -1,14 +1,17 @@
 import { randomUUID } from 'crypto';
 import * as path from 'path';
-import { Construct } from 'constructs';
+import { Construct, IConstruct } from 'constructs';
 import { ISourceApiAssociation } from './source-api-association';
 import { PolicyStatement, Effect } from '../../aws-iam';
 import { Runtime, Function } from '../../aws-lambda/lib';
 import { NodejsFunction } from '../../aws-lambda-nodejs';
-import { CfnResource, CustomResource, Duration, Resource } from '../../core';
+import { CfnResource, CustomResource, Duration } from '../../core';
 import { Provider } from '../../custom-resources/lib/provider-framework';
 
-export interface ISourceApiAssociationMergeOperationProvider {
+/**
+ * This interface for the provider of the custom resource that will be used to initiate a merge operation during Cloudformation update.
+ */
+export interface ISourceApiAssociationMergeOperationProvider extends IConstruct {
 
   /**
   * Service token which is used for identifying the handler used for the merge operation custom resource.
@@ -23,7 +26,11 @@ export interface ISourceApiAssociationMergeOperationProvider {
   associateSourceApiAssociation(sourceApiAssociation: ISourceApiAssociation): void;
 }
 
-export class SourceApiAssociationMergeOperationProvider extends Resource implements ISourceApiAssociationMergeOperationProvider {
+/**
+ * SourceApiAssociationMergeProvider class is responsible for constructing the custom resource that will be used for initiating the
+ * source API merge during a Cloudformation update.
+ */
+export class SourceApiAssociationMergeOperationProvider extends Construct implements ISourceApiAssociationMergeOperationProvider {
 
   /**
    * Service token for the resource provider.
@@ -92,13 +99,16 @@ export interface SourceApiAssociationMergeOperationProps {
   readonly sourceApiAssociation: ISourceApiAssociation;
 
   /**
-   *
+   * The merge operation provider construct which is responsible for configuring the Lambda resource that will be invoked during
+   * Cloudformation update.
    */
   readonly mergeOperationProvider: ISourceApiAssociationMergeOperationProvider;
 
   /**
    * The version identifier for the schema merge operation. Any change to the version identifier will trigger a merge on the next
    * update. Use the version identifier property to control when the source API metadata is merged.
+   *
+   * @default null
    */
   readonly versionIdentifier?: string;
 
@@ -106,7 +116,7 @@ export interface SourceApiAssociationMergeOperationProps {
    * Flag indicating whether the source api should be merged on every CFN update or not.
    * If set to true and there are no changes to the source API, this will result in a no-op merge operation.
    *
-   * @default - False
+   * @default False
    */
   readonly alwaysMergeOnStackUpdate?: boolean;
 }
@@ -148,7 +158,7 @@ type MergeResourceProperties = {
  * This can be used to propagate changes from the source API to the Merged API when the association is using type MANUAL_MERGE.
  * If the merge operation fails, it will fail the Cloudformation update and rollback the stack.
  */
-export class SourceApiAssociationMergeOperation extends Resource {
+export class SourceApiAssociationMergeOperation extends Construct {
 
   constructor(scope: Construct, id: string, props: SourceApiAssociationMergeOperationProps) {
     super(scope, id);
--- a/packages/aws-cdk-lib/package.json
+++ b/packages/aws-cdk-lib/package.json
@@ -146,7 +146,7 @@
     "@aws-cdk/spec2cdk": "0.0.0",
     "@aws-sdk/client-acm": "^3.414.0",
     "@aws-sdk/client-account": "^3.414.0",
-    "@aws-sdk/client-appsync": "^3.405.0",
+    "@aws-sdk/client-appsync": "^3.414.0",
     "@aws-sdk/client-codepipeline": "^3.414.0",
     "@aws-sdk/client-cloudwatch-logs": "^3.414.0",
     "@aws-sdk/client-dynamodb": "^3.415.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -282,46 +282,47 @@
     "@smithy/util-utf8" "^2.0.0"
     tslib "^2.5.0"
 
-"@aws-sdk/client-appsync@^3.405.0":
-  version "3.405.0"
-  resolved "https://registry.npmjs.org/@aws-sdk/client-appsync/-/client-appsync-3.405.0.tgz#31659d5336a59b9b2d1e5d31cae9a910085a189b"
-  integrity sha512-4Ub/Mb5D9XGN6+A2gRvplWsoyTBdieE7khRkz1+nmfms+JREmvfRKu0ZmMWonWdrl0tlYN71cm6XTpwG2Vqbug==
+"@aws-sdk/client-appsync@^3.414.0":
+  version "3.418.0"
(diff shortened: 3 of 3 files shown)


MrArnoldPalmer:
Discussing this offline, we are moving this functionality to `awscdk-appsync-utils`. If this is something that users really want and the implementation there is successful, we can consider stabilizing this in aws-cdk-lib in the future. This will give us a chance to gather feedback and iterate on the implementation. Once released there, we will add some links from the cdk appsync docs so users will be able to find it.

PR #27231, opened 2023-09-21
chore: add label replacement action

scanlonp:
Draft: This is something I would like to do in the future. However, possible dependencies on our current labels (metrics, actions, etc) makes this something that does not need to be done at the moment.

This action replaces labels in the repository by renaming them. This changes the label at the repository level, so all issues and PRs with the label currently will be updated. It runs the action here [scanlonp/github-action-replace-labels](https://github.com/scanlonp/github-action-replace-labels).

These specific parameters will update our area labels to reflect the repository restructure.

The action is on workflow dispatch, so it will only run when manually triggered.

----


rix0rrr:
Excellent, but can't we run this once, on demand, from a workstation? Why does it need to be a GHA?

kaizencc:
Closing cuz it will take some additional work to get this up to snuff, and no one is available to do that right now.

PR #27232, opened 2023-09-21
chore: update area labels to match repository restructure

scanlonp:
Draft: This is something I would like to do in the future. However, possible dependencies on our current labels (metrics, actions, etc) makes this something that does not need to be done at the moment.

All of our "area" labels currently are prefixed with `@aws-cdk`. Since the repository was restructured to have stable modules live in `aws-cdk-lib` and alpha modules live in `@aws-cdk`, our labels should reflect that.

Additionally, updated several labels for modules that have been graduated from alpha to stable.

Dependent on #27231 and 
- [ ] running the associated action before merging this workflow update.
- [ ] labels for `@aws-cdk/aws-kinesis-firehose-destinations-alpha`, `@aws-cdk/aws-scheduler-alpha`, and `@aws-cdk/aws-scheduler-targets-alpha` need to be created.

----


.github/workflows/issue-label-assign.yml
```diff
@@ -63,224 +63,226 @@ env:
 
  AREA_AFFIXES: >
            {
-             "prefixes":["@aws-cdk/"]
+             "prefixes":["@aws-cdk/", "aws-cdk-lib/", "aws-cdk-lib."]
```
rix0rrr, line 66:
```suggestion
             "prefixes":["@aws-cdk/", "aws-cdk-lib/", "aws-cdk-lib."]
```

Can we turn 'em in to `module/xxx` or `service/xxx` ?

kaizencc:
Closing cuz it will take some additional work to get this up to snuff, and no one is available to do that right now.

PR #27236, opened 2023-09-21
feat(scheduler): disable Schedule on creation

MoartnW:
Small change to make it possible to disable a Schedule. It basically works the same as in Rule in aws-events.

----


MoartnW:
Exemption Request

Not sure what to add to the README file for this change. Also I think integration test is not required here (yet)?

lpizzinidev (changes requested):
Thanks for the contribution! 💪
I left you a couple of comments for some minor changes.
**README changes**
You can specify your new attribute in the [related README](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-scheduler-alpha/README.md).
For example, adding a new section `Disabling Schedules` after [`Grouping Schedules`](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-scheduler-alpha/README.md#grouping-schedules) explaining how to use the attribute and the default behavior (with an example).
**Integration test**
You should be able to implement an integration test in the [test folder](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/aws-scheduler-alpha/test) of the construct by following the [specified guidelines](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md).

MoartnW:
Thanks for your feedback! I did the proposed code changes, and updated the README file. I will now work on adding the integration test and will update this PR once done.

packages/@aws-cdk/aws-scheduler-alpha/test/schedule.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      role: this.role,
+    };
+  }
+}
+
+describe('Schedule', () => {
+  let stack: Stack;
+  let func: lambda.IFunction;
+  const expr = ScheduleExpression.at(new Date(Date.UTC(1969, 10, 20, 0, 0, 0)));
+  let role: iam.IRole;
```
lpizzinidev, line 25:
```suggestion
  let role: iam.IRole;
  const expr = ScheduleExpression.at(new Date(Date.UTC(1969, 10, 20, 0, 0, 0)));
```
`role` declaration should be moved together with other `let` variables

lpizzinidev (changes requested):
Nice job! 🎉
Thanks for the quick changes.
Just a super-minor change on the unit test and then it will be good to go for me.

kaizencc (approved):
Thanks for your contribution @MoartnW! This looks good to me. However, here's a quick PSA: please create open-source PRs off of a non-main branch, so we're guaranteed to have permission to push to your branch. I'm unsure if it will work this time but we're going to give it a try.

More information [here](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request)

PR #27237, opened 2023-09-21
feat(s3-deployment): support Fn::Select in renderData()

### LINKED ISSUE #25504: aws-s3-deployment: `Source.jsonData` fails for certain types of deploy-time values
### Describe the bug

Some types of deploy-time values cause `Source.jsonData` (and presumably the other deploy-time static methods for `Source`) to fail since these deploy-time values contain CloudFormation functions other than `Ref` or `Fn::Get`.

For example, the `secretName` field on a `Secret` becomes an `Fn::Join` function call which gets rejected by `Source.jsonData` (see `Reproduction Steps` for example code).

### Expected Behavior

For the rendered JSON file containing the secret ID to be deployed to the S3 bucket.

### Current Behavior

For the code for reproducing the issue, I get this error:

```
Error: Invalid CloudFormation reference. "Ref" or "Fn::GetAtt". Got {"Fn::Join":["-",[{"Fn::Select":[0,{"Fn::Split":["-",{"Fn::Select":[6,{"Fn::Split":[":",{"Ref":"testsecretF8BBC644"}]}]}]}]},{"Fn::Select":[1,{"Fn::Split":["-",{"Fn::Select":[6,{"Fn::Split":[":",{"Ref":"testsecretF8BBC644"}]}]}]}]}]]}
```

### Reproduction Steps

```typescript
const bucket = new Bucket(this, 'test-bucket');
const secret = new Secret(this, 'test-secret');
new BucketDeployment(this, 'test-bucket-deployment', {
  destinationBucket: bucket,
  sources: [Source.jsonData('key.json', { secretId: secret.secretName })],
});
```

### Possible Solution

_No response_

Seems to be rejected by this code: https://github.com/aws/aws-cdk/blob/dde8a4a423adbf5c543e93f8729af10909335100/packages/aws-cdk-lib/aws-s3-deployment/lib/render-data.ts#L61-L73

### CDK CLI Version

2.75.0 (build 37c53d6)

### Framework Version

_No response_

### Node.js Version

18

### OS

macOS

### Language

Typescript

### Language Version

TypeScript (4.9.5)

### Other information

_No response_

Justinon:
Closes #25504 

The reason for this change is to support more complex Cloudformation references used within `Source.data` in `aws-s3-deployment`. The objects today only support `Ref` or `Fn::GetAtt` Cfn references, which is limiting when it comes to attempting to manipulate Cfn references at deploy-time, such as via `Fn::Split` or `Fn::Select`. Many AWS CDK functions return tokens that must be evaluated using these complex Cfn functions (see [ApplicationTargetGroup's firstLoadBalancerFullName attribute](https://github.com/aws/aws-cdk/blob/3edd2400bc0c8a86366a29d3a7eef1ef4fa5e016/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-target-group.ts#L438)), but they are incompatible with `renderData`!

This is a blocking issue for CDK projects which rely on generating S3 objects using `BucketDeployment`, wherein the rendered data is generated from native functions which utilize Cfn functions under-the-hood to dynamically construct values at deploy-time.

----


rix0rrr (dismissed):
Honestly this whole function seems a little silly. Feels like we could accept EVERY intrinsic.

But thanks for the contribution, we're at least better off than we were before!

Justinon:
@mergifyio refresh

Justinon:
@rix0rrr The Mergify job `Queue: Embarked in merge queue` is failing for this reason:
```
Mergify doesn't have permission to update

For security reasons, Mergify can't update this pull request. Try updating locally.
GitHub response: refusing to allow a GitHub App to create or update workflow .github/workflows/request-cli-integ-test.yml without workflows permission
```
I see from the comment on this PR by the Mergify bot that:
> Your pull request will be updated from main and then merged automatically (do not update manually)

However the GitHub Action is suggesting that I try updating locally. How do I resolve this to get it embarked?

Justinon:
@Mergifyio update

scanlonp:
@Mergifyio update

scanlonp (approved):
Refresh Approval

PR #27242, opened 2023-09-21
chore(cdk): remove mockery

mrgrain:
Replace usage of mockery in some tests with a more modern way.

Closes https://github.com/aws/aws-cdk/security/dependabot/181

----


packages/aws-cdk/package.json
```diff
@@ -86,9 +86,9 @@
     "constructs": "^10.0.0",
     "fast-check": "^3.13.0",
     "jest": "^29.7.0",
+    "jest-mock": "^29.7.0",
```
mrgrain, line 89: Note, this is already installed with `jest` anyway. Just needs be explicit now, since we are importing a utility function from it.

TheRealAmazonKendra (approved):

PR #27243, opened 2023-09-21
fix(appconfig): allow multiple environment monitor roles to be created

chenjane-dev:
Currently, when adding multiple monitors under an environment, users are seeing the following error

```
Error: There is already a Construct with name 'Role' in Environment [MyEnvironment]
```

Before the logical id for any monitor role that is created was just `Role`. In this change, we use the monitor index in the logical id so there aren't any repeats in logical id's. This is also why there is a change in the integ test snapshot, since the logical id was refactored.

----


chenjane-dev:
`Exemption Request`: This snapshot change is necessary because of the logical id mentioned in the issue description

otaviomacedo (dismissed):

otaviomacedo (dismissed):

PR #27268, opened 2023-09-25
fix(custom-resources): not work fromSdkCalls using sdk v3 format

### LINKED ISSUE #27255: (custom-resources): AwsCustomResourcePolicy.fromSdkCalls does not work with v3 command format
### Describe the bug

Creating an AwsCustomResource with a service/action using the AWS CDK v3 format and a policy AwsCustomResourcePolicy.fromSdkCalls fails

### Expected Behavior

A policy is correctly created

### Current Behavior

```
Error: Action '@aws-sdk/client-ec2:ModifyInstanceMetadataOptionsCommand' is invalid. An action string consists of a service namespace, a colon, and the name of an action. Action names can include wildcards.
    at PolicyStatement.validatePolicyActions (/home/jonathan/Development/eterna/infra/node_modules/aws-cdk-lib/aws-iam/lib/policy-statement.js:2:3414)
    at PolicyStatement.addActions (/home/jonathan/Development/eterna/infra/node_modules/aws-cdk-lib/aws-iam/lib/policy-statement.js:2:1290)
    at new PolicyStatement (/home/jonathan/Development/eterna/infra/node_modules/aws-cdk-lib/aws-iam/lib/policy-statement.js:2:579)
    at new AwsCustomResource (/home/jonathan/Development/eterna/infra/node_modules/aws-cdk-lib/custom-resources/lib/aws-custom-resource/aws-custom-resource.js:1:5701
```

### Reproduction Steps

The call I'm currently running into this with:

```ts
new AwsCustomResource(instance, `enableMetadataTags`, {
  onCreate: {
    service: '@aws-sdk/client-ec2',
    action: 'ModifyInstanceMetadataOptionsCommand',
    parameters: {
      InstanceId: instance.instanceId,
      InstanceMetadataTags: 'enabled'
    },
    physicalResourceId: PhysicalResourceId.of('EnableMetadataTags')
  },
  onDelete: {
    service: '@aws-sdk/client-ec2',
    action: 'ModifyInstanceMetadataOptionsCommand',
    parameters: {
      InstanceId: instance.instanceId,
      InstanceMetadataTags: 'disabled'
    }
  },
  policy: AwsCustomResourcePolicy.fromSdkCalls({
    resources: AwsCustomResourcePolicy.ANY_RESOURCE
  })
});
```

### Possible Solution

I'm assuming to construct the correct policy, the service name needs to strip `@aws-sdk/client-` and the action needs to strip `Command`

_No response_

### CDK CLI Version

2.94.0 (build 987c329)

### Framework Version

_No response_

### Node.js Version

v18.17.1

### OS

Linux

### Language

Typescript

### Language Version

_No response_

### Other information

If my assessment is correct, this should be easy for me to submit a PR for if that would be helpful!

WinterYukky:
## Summary

Add support AWS SDK v3 format to `AwsCustomResourcePolicy.fromSdkCalls`.

## How to fix this?

There are services that cannot be addressed simply by excluding `@aws-sdk/client-` from the service name like a `@aws-sdk/client-emr` (in iam service is `elasticmapreduce`). There is a similar issue with AWS SDK v2, and in AWS SDK v2, IAM service names are resolved from API metadata. However, API Metadata is only compatible with AWS SDK v2, and since it cannot be used with v3, I created a script to register the v3 package name in API Metadata and made it compatible.

## Other
There are integration tests for AWS SDK v2 and v3 in `packages/@aws-cdk-testing/framework-integ/test/test/aws-custom-resource/integ.aws-customs-custom-resource.ts`, but the AWS SDK has already been changed to V3 and both were an AWS SDK v3 test using the v2 format. So I updated my tests for v3 and reused them for testing the v3 format.

Closes #27255.

----


rix0rrr (dismissed):
Thanks for helping out, but wait a minute.

It seems these changes are predicated on the fact that the user would be specifying the SDKv3 service name. For backwards compatibility, we also need to let people specify the SDKv2 service name (*even though* the call will be serviced using SDKv3).

So I'd rather see the following transform:

```
SDKv3 package name -> SDKv2 service name -> IAM permissions name
```

Or the alternative:

```
SDKv2 service name -> SDKv3 package name -> IAM permissions name
```

In any case: a deterministic order of mapping. 

As opposed to what it feels to me like we have now:

```
SDKv2 service name (if using) -> IAM permissions name
SDKv3 package name (if using) -> IAM permissions name
```

Because the current is harder to test.

rix0rrr:
Thanks for the pointers! I will take it from here myself I think. I'm bit bothered by the current state of the code.

rix0rrr:
> By the way, If we can add @aws-cdk/sdk-v2-to-v3 package to dependencies we will realization it... I could not it :(

I know. `sdk-v2-to-v3` is intended to be bundled into custom resource handlers. It cannot be used as a dependency of CDK construct libraries. It's all good, I'll take care of it.

PR #27274, opened 2023-09-25
feat(s3): Reduce notifications handler permissions (#5925)

### LINKED ISSUE #5925: s3 bucket notifications creates IAM policy that has no resource boundary
Forked off from https://github.com/aws/aws-cdk/issues/2781, specifically [this comment](https://github.com/aws/aws-cdk/issues/2781#issuecomment-543759478)

S3 bucket notifications creates an IAM role holding a policy that contains no resource boundary, specifically `"Resource": "*"`.

Companies typically enforce that all IAM policies should be well bounded in their actions and resource. 

https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-s3-notifications/test/integ.notifications.expected.json#L188-L208

faridnsh:
This restricts them to resources only the buckets that is required. To ensure permissions don't get too large minimize is also enabled on the policy.

Closes #5925

----

### DIFF
--- a/packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource-handler.ts
+++ b/packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource-handler.ts
@@ -52,24 +52,28 @@ export class NotificationsResourceHandler extends Construct {
   public readonly functionArn: string;
 
   /**
-   * The role of the handler's lambda function.
+   * The default policy of role of the handler's lambda function.
    */
-  public readonly role: iam.IRole;
+  private readonly policy: iam.Policy;
 
   constructor(scope: Construct, id: string, props: NotificationsResourceHandlerProps = {}) {
     super(scope, id);
 
-    this.role = props.role ?? new iam.Role(this, 'Role', {
+    const role = props.role ?? new iam.Role(this, 'Role', {
       assumedBy: new iam.ServicePrincipal('lambda.amazonaws.com'),
     });
 
-    this.role.addManagedPolicy(
+    role.addManagedPolicy(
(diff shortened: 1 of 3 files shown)


alFReD-NSH:
Clarification Request: I'm not sure if this falls as a feature or a bugfix. In either case no changes to readme or integration tests are required as there's no changes to end users and there isn't any changes in how the reosource reacts.

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource-handler.ts
```diff
@@ -52,24 +52,28 @@ export class NotificationsResourceHandler extends Construct {
   public readonly functionArn: string;
 
   /**
-   * The role of the handler's lambda function.
+   * The default policy of role of the handler's lambda function.
    */
-  public readonly role: iam.IRole;
+  private readonly policy: iam.Policy;
```
vinayak-kukreja, line 57: This is breaking. A user could be using this property which now would become `private`

vinayak-kukreja (changes requested):
Hey, I see the build is failing right now. We would not be able to prioritize review for this PR till the build succeeds. 

This could be a high impact PR since it is modifying permissions. I am not sure if this can be merged in. It looks like the integration tests were not run with this change. Could you run those?

alFReD-NSH:
Hi @vinayak-kukreja, Thanks for the review. I had a busy week but now I'm happy to make the PR again, while adding back the `role` as public to make sure it's not breaking. In the meantime, other than locally running the integration tests, what needs to be done here?

Could you also let me know whether this should a bugfix or feature and what kind of change I can put in the README and also integration test file?

PR #27287, opened 2023-09-25
feat(stepfunctions-tasks): add `ExecutionParameters` to `AthenaStartQueryExecution`

### LINKED ISSUE #27286: aws-stepfunctions-tasks: `AthenaStartQueryExecution` is missing `ExecutionParameters`
### Describe the feature

The `StartQueryExecution` call from Athena these days (Since July 2022) supports parameterized queries, which means passing in `ExecutionParameters`. These are also supported inside Step Functions as described in this [documentation](https://docs.aws.amazon.com/step-functions/latest/dg/connect-athena.html).

However, the CDK wrappering of Step Functions tasks for Athena does not support passing `ExecutionParameters`.

### Use Case

I want to use a parameterized Athena query inside Step Functions.

### Proposed Solution

Simply add `ExecutionParameters`, which is a string list, similarily to the existing `QueryString` and other parameters.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.97.0 (build d7cf3be)

irrelevant

nakedible-p:
Add missing `ExecutionParameters` option to `AthenaStartQueryExecution` step.

Closes #27286.

----


nakedible-p:
Need to manage to regenerate the integration test files and this should be good to go...

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
```diff
@@ -42,6 +42,13 @@ export interface AthenaStartQueryExecutionProps extends sfn.TaskStateBaseProps {
    * @default - No work group
    */
   readonly workGroup?: string;
+
+  /**
+   * A list of values for the parameters in a query.
```
lpizzinidev, line 47:
```suggestion
   * A list of values for the parameters in a query.
   * The values are applied sequentially to the parameters in the query in the order
   * in which the parameters occur.
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
```diff
@@ -209,6 +216,7 @@ export class AthenaStartQueryExecution extends sfn.TaskStateBase {
           OutputLocation: this.props.resultConfiguration?.outputLocation ? `s3://${this.props.resultConfiguration.outputLocation.bucketName}/${this.props.resultConfiguration.outputLocation.objectKey}/` : undefined,
         },
         WorkGroup: this.props?.workGroup,
+        ExecutionParameters: this.props.executionParameters,
```
lpizzinidev, line 219: A `private validateExecutionParameters(executionParameters?: string[])` method should be created to apply validation according to [this](https://docs.aws.amazon.com/athena/latest/APIReference/API_StartQueryExecution.html#athena-StartQueryExecution-request-ExecutionParameters) specs.
The method should be invoked in the constructor.

lpizzinidev (changes requested):
Thanks for the contribution 💪
Left you some notes on a couple of minor fixes and a missing validation.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
```diff
@@ -209,6 +216,7 @@ export class AthenaStartQueryExecution extends sfn.TaskStateBase {
           OutputLocation: this.props.resultConfiguration?.outputLocation ? `s3://${this.props.resultConfiguration.outputLocation.bucketName}/${this.props.resultConfiguration.outputLocation.objectKey}/` : undefined,
         },
         WorkGroup: this.props?.workGroup,
```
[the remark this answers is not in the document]
nakedible-p: Spotted this but didn't want to do unrelated changes in the same pull. However, happy to do so.

nakedible-p:
@lpizzinidev Thank you for the review!

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
```diff
@@ -66,10 +76,21 @@ export class AthenaStartQueryExecution extends sfn.TaskStateBase {
     this.integrationPattern = props.integrationPattern ?? sfn.IntegrationPattern.REQUEST_RESPONSE;
 
     validatePatternSupported(this.integrationPattern, AthenaStartQueryExecution.SUPPORTED_INTEGRATION_PATTERNS);
+    this.validateExecutionParameters(props.executionParameters);
 
     this.taskPolicies = this.createPolicyStatements();
   }
 
+  private validateExecutionParameters(executionParameters?: string[]) {
+    if (executionParameters == null) return;
```
lpizzinidev, line 85:
```suggestion
    if (executionParameters === undefined || Token.isUnresolved(executionParameters)) return;
```
Validation should be skipped also when the parameter is a Token.
For example when specified as [`JsonPath.listAt()`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions.JsonPath.html#static-listwbratpath).

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
```diff
@@ -66,10 +76,21 @@ export class AthenaStartQueryExecution extends sfn.TaskStateBase {
...
+    if (executionParameters.length == 0) {
+      throw new Error('Query executionParameters must not be an empty list');
+    }
+    if (executionParameters.length > 1024) {
+      throw new Error('Too many executionParameters for query, maximum 1024 supported');
+    }
```
lpizzinidev, line 91:
```suggestion
    const invalidExecutionParameters = executionParameters.some(p => p.length < 1 || p.length > 1024);
    if (invalidExecutionParameters) {
        throw new Error('\'executionParameters\' items\'s length must be between 1 and 1024 characters');
    }
```
> Length Constraints: Minimum length of 1. Maximum length of 1024.

This refers to the length of the single strings in the array.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
```diff
@@ -66,10 +76,21 @@ export class AthenaStartQueryExecution extends sfn.TaskStateBase {
...
     validatePatternSupported(this.integrationPattern, AthenaStartQueryExecution.SUPPORTED_INTEGRATION_PATTERNS);
+    this.validateExecutionParameters(props.executionParameters);
 
     this.taskPolicies = this.createPolicyStatements();
   }
 
+  private validateExecutionParameters(executionParameters?: string[]) {
+    if (executionParameters == null) return;
+    if (executionParameters.length == 0) {
+      throw new Error('Query executionParameters must not be an empty list');
```
lpizzinidev, line 87:
```suggestion
      throw new Error('\'executionParameters\' must be a non-empty list');
```
Clearer message.

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/athena/start-query-execution.test.ts
```diff
@@ -224,4 +224,58 @@ describe('Start Query Execution', () => {
       }),
     });
   });
+
+  test('execution parameters', () => {
```
lpizzinidev, line 228: Test cases should be added to check that validation rules are enforced.

lpizzinidev (changes requested):
Thanks for applying the changes! 
A couple of adjustments are needed on validation and some missing unit tests.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
nakedible-p: I have a question about that actually... I think right now this will fail if using things like `JsonPath.listAt()`, because the typescript schema says `string[]` and that will not match. I don't know enough of the machinery to say what it should be – something like `string[] | Token`?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
nakedible-p: Good catch, didn't understand that from the documentation.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
mrgrain: Nah that's fine. Tokens are pretending to be strings and string arrays.

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/athena/start-query-execution.test.ts
nakedible-p: Okay, will add them. Should probably also add a test case to see that "token" works for example with `JsonPath.array`

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
nakedible-p: No, actually `JsonPath.listAt()` returns `string[]`. But `JsonPath.array()` returns `string`. Not sure if there's anything to fix on that front or not.

lpizzinidev: `executionParameters` is `string[]` so using `JsonPath.array()` will not produce an accepted value in this case. 
It's up to the user to specify the correct `JsonPath` function, we don't need to support every `JsonPath` returned type.

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/athena/start-query-execution.test.ts
nakedible-p: Right, understood.

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/athena/start-query-execution.test.ts
```diff
@@ -224,4 +224,138 @@ describe('Start Query Execution', () => {
       }),
     });
   });
+
+  test('execution parameters succeeds', () => {
```
lpizzinidev, line 228: A couple of missing test cases:
1. `executionParameters: []`
2. `executionParameters: ['']`

lpizzinidev (changes requested):
Thanks for the changes. 
A couple of unit tests are needed to complete validation coverage.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/athena/start-query-execution.ts
```diff
@@ -66,10 +76,22 @@ export class AthenaStartQueryExecution extends sfn.TaskStateBase {
...
   }
 
+  private validateExecutionParameters(executionParameters?: string[]) {
+    if (executionParameters === undefined || cdk.Token.isUnresolved(executionParameters)) return;
+    if (executionParameters.length == 0) {
+      throw new Error('\'executionParameters\' must be a non-empty list');
+    }
+    const invalidExecutionParameters = executionParameters.some(p => p.length < 1 || p.length > 1024);
+    if (invalidExecutionParameters) {
+        throw new Error('\'executionParameters\' items\'s length must be between 1 and 1024 characters');
```
lpizzinidev, line 91:
```suggestion
      throw new Error('\'executionParameters\' items\'s length must be between 1 and 1024 characters');
```
Wrong indentation.

lpizzinidev (approved):

vinayak-kukreja (approved):
This is great. Thank you for your contribution. 

And also, thank you @lpizzinidev for reviewing the PR.

vinayak-kukreja (approved):

PR #27293, opened 2023-09-26
fix(cli): cdk destroy does not throw error for a non-existent stack

### LINKED ISSUE #27179: No exception when stack with wrong cases is deployed
### Describe the bug

When invalid stack name is given to cdk destroy, it does not give an error, but does nothing.

### Expected Behavior

Should give an error: "Stack does not exist"

### Current Behavior

Nothing happens

### Reproduction Steps
Stack "IAMStack" exists, but not "iamstack".

```
> cdk destroy iamstack
Are you sure you want to delete:  (y/n)? y

```

Also notice that it skips the stack name after the word "delete", as it is invalid.

### Possible Solution

Just a check if the stack exists, if not throw error.

_No response_

### CDK CLI Version

2.89.0 (build 2ad6683)

### Framework Version

_No response_

### Node.js Version

20

### OS

Mac M1 13.5.2 (22G91)

### Language

Typescript

### Language Version

_No response_

### Other information

I'm willing to fix the issue if core team approves :)

go-to-k:
This PR for cli is to throw an error if stacks with wrong cases (=not exist) specified in `cdk destroy`.

Closes #27179.

----


go-to-k:
Exemption Request: this should be covered with cli-integ tests. In my environment, the behavior was confirmed as expected.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -769,7 +770,10 @@ export class CdkToolkit {
       defaultBehavior: DefaultSelection.OnlySingle,
     });
 
-    // No validation
+    const notExistPatterns = selector.patterns.filter(p => !stacks.stackArtifacts.find(s => minimatch(s.hierarchicalId, p)));
+    if (notExistPatterns.length > 0) {
+      throw new Error(`Stacks not exist: ${notExistPatterns.join(', ')}`);
```
SankyRed, line 775: Can you please change the error message to be something more informative as to what does not exist and how they might be able to fix it?

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -769,7 +770,10 @@ export class CdkToolkit {
       defaultBehavior: DefaultSelection.OnlySingle,
     });
 
-    // No validation
+    const notExistPatterns = selector.patterns.filter(p => !stacks.stackArtifacts.find(s => minimatch(s.hierarchicalId, p)));
```
SankyRed, line 773: Could you also add code such that we error out before it we even try deleting any stack, if at least one of the stacks provided does not exist?

packages/aws-cdk/lib/cdk-toolkit.ts
go-to-k: @SankyRed 

For example, if you specify stacks `stack-A`, `stack-X`, and `stack-Y` in `cdk destroy`, and only `stack-A` exists in your account.

In this code, the error message is to be following, `Stacks not exist: stack-X, stack-Y`. So you can know which of stacks do not exist.

`how they might be able to fix it?`

It is to difficult to know this. If you enter non-existent stacks, CDK does not have the way to fix it unless it gets all the stacks in the AWS account and does a fuzzy search on all of them. Even if we did that, you probably wouldn't get the expected and accurate output.

packages/aws-cdk/lib/cdk-toolkit.ts
go-to-k: @SankyRed 

In the current code, if any of the stacks are not existing, none of the stacks are deleted. And the cdk occurs an error.

Is this not what you are assuming?

packages/aws-cdk/lib/cdk-toolkit.ts
TheRealAmazonKendra: A tad more context here might be helpful, if we can provide it. Maybe something like `Cannot run cdk destroy on stack(s) ${stacks.join(', ')}. ${notExistPatterns.join(', ')} are not deployed in account: ${account}, region: ${region}.`

TheRealAmazonKendra (changes requested):

TheRealAmazonKendra:
> Exemption Request: this should be covered with cli-integ tests. In my environment, the behavior was confirmed as expected.

I would like to see this added as a failure case in the cli integ tests. Could you please add a test case there?

go-to-k:
@TheRealAmazonKendra 

> A tad more context here might be helpful, if we can provide it. Maybe something like Cannot run cdk destroy on stack(s) ${stacks.join(', ')}. ${notExistPatterns.join(', ')} are not deployed in account: ${account}, region: ${region}.

This `notExistPattern` handling checks the non-existence stacks by looking stack instance definitions in the cdk app (that is cloud-assembly), so I think whether are deployed or not in an account with `account` and `region` cannot be checked.

How about the following message?:

```
Cannot run cdk destroy on stack(s) ${stacks.join(', ')}. ${notExistPatterns.join(', ')} don't exist.
```

go-to-k:
@TheRealAmazonKendra 

I added cli integ tests and changed an error message for now. 

This is my first time doing cli-integ-tests, so please let me know if I am missing something.

https://github.com/aws/aws-cdk/pull/27293/commits/3dbc266fc070e095c306f5a7ed22db41bac5ca6b

packages/aws-cdk/lib/cdk-toolkit.ts
go-to-k: @TheRealAmazonKendra 

> A tad more context here might be helpful, if we can provide it. Maybe something like Cannot run cdk destroy on stack(s) ${stacks.join(', ')}. ${notExistPatterns.join(', ')} are not deployed in account: ${account}, region: ${region}.

I think the format is difficult.

This `notExistPattern` handling and `cdk destroy` check the non-existence stacks by looking not the AWS account but stack instance definitions in the cdk app (that is cloud-assembly).

My understanding is that `cdk destroy` does not run the command by specifying an account or region, but looks at the stack list ([`CloudFormationStackArtifact[]`](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/cx-api/lib/artifacts/cloudformation-artifact.ts)) in cloud-assembly and if each stack has a region and account tied to it If it is tied to a region and account, it is used; otherwise, the default region is used.
So we know that the stack entered does not exist in the CDK app in the first place, but we do not know which account/region it does not exist in.

Then, how about the following message?:

```
Cannot run cdk destroy on stack(s) ${stacks.join(', ')}. ${notExistPatterns.join(', ')} not exist.
```

go-to-k:
@TheRealAmazonKendra 

Could you please review my changes and answers?

Issue #23532, opened 2023-01-02
awsstepfunctions|awsstepfunctionstasks: Allow custom state names


jerrydevin96:
### Describe the feature

We use the go bindings for AWS CDK (https://github.com/aws/aws-cdk-go) to deploy state machines. The functions provided to create a state machine in the library do not have the option to add a custom state name, which is possible while creating a state machine from the console. instead while using the CDK, the state name is auto-generated from other inputs provided while defining the state machine. This auto-generated state name is too long and confusing.

### Use Case

Too long state names are confusing. Having the possibility to add a custom user-defined state name would be really good.

### Proposed Solution

Add additional struct field to define custom state names.

### Other Information

I am raising this issue here, as the Go Bindings (https://github.com/aws/aws-cdk-go) do not have issues enabled.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.49.1

### Environment details (OS name and version, etc.)

macOS v12.5.1

jerrydevin96:
Will this have to be implemented in the current repository (https://github.com/aws/aws-cdk) or should it be implemented in the go bindings repository?

Jacco:
I do not understand the issue raised here. Looking at the code it seems to me that everywhere the `logical id` of the `State` construct is used as the name of the `State` in the rendered JSON.

bun913:
Hello!
May I challenge this issue?

bun913:
I was going to start working on this Issue, but as per the following comment, I reconsidered whether it is a problem on the naming convention side.

https://github.com/aws/aws-cdk/issues/23532#issuecomment-1374506128

I was going to start working on this issue, but then I reconsidered and wondered if it would be possible to write the logic on the naming side so that it would not be too long.

PR #27306, opened 2023-09-27
feat(stepfunctions): adding custom state name prop

SankyRed:
Closes https://github.com/aws/aws-cdk/issues/23532

----


packages/aws-cdk-lib/aws-stepfunctions/lib/states/fail.ts
```diff
@@ -9,6 +9,8 @@ import { INextable } from '../types';
 export interface FailProps {
   /**
    * Optional name for this state
+   *
+   * @default LogicalResourceId
```
rix0rrr, line 13: I don't think that this is accurate. It's more accurate to say something like:

```ts
* @default - The construct ID will be used as state name
```

packages/aws-cdk-lib/aws-stepfunctions/lib/states/succeed.ts
```diff
@@ -54,6 +61,7 @@ export class Succeed extends State {
   public toStateJson(): object {
     return {
       Type: StateType.SUCCEED,
+      StateName: this.stateName,
```
rix0rrr, line 64: If we put `StateName` in this return object, what ultimately happens to it?

Have you seen the documentation for [Amazon States Language](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-amazon-states-language.html)?

This is what we are targeting. It looks like this:

```ts
{
  "StartAt": "<StateName>",
  "States": {
    "<StateName1>": {
      "Type": "<Whatever>",
      <Properties>
    },
    "<StateName2>": {
      "Type": "<MoreWhatever>",
    }
}
```

That is ultimately what we want to change.

rix0rrr (changes requested):

packages/aws-cdk-lib/aws-stepfunctions/lib/states/fail.ts
SankyRed: Okay, will update that. Thanks Rico.

kaizencc (commented):
Can you try again with the title? Couple things:

- this is a feature, not a chore. if you feel like you need exemptions for integ tests, you can add the relevant labels yourself.
- standard is to not capitalize words
- since you are working in a particular module, the standard is to include the module name in the title (i.e. `feat(stepfunctions):`)
- try to rework the title to be more descriptive and describe the feature that has been added. if you've followed the other steps we're now at "feat(stepfunctions): adding custom state name prop". I don't think people who read the title will immediately know where the custom state name prop can be added.

packages/aws-cdk-lib/aws-stepfunctions/README.md
```diff
@@ -124,6 +124,7 @@ becomes new the new Data object. This behavior can be modified by supplying valu
 
 These properties impact how each individual step interacts with the state machine data:
 
+* `stateName`: the custom state name to each state if required else we used the construct id
```
rix0rrr, line 127: "if required"

Required by who?

rix0rrr: Please use punctuation as well.

"the name of the state in the state machine definition. If not supplied, defaults to the construct id."

Or something.

packages/aws-cdk-lib/aws-stepfunctions/README.md
```diff
@@ -249,6 +250,8 @@ An arbitrary JSON object (specified at execution start) is passed from state to
 state and transformed during the execution of the workflow. For more
 information, see the States Language spec.
 
+You can add custom state names to all states and if not it defaults to construct id.
```
rix0rrr, line 253: Did we not already say this?

packages/aws-cdk-lib/aws-stepfunctions/README.md
```diff
@@ -281,10 +284,12 @@ The `Pass` state also supports passing key-value pairs as input. Values can
 be static, or selected from the input with a path.
 
 The following example filters the `greeting` field from the state input
-and also injects a field called `otherData`.
+and also injects a field called `otherData`. You can also add custom state name
+to Pass state.
```
rix0rrr, line 288: I don't think we need to add this everywhere.

packages/aws-cdk-lib/aws-stepfunctions/README.md
```diff
@@ -281,10 +284,12 @@ The `Pass` state also supports passing key-value pairs as input. Values can
...
 
 ```ts
 const pass = new sfn.Pass(this, 'Filter input and inject data', {
+  stateName: 'my-pass-state',
```
rix0rrr, line 292: Leave this out I think? Or give a single example on `stateName`, adding a comment saying what the default is.

packages/aws-cdk-lib/aws-stepfunctions/lib/states/choice.ts
```diff
@@ -47,7 +54,7 @@ export class Choice extends State {
   public readonly endStates: INextable[] = [];
 
   constructor(scope: Construct, id: string, props: ChoiceProps = {}) {
-    super(scope, id, props);
+    super(scope, props.stateName? props.stateName: id, props);
```
rix0rrr, line 57: This changes the construct ID though.

So passing `stateName` now has 2 effects, affecting the construct ID as well as the state name.

It's just an override for `id` now, which is not what we want.

rix0rrr (changes requested):

rix0rrr (approved):
Excellent!

PR #27340, opened 2023-09-28
feat(events): add multiple event bus policies on a single event bus

### LINKED ISSUE #24671: aws-events: EventBus and EventBusPolicy only accept a single statement
### Describe the bug

The EventBus only accepts a single IAM statement in the policy. If I want to grant multiple principals different access to a bus, it should be possible with `addToResourcePolicy`, but that method explicitly denies additional statements. Defining an `EventBusPolicy` does not alleviate the issue, because `EventBusPolicy` also only accepts a single iam.PolicyStatement.

### Expected Behavior

The documentation for `EventBusPolicy` and `EventBus` indicate that additional permissions should be added through the `addToResourcePolicy` method.
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_events.EventBus.html#applywbrremovalwbrpolicypolicy
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_events.EventBusPolicy.html
> Policies define the operations that are allowed on this resource.
>
>You almost never need to define this construct directly.
>
>All AWS resources that support resource policies have a method called addToResourcePolicy(), which will automatically >create a new resource policy if one doesn't exist yet, otherwise it will add to the existing policy.
>
>Prefer to use addToResourcePolicy() instead.

Expected behavior would be that `addToResourcePolicy` creates or updates the existing resource policy with additional statements.

### Current Behavior

New statements are silently ignored, and the construct does not support adding additional permissions. 
It looks like the behavior to only accept a single statement is codified in the `addToResourcePolicy` implementation. Shouldn't the policy contain a policy document rather than a single statement?

https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/aws-events/lib/event-bus.ts#L341

It seems odd that not having a SID would cause an error to be thrown, while not adding new permissions would allow execution to continue and potentially deploy a change that doesn't have the correct policy.

### Reproduction Steps

Create a bus, add multiple resource policies
```typescript
import * as events from 'aws-cdk-lib/aws-events';
import * as iam from 'aws-cdk-lib/aws-iam';

export class BusStack extends cdk.Stack {
  constructor(scope: Construct, id: string, props: cdk.StackProps) {
    super(scope, id, props);
    const bus = new events.EventBus(this, 'my-bus', {
      eventBusName: 'my-bus'
    });

    const externalPrincipal1 = new iam.AccountPrincipal('someaccountid');
    const addPublishForFooSource = new iam.PolicyStatement({
      effect: iam.Effect.ALLOW,
      principals: [externalPrincipal1],
      sid: `uniquesid1`,
      resources: [bus.eventBusArn],
      actions: ['events:PutEvents'],
    });
    addPublishForFooSource.addCondition('StringEquals', { 'events:source': ['foo'] });
    //succeeds
    bus.addToResourcePolicy(addPublishForFooSource);

    const externalPrincipal2 = new iam.AccountPrincipal('someotheraccountid');
    const  addPublishForBarSource = new iam.PolicyStatement({
      effect: iam.Effect.ALLOW,
      principals: [externalPrincipal2],
      sid: `uniquesid2`,
      resources: [bus.eventBusArn],
      actions: ['events:PutEvents'],
    });
    addPublishForFooSource.addCondition('StringEquals', { 'events:source': ['bar'] });
    //no error thrown, but no policy update
    bus.addToResourcePolicy(addPublishForBarSource);
  }
}
```

### Possible Solution

Modify the `EventBusPolicy` to accept a PolicyDocument rather than a PolicyStatement, and update the `addToResourcePolicy` to append new statements to the existing policy.

_No response_

### CDK CLI Version

2.69

### Framework Version

_No response_

### Node.js Version

v16.17.1

### OS

OSX

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

postsa:
Enable the creation of multiple event bus policies on a single event bus.

Closes #24671.

The result of the Policies created by the integration test is a resource policy on the event bus that looks like 

```json
{
  "Version": "2012-10-17",
  "Statement": [{
    "Sid": "Statement2",
    "Effect": "Allow",
    "Principal": {
      "AWS": "arn:aws:iam::<account-id>:root"
    },
    "Action": "events:PutRule",
    "Resource": "arn:aws:events:us-west-2:<account-id>:event-bus/StackBusAA0A1E4B"
  }, {
    "Sid": "Statement1",
    "Effect": "Allow",
    "Principal": {
      "AWS": "arn:aws:iam::<account-id>:root"
    },
    "Action": "events:PutEvents",
    "Resource": "arn:aws:events:us-west-2:<account-id>:event-bus/StackBusAA0A1E4B"
  }]
}
```

----


packages/aws-cdk-lib/aws-events/lib/event-bus.ts
```diff
@@ -309,7 +309,10 @@ export class EventBus extends EventBusBase {
    */
   public readonly eventSourceName?: string;
 
-  private policy?: EventBusPolicy;
+  /**
+   * The EventBusPolicies attached to this event bus
+   */
+  public readonly policies: EventBusPolicy[];
```
lpizzinidev, line 315:
```suggestion
```
I think the variable should be removed since it is not used at the moment.

packages/aws-cdk-lib/aws-events/test/event-bus.test.ts
```diff
@@ -580,27 +580,35 @@ describe('event bus', () => {
...
     // WHEN
-    const add1 = bus.addToResourcePolicy(statement);
-    const add2 = bus.addToResourcePolicy(statement);
+    const add1 = bus.addToResourcePolicy(statement1);
+    const add2 = bus.addToResourcePolicy(statement2);
 
     // THEN
     expect(add1.statementAdded).toBe(true);
-    expect(add2.statementAdded).toBe(false);
+    expect(add2.statementAdded).toBe(true);
```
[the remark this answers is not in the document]
postsa: done

packages/aws-cdk-lib/aws-events/lib/event-bus.ts
postsa: Not by the lib, but since we are transforming policy statements into event bus policies behind the scenes it seems considerate to provide an accessor for the result. Happy to remove it if you believe strongly otherwise.

lpizzinidev: The pre-existing `policy` variable was `private` and used only to avoid creating duplicate policies.
I don't think it is a good idea to expose a variable that could potentially be changed/removed in the future (considering the note above) and it currently serves no purpose.
The only reason I would keep it is to check for constraints on policies applied on an event bus ([should not exceed 10KB](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-eventbuspolicy.html#:~:text=The%20permission%20policy%20on%20the%20event%20bus%20cannot%20exceed%2010%20KB%20in%20size.)), but still, I would keep it `private`.
I don't think that this validation is necessary to be honest, so I would just remove the variable.

lpizzinidev (approved):
Looks good, thanks 👍
But could need [higher-level evaluation](https://github.com/aws/aws-cdk/pull/27340#pullrequestreview-1660038522).

bs-u27:
+1

kaizencc (approved):
I think this is the best we're going to get, so I don't mind adding this in. It's not really a hack to just specify more `EventBusPolicy` resources, we do this elsewhere though I can't come up with an example off the top of my head.

PR #27357, opened 2023-09-29
docs(s3-deployment): dataJson should be jsonData

mtfurlan:
The aws_s3_deployment.Source has jsonData and yamlData, but one of the
comments that generates documentation says dataJson and dataYaml, which
don't exist.

----


lpizzinidev (approved):
Nice catch 😎

kaizencc (approved):

PR #27424, opened 2023-10-05
feat(scheduler-targets): step function start execution target

### LINKED ISSUE #27377: aws-scheduler-targets-alpha: Add StepFunctionsStartExecution
### Describe the feature

Currently the only impelmentation of [IScheduleTarget](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-scheduler-alpha.IScheduleTarget.html) is LambdaInvoke.

It would be great if we could have also StepFunctionsStartExecution as one of the templated apis in the aws console.

grafik

### Use Case

You can have a lot of targets in event scheduler, but in this alpha module its only possible to trigger a lambda. I would like to be able to trigger a Step Function/State Machine from there.

### Proposed Solution

The implementation can be very similar/almost identical like [LambdaInvoke](https://github.com/aws/aws-cdk/blob/b2a895ef285e5451e64c21e179172e998c479582/packages/%40aws-cdk/aws-scheduler-targets-alpha/lib/lambda-invoke.ts#L8).

I've tested following implementation in a private project succesfully:

```typescript
export class StepFunctionsStartExecution
  extends ScheduleTargetBase
  implements IScheduleTarget
{
  constructor(
    private readonly stateMachine: IStateMachine,
    private readonly props: ScheduleTargetBaseProps,
  ) {
    super(props, stateMachine.stateMachineArn);
  }

  protected addTargetActionToRole(schedule: ISchedule, role: IRole): void {
    const stateMachineEnv = this.stateMachine.env;
    if (!sameEnvDimension(stateMachineEnv.region, schedule.env.region)) {
      throw new Error(`Cannot assign stateMachine in region ${this.stateMachine.env.region} to the schedule ${Names.nodeUniqueId(schedule.node)} in region ${schedule.env.region}. Both the schedule and the stateMachine must be in the same region.`);
    }

    if (!sameEnvDimension(stateMachineEnv.account, schedule.env.account)) {
      throw new Error(`Cannot assign stateMachine in account ${this.stateMachine.env.account} to the schedule ${Names.nodeUniqueId(schedule.node)} in account ${schedule.env.region}. Both the schedule and the stateMachine must be in the same account.`);
    }

    if (this.props.role && !sameEnvDimension(this.props.role.env.account, stateMachineEnv.account)) {
        throw new Error(`Cannot grant permission to execution role in account ${this.props.role.env.account} to invoke target ${Names.nodeUniqueId(this.stateMachine.node)} in account ${this.stateMachine.env.account}. Both the target and the execution role must be in the same account.`);
      }

    this.stateMachine.grantStartExecution(role);
  }
}

```

### Other Information

I would like to contribute this feature during the [hacktoberfest](https://hacktoberfest.com). 
It would be great if this repository can be marked with the topic "hacktoberfest" so that it gets counted towards the goal (see https://hacktoberfest.com/participation/#maintainers for more info) :)

### Acknowledgements

### CDK version used

2.99.1

windows, wsl

WtfJoke:
A StepFunctionStartExecution ScheduleTarget was implemented similar to the already existing [LambdaInvoke](https://github.com/aws/aws-cdk/blob/b2a895ef285e5451e64c21e179172e998c479582/packages/%40aws-cdk/aws-scheduler-targets-alpha/lib/lambda-invoke.ts#L8) target.

I've added an integration test, which will trigger an eventbridge schedule once. This schedule has a step function as a target. The step function creates a parameter with a given value, which will get verified by the test

Closes #27377 

----


packages/cdk-cli-wrapper/.warnings.jsii.js
```diff
@@ -0,0 +1,73 @@
+function cdk_cli_wrapper_ICdk(p) {
```
filletofish, line 1: Looks like cdk-cli-wrapper folder was committed by mistake?

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/stepfunctions-start-execution.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Names } from 'aws-cdk-lib';
+import { IRole } from 'aws-cdk-lib/aws-iam';
+import { IStateMachine } from 'aws-cdk-lib/aws-stepfunctions';
+import { ScheduleTargetBase, ScheduleTargetBaseProps } from './target';
+import { sameEnvDimension } from './util';
+
+/**
+ * Use an AWS Step function as a target for AWS EventBridge Scheduler.
+ */
+export class StepFunctionsStartExecution extends ScheduleTargetBase implements IScheduleTarget {
```
filletofish, line 11: I would recommend to add unit tests for this target too. See https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-scheduler-targets-alpha/test/target.test.ts#L9. We probably need to rename that test file and create one file with unit tests per each target.

filletofish (changes requested):
Thank you for the commit! It motivated me to add more issues for Scheduler Targets to get help from the community to build all targets for L2 Scheduler constructs.

packages/cdk-cli-wrapper/.warnings.jsii.js
WtfJoke: Yeah, I commited this by mistake. Good spot! Thanks

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/stepfunctions-start-execution.ts
WtfJoke: I created a new unit test which contains almost identical unit tests for step functions.

I also renamed the existing file `target.test.ts` into `lambda-invoke.test.ts`.

WtfJoke:
@filletofish thanks for the review! I'm glad you created more issues for it, I thought about contributing more targets, but have chosen the one I needed the most :)

I've addressed your comments, please have a look on it again. 
Thank you!

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/stepfunctions-start-execution.ts
```diff
@@ -0,0 +1,35 @@
...
+      throw new Error(`Cannot assign stateMachine in region ${stateMachineEnv.region} to the schedule ${Names.nodeUniqueId(schedule.node)} in region ${schedule.env.region}. Both the schedule and the stateMachine must be in the same region.`);
+    }
+
+    if (!sameEnvDimension(stateMachineEnv.account, schedule.env.account)) {
+      throw new Error(`Cannot assign stateMachine in account ${stateMachineEnv.account} to the schedule ${Names.nodeUniqueId(schedule.node)} in account ${schedule.env.region}. Both the schedule and the stateMachine must be in the same account.`);
+    }
+
+    if (this.props.role && !sameEnvDimension(this.props.role.env.account, stateMachineEnv.account)) {
+      throw new Error(`Cannot grant permission to execution role in account ${this.props.role.env.account} to invoke target ${Names.nodeUniqueId(this.stateMachine.node)} in account ${stateMachineEnv.account}. Both the target and the execution role must be in the same account.`);
+    }
```
WtfJoke, line 31: It might make sense to extract this common logic block somewhere, do you have a favorite spot for it? Should it go to `ScheduleTargetBase` or in a new class/function?

filletofish: @kaizencc do you have any opinion on this? I personally do not mind the duplication of the code

filletofish (approved):

kaizencc (approved):
Thanks for the submission @WtfJoke!

PR #27425, opened 2023-10-06
feat(ec2): Add SubnetFilter for CIDR Range

### LINKED ISSUE #27421: aws-ec2: Filter subnets by CIDR Range
### Describe the feature

I would like the ability to filter a SubnetSelection based on a CIDR range in a VPC. While SubnetFilter does allow to filter by CIDR mask and IP address, a VPC can have multiple CIDR Ranges associated with it and the CIDR masks can be the same while the range can be different.

### Use Case

I would like this feature because some of the VPCs I have worked with have 2 or more CIDR Ranges associated and the current subnet filter options don't really work unless I filter by Id.

Simple Example: 

VPC A can have CIDR ranges: `10.2.0.0/16` and `100.64.0.0/16`

Its subnets can be as follows: 

Subnet A: `10.2.0.0/18`
Subnet B: `10.2.64.0/18`
Subnet C: `100.64.0.0/18`
Subnet D: `100.64.64.0/18`

I want to only use the subnets under the `10.2.0.0/16` CIDR range.

### Proposed Solution

I think the feature would look something like this

```
const subnetsByCidrRanges: SubnetSelection = {
      subnetFilters: [SubnetFilter.byCidrRanges([10.2.0.0/16])]
    }
````

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.93.0

Mac OS

### PR DESCRIPTION
**Pull request by AlemanCS**
This PR adds a new SubnetFilter ,`CidrRangesSubnetFilter`, which filter subnets if they are included in a given list of CIDR ranges. This becomes helpful if a VPC has multiple CIDR ranges and you want to select subnets for only a specific CIDR range or multiple.

Note: It is a bit difficult to write unit tests as CDK doesn't have the ability to add a secondary CIDR to a VPC.

Closes #27421 

----

### CODE REVIEW

File: packages/aws-cdk-lib/aws-ec2/test/vpc.test.ts [hunk 1]
```diff
@@ -2261,6 +2261,46 @@ describe('vpc', () => {
 
     });
 
+    test('can filter by multiple CIDR Ranges', () => {
```
lpizzinidev, line 2264: > Note: It is a bit difficult to write unit tests as CDK doesn't have the ability to add a secondary CIDR to a VPC.

Adding a test case to check that [subnet creation fails](https://github.com/aws/aws-cdk/blob/06bea3159e137a8e85e362795ef2cf7e219e381c/packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts#L748-L750) with no subnets if non-overlapping CIDR ranges are supplied could help increase coverage.

packages/aws-cdk-lib/aws-ec2/README.md
```diff
@@ -180,6 +180,7 @@ Which subnets are selected is evaluated as follows:
   * `onePerAz`: chooses at most one subnet per availability zone
   * `containsIpAddresses`: chooses a subnet which contains *any* of the listed ip addresses
   * `byCidrMask`: chooses subnets that have the provided CIDR netmask
+  * `byCidrRanges`: chooses subnets which are inside any of the specified CIDR Ranges
```
lpizzinidev, line 183:
```suggestion
  * `byCidrRanges`: chooses subnets which are inside any of the specified CIDR ranges
```
Unnecessary capitalization (also in `byCidrRanges` function and `CidrRangesSubnetFilter` class heading comments)

lpizzinidev (changes requested):
Nice job 👍
Just some minor adjustments and a test case to be added in my opinion.

packages/aws-cdk-lib/aws-ec2/test/vpc.test.ts
```diff
@@ -2261,6 +2261,46 @@ describe('vpc', () => {
...
+      // IP space is split into 6 pieces, one public/one private per AZ
+      const vpc = new Vpc(stack, 'VPC', {
+        ipAddresses: IpAddresses.cidr('10.0.0.0/16'),
+        maxAzs: 3,
+      });
+
+      // WHEN
+      // We want to place this endpoint in the same subnets as these IPv4
+      // address.
+      // WHEN
```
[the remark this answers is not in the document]
AlemanCS: Removed duplicate comment

packages/aws-cdk-lib/aws-ec2/lib/subnet.ts
```diff
@@ -169,3 +176,33 @@ class CidrMaskSubnetFilter extends SubnetFilter {
...
+   * Executes the subnet filtering logic.
+   */
+  public selectSubnets(subnets: ISubnet[]): ISubnet[] {
+    return this.checkCidrRanges(subnets, this.cidrRanges);
+  }
+
+  private checkCidrRanges(subnets: ISubnet[], cidrRanges: string[]): ISubnet[] {
+    const vpcCidrs = cidrRanges.map(cidr => {
+      return new CidrBlock(cidr);
+    });
```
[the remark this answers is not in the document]
AlemanCS: Fixed!

packages/aws-cdk-lib/aws-ec2/test/vpc.test.ts
AlemanCS: Added additional test case

packages/aws-cdk-lib/aws-ec2/README.md
AlemanCS: Fixed capitalization!

lpizzinidev (approved):
Thanks 👍

packages/aws-cdk-lib/aws-ec2/lib/subnet.ts
```diff
@@ -43,6 +43,13 @@ export abstract class SubnetFilter {
     return new CidrMaskSubnetFilter(mask);
   }
 
+  /**
```
vinayak-kukreja, line 46: NIT: Hey, could you add `@params` for the function argument here?

AlemanCS: Just added this `@param cidrs List of CIDR ranges to filter subnets from`

sumupitchayan (approved):

PR #27427, opened 2023-10-06
feat(eks): add neuron device plugin for Inferentia managed node groups

freschri:
In an EKS cluster with Inferentia instances, Neuron device plugin is only installed in case of autoscaling groups.
The change expands the logic to managed node groups.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-eks/lib/cluster.ts
+++ b/packages/aws-cdk-lib/aws-eks/lib/cluster.ts
@@ -1800,10 +1800,16 @@ export class Cluster extends ClusterBase {
    * @param options options for creating a new nodegroup
    */
   public addNodegroupCapacity(id: string, options?: NodegroupOptions): Nodegroup {
-    return new Nodegroup(this, `Nodegroup${id}`, {
+    const mng = new Nodegroup(this, `Nodegroup${id}`, {
       cluster: this,
       ...options,
     });
+
+    if (options?.instanceTypes?.some(instanceType => nodeTypeForInstanceType(instanceType) === NodeType.INFERENTIA)) {
+      this.addNeuronDevicePlugin();
+    }
+
+    return mng;
   }
 
   /**


kaizencc (changes requested):
I'm not sure what is going on here re: `addNeuronDevicePlugin`, and there's not really any additional information for me to figure it out. Is there a linked issue that can provide more background? Can you state more clearly what the intention is of this PR in the PR description? Lastly, what we definitely will need is at least 1 unit test

freschri:
> I'm not sure what is going on here re: `addNeuronDevicePlugin`, and there's not really any additional information for me to figure it out. Is there a linked issue that can provide more background? Can you state more clearly what the intention is of this PR in the PR description? Lastly, what we definitely will need is at least 1 unit test

You can find a description of what the Neuron device plugin is, in here: https://awsdocs-neuron.readthedocs-hosted.com/en/latest/containers/tutorials/k8s-setup.html (tap on Deploy Neuron...):
"Neuron device plugin exposes Neuron cores & devices to kubernetes as a resource. aws.amazon.com/neuroncore, aws.amazon.com/neurondevice, aws.amazon.com/neuron are the resources that the neuron device plugin registers with the kubernetes. aws.amazon.com/neuroncore is used for allocating neuron cores to the container. aws.amazon.com/neurondevice is used for allocating neuron devices to the container. When neurondevice is used all the cores belonging to the device will be allocated to container."
The device plugin is then required not only when instances are launched in an auto scaling group but also in managed group.

The change is replicating the same logic present in addAutoScalingGroupCapacity, see here: https://github.com/aws/aws-cdk/blob/7d11e7b3beb714fa008414b2c7e14dcf94ecd8d2/packages/aws-cdk-lib/aws-eks/lib/cluster.ts#L1786

Note that e.g. eksctl takes care of that, see here https://docs.aws.amazon.com/eks/latest/userguide/inferentia-support.html:
"The eksctl utility detects that you are launching a node group with an Inf1 instance type and will start your nodes using one of the Amazon EKS optimized accelerated Amazon Linux AMIs. "

The CDK needs to implement the same behaviour, otherwise managed node groups with Inferentia and Trainium devices will not be fully supported.

PR #27442, opened 2023-10-07
feat(lambda-python-alpha): add without-urls option for poetry

### LINKED ISSUE #27103: @aws-cdk/aws-lambda-python-alpha: allow "--without-urls" option to poetry export
### Describe the feature

I would like the poetry export command to be able to export without the url.

### Use Case

the projects I would like to use cdk and poetry with use an authenticated codeartifact url for dependencies which are sourced in the pyproject.toml.

When Poetry exports a requirements.txt it has an un-auth'd URL in the file which causes any install to fail waiting for username and password input.  Being able to export without a url would fix this problem.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.85.0

Mac OS 13.5.1 (22G90)

msambol:
Added `without-urls` option for poetry. Also added Python versions for the integration tests.

Closes #27103.

----

### DIFF
--- a/.github/workflows/issue-label-assign.yml
+++ b/.github/workflows/issue-label-assign.yml
@@ -239,6 +239,8 @@ env:
             {"area":"@aws-cdk/aws-s3outposts","keywords":["aws-s3outposts","s3outposts"],"labels":["@aws-cdk/aws-s3outposts"]},
             {"area":"@aws-cdk/aws-sagemaker","keywords":["aws-sagemaker","sagemaker"],"labels":["@aws-cdk/aws-sagemaker"]},
             {"area":"@aws-cdk/aws-sam","keywords":["serverless-application-model","aws-sam","sam"],"labels":["@aws-cdk/aws-sam"]},
+            {"area":"@aws-cdk/aws-scheduler","keywords":["aws-scheduler","scheduler"],"labels":["@aws-cdk/aws-scheduler"]},
+            {"area":"@aws-cdk/aws-scheduler-targets","keywords":["aws-scheduler-targets","scheduler-targets"],"labels":["@aws-cdk/aws-scheduler-targets"]},
             {"area":"@aws-cdk/aws-sdb","keywords":["simpledb","aws-sdb","sdb"],"labels":["@aws-cdk/aws-sdb"]},
             {"area":"@aws-cdk/aws-secretsmanager","keywords":["secret","aws-secretsmanager","secrets-manager"],"labels":["@aws-cdk/aws-secretsmanager"]},
             {"area":"@aws-cdk/aws-securityhub","keywords":["aws-securityhub","security-hub"],"labels":["@aws-cdk/aws-securityhub"]},
--- a/.github/workflows/request-cli-integ-test.yml
+++ b/.github/workflows/request-cli-integ-test.yml
@@ -19,7 +19,7 @@ jobs:
           persist-credentials: false
       - name: Find changed cli files
         id: changed-cli-files
-        uses: tj-actions/changed-files@8238a4103220c636f2dad328ead8a7c8dbe316a3
+        uses: tj-actions/changed-files@db153baf731265ad02cd490b07f470e2d55e3345
         with:
           base_sha: ${{ github.event.pull_request.base.sha }}
           files_yaml: |
--- a/.github/workflows/v2-pull-request.yml
+++ b/.github/workflows/v2-pull-request.yml
@@ -32,6 +32,6 @@ jobs:
           yarn install --frozen-lockfile
           yarn pkglint
       - name: push
-        uses: stefanzweifel/git-auto-commit-action@v4
+        uses: stefanzweifel/git-auto-commit-action@v5
         with:
           commit_message: 'automatic pkglint fixes'
--- a/packages/@aws-cdk/aws-glue-alpha/lib/job.ts
+++ b/packages/@aws-cdk/aws-glue-alpha/lib/job.ts
@@ -384,13 +384,13 @@ export interface SparkUIProps {
   /**
    * The bucket where the Glue job stores the logs.
    *
-   * @default a new bucket will be created.
+   * @default - a new bucket will be created.
    */
   readonly bucket?: s3.IBucket;
 
   /**
    * The path inside the bucket (objects prefix) where the Glue job stores the logs.
-   * Use format `'/foo/bar'`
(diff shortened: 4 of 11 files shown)


packages/@aws-cdk/aws-lambda-python-alpha/README.md
```diff
@@ -124,6 +124,20 @@ new python.PythonFunction(this, 'function', {
...
+```ts
+new python.PythonFunction(this, 'function', {
+  entry: '/path/to/poetry-function',
+  runtime: Runtime.PYTHON_3_8,
+  bundling: {
+    poetryIncludeHashes: true,
+    poetryWithoutUrls: true,
+  },
+});
+```
```
lpizzinidev, line 140:
```suggestion
**Including hashes*

You can include hashes in `poetry` using the optional boolean parameter `poetryIncludeHashes`:

```ts
new python.PythonFunction(this, 'function', {
  entry: '/path/to/poetry-function',
  runtime: Runtime.PYTHON_3_8,
  bundling: {
    poetryIncludeHashes: true,
  },
});
```
Clearer by keeping the two options separate.

packages/@aws-cdk/aws-lambda-python-alpha/README.md
```diff
@@ -124,6 +124,20 @@ new python.PythonFunction(this, 'function', {
...
```
lpizzinidev, line 141:
```suggestion

**Excluding URLs*

You can exclude URLs in `poetry` using the optional boolean parameter `poetryWithoutUrls`:

```ts
new python.PythonFunction(this, 'function', {
  entry: '/path/to/poetry-function',
  runtime: Runtime.PYTHON_3_8,
  bundling: {
    poetryWithoutUrls: true,
  },
});
```
```

File: packages/@aws-cdk/aws-lambda-python-alpha/test/bundling.test.ts [hunk 3]
```diff
@@ -322,13 +322,27 @@ test('Bundling a function with poetry dependencies, with hashes', () => {
...
     }),
   }));
+});
 
-  const files = fs.readdirSync(assetCode.path);
-  expect(files).toContain('index.py');
-  expect(files).toContain('pyproject.toml');
-  expect(files).toContain('poetry.lock');
-  // Contains hidden files.
-  expect(files).toContain('.ignorefile');
```
lpizzinidev, line 331: Any reason for removing these assertions from the above test case?

packages/@aws-cdk/aws-lambda-python-alpha/test/integ.function.poetry.ts
```diff
@@ -17,89 +17,116 @@ class TestStack extends Stack {
 
     const pythonFunction39 = new lambda.PythonFunction(this, 'my_handler_inline', {
```
lpizzinidev, line 18: I think it's better to keep the existing test cases as they are and:
1. Add `pythonFunction37WithoutUrls`, `pythonFunction38WithoutUrls`, `pythonFunction39WithoutUrls`
2. Add `pythonFunction310`, `pythonFunction310WithHashes`, `pythonFunction310Excludes`, `pythonFunction310WithoutUrls`, `pythonFunction311`, `pythonFunction311WithHashes`, `pythonFunction311Excludes`, `pythonFunction311WithoutUrls` for 3.10 and 3.11

lpizzinidev (changes requested):
Thanks 💪
The implementation looks good.
Some adjustments are needed on documentation and tests in my opinion.

packages/@aws-cdk/aws-lambda-python-alpha/test/integ.function.poetry.ts
msambol: @lpizzinidev The older versions of Python are erroring out:

```
{
  "ExecutedVersion": "$LATEST",
  "FunctionError": "Unhandled",
!! Expected 200 but received {"errorMessage": "Unable to import module 'index': urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'OpenSSL 1.0.2k-fips  26 Jan 2017'. See: https://github.com/urllib3/urllib3/issues/2168", "errorType": "Runtime.ImportModuleError", "stackTrace": []}
  "Payload": "{\"errorMessage\": \"Unable to import module 'index': urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'OpenSSL 1.0.2k-fips  26 Jan 2017'. See: https://github.com/urllib3/urllib3/issues/2168\", \"errorType\": \"Runtime.ImportModuleError\", \"stackTrace\": []}",
  "StatusCode": 200
}
```

I can look at updating dependencies if you want?

lpizzinidev: I'm not a Python expert, but maybe this could be solved by creating two handlers.
A `lambda-handler-poetry` with `urllib3` v1.x for runtimes with Python <= 3.9 and a `lambda-handler-poetry` with `urllib3` v2.x for runtimes with Python >= 3.10.
Not sure if it'll work or if it's the right approach.

packages/@aws-cdk/aws-lambda-python-alpha/test/lambda-handler-poetry/basic.py
```diff
@@ -0,0 +1,4 @@
+def handler(event, context):
+  print('Basic handler')
+
+  return 200
```
msambol, line 4: Add this basic handler with no dependencies for Python <= 3.8.

packages/@aws-cdk/aws-lambda-python-alpha/test/integ.function.poetry.ts
msambol: Done. Added a simple Python function with no dependencies.

packages/@aws-cdk/aws-lambda-python-alpha/test/bundling.test.ts
msambol: I thought there was a good reason for this but I can't remember why—I added it back.

lpizzinidev (approved):
Thanks 💪

sumupitchayan (approved):
This looks good to me - thank you for your contribution!

PR #27486, opened 2023-10-10
docs(s3): Bad type definition for objectsKeyPattern parameter

### LINKED ISSUE #27481: aws-s3: Bad type definition for objectsKeyPattern parameter
### Describe the issue

The parameter `objectsKeyPattern` for `grantRead(identity, objectsKeyPattern?)` has the type `any`; But it seems it needs to be a `string` and nothing else.
I initially thought the parameter had to be a `object` due to the name.

Same wrong type is found for 

- `grantDelete(identity, objectsKeyPattern?)`
- `grantPut(identity, objectsKeyPattern?)`
- `grantReadWrite(identity, objectsKeyPattern?)`
- `grantWrite(identity, objectsKeyPattern?, allowedActionPatterns?)`

Incidentally `grantPutAcl(identity, objectsKeyPattern?)` has correct type.

### Links

- https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Bucket.html#grantwbrdeleteidentity-objectskeypattern
- https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Bucket.html#grantwbrputidentity-objectskeypattern
- https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Bucket.html#grantwbrputwbraclidentity-objectskeypattern
- https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Bucket.html#grantwbrreadidentity-objectskeypattern
- https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Bucket.html#grantwbrreadwbrwriteidentity-objectskeypattern
- https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Bucket.html#grantwbrwriteidentity-objectskeypattern-allowedactionpatterns

alinasir85:
Fixed the type of objectsKeyPattern parameter as string

Closes #27481

----

### DIFF
--- a/packages/aws-cdk-lib/aws-codepipeline-actions/test/cloudformation/pipeline-actions.test.ts
+++ b/packages/aws-cdk-lib/aws-codepipeline-actions/test/cloudformation/pipeline-actions.test.ts
@@ -446,15 +446,15 @@ class RoleDouble extends iam.Role {
 }
 
 class BucketDouble extends s3.Bucket {
-  public grantRead(identity: iam.IGrantable, _objectsKeyPattern: any = '*'): iam.Grant {
+  public grantRead(identity: iam.IGrantable, _objectsKeyPattern: string = '*'): iam.Grant {
     return iam.Grant.drop(identity, '');
   }
 
-  public grantWrite(identity: iam.IGrantable, _objectsKeyPattern: any = '*'): iam.Grant {
+  public grantWrite(identity: iam.IGrantable, _objectsKeyPattern: string = '*'): iam.Grant {
     return iam.Grant.drop(identity, '');
   }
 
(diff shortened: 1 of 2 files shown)


kaizencc:
We cannot do this because it is a breaking change to a stable module (retyping `any` as `string`). Even if we argue that only a `string` is valid here, it's unlikely but possible that there's some implementation of `IBucket` out there that uses `objectsKeyPattern: any` and would be broken by this. Closign this PR because we cannot accept breaking changes here.

While I agree that this is an oversight the ship has sailed and the best we can do is provide better documentation to what types our implementation of `Bucket.grantXxx` accepts. 

Also, this is neither here or there but this is clearly not a "docs" change.

PR #27497, opened 2023-10-11
fix(secretsmanager): RotateSchedule is not disabled when automaticallyAfter is 0

### LINKED ISSUE #27460: (aws_secretsmanager): automaticallyAfter is not disabling the automatic rotation of secrets
### Describe the bug

As per doc, if the property automaticallyAfter is set to Duration.days(0) then it should disable automatic rotation, however it is removing the property 

```
 RotationRules:
        AutomaticallyAfterDays: 365
```
from the resource AWS::SecretsManager::RotationSchedule

### Expected Behavior

The resource AWS::SecretsManager::RotationSchedule completely removed from the stack template which will make the call API [CancelRotateSecret](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_CancelRotateSecret.html)

### Current Behavior

It is removing the below property which is not making any change in the secret's rotation.

```
 RotationRules:
        AutomaticallyAfterDays: 365
```

### Reproduction Steps

```
secret_test = secretsmanager.Secret(self, "Secret",
                                            secret_object_value={
                                                "username": cdk.SecretValue.unsafe_plain_text("foo"),
                                                "database": cdk.SecretValue.unsafe_plain_text("foo"),
                                            }
                                            )

rotation_fn = _lambda.Function(self, 'RotationFN', runtime=_lambda.Runtime.PYTHON_3_9,
                                       handler='index.handler', code=_lambda.Code.from_inline("hello world"))

secret_test.add_rotation_schedule(
            "SecretRotation", automatically_after=cdk.Duration.days(0), rotate_immediately_on_update=False, rotation_lambda=rotation_fn)
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.97.1

### Framework Version

_No response_

### Node.js Version

20.7.0

### OS

macos

### Language

Python

### Language Version

_No response_

### Other information

_No response_

go-to-k:
This PR fixes the bug that a RotateSchedule is not disabled when automaticallyAfter is 0.

Closes #27460.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-secretsmanager/lib/rotation-schedule.ts
+++ b/packages/aws-cdk-lib/aws-secretsmanager/lib/rotation-schedule.ts
@@ -127,21 +127,19 @@ export class RotationSchedule extends Resource {
     // Prevent secrets deletions when rotation is in place
     props.secret.denyAccountRootDelete();
 
-    if (props.automaticallyAfter && props.automaticallyAfter.toMilliseconds() === 0) {
-      return;
+    if (!props.automaticallyAfter || props.automaticallyAfter.toMilliseconds() !== 0) {
+      const rotationRules = {
+        automaticallyAfterDays: props.automaticallyAfter?.toDays() || 30,
+      };
+
+      new CfnRotationSchedule(this, 'Resource', {
+        secretId: props.secret.secretArn,
+        rotationLambdaArn: props.rotationLambda?.functionArn,
+        hostedRotationLambda: props.hostedRotation?.bind(props.secret, this),
+        rotationRules,
+        rotateImmediatelyOnUpdate: props.rotateImmediatelyOnUpdate,
+      });
     }
-
-    const rotationRules = {
-      automaticallyAfterDays: props.automaticallyAfter?.toDays() || 30,
-    };
-
-    new CfnRotationSchedule(this, 'Resource', {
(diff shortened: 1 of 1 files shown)


go-to-k:
Duration.days(0) should disable automatic rotation.

```ts
  /**
   * Specifies the number of days after the previous rotation before
   * Secrets Manager triggers the next automatic rotation.
   *
   * A value of zero will disable automatic rotation - `Duration.days(0)`.
   *
   * @default Duration.days(30)
   */
  readonly automaticallyAfter?: Duration;
```

https://github.com/aws/aws-cdk/blob/f9dcef27b90c6730a01498bd3fe294e765771219/packages/aws-cdk-lib/aws-secretsmanager/lib/rotation-schedule.ts#L36-L44

lpizzinidev (approved):
Thanks 👍

scanlonp:
I believe that the current behavior is intended (see discussion on issue https://github.com/aws/aws-cdk/issues/27460#issuecomment-1765130471). Using the 'magic' value of 0 on `automaticallyAfter` was designed to get around a limitation of being forced into the default rather than leaving it unset.

As of now, I do not think this change is necessary. There may still be a case for improving the doc string to be more descriptive of what the actual behavior is.

go-to-k:
@scanlonp 

OK, I will another PR for modifying the doc.

PR #17984, opened 2021-12-13
fix(region-info): ssm service principal is wrong in majority of regions

rix0rrr:
The SSM service principal format depends on the region. Older regions have a "global" service principal (`ssm.amazonaws.com`), while newer regions have only regional service principals (`ssm.ap-east-1.amazonaws.com`).

A number of things have been changed to address this:

* Add the notion of a "region order" into the `region-info` library. This allows us to express things like "does this region predate or postdate the change of some convention", and allows us to express that certain regions are *after* SSM introduced this change.
* For region-agnostic stacks, it is no longer possible to supply a single value for the template that will suffice in all regions, as the *format itself* will have changed (neither `"ssm.amazonaws.com"` nor `"ssm.$REGION.amazonaws.com"` will work in all regions). That means we must always introduce a lookup map for region-agnostic stacks. Add `stack.regionalFact()` to generate lookup maps from facts in case it is necessary.
  * Detect if all map values are just an instantiation of a token pattern, and return the simplification if possible (e.g.: if the lookup values are `service.us-east-1.amazonaws.com`, `service.us-east-2.amazonaws.com`, etc, then simplify to `service.$REGION.$URL_SUFFIX` and avoid emitting a lookup).   
  * Simplify existing usage sites of `RegionInfo.regionMap()` in Lambda and CodeBuild to use the new `stack.regionalFact()`.
* Because lookup maps would always include information for all regions, including GovCloud regions, and those are only rarely necessary: add the infrastructure for users to restrict what partitions they want to include information for, by means of a context flag. Defaults to all regions if not specified (so we don't break old templates), but for new projects restricts itself to `['aws', 'aws-cn']`. Set to just `['aws']` for integration tests so we don't break all of our snapshot tests.

Fixes #16188, fixes #17646.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #27506, opened 2023-10-11
fix(region-info): facts only returned from constant region list

### LINKED ISSUE #27260: region-info: Fact.regions behavior does not match description.
### Describe the bug

In fact.ts

https://github.com/aws/aws-cdk/blob/ab52bb55aae8cf8510bb6788f3bd88a71b648567/packages/aws-cdk-lib/region-info/lib/fact.ts#L7-L14

the description of `regions` it states it returns any region that has at least one fact registered. This is untrue, it is only listing from the fixed list defined in `AWS_REGIONS` from aws-entities.ts.

Either the description needs to be updated, or the logic needs to pull the list of regions from `this.database` instead of the `AWS_REGIONS` export.

### Expected Behavior

1. Call `Fact.regions` and save the array as `call1`
3. Register a new fact with a region that has yet to be added to the list in `AWS_REGIONS`
4. Call `Fact.regions` as `call2`
5. The `call2` array would include the name of my newly added region and the `call1` array would not.

### Current Behavior

1. Call `Fact.regions` and save the array as `call1`
3. Register a new fact with a region that has yet to be added to the list in `AWS_REGIONS`
4. Call `Fact.regions` as `call2`
5. The `call2` array is identical to the `call1` array.

### Reproduction Steps

```
  it('can find new region in `Fact.region`', () => {
    Fact.register({
      region: 'us-unreleased-1',
      name: FactName.PARTITION,
      value: 'aws',
    });
    expect(Fact.regions.includes('us-unreleased-1')).toBeTruthy();
  })
  ```

### Possible Solution

Switch the code from `Array.from(AWS_REGIONS)` to `Array.from(Object.keys(this.database))` or update the documentation to say it is a fixed list based on `AWS_REGIONS` in aws-entities.ts

_No response_

### CDK CLI Version

2.95.0 (build cfa7e88)

### Framework Version

_No response_

### Node.js Version

v18.17.1

### OS

OSX 13.5.2 (22G91)

### Language

Typescript

### Language Version

TypeScript (4.3.0)

### Other information

You can reach out to me on the Amazon internal slack as `@chamsco`

msambol:
Facts are only being returned from the regions in a constant list.
This PR returns facts for all Regions in `AWS_REGIONS` + `this.database`.

Closes #27260.

----

### DIFF
--- a/packages/@aws-cdk/app-staging-synthesizer-alpha/package.json
+++ b/packages/@aws-cdk/app-staging-synthesizer-alpha/package.json
@@ -83,10 +83,6 @@
       "AWSLINT_BASE_CONSTRUCT": true
     }
   },
-  "dependencies": {
-    "aws-cdk-lib": "0.0.0",
-    "constructs": "^10.0.0"
-  },
   "devDependencies": {
     "aws-cdk-lib": "0.0.0",
     "@aws-cdk/integ-runner": "0.0.0",
--- a/packages/@aws-cdk/aws-appconfig-alpha/package.json
+++ b/packages/@aws-cdk/aws-appconfig-alpha/package.json
@@ -87,8 +87,6 @@
     "jest": "^29.7.0"
   },
   "dependencies": {
-    "aws-cdk-lib": "0.0.0",
-    "constructs": "^10.0.0",
     "mime-types": "^2.1.35"
   },
   "homepage": "https://github.com/aws/aws-cdk",
--- a/packages/@aws-cdk/cx-api/package.json
+++ b/packages/@aws-cdk/cx-api/package.json
@@ -79,7 +79,6 @@
     "organization": true
   },
   "dependencies": {
-    "@aws-cdk/cloud-assembly-schema": "0.0.0",
     "semver": "^7.5.4"
   },
   "peerDependencies": {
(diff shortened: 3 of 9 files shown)


lpizzinidev (commented):
> @returns the list of names of AWS regions for which there is at least one registered fact.

In my opionion updating the `regions()` method to return `Array.from(Object.keys(this.database))` is the way to go here.
Also, the PR should be a `fix`.

msambol:
@lpizzinidev Updated per your feedback—thanks!

msambol:
@lpizzinidev For this, I _think_ it makes sense to ignore the linter error about integration tests?

lpizzinidev (changes requested):
Thanks 👍
I agree that an integration test is not necessary in this case.
Also, the title should be updated [to describe the bug fixed](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request).

msambol:
@lpizzinidev I updated the title per your recommendation.

lpizzinidev (approved):
Thanks! 👍

msambol:
@kaizencc Can you this be approved/merged without the linter passing?

msambol:
Exemption Request

`region-info` doesn't have integration tests.

mrgrain:
This has been "static" for a long time now:
https://github.com/aws/aws-cdk/pull/17984

I'm also concerned that we are using this list in other places and rely on it containing _all_ regions.

Can we change the code and description so that it includes all AWS Regions _plus_ any custom defined regions?

msambol:
@mrgrain Updated per your recommendation. Thanks!

mrgrain (approved):

PR #27532, opened 2023-10-13
fix(lambda-event-sources): does not support setting maxRecordAge to infinity

### LINKED ISSUE #27321: aws_lambda_event_sources: Missing support for maxRecordAge=-1 value in  KinesisEventSourceProps
### Describe the bug

The service supports -1 as a value but CDK imposes a Duration object which does not support negative value. This can become an issue in case the Kinesis Stream Retention window is set to > 7 days (which is the max allowed for maxRecordAge). when this happens customer may lose records.

### Expected Behavior

the KinesisEventSourceProps should support  a way to set the maxRecordAge value to -1. (Duration(-1)?)

### Current Behavior

Duration(-1) is not allowed and no other way to set this value

image

### Reproduction Steps

self explanatory

### Possible Solution

allow Duration.seconds(-1)  as a way to pass through the value? 

_No response_

### CDK CLI Version

latest

### Framework Version

_No response_

### Node.js Version

all

### OS

all

### Language

Typescript

### Language Version

all

### Other information

_No response_

lpizzinidev:
Allows explicitly setting `maxRecordAge` to infinity (`-1`).

Closes #27321.

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-lambda-event-sources/test/integ.kinesis.js.snapshot/lambda-event-source-kinesis.template.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-lambda-event-sources/test/integ.kinesis.js.snapshot/lambda-event-source-kinesis.template.json
@@ -99,6 +99,7 @@
     "FunctionName": {
      "Ref": "FC4345940"
     },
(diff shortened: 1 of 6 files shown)


lpizzinidev:
Wrong approach

PR #27559, opened 2023-10-16
fix(cdk): Add AppSync:Api_Key as hot swappable and fix a bug with AppSync.function

Amplifiyer:
Add AppSync:Api_Key as hots wappable and fix the bug where AppSync.function doesn't allow setting version and runtime

1. Allow `expires` property of resource `AWS::AppSync::Api_Key` to be hot-swappable
2. Read the Api_Key_id from the physical ARN if not available from resource properties. (It's optional in CFN but mandatory in SDK)
3. UpdateFunction doesn't [allow](https://docs.aws.amazon.com/appsync/latest/APIReference/API_UpdateFunction.html) setting both `functionVersion` and `runtime` in the SDK (allowed in CFN). Update to remove one based on if `code` is provided or `mappingTemplates`
4. Fix a bug where the file returned from S3 was not being decoded from buffer.
5. Increase the timeout and number of retries for concurrent modification of AppSync.Functions.

----


packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -124,7 +144,7 @@ async function simpleRetry(fn: () => Promise<any>, numOfRetries: number, errorCo
     await fn();
   } catch (error: any) {
     if (error && error.code === errorCodeToRetry && numOfRetries > 0) {
-      await sleep(500); // wait half a second
+      await sleep(1000); // wait half a second
```
comcalvi, line 147:
```suggestion
      await sleep(1000); // wait a whole second
```

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -104,6 +115,15 @@ export async function isHotswappableAppSyncChange(
           if (schemaCreationResponse.status === 'FAILED') {
             throw new Error(schemaCreationResponse.details);
           }
+        } else { //isApiKey
+          if (!sdkRequestObject.id) {
+            // ApiKeyId is optional in CFN but required in SDK. Grab the KeyId from physicalArn if not available as part of CFN template
+            const arnParts = physicalName?.split('/');
+            if (arnParts && arnParts.length === 4) {
+              sdkRequestObject.id = arnParts[3];
+            }
```
comcalvi, line 124: can we use https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.Arn.html here and do `Arn.split` instead?

comcalvi (changes requested):
only minor comments, looks great!

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
Amplifiyer: it seems the `Arn.split` only considers one resourceType or resourceName but not "nested" resource such as in this case `arn:aws:appsync:us-east-1:111111111111:apis/apiId/apikeys/api-key-id`
`Arn.split` returns the entire `apiId/apikeys/api-key-id` as the resourceName which then requires a split anyways.

```js
{
    service: 'appsync',
    resource: 'apis',
    partition: 'aws',
    region: 'us-east-1',
    account: '111111111111',
    resourceName: 'apiId/apikeys/api-key-id',
    sep: '/',
    arnFormat: 'arn:aws:service:region:account:resource/resourceName'
}
```
Moreover, `Arn.split` is in `aws-cdk-lib` and the change will require we bump the dependency from dev to runtime which might cause cyclic dependency issue? Let me know what you think

comcalvi (approved):

comcalvi (approved):

PR #27569, opened 2023-10-17
fix(events-targets): CloudWatch Log Group Resource Policy should be disabled (under feature flag)

### LINKED ISSUE #17002: (aws-events-targets): Use custom resource policy for CloudWatch Log Group event target
### General Issue

CloudWatch Log Group event target always creates resource Policy

### The Question

When I deploy a stack using the following code, CDK is automatically creating a custom resource to create a CloudWatch resource policy and results in hitting the maximum CloudWatch resource policies of 10 per region:
```
  test_fanout_event_log_group = LogGroup(stack, "TestFanout", retention=RetentionDays.ONE_DAY)

  test_fanout_event_rule = Rule(stack, "TestFanoutRule",
      event_pattern=EventPattern(
          source=['bb.identity-provider-test'],
          detail_type=['Identity Provider Created']
      ),
      event_bus=event_bus,
  )
  test_fanout_event_rule.add_target(
      CloudWatchLogGroup(test_fanout_event_log_group)
  )
```
I'm wondering if there is a way to use an existing CloudWatch resource policy instead of having one generated automatically.

### CDK CLI Version

1.124.0

### Framework Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

### Language

Python

### Language Version

3.8.12

### Other information

_No response_

aritola-sxm:
This is a proposed solution to #17002. In summary, aws-events-targets.CloudWatchLogGroup will always create a CloudWatch Log Group Resource Policy, which have a max limit of 10 per AWS environment. This behavior is untenable in large environments and instead a single Resource Policy should be created separately, similar to how the AWS Console will do it.

I want to run the approach by the CDK team before I do the polish work.

TODOs:
- [ ] Update FF README
- [ ] Write and run tests

Closes #17002.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-events-targets/lib/log-group.ts
+++ b/packages/aws-cdk-lib/aws-events-targets/lib/log-group.ts
@@ -5,7 +5,8 @@ import { RuleTargetInputProperties, RuleTargetInput, EventField, IRule } from '.
 import * as iam from '../../aws-iam';
 import * as logs from '../../aws-logs';
 import * as cdk from '../../core';
-import { ArnFormat, Stack } from '../../core';
+import { ArnFormat, Stack, FeatureFlags} from '../../core';
+import { EVENTS_TARGET_DISABLE_LOG_GROUP_RESOURCE_POLICY } from '../../cx-api';
 
 /**
  * Options used when creating a target input template
@@ -78,6 +79,14 @@ export interface LogGroupProps extends TargetBaseProps {
    * @default - the entire EventBridge event
    */
   readonly logEvent?: LogGroupTargetInput;
+
+  /**
+   * Automatically configure an AWS CloudWatch Log Group Resource Policy for Events Target.
+   *
+   * @default true
+   * @default - false if `@aws-cdk/aws-events:eventsTargetDisableLogGroupResourcePolicy` is enabled, true otherwise
+   */
+  readonly logGroupResourcePolicy?: boolean;
 }
 
 /**
@@ -91,9 +100,6 @@ export class CloudWatchLogGroup implements events.IRuleTarget {
    * Returns a RuleTarget that can be used to log an event into a CloudWatch LogGroup
    */
   public bind(_rule: events.IRule, _id?: string): events.RuleTargetConfig {
-    // Use a custom resource to set the log group resource policy since it is not supported by CDK and cfn.
-    const resourcePolicyId = `EventsLogGroupPolicy${cdk.Names.nodeUniqueId(_rule.node)}`;
-
     const logGroupStack = cdk.Stack.of(this.logGroup);
 
     if (this.props.event && this.props.logEvent) {
@@ -107,15 +113,10 @@ export class CloudWatchLogGroup implements events.IRuleTarget {
 
     _rule.node.addValidation({ validate: () => this.validateInputTemplate() });
 
-    if (!this.logGroup.node.tryFindChild(resourcePolicyId)) {
-      new LogGroupResourcePolicy(logGroupStack, resourcePolicyId, {
-        policyStatements: [new iam.PolicyStatement({
-          effect: iam.Effect.ALLOW,
-          actions: ['logs:PutLogEvents', 'logs:CreateLogStream'],
-          resources: [this.logGroup.logGroupArn],
-          principals: [new iam.ServicePrincipal('events.amazonaws.com')],
-        })],
-      });
+    const logGroupResourcePolicyDefault = FeatureFlags.of(this.logGroup).isEnabled(EVENTS_TARGET_DISABLE_LOG_GROUP_RESOURCE_POLICY) ? false : true;
+    const logGroupResourcePolicy = this.props.logGroupResourcePolicy ?? logGroupResourcePolicyDefault;
+    if (logGroupResourcePolicy) {
+      this._configureLogGroupResourcePolicy(_rule, logGroupStack);
     }
 
     return {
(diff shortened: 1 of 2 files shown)


packages/aws-cdk-lib/aws-events-targets/lib/log-group.ts
```diff
@@ -5,7 +5,8 @@ import { RuleTargetInputProperties, RuleTargetInput, EventField, IRule } from '.
 import * as iam from '../../aws-iam';
 import * as logs from '../../aws-logs';
 import * as cdk from '../../core';
-import { ArnFormat, Stack } from '../../core';
+import { ArnFormat, Stack, FeatureFlags} from '../../core';
```
scanlonp, line 8:
```suggestion
```
This was not needed even before your change. We can get rid of it now.

packages/aws-cdk-lib/aws-events-targets/lib/log-group.ts
```diff
@@ -78,6 +79,14 @@ export interface LogGroupProps extends TargetBaseProps {
...
    */
   readonly logEvent?: LogGroupTargetInput;
+
+  /**
+   * Automatically configure an AWS CloudWatch Log Group Resource Policy for Events Target.
+   *
+   * @default true
+   * @default - false if `@aws-cdk/aws-events:eventsTargetDisableLogGroupResourcePolicy` is enabled, true otherwise
+   */
+  readonly logGroupResourcePolicy?: boolean;
```
scanlonp, line 89:
```suggestion
  /**
   * Automatically configure an AWS CloudWatch Log Group Resource Policy for Events Target.
   *
   * Note: if set to true, you must create and configure a CloudWatch Log Group Resource Policy separately.
   *
   * Can be used to avoid hitting the limit of 10 CloudWatch Log Group Resources in an AWS environment.
   *
   * @default false
   */
  readonly disableLogGroupResourcePolicyCreation?: boolean;
```
Open to discussion on the prop name + doc string, but I think this is closer to what we want to convey.

scanlonp (changes requested):
This looks good! Please go ahead with the readme and tests.

However, I do not think this needs to be behind a feature flag. The new property will allow the user to control if the policy is created or not, and keeping the default behavior consistent with the current behavior will not be breaking. If you have a reason for keeping this behind a feature flag, I am open to discussion.

aritola-sxm:
Hey @scanlonp, I appreciate you taking an early look! I will make time to add the tests, README, and naming changes.

> If you have a reason for keeping this behind a feature flag, I am open to discussion.

IMO the feature flag is important to be able to change what the default behavior is, similar to how the `@aws-cdk/aws-apigateway:disableCloudWatchRole` FF works. Large AWS environments can easily end up hitting the limit.

If you disagree with this reasoning, I'm happy to remove the FF because it's an improvement regardless.

scanlonp:
@aritola-sxm, after some discussion with the team, I think that there are 3 states for how we handle policy creation:

1. Create policy for each log group (legacy default)
2. Create 1 policy for all log groups in a stack (possible feature flag default) 
4. Disable creation of policies; depend on user to configure their own policy

State 1 is what we currently have, and it will stay as the default for backwards compatibility.

State 2 may be a more ideal default. We can check if a log group policy has been made in the stack, and add new log groups to that policy automatically. It looks like that may have been the intention of the current logic, but looking for the `resourcePolicyId` in different scopes (`this.logGroup.node.tryFindChild` rather than `this.logGroupStack.node.tryFindChild`) may be why it is not working. If you would like to implement this as well, that would be terrific! This could be the default behind a feature flag.

State 3 is your proposal. This would take the guardrails off policy creation. This will not be a default (even behind a feature flag), but will cover the cases described in the original issue.

This will require your new optional prop to be an enum type rather than a boolean (even if state 2 is not implemented right now).
If you have any questions, feel free to ask!

aritola-sxm:
@scanlonp Thanks for your response.

All 3 states make sense to me. I'm happy to take a stab at fixing state 2 per your recommendation. In my context, we require state 3 as the resource policy will generally not exist in the stack(s) using this trigger. 

Could you explain more of the rationalization behind why making state 3 a configurable default is not tenable? It means without additional code changes, teams cannot deploy more than 10 distinct stacks in a given AWS environment containing this trigger. That limit seems a bit low to me, even for companies not using shared AWS environments.

Regardless, I have blocked off time early next week to implement the changes we're aligned on.

scanlonp:
@aritola-sxm, sure. The general idea is that if you want behavior other than the default, you should configure it when defining the resource in the code. Your use case is an opt-out of creating a policy. We see that as something that should be defined in code.

On the other hand, if we discover that a default value should be changed, we define a new default, but put it behind a feature flag for backwards compatibility. We do not use feature flags to enable non-default behavior across the app / stack.

I understand it is not the most convenient to change the code for each individual resource that you want to have this new behavior, but that is our practice.

scanlonp:
@aritola-sxm I do see the benefit of changing a default value for any given resource property. I think investigating ways we can let users easily define thin wrappers to edit specific resource defaults across an app / stack would be worthwhile.

aritola-sxm:
> I understand it is not the most convenient to change the code for each individual resource that you want to have this new behavior, but that is our practice.

No problem! I'll create a new PR for this as soon as I can, per the decisions from this PR.

I've had a series of crunch-time projects and have struggled to make the time to wrap this up. It's still important, just not as urgent as other things on my plate right now 😅

scanlonp:
@aritola-sxm, do you want to make a new PR? I can reopen this one if you would like. If you prefer to start clean, more power to you.

PR #27572, opened 2023-10-17
fix: remove omitEmptyArray default in Ec2Service constructor

### LINKED ISSUE #27555: aws_cdk.aws_ecs: empty placement constraints / placement strategies not allowed
### Describe the bug

Altering PlacementConstraints and PlacementStrategies in CDK to remove them does not remove PlacementConstraints and Placement Strategies

**Findings**: the ECS UpdateService API used by CloudFormation allows for the use of an empty list passed to PlacementConstraints to remove existing placement constraints. The same applies to PlacementStrategies (though the documentation is less clear, and indicates "to remove an existing placement strategy, specify an empty object"), which implies CDK should emit a null here if strategies are empty/ should be removed.

Source: [ECS API Reference](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_UpdateService.html#ECS-UpdateService-request-placementConstraints)

**Status quo**: CDK does not allow the use of an empty array for these properties as it evaluates PlacementConstraints using:

`Lazy.any({ produce: () => this.constraints }, { omitEmptyArray: true })`

Reference: [Implementation](https://github.com/aws/aws-cdk/blob/1b5435359a4d8a5d2d83244b13e4f27b5dc46d25/packages/aws-cdk-lib/aws-ecs/lib/ec2/ec2-service.ts#L182)

Therefore, setting PlacementConstraints to an empty array results in no property being generated. Furthermore, the use of an escape hatch (setting PlacementConstraints via `addPropertyOverride` does not seem to update this property.

### Expected Behavior

I expect to be able to update/ remove placement strategies and placement constraints. Removing the constraints triggers an update call without the parameter set, which means that the properties are not updated. 

### Current Behavior

Placement constraints cannot be removed easily. Placement strategies cannot be removed once set. 

### Reproduction Steps

Python code to reproduce the issue:

```python
#!/usr/bin/env python3

import aws_cdk as cdk
import aws_cdk.aws_ec2 as EC2
import aws_cdk.aws_ecs as ECS

app = cdk.App()
stack = cdk.Stack(app, "Stack")
vpc = EC2.Vpc(stack, "Vpc")
cluster = ECS.Cluster(stack, "Cluster", vpc=vpc)
asg = cluster.add_capacity("DefaultAutoScalingGroup", instance_type=EC2.InstanceType("t2.micro"))

task_definition = ECS.Ec2TaskDefinition(stack, "TaskDefinition")
container = task_definition.add_container("web", image=ECS.ContainerImage.from_registry("amazon/amazon-ecs-sample"), memory_limit_mib=256)

# the constraints and strategies are not set
service = ECS.Ec2Service(stack, "Service", cluster=cluster, task_definition=task_definition, placement_constraints=[], placement_strategies=[])

# this does allow us to set the placement constraints to an empty array
cfn_service: ECS.CfnService = service.node.default_child
cfn_service.placement_constraints = []

# we have no way to set the placement strategies to None - this doesn't work
cfn_service.placement_strategies = None

# the escape hatch for placement strategies also leads to some surprise - this doesn't work
cfn_service.add_property_override("PlacementStrategies", None)

app.synth()
```

### Possible Solution

CDK should always provide empty values for these parameters if they are not set, rather than not providing the parameters to avoid surprise on stack update when removing placement constraints. 

_No response_

### CDK CLI Version

2.99.1

### Framework Version

2.98.0

### Node.js Version

18.16.0

### OS

Mac OS

### Language

Python

### Language Version

Python (3.11.3)

### Other information

_No response_

123MwanjeMike:
[Removes omitEmptyArray default in Ec2Service constructor](https://github.com/aws/aws-cdk/pull/27572/commits/f39f743073d3e56b5c576b1d7f2b75f6928eeca7) to fix #27555 

Closes #27555 

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ecs/test/ec2/ec2-service.test.ts
+++ b/packages/aws-cdk-lib/aws-ecs/test/ec2/ec2-service.test.ts
@@ -2098,32 +2098,6 @@ describe('ec2 service', () => {
       }).toThrow();
     });
 
-    test('with an empty list in placement strategies', () => {
-      // GIVEN
-      const stack = new cdk.Stack();
-      const vpc = new ec2.Vpc(stack, 'MyVpc', {});
-      const cluster = new ecs.Cluster(stack, 'EcsCluster', { vpc });
-      addDefaultCapacityProvider(cluster, stack, vpc);
-      const taskDefinition = new ecs.Ec2TaskDefinition(stack, 'Ec2TaskDef');
-
-      taskDefinition.addContainer('web', {
-        image: ecs.ContainerImage.fromRegistry('amazon/amazon-ecs-sample'),
-        memoryLimitMiB: 512,
-      });
-
-      new ecs.Ec2Service(stack, 'Ec2Service', {
-        cluster,
-        taskDefinition,
-        placementStrategies: [],
-      });
-
-      // THEN
-      Template.fromStack(stack).hasResourceProperties('AWS::ECS::Service', {
-        PlacementStrategies: Match.arrayEquals([]),
-      });
-
-    });
-
     test('throws an exception if non-DAEMON service is added but no EC2 capacity is associated with the cluster', () => {
       // GIVEN
       const stack = new cdk.Stack();


123MwanjeMike:
Clarification Request

pwrmiller:
**One note**: while empty PlacementConstraints will clear all placement constraints, an empty PlacementStrategies is not respected by CloudFormation/ does not clear PlacementStrategies on stack update

There are a couple of fixes and/or workarounds we could consider here:

1. Have ECS/ CloudFormation coordinate a fix for how PlacementStrategies is handled (CloudFormation does not allow null values, though it seems that PlacementStrategies needs to be set to null to clear existing strategies) - so this is a problem / bug between CloudFormation and ECS 
2. Have CDK pass a single PlacementStrategy, even if none are specified, that represents the default taken by ECS. The ECS documentation isn't clear as to what this strategy is, so I cannot offer further speculation as to the viability of this workaround.

123MwanjeMike:
> **One note**: while empty PlacementConstraints will clear all placement constraints, an empty PlacementStrategies is not respected by CloudFormation/ does not clear PlacementStrategies on stack update
> 
> There are a couple of fixes and/or workarounds we could consider here:
> 
>     1. Have ECS/ CloudFormation coordinate a fix for how PlacementStrategies is handled (CloudFormation does not allow null values, though it seems that PlacementStrategies needs to be set to null to clear existing strategies) - so this is a problem / bug between CloudFormation and ECS
> 
>     2. Have CDK pass a single PlacementStrategy, even if none are specified, that represents the default taken by ECS. The ECS documentation isn't clear as to what this strategy is, so I cannot offer further speculation as to the viability of this workaround.
Thanks @pwrmiller
We can stick to the placement constraints scope for this PR.
Let me revert the PlacementStrategy changes made

Issue #27587, opened 2023-10-17
aws-cdk-lib/aws-ec2: "Malformed instance type identifier" when creating ec2 instance of type 'm7i-flex.large'


matt-mclaughlin-quantum-si:
### Describe the bug

I'm trying to add an `m7i-flex.large` instance to a managed node group per
`createManagedComputeNodes(
        name,
        subnet,
        [
          new ec2.InstanceType('m7i-flex.large'),
        ],
        eks.CapacityType.SPOT,
      );`

But receive a `Malformed instance type identifier` when doing so.

### Expected Behavior

I'm able to add the `m7i-flex.large` instance type as an instance choice for the managed node group

### Current Behavior

```
Error: Malformed instance type identifier
    at InstanceType.sameInstanceClassAs (/node_modules/aws-cdk-lib/aws-ec2/lib/instance-types.js:1:17751)
    at /node_modules/aws-cdk-lib/aws-eks/lib/managed-nodegroup.js:1:7443
    at Array.some (<anonymous>)
    at isGpuInstanceType (/node_modules/aws-cdk-lib/aws-eks/lib/managed-nodegroup.js:1:7422)
    at typeToArch (/node_modules/aws-cdk-lib/aws-eks/lib/managed-nodegroup.js:1:8182)
    at Array.map (<anonymous>)
    at getPossibleAmiTypes (/node_modules/aws-cdk-lib/aws-eks/lib/managed-nodegroup.js:1:8442)
    at new Nodegroup (/node_modules/aws-cdk-lib/aws-eks/lib/managed-nodegroup.js:1:3543)
    at Cluster.addNodegroupCapacity (/node_modules/aws-cdk-lib/aws-eks/lib/cluster.js:1:19557)

Node.js v20.5.0
```

### Reproduction Steps

Add a `m7i-flex.large` instance type to a managed node group per "Describe the bug".  Then run `cdk synthesize`.

### CDK CLI Version

2.101.1 (build 16ddad1)

### Node.js Version

20.5.0

### OS

MAC OS Monterey 12.6.1

### Language

TypeScript

peterwoodworth:
The regex doesn't allow for hyphens, ths would need to change https://github.com/aws/aws-cdk/blob/b2a895ef285e5451e64c21e179172e998c479582/packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts#L1393

bibhu107:
I would like to contribute here!
Thanks

[...]

PR #27616, opened 2023-10-19
fix(aws-cdk-lib): fixing regex value to include hyphen in instanceType

bibhu107:
Correcting regex value to include hyphen

Closes https://github.com/aws/aws-cdk/issues/27587

----


scanlonp (changes requested):
Is there a reason instantiation of the form `ec2.InstanceType.of(ec2.InstanceClass.M7I_FLEX, ec2.InstanceSize.LARGE)` is not possible? I left the same comment on the linked issue. That looks to be the preferred syntax, and this change would not be needed.

PR #27621, opened 2023-10-20
chore(logs): remove hardcoded account id and region from snapshot

tam0ri:
In #26498, I injected specific account id and region in integration test snapshot by mistake. This PR replaces them with variables by taking snapshot correctly.

----


kaizencc (approved):
Whoopsie-daisy :) I'm surprised that our linter that looks for 12 digits didn't alert us to that. Might need further investigation.

PR #27623, opened 2023-10-20
fix(stepfunctions-tasks): cloudwatchlogs service generates wrong action in role policy

### LINKED ISSUE #27573: aws_stepfunctions_tasks : Generates wrong action in role policy
### Describe the bug

Following code is properly creates a step in the Step Function
```
  // Create log stream step  using arn:aws:states:::aws-sdk:cloudwatchlogs:createLogStream
  const createLogStream = new tasks.CallAwsService(scope, 'Create log stream', {
    service: 'cloudwatchlogs',
    action: 'createLogStream',
    parameters: {
      LogGroupName: myLogGroup.logGroupName,
      LogStreamName: sfn.JsonPath.stringAt('$$.Execution.Name'),
    },
    resultPath: sfn.JsonPath.DISCARD,
    iamResources: [myLogGroup.logGroupArn],
  });
```

However, it creates wrong statement in the role action 

```
{
 "Action": "cloudwatchlogs:createLogStream",
 "Resource": "arn:aws:logs:eu-west-1:408064982279:log-group:some_name:*",
 "Effect": "Allow"
},
```

### Expected Behavior

action should be `logs:CreateLogStream`

### Current Behavior

Action is cloudwatchlogs:createLogStream

### Reproduction Steps

Create a step function with provided code.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

v2.101.1

### Framework Version

_No response_

### Node.js Version

v18.16.0

### OS

MacOS

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

go-to-k:
This PR fixes the bug that a wrong action in role policy is generated when `cloudwatchlogs` service is specified.

A correct action is `logs:xxx`, but current behavior is to `cloudwatchlogs:xxx` by using the `service` property.

Closes #27573.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
+++ b/packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
@@ -599,6 +599,16 @@ export enum InstanceClass {
   /**
    * Storage optimized instances powered by Graviton2 processor, 4th generation
    */
+  STORAGE4_GRAVITON = 'storage4_graviton',
+
+  /**
+   * Storage optimized instances powered by Graviton2 processor, 4th generation
+   */
+  I4G = 'i4g',
+
+  /**
+   * Storage optimized instances powered by Graviton2 processor, 4th generation
+   */
   STORAGE4_GRAVITON_NETWORK_OPTIMIZED = 'storage4-graviton-network-optimized',
 
   /**
@@ -1291,6 +1301,8 @@ export class InstanceType {
       [InstanceClass.I3]: 'i3',
       [InstanceClass.IO3_DENSE_NVME_DRIVE]: 'i3en',
       [InstanceClass.I3EN]: 'i3en',
+      [InstanceClass.STORAGE4_GRAVITON]: 'i4g',
+      [InstanceClass.I4G]: 'i4g',
       [InstanceClass.STORAGE4_GRAVITON_NETWORK_OPTIMIZED]: 'im4gn',
       [InstanceClass.IM4GN]: 'im4gn',
       [InstanceClass.STORAGE4_GRAVITON_NETWORK_STORAGE_OPTIMIZED]: 'is4gen',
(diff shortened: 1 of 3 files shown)


scanlonp (changes requested):
Hey @go-to-k, this change looks great. Before approval, do you have a link to docs that state the correct action is `logs:createLogStream`? 
Also in the original issue, the user states that the action should be `logs:CreateLogStream` with a capital C on Create. Another case where the docs would be helpful in confirming the correct action is being created.

go-to-k:
@scanlonp 

Thanks for your review. The following description is in the UserGuide reference.

> The prefix and the action name are case insensitive.

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_action.html

The documentation link and the explanation are in the CDK codes.

https://github.com/aws/aws-cdk/blob/09c809b52fd2eeb27ac5bbc91d425ecf54e31bf9/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/aws-sdk/call-aws-service.ts#L92-L94

https://github.com/aws/aws-cdk/blob/09c809b52fd2eeb27ac5bbc91d425ecf54e31bf9/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/aws-sdk/call-aws-service.ts#L21-L26

But CloudWatch Logs are **insensitive** for action names, while StepFunctions SDK service integrations are case **sensitive** for API actions. The action for StepFunctions must be camel case only.

> The API action will always be camel case, and parameter names will be Pascal case. For example, you could use Step Functions API action startSyncExecution and specify its parameter as StateMachineArn.

https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html

So, if you use `logs:CreateLogStream` (Pascal case) for the `action` in CDK props, an error will occur in StepFunction definitions.

```
Deployment failed: Error: The stack named aws-stepfunctions-tasks-call-aws-service-logs-integ failed to deploy: UPDATE_ROLLBACK_COMPLETE: Resource handler returned message: "Invalid State Machine Definition: 'SCHEMA_VALIDATION_FAILED: The resource provided arn:aws:states:::aws-sdk:cloudwatchlogs:CreateLogStream is not recognized. The value is not a valid resource ARN, or the resource is not available in this region. at /States/SendTaskSuccess/Resource' (Service: AWSStepFunctions; Status Code: 400; Error Code: InvalidDefinition; Request ID: xxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx; Proxy: null)" (RequestToken: xxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx, HandlerErrorCode: InvalidRequest)
```

In this regard, I will submit an another PR that the construct translates to camel cases in StepFunction props if you use pascal cases for `action` in props.

scanlonp (approved):
Perfect, thanks for the docs links! Looks good.

scanlonp (approved):

PR #27629, opened 2023-10-20
feat(scheduler-targets): eventBridge putEvents target

### LINKED ISSUE #27454: (aws-scheduler-targets-alpha): Add EventBridgePutEvents Target
### Describe the feature

Work to support L2 constructs for AWS Scheduler is in progress (https://github.com/aws/aws-cdk/issues/23394). See the approved [RFC](https://github.com/aws/aws-cdk-rfcs/pull/486/files). RFC planned to add 12 templates targets, but only Lambda Invoke is currently implemented (https://github.com/aws/aws-cdk/pull/26575). 

This issue tracks implementation of  `EventBridgePutEvents` target to put an event to AWS Event Bridge Event Bus.

### Use Case

Customers would like to use templated target `EventBridgePutEvents` to be able to put events to Event Bus on schedule. L2 target construct should grant required permissions to the `AWS Scheduler` to put events to a Event Bus.

### Proposed Solution

The proposed solution needs to be adopted to the recent examples of `LambdaInvoke` (https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-scheduler-targets-alpha/lib/lambda-invoke.ts). 

Solution should also include unit and integration tests. 

Class `EventBridgePutEvents ` should: 
1. Grant Scheduler Execution Role permissions to Put Events via `addTargetActionToRole`
2. Override `bindBaseTargetConfig` to return `eventBridgeParameters`  as part of `ScheduleTargetConfig`.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.99.1

MacOS

WtfJoke:
An eventBridgePutEvents target was implemented similar to the already existing LambdaInvoke/StepFunctionStartExecution target.

I needed to change some properties and methods from Target.ts from private to protected so that the logic could be reused (hope that is ok).

Some design choices to outline (let me know if you disagree or have improvements I could take :) ):
1. PutEvents would accept multiple Entries (eg. an array), but I decided to support just one single event, because how Target is currently designed (to support only one target arn). It also aligns with the templated integration in the aws management console.
2. It throws an error in the constructor if the base prop `input` is used. All the props should be delivered by the new `EventBridgePutEventsEntry`. It felt not right for the developer experience to split the object (detail to `input` and `source`, `detailType` to `EventBridgePutEventsEntry` ).

Closes #27454.

----

### DIFF
--- a/packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
+++ b/packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
@@ -2,7 +2,6 @@ import { IScheduleTarget, ISchedule, ScheduleTargetInput, ScheduleTargetConfig }
 import { Names } from 'aws-cdk-lib';
 import * as events from 'aws-cdk-lib/aws-events';
 import { IRole } from 'aws-cdk-lib/aws-iam';
-import * as iam from 'aws-cdk-lib/aws-iam';
 import { ScheduleTargetBase, ScheduleTargetBaseProps } from './target';
 import { sameEnvDimension } from './util';
 
@@ -56,7 +55,7 @@ export class EventBridgePutEvents extends ScheduleTargetBase implements ISchedul
     private readonly props: ScheduleTargetBaseProps,
   ) {
     super(props, entry.eventBus.eventBusArn);
-    if (this.baseProps.input) {
+    if (this.props.input) {
       throw new Error('ScheduleTargetBaseProps.input is not supported for EventBridgePutEvents. Please use entry.detail instead.');
     }
   }
@@ -80,26 +79,12 @@ export class EventBridgePutEvents extends ScheduleTargetBase implements ISchedul
   }
 
   protected bindBaseTargetConfig(_schedule: ISchedule): ScheduleTargetConfig {
-    const role: iam.IRole = this.baseProps.role ?? this.singletonScheduleRole(_schedule, this.targetArn);
-    this.addTargetActionToRole(_schedule, role);
-
-    if (this.baseProps.deadLetterQueue) {
-      this.addToDeadLetterQueueResourcePolicy(_schedule, this.baseProps.deadLetterQueue);
-    }
-
-    const { source, detailType, detail } = this.entry;
-
     return {
-      arn: this.targetArn,
-      role,
-      deadLetterConfig: this.baseProps.deadLetterQueue ? {
-        arn: this.baseProps.deadLetterQueue.queueArn,
-      } : undefined,
-      retryPolicy: this.renderRetryPolicy(this.baseProps.maxEventAge, this.baseProps.retryAttempts),
-      input: detail,
+      ...super.bindBaseTargetConfig(_schedule),
+      input: this.entry.detail,
       eventBridgeParameters: {
-        detailType,
-        source,
+        detailType: this.entry.detailType,
+        source: this.entry.source,
       },
     };
   }
--- a/packages/@aws-cdk/aws-scheduler-targets-alpha/lib/target.ts
+++ b/packages/@aws-cdk/aws-scheduler-targets-alpha/lib/target.ts
@@ -68,7 +68,7 @@ export interface ScheduleTargetBaseProps {
 export abstract class ScheduleTargetBase {
 
   constructor(
-    protected readonly baseProps: ScheduleTargetBaseProps,
+    private readonly baseProps: ScheduleTargetBaseProps,
     protected readonly targetArn: string,
   ) {
   }
@@ -97,7 +97,7 @@ export abstract class ScheduleTargetBase {
   /**
    * Create a return a Schedule Target Configuration for the given schedule
    * @param schedule
-   * @returnn
+   * @returns a Schedule Target Configuration
    */
   bind(schedule: ISchedule): ScheduleTargetConfig {
     return this.bindBaseTargetConfig(schedule);
@@ -109,7 +109,7 @@ export abstract class ScheduleTargetBase {
    * If a role already exists, it will be returned. This ensures that if multiple
    * events have the same target, they will share a role.
    */
-  protected singletonScheduleRole(schedule: ISchedule, targetArn: string): iam.IRole {
+  private singletonScheduleRole(schedule: ISchedule, targetArn: string): iam.IRole {
     const stack = Stack.of(schedule);
     const arn = Token.isUnresolved(targetArn) ? stack.resolve(targetArn).toString() : targetArn;
     const hash = md5hash(arn).slice(0, 6);
@@ -141,7 +141,7 @@ export abstract class ScheduleTargetBase {
    * @param schedule schedule to add DLQ to
    * @param queue the DLQ
    */
-  protected addToDeadLetterQueueResourcePolicy(schedule: ISchedule, queue: sqs.IQueue) {
+  private addToDeadLetterQueueResourcePolicy(schedule: ISchedule, queue: sqs.IQueue) {
     if (!sameEnvDimension(schedule.env.region, queue.env.region)) {
       throw new Error(`Cannot assign Dead Letter Queue in region ${queue.env.region} to the schedule ${Names.nodeUniqueId(schedule.node)} in region ${schedule.env.region}. Both the queue and the schedule must be in the same region.`);
     }
@@ -164,7 +164,7 @@ export abstract class ScheduleTargetBase {
     }
   }
 
-  protected renderRetryPolicy(maximumEventAge: Duration | undefined, maximumRetryAttempts: number | undefined): CfnSchedule.RetryPolicyProperty {
+  private renderRetryPolicy(maximumEventAge: Duration | undefined, maximumRetryAttempts: number | undefined): CfnSchedule.RetryPolicyProperty {
     const maxMaxAge = Duration.days(1).toSeconds();
     const minMaxAge = Duration.minutes(15).toSeconds();
     let maxAge: number = maxMaxAge;


packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
```diff
@@ -0,0 +1,107 @@
...
+   * @see https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events.html
+   */
+  readonly source: string;
+}
+
+/**
+ * Send an event to an AWS EventBridge by AWS EventBridge Scheduler.
+ */
+export class EventBridgePutEvents extends ScheduleTargetBase implements IScheduleTarget {
+
```
kaizencc, line 54:
```suggestion
```

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
```diff
@@ -0,0 +1,107 @@
...
+  constructor(
+    private readonly entry: EventBridgePutEventsEntry,
+    private readonly props: ScheduleTargetBaseProps,
+  ) {
+    super(props, entry.eventBus.eventBusArn);
+    if (this.baseProps.input) {
+      throw new Error('ScheduleTargetBaseProps.input is not supported for EventBridgePutEvents. Please use entry.detail instead.');
+    }
```
kaizencc, line 62: I don't quite understand this even with your pr description. `detail` and `input` seem to both have the same type. Why do we accept one but not the other?

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    if (this.props.role && !sameEnvDimension(this.props.role.env.account, eventBusEnv.account)) {
+      throw new Error(`Cannot grant permission to execution role in account ${this.props.role.env.account} to invoke target ${Names.nodeUniqueId(eventBus.node)} in account ${eventBusEnv.account}. Both the target and the execution role must be in the same account.`);
+    }
+
+    eventBus.grantPutEventsTo(role);
+  }
+
+  protected bindBaseTargetConfig(_schedule: ISchedule): ScheduleTargetConfig {
```
kaizencc, line 83: I don't like the duplication here. let's figure out how to modify `super.bindBaseTargetConfig()` so that it can be called via this method and we can reuse as much as we can

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Send an event to an AWS EventBridge by AWS EventBridge Scheduler.
+ */
+export class EventBridgePutEvents extends ScheduleTargetBase implements IScheduleTarget {
+
+  constructor(
+    private readonly entry: EventBridgePutEventsEntry,
+    private readonly props: ScheduleTargetBaseProps,
+  ) {
+    super(props, entry.eventBus.eventBusArn);
+    if (this.baseProps.input) {
```
kaizencc, line 60: this can also be `this.props.input`, i don't think we need to expose `baseProps` at all

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/target.ts
```diff
@@ -68,7 +68,7 @@ export interface ScheduleTargetBaseProps {
 export abstract class ScheduleTargetBase {
 
   constructor(
-    private readonly baseProps: ScheduleTargetBaseProps,
+    protected readonly baseProps: ScheduleTargetBaseProps,
```
kaizencc, line 71: after some of these changes i suspect we can revert this line and the rest of the changes in this file

kaizencc (changes requested):
Hi @WtfJoke! Thanks for another scheduler target! A few comments here and there, but I appreciate your dedication to this module :)

packages/@aws-cdk/aws-scheduler-targets-alpha/README.md
```diff
@@ -102,3 +102,30 @@ new Schedule(this, 'Schedule', {
...
+  source: 'PetService',
+  detail: ScheduleTargetInput.fromObject({ Name: 'Fluffy' }),
+  detailType: '🐶',
+};
+
+new Schedule(this, 'Schedule', {
+  schedule: ScheduleExpression.rate(Duration.hours(1)),
+  target: new targets.EventBridgePutEvents(eventEntry, {}),
+});
+```
```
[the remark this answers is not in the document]
WtfJoke: StepFunction was already there. I've added eventbridge

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
WtfJoke: I've introduced a new method overrideTargetConfig which tries to minimize the duplication here. Im open for better ideas, in case it doesnt satisfy your needs :)

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
WtfJoke: Yeah, did that

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/target.ts
WtfJoke: Reverted the visibility changes back with the introduction of the method `overrideTargetConfig`

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
WtfJoke: Theoretically we could accept both (`input` and `detail`). However if somebody is supplying both, which should we pick? 
Then you might think that we can make both optional and throw an exception if both are supplied, but I dont see any advantages going that route.

We could also omit `detail` from `EventBridgePutEventsEntry`, but I think that might be confusing for users, who know the structure from the [PutEvents API Call](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEvents.html), eg. `detail`. And I think its a bit unintuitive to put most of the payload in one object and the main part (e.g detail in input). So this was my reasoning for the current implementation.

Feel free to question me further or share me how you think about this. :)
I hope I was able to describe it good enough to understand.

WtfJoke:
Hey @kaizencc 
Thanks for the review! Appreciate the feedback. I've addressed most of your feedback, please have a look at it again.

Yeah I initially needed StepFunction for a Demo/Talk, so I was keen in contributing this back (after hacking it together on my demo repo) and figured its quiet straight forward to add more targets (Im thinking about adding the Universal target as well, but I could imagine this will be a bit harder) :D

WtfJoke:
@kaizencc I've fixed the merge conflicts and rebased my changes. Please have a look at it.

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
```diff
@@ -0,0 +1,91 @@
...
+ * An entry to be sent to EventBridge
+ *
+ * @see https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEventsRequestEntry.html
+ */
+export interface EventBridgePutEventsEntry {
+  /**
+   * The event body
+   *
+   * Can either be provided as an object or as a JSON-serialized string
+   * @example
```
kaizencc, line 18: with `@example`, eventally `jsii-rosetta` (a tool we have that aggregates all examples that show up in our documentation) will want this to compile. Unless you want to figure out how to get this to compile, I suggest the following:

```suggestion
   * For example,
   *
   * ```
   * // your uncompiled example code here
   * ```
```

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
```diff
@@ -0,0 +1,91 @@
...
+   */
+  readonly detail: ScheduleTargetInput;
+
+  /**
+   * Used along with the source field to help identify the fields and values expected in the detail field
+   *
+   * For example, events by CloudTrail have detail type "AWS API Call via CloudTrail"
+   * @see https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events.html
+   */
+  readonly detailType: string;
```
kaizencc, line 31: I think we can type this better than a `string`. From the docs you've provided, the only thing that I see as a possible input here is `"AWS API Call via CloudTrail"`. Lol. So here is what I propose:

```ts
readonly detailType: DetailType;

////


export class DetailType {
  public static API_CALL_VIA_CLOUDTRAIL() {
    return new DetailType('AWS API Call via CloudTrail');
  }

  public constructor (public readonly type: string) {}
}
```

This way, we can

a) give users an easy API to hit rather than getting the specific string correct (`DetailType.API_CALL_VIA_CLOUDTRAIL`),
b) provide space for future additions if we know of other valid detail types as static methods,
c) users can still specify whatever they want via `new DetailType('my special type')`

kaizencc (dismissed):
Hi @WtfJoke, thanks for your continued patience here. I have a few more comments, but this is looking good!

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
WtfJoke: The build is green as far as I know. Or which build (or command) do you refer to?

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+export class EventBridgePutEvents extends ScheduleTargetBase implements IScheduleTarget {
+  constructor(
+    private readonly entry: EventBridgePutEventsEntry,
+    private readonly props: ScheduleTargetBaseProps,
+  ) {
+    super(props, entry.eventBus.eventBusArn);
+    if (this.props.input) {
+      throw new Error('ScheduleTargetBaseProps.input is not supported for EventBridgePutEvents. Please use entry.detail instead.');
+    }
```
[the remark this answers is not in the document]
WtfJoke: I got you. However currently its not possible to not set entry.Detail as its required. Do you suggest weakening the type of EventBridgePutEventsEntry.detail so that we can pass null there? And then add this check? It seems to be not ideal.

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
WtfJoke: Will do.

There seems to be [more](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-service-event-list.html), but not documented which `detail-type` they emit.
I found only one additional example in the docs [here](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events-structure.html) `"detail-type": "EC2 Instance State-change Notification"`

Cant speak for others but our team uses only own detail types (e.g your point c).

For context:
I followed the existing [EventBridgePutEvents](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions_tasks.EventBridgePutEvents.html) implementation of the step function tasks package, which uses string as types.

WtfJoke: Done :)

WtfJoke:
@kaizencc Great to hear. Looking forward to getting this merged.

Would be glad if you can answer my questions, as im stuck in 2/3 review comments.

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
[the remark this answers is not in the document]
kaizencc: Ohhh ok got you. This is okay then.

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
kaizencc: Andddd I am now wavering on this suggestion. I still think it could be a good idea, but might be overengineered. I didn't see the nuance at first that most `detail-types` are user-owned, and then for events from AWS there maybe detail types with specific strings.

I think we should go back to using `string` here.

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
kaizencc: It does compile :). Nevermind.

packages/@aws-cdk/aws-scheduler-targets-alpha/lib/event-bridge-put-events.ts
WtfJoke: I just dropped my last commit, which effectively reverts the change.

WtfJoke:
@kaizencc I just dropped my last commit. Also rebased my changes on to latest main and fixed all the conflicts.

Please have a look :)

kaizencc (approved):
Thanks @WtfJoke and sorry for the many rounds of fruitless reviews :)

PR #27632, opened 2023-10-20
docs: update INTEGRATION_TESTS.md

scanlonp:
This updates our docs surrounding integ tests:

- gives the correct location of stable module integ tests
- standardizes code examples
- adds example for using `assertAtPath()`
- warns against using `outputPaths` in `awsApiCall()`
- other small updates

----


INTEGRATION_TESTS.md
```diff
@@ -262,12 +259,32 @@ const describe = testCase.assertions.awsApiCall('StepFunctions', 'describeExecut
...
 
 // assert the results
-describe.expect(ExpectedResult.objectLike({
+describe.expect(integ.ExpectedResult.objectLike({
   status: 'SUCCEEDED',
 }));
 ```
 
-Not every test requires an assertion. We typically do not need to assert CloudFormation behavior. For example, if we create an S3 Bucket
+If we want to pick out certain values from the api call response, we can use the `assertAtPath()` method, as in the [integ.pipeline-with-additional-inputs.ts](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/framework-integ/test/pipelines/test/integ.pipeline-with-additional-inputs.ts) integ test. Note that using the `outputPaths` optional parameter on the `awsApiCall()` function often interacts poorly with the `expect()` function.
```
vinayak-kukreja, line 267: > Note that using the `outputPaths` optional parameter on the `awsApiCall()` function often interacts poorly with the `expect()` function.

Why is this the case?

INTEGRATION_TESTS.md
```diff
@@ -262,12 +259,32 @@ const describe = testCase.assertions.awsApiCall('StepFunctions', 'describeExecut
...
 
-Not every test requires an assertion. We typically do not need to assert CloudFormation behavior. For example, if we create an S3 Bucket
+If we want to pick out certain values from the api call response, we can use the `assertAtPath()` method, as in the [integ.pipeline-with-additional-inputs.ts](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/framework-integ/test/pipelines/test/integ.pipeline-with-additional-inputs.ts) integ test. Note that using the `outputPaths` optional parameter on the `awsApiCall()` function often interacts poorly with the `expect()` function. 
+
+```ts
+declare const pipelineName: string;
+declare const expectedString: string;
+
+const testCase = new integ.IntegTest(app, 'PipelineAdditionalInputsTest', {
+  testCases: [stack],
```
vinayak-kukreja, line 274: We can add, 
```
declare const stack: Stack;
```
and 
```
```

INTEGRATION_TESTS.md
```diff
@@ -262,12 +259,32 @@ const describe = testCase.assertions.awsApiCall('StepFunctions', 'describeExecut
...
+
+const source = testCase.assertions.awsApiCall('CodePipeline', 'GetPipeline', {
+  name: pipelineName,
+});
+
+// assert the value at the given path matches the expected string
+// the numbers index arryas in the json response object
+source.assertAtPath('pipeline.stages.0.actions.0.name', integ.ExpectedResult.stringLikeRegexp(expectedString));
+```
+A helpful trick is to deploy the integ test with `--no-clean` and then make the api call locally. We can then trace the path to specific values easily.
```
vinayak-kukreja, line 285: > A helpful trick is to deploy the integ test with `--no-clean` and then make the api call locally.

Can you add an example of this? For instance, how can they make the api call locally? Do they need to setup something, etc?

INTEGRATION_TESTS.md
```diff
@@ -262,12 +259,32 @@ const describe = testCase.assertions.awsApiCall('StepFunctions', 'describeExecut
...
+
+Adding assertions is prefered on all new integ tests; however, it is not strictly required. We typically do not need to assert CloudFormation behavior. For example, if we create an S3 Bucket
 with Encryption, we do not need to assert that Encryption is set on the bucket. We can trust that the CloudFormation behavior works.
 Some things you should look for in deciding if the test needs an assertion:
```
vinayak-kukreja, line 289: > Some things you should look for in deciding if the test needs an assertion:

I think we can add `custom resources` to this list. WDYT?

vinayak-kukreja (changes requested):
Hey, thank you for improving our docs. 🙌 

I have some suggestions, let me know if I can clarify anything.

INTEGRATION_TESTS.md
scanlonp: No idea, but it did not work for me, and I have heard from others that it has not worked smoothly for them.

INTEGRATION_TESTS.md
scanlonp: I will see what I can do. Not sure the mechanics of making api calls with the aws cli is our main focus, but can give a small code example.

INTEGRATION_TESTS.md
scanlonp: Yes.

INTEGRATION_TESTS.md
vinayak-kukreja: Ok, it may be better to deprecate this if it is inconsistent. But not blocking this PR :)

vinayak-kukreja (approved):

PR #27633, opened 2023-10-21
feat(assertions): add stack tagging assertions

### LINKED ISSUE #27620: (aws-cdk-lib/assertions): Simplify assertions on stack tags
### Describe the feature

Tags associated with a Cloudformation stack are recorded in the cloud assembly manifest. The existing `Template` and `Annotations` classes in `aws-cdk-lib/assertions` allow for simplified testing of some parts of the cloud assembly, but not the `tags` member.

It is possible to interrogate the tags of a Resource, but I'm unaware of how to do that for the Stack tags.

### Use Case

We have constructs libraries that ensure our stacks are tagged appropriately for a project. Currently tests for this behaviour are possible but not straightforward.

### Proposed Solution

I suggest additions the API such that it would be possible to write assertions similar to:

```ts
// Tags aren't technically part of the template, though this kinda makes sense
Template.fromStack(foo).hasTags({
  "tag-name":"tag-value",
  "other-tag": Matcher.absent()
})

// OR - this is targeted carefully at this case
Tags.fromStack(foo).hasTags(Matcher.anyValue())

// OR - more generally, could be use for other manifest properties but I don't know what they would be.
StackProperties.fromStack(foo).hasTags(Matcher.anyValue())
```

### Other Information

It is possible to write these tests outside of the CDK assertions library by interrogating the Cloud Assembly API. 4

Something like the following works:

```ts
  const app = new TestApp()
  const stack = new TestStack(app, "name")

  const stage = Stage.of(stack)
  if (!Stage.isStage(stage)) {
    throw new Error("unexpected: all stacks must be part of a Stage or an App")
  }

  const assembly = stage.synth()
  const tags = assembly.getStackArtifact(stack.artifactId).tags

  expect(tags).toMatchObject({
    "aws:cdk:path": "/name",
    "aws:cdk:logicalId": "name",
  })
```

There are some special cases that occur when looking up details in the assembly in both `Template` and `Annotations` classes. Respectively, handling nested stacks and forcing synthesis. I'm not sure which of these special cases should be handled in the tag case.

I've started on a branch to feel out what an implementation would look like. If the `Tags` API is selected, the PR would be quite small by the looks.

### Acknowledgements

### CDK version used

2.102.0

N/A

jamestelfer:
Adds a `Tag` class to the assertions library that permits assertions against tags on synthesized CDK stacks.

Tags on AWS resources can be checked via assertions with the Template class, but since stack tags only appear on the cloud assembly manifest as a sibling of the template, a separate assertion mechanism is required.

### API

```ts
class Tags {
  public static fromStack(stack: Stack) : Tags;
  public hasValues(props: any): void;
  public all(): { [key: string]: string };
}
```

### Usage

This new class permits tests of the form:

```ts
import { App, Stack } from 'aws-cdk-lib';
import { Tags } from 'aws-cdk-lib/assertions';

const app = new App();
const stack = new Stack(app, 'MyStack', {
  tags: {
    'tag-name': 'tag-value'
  }
});

const tags = Tags.fromStack(stack);

// using a default 'objectLike' Matcher
tags.hasValues({
  'tag-name': 'tag-value'
});

// or another Matcher
tags.hasValues({
  'tag-name': Match.anyValue()
});
```

You can also get the set of tags to test them in other ways:

```ts
tags.all()
```

### No tags case

One might expect that the case where no tags are present would match `undefined` or `null`, but since the Cloud Assembly API defaults tags to `{}` when none are present, this isn't possible. It's also not practical to directly test the `artifact.manifest.properties.tags` value directly, as there is a legacy case that the API handles. This means that the `artifact.tags` property is the best value to check against.

The tests for this PR show that matching with `Match.absent()` will fail when there are no tags, but testing against the empty object will succeed.

I think that this behaviour (defaulting to empty) will be OK, but potentially require a callout on the assertion method.

### API method naming

The current suggested API went through some evolution, starting with:

```ts
class Tags {
  public static fromStack(stack: Stack) : Tags;
  public hasTags(props: any): void;
  public getTags():  { [key: string]: string };
}
```
But this stuttered, and `getTags()` wasn't compatible with Java.

I considered:

```ts
class Tags {
  public static fromStack(stack: Stack) : Tags;
  public hasValues(props: any): void;
  public values():  { [key: string]: string };
}
```
and
```ts
class Tags {
  public static fromStack(stack: Stack) : Tags;
  public has(props: any): void;
  public all():  { [key: string]: string };
}
```

... before settling on a mix of the two. I think the current iteration fits with the rest of the `assertions` API and makes sense by itself, but very open to changes.

Closes #27620.

----

### DIFF
--- a/packages/aws-cdk-lib/assertions/README.md
+++ b/packages/aws-cdk-lib/assertions/README.md
@@ -595,3 +595,37 @@ Annotations.fromStack(stack).hasError(
   Match.stringLikeRegexp('.*Foo::Bar.*'),
 );
 ```
+
+## Asserting Stack tags
+
+Tags applied to a `Stack` are not part of the rendered template: instead, they
+are included as properties in the Cloud Assembly Manifest. To test that stacks
+are tagged as expected, simple assertions can be written.
+
+Given the following setup:
+
+```ts nofixture
+import { App, Stack } from 'aws-cdk-lib';
+import { Tags } from 'aws-cdk-lib/assertions';
+
+const app = new App();
+const stack = new Stack(app, 'MyStack', {
+  tags: {
+    'tag-name': 'tag-value'
+  }
+});
+```
+
+It is possible to test against these values:
+
+```ts
+// using a default 'objectLike' Matcher
+Tags.fromStack(stack).hasValues({
+  'tag-name': 'tag-value'
+});
+
+// or another Matcher
+Tags.fromStack(stack).hasValues({
+  'tag-name': Match.anyValue()
+});
+```
--- a/packages/aws-cdk-lib/assertions/lib/index.ts
+++ b/packages/aws-cdk-lib/assertions/lib/index.ts
@@ -3,3 +3,4 @@ export * from './template';
 export * from './match';
 export * from './matcher';
 export * from './annotations';
+export * from './tags';
--- a/packages/aws-cdk-lib/assertions/lib/tags.ts
+++ b/packages/aws-cdk-lib/assertions/lib/tags.ts
@@ -1,0 +1,69 @@
+import { Match } from './match';
+import { Matcher } from './matcher';
+import { Stack, Stage } from '../../core';
+
+type ManifestTags = { [key: string]: string };
+
+/**
+ * Allows assertions on the tags associated with a synthesized CDK stack's
+ * manifest. Stack tags are not part of the synthesized template, so can only be
+ * checked from the manifest in this manner.
+ */
+export class Tags {
+  /**
+   * Find tags associated with a synthesized CDK `Stack`.
+   *
+   * @param stack the CDK Stack to find tags on.
+   */
+  public static fromStack(stack: Stack): Tags {
+    return new Tags(getManifestTags(stack));
+  }
+
+  private readonly _tags: ManifestTags;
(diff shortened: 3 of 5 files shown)


jamestelfer:
Exemption Request

As far as I can tell, the assertions library does not have integration tests, but I'm happy to be corrected.

jamestelfer:
@SankyRed thanks for picking this up. 

As far as I know there's nothing more I can do before review: tests pass and I don't think integration tests are required (see the exemption request).

Have I missed something that would allow a review?

jamestelfer:
@SankyRed / @pahud while the automation is pointing to this PR being stale, to the best of my knowlegde this PR is only awaiting (a) a response to the exemption request and (b) a code review.

Would it be possible for someone to indicate what further action I need to take?

jamestelfer:
I've rebased this PR on the latest `main`.

jamestelfer:
@SankyRed / @pahud I'm not sure why this was closed automatically: it was sitting awaiting review. As far as I knew, there was nothing more I could do, and I didn't want to ping constantly as that seemed rude. 

What's the expected etiquette in this repo? Apologies, I'm new here!

PR #27635, opened 2023-10-21
chore(stepfunctions-tasks): validate cases of action and parameters in CallAwsService

go-to-k:
This PR adds the following validations in CallAwsService.

- `action` must be camelCase.
- parameter names in `parameters` must be PascalCase.

See the doc: https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html

> The API action will always be camel case, and parameter names will be Pascal case. For example, you could use Step Functions API action startSyncExecution and specify its parameter as StateMachineArn.

CloudFormation fails with a following error if there are not these validations.

```
Deployment failed: Error: The stack named aws-stepfunctions-tasks-call-aws-service-logs-integ failed to deploy: UPDATE_ROLLBACK_COMPLETE: Resource handler returned message: "Invalid State Machine Definition: 'SCHEMA_VALIDATION_FAILED: The resource provided arn:aws:states:::aws-sdk:cloudwatchlogs:CreateLogStream is not recognized. The value is not a valid resource ARN, or the resource is not available in this region. at /States/SendTaskSuccess/Resource' (Service: AWSStepFunctions; Status Code: 400; Error Code: InvalidDefinition; Request ID: xxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx; Proxy: null)" (RequestToken: xxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx, HandlerErrorCode: InvalidRequest)
```

I think it is a good thing to make these errors in the synth phase, since there were actually cases of confusion as follows.

https://github.com/aws/aws-cdk/pull/27623#issuecomment-1773761113

I also thought to not validate but translate to camel (or pascal) cases. However I thought it would allow input that violates the explanation defined in the API documentation, so I decided not to.

On the other hands, the `action` is also used for IAM actions so the IAM actions will be to camel cases (like `logs:createLogStream`). But I allowed it because IAM actions are case insensitive. If a translation is a better way to do it rather than the validation, I will consider that as well.

https://github.com/aws/aws-cdk/blob/09c809b52fd2eeb27ac5bbc91d425ecf54e31bf9/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/aws-sdk/call-aws-service.ts#L92-L94

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
+++ b/packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
@@ -599,6 +599,16 @@ export enum InstanceClass {
   /**
    * Storage optimized instances powered by Graviton2 processor, 4th generation
    */
+  STORAGE4_GRAVITON = 'storage4_graviton',
+
+  /**
+   * Storage optimized instances powered by Graviton2 processor, 4th generation
+   */
+  I4G = 'i4g',
+
+  /**
+   * Storage optimized instances powered by Graviton2 processor, 4th generation
+   */
   STORAGE4_GRAVITON_NETWORK_OPTIMIZED = 'storage4-graviton-network-optimized',
 
   /**
@@ -1291,6 +1301,8 @@ export class InstanceType {
       [InstanceClass.I3]: 'i3',
       [InstanceClass.IO3_DENSE_NVME_DRIVE]: 'i3en',
       [InstanceClass.I3EN]: 'i3en',
+      [InstanceClass.STORAGE4_GRAVITON]: 'i4g',
+      [InstanceClass.I4G]: 'i4g',
       [InstanceClass.STORAGE4_GRAVITON_NETWORK_OPTIMIZED]: 'im4gn',
       [InstanceClass.IM4GN]: 'im4gn',
       [InstanceClass.STORAGE4_GRAVITON_NETWORK_STORAGE_OPTIMIZED]: 'is4gen',


packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/aws-sdk/call-aws-service.ts
```diff
@@ -79,6 +79,15 @@ export class CallAwsService extends sfn.TaskStateBase {
     if (this.props.integrationPattern === sfn.IntegrationPattern.RUN_JOB) {
       throw new Error('The RUN_JOB integration pattern is not supported for CallAwsService');
     }
+    if (!Token.isUnresolved(this.props.action) && !this.props.action.startsWith(this.props.action[0]?.toLowerCase())) {
+      throw new Error(`action must be CamelCase, got: ${this.props.action}`);
```
scanlonp, line 83:
```suggestion
      throw new Error(`action must be camelCase, got: ${this.props.action}`);
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/aws-sdk/call-aws-service.test.ts
```diff
@@ -161,6 +161,30 @@ test('throws with invalid integration pattern', () => {
   })).toThrow(/The RUN_JOB integration pattern is not supported for CallAwsService/);
 });
 
+test('throws if action is not CamelCase', () => {
```
scanlonp, line 164:
```suggestion
test('throws if action is not camelCase', () => {
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/aws-sdk/call-aws-service.test.ts
```diff
@@ -161,6 +161,30 @@ test('throws with invalid integration pattern', () => {
...
+  expect(() => new tasks.CallAwsService(stack, 'GetObject', {
+    service: 's3',
+    action: 'GetObject',
+    parameters: {
+      Bucket: 'my-bucket',
+      Key: sfn.JsonPath.stringAt('$.key'),
+    },
+    iamResources: ['*'],
+  })).toThrow(/action must be CamelCase, got: GetObject/);
```
scanlonp, line 173:
```suggestion
  })).toThrow(/action must be camelCase, got: GetObject/);
```

scanlonp (changes requested):
This looks great. Validation is the way to go. Excellent error messages! Just one small change; let's write camelCase in camelCase!

scanlonp (approved):
Thank you!

PR #27645, opened 2023-10-23
feat(apigateway): add useDefaultMethodResponses property for StepFunctionsIntegrations

### LINKED ISSUE #27520: apigateway: StepFunctionsIntegrations add default methodResponses
### Describe the feature

When StepFunctionsIntegrations is used, it creates methodResponses for 200, 400 and 500 and it's not possible to override this behaviour.

### Use Case

When custom responses are provided they may clash with default ones.
For example one may want setup a different responseTemplate from the one provided.
For example I have a state machine that returns an URL and my responseTemplate set the http status to 302.

### Proposed Solution

Adding a `useDefaultMethodResponses?: boolean` prop (default true) to the `StepFunctionsExecutionIntegrationOptions` interface to avoid adding default responses:

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-apigateway/lib/integrations/stepfunctions.ts#L160

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.100

MacOs Sonoma

go-to-k:
In `StepFunctionsIntegration`, method response models (`methodResponses`) with 200, 400 and 500 status codes are added by default.

This PR adds `useDefaultMethodResponses` property not to add default method response models by set it to `false`.

Closes #27520.

----


lpizzinidev (changes requested):
Nice 👍
Just some documentation and test minor-changes needed in my opinion.

go-to-k:
@lpizzinidev 

Thanks, I changed. Please review them again.

packages/aws-cdk-lib/aws-apigateway/test/stepfunctions-api.test.ts
```diff
@@ -157,6 +157,75 @@ describe('Step Functions api', () => {
...
+              },
+              ':apigateway:',
+              {
+                Ref: 'AWS::Region',
+              },
+              ':states:action/StartSyncExecution',
+            ],
+          ],
+        },
+        PassthroughBehavior: 'NEVER',
```
lpizzinidev, line 224:
```suggestion
```
Let's match only on `HttpMethod` and `MethodResponses` to keep this concise.

lpizzinidev (changes requested):
Thanks, just a minor adjustment and it will be good to go for me.

go-to-k:
Thanks, I changed it.

scanlonp (dismissed):
Any reason to not have an integ test for the `StepFunctionsRestApi` invocation? Otherwise looks good to go.

go-to-k:
@scanlonp 

I was satisfied with just the `StepFunctionsIntegration.startExecution` test. I added that integ test.

I did not want to interfere with the behavior of the existing StepFunctionsRestApi test with `useDefaultMethodResponses` is true as a default value, so I added a new test.

scanlonp (approved):
Great, thanks for the new integ test. The optional parameter is a good way to disable automatic behavior, and the tests cover both the template validation and deploy success.

PR #27649, opened 2023-10-23
feat(dynamodb): add tagging support to TableV2

### LINKED ISSUE #27146: (dynamodb): Dynamo TableV2 does not support tagging replicas
### Describe the feature

I would like to be able to supply tags to the primary and secondary replicas within the TableV2 GlobalTable construct.

### Use Case

I need to be able to tag specific replicas in order to allow those replicas to be detected by other applications.

### Proposed Solution

Tags already exist on the CfnGlobalTable.ReplicaSpecificationProperty type so we simply need to allow the property to be passed to either the TablePropsV2 (for the primary replica) or ReplicaTableProps (for the secondary replicas). Both of these interfaces extend the TableOptionsV2 interface so I would propose the update be added there. Additionally there are some small changes needed to use the tags when configuring the promary replica and  when performing the configureReplicaTable method. All these changes are outlined in the attached screenshot of the proposed diff.

Proposed Changes

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.96.0

Node 20.6.1, Mac

mikevec23:
This change adds the ability to tag GlobalTable replicas created using the TableV2 construct. The top level "tags" will apply tags to the primary table while tags for each replica can be specified in the respective replica definition.

Closes #27146.

----


colifran (commented):
@mikevec23 I think this looks great! My only concern is with using `CfnTag` vs. `Tags.of(...).add(...)`. I recognize that the underlying L1 doesn't implement the `ITaggable` interface which would prevent the use of `Tags`, but I'd like to get another opinion on whether this is the right approach or if we should find a way to do this using `Tags`. I'll get an update for you soon.

colifran (approved):
Thanks for your contribution @mikevec23 - this looks good!

PR #27652, opened 2023-10-23
chore: fix the use of mixed dep/devdeps

rix0rrr:
`fs-extra` is sometimes not included in the `npm pack` tarball, depending on the version of NPM.

The reason is that it occurs in both `dependencies` and `devDependencies`, which is not a useful dependency specification.

This is breaking jsii builds, depending on the NPM version used.

----


MrArnoldPalmer (approved):

packages/aws-cdk-lib/package.json
```diff
@@ -108,7 +108,7 @@
     "organization": true
   },
   "license": "Apache-2.0",
-  "bundledDependencies": [
+  "bundleDependencies": [
```
mrgrain, line 111: 😂

PR #27681, opened 2023-10-25
chore(spec2cdk): generate ICfnResource interface

kaizencc:
Generates something that looks like this for all resources:

```ts
interface ICfnResource {
  readonly attrName: string;    // { Ref }, i.e. the primaryIdentifier
  readonly attrArn: string;     // { Fn::GetAtt }, if applicable
}
```

In a separate PR, we will update all `CfnResource`s to extend `ICfnResource`, and then later on, update some `Resource`s to extend `ICfnResource` as well.

**This PR targets `conroy/generate` as a horribly named branch off of `main` which will house all pieces of this puzzle before unleashing onto `main`.**

----

### DIFF
--- a/packages/awslint/lib/rules/core-types.ts
+++ b/packages/awslint/lib/rules/core-types.ts
@@ -90,6 +90,7 @@ export class CoreTypes {
    */
   public static isCfnType(interfaceType: reflect.Type) {
     return interfaceType.name.startsWith('Cfn')
+      || interfaceType.name.startsWith('ICfn')
       || (interfaceType.namespace && interfaceType.namespace.startsWith('Cfn'))
       // aws_service.CfnTheResource.SubType
       || (interfaceType.namespace && interfaceType.namespace.split('.', 2).at(1)?.startsWith('Cfn'));


tools/@aws-cdk/spec2cdk/test/services.test.ts
```diff
@@ -25,3 +25,106 @@ test('can codegen service with arbitrary suffix', () => {
...
+    '   * Catalog id string, not less than 1 or more than 255 bytes long, matching the [single-line string pattern](https://docs.aws.amazon.com/lake-formation/latest/dg/aws-lake-formation-api-aws-lake-formation-api-common.html) .',
+    '   *',
+    '   * The ID of the catalog to which the table belongs.',
+    '   *',
+    '   * @cloudformationRef tableCatalogId',
+    '   *',
+    '   * @see http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lakeformation-datacellsfilter.html#cfn-lakeformation-datacellsfilter-tablecatalogid',
+    '   */',
+    '  readonly attrTableCatalogId: string;',
+    // '  /**',
```
kaizencc, line 104: Inexplicably this line fails the comparison even though it definitely is exactly right...

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
```diff
@@ -37,12 +38,47 @@ export class ResourceDecider {
...
+      if (att) {
+        this.primaryIdentifier.push(att);
+      } else if (prop) {
+        // rename the prop name as an attribute name, since it is gettable by ref
+        this.primaryIdentifier.push({
+          ...prop,
+          name: attributePropertyName(prop.name[0].toUpperCase() + prop.name.slice(1)),
+          docs: {
+            ...prop.docs,
+            remarks: prop.docs?.remarks?.concat(['\n', `@cloudformationRef ${prop.name}`].join('\n')),
```
kaizencc, line 61: not sure what remark we want to add here... does `@cloudformationRef` make sense? for attributes, its `@cloudformationAttribute`. And do we want to do the same for primaryIdentifiers found in attributes? I think yes, but wanna check

tools/@aws-cdk/spec2cdk/test/services.test.ts
kaizencc, line 1: looking for a better way to test these without using full on snapshots. How about i store each code block as a `.txt` file and then load it into the test?

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
mrgrain: Pretty sure `@cloudformationAttribute` is functionally used during linting. So we might not need anything here.

tools/@aws-cdk/spec2cdk/test/services.test.ts
mrgrain: You can use `AstBuilder.forResource` to make the snapshot smaller.

What's your issue with snapshots?

tools/@aws-cdk/spec2cdk/test/services.test.ts
mrgrain: It clearly isn't 😝 

I'd probably prefer these assertions to be less focused on the exact format and more on the content. Currently you really are doing snapshot testing while trying to avoid using snapshots - and making it even harder.

tools/@aws-cdk/spec2cdk/test/services.test.ts
kaizencc: `// Snapshot tests will fail every time the docs get updated`

And in general, i prefer to avoid snapshots to be intentional on exactly what we are asserting. but it may be unavoidable in this case.

mrgrain: > // Snapshot tests will fail every time the docs get updated
Yeah that's a fair point 😅

tools/@aws-cdk/spec2cdk/test/services.test.ts
```diff
@@ -25,3 +25,147 @@ test('can codegen service with arbitrary suffix', () => {
...
+test('resource interface when primaryIdentifier is an attribute', () => {
+  const service = db.lookup('service', 'name', 'equals', 'aws-voiceid').only();
+
+  const ast = AstBuilder.forService(service, { db });
+
+  const rendered = renderer.render(ast.module);
+
+  expect(rendered).toContain([
+    '/**',
+    ' * Shared attributes for both `CfnDomain` and `Domain`.',
```
rix0rrr, line 38: This implies knowledge of L2s, which doesn't make sense yet. Don't refer to it as "shared attributes" I guess

tools/@aws-cdk/spec2cdk/test/services.test.ts
```diff
@@ -25,3 +25,147 @@ test('can codegen service with arbitrary suffix', () => {
...
+
+  const ast = AstBuilder.forService(service, { db });
+
+  const rendered = renderer.render(ast.module);
+
+  expect(rendered).toContain([
+    '/**',
+    ' * Shared attributes for both `CfnDataCellsFilter` and `DataCellsFilter`.',
+    ' *',
+    ' * @struct',
```
rix0rrr, line 90: It's definitely NOT a struct

kaizencc: Can i get clarification on what is/isn't a struct? I was confused as to why our property interfaces are typed as `StructType` and not `InterfaceType`, so I copied that behavior over. Right now, all our property interfaces have `@struct` too.

kaizencc: Figured the nuances out w @mrgrain

tools/@aws-cdk/spec2cdk/test/__snapshots__/resources.test.ts.snap
```diff
@@ -0,0 +1,940 @@
...
+ * @stability external
+ */
+export interface ICfnResource {
+  /**
+   * The identifier of the resource.
+   *
+   * @see http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-some-resource.html#cfn-some-resource-id
+   */
+  readonly attrId?: string;
+}
```
kaizencc, line 96: ^

tools/@aws-cdk/spec2cdk/test/__snapshots__/resources.test.ts.snap
```diff
@@ -0,0 +1,940 @@
...
+ * @stability external
+ */
+export interface ICfnResource {
+  /**
+   * The identifier of the resource
+   *
+   * @cloudformationAttribute Id
+   */
+  readonly attrId: string;
+}
```
kaizencc, line 250: ^

tools/@aws-cdk/spec2cdk/test/__snapshots__/resources.test.ts.snap
```diff
@@ -0,0 +1,940 @@
...
+   */
+  readonly attrId: string;
+
+  /**
+   * The arn for something
+   *
+   * @cloudformationAttribute SomethingArn
+   */
+  readonly attrSomethingArn: string;
+}
```
kaizencc, line 410: ^

tools/@aws-cdk/spec2cdk/test/__snapshots__/resources.test.ts.snap
```diff
@@ -0,0 +1,940 @@
...
+   */
+  readonly attrId: string;
+
+  /**
+   * The arn for the resource
+   *
+   * @cloudformationAttribute Arn
+   */
+  readonly attrArn: string;
+}
```
kaizencc, line 570: ^

tools/@aws-cdk/spec2cdk/test/__snapshots__/resources.test.ts.snap
```diff
@@ -0,0 +1,940 @@
...
+   */
+  readonly attrId: string;
+
+  /**
+   * The arn for the resource
+   *
+   * @cloudformationAttribute ResourceArn
+   */
+  readonly attrResourceArn: string;
+}
```
kaizencc, line 730: ^

tools/@aws-cdk/spec2cdk/lib/cdk/resource-class.ts
```diff
@@ -105,6 +129,22 @@ export class ResourceClass extends ClassType {
...
+    // Build the shared interface
+    for (const identifier of this.decider.primaryIdentifier ?? []) {
+      this.resourceInterface.addProperty({
+        ...identifier,
+        immutable: true,
+      });
+    }
+
+    // Add the arn too, unless it is duplicated in the resourceIdentifier already
+    if (this.decider.arn && this.resourceInterface.properties.every((p) => p.name !== this.decider.arn!.name)) {
```
kaizencc, line 141: @mrgrain  this fixes the duplicate identifier situation in vpc lattice (i'm 99% sure) with the test case to show it

mrgrain (approved):

PR #27692, opened 2023-10-26
fix(cli): fix stack monitoring when the stack events do not have phsical resource id set

Amplifiyer:
`PhysicalResourceId` for a `StackEvent` is [not required](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_StackEvent.html) and can be empty. This causes error in stack deployment activty monitoring where we try to monitor nested stacks but empty `PhysicalResourceId` in the `StackEvent` causes it to fail with the error

```
[ValidationError]: 2 validation errors detected: 
Value '' at 'stackName' failed to satisfy constraint: Member must have length greater than or equal to 1; 
Value '' at 'stackName' failed to satisfy constraint: Member must satisfy regular expression pattern: [a-zA-Z][-a-zA-Z0-9]*|arn:[-a-zA-Z0-9:/._+]*
```
This PR adds a check to avoid that.

----


madeline-k (approved):
The diff looks good! We just need to run this change through our full CI pipeline before it can be merged since it changes CLI code.

madeline-k:
The CLI tests are now in progress, I will update this PR with the results.

PR #27693, opened 2023-10-26
fix(config): explicit `ruleScope` value overwritten on `CloudFormationStackDriftDetectionCheck`

sHtev:
If `ruleScope` is set in the props of `CloudFormationStackDriftDetectionCheck`, and `ownStackOnly` is not set then the current code:

```ts
    this.ruleScope = RuleScope.fromResource( ResourceType.CLOUDFORMATION_STACK, props.ownStackOnly ? Stack.of(this).stackId : undefined );
```
overwrites  `this.ruleScope` with a resource `RuleScope` with an `undefined` resource id.

This makes it impossible to scope the drift check based on tags or resource ids other than that of the current stack.

The change in this PR explicitly checks for the `ownStackOnly` prop before overriding the ruleScope, and sets a sensible default if `ruleScope` itself is undefined to mirror existing default behaviour.

An example with a tag-filtering `RuleScope` has been added to the unit tests.

----


sHtev:
Note: the integration test currently has a warning it is undeployable, I hope this does not stop this needed fix being applied

sHtev:
Exemption Request if someone could help me understand why the codebuild fails when integration works locally that would be great

paulhcsun:
Hi @sHtev, apologies for not getting around to this, if you're still interested in working on this I've taken a look at the codebuild failure, usually integ test failures in the PR pipeline are because the snapshots you've added don't match up with what's being created by the pipeline. Could you just double check that the snapshots included in the PR are the most recent snapshots you have from running the integ tests locally?

paulhcsun:
Closing this PR for staleness. Please open a new PR if you'd like to continue working on this @sHtev.

PR #27611, opened 2023-10-19
feat(ec2): support for c7i and r7iz instance types in aws-ec2

kevntao:
Support of C7i and R7iz EC2-Instance types in aws-ec2.

The [C7i](https://aws.amazon.com/ec2/instance-types/c7i/) and the [R7iz](https://aws.amazon.com/ec2/instance-types/r7iz/) instance types were introduced in September 2023.
Both instance types are confirmed to be supported in [AWS-Cloudformation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-instancetype).

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #27737, opened 2023-10-27
feat(ec2): support for r7i instance type in aws-ec2

kevntao:
Support of R7i EC2-Instance type in aws-ec2.

The [R7i](https://aws.amazon.com/ec2/instance-types/r7i/) was recently introduced in October 2023.
The R7i instance type has been confirmed to be supported in [AWS-Cloudformation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-instancetype).

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
+++ b/packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
@@ -357,7 +357,7 @@ export enum InstanceClass {
   R7I = 'r7i',
 
   /**
-   * Memory optimized instances based on Intel Xeon Scalable (Sapphire Rapids) processors, 7th generation w/ sustained 3.9GHz turbo frequency
+   * Memory optimized instances based on Intel Xeon Scalable (Sapphire Rapids) processors, 7th generation, with sustained 3.9GHz turbo frequency
    */
   MEMORY7_INTEL = 'memory7-intel',
 


kevntao:
Exemption Request

This PR is similar to this [PR](https://github.com/aws/aws-cdk/pull/27611), which didn't need changes to the README or tests.

packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
```diff
@@ -326,6 +326,11 @@ export enum InstanceClass {
    */
   R7GD = 'r7gd',
 
+  /**
+   * Memory optimized instances based on Intel Xeon Scalable (Sapphire Rapids) processors, 7th generation
```
scanlonp, line 330: Lets change this to not be the same as r7iz. Or if the better change would be to that doc string, do that.

scanlonp (changes requested):
The doc string for r7i is the same as r7iz. There is also no "word" representation of the instance. Can we add one?

I suspect that `MEMORY7_INTEL` should have corresponded to `r7i`, but it is coupled with `r7iz`. We cannot change that now, but maybe we can come up with a good name for `r7i`. (Also we can put a comment that this is a 1-time deviation from our pattern.)

packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
kevntao: i agree, i had difficulty finding the nuanced difference between the `r7i` and `r7iz`, unfortunately you're correct, the `MEMORY7_INTEL` should have gone to the `r7i` instead of the `r7iz`, but I had thought the `r7iz` was the replacement for the `r7i` since it used the same processor generation as the `c7i` and `m7i`. 

I'm totally open to suggestions, but happy to take a stab at it for community review.

scanlonp: Yes, that mistake is on me. Do what you think looks good and we can go from there!

packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
```diff
@@ -347,12 +347,22 @@ export enum InstanceClass {
...
+  R7I = 'r7i',
+
+  /**
+   * Memory optimized instances based on Intel Xeon Scalable (Sapphire Rapids) processors, 7th generation w/ sustained 3.9GHz turbo frequency
    */
   MEMORY7_INTEL = 'memory7-intel',
 
   /**
-   * Memory optimized instances based on Intel Xeon Scalable (Sapphire Rapids) processors, 7th generation
+   * Memory optimized instances based on Intel Xeon Scalable (Sapphire Rapids) processors, 7th generation w/ sustained 3.9GHz turbo frequency
```
scanlonp, line 365:
```suggestion
   * Memory optimized instances based on Intel Xeon Scalable (Sapphire Rapids) processors, 7th generation, with sustained 3.9GHz turbo frequency
```

packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
```diff
@@ -347,12 +347,22 @@ export enum InstanceClass {
...
+   */
+  MEMORY7_INTEL_BASE = 'memory7-intel-base',
+
+  /**
+   * Memory optimized instances based on Intel Xeon Scalable (Sapphire Rapids) processors, 7th generation w/  3.2GHz turbo frequency
+   */
+  R7I = 'r7i',
+
+  /**
+   * Memory optimized instances based on Intel Xeon Scalable (Sapphire Rapids) processors, 7th generation w/ sustained 3.9GHz turbo frequency
```
scanlonp, line 360:
```suggestion
   * Memory optimized instances based on Intel Xeon Scalable (Sapphire Rapids) processors, 7th generation, with sustained 3.9GHz turbo frequency
```

scanlonp (approved):
I really like the `MEMORY7_INTEL_BASE`. Hopefully we get back on track with R8!

PR #27764, opened 2023-10-30
chore(cdk): boostrap template uses deprecated AccessControl property

scarytom:
Use of AccessControl is deprecated — see
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-s3-bucket.html#cfn-s3-bucket-accesscontrol


scarytom:
Clarification Request

kylelaker:
To fix the linter error, consider `chore(core): Use of AccessControl is deprecated`. The existing title seems to have lost the colon when the scope ("core") was added.

While the use of `AccessControl` is deprecated, it [defaults to `Private`](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl). It looks like this field has been in the template basically since [the beginning of time](https://github.com/aws/aws-cdk/commit/4d7d88bbbf17872280eae804d21f59bc970dd657#diff-fd2be8a383afcae4006cd8da81ef66921ac70ea59e87f2bee8f4db44746e3137R20). So it should not only be safe to remove this but functionally equivalent with the newer defaults for CFN and S3.

scarytom:
Thanks for the advice @kylelaker.  I've adjusted the title as suggested.

I was put onto this because cfn-lint has recently started warning about the use of `AccessControl`.  As you say, removing it should be functionally equivalent.

scarytom:
> ```
> ❌ Fixes must contain a change to a test file.
> ```

I don't think this trivial fix, which doesn't alter behavior in any way, warrants a test file change, so I'd like to make an Exemption Request please.

kylelaker (approved):

scanlonp (commented):
Not 100% sure about this change. I did want to confirm that this passed the testing pipeline.

scarytom:
> Not 100% sure about this change.

What aren't you sure about?

scanlonp (approved):
I wanted to test this locally. I bootstrapped a new region with this template, deploying a lambda function whose code is stored in the s3 staging bucket works normally.

Additionally, this change should have no impact for existing users since their environments are already bootstrapped.

Thanks!

scanlonp (approved):

PR #27765, opened 2023-10-30
fix(opensearchservice): IM4GN instances don't support EBS

### LINKED ISSUE #27757: aws-opensearchservice: No support for IM4GN instances
### Describe the bug

No support for IM4GN instances:

https://github.com/aws/aws-cdk/blob/72bf499071dd36ef54c24fce0ac265140291fb4f/packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts#L1544-L1548

IM4GN instances don't support EBS. Upgrading to IM4GN makes cdk deployments fail.

### Expected Behavior

IM4GN deployments are supported.

### Current Behavior

Error message on build:

`Error: EBS volumes are required when using instance types other than r3, i3 or r6gd.`

### Reproduction Steps

Change the instances type to any IM4GN instance in OS.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

Latest

### Framework Version

_No response_

### Node.js Version

Latest

### OS

MacOS

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

msambol:
Relevant docs https://aws.amazon.com/opensearch-service/pricing/

Closes #27757.

----


packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
```diff
@@ -1525,8 +1525,8 @@ export class Domain extends DomainBase implements IDomain, ec2.IConnectable {
 
     // Validate against instance type restrictions, per
     // https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-instance-types.html
-    if (isSomeInstanceType('i3', 'r6gd') && ebsEnabled) {
-      throw new Error('I3 and R6GD instance types do not support EBS storage volumes.');
+    if (isSomeInstanceType('r3', 'i3', 'r6gd', 'im4gn') && ebsEnabled) {
+      throw new Error('R3, I3, R6GD and IM4GN instance types do not support EBS storage volumes.');
```
msambol, line 1529: this `if` was missing `r3`

packages/aws-cdk-lib/aws-opensearchservice/test/domain.test.ts
```diff
@@ -1851,6 +1871,9 @@ each(testedOpenSearchVersions).describe('custom error responses', (engineVersion
       encryptionAtRest: {
         enabled: true,
       },
+      ebs: {
+        enabled: false,
+      },
```
msambol, line 1876: I changed this so it didn't error on the EBS check.

msambol:
Exemption Request: I don't think integration test changes are needed for these logic checks but happy to add them if needed.

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
kylelaker: The docs linked above say that `r3` is incompatible with encryption at rest but is silent about EBS specifically. Are the service docs missing that fact, is there another source for that limitation? https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-instance-types.html

msambol: I _think_ that doc is missing it? On the [pricing page](https://aws.amazon.com/opensearch-service/pricing/), it doesn't have EBS and is similar to `i3`, `r6gd`, and `im4gn`.

msambol: @kylelaker Let me fix this. I tested in the Console and you are allowed to have EBS for R3.

msambol: @kylelaker Thanks for catching this! I confirmed `r3` supports both instance and EBS.

msambol:
Exemption Request: I don't think integration test changes are needed for these logic checks but happy to add them if needed.

kylelaker (approved):
Thanks for looking into the R3 instances more! This looks good for adding the IM4GN support.

scanlonp (approved):
Good addition. Thanks!

I do not think we need the integ test in this case. I do not see instance type checks in any of our integ tests. The unit test should be sufficient.

PR #27768, opened 2023-10-30
chore(cloudfront): create unit tests to mock edge function handler

### LINKED ISSUE #27724: (cloudfront): create unit tests for edge function handler
### Describe the feature

We need to improve testing coverage of our Lambda handlers. The Lambda handler for [edge functions](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-cloudfront/lib/experimental/edge-function/index.js) does not have corresponding unit tests. Unit tests should be created specifically for this handler.

### Use Case

Unit tests for the edge function Lambda handler would improve test coverage and help us find bugs and prevent breaking changes.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.103.1

MacOS

colifran:
This PR adds three unit tests that mock functionality found in the Lambda handler for edge functions. Specifically, this PR is adding one unit test for a create event, one unit test for an update event, and one unit test for a delete event to verify that the handler only executes for create and update events.

Closes #27724

----


MrArnoldPalmer (commented):
What do you think of changing these to use [aws-sdk-mock](https://www.npmjs.com/package/aws-sdk-mock). It would be nice to start standardizing on this.

MrArnoldPalmer (approved):
Approving as there was an issue with aws-sdk-mock we need to work out separately.

PR #27771, opened 2023-10-31
chore: check region-info completeness

rix0rrr:
During the compatibility check, also check that no facts have disappeared from the `region-info` database.

----


scripts/check-region-info-compatibility.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const oldPkg: RegionInfoPackage = (require(oldPackage) as AwsCdkLib).region_info;
+  const newPkg: RegionInfoPackage = (require(newPackage) as AwsCdkLib).region_info;
+
+  const oldFacts = definedFacts(oldPkg.Fact);
+  const newFacts = definedFacts(newPkg.Fact);
+
+  const disappearedFacts = oldFacts.filter((oldFact) => !newFacts.some((newFact) => factEq(oldFact, newFact)));
+  if (disappearedFacts.length > 0) {
+    console.log('Facts have disappeared from region fact database (add to allowed-breaking-changes.txt to ignore):');
```
otaviomacedo, line 23: Shouldn't we log this only if we are sure to exit with code 1?

rix0rrr: Yessir

otaviomacedo (approved):

PR #27781, opened 2023-10-31
chore(spec2cdk): Resource class implements ICfnResource

kaizencc:
This PR implements the following:

- If all `primaryIdentifiers` are `attributes`, there is nothing to do; `ICfnResource` is already implemented by `Resource`
- if all `primaryIdentifiers` are `properties`, then
  - if there is 1 `primaryIdentifier` initialize it as `this.attrId = this.ref;`
  - if there are multiple `primaryIdentifiers` initialize each as `this.attrId = cdk.Fn.select(0, cdk.Fn.split('|', this.ref));` and so on

----


tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
```diff
@@ -73,23 +73,53 @@ export class ResourceDecider {
...
-            ...prop.docs,
-            remarks: prop.docs?.remarks?.concat(['\n', `@cloudformationRef ${prop.name}`].join('\n')),
+            ...propSpec.docs,
+            remarks: propSpec.docs?.remarks?.concat(['\n', '@cloudformationAttribute Ref'].join('\n')),
           },
+        };
+
+        // Add the new attribute to the relevant places
+        // TODO: we probably want this to be attribute properties but initializers are different
+        this.classProperties.push({
```
kaizencc, line 108: the type for `initializer` for `classProperties` is `(props: Expression): Expression` while the `initializer` for `attrProperties` is simply `Expression`. Since we need the function, I'm using `classProperties` for now.

rix0rrr:
> if all primaryIdentifiers are required, then initialize them as this.attrId = prop.Id;

This feels unsafe. The property can change over time, are you accounting for that? Seems safer to just take it from `{ Ref }`, potentially by picking apart the ref using `Fn.split()` and `Fn.select()` (the combination character is always a `|`):

```ts
// primaryIdentifier = ['ClusterName', 'AddonName']

this.attrClusterName = Fn.select(0, Fn.split('|', this.ref));
this.attrAddonName = Fn.select(1, Fn.split('|', this.ref));
```

mrgrain:
> > if all primaryIdentifiers are required, then initialize them as this.attrId = prop.Id;
> 
> This feels unsafe. The property can change over time, are you accounting for that? 

Howe can a prop change over time?

> Seems safer to just take it from `{ Ref }`, potentially by picking apart the ref using `Fn.split()` and `Fn.select()` (the combination character is always a `|`):

tl;dr We didn't find any information on the combination character and the order. If it's always `|` and the primary identifier list is ordered, then we can absolutely do this.

mrgrain:
> I could only find `expr.cond`, there's no `expr.nullish_coalescer` yet :)

I think we had some vague ideas about Typewriter generating Python code as well. So the expressions aimed to be as basic as possible.

mrgrain:
@kaizencc this is still creating duplicate `readonly attrArn: string;`

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
```diff
@@ -58,39 +61,73 @@ export class ResourceDecider {
...
     const possibleArnNames = ['Arn', 'ResourceArn', `${this.resource.name}Arn`];
     for (const arn of possibleArnNames) {
-      const att = this.classAttributeProperties.filter((a) => a.propertySpec.name === attributePropertyName(arn));
-      const prop = this.propsProperties.filter((p) => p.propertySpec.name === propertyNameFromCloudFormation(arn));
-      if (att.length > 0 || prop.length > 0) {
-        return att[0] ? att[0].propertySpec : prop[0].propertySpec;
-      }
+      const att = this.classAttributeProperties.find((a) => a.propertySpec.name === attributePropertyName(arn));
+      if (att) { return att.propertySpec; }
+      // TODO: double check -- if an arn is a property, not an attribute, and not a primaryIdentifier
```
kaizencc, line 66: need a double check here for what to do when the `arn` is from the `property` bag and not a `primaryIdentifier`. The problem is that this is currently generated as `this.arn = props.arn` and based on the spec it doesn't feel like the arn is actually returned via cfn.

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
```diff
@@ -58,39 +61,73 @@ export class ResourceDecider {
...
+
   private convertPrimaryIdentifier() {
-    for (const cfnName of this.resource.primaryIdentifier ?? []) {
+    if (this.resource.primaryIdentifier === undefined) { return; }
+    for (let i = 0; i < (this.resource.primaryIdentifier).length; i++) {
+      const cfnName = this.resource.primaryIdentifier[i];
       const att = this.findAttributeByName(attributePropertyName(cfnName));
       const prop = this.findPropertyByName(propertyNameFromCloudFormation(cfnName));
       if (att) {
         this.primaryIdentifier.push(att);
```
rix0rrr, line 95: What is `this.primaryIdentifier` used for?

Aren't we just saying that for primary identifier parts that are ALSO attributes, we don't need to do anything because they will already be initialized as part of some other code somewhere?

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
```diff
@@ -58,39 +61,73 @@ export class ResourceDecider {
...
+        const propSpec = prop.propertySpec;
+
+        // Build an attribute out of the property we're getting
+        // Create initializer for new attribute, if possible
+        let initializer: Expression | undefined = undefined;
+        if (propSpec.type === Type.STRING) { // handling only this case for now
+          if (this.resource.primaryIdentifier!.length === 1) {
+            initializer = CDK_CORE.tokenAsString($this.ref);
+          } else {
+            initializer = CDK_CORE.tokenAsString($Fn.select(expr.directCode(i.toString()), $Fn.split(expr.directCode('\'|\''), $this.ref)));
```
rix0rrr, line 106: Please avoid `expr.directCode()` as much as possible.

```suggestion
            initializer = CDK_CORE.tokenAsString($Fn.select(expr.lit(i), $Fn.split(expr.lit('|'), $this.ref)));
```

rix0rrr (changes requested):

rix0rrr:
> Howe can a prop change over time?

Properties are mutable fields. Think UI controls.

```ts
const bucket = new CfnBucket(this, 'Bucket', {
  bucketName: 'hello'
});

bucket.bucketName = 'bye'; // Totally allowed
```

rix0rrr (approved):

PR #27787, opened 2023-11-01
fix(logs): adding a resource policy statement with `AnyPrincipal` fails

### LINKED ISSUE #27783: aws-logs: LogGroup.addToResourcePolicy 'ARNs must start with "arn:" and have at least 6 components: *'
### Describe the bug

When I create a new LogGroup and then call addToResourcePolicy with a valid policy, I get the error message
```
    ARNs must start with "arn:" and have at least 6 components: *
```

### Expected Behavior

I should be able to `addToResourcePolicy`.

### Current Behavior

Error message.

### Reproduction Steps

https://github.com/ahammond/repro-loggroup-addtoresourcepolicy

```
import { App, RemovalPolicy, Stack, StackProps } from 'aws-cdk-lib';
import { AnyPrincipal, PolicyStatement } from 'aws-cdk-lib/aws-iam';
import { LogGroup } from 'aws-cdk-lib/aws-logs';

    super(scope, id, props);

    const r = new LogGroup(this, 'MyLogGroup', {
    });

    // This fails, no matter what the policy.
    r.addToResourcePolicy(new PolicyStatement({
      actions: ['logs:CreateLogGroupLogStream', 'logs:DescribeLogStreams', 'logs:PutLogEvents'],
      principals: [new AnyPrincipal()],
      resources: ['*'],
    }));
  }
}

// for development, use account/region from cdk cli
const devEnv = {
  region: process.env.CDK_DEFAULT_REGION,
};

const app = new App();
new MyStack(app, 'MyStack',  { env: devEnv });

app.synth();
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.103.1 (build 3bb19ac)

### Framework Version

same

### Node.js Version

20.5.1

### OS

MacOS latest

### Language

TypeScript

### Language Version

4.9.5

### Other information

_No response_

laurelmay:
Because `AnyPrincipal` extends `ArnPrincipal` it gets caught up in the checks for parsing the ARN from the principal to get the account. This check should be skipped when the ARN is set to `"*"` because that can't be parsed.

Closes #27783.

----


packages/@aws-cdk-testing/framework-integ/test/aws-logs/test/integ.log-resource-policy-any-principal.ts
```diff
@@ -0,0 +1,22 @@
...
+import { LogGroup } from 'aws-cdk-lib/aws-logs';
+import * as iam from 'aws-cdk-lib/aws-iam';
+
+class LogGroupIntegStack extends Stack {
+  constructor(scope: App, id: string, props?: StackProps) {
+    super(scope, id, props);
+    const logGroup = new LogGroup(this, 'LogGroup');
+    logGroup.addToResourcePolicy(new iam.PolicyStatement({
+      effect: iam.Effect.ALLOW,
+      resources: ['*'],
```
msambol, line 12: since this int test may be used as an example, maybe this should be `[logGroup.logGroupArn]` ?  The way it's written now, it results in a policy that says any anonymous principal can write to any log group in the account.

msambol: One more thought... is `addToResourcePolicy` misleading? It's a method on `LogGroup`, but it actually creates a `AWS::Logs::ResourcePolicy` which is not tied to an `LogGroup` at all, but at the account level. I almost think resources for this method should be set to `logGroup.logGroupArn` and there should be another mechanism to create a resource policy with parameterized resources.

kylelaker: I have changed this to the suggested value. I'd recommend that if there are larger concerns about the `addToResourcePolicy` API that they be addressed in a separate issue to make sure that they get the needed attention rather than on a PR targeted at a tangentially related bug fix.

packages/aws-cdk-lib/aws-logs/lib/log-group.ts
```diff
@@ -236,7 +236,7 @@ abstract class LogGroupBase extends Resource implements ILogGroup {
       return new iam.ArnPrincipal(principal.principalAccount);
     }
 
-    if (principal instanceof iam.ArnPrincipal) {
+    if (principal instanceof iam.ArnPrincipal && principal.arn !== '*') {
```
lpizzinidev, line 239: Can you please fix the typo in the function name? `convertArnPrincipalToAccountId`

lpizzinidev (changes requested):
Looks good to me 👍
Just a super-minor nit (it was annoying 🤓)

paulhcsun (dismissed):
Great work, thanks for adding this fix @kylelaker, and thanks for reviewing @lpizzinidev @msambol!

packages/aws-cdk-lib/aws-logs/lib/log-group.ts
paulhcsun: I'll push a change for this in a separate PR. Agreed that it's annoying but I don't want to hold this PR up because of it.

paulhcsun:
Not sure why tests are failing, I'll push an empty commit in a sec to trigger the build again.

paulhcsun (approved):

PR #27790, opened 2023-11-01
chore(awslint): new rules for reference interfaces

mrgrain:
Introducing two new rules, and immediately accepting all current violations.

- `resource-interface-extends-resource-ref` - all resources interfaces like `IBucket` must also implement the reference interface, e.g. `ICfnBucket` 
- `ref-via-ref-interface` - all APIs should accept the reference interface `ICfnBucket` over the resource interface `IBucket` 

----


packages/awslint/lib/rules/api.ts
```diff
@@ -3,143 +3,220 @@ import { ConstructReflection } from './construct';
...
+});
 
 apiLinter.add({
   code: 'ref-via-interface',
-  message: 'API should use interface and not the concrete class (%s). ' +
+  message:
+    'API should use interface and not the concrete class (%s). ' +
     `If this is intentional, add "${EXCLUDE_ANNOTATION_REF_VIA_INTERFACE}" to element's jsdoc`,
-  eval: e => {
+  eval: (e) => {
```
mrgrain, line 53: This rule hasn't changed. I just refactored the code used to visit spec nodes to be reusable.

packages/awslint/lib/rules/api.ts
```diff
@@ -3,143 +3,220 @@ import { ConstructReflection } from './construct';
 import { CoreTypes } from './core-types';
 import { Linter } from '../linter';
 
-const EXCLUDE_ANNOTATION_REF_VIA_INTERFACE = '[disable-awslint:ref-via-interface]';
+const EXCLUDE_ANNOTATION_REF_VIA_INTERFACE =
+  '[disable-awslint:ref-via-interface]';
+
+const EXCLUDE_ANNOTATION_REF_VIA_REF_INTERFACE =
+  '[disable-awslint:ref-via-ref-interface]';
```
mrgrain, line 10: I considered wrapping them both into a single rule, but found it too confusing and complicated.
Now we have two rules:
- `ref-via-interface` guarantees that we are not expecting concrete instances of constructs
- `ref-via-ref-interface` guarantees that we prefer `ICfnWhatever` reference interfaces over `IWhatever` interfaces

mikewrighton:
Generally looks good to me - did you see what impact this had on the lint times?

packages/awslint/lib/rules/construct.ts
```diff
@@ -184,6 +184,8 @@ constructLinter.add({
   message: 'construct interface must extend core.IConstruct',
   eval: e => {
     if (!e.ctx.interfaceType) { return; }
+    // ICfn interfaces are not construct interfaces
```
rix0rrr, line 187: I wonder if they should be though? `IBucket extends IConstruct`, I wonder if we should `ICfnBucket extends IConstruct` as well.

mrgrain: @kaizencc

mikewrighton (approved):

mrgrain:
Adds about 5s to awslint. Deeming this as acceptable. I am working on some speed-ups for the linter.

PR #27799, opened 2023-11-01
feat(scheduler-targets): add CodePipeline as target for scheduler

### LINKED ISSUE #27449: (aws-scheduler-targets-alpha): Add targets.CodePipelineStartPipelineExecution Target
### Describe the feature

Work to support L2 constructs for AWS Scheduler is in progress (https://github.com/aws/aws-cdk/issues/23394). See the approved [RFC](https://github.com/aws/aws-cdk-rfcs/pull/486/files). RFC planned to add 12 templates targets, but only Lambda Invoke is currently implemented (https://github.com/aws/aws-cdk/pull/26575). 

This issue tracks implementation of  `CodePipelineStartPipelineExecution` target to start an AWS CodePipeline execution. 

### Use Case

Customers would like to use templated target `CodePipelineStartPipelineExecution ` to be able to start AWS CodePipeline execution with AWS Scheduler. L2 target construct should grant required permissions to the `AWS Scheduler` to start the CodePipeline execution.  

### Proposed Solution

The proposed solution needs to be adopted to the recent examples of `LambdaInvoke` (https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-scheduler-targets-alpha/lib/lambda-invoke.ts). 

Solution should also include unit and integration tests. 

```ts
export class CodePipelineStartPipelineExecution  extends ScheduleTargetBase implements IScheduleTarget {
    constructor(
      private readonly pipeline: codepipeline.IPipeline,
      private readonly props: ScheduleTargetBaseProps,
    ) {
      super(props, pipeline.pipelineArn);
    }

   protected addTargetActionToRole(schedule: ISchedule, role: IRole): void {
      // TODO: Check if target and schedule are in the same account and region
      // TODO: Check if target and role are in the same account

      // TODO: Check if L2 grant methods can be used      
      role.addToPrincipalPolicy(new iam.PolicyStatement({
        actions: ['codepipeline:StartPipelineExecution'],
        resources: [this.pipeline.pipelineArn],
      }));
    }
  }
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.99.1

MacOs

daschaa:
Closes #27449

----

### DIFF
--- a/packages/aws-cdk-lib/core/lib/annotations.ts
+++ b/packages/aws-cdk-lib/core/lib/annotations.ts
@@ -143,7 +143,8 @@ export class Annotations {
   private addMessage(level: string, message: string) {
     const isNew = !this.scope.node.metadata.find((x) => x.data === message);
     if (isNew) {
-      this.scope.node.addMetadata(level, message, { stackTrace: this.stackTraces });
+      let normalizedMessage = typeof message === 'string' ? message : JSON.stringify(message);
+      this.scope.node.addMetadata(level, normalizedMessage, { stackTrace: this.stackTraces });
     }
   }
 }


packages/@aws-cdk/aws-scheduler-targets-alpha/lib/codepipeline-start-pipeline-execution.ts
```diff
@@ -0,0 +1,41 @@
...
+    private readonly pipeline: IPipeline,
+    private readonly props: ScheduleTargetBaseProps = {},
+  ) {
+    super(props, pipeline.pipelineArn);
+  }
+
+  protected addTargetActionToRole(schedule: ISchedule, role: IRole): void {
+    const pipelineArn = Arn.split(this.pipeline.pipelineArn, ArnFormat.NO_RESOURCE_NAME);
+    const region = pipelineArn.region ?? '';
+    const account = pipelineArn.account ?? '';
```
lpizzinidev, line 22: Isn't better to parse ARN attributes in [`fromPipelineArn`](https://github.com/aws/aws-cdk/blob/7c497ee998976b69ad1e2cd3e68f30cb593f8ded/packages/aws-cdk-lib/aws-codepipeline/lib/pipeline.ts#L322) directly?

lpizzinidev (commented):
Thanks, looks good!
Just a doubt on region and account values for imported pipelines.

daschaa:
> Thanks, looks good! 💪
> 
> Just an adjustment on region and account values for imported pipelines.

I have created an issue for that: https://github.com/aws/aws-cdk/issues/27816

daschaa:
@lpizzinidev Do you think I should "fix" the `fromPipelineArn` function in this PR or do this seperately in https://github.com/aws/aws-cdk/issues/27816 ?
The issue still under investigating, so i'm not sure if the implementation in `fromPipelineArn` is intended?

What are your thoughts on that?

lpizzinidev:
@daschaa 
I encountered the same problem while implementing https://github.com/aws/aws-cdk/pull/27837 for `SnsPublish`.
In my opinion, this is a bug on imported resources from ARNs, as you state in https://github.com/aws/aws-cdk/issues/27816 and, for the CodePipeline target, it should be fixed in this PR since it's directly related to the implementation.
However, it's also good to have an issue open to keep track of the "bug", as it does affect multiple resources and some discussion might be needed to get to an ideal and consistent solution.

daschaa:
@lpizzinidev Can you re-review this PR? :)

packages/aws-cdk-lib/aws-codepipeline/lib/pipeline.ts
```diff
@@ -325,7 +332,9 @@ export class Pipeline extends PipelineBase {
       public readonly pipelineArn = pipelineArn;
     }
 
-    return new Import(scope, id);
+    return new Import(scope, id, {
+      environmentFromArn: pipelineArn,
+    });
```
lpizzinidev, line 337: Can you please add unit tests to verify that the imported pipelines resolve account and region correctly?

lpizzinidev (changes requested):
Looks good, thanks!
Just some tests following the last changes are required in my opinion.

packages/aws-cdk-lib/aws-codepipeline/test/pipeline.test.ts
```diff
@@ -60,6 +60,21 @@ describe('', () => {
...
+          region: 'eu-west-1',
+          account: '234567890123',
+        },
+      });
+
+      const pipeline = codepipeline.Pipeline.fromPipelineArn(stack, 'Pipeline',
+        'arn:aws:codepipeline:us-east-1:123456789012:MyPipeline');
+
+      expect(pipeline.env.region).toEqual('us-east-1');
+      expect(pipeline.env.account).toEqual('123456789012');
```
lpizzinidev, line 75:
```suggestion
      const stack = new cdk.Stack();

      const pipeline = codepipeline.Pipeline.fromPipelineArn(stack, 'Pipeline',
        'arn:aws:codepipeline:us-west-1:123456789012:MyPipeline');

      expect(pipeline.env.region).toEqual('us-west-1');
      expect(pipeline.env.account).toEqual('123456789012');
```
Let's keep it simple.

lpizzinidev (changes requested):

packages/aws-cdk-lib/aws-codepipeline/test/pipeline.test.ts
daschaa: I prefer adding the env to the stack explicitly, because it shows that the import really does not use the environment of the stack 🤔

lpizzinidev (approved):

packages/aws-cdk-lib/aws-codepipeline/lib/pipeline.ts
kaizencc, line 337: This is just a bug fix to codepipeline that we should have had in a long time ago lol

kaizencc (dismissed):
Thanks @daschaa

kaizencc (approved):

PR #27803, opened 2023-11-01
test(ecs): add unit tests for drain hook lambda

### LINKED ISSUE #27725: (ecs): create unit tests for drain hook lambda handler
### Describe the feature

We need to improve testing coverage of our Lambda handlers. The Lambda handler for [drain hooks](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ecs/lib/drain-hook/lambda-source/index.py) does not have corresponding unit tests. Unit tests should be created specifically for this handler.

### Use Case

Unit tests for the drain hook Lambda handler would improve test coverage and help us find bugs and prevent breaking changes.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.103.1

MacOS

msambol:
Closes #27725.

----


msambol:
I need some help on this one as I don't normally test Python functions this way.

1. I'm not using `MagicMock`. Should I be?
2. Is there something I need to do in the build to make sure this gets run?

scanlonp:
@msambol it looks like the integ tests are failing since the handler code changed. That will cause the snapshots to be different. Simply running those integ tests should be good to get the build passing.

I am not as sure about needing MagicMock, I can look into it.

PR #27812, opened 2023-11-02
fix(ecs-patterns): scheduled fargate task runtime patform

hoonoh:
Apply runtime platform props to fargate task definition from `scheduledFargateTaskImageOptions` prop.

Without this patch, I was never unable to define Fargate CPU architecture.

This PR includes destructive changes with integrity tests. (adds fargate task definition with runtime platform option)

----


scanlonp (changes requested):
I see we have a warning stating that `runtimePlatform` is ignored in V2. I can look deeper into this warning, but it seems intentional. Therefore this is likely not something we want to support as is.

At best, this could go behind a new optional property or a feature flag, although neither seem quite right here. Destructive changes here will be prohibitive.

PR #27813, opened 2023-11-02
feat(cloudformation-diff): use awscdk-service-spec as data source

mrgrain:
We already changed the datasource for generated resource to use `@aws-cdk/awscdk-service-spec`.
However `cloudformation-diff` was still using the old data source. 

This PR swaps out the source, with minimal further changes.
Hence no additional tests are required.

Note that this new implementation does not apply `temporary-schemas` defined in `spec2cdk`.
This is because `temporary-schemas` is a mechanism to unblock development of unreleased features, and not a mechanism to publish schema changes.
For the latter, changes must be made in [cdklabs/awscdk-service-spec](https://github.com/cdklabs/awscdk-service-spec).

I ran a local benchmark and `cdk diff` is now running ~1.03 slower than before.
I investigated why and this appears to be mostly due to the gzip compression of the database file.
If we load the DB from an uncompressed source, the new code path performs ~1.07 times faster than the old code path. This trade-off is acceptable.

Compressed package size of `aws-cdk` is reducing with this change from `11.7MB` to `8.9MB`

----

### DIFF
--- a/.github/workflows/spec-update.yml
+++ b/.github/workflows/spec-update.yml
@@ -30,11 +30,11 @@ jobs:
       - name: Install ncu tool
         run: npm -g install lerna npm-check-updates
       - name: Run "ncu" for service spec packages
-        run: lerna exec --parallel ncu -- --upgrade --filter='@aws-cdk/aws-service-spec,@aws-cdk/service-spec-types' --target=latest
+        run: lerna exec --parallel ncu -- --upgrade --filter='@aws-cdk/aws-service-spec,@aws-cdk/service-spec-importers,@aws-cdk/service-spec-types' --target=latest
 
       # This will ensure the current lockfile is up-to-date with the dependency specifications
       - name: Install latest version & update lockfile
-        run: yarn upgrade @aws-cdk/aws-service-spec @aws-cdk/service-spec-types
+        run: yarn upgrade @aws-cdk/aws-service-spec @aws-cdk/service-spec-importers @aws-cdk/service-spec-types
 
       # Build @aws-cdk/spec2cdk and run L1 gen script to generate base files for new modules
       - name: Build @aws-cdk/spec2cdk
--- a/tools/@aws-cdk/spec2cdk/package.json
+++ b/tools/@aws-cdk/spec2cdk/package.json
@@ -33,7 +33,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@aws-cdk/aws-service-spec": "^0.0.26",
-    "@aws-cdk/service-spec-importers": "^0.0.1",
+    "@aws-cdk/service-spec-importers": "^0.0.4",
     "@aws-cdk/service-spec-types": "^0.0.26",
     "@cdklabs/tskb": "^0.0.1",
     "@cdklabs/typewriter": "^0.0.2",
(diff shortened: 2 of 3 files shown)


packages/@aws-cdk/cloudformation-diff/lib/diff/index.ts
```diff
@@ -36,8 +37,7 @@ export function diffResource(oldValue?: types.Resource, newValue?: types.Resourc
 
   if (resourceType.oldType !== undefined && resourceType.oldType === resourceType.newType) {
     // Only makes sense to inspect deeper if the types stayed the same
-    const typeSpec = cfnspec.filteredSpecification(resourceType.oldType);
-    const impl = typeSpec.ResourceTypes[resourceType.oldType];
+    const impl = database().lookup('resource', 'cloudFormationType', 'equals', resourceType.oldType).only();
```
rix0rrr, line 40: Same question

packages/@aws-cdk/cloudformation-diff/lib/model.ts
```diff
@@ -0,0 +1,11 @@
+import { loadAwsServiceSpecSync } from '@aws-cdk/aws-service-spec';
+import { SpecDatabase } from '@aws-cdk/service-spec-types';
+
+let DATABASE: SpecDatabase;
```
rix0rrr, line 4:
```suggestion
let DATABASE: SpecDatabase | undefined;
```

packages/@aws-cdk/cloudformation-diff/lib/diff/types.ts
```diff
@@ -119,16 +120,23 @@ export class TemplateDiff implements ITemplateDiff {
...
+      const newTypeProps = database().lookup('resource', 'cloudFormationType', 'equals', resourceChange.newResourceType).only().properties || {};
+      for (const [propertyName, prop] of Object.entries(newTypeProps)) {
+        const propScrutinyType = prop.scrutinizable || PropertyScrutinyType.None;
+        if (scrutinyTypes.includes(propScrutinyType)) {
+          ret.push({
+            resourceLogicalId,
+            propertyName,
+            resourceType: resourceChange.resourceType,
+            scrutinyType: propScrutinyType,
+            oldValue: resourceChange.oldProperties && resourceChange.oldProperties[propertyName],
```
rix0rrr, line 136: This can now be written as 
```suggestion
            oldValue: resourceChange.oldProperties?.[propertyName],
```

But I won't fault you if you don't care.

mrgrain: Nice catch! So far I was trying to keep the changes to a minimum.

packages/@aws-cdk/cloudformation-diff/lib/diff/types.ts
```diff
@@ -119,16 +120,23 @@ export class TemplateDiff implements ITemplateDiff {
...
-          resourceType: resourceChange.resourceType,
-          scrutinyType: cfnspec.propertySpecification(resourceChange.resourceType, propertyName).ScrutinyType!,
-          oldValue: resourceChange.oldProperties && resourceChange.oldProperties[propertyName],
-          newValue: resourceChange.newProperties && resourceChange.newProperties[propertyName],
-        });
+      if (!resourceChange.newResourceType) {
+        continue;
+      }
+
+      const newTypeProps = database().lookup('resource', 'cloudFormationType', 'equals', resourceChange.newResourceType).only().properties || {};
```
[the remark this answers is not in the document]
mrgrain: Good point.

PR #27744, opened 2023-10-29
feat(glue): database description property

msambol:
Closes #27740.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #27819, opened 2023-11-03
chore(prlint): don't mark PRs as 'not ready' on community comment

laurelmay:
Community reviewers have the ability to choose any of the
approve/comment/request changes buttons that are available in the review
tab. Prior to this change, `prlint` would consider a comment from a
community reviewer as being equivalent to requesting changes (and in
fact, didn't consider the requesting changes case). This would remove
the `pr/needs-community-review` label which surprised some reviewers.

With this change, `prlint` will only remove the
`pr/needs-community-review` label when a community reviewer specifically
chooses "request changes". Additionally, reviewers are now able to
switch from approving to requesting changes (this doesn't override any
other reviewers' approvals, just that reviewer's own).

Additionally, this adds mocks for the `getTrustedCommunityMembers`
method and avoids hardcoding the logins of multiple community reviewers
into the tests. As a side effect, the tests also run faster since `curl`
isn't being invoked so frequently.

Here is a screenshot of what I see when reviewing on this repo:
image

----


kylelaker:
It also seems unnecessary to filter out reviews from the PR author as GitHub already doesn't allow self-approvals or self-requests (though you can comment). Here is what I see on _this_ PR's review tab:

image

kylelaker:
For an example of a community member requesting changes, see https://github.com/aws/aws-cdk/pull/27744#pullrequestreview-1704173268. Even though GitHub shows it as "suggested changes", the API `state` field is still `CHANGES_REQUESTED`. https://api.github.com/repos/aws/aws-cdk/pulls/27744/reviews

go-to-k (approved):
Great! Thank you for addressing this matter.

vinayak-kukreja (approved):
Hey, this is great. Thank you for contributing this 🙌

PR #27839, opened 2023-11-04
chore(ec2): deprecate duplicate InterfaceVpcEndpointAwsService

### LINKED ISSUE #27662: aws-ec2: duplicate InterfaceVpcEndpointAwsService for SES, EMAIL_SMTP
### Describe the bug

The definition of `InterfaceVpcEndpointAwsService` contains definitions for [SES](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts#L441C15-L441C15) and [EMAIL_SMTP](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts#L465), but these both create services with the name `email-smtp[.suffix]`.

It is not clear what, if any, functional difference there is between these two.

### Expected Behavior

Either of the following:
* There only to be one service present for `email-smtp`
* The difference, if there is one, to be explained

### Current Behavior

Both are available, which means if in confusion or hedging your bets you try to use both, you cannot because CloudFormation will return:

```
Resource handler returned message: "private-dns-enabled cannot be set because there is already a conflicting DNS domain for email-smtp.[region].amazonaws.com in the VPC [VPC name]
```

### Reproduction Steps

```
import * as cdk from "aws-cdk-lib";
import * as ec2 from "aws-cdk-lib/aws-ec2";
import { Construct } from "constructs";


export class DemoStack extends cdk.Stack {
  constructor(
    scope: Construct,
    id: string,
    props: cdk.StackProps
  ) {
    super(scope, id, props);

    const vpc = new ec2.Vpc(this, "vpc", {
      ipAddresses: ec2.IpAddresses.cidr("10.1.0.0/16"),
      availabilityZones: ["eu-west-1a", "eu-west-1b", "eu-west-1c"],
      enableDnsHostnames: true,
      enableDnsSupport: true,
      subnetConfiguration: [
        {
          cidrMask: 26,
          name: "Endpoints",
          subnetType: ec2.SubnetType.PRIVATE_ISOLATED,
        },
      ],
    });

    new ec2.InterfaceVpcEndpoint(this, "sesEndpoint", {
      vpc: vpc,
      service: ec2.InterfaceVpcEndpointAwsService.SES,
      subnets: { subnetGroupName: "Endpoints" },
      privateDnsEnabled: true,
    });

    new ec2.InterfaceVpcEndpoint(this, "smtpEndpoint", {
      vpc: vpc,
      service: ec2.InterfaceVpcEndpointAwsService.EMAIL_SMTP,
      subnets: { subnetGroupName: "Endpoints" },
      privateDnsEnabled: true,
    });
  }
}
```

### Possible Solution

Document the reason for the presence of both, or, deprecate one of them.

As a user of SES, I would like to be clear if the interface endpoints provided give access to the SMTP host, the SES REST API, or both. It _appears_ to be the former, but one might expect the `SES` interface endpoint to provide access to the [API](https://docs.aws.amazon.com/ses/latest/APIReference/API_Operations.html) and the `EMAIL_SMTP` interface endpoint to provide access to the SMTP host (e.g. `email-smtp.eu-west-1.amazonaws.com`).

### CDK CLI Version

2.83.1 (build 006b542)

### Framework Version

_No response_

### Node.js Version

18.17.1

### OS

Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

Previous commits to add the existing options are [here](https://github.com/aws/aws-cdk/commit/06469a116ffe2b8aa6fb9639efc26c1c8ff1c6ac) and [here](https://github.com/aws/aws-cdk/commit/98386f58234e503066e95975cad5a4ec08eaafe7).

Both commits were released in version `2.47.0`. It feels likely that the release should have contained only one entry. I suspect `EMAIL_SMTP` makes the most sense because it explicitly refers to the SMTP host, I think.

@watany-dev and @clueleaf might have additional context here.

lpizzinidev:
Deprecates endpoints with the same name and reorganizes properties alphabetically.

Closes #27662.

----


daschaa (approved):

scanlonp (commented):
What was the rationale of deprecating `SES` over `Email_SMTP`?

lpizzinidev:
@scanlonp 
Naming. It depends on wheter the variable should have the service or the interface name.

scanlonp (approved):
Thanks for sharing your reasoning. I agree `EMAIL_SMTP` is the better choice. Looks good!

PR #27843, opened 2023-11-05
feat(cli): abort cli execution on invalid options

### LINKED ISSUE #27723: cdk: cdk CLI should abort in case an invalid option is provided
### Describe the bug

`cdk COMMAND` should fail in case a provided option is not known.

For example `cdk synth --foo-bar` or `cdk synth --output-file out.json` should fail so that the user is able to spot a mistake in the command.

### Expected Behavior

The command should fail in case a specified option is unknown.

### Current Behavior

The command ignores the failure and proceeds, the mistyped or unknown option is ignored.

In case of typos, the user will realize later, only after some investigation.

### Reproduction Steps

Run `cdk synth --foo-bar` or any command with an unknown/mistyped command line option. 

### Possible Solution

Implement a check in case of invalid options, or at least a big warning so that the user can notice the issue.

_No response_

### CDK CLI Version

any, but tested with cdk 2.101.0.

### Framework Version

_No response_

### Node.js Version

any

### OS

macos

### Language

### Language Version

_No response_

### Other information

_No response_

daschaa:
Adds `.strictOptions()` to yargs.

Closes #27723

----

### DIFF
--- a/packages/aws-cdk/test/options.test.ts
+++ b/packages/aws-cdk/test/options.test.ts
@@ -26,7 +26,7 @@ describe('options', () => {
     await expect(async () => {
       await exec(['version', '--foo-bar', '-v']);
     }).rejects.toThrowError(/Process exited/);
-    expect(mockedConsoleError).toHaveBeenCalledWith('Unknown arguments: foo-bar, fooBar');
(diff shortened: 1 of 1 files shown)


scanlonp (changes requested):
I think this change is good conceptually but may have some unintended consequences. Would this break user pipelines or other workflows if their cli commands have extra options?
There are also some of our existing tests failing with this change. Can you look into why those are failing?

PR #27844, opened 2023-11-05
chore(integ-tests): add waiterProvider to IApiCall

go-to-k:
This PR changes to add the `waiterProvider` property to an `IApiCall` for `awsApiCall` in integ-tests-alpha.

By default `awsApiCall` in integ tests, the AwsApiCall construct will automatically add the correct IAM policies to allow the Lambda function to make the API call. It does this based on the service and api that is provided. In the following example the service is SQS and the api is receiveMessage so it will create a policy with Action: 'sqs:ReceiveMessage'.

```ts
const integ = new IntegTest(app, 'Integ', {
  testCases: [stack],
});
integ.assertions.awsApiCall('SQS', 'receiveMessage', {
  QueueUrl: 'url',
});
```

There are some cases where the permissions do not exactly match the service/api call, for example the S3 listObjectsV2 api. In these cases it is possible to add the correct policy by accessing the `provider` object.

```ts
const apiCall = integ.assertions.awsApiCall('S3', 'listObjectsV2', {
  Bucket: 'mybucket',
});

apiCall.provider.addToRolePolicy({
  Effect: 'Allow',
  Action: ['s3:GetObject', 's3:ListBucket'],
  Resource: ['*'],
});
```

On the other hand, there is the case to use `waitForAssertions` when using `awsApiCall` in integ tests. This causes `apiCall` to have a `waiterProvider` property in addition to `provider`.

```ts
const apiCall = integ.assertions.awsApiCall('S3', 'listObjectsV2', {
  Bucket: 'mybucket',
}).expect(ExpectedResult.objectLike({
  KeyCount: 1,
})).waitForAssertions({
  interval: cdk.Duration.seconds(30),
  totalTimeout: cdk.Duration.minutes(10),
});
```

In the case, `waiterProvider` actually calls to the service/api, so it should have the proper policies.

However a type of a return value of `apiCall` is `IApiCall` interface so that the interface has a `provider` property, `waiterProvider` is not in `IApiCall` but in `AwsApiCall`.

Then it cannot take the policies without casting the following. (`apiCall instanceof AwsApiCall`)

```ts
if (apiCall instanceof AwsApiCall) {
  apiCall.waiterProvider?.addToRolePolicy({
    Effect: 'Allow',
    Action: ['s3:GetObject', 's3:ListBucket'],
    Resource: ['*'],
  });
}
```

So I add `waiterProvider` to `IApiCall`, so that it can take the policies without casting:

```ts
// if (apiCall instanceof AwsApiCall) {
  apiCall.waiterProvider?.addToRolePolicy({
    Effect: 'Allow',
    Action: ['s3:GetObject', 's3:ListBucket'],
    Resource: ['*'],
  });
//}
```

In my opinion, I see no negative impact from this.

----


lpizzinidev (changes requested):
Thanks, just a minor documentation adjustment on the current implementation.

Also:
* I think that an integration test should be added to verify that the policy is working as expected.
* Any reason why `waiterProvider` shouldn't be an abstract property in [`ApiCallBase`](https://github.com/aws/aws-cdk/blob/e6b59d6079fbaf8b2b79f47335b83902b1cda257/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/api-call-base.ts#L123) and used in [`HttpApiCall`](https://github.com/aws/aws-cdk/blob/e6b59d6079fbaf8b2b79f47335b83902b1cda257/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/http-call.ts#L16) as well after this change?

go-to-k:
> I think that an integration test should be added to verify that the policy is working as expected.

Yes, I will add the test.

> Any reason why waiterProvider shouldn't be an abstract property in [ApiCallBase](https://github.com/aws/aws-cdk/blob/e6b59d6079fbaf8b2b79f47335b83902b1cda257/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/api-call-base.ts#L123) and used in [HttpApiCall](https://github.com/aws/aws-cdk/blob/e6b59d6079fbaf8b2b79f47335b83902b1cda257/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/http-call.ts#L16) as well after this change?

I think there is probably no case to use `waiterProvider` with `HttpApiCall`. This is because `waiterProvider` can be used primarily to add additional IAM policies to a provider's role policy, but `HttpApiCall` does not require additional IAM policies. So I don't think `waiterProvider` needs to be an abstract property in `ApiCallBase`.

go-to-k:
> I think that an integration test should be added to verify that the policy is working as expected.

This `integ-tests-alpha` module does not have its own integ tests. So, do I just include the code that confirms this change in the integration test for the module of my choice somewhere?

go-to-k:
> I think that an integration test should be added to verify that the policy is working as expected.

This `integ-tests-alpha` module does not have its own integ tests. So, do I just include the code that confirms this change in the integration test for the module of my choice somewhere?

Also, is that not enough unit testing?

go-to-k:
> I think that an integration test should be added to verify that the policy is working as expected.

This `integ-tests-alpha` module does not have its own integ tests. So, do I just include the code that confirms this change in the integration test for the module of my choice somewhere?

Also, is unit testing alone not enough? If it is just a policy check, it seems to me that a unit test alone would suffice.

lpizzinidev:
@go-to-k 
You can write an integration test in the [test directory](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/integ-tests-alpha/test/assertions/providers/lambda-handler) of the module (similar to what is done [here](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/integ-tests-alpha/test/assertions/providers/integ.assertions.ts)).

We've integration tests for the `provider.addToRolePolicy` method, but not for the `waiterProvider` one, so I think it would be nice to have (given the fact that it will be only used in other integration tests).

Also, it should be documented [here](https://github.com/aws/aws-cdk/blob/38f3b8992f58d26947bd8c7b5ec64b973ccc8586/packages/%40aws-cdk/integ-tests-alpha/README.md?plain=1#L336).

go-to-k:
@lpizzinidev 

> You can write an integration test in the [test directory](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/integ-tests-alpha/test/assertions/providers/lambda-handler) of the module (similar to what is done [here](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/integ-tests-alpha/test/assertions/providers/integ.assertions.ts)).

Thanks for letting me know!  I missed it, I'll just put it here: integ-tests-alpha/test/assertions/providers/lambda-handler/integ.xx.ts.

> We've integration tests for the provider.addToRolePolicy method

I can't find this test in `providers/integ.assertions.ts`, which file is it? 

> Also, it should be documented [here](https://github.com/aws/aws-cdk/blob/38f3b8992f58d26947bd8c7b5ec64b973ccc8586/packages/%40aws-cdk/integ-tests-alpha/README.md?plain=1#L336).

Yes, I will take it.

go-to-k:
@lpizzinidev 

In the meantime, I've created an integ test, what do you think?

https://github.com/aws/aws-cdk/pull/27844/commits/eaf00746c8f308502333ba40b05b8b6d21b77336

packages/@aws-cdk/integ-tests-alpha/test/assertions/providers/lambda-handler/integ.waiter-provider.ts
```diff
@@ -0,0 +1,31 @@
...
+  interval: Duration.seconds(10),
+  totalTimeout: Duration.minutes(3),
+}).waiterProvider?.addToRolePolicy({
+  Effect: 'Allow',
+  Action: ['s3:GetObject', 's3:ListBucket'],
+  Resource: [
+    `${bucket.bucketArn}`,
+    `${bucket.bucketArn}/*`,
+  ],
+});
```
lpizzinidev, line 31:
```suggestion
const listObjectsCall = integ.assertions.awsApiCall('S3', 'listObjectsV2', {
  Bucket: bucket.bucketName,
  MaxKeys: 1,
}).expect(ExpectedResult.objectLike({
  KeyCount: 0,
})).waitForAssertions({
  interval: Duration.seconds(10),
  totalTimeout: Duration.minutes(3),
});

listObjectsCall.waiterProvider?.addToRolePolicy({
  Effect: 'Allow',
  Action: ['s3:GetObject', 's3:ListBucket'],
  Resource: [
    `${bucket.bucketArn}`,
    `${bucket.bucketArn}/*`,
  ],
});
```
To keep it cleaner.

lpizzinidev (changes requested):
The integration test looks good (just a minor adjustment is needed in my opinion) 👍
I'll review the README documentation when ready.

> I can't find this test in providers/integ.assertions.ts, which file is it?

`provider.addToRolePolicy` is already used in many integration tests across the project.

go-to-k:
@lpizzinidev 

Thanks for your review.

I changed them, and added the case to README. Please check them.

packages/@aws-cdk/integ-tests-alpha/README.md
```diff
@@ -522,3 +522,32 @@ const describe = testCase.assertions.awsApiCall('StepFunctions', 'describeExecut
...
+
+const apiCall = integ.assertions.awsApiCall('S3', 'listObjectsV2', {
+  Bucket: 'mybucket',
+}).waitForAssertions({
+  totalTimeout: Duration.minutes(5),
+  interval: Duration.seconds(15),
+  backoffRate: 3,
+});
+
+apiCall?.waiterProvider.addToRolePolicy({
```
lpizzinidev, line 548:
```suggestion
apiCall.waiterProvider?.addToRolePolicy({
```

packages/@aws-cdk/integ-tests-alpha/README.md
```diff
@@ -522,3 +522,32 @@ const describe = testCase.assertions.awsApiCall('StepFunctions', 'describeExecut
 });
 ```
 
+In cases where the `waitForAssertions()` is used for the `awsApiCall`, the actual API call is executed
+by the `waiterProvider`.
+
+Same as `provider`, by default, the `AwsApiCall` construct will automatically add the correct IAM policies
```
lpizzinidev, line 528:
```suggestion
By default, the `AwsApiCall` construct will automatically add the correct IAM policies
```

packages/@aws-cdk/integ-tests-alpha/README.md
```diff
@@ -522,3 +522,32 @@ const describe = testCase.assertions.awsApiCall('StepFunctions', 'describeExecut
 });
 ```
 
+In cases where the `waitForAssertions()` is used for the `awsApiCall`, the actual API call is executed
+by the `waiterProvider`.
```
lpizzinidev, line 526:
```suggestion
When `waitForAssertions()` is used for the `awsApiCall`, the actual API call is executed
by the `waiterProvider` assertion provider.
```

lpizzinidev (changes requested):

go-to-k:
@lpizzinidev 

Thanks! I changed.

go-to-k:
@lpizzinidev 

Thanks! I changed.

lpizzinidev (approved):

SankyRed:
Thank you for the contribution @go-to-k. The changes look good to me but, merge from `main` failed the build. Could you please take a look at this?

packages/@aws-cdk/integ-tests-alpha/lib/assertions/sdk.ts
```diff
@@ -50,19 +50,6 @@ export interface AwsApiCallProps extends AwsApiCallOptions { }
...
-   *
-   * @example
-   * declare const apiCall: AwsApiCall;
-   * apiCall.waiterProvider?.addToRolePolicy({
-   *   Effect: 'Allow',
-   *   Action: ['s3:GetObject'],
-   *   Resource: ['*'],
-   * });
-   */
   public waiterProvider?: AssertionsProvider;
```
kaizencc, line 53: Looks like the build is failing here. Can't remove this docstring entirely since its a public API. Why was it removed in the first place?

go-to-k: I see, I added it anyway.

I've moved the same explanation to `IApiCall`, and at few lines up `public readonly provider: AssertionsProvider;` had no docstring, so I removed this too. Why is it good that there is no docstring there (`provider`)?

go-to-k:
@SankyRed @kaizencc 

Thanks, I fixed and confirmed the CodeBuild success.

packages/@aws-cdk/integ-tests-alpha/lib/assertions/sdk.ts
go-to-k: I see, that's because it's readonly.

SankyRed (approved):
The change looks good to me. Thanks for the contribution @go-to-k

PR #27857, opened 2023-11-06
feat(gamelift): support Build serverSdkVersion, updated OperatingSystem values

### LINKED ISSUE #27655: (@aws-cdk/aws-gamelift-alpha): Missing properties to support GameLift Server SDK 5+
### Describe the bug

Two similar issues:

You can't set the `ServerSdkVersion` property when creating a build, which is mandatory to use SDK 5+ (since the default is 4 something). Here's [documentation on the property](https://docs.aws.amazon.com/gamelift/latest/apireference/API_CreateBuild.html#gamelift-CreateBuild-request-ServerSdkVersion) and [the property missing in the CDK documentation](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-gamelift-alpha.Build.html).

Similarly, the [enum for operating system](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-gamelift-alpha.OperatingSystem.html) doesn't allow setting `AMAZON_LINUX_2023` as [documented here](https://docs.aws.amazon.com/gamelift/latest/apireference/API_FleetAttributes.html#gamelift-Type-FleetAttributes-OperatingSystem).

Thank you!

### Expected Behavior

To be able to set `ServerSdkVersion` and `OperatingSystem`.

### Current Behavior

You can't set `ServerSdkVersion` and `OperatingSystem`.

### Reproduction Steps

N/A

### Possible Solution

- Add a `serverSdkVersion` prop to `Build` and pass it along when calling `CfnBuild`.
- Add `AMAZON_LINUX_2023` to the `OperatingSystem` enum. No logic changes seem to be necessary.

_No response_

### CDK CLI Version

2.101.1

### Framework Version

_No response_

### Node.js Version

v20.8.1

### OS

Ubuntu (WSL2 on Windows 11)

### Language

TypeScript

### Language Version

5.2.2

### Other information

_No response_

lpizzinidev:
Some adjustments to `Build` properties:
* Adds support for [`serverSdkVersion`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-gamelift-build.html#cfn-gamelift-build-serversdkversion) attribute
* Updates list of available [`operatingSystem`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-gamelift-build.html#cfn-gamelift-build-operatingsystem) and deprecates Windows Server 2012

Closes #27655.

----


packages/@aws-cdk/aws-gamelift-alpha/lib/build.ts
```diff
@@ -276,4 +313,13 @@ export class Build extends BuildBase {
     });
   }
 
+  private validateServerSdkVersion(serverSdkVersion?: string) {
+    if (!serverSdkVersion) return;
+    if (!serverSdkVersion.match(/^\d+\.\d+\.\d+$/)) {
+      throw new Error(`serverSdkVersion must be in the 0.0.0 format, got ${serverSdkVersion}.`);
+    }
+    if (serverSdkVersion.length > 128) {
+      throw new Error(`serverSdkVersion length must be smaller than 128, got ${serverSdkVersion}.`);
```
scanlonp, line 322:
```suggestion
      throw new Error(`serverSdkVersion length must be smaller than or equal to 128, got ${serverSdkVersion}.`);
```

packages/@aws-cdk/aws-gamelift-alpha/test/build.test.ts
```diff
@@ -183,6 +185,20 @@ describe('build', () => {
...
+          content,
+          serverSdkVersion: 'invalid',
+        })).toThrow(/serverSdkVersion must be in the 0.0.0 format, got invalid./);
+      });
+
+      test('with an incorrect serverSdkVersion length', () => {
+        expect(() => new gamelift.Build(stack, 'BuildWithInvalidServerSdkVersion', {
+          content,
+          serverSdkVersion: '1'.repeat(50) + '.' + '1'.repeat(50) + '.' + '1'.repeat(50),
+        })).toThrow(/serverSdkVersion length must be smaller than 128/);
```
scanlonp, line 200:
```suggestion
        })).toThrow(/serverSdkVersion length must be smaller than or equal to 128/);
```

scanlonp (dismissed):
Looks great! ~A minor nit on the error message, and then it should be good to go.~ Edited the error message.

One concern. Should `serverSdkVersion` be a public attribute on the `Build` and `Import` classes? It looks like the rest of the parameters are.

packages/@aws-cdk/aws-gamelift-alpha/lib/build.ts
```diff
@@ -276,4 +313,13 @@ export class Build extends BuildBase {
     });
   }
 
+  private validateServerSdkVersion(serverSdkVersion?: string) {
+    if (!serverSdkVersion) return;
```
go-to-k, line 317: It will be passed if a string is empty, so I would prefer an undefined check. And I think it should take the case that the param is Token by CfnParameter, etc., what do you think?

```suggestion
    if (serverSdkVersion === undefined || Token.isUnresolved(serverSdkVersion)) return;
```

packages/@aws-cdk/aws-gamelift-alpha/lib/build.ts
```diff
@@ -276,4 +313,13 @@ export class Build extends BuildBase {
...
+    if (!serverSdkVersion.match(/^\d+\.\d+\.\d+$/)) {
+      throw new Error(`serverSdkVersion must be in the 0.0.0 format, got ${serverSdkVersion}.`);
+    }
+    if (serverSdkVersion.length > 128) {
+      throw new Error(`serverSdkVersion length must be smaller than or equal to 128, got ${serverSdkVersion}.`);
```
go-to-k, line 322: I think it would be better to output the length of a character in the error message. I have seen other cases like this.

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/sagemaker/create-training-job.ts#L341
https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-events-targets/lib/batch.ts#L118
https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-glue-alpha/lib/database.ts#L136


```suggestion
      throw new Error(`serverSdkVersion length must be smaller than or equal to 128, got ${serverSdkVersion.length}.`);
```

go-to-k (changes requested):
Looks good to me. I had to comment on a couple of points.

go-to-k:
Oh, sorry, I timed my comment with @scanlonp.

lpizzinidev:
@go-to-k 
Nice catches! I missed both of them 😄

@scanlonp 
Some of the props are exposed, but others are not.
Unless there is a specific need to provide a reference to `serverSdkVersion` outside of the class I would keep it private.

go-to-k (approved):
@lpizzinidev Thanks! I approved.

scanlonp (approved):

PR #27860, opened 2023-11-06
fix(awslint): linting is slow

mrgrain:
Reduces runtime of `awslint` against `aws-cdk-lib` from ~70s down to ~15s.

Speed up 1: Reduce rule definitions (~1s)
Speed up 2: Make core checks fqn based only (~5s)
Speed up 3: Optimize code paths to defer expensive checks (~4s)
Speed up 4: Locked typesystem (~25s)
Speed up 5: Faster camel casing (~15s)

giphy

----

### DIFF
--- a/packages/aws-cdk-lib/aws-lambda/lib/function.ts
+++ b/packages/aws-cdk-lib/aws-lambda/lib/function.ts
@@ -1048,6 +1048,7 @@ export class Function extends FunctionBase {
     const reservedEnvironmentVariables = [
       '_HANDLER',
       '_X_AMZN_TRACE_ID',
+      'AWS_DEFAULT_REGION',
       'AWS_REGION',
       'AWS_EXECUTION_ENV',
       'AWS_LAMBDA_FUNCTION_NAME',


packages/aws-cdk-lib/awslint.json
mrgrain, line 1: Speed up 1

packages/awslint/.eslintrc.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
   parserOptions: {
     ecmaVersion: '2018',
     sourceType: 'module',
-    project: './tsconfig.json',
+    project: __dirname + '/tsconfig.json',
```
mrgrain, line 18: This was just broken before 🤷🏻

packages/awslint/bin/awslint.ts
```diff
@@ -50,7 +50,8 @@ async function main() {
 
   if (args._.length > 1) {
     argv.showHelp();
-    process.exit(1);
+    process.exitCode = 1;
+    return;
```
mrgrain, line 54: Unrelated change to use best practice.

packages/awslint/bin/awslint.ts
```diff
@@ -241,14 +242,17 @@ main().catch(e => {
...
+  process.exitCode = 1;
 });
 
 async function loadModule(dir: string) {
   const ts = new reflect.TypeSystem();
   await ts.load(dir, { validate: false }); // Don't validate to save 66% of execution time (20s vs 1min).
   // We run 'awslint' during build time, assemblies are guaranteed to be ok.
 
+  // We won't load any more assemblies. Lock the typesystem to benefit from performance improvements.
+  ts.lock();
```
mrgrain, line 254: This enables performance improvements in `jsii-reflect` that rely on a locked typesystem.

packages/awslint/lib/case.ts
```diff
@@ -0,0 +1,18 @@
+import * as changeCase from 'change-case';
```
mrgrain, line 1: I tested a bunch. 

`fast-case` is the fastest, but cannot deal with cases like `DBProxy` (converts it to `dBProxy`)
`lodash` and `change-case` are of similar speed
`camelcase` is by far the slowest.

packages/awslint/lib/case.ts
```diff
@@ -0,0 +1,18 @@
...
+
+function cachedTransform(transform: (x: string) => string): (x: string) => string {
```
mrgrain, line 3: Throwing this in for good measure, so we never convert the same string twice. It saves 3-4s.

packages/awslint/lib/rules/api.ts
```diff
@@ -126,7 +126,7 @@ apiLinter.add({
           return;
         }
 
-        if (type.type.fqn === e.ctx.core.constructClass.fqn) {
+        if (type.type.fqn === e.ctx.core.baseConstructClassFqn) {
```
mrgrain, line 129: By passes loading of the type just to compare the fqn.
I am making the assumption here that our core types are well tested and definitely exist in the assembly. If not jsii would blow up anyway.

packages/awslint/lib/rules/construct.ts
```diff
@@ -99,16 +93,9 @@ constructLinter.add({
     }
 
     const expectedParams = new Array<MethodSignatureParameterExpectation>();
-
-    let baseType;
-    if (process.env.AWSLINT_BASE_CONSTRUCT && !CoreTypes.isCfnResource(e.ctx.classType)) {
-      baseType = e.ctx.core.baseConstructClass;
-    } else {
-      baseType = e.ctx.core.constructClass;
-    }
```
mrgrain, line 108: This check was redundant because `e.ctx.core.baseConstructClass` and `e.ctx.core.constructClass` are the same.

packages/awslint/lib/rules/core-types.ts
```diff
@@ -158,11 +193,5 @@ export class CoreTypes {
       // disable-all-checks
       return;
     }
-
-    for (const fqn of Object.values(CoreTypesFqn)) {
-      if (!this.sys.tryFindFqn(fqn)) {
-        throw new Error(`core FQN type not found: ${fqn}`);
-      }
-    }
```
mrgrain, line 166: This is already checked by jsii

packages/awslint/lib/rules/imports.ts
```diff
@@ -70,8 +70,7 @@ importsLinter.add({
       // "fromRoleArn" => "roleArn"
       const argName = e.ctx.resource.basename[0].toLocaleLowerCase() + method.name.slice('from'.length + 1);
 
-      const baseType = process.env.AWSLINT_BASE_CONSTRUCT ? e.ctx.resource.core.baseConstructClass :
-        e.ctx.resource.core.constructClass;
+      const baseType = e.ctx.resource.core.baseConstructClass;
```
mrgrain, line 73: Both values are the same.

packages/awslint/lib/rules/imports.ts
```diff
@@ -92,8 +91,7 @@ importsLinter.add({
       return;
     }
 
-    const baseType = process.env.AWSLINT_BASE_CONSTRUCT ? e.ctx.resource.core.baseConstructClass
-      : e.ctx.resource.core.constructClass;
+    const baseType = e.ctx.resource.core.baseConstructClass;
```
mrgrain, line 94: Both values are the same.

packages/awslint/lib/rules/resource.ts
```diff
@@ -31,10 +31,9 @@ export class ResourceReflection {
       return []; // not part of the dep stack
     }
 
-    return ConstructReflection
-      .findAllConstructs(assembly)
-      .filter(c => CoreTypes.isResourceClass(c.classType))
```
mrgrain, line 36: This loops twice over all constructs.

packages/awslint/lib/rules/resource.ts
```diff
@@ -266,7 +265,7 @@ function tryResolveCfnResource(resourceClass: reflect.ClassType): CfnResourceRef
   }
 
   // try to resolve through ancestors
-  for (const base of resourceClass.getAncestors()) {
+  for (const base of resourceClass.ancestors) {
```
mrgrain, line 268: Using the memoized version of `getAncestors`

packages/aws-cdk-lib/awslint.json
mrgrain, line 1: Folding a bunch of rules into `*` rules. This is mostly enums.

vinayak-kukreja (commented):
This is amazing 😍

packages/awslint/lib/rules/construct.ts
```diff
@@ -24,11 +24,6 @@ export class ConstructReflection {
     return typeRef.fqn;
   }
 
-  /**
-   * @deprecated - use `CoreTypes.constructClass()` or `CoreTypes.baseConstructClass()` as appropriate
-   */
-  public readonly ROOT_CLASS: reflect.ClassType; // constructs.Construct
```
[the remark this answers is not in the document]
mrgrain: It's public but in "dev preview". Has been for ages.
https://www.npmjs.com/package/awslint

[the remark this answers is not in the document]
mrgrain: Never mind, this isn't even a public API. It's not exported at the top-level.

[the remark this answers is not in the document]
mrgrain: It's also not a library. It's a cli tool. 😅

vinayak-kukreja (approved):

PR #27861, opened 2023-11-06
fix(apigatewayv2-integrations): integrations of imported lambda functions do not get configured

### LINKED ISSUE #18965: (aws_apigatewayv2.HttpLambdaIntegration): add_routes integration to subroute doesn't take affect
### What is the problem?

Given (Pre configuration required): 
`from aws_cdk.aws_apigatewayv2_integrations import HttpLambdaIntegration`
API gateway configured with routes & subroutes, connected to a lambda function via CDK (Python)

Looks like a stack is not accepting multiple `aws_apigatewayv2.HttpApi.add_routes()` as done below in test.py

When (Action taken): Function looked up via `Function.from_function_arn` and applied to integration parameter of `http_api.add_routes`

### Reproduction Steps

app.py
```
import os
from aws_cdk import core as cdk
from src.aws_modules.APIGateway_test.test import testStack as test

app = cdk.App()
environment = pc.getStackEnvironment()

if __name__ == "__main__":
    test(scope=app, id="test-stack-id",env=environment).run_stack()
    app.synth()

```
test.py
```
import os
from aws_cdk.aws_apigatewayv2_integrations import HttpLambdaIntegration
from aws_cdk import (aws_apigatewayv2 as apigwv2,aws_lambda as lambda_)
from aws_cdk import core as cdk
from aws_cdk.aws_apigatewayv2_authorizers import HttpLambdaAuthorizer, HttpLambdaResponseType

class testStack(cdk.Stack):


    def run_stack(self):

        current_region = os.environ['CDK_DEFAULT_REGION']
        current_account = os.environ['CDK_DEFAULT_ACCOUNT']
        authorizer_lambda_name = "test-authorizer-api-dev"
        integration_lambda_name = "test-service-dev"

        api_name = "my-test-API-for-AWS"
        http_api = apigwv2.HttpApi(self, "test-api-gatewayv2",
        cors_preflight=apigwv2.CorsPreflightOptions(
        allow_headers=["*"],
        allow_methods=[apigwv2.CorsHttpMethod.ANY],
        allow_origins=["*"],
        max_age=cdk.Duration.days(10)

    )
        )


        authorizer_function = lambda_.Function.from_function_arn(
            self,
            id="found-function-authorizer",
            function_arn="arn:aws:lambda:{}:{}:function:{}".format(current_region,current_account,authorizer_lambda_name)
            )
        authorizer_test = HttpLambdaAuthorizer("test-authorizer", authorizer_function,
            response_types=[HttpLambdaResponseType.SIMPLE]
        )

        integration_function = lambda_.Function.from_function_arn(
            self,
            id="found-function-integration",
            function_arn="arn:aws:lambda:{}:{}:function:{}".format(current_region,current_account,integration_lambda_name)
            )
        function_integration = HttpLambdaIntegration("Test-FunctionIntegration", integration_function)


        http_api.add_routes(
            path="/companies",
            methods=[apigwv2.HttpMethod.GET],
            integration=function_integration,
            authorizer=authorizer_test
            )

        http_api.add_routes(
            path="/companies/swagger",
            methods=[apigwv2.HttpMethod.GET],
            integration=function_integration
            )
```

Then (Expected behavior, per documentation): Route has lambda integration defined, attached and working.

### What actually happened?

Actual (Observed behavior): Route gains integration ID and appears to have function assigned to it but this is not the case. This is verified by navigating to the lambda function under Configuration, the sub-route is not defined as an endpoint. 

### CDK CLI Version

2.8.0 (build 8a5eb49)

### Framework Version

_No response_

### Node.js Version

v14.17.3

### OS

MacOS

### Language

Python

### Language Version

Python 3.9.8

### Other information

This looks to be an issue with the `add_routes` method as the `from_function_arn` works for the authorizer. 

Workaround: 
1. CDK deploys API gateway 
2. Navigate API Gateway > integrations > Select sub route > Select HTTP method
3. Select "Manage integration" button
4. Select "Edit"
5. Select "Save" button at bottom right
6. Check lambda function > Configuration
7. You will see the route is now there.

sumupitchayan:
Closes #18965 

`HTTPLambdaIntegration` using imported lambda functions are currently not being configured. This is due to the fact that the `canCreatePermissions` property is set to false in imported lambdas, so the lambda permissions are never created.

This PR fixes this issue by:
- Setting the `sameEnvironment` property to `true` for lambdas imported using `fromFunctionName()`
- Adding clarification in documentation about imported lambdas needing to belong to the same stack account and region as that of the lambda being imported from be able to add permissions
- Adding a new warning to be thrown when the `addPermissions()` function is called on a lambda with `canCreatePermissions` set to false.

The new integ test file tests that imported lambdas using `fromFunctionName()` and `fromFunctionAttributes()` (with `sameEnvironment` set to `true`) work as expected, including on sub-route integrations.

----


packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -500,11 +500,15 @@ export class Function extends FunctionBase {
...
+      sameEnvironment: true,
     });
   }
 
   /**
-   * Import a lambda function into the CDK using its ARN
+   * Import a lambda function into the CDK using its ARN.
+   *
+   * To be able to add permissions to this imported lambda, make sure that is in the same account
+   * and region as the stack you are importing it into.
```
rix0rrr, line 511: Maybe phrase this as `Function.addPermissions()` specifically, instead of "add permissions", because it depends on how you do it.

I think we also need to be a bit more specific on when it works and when it doesn't. Use the blurb I typed for some inspiration.

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -514,6 +518,9 @@ export class Function extends FunctionBase {
    * Creates a Lambda function object which represents a function not defined
    * within this stack.
    *
+   * To be able to add permissions, set the sameEnvironment property to true if this imported lambda is
```
rix0rrr, line 521: Same here.

rix0rrr (approved):

PR #27871, opened 2023-11-07
chore(aws-cdk-lib): remove superfluous script

mrgrain:
This checks was introduced when aws-cdk-lib was still a "virtual" package and created by combing different sources. Since this is not the case anymore, we can remove this script.

Fixes failing build.

----


rix0rrr (approved):

mrgrain:
@mergifyio update

PR #27877, opened 2023-11-07
chore: fix broken link in README.md

blackcat1383:
The link to nodejs.org is broken.

From: https://nodejs.org/en/about/releases
image

To: https://nodejs.org/en/about/previous-releases
image

Please let me know if the [download](https://nodejs.org/en/download) page is a better fit for this...

----


scanlonp (approved):
Good find! That link looks like the right one.

PR #27886, opened 2023-11-08
fix(lambda-nodejs): docker image is built when not required

tmokmss:
This is a small optimization to get rid of unnecessary building of container images. Currently the docker image required for NodejsFunction bundling (e.g. `public.ecr.aws/sam/build-nodejs18.x`) is always built regardless of necessity, which sometimes makes cdk commands slow.

This PR fixes the current behavior; If bundling is not required for an NodejsFunction's asset, we just skip it, which is also a consistent behavior with PythonFunction.

https://github.com/aws/aws-cdk/blob/de0a9e218eeccbe4f685e45f93e25a250733dc51/packages/%40aws-cdk/aws-lambda-python-alpha/lib/bundling.ts#L56-L62

https://github.com/aws/aws-cdk/blob/de0a9e218eeccbe4f685e45f93e25a250733dc51/packages/%40aws-cdk/aws-lambda-python-alpha/lib/function.ts#L76

edit) turns out it's not very easy to fix, because currently the construct seems to expect the bundling preprocess always run to work properly.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-lambda-nodejs/lib/bundling.ts
+++ b/packages/aws-cdk-lib/aws-lambda-nodejs/lib/bundling.ts
@@ -49,13 +49,6 @@ export interface BundlingProps extends BundlingOptions {
    * @default - BundlingFileAccess.BIND_MOUNT
    */
   readonly bundlingFileAccess?: cdk.BundlingFileAccess;
-
-  /**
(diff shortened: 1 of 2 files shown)


tmokmss:
Exemption Request: As far as I know this is kind of a change that cannot be tested with integ-runner, because the change affects cases when we don't deploy a stack.

PR #27891, opened 2023-11-08
fix(stepfunctions-tasks): missing permissions for running tasks on ecs

msambol:
While working on [#27803](https://github.com/aws/aws-cdk/pull/27803), I noticed the integration tests for `aws-stepfunctions-tasks/ecs` were not fully working (they deployed but the state machines did not run successfully). This PR addresses two issues:

1. Missing permissions for `ecs:RunTask` on the task definition version.
sfn-role

2. The sample container was from a Lambda image. This resulted in the following error: `entrypoint requires the handler name to be the first argument`. I changed the image to `docker/library/python:3.12`.

These changes result in the successful execution of all four state machines in `aws-stepfunctions-tasks/ecs`.

----


packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/ecs/eventhandler-image/Dockerfile
```diff
@@ -1,5 +1,3 @@
-FROM public.ecr.aws/lambda/python:3.6
-EXPOSE 8000
-WORKDIR /src
-ADD . /src
-CMD python3 index.py
+FROM --platform=x86-64 public.ecr.aws/docker/library/python:3.12
+ADD index.py .
+CMD [ "python3", "./index.py" ]
```
msambol, line 3: I needed `--platform=x86-64` since I am building on a Mac M2.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
```diff
@@ -340,7 +340,10 @@ export class EcsRunTask extends sfn.TaskStateBase implements ec2.IConnectable {
     const policyStatements = [
       new iam.PolicyStatement({
         actions: ['ecs:RunTask'],
-        resources: [this.getTaskDefinitionFamilyArn()],
+        resources: [
+          this.getTaskDefinitionFamilyArn(),
+          `${this.getTaskDefinitionFamilyArn()}:*`,
```
msambol, line 345: Drawing attention to the bits that add permission to run versions of the task definition.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/ecs/eventhandler-image/Dockerfile
lpizzinidev: Instead of changing the image have you tried changing `CMD`:
```
CMD [ "index.handler" ]
```
and defining a handler function in `index.py`:
```
import os
import print

def handler(event, context):
    print('Hello from ECS!')
    pprint.pprint(dict(os.environ))
```
?

lpizzinidev (changes requested):
Thanks!
Some minor comments and a note on the Docker image change.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/ecs/eventhandler-image/Dockerfile
msambol: I did try that but didn't spend a bunch of time on it. I can revisit.. but my thinking is, why is it a Lambda image when this is running on ECS and not in Lambda?

lpizzinidev: Good point, thanks for clarifying.

lpizzinidev (approved):

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/ecs/eventhandler-image/Dockerfile
vinayak-kukreja, line 1: Could you tell me why some of these were removed?

And if we can remove `--platform=x86-64`?

vinayak-kukreja (commented):
Hey @msambol , this is great. Just a clarification comment.

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/ecs/eventhandler-image/Dockerfile
msambol: We don't need some of these commands in the `Dockerfile`. 

I needed `--platform=x86-64` because I am building on an M2 and got an architecture mismatch. Unless I'm doing something wrong?

msambol:
@vinayak-kukreja Any additional feedback on this?

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/ecs/eventhandler-image/Dockerfile
kaizencc: I'm not too familiar with Docker but I'm also concerned with this. What about other users who are not on an M2 but on a computer incompatible with this platform? Maybe @mrgrain can weigh in further because I'm not well-versed myself in Docker stuff.

msambol: @mrgrain Have some time to look at this?

msambol: @mrgrain I'm not sure if I was doing something wrong or something changed, but it's working now without "platform" 😅

msambol:
@mrgrain Have time for another look at this one ?

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/ecs/eventhandler-image/Dockerfile
msambol: @mrgrain wait... I think this is still an issue. I get `exec /usr/local/bin/python3: exec format error` when I run the task without `--platform=linux/amd64`. Let me keep testing..

msambol: @mrgrain If I don't use `--platform=linux/amd64`, I get the following error when the task runs: `exec /usr/local/bin/python3: exec format error`. Similar here: https://github.com/aws/aws-cdk/pull/28898.

msambol:
@mrgrain Mind taking another look here?

kaizencc (approved):
Thanks and sorry for the delay @msambol

PR #27903, opened 2023-11-08
fix(lambda-python-alpha): bundling hash logic includes auth tokens in PIP urls, causing an unnecessary rebuild

### LINKED ISSUE #27331: aws_lambda_python_alpha: Using PIP_INDEX_URL and Codeartifact cause Docker to rebuild lambda every synth
### Describe the bug

When following the instructions for using [CodeArtifact](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda_python_alpha/README.html#custom-bundling-with-code-artifact), the lambda artifact is rebuilt anytime the CodeArtifact token changes, despite the actual code not changing at all.

### Expected Behavior

The lambda shouldn't be rebuilt unless the actual code changes.

### Current Behavior

The lambda is rebuilt with a new hash, which causes an unnecessary redeploy. This slows down the synth and deploy steps, and causes cold starts for all lambdas.

### Reproduction Steps

Define any lambda function using the Custom Bundling with Code Artifact instructions. Observe that the lambda is rebuilt on each synth, even with no code changes.

### Possible Solution

The Token included in the CodeArtifact URL should be ignored when computing the hash that determines if the lambda is rebuilt. Alternately, provide an option to ignore the PIP_INDEX_URL entirely.

_No response_

### CDK CLI Version

2.95.1

### Framework Version

_No response_

### Node.js Version

v20.6.1

### OS

OSX 12.2

### Language

Python

### Language Version

Python 3.9.16

### Other information

_No response_

kcp-chewie:
Update the bundler hash logic to ignore the secret token that is included in the URL when fetching packages from Code Artifact. This token changes constantly, and prevents the results of a previous build from being reused, along with causing lambdas to be unnecessarily redeployed anytime the CDK is built, even if no code is changed.

This implementation strips the token from the hash, but does not change anything else about the hash. 

Open question
Currently this logic will error if an invalid URL string is passed for PIP_INDEX_URL or PIP_EXTRA_INDEX_URL. The build would fail later anyway when those URLs are being used, but I'm happy to try/catch this logic block to be more robust if that is preferred.

Closes #27331.

----


lpizzinidev (changes requested):
Thanks 👍
Just some minor adjustments in my opinion.
Also, can you please change the title to describe the bug (not the solution) as per [guidelines](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request)?

kcp-chewie (commented):
Thanks for the review!

lpizzinidev (approved):
Looks good to me, thanks 👍

kcp-chewie:
Anything I can do to help this one get reviewed/merged?

CC @lpizzinidev

kaizencc (approved):
Thanks @kcp-chewie. I surmise that this will cause unnecessary rebuilds for the first time people using `PIP_INDEX_URL` update their CDK with this change, but I suppose we are okay with that.

kaizencc (approved):

kcp-chewie:
Thanks @kaizencc ! 

For reference, this should only cause an unnecessary rebuild for anyone PIP_INDEX_URL with basic auth username/password and a static password. Folks using PIP_INDEX_URL without any auth should see no rebuild, and folks using PIP_INDEX_URL with a rotating token (like AWS CodeArtifact users) will see 1 final rebuild.

PR #27907, opened 2023-11-09
fix(events): imported event rule ignores environment from arn

go-to-k:
This PR fixes the bug that `fromEventRuleArn` does not extract region and account from the arn.

- Behavior until now

```ts
    const stack = new Stack(); // <- region: us-east-1, account: 123456789012
    const imported = Rule.fromEventRuleArn(stack, 'Imported', 'arn:aws:events:us-west-2:999999999999:rule/example');

    // imported.env.region : us-east-1("Ref": "AWS::Region"). But it should be us-west-2.
    // imported.env.account : 123456789012("Ref": "AWS::AccountId"). But it should be 999999999999.
```

----


go-to-k:
Exemption Request, because I think unit tests could cover it.

lpizzinidev (approved):
Thanks 👍

vinayak-kukreja:
@Mergifyio update

vinayak-kukreja (approved):
Hey, this is great. Thank you for your contribution.

PR #27909, opened 2023-11-09
fix(eks): neuron plugin not added with addNodegroupCapacity()

gdamjan:
Previously, the neuron plugin is already added in the addAutoScalingGroupCapacity() method, when the instanceType was some of the INFERENTIA types.

Let's also add the plugin if at least one (some/any) of the instanceTypes in addNodegroupCapacity() is of the INFERENTIA types.

----


gdamjan:
# Clarification Request

❌ Features must contain a change to a README file.

The neuron plugin was not mentioned previously in the aws-eks README file, so I didn't know where to document this semi-new behaviour. Should I add a note that the plugin is added when using both `addNodegroupCapacity()` and `addAutoScalingGroupCapacity()` ?

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
```diff
@@ -1800,6 +1800,9 @@ export class Cluster extends ClusterBase {
    * @param options options for creating a new nodegroup
    */
   public addNodegroupCapacity(id: string, options?: NodegroupOptions): Nodegroup {
+    if (options?.instanceTypes?.some(i => nodeTypeForInstanceType(i) === NodeType.INFERENTIA)) {
+      this.addNeuronDevicePlugin();
+    }
```
lpizzinidev, line 1805:
```suggestion
    const hasInferentiaInstanceType = [
      options?.instanceType,
      ...options?.instanceTypes ?? [],
    ].some(i => i && nodeTypeForInstanceType(i) === NodeType.INFERENTIA);
    if (hasInferentiaInstanceType) {
      this.addNeuronDevicePlugin();
    }
```
Let's also check the deprecated `instanceType` attribute.
Can you please add unit tests as well?

lpizzinidev (changes requested):
Thanks for the contribution 👍

I think that this is more a `fix` than a `feature`, can please you update the title following the [guidelines](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request)?

Also, an integration test is needed.
You can use [this](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/framework-integ/test/aws-eks/test/integ.eks-inference.ts) as a template and check the [documentation](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md) for its execution.

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
gdamjan: sure

lpizzinidev (approved):
Thanks, looks good 👍
Can you please update the title to `fix(eks): neuron plugin not added with addNodegroupCapacity`?
It should [describe the bug (not the solution)](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request).

gdamjan:
> Thanks, looks good 👍 Can you please update the title to `fix(eks): neuron plugin not added with addNodegroupCapacity`? It should [describe the bug (not the solution)](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request).

should I change the commit itself too, or you'll squash merge with using the PR topic?

lpizzinidev:
@gdamjan 
No, it's fine like this, thanks for the update 👍
The reason is that `fix` and `feature` titles will appear on the release CHANGELOG, so they need to follow a certain format.

tests.txt
```diff
@@ -303,6 +303,7 @@ aws-eks/test/integ.eks-cluster-private-endpoint.js
 aws-eks/test/integ.eks-cluster.js
 aws-eks/test/integ.eks-helm-asset.js
 aws-eks/test/integ.eks-inference.js
+aws-eks/test/integ.eks-inference-nodegroup.js
```
scanlonp, line 306:
```suggestion

```
Do not think we need to update this file.

scanlonp (changes requested):
One nit.
Do you think it would be helpful to have information about neuron plugins in the readme? If so add it, if not, then I think the exemption is fine.
Other than that, I am good to approve.

tests.txt
gdamjan: afaik, the integration tests were tying to run the `.ts` file in that case and failing. I can try again.

kaizencc (approved):

PR #27913, opened 2023-11-09
feat(stepfunctions): support Map ItemProcessor

### LINKED ISSUE #27878: aws-cdk-lib/aws-stepfunctions: AWS StepFunctions recommends using ItemProcessor vs Iterator
### Describe the feature

The StepFunction documentation has the following note [here](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html)

```
The ItemProcessor field replaces the now deprecated [Iterator](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#iterator) field. Although you can continue to include Map states that use the Iterator field, we highly recommend that you replace this field with ItemProcessor.
```

### Use Case

Unable to use [Distributed mode](https://docs.aws.amazon.com/step-functions/latest/dg/use-dist-map-orchestrate-large-scale-parallel-workloads.html) for Map. 

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.104.x

Unix

lpizzinidev:
Adds support for Map's [`ItemProcessor`](https://docs.aws.amazon.com/step-functions/latest/dg/use-dist-map-orchestrate-large-scale-parallel-workloads.html#distitemprocessor) required field and deprecates [`Iterator`](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#iterator).

Closes #27878.

----


packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
```diff
@@ -411,11 +427,9 @@ export abstract class State extends Construct implements IChainable {
    * Render map iterator in ASL JSON format
    */
   protected renderIterator(): any {
-    if (!this.iteration) {
-      throw new Error('Iterator must not be undefined !');
-    }
+    if (!this.iteration) return undefined;
```
paulhcsun, line 430: Is it safe to remove the error here? Maybe we could add an additional check for `!this.processor` so that we can still throw the error if neither are set.

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
```diff
@@ -438,6 +452,40 @@ export abstract class State extends Construct implements IChainable {
     });
   }
 
+  /**
+   * Render ItemProcessor in ASL JSON format
+   */
+  protected renderItemProcessor(): any {
+    if (!this.processor) return undefined;
```
paulhcsun, line 459: Same as above, do you think we should error here if `!this.processor && !this.iteration`?

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
lpizzinidev: I removed validation here to handle it in [`validateState`](https://github.com/aws/aws-cdk/pull/27913/files#diff-cca37caeb11e88e3ed977617a93def71dc433b7b12428ab0c7a9d6e964d76e80L194).
I think it's clearer to have it centralized and we shouldn't miss coverage unless I'm missing something.

packages/aws-cdk-lib/aws-stepfunctions/README.md
```diff
@@ -518,9 +518,29 @@ const definition = choice
     .afterwards()
     .next(finish);
 
-map.iterator(definition);
+map.itemProcessor(definition);
 ```
 
+To define a distributed `Map` state use the configuration parameter.
```
kaizencc, line 524: unclear to me what the "configuration parameter" is

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map.ts
```diff
@@ -194,8 +207,16 @@ export class Map extends State implements INextable {
   protected validateState(): string[] {
     const errors: string[] = [];
 
-    if (this.iteration === undefined) {
-      errors.push('Map state must have a non-empty iterator');
+    if (this.iteration === undefined && this.processor === undefined) {
```
kaizencc, line 210: I mean, this is fine, but why not `if (!this.iteration && !this.processor)`

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map.ts
```diff
@@ -162,13 +162,25 @@ export class Map extends State implements INextable {
 
   /**
    * Define iterator state machine in Map
+   *
+   * @deprecated - use `itemProcessor` instead.
    */
   public iterator(iterator: IChainable): Map {
```
kaizencc, line 168: Docstring should mention mutex with itemProcessor

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map.ts
```diff
@@ -162,13 +162,25 @@ export class Map extends State implements INextable {
...
     const name = `Map ${this.stateId} Iterator`;
     super.addIterator(new StateGraph(iterator.startState, name));
     return this;
   }
 
+  /**
+   * Define item processor in Map
+   */
+  public itemProcessor(processor: IChainable, config: ProcessorConfig = {}): Map {
```
kaizencc, line 177: Docstring should mention mutex with iterator

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
```diff
@@ -411,11 +427,9 @@ export abstract class State extends Construct implements IChainable {
    * Render map iterator in ASL JSON format
    */
   protected renderIterator(): any {
-    if (!this.iteration) {
-      throw new Error('Iterator must not be undefined !');
-    }
+    if (!this.iteration) return undefined;
     return {
-      Iterator: this.iteration.toGraphJson(),
+      Iterator: this.iteration?.toGraphJson(),
```
kaizencc, line 432: don't need this `?` right? typescript should know whats up based on line 430

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
```diff
@@ -438,6 +452,40 @@ export abstract class State extends Construct implements IChainable {
...
 
+  /**
+   * Render ItemProcessor in ASL JSON format
+   */
+  protected renderItemProcessor(): any {
+    if (!this.processor) return undefined;
+    return {
+      ItemProcessor: {
+        ...this.renderProcessorConfig(),
+        ...this.processor?.toGraphJson(),
```
kaizencc, line 463: Shouldn't TS know `this.processor` exists without the `?`?

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
```diff
@@ -438,6 +452,40 @@ export abstract class State extends Construct implements IChainable {
...
+      },
+    };
+  }
+
+  /**
+   * Render ProcessorConfig in ASL JSON format
+   */
+  private renderProcessorConfig(): any {
```
kaizencc, line 471:
```suggestion
  private renderProcessorConfig() {
```

`any` doesn't help much, i'd rather TS guses at the return type

kaizencc (dismissed):
Some preliminary thoughts @lpizzinidev, but thanks for getting this started!

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
lpizzinidev: Makes sense, I specified the return type as this is the approach used for the other rendering functions.

kaizencc (approved):

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
kaizencc: Fair :). If there's a reason for it I can't think of one, but it's whatever either way.

PR #27934, opened 2023-11-10
feat(codebuild): Lambda compute for codebuild projects

### LINKED ISSUE #28418: aws-codebuild: CodeBuild lambda images/compute types missing from CDK
### Describe the feature

Recently, CodeBuild released a new set of lambda images which can be found [here](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html) under `Lambda compute images`. Along with these new images, CodeBuild released new lambda compute types listed [here](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-compute-types.html) as the BUILD_LAMBDA_1GB-10GB compute types. The CodeBuild CDK code should be updated to reflect these new images/compute types.

### Use Case

When trying to run lambda CodeBuild builds using CDK, the lambda project cannot be created as the relevant CDK values are missing.

### Proposed Solution

Create new ARM/Linux image files with the new lambda images and compute types

### Other Information

_No response_

### Acknowledgements

### CDK version used

Any

Any

sakurai-ryo:
CodeBuild has added support for Lambda compute.
CloudFormation can be deployed as follows.
```yaml
Resources:
  CodeBuildProject:
    Type: AWS::CodeBuild::Project
    Properties:
      Artifacts:
        Type: NO_ARTIFACTS
      ServiceRole: !GetAtt CodeBuildRole.Arn
      Source:
        # 
      Environment:
        Type: LINUX_LAMBDA_CONTAINER
        ComputeType: BUILD_LAMBDA_1GB
        Image: aws/codebuild/amazonlinux-x86_64-lambda-standard:go1.21
  CodeBuildRole:
    Type: AWS::IAM::Role
    Properties:
        # 
```
https://aws.amazon.com/about-aws/whats-new/2023/11/aws-codebuild-lambda-compute

This PR implements Lambda ComputeType by adding Classes (`LinuxArmLambdaBuildImage`, `LinuxLambdaBuildImage`) that extend the IBuildImage interface.

Supported Docker Images and ComputeTypes are listed below.
https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html
https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-compute-types.html#environment.types

Also, Lambda compute has some limitations and I have added validation for them.
https://docs.aws.amazon.com/codebuild/latest/userguide/lambda.html#lambda.limitations

closes #28418

----


packages/aws-cdk-lib/aws-codebuild/test/linux-arm-lambda-build-image.test.ts
```diff
@@ -0,0 +1,178 @@
+import { Template } from '../../assertions';
+import * as codecommit from '../../aws-codecommit';
+import * as cdk from '../../core';
+import * as codebuild from '../lib';
+
+describe('Linux ARM Lambda build image', () => {
```
sakurai-ryo, line 6: Since there are no changes to the test content, I've written the tests only in Node.js Image.
If it looks like I should add other Images, I will add tests using jest.each, etc.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1903,7 +1935,15 @@ export class LinuxBuildImage implements IBuildImage {
...
-    return [];
+    const ret = [];
+
+    const lambdaComputeTypes = Object.values(ComputeType).filter(value => value.startsWith('BUILD_LAMBDA'));
+    if (_env.computeType && lambdaComputeTypes.includes(_env.computeType)) {
+      ret.push(`x86-64 images only support ComputeTypes between '${ComputeType.SMALL}' and '${ComputeType.X2_LARGE}' - ` +
+               `'${_env.computeType}' was given`);
+    }
+
+    return ret;
```
lpizzinidev, line 1946:
```suggestion
    const ret = [];

    const lambdaComputeTypes = Object.values(ComputeType).filter(value => value.startsWith('BUILD_LAMBDA'));
    if (this.isLambdaComputeType(env.computeType)) {
      ret.push('x86-64 images do not support Lambda compute mode');
    }

    return ret;
```
It's better to declare a separate function to check if the compute type is a Lambda one and re-use it.
```
private isLambdaComputeType(computeType?: string): boolean {
  if (!computeType) return false;
  return [
    ComputeType.LAMBDA_1GB,
    ComputeType.LAMBDA_2GB,
    ...
  ].includes(computeType);
}

File: packages/aws-cdk-lib/aws-codebuild/lib/project.ts [hunk 3]
```diff
@@ -1542,6 +1562,10 @@ export class Project extends ProjectBase {
       throw new Error('Both source and artifacts must be set to CodePipeline');
     }
   }
+
+  private isLambdaComputeType(buildImage: IBuildImage): boolean {
```
lpizzinidev, line 1566:
```suggestion
  private isLambdaBuildImage(buildImage: IBuildImage): boolean {
```
Naming.

packages/aws-cdk-lib/aws-codebuild/README.md
```diff
@@ -396,6 +401,24 @@ new codebuild.Project(this, 'Project', {
...
+  environment: {
+    buildImage: codebuild.LinuxLambdaBuildImage.AMAZON_LINUX_2_NODE_18,
+  },
+  // ...
+})
+```
+
+Custom images are not supported, so please use a static method such as `AMAZON_LINUX_2_NODE_18`.
+See documentation for other constraints.
+https://docs.aws.amazon.com/codebuild/latest/userguide/lambda.html#lambda.limitations
```
lpizzinidev, line 420:
```suggestion
> Visit [AWS Lambda compute in AWS CodeBuild](https://docs.aws.amazon.com/codebuild/latest/userguide/lambda.html) for more details.
```

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1334,13 +1334,33 @@ export class Project extends ProjectBase {
...
+
     if (errors.length > 0) {
       throw new Error('Invalid CodeBuild environment: ' + errors.join('\n'));
     }
 
-    const imagePullPrincipalType = this.buildImage.imagePullPrincipalType === ImagePullPrincipalType.CODEBUILD
-      ? ImagePullPrincipalType.CODEBUILD
-      : ImagePullPrincipalType.SERVICE_ROLE;
+    // For Lambda compute, specifying imagePullPrincipalType is not supported
+    const imagePullPrincipalType = this.isLambdaComputeType(this.buildImage)
```
lpizzinidev, line 1359: Is this needed?
If CFN does not deploy it should be validated in `validateLambdaBuildImage` otherwise I think we can just ignore the value.

lpizzinidev (changes requested):
Thanks for the contribution! 👍
Looks good overall, just some notes on documentation and code structure.

packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Validates by checking unsupported property and compute type
+   * @param buildEnvironment BuildEnvironment
+   */
+  public validate(buildEnvironment: BuildEnvironment): string[] {
+    const ret = [];
+
+    if (buildEnvironment.privileged) {
+      ret.push('Lambda compute type does not support Privileged mode');
```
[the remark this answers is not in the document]
sakurai-ryo: Same validation and testing added for `WindowsBuildImage`.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
sakurai-ryo: added implementation as a function instead of a method so that it can be used in other files.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
sakurai-ryo: In the original implementation, `SERVICE_ROLE` is specified as the default value even if `imagePullPrincipalType` is not specified.
https://github.com/aws/aws-cdk/blob/0178cd39fca2dfe0ae1872b823e6a79dc91a1c78/packages/aws-cdk-lib/aws-codebuild/lib/project.ts#L1341

```ts
    const imagePullPrincipalType = this.buildImage.imagePullPrincipalType === ImagePullPrincipalType.CODEBUILD
        ? ImagePullPrincipalType.CODEBUILD
        : ImagePullPrincipalType.SERVICE_ROLE;

    return {
      imagePullCredentialsType: imagePullPrincipalType,
    };
```

Therefore, I think some kind of modification other than validation is necessary.
Any advice on a better implementation would be appreciated!

sakurai-ryo:
@lpizzinidev 
Thanks for the review! I've learned a lot.
Please check out some of the changes and comments I have made.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
lpizzinidev: I think that `imagePullPrincipalType`, `secretsManagerCredentials`, and `repository` will just be ignored if a Lambda-compute image is passed (I might be wrong on this).
If this is the case we could provide some feedback with `Annotations.addWarningV2()` to notify the users that those properties will not be used.
I don't think we should set `imagePullPrincipalType` to `undefined` if `this.isLambdaBuildImage(this.buildImage)` because it would either unset an already ignored value or mask a wrong property specification that should be validated.

lpizzinidev (changes requested):
Thanks for the changes 👍
Just some minor adjustments are needed.
Also, I would change the title to `feat(codebuild): support Lambda compute`.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1715,8 +1752,11 @@ interface LinuxBuildImageProps {
 }
 
 // Keep around to resolve a circular dependency until removing deprecated ARM image constants from LinuxBuildImage
-// eslint-disable-next-line no-duplicate-imports, import/order
+/* eslint-disable no-duplicate-imports, import/order */
 import { LinuxArmBuildImage } from './linux-arm-build-image';
+import { LinuxArmLambdaBuildImage } from './linux-arm-lambda-build-image';
+import { LinuxLambdaBuildImage } from './linux-lambda-build-image';
+/* eslint-enable no-duplicate-imports, import/order */
```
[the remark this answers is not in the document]
sakurai-ryo: Since `LinuxArmLambdaBuildImage` and `LinuxLambdaBuildImage` reference `ComputeType`, it's necessary to place the import statements below `ComputeType`. I've adjusted their placement just beneath `ComputeType`. How does that look?

However, these classes only use `ComputeType` when generating error messages. It's also possible to move the import statements to the top of the file by modifying the error messages that don't use `ComputeType`. This option might be more appropriate.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
sakurai-ryo: > I think that imagePullPrincipalType, secretsManagerCredentials, and repository will just be ignored if a Lambda-compute image is passed (I might be wrong on this).

I agree with you.
`imagePullPrincipalType`, `secretsManagerCredentials`, and `repository` are all specified within `IBuildImage`, but the Lambda Build Image doesn't have these properties. Therefore, users generally cannot explicitly specify these values.

However, when it comes to `imagePullPrincipalType`, even if these values are not specified within the Lambda Build Image, the `Project.renderEnvironment` method defaults to setting `imagePullPrincipalType` to `SERVICE_ROLE`. This causes deployment to result in an error message `Cannot specify imagePullPrincipalType for lambda compute`.

So, I think some modifications is needed in lines like the following within the `renderEnvironment` method for the deployment to work correctly.
https://github.com/aws/aws-cdk/blob/14662072692e5555cdfe44d9cfa5d9d66e03c7bf/packages/aws-cdk-lib/aws-codebuild/lib/project.ts#L1341
https://github.com/aws/aws-cdk/blob/14662072692e5555cdfe44d9cfa5d9d66e03c7bf/packages/aws-cdk-lib/aws-codebuild/lib/project.ts#L1364

sakurai-ryo:
@lpizzinidev 
Thank you very much for the review! I have made the necessary corrections and added comments.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -2066,6 +2115,15 @@ export class WindowsBuildImage implements IBuildImage {
 
   public validate(buildEnvironment: BuildEnvironment): string[] {
     const ret: string[] = [];
+
+    if (buildEnvironment.privileged) {
+      ret.push('Windows images do not support privileged mode');
+    }
+
+    if (buildEnvironment.computeType && isLambdaComputeType(buildEnvironment.computeType)) {
+      ret.push('Windows images do not support Lambda compute mode');
```
lpizzinidev, line 2124:
```suggestion
      ret.push('Windows images do not support Lambda compute types');
```

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -2173,3 +2231,14 @@ export enum ProjectNotificationEvents {
 function isBindableBuildImage(x: unknown): x is IBindableBuildImage {
   return typeof x === 'object' && !!x && !!(x as any).bind;
 }
+
+export function isLambdaComputeType(computeType?: ComputeType): boolean {
+  if (!computeType) return false;
```
lpizzinidev, line 2236:
```suggestion
export function isLambdaComputeType(computeType: ComputeType): boolean {
```
On second thought, let's make the property non-optional.
After all, we want to check only defined values.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1334,13 +1334,19 @@ export class Project extends ProjectBase {
...
 
-    const imagePullPrincipalType = this.buildImage.imagePullPrincipalType === ImagePullPrincipalType.CODEBUILD
-      ? ImagePullPrincipalType.CODEBUILD
-      : ImagePullPrincipalType.SERVICE_ROLE;
+    // For Lambda compute, specifying imagePullPrincipalType is not supported
+    const imagePullPrincipalType = this.isLambdaBuildImage(this.buildImage)
+      ? undefined
+      : this.buildImage.imagePullPrincipalType === ImagePullPrincipalType.CODEBUILD
+        ? ImagePullPrincipalType.CODEBUILD
+        : ImagePullPrincipalType.SERVICE_ROLE;
```
lpizzinidev, line 1349:
```suggestion
    const imagePullPrincipalType = this.buildImage.imagePullPrincipalType === ImagePullPrincipalType.CODEBUILD
      ? ImagePullPrincipalType.CODEBUILD
      : ImagePullPrincipalType.SERVICE_ROLE;
```
Let's keep it as it was. See https://github.com/aws/aws-cdk/pull/27934#discussion_r1390885565.

packages/aws-cdk-lib/aws-codebuild/test/codebuild.test.ts
```diff
@@ -1649,6 +1649,47 @@ describe('Windows2019 image', () => {
...
+      const stack = new cdk.Stack();
+
+      expect(() => {
+        new codebuild.PipelineProject(stack, 'Project', {
+          environment: {
+            buildImage: codebuild.WindowsBuildImage.WIN_SERVER_CORE_2019_BASE,
+            computeType: codebuild.ComputeType.LAMBDA_1GB,
+          },
+        });
+      }).toThrow(/Invalid CodeBuild environment: Windows images do not support Lambda compute mode/);
```
lpizzinidev, line 1663:
```suggestion
      }).toThrow(/Invalid CodeBuild environment: Windows images do not support Lambda compute types/);
```

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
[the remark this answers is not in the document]
lpizzinidev: Thanks for clarifying, let's just keep them together as you did in the original commit then. It's clearer.

lpizzinidev (changes requested):
Thanks, just some minor adjustments and it will be good to go for me.
Note https://github.com/aws/aws-cdk/pull/27934#discussion_r1392125468.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1542,6 +1545,32 @@ export class Project extends ProjectBase {
...
+  }
+
+  /**
+   * Validates a Lambda build image given the project properties.
+   * @see https://docs.aws.amazon.com/codebuild/latest/userguide/lambda.html#lambda.limitations
+   */
+  private validateLambdaBuildImage(buildImage: IBuildImage, props: ProjectProps): string[] {
+    if (!this.isLambdaBuildImage(buildImage)) return [];
+    const ret = [];
+    if (props.timeout) {
```
lpizzinidev, line 1560:
```suggestion
    if (buildImage.imagePullPrincipalType) {
      ret.push('Cannot specify imagePullPrincipalType for Lambda compute images. Lambda compute doesn\'t support custom images.');
    }
    if (buildImage.repository) {
      ret.push('Cannot specify repository for Lambda compute images. Lambda compute doesn\'t support custom images.');
    }
    if (buildImage.secretsManagerCredentials) {
      ret.push('Cannot specify secretsManagerCredentials for Lambda compute images. Lambda compute doesn\'t support custom images.');
    }
    if (props.timeout) {
```
We should just validate `imagePullPrincipalType`, `repository`, and `secretsManagerCredentials` for the Lambda compute image (will also need related unit tests).

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1361,7 +1364,7 @@ export class Project extends ProjectBase {
     return {
       type: this.buildImage.type,
       image: this.buildImage.imageId,
-      imagePullCredentialsType: imagePullPrincipalType,
+      imagePullCredentialsType: this.isLambdaBuildImage(this.buildImage) ? undefined : imagePullPrincipalType,
```
lpizzinidev, line 1367:
```suggestion
      imagePullCredentialsType: imagePullPrincipalType,
```
Will remain unchanged after validating the image.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -2173,3 +2228,14 @@ export enum ProjectNotificationEvents {
 function isBindableBuildImage(x: unknown): x is IBindableBuildImage {
   return typeof x === 'object' && !!x && !!(x as any).bind;
 }
+
+export function isLambdaComputeType(computeType: ComputeType): boolean {
+  if (!computeType) return false;
```
lpizzinidev, line 2233:
```suggestion
```
Always defined.

lpizzinidev (changes requested):
Thanks for the changes and the clarifications.
I missed https://github.com/aws/aws-cdk/pull/27934#discussion_r1391322308 (sorry for that). 
I think we should just validate those properties in `validateLambdaBuildImage`.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
sakurai-ryo: Previously, setting `imagePullPrincipalType` to `CODEBUILD` seemed to cause an error (really sorry if I misunderstood). However, I just rechecked the behavior and found that deploying with `imagePullPrincipalType` set to `CODEBUILD` succeeds.
Therefore, I've changed the default value of `imagePullPrincipalType` in the `LinuxLambdaBuildImage` and `LinuxArmLambdaBuildImage` classes to `CODEBUILD`.

For `repository` and `secretsManagerCredentials`, these properties are not specified in the Lambda Build Image and are read-only, i.e., they cannot be changed after initialization in the Lambda Build Image. So, I don't think I can write unit tests to check for validation errors.
Do you mean to add own class that implements IBuildImage (like the lambda build image classes) and test it?

sakurai-ryo:
@lpizzinidev 
Thanks!
I've made the changes and added comments.
https://github.com/aws/aws-cdk/pull/27934#discussion_r1395149725

packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
```diff
@@ -0,0 +1,83 @@
...
+import { runScriptLinuxBuildSpec } from './private/run-script-linux-build-spec';
+import { BuildEnvironment, ComputeType, IBuildImage, ImagePullPrincipalType, isLambdaComputeType } from './project';
+
+/**
+ * Construction properties of `LinuxArmLambdaBuildImage`.
+ * Module-private, as the constructor of `LinuxArmLambdaBuildImage` is private.
+ */
+interface LinuxArmLambdaBuildImageProps {
+  readonly imageId: string;
+  readonly imagePullPrincipalType?: ImagePullPrincipalType;
```
lpizzinidev, line 11:
```suggestion
```

packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
```diff
@@ -0,0 +1,83 @@
...
+   * @see https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html
+   *
+   * @param id The image identifier
+   * @example 'aws/codebuild/amazonlinux-aarch64-lambda-standard:nodejs18'
+   * @returns A Docker image provided by CodeBuild.
+   */
+  private static fromCodeBuildImageId(id: string): IBuildImage {
+    return new LinuxArmLambdaBuildImage({
+      imageId: id,
+      imagePullPrincipalType: ImagePullPrincipalType.CODEBUILD,
```
lpizzinidev, line 51:
```suggestion
```

packages/aws-cdk-lib/aws-codebuild/lib/linux-lambda-build-image.ts
```diff
@@ -0,0 +1,83 @@
...
+import { runScriptLinuxBuildSpec } from './private/run-script-linux-build-spec';
+import { BuildEnvironment, ComputeType, IBuildImage, ImagePullPrincipalType, isLambdaComputeType } from './project';
+
+/**
+ * Construction properties of `LinuxLambdaBuildImage`.
+ * Module-private, as the constructor of `LinuxLambdaBuildImage` is private.
+ */
+interface LinuxLambdaBuildImageProps {
+  readonly imageId: string;
+  readonly imagePullPrincipalType?: ImagePullPrincipalType;
```
lpizzinidev, line 11:
```suggestion
```

packages/aws-cdk-lib/aws-codebuild/lib/linux-lambda-build-image.ts
```diff
@@ -0,0 +1,83 @@
...
+   * @see https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html
+   *
+   * @param id The image identifier
+   * @example 'aws/codebuild/amazonlinux-x86_64-lambda-standard:nodejs18'
+   * @returns A Docker image provided by CodeBuild.
+   */
+  private static fromCodeBuildImageId(id: string): IBuildImage {
+    return new LinuxLambdaBuildImage({
+      imageId: id,
+      imagePullPrincipalType: ImagePullPrincipalType.CODEBUILD,
```
lpizzinidev, line 51:
```suggestion
```

packages/aws-cdk-lib/aws-codebuild/lib/linux-lambda-build-image.ts
```diff
@@ -0,0 +1,83 @@
...
+    });
+  }
+
+  public readonly type = 'LINUX_LAMBDA_CONTAINER';
+  public readonly defaultComputeType = ComputeType.LAMBDA_1GB;
+  public readonly imageId: string;
+  public readonly imagePullPrincipalType?: ImagePullPrincipalType;
```
lpizzinidev, line 58:
```suggestion
```

packages/aws-cdk-lib/aws-codebuild/lib/linux-lambda-build-image.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  private constructor(props: LinuxLambdaBuildImageProps) {
+    this.imageId = props.imageId;
+    this.imagePullPrincipalType = props.imagePullPrincipalType;
```
lpizzinidev, line 62:
```suggestion
```

packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  public readonly type = 'ARM_LAMBDA_CONTAINER';
+  public readonly defaultComputeType = ComputeType.LAMBDA_1GB;
+  public readonly imageId: string;
+  public readonly imagePullPrincipalType?: ImagePullPrincipalType;
+
+  private constructor(props: LinuxArmLambdaBuildImageProps) {
+    this.imageId = props.imageId;
+    this.imagePullPrincipalType = props.imagePullPrincipalType;
```
lpizzinidev, line 62:
```suggestion
```

packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
```diff
@@ -0,0 +1,83 @@
...
+    return new LinuxArmLambdaBuildImage({
+      imageId: id,
+      imagePullPrincipalType: ImagePullPrincipalType.CODEBUILD,
+    });
+  }
+
+  public readonly type = 'ARM_LAMBDA_CONTAINER';
+  public readonly defaultComputeType = ComputeType.LAMBDA_1GB;
+  public readonly imageId: string;
+  public readonly imagePullPrincipalType?: ImagePullPrincipalType;
```
lpizzinidev, line 58:
```suggestion
```

lpizzinidev (changes requested):
> However, I just rechecked the behavior and found that deploying with imagePullPrincipalType set to CODEBUILD succeeds.

Thanks for clarifying.
We can just leave the logic for handling `imagePullPrincipalType` in `renderEnvironment` (with no code changes).
Just remove `imagePullPrincipalType` from `LinuxArmLambdaBuildImage` and `LinuxLambdaBuildImageProps` since it is redundant and it will be good to go for me.

sakurai-ryo:
@lpizzinidev 
Thank you for your comment!
This modification may be redundant, but the `imagePullPrincipalType` specifies `SERVICE_ROLE` by default, so as I commented before, the deployment will fail as it is.
If I make the modification as per the review comments, integ-tests fail as follows.

スクリーンショット 2023-11-17 14 41 06

If `renderEnvironment` should not be modified, I think it is necessary to specify the `imagePullPrincipalType` explicitly, as in `LinuxBuildImage` and `LinuxArmBuildImage` (for EC2).

lpizzinidev:
@sakurai-ryo 
After re-reviewing this, I think you got it right on the first try and
```
const imagePullPrincipalType = this.isLambdaBuildImage(this.buildImage) ? undefined :
      this.buildImage.imagePullPrincipalType === ImagePullPrincipalType.CODEBUILD
        ? ImagePullPrincipalType.CODEBUILD
        : ImagePullPrincipalType.SERVICE_ROLE;
```
should be the way to go here.
Sorry for the inconvenience and misunderstanding (100% my fault). 
If you could adjust this I will mark it as ready for maintainers to review.
Thanks!

sakurai-ryo:
@lpizzinidev 
Please don’t worry about it.
I appreciate your thorough review and helpful feedback!

Please check it out as I've made changes.

lpizzinidev (approved):

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1718,6 +1753,12 @@ interface LinuxBuildImageProps {
 // eslint-disable-next-line no-duplicate-imports, import/order
 import { LinuxArmBuildImage } from './linux-arm-build-image';
 
+// LinuxArmLambdaBuildImage and LinuxLambdaBuildImage reference ComputeType, so these imports should be placed below ComputeType.
+/* eslint-disable no-duplicate-imports, import/order */
+import { LinuxArmLambdaBuildImage } from './linux-arm-lambda-build-image';
+import { LinuxLambdaBuildImage } from './linux-lambda-build-image';
+/* eslint-enable no-duplicate-imports, import/order */
```
kaizencc, line 1760: What are the duplicate imports?

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1902,8 +1943,14 @@ export class LinuxBuildImage implements IBuildImage {
     this.repository = props.repository;
   }
 
-  public validate(_env: BuildEnvironment): string[] {
-    return [];
+  public validate(env: BuildEnvironment): string[] {
+    const ret = [];
```
kaizencc, line 1947:
```suggestion
    const errors = [];
```

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -2068,6 +2115,15 @@ export class WindowsBuildImage implements IBuildImage {
 
   public validate(buildEnvironment: BuildEnvironment): string[] {
     const ret: string[] = [];
```
kaizencc, line 2117:
```suggestion
    const errors: string[] = [];
```

kaizencc (changes requested):
Thanks for starting this @sakurai-ryo. Looks like Luca has provided some detailed reviews -- that's awesome. I have some preliminary comments, nothing too serious. Will take a deeper look in a bit.

sakurai-ryo:
@kaizencc 
Thanks for your review 👍 
I fixed them.

packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
```diff
@@ -0,0 +1,84 @@
...
+   * @returns A Docker image provided by CodeBuild.
+   */
+  private static fromCodeBuildImageId(id: string): IBuildImage {
+    return new LinuxArmLambdaBuildImage({
+      imageId: id,
+    });
+  }
+
+  public readonly type = 'ARM_LAMBDA_CONTAINER';
+  public readonly defaultComputeType = ComputeType.LAMBDA_1GB;
```
kaizencc, line 54: what is this for / where is this used / why is this the default?

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1715,9 +1750,15 @@ interface LinuxBuildImageProps {
...
 // Keep around to resolve a circular dependency until removing deprecated ARM image constants from LinuxBuildImage
-// eslint-disable-next-line no-duplicate-imports, import/order
+// eslint-disable-next-line import/order
 import { LinuxArmBuildImage } from './linux-arm-build-image';
 
+// LinuxArmLambdaBuildImage and LinuxLambdaBuildImage reference ComputeType, so these imports should be placed below ComputeType.
+/* eslint-disable import/order */
+import { LinuxArmLambdaBuildImage } from './linux-arm-lambda-build-image';
+import { LinuxLambdaBuildImage } from './linux-lambda-build-image';
+/* eslint-enable import/order */
```
kaizencc, line 1760: I don't love this actually. I think the right thing to do here is to pull `ComputeType` out into its own file. Then we can import normally.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -2175,3 +2231,13 @@ export enum ProjectNotificationEvents {
...
 }
+
+export function isLambdaComputeType(computeType: ComputeType): boolean {
+  return [
+    ComputeType.LAMBDA_1GB,
+    ComputeType.LAMBDA_2GB,
+    ComputeType.LAMBDA_4GB,
+    ComputeType.LAMBDA_8GB,
+    ComputeType.LAMBDA_10GB,
+  ].includes(computeType);
```
kaizencc, line 2242: We're going to have to continue to maintain this list every time we add a new Lambda compute type. Ideally we'd have this be an enum-like class but can't change that now. The only thing I can think of here is that instead of maintaining a list, we commit to having the `BUILD_LAMBDA` prefix for all Lambda Compute types and then check for that in this function.

packages/aws-cdk-lib/aws-codebuild/lib/linux-lambda-build-image.ts
```diff
@@ -0,0 +1,84 @@
...
+  public validate(buildEnvironment: BuildEnvironment): string[] {
+    const errors = [];
+
+    if (buildEnvironment.privileged) {
+      errors.push('Lambda compute type does not support privileged mode');
+    }
+
+    if (buildEnvironment.computeType && !isLambdaComputeType(buildEnvironment.computeType)) {
+      errors.push([
+        'Lambda images only support ComputeTypes between',
```
kaizencc, line 70:
```suggestion
        'Lambda images only support Lambda ComputeTypes between',
```

packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
```diff
@@ -0,0 +1,84 @@
...
+  public validate(buildEnvironment: BuildEnvironment): string[] {
+    const errors = [];
+
+    if (buildEnvironment.privileged) {
+      errors.push('Lambda compute type does not support privileged mode');
+    }
+
+    if (buildEnvironment.computeType && !isLambdaComputeType(buildEnvironment.computeType)) {
+      errors.push([
+        'Lambda images only support ComputeTypes between',
```
kaizencc, line 70:
```suggestion
        'Lambda images only support Lambda ComputeTypes between',
```

packages/@aws-cdk-testing/framework-integ/test/aws-codebuild/test/integ.project-linux-lambda-image.lit.ts
kaizencc, line 1: Besides ensuring that this deploys successfully, are there any assertions we can make to confirm everything is running smoothly?

kaizencc (dismissed):
A few final minor comments, and then we're good to go. Thanks @sakurai-ryo

packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
sakurai-ryo: The `defaultComputeType` is required for `IBuildImage` and is used if the user does not specify a `computeType`.
Here is the relevant codes.
https://github.com/aws/aws-cdk/blob/4ab5bb2e9aab4cee4e8a2bcd81871254f2c3065f/packages/aws-cdk-lib/aws-codebuild/lib/project.ts#L1639
https://github.com/aws/aws-cdk/blob/4ab5bb2e9aab4cee4e8a2bcd81871254f2c3065f/packages/aws-cdk-lib/aws-codebuild/lib/project.ts#L1375

packages/@aws-cdk-testing/framework-integ/test/aws-codebuild/test/integ.project-linux-lambda-image.lit.ts
sakurai-ryo: It is indeed required.
Added execution of the `startBuild` API on `awsApiCall`.

sakurai-ryo:
@kaizencc 
Thank you always for your code reviews.
Fixed them and left some comments.

kaizencc (approved):
Thanks for your work @sakurai-ryo!

PR #27938, opened 2023-11-10
fix(elasticloadbalancingv2): logaccesslogs throws error if region is not set using environment

### LINKED ISSUE #27432: elasticloadbalancing: allow logAccessLogs on environment agnostic stack
### Describe the feature

Currently `BaseLoadBalancer.logAccessLogs` requires that the Stack has a specific region specified on the Environment.

This seems a pretty arbitrary limitation - why should this be necessary? Looking at the code, it seems to be in order to set a principal, but other logging constructs like flowLog allow specifying roles, and indeed [if the region doesn't resolve to an account in `BaseLoadBalancer.resourcePolicyPrincipal` it just returns `iam.ServicePrincipal('logdelivery.elasticloadbalancing.amazonaws.com')`](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts#L315), so why shouldn't it do that if the region is unresolved too, as it's only using the region to find an account?

### Use Case

I want to enable access logging on an ALB created in an environment agnostic stack, which seems a reasonable thing to do.

### Proposed Solution

Change https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts#L306-L309
as so:
```
    const region = Stack.of(this).region;
    if (Token.isUnresolved(region)) {
      return new iam.ServicePrincipal('logdelivery.elasticloadbalancing.amazonaws.com');
    }
```

### Other Information

Context - we can't set a region. In our case we have a CDK setup that has been running fine for over a year. A PEN test has flagged that we should have access logs for our ALBs. We have a Product Stack as part of our Stack, and setting a region on the Environment for the Stack completely breaks it; if I set just a region I get this:
```
Error: Stack "my-stack/my-product" cannot reference {my-stack/my-stack-vpc/publicSubnet1/Subnet[Ref]} in stack "my-stack". Cross stack references are only supported for stacks deployed to the same environment or between nested stacks and their parent stack. Set crossRegionReferences=true to enable cross region references
```
(I am setting `crossRegionReferences(true)`)

If I set a region and an account I get this:
```
Resolution error: Cannot generate a physical name for my-stack/my-product/my-stack-my-product-ecs-service-task-def/ExecutionRole, because the region is un-resolved or missing.
```

There seems no way to specify an environment on a product stack.

So currently I'm completely stymied. I guess I'll just have to set up the access logging manually as a handcrafted little snowflake.

### Acknowledgements

### CDK version used

2.99.1

macOs & linux

jakob-lj:
When calling logAccessLogs on an Application Load balancer (v2), there is a check getting the region of the stack. This is later used for getting the account where the stack is deployed. However, if the account is undefined, the service principal (logdelivery.elasticloadbalancing.amazonaws.com) is used. So the region should not be required, as it is only used for finding the account, which is not required. 

Therefore, it should make sense to return the service principal also if the region is undefined, which is described in the issue. 

Closes #27432 .

----


packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts
```diff
@@ -307,7 +307,7 @@ export abstract class BaseLoadBalancer extends Resource {
   protected resourcePolicyPrincipal(): iam.IPrincipal {
     const region = Stack.of(this).region;
     if (Token.isUnresolved(region)) {
-      throw new Error('Region is required to enable ELBv2 access logging');
+      return new iam.ServicePrincipal('logdelivery.elasticloadbalancing.amazonaws.com');
```
tmokmss, line 310: Only newer regions use this principal for elb logging ([doc](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html#attach-bucket-policy)), and other (older) regions require the `elb-account-id` for each region.

So with this change, you _can_ synthesize an environment-agnostic template using elb access logging, but it won't work in those older regions. I guess that's not the desired result. Correct me if I misunderstand something 🙏

jakob-lj: I see. 

Then the proposed solution in the issue wont work. I should have done some more research here before opening PR. Thank you for review @tmokmss . Would you like to inform in the issue as well?

tmokmss: Sure I commented to the original issue, and thank you for working on the cdk issues!

Issue #27929, opened 2023-11-10
aws-cdk-s3: Public access setup as default.


MichalLipski95:
### Describe the bug

When running bootstrap with version 2.106.0 of aws-cdk it appears that s3 cdk bucket is now created by default has set up all public access as enabled and all ACLs are enabled. 
As checked with aws-cli manual, it still says that public access should be disabled by default.
When checked previous version 2.105.0, the behavior was that s3 cdk bucket had blocked public access and object ownership was set up to ACLs disabled. 

### Expected Behavior

S3 cdk bucket should have blocked public access and object ownership set up to ACLs disabled

### Current Behavior

S3 cdk bucket has open public access and object ownership set up to ACLs enabled

### Reproduction Steps

Run cdk bootstrap command with aws-cdk version 2.106.0 which will result with creating a public access enabled bucket and enabled ACLs

### CDK CLI Version

2.106.0

### Node.js Version

10

### OS

Linux

### Language

Java

### Language Version

Java 17

pahud:
I think it has blocked all public access by default.

<img width="517" alt="image" src="https://github.com/aws/aws-cdk/assets/278432/c65cc3ea-4ea8-4f14-b59f-ddfb43f1e3cc">

What made you think it does not? Can you share some screenshots?

scanlonp:
Likely related to #27764.

colifran:
@MichalLipski95 We've tested this with both external and internal accounts and we're seeing that the CDK S3 bucket has public access blocked and object ownership is set up to disable ACLs. Can you share some screenshots of what you're seeing?

mgerlach:
In all our repos using CDK, PR checks for renovate PRs upgrading CDK to 2.106.0 fail in CDK bootstrap because of this. Our AWS org does not allow setting S3 Bucket ACLs causing the build to fail even if the deployment role has the s3:PutBucketAcl permission. 

Please revert to the previous behavior or provide a way to fix. 

#27764 removed a template line setting the deprecated AccessControl property to false. Maybe the deleted line should better have been replaced by an equivalent non-deprecates setting?

mgerlach:
Looks like this was fixed/reverted in #2.106.1. Our PR workflows run successfully with that release.

yt-pooyan:
I just faced this issue for an old project that I just upgraded. (works fine for new projects, though)

[...]

PR #27939, opened 2023-11-10
revert: "chore(cdk): boostrap template uses deprecated AccessControl property"

colifran:
Reverts aws/aws-cdk#27764


vinayak-kukreja (approved):

scarytom:
@colifran why did my change get reverted?

This AccessControl property is deprecated and is being flagged by cfn-lint

mrgrain:
> @colifran why did my change get reverted?
> 
> This AccessControl property is deprecated and is being flagged by cfn-lint

See https://github.com/aws/aws-cdk/issues/27964

The bootstrap stack template has been widly reviewed and assessed. If you need to to comply with further requirements, like e passing and uncomfigured cfn-lint check, I recommend rolling [your own custom version](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html#bootstrapping-customizing) for it. 

Appreciate you are taking the time the review this from a security perspective though.

scarytom:
Thanks for the reply.

I had a look at #27964, and I'm really surprised by that.  The original pull request was reviewed, and it was agreed that this legacy property doesn't do anything.  Seems strange to have a deprecated property that is not actually deprecated.

Thanks for the link to rolling our own template.  We already do that, but I like to try and contribute improvement upstream, which was the case here.

mrgrain:
Found this. I don't have all the details, but looks like it might have be an issue with migrating existing buckets with certain account configurations.

https://github.com/aws/aws-cdk/issues/27929

scarytom:
I guess backwards compatibility with imperfect existing setups is important, so I can understand how we got here, thanks.  I guess we live with it, as there is no easy way to make a breaking change.

PR #27944, opened 2023-11-11
feat(core): option to disable cache for docker build

### LINKED ISSUE #27916: DockerImageAsset: Support --no-cache
### Describe the feature

Currently there's no option to build docker images without caching. The last comment on https://github.com/aws/aws-cdk/issues/20555 explains that this doesn't function as intended.

### Use Case

I have a `pyproject.toml` that requires installing a package from mainline. It currently has no idea if/when mainline has changed and therefore uses cached layers. I need to manually delete local and ECR layers each time to force rebuild.

### Proposed Solution

Add ability to pass args that don't take values.

### Other Information

_No response_

### Acknowledgements

### CDK version used

latest

MacOS

msambol:
Closes #27916.

----


msambol:
This touches a few different packages.. I put `feat(core)` but let me know if that's not right.

msambol:
@lpizzinidev Could I get poke you for a review here? 😄

packages/aws-cdk-lib/core/lib/assets.ts
lpizzinidev, line 1: We could add a test [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/test/assets.test.ts) for this.

packages/aws-cdk-lib/cx-api/lib/assets.ts
lpizzinidev, line 1: We could add a test [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-lambda/test/code.test.ts) for this.

lpizzinidev (changes requested):
Looks good to me 👍
I left some minor suggestions for increasing test coverage.

packages/aws-cdk-lib/core/lib/assets.ts
msambol: added 👍🏼

packages/aws-cdk-lib/cx-api/lib/assets.ts
msambol: added 👍🏼

msambol:
@lpizzinidev Updated with your feedback 👍🏼

lpizzinidev (approved):

kaizencc (approved):
Thanks @msambol

kaizencc (approved):

PR #27948, opened 2023-11-11
fix(elasticloadbalancingv2): access logging fails on imported bucket

### LINKED ISSUE #27928: ApplicationLoadBalancer: failing call to logAccessLogs after version change from 2.105 to 2.106
### Describe the bug

I have a code that is creating ALBs with logAccessLogs enabled. I am passing IBucket to the method which I'm getting by referencing the bucket name with S3.fromBucketName method. In version 2.105 this piece of code has worked without any issues. Only after I switched to 2.106 I started getting issues with this. My assumption is that this is related to commit https://github.com/aws/aws-cdk/commit/1b86634c75623f34856a3578e361d81dae278f86 which fixed issue #27528

### Expected Behavior

Attach bucket for ALB to use to write access logs.

### Current Behavior

Fails with following errors:
```log
File "/codebuild/output/src3432/src/s3/01/lib/alb.py", line 108, in __init__
    self.alb.log_access_logs(
  File "/root/.pyenv/versions/3.11.4/lib/python3.11/site-packages/aws_cdk/aws_elasticloadbalancingv2/__init__.py", line 2389, in log_access_logs
    return typing.cast(None, jsii.invoke(self, "logAccessLogs", [bucket, prefix]))
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.pyenv/versions/3.11.4/lib/python3.11/site-packages/jsii/_kernel/__init__.py", line 149, in wrapped
    return _recursize_dereference(kernel, fn(kernel, *args, **kwargs))
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.pyenv/versions/3.11.4/lib/python3.11/site-packages/jsii/_kernel/__init__.py", line 399, in invoke
    response = self.provider.invoke(
               ^^^^^^^^^^^^^^^^^^^^^
  File "/root/.pyenv/versions/3.11.4/lib/python3.11/site-packages/jsii/_kernel/providers/process.py", line 380, in invoke
    return self._process.send(request, InvokeResponse)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.pyenv/versions/3.11.4/lib/python3.11/site-packages/jsii/_kernel/providers/process.py", line 342, in send
    raise RuntimeError(resp.error) from JavaScriptError(resp.stack)
RuntimeError: TypeError: Cannot read properties of undefined (reading 'shouldSynthesize')
```

### Reproduction Steps

```python
self.alb = elbv2.ApplicationLoadBalancer(
    scope, "ALB",
    vpc=vpc,
    internet_facing=internet_facing,
    vpc_subnets=subnets,
    security_group=security_groups[0],
    load_balancer_name=self.name,
    **kwargs
)
self.alb.log_access_logs(
    aws_s3.Bucket.from_bucket_name(
        scope, "ALBLogsBucket", get_alb_bucket_name(vpc_name)
    ),
    ""
)
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.106.0

### Framework Version

2.106.0

### Node.js Version

18.16.1

### OS

aws-codebuild-docker-images/ubuntu/standard/7.0

### Language

Python

### Language Version

3.11.4

### Other information

_No response_

lpizzinidev:
https://github.com/aws/aws-cdk/pull/27558 fixed a circular dependency issue between the bucket and the ELB when using `logAccessLogs` by adding the dependency at the L1 level.
The dependency was added on both the bucket and the bucket policy (which already refers to the bucket).

This introduced a new bug when adding a dependency for imported buckets (as `bucket.node.defaultChild` is `undefined`).

This PR fixes the issue by adding the dependency at the L1 level on the bucket policy directly.

Closes #27928.

----


packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts
```diff
@@ -252,16 +252,21 @@ export abstract class BaseLoadBalancer extends Resource {
...
+      actions: ['s3:PutObject'],
+      principals: [this.resourcePolicyPrincipal()],
+      resources: [
+        bucket.arnForObjects(`${prefix ? prefix + '/' : ''}AWSLogs/${Stack.of(this).account}/*`),
+      ],
+    });
+    bucket.addToResourcePolicy(putObjectStatement);
+    if (!bucket.policy) {
+      // Imported buckets have `autoCreatePolicy` disabled
+      bucket.policy = new s3.BucketPolicy(bucket, 'Policy', { bucket });
```
tmokmss, line 265: Afaik this may fail if the imported bucket already has its bucket policy configured outside of cfn because there can only be one bucket policy for each bucket. IMO we'd better leave the bucket policy as it is, at least for this PR.

tmokmss (commented):
Thank you for the speedy fix! One comment for concern 🙏

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts
lpizzinidev: Thanks for the review.
Should we provide users some feedback about it or assume they know how to manage bucket policies?

tmokmss: Given that there's been no issue related to that, I think we can assume they know it. Additional documents about that may help, but I found that this behavior is explicitly stated [here](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Bucket.html#policy).

lpizzinidev: Ok, just wanted feedback on this.
I think that the current documentation is clear enough.

tmokmss (approved):
LGTM, thanks!

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts
```diff
@@ -283,9 +281,13 @@ export abstract class BaseLoadBalancer extends Resource {
...
-    // at the L1 level to avoid creating a circular dependency (see https://github.com/aws/aws-cdk/issues/27528)
-    const lb = this.node.defaultChild as CfnLoadBalancer;
-    lb.addDependency(bucket.node.defaultChild as s3.CfnBucket);
+    // at the L1 level to avoid creating a circular dependency (see https://github.com/aws/aws-cdk/issues/27528
+    // and https://github.com/aws/aws-cdk/issues/27928)
+    const lb = this.node.defaultChild;
+    const bucketPolicy = bucket.policy?.node.defaultChild;
+    if (lb && bucketPolicy && CfnResource.isCfnResource(lb) && CfnResource.isCfnResource(bucketPolicy)) {
+      lb.addDependency(bucketPolicy);
+    }
```
kaizencc, line 290: we started off with this (before #27558)

```ts
    this.node.addDependency(bucket);
```

Now, we are (among other changes) only adding the dependency if the policy exists. So I am concerned about the situation where the bucket does not have a policy. In this situation, it seems we used to add `this.node.addDependency(bucket)`, but now skip adding the dependency altogether. 

@lpizzinidev is that something you've considered?

kaizencc (commented):
Hi @lpizzinidev thanks for the quick fix for this issue! Just one comment, and also a general one: I would love if you described the issue you are solving in more detail in the pr description. It is quite hard to have to figure out what the actual problem is before looking into the code changes.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts
lpizzinidev: Thanks for the feedback.

If the bucket does not have a policy, then adding a dependency on it would not create a circular dependency.
Are there cases where we want a dependency on the bucket without the policy?
I assume that, if the bucket does not have a policy attached by CDK, the policy must already exist, otherwise logs would not be delivered to S3.

kaizencc: Gotcha. We were depending on the bucket before but really only cared about depending on the policy. Thats the nuance I was missing.

kaizencc (approved):

PR #27977, opened 2023-11-14
docs(rds): change default value for deleteAutomatedBackups from false to true 

### LINKED ISSUE #27973: (aws_rds): Default value of deleteAutomatedBackups of DBInstance is mismatched from the Cfn and CDK Readme documentation
### Describe the issue

Issue created for the SIM -14222059591

The customer has noticed that there is a discrepancy between the documentation for RDS DB Instances in CDK and CFN resulting in what appears to be unexpected behavior with this resource in regards to the deleteAutomatedBackups property for this resource. It appears that the documentation for this property in CDK is not correct.

In the [CDK docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.DatabaseInstance.html#deleteautomatedbackups), this property is described as the following:

deleteAutomatedBackups?
    Type: boolean (optional, default: false)
    Indicates whether automated backups should be deleted or retained when you delete a DB instance.

And in the [CFN docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-deleteautomatedbackups) for this property of the AWS::RDS::DBInstance resource, it says this:

```
DeleteAutomatedBackups
    A value that indicates whether to remove automated backups immediately after the DB instance is deleted. This parameter isn't case-sensitive. The default is to remove automated backups immediately after the DB instance is deleted.
```

Therefore, as stated in the documentation for this property, the default behavior for this property when not defined is:

    CDK: false, which means the backups would not be deleted
    CFN: true, which means the backups would be deleted

In this case the customer had not defined this property in their CDK stack, and then when they deleted the stack saw that the backups were deleted, which was unexpected as per the CDK docs.

### Links

Default value for deleteAutomatedBackups for DBInstance as per CFN docs - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-deleteautomatedbackups

As mentioned in CDK Documentation - https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.DatabaseInstance.html#deleteautomatedbackups

As mentioned in CDK Code readme -https://github.com/aws/aws-cdk/blob/9d17fe2414b72595802ff94057af5a7efe711268/packages/aws-cdk-lib/aws-rds/lib/props.ts#L96C1-L96C1

tam0ri:
Currently, the default value for deleteAutomatedBackups prop is described as `false` in CDK document.
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.DatabaseInstance.html#deleteautomatedbackups

However, from underlying CloudFormation resource and RDS API perspective, `true` is correct default value.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-deleteautomatedbackups
> A value that indicates whether to remove automated backups immediately after the DB instance is deleted. This parameter isn't case-sensitive. The default is to remove automated backups immediately after the DB instance is deleted.

https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DeleteDBInstance.html
> Specifies whether to remove automated backups immediately after the DB instance is deleted. This parameter isn't case-sensitive. The default is to remove automated backups immediately after the DB instance is deleted.

Also, From CDK implementation perspective, deleteAutomatedBackups prop is simply passed to deleteAutomatedBackups parameter for CfnDBInstance (AWS::RDS::DBInstance).
https://github.com/aws/aws-cdk/blob/5665a959c6628e8786a14863f8d3177b4a1d9f7a/packages/aws-cdk-lib/aws-rds/lib/instance.ts#L856

This PR solves this discrepancy between the documentation for RDS DB Instances in CDK and CFn.

Closes #27973

----

### DIFF
--- a/packages/aws-cdk-lib/aws-rds/lib/instance.ts
+++ b/packages/aws-cdk-lib/aws-rds/lib/instance.ts
@@ -510,7 +510,7 @@ export interface DatabaseInstanceNewProps {
    * Indicates whether automated backups should be deleted or retained when
    * you delete a DB instance.
    *
-   * @default false
+   * @default true
    */
   readonly deleteAutomatedBackups?: boolean;
 


lpizzinidev (approved):
Thanks 👍

mikewrighton (approved):

PR #27983, opened 2023-11-14
fix(integ-runner): fails with "no such file or directory, open 'node_modules/@aws-cdk/integ-runner/lib/workers/db.json.gz'"

mrgrain:
Mark the `@aws-cdk/aws-service-spec` as external, because it cannot be bundled.

Changes an import from `cdk-assets` to be more specific, so we don't bundle the whole of `aws-sdk` anymore.
This reduced the worker entrypoint size from 70MB -> 1.5MB

----

### DIFF
--- a/packages/aws-cdk-lib/aws-rds/lib/instance.ts
+++ b/packages/aws-cdk-lib/aws-rds/lib/instance.ts
@@ -510,7 +510,7 @@ export interface DatabaseInstanceNewProps {
    * Indicates whether automated backups should be deleted or retained when
    * you delete a DB instance.
    *
-   * @default false
+   * @default true
    */
   readonly deleteAutomatedBackups?: boolean;
 


packages/@aws-cdk/integ-runner/lib/runner/private/cloud-assembly.ts
```diff
@@ -1,6 +1,6 @@
 import * as path from 'path';
 import { AssemblyManifest, Manifest, ArtifactType, AwsCloudFormationStackProperties, ArtifactManifest, MetadataEntry, AssetManifestProperties, ArtifactMetadataEntryType, ContainerImageAssetMetadataEntry, FileAssetMetadataEntry } from '@aws-cdk/cloud-assembly-schema';
-import { AssetManifest, FileManifestEntry, DockerImageManifestEntry } from 'cdk-assets';
+import { AssetManifest, FileManifestEntry, DockerImageManifestEntry } from 'cdk-assets/lib/asset-manifest';
```
kaizencc, line 3: nice :)

kaizencc (approved):

packages/@aws-cdk/integ-runner/lib/runner/private/cloud-assembly.ts
mrgrain: It's ugly. But since it's a monorepo we can do it!

AntonioAngelino:
@mrgrain The cli-lib-alpha is still broken for the same reason.

`Deployment failed: Error: ENOENT: no such file or directory, open '/home/xxx/yyyy/node_modules/@aws-cdk/cli-lib-alpha/db.json.gz'`

mrgrain:
Thanks! I know why. Looking into it when I'm back at work next week.

AntonioAngelino:
@mrgrain let me know if I can help. This issue is a blocker for us. Thank you!

mrgrain:
I have the fix ready. Currently making sure this won't happen again. Unfortunately due to re:Invent it won't be released until next week.

AntonioAngelino:
Ouch. Which old version of aws cdk is not affected by the bug? 

We can think of using an older version or a nightly build (if it's possible).

mrgrain:
[v2.106.0](https://github.com/aws/aws-cdk/releases/tag/v2.106.0) should be the one that introduced the issue, so anything before that should still work. There isn't anything that's strictly forcing you to keep `aws-cdk-lib` at the same version, but if you can I would.

We don't have nightly builds at the moment. If you want to hack it, [this package](https://www.npmjs.com/package/@aws-cdk/aws-service-spec) contains a `db.json.gz` file that needs to be copied to `./node_modules/@aws-cdk/cli-lib-alpha/db.json.gz`.

AntonioAngelino:
Ok, thank you. Hacking is fine for now since we're just developing the solution, ty.  👍

PR #27989, opened 2023-11-14
chore: workflow to protect the metadata.ts file against manual changes

otaviomacedo:
This new workflow protects the region-info metadata file, so that only the user `github-actions[bot]` can change it.

----

### DIFF
--- a/.github/workflows/check-region-changes.yml
+++ b/.github/workflows/check-region-changes.yml
@@ -1,0 +1,19 @@
+name: Check Metadata region changes
+
+on:
+  pull_request:
+    branches:
+      - main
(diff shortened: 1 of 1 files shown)


rix0rrr (changes requested):
Did we investigate making this part of the PR linter? I'd rather have it all in once place than spread around a lot of workflow scripts.

otaviomacedo:
> Did we investigate making this part of the PR linter? I'd rather have it all in once place than spread around a lot of workflow scripts.

You're absolutely right. There you go: https://github.com/aws/aws-cdk/pull/28007.

PR #27990, opened 2023-11-14
docs(s3): clarify objectSizeGreaterThan attribute for lifecycle rules

### LINKED ISSUE #27969: S3: object lifecycle rule size ambiguity
### Describe the issue

It's possible to filter what objects a lifecycle rule applies to based on the size of the object using `objectSizeGreaterThan` and `objectSizeLessThan`.

So if I want the rule to apply to objects 10 bytes or bigger I should be able to set `objectSizeGreaterThan: 9`. Objects that are 9 bytes should be left untouched since 10 is greater than 9, but 9 is not greater than 9.

But when I look at the documentation for `objectSizeGreaterThan` it reads
> Specifies the minimum object size in bytes for this rule to apply to.

So if I set it to 9 it would apply to objects of size 9 as well?

I don't know if the attribute should be renamed to `objectSizeGreaterThanOrEqualTo` or if the documentation for the attribute needs to be updated.

### Links

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.LifecycleRule.html#objectsizegreaterthan

msambol:
Docs: [cloudformation s3 bucket rule](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-rule.html).

Closes #27969.

----


scanlonp (dismissed):
Let's also add the corresponding line to `objectSizeLessThan`, which is right above.

scanlonp (approved):
Thanks!

PR #28001, opened 2023-11-15
chore(core): isCfnResource allows any type as input

tmokmss:
It is more or less frustrating that we have to check if a variable is undefined or not before calling the `CfnResource.isCfnResource` method. For example,

```ts
const bucket1 = new Bucket(stack, 'Bucket1');
const bucket1Resource = bucket1.node.defaultChild;
if (bucket1Resource !== undefined &&  // Currently we need this!
    cdk.CfnResource.isCfnResource(bucket1Resource)
) {
    bucket1Resource.addDependency(...);
}
```

With this PR, `isCfnResource` now accepts `any` type as input and performs the necessary validations inside.

```ts
const bucket1 = new Bucket(stack, 'Bucket1');
const bucket1Resource = bucket1.node.defaultChild;
if (cdk.CfnResource.isCfnResource(bucket1Resource)) { // much smoother
    bucket1Resource.addDependency(...);
}
```

Actually, other `isXxx` methods have consistent signatures like the one below:

```ts
public static isStack(x: any): x is Stack
public static isReference(x: any): x is Reference
public static isCfnElement(x: any): x is CfnElement
// and more...
```

This change also makes the `isCfnResource` consistent with these signatures.

Note that this is not a breaking change, because the input constraint is relaxed, not tightened, so all the old code will work without change.

----


packages/aws-cdk-lib/core/lib/cfn-resource.ts
```diff
@@ -34,10 +34,10 @@ export interface CfnResourceProps {
  */
 export class CfnResource extends CfnRefElement {
   /**
-   * Check whether the given construct is a CfnResource
+   * Check whether the given object is a CfnResource
    */
-  public static isCfnResource(construct: IConstruct): construct is CfnResource {
-    return (construct as any).cfnResourceType !== undefined;
+  public static isCfnResource(x: any): x is CfnResource {
+    return x !== null && typeof(x) === 'object' && x.cfnResourceType !== undefined;
```
go-to-k, line 40: Other similar isXxx methods check by `Symbol`, is there any reason not to do the following? I thought it would work as it is more safe and also keeps the consistent signatures.

Because the `cfnResourceType` can be considered unique property?

```ts
const CFN_RESOURCE_SYMBOL = Symbol.for('@aws-cdk/core.CfnResource');
// ...
  public static isCfnResource(x: any): x is CfnResource {
    return x !== null && typeof(x) === 'object' && CFN_RESOURCE_SYMBOL in x;
  }
  // ...
  constructor(scope: Construct, id: string, props: CfnResourceProps) {
    // ...
    Object.defineProperty(this, CFN_RESOURCE_SYMBOL, { value: true });
```

tmokmss: Thanks. I'd like to keep this as it is to prevent from possible breaking change unless there is a strong reason to do so.

go-to-k: I see. Thanks!

go-to-k (approved):

vinayak-kukreja (approved):

vinayak-kukreja (approved):

PR #28012, opened 2023-11-15
fix(s3-notifications): s3 notifications not publishing to sns

### LINKED ISSUE #27994: S3 Event Notification: SNS destination access policy not setting "aws:SourceAccount"
### Describe the bug

Ref: https://github.com/aws/aws-cdk/blob/v2.105.0/packages/aws-cdk-lib/aws-s3-notifications/lib/sns.ts

As per the documentation - https://docs.aws.amazon.com/AmazonS3/latest/userguide/ways-to-add-notification-config-to-bucket.html#step1-create-sns-topic-for-notification

The expected Access policy should include in "Condition", but missing. 

```
"StringEquals": {
    "aws:SourceAccount": "bucket-owner-account-id"
}
```

I believe this is causing S3 object create event to not trigger SNS notification

### Expected Behavior

SNS notification triggered on S3 object creation

### Current Behavior

SNS notification not triggered 

### Reproduction Steps

```
const bucket = new Bucket(this, `some-bucket-name`, {
      bucketName: 'some-bucket',
      encryption: BucketEncryption.S3_MANAGED,
      blockPublicAccess: BlockPublicAccess.BLOCK_ALL,
      enforceSSL: true,
      autoDeleteObjects: true,
    });\

    //Setup SNS
    const topicName = `some-topic-name`
    const topic = new Topic(this, 'some-id', {
      displayName: `some name`,
      topicName: topicName
    });

    //Setup S3 event notification publish to SNS
    bucket.addEventNotification(EventType.OBJECT_CREATED, new SnsDestination(topic), {prefix: 'folder/subfolder/'});
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

v2.105.0

### Framework Version

_No response_

### Node.js Version

v2.105.0

### OS

MacOS

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

msambol:
I tested the integration test with `--no-clean` and it successfully published.

Oddly, when you create the S3 notification via the Console no `Conditions` are created but it still works.
Maybe both are needed if you include one. 😓 

Closes #27994.

----


rix0rrr (changes requested):
> Oddly, when you create the S3 notification via the Console no Conditions are created but it still works.

They are never necessary.

One might argue they prevent a confused deputy attack, but since we already have a condition on `aws:SourceArn`, that confused deputy attack is already prevented.

I'm seeing no reason to merge this. Anyone feel differently?

msambol:
@rix0rrr I think the options are remove the existing condition or add this one. The [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/grant-destinations-permissions-to-s3.html) have both, but for some reason it doesn't work if you just have the one.

rix0rrr:
Hi Mike,

At the risk of repeating myself, `Conditions` never make actions work. The only thing they can do is prevent certain actions that would *otherwise* be allowed from working. They are like **additional filters** to the statement's `Allow` effect.

For example:

```json
{
  "Effect": "Allow",
  "Action": "s3:GetObject",
  "Resource": "*",
  "Principal": "You"
}
```

Means: *"you can read any S3 object"*. 

In contrast:

```json
{
  "Effect": "Allow",
  "Action": "s3:GetObject",
  "Resource": "*",
  "Principal": "You",
  "Condition": {               
    "IpAddress": {"aws:SourceIp": "1.2.3.4"}
   }
}
```

Means: *"you can read any S3 object, but ONLY if you are making API calls from `1.2.3.4`"*.

See how the `Condition` only takes away from the powers that the statement confers? By default, the statement applies *always*, but using `Conditions` you can make the statement only apply *sometimes*. Adding more `Conditions` only makes the statement apply successfully in *fewer* cases.

That's why adding a `Condition` to a statement that's not working, is never going to make it work. 

> Conditions are great, but they're somewhat complicated by the fact that **you don't get to inspect the variables that are used for the evaluation**. That information is not visible or logged anywhere. So if a condition doesn't do what you want, you're going to have to guess what the value of `aws:SourceIp` or `aws:SourceArn` was that the server used in its evaluation of that policy statement. This can be very frustrating! (Ask me how I know)

------------

Now, I believe you if you say the statement is not working (although it's a bit odd that it would. It sure seems like it should have worked at some point, I wonder if something has changed?). Our job is to figure out why the statement is not working as it is.

I see in the documentation that the recommended ARN pattern has `*` in it:

```
            "Condition": {
                "ArnLike": {
                    "aws:SourceArn": "arn:aws:s3:*:*:bucket-name"
                },
                "StringEquals": {
                    "aws:SourceAccount": "bucket-owner-account-id"
                }
            }
```

This looks super odd to me, as S3 ARNs normally don't have anything in those locations blanked out by the `*`s. An [S3 bucket ARN looks like this](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html):

```
arn:aws:s3:::bucket_name
```

As you can see, it doesn't have `region` or `account` fields, which are commonly present in all other ARNs. I wonder if that's the bug? That the ARN that the SNS team uses for `aws:SourceArn` *does* in fact have fields there, but the ARN we're putting into the Condition by means of `{ "Fn::GetAtt": ["MyBucket", "Arn"] }` does NOT, and that's why the Condition is not matching? (Again, extremely complicated by the fact that we can't look at the value of `aws:SourceArn` used in the evaluation of this policy anywhere)

Can you try that? Can you try building an ARN that has the `*` in there, using the `bucket.bucketName` to put the string together yourself (as opposed to using `bucket.bucketArn` which might have the wrong format)

rix0rrr:
(Now of course, all of what I said could easily be invalidated by a single additional `if` in S3's code that *requires* the presence of an `aws:SourceAccount` Condition in the Policy Statement, and refuses to even perform the statement evaluation if that Condition isn't there. AWS is a land of exceptions upon exceptions 🤪 but what I said above applies broadly to 99% of IAM usage, and I would need some solid proof to be convinced otherwise)

msambol:
@rix0rrr – I 100% agree with you: adding a condition should not *make* this work. That said, I tested again with the current code and I can get messages to fire to SNS. When I tried last week, it was not working until I added the code in the PR. It wasn't working for @rkraman-leo as well, who originally filed the issue. @rix0rrr, is it possible to check if there was a bug or issue last week that caused this to break?

@rkraman-leo – Can you verify that you can get messages to publish to SNS with the current code?

rix0rrr:
> I tested again with the current code and I can get messages to fire to SNS

Aha! Is what you're saying, even without the fix in this PR the code magically started working again? I'll create an internal ticket to ask around.

rix0rrr:
Huh. If you have an approximate date and time, and a region you observed this behavior in, that would be helpful

msambol:
@rix0rrr It was in Oregon, ~30 minutes before the PR was opened. 😅

rkraman-leo:
> @rix0rrr – I 100% agree with you: adding a condition should not _make_ this work. That said, I tested again with the current code and I can get messages to fire to SNS. When I tried last week, it was not working until I added the code in the PR. It wasn't working for @rkraman-leo as well, who originally filed the issue. @rix0rrr, is it possible to check if there was a bug or issue last week that caused this to break?
> 
> @rkraman-leo – Can you verify that you can get messages to publish to SNS with the current code?

Apologies for getting back late on this. 
Apparently, for the issues that I was facing, it had nothing to do with the SNS Access Policy. The problem was with S3 prefix having special character ("=") that require special handling. 

Ref: 
* https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-guidelines-special-handling
* https://www.geeksforgeeks.org/html-url-encoding/

After I changed the S3 prefix in the event notification from "business_config_snapshot/schemaVersion=1/" to "business_config_snapshot/schemaVersion%3D1/", it worked for me.

msambol:
@rix0rrr I think this can be closed unless we want to close the loop with the S3 team if there was a glitch in sending notifications to SNS.

PR #28039, opened 2023-11-16
feat(lambda): support for advanced logging

otaviomacedo:
Structured logging enables automated analysis of the logs, thus enables customers to perform quick exploratory and automated monitoring the health of their applications, and helps them troubleshoot their production issues faster. Structured logging is now considered to be a monitoring best practice.

Lambda is developing Advance Logging Controls to give developers and operators more control over their function logs. The feature allows customers to capture logs in JSON format and control the level of details in emitted logs. To ensure compatibility and enable these capabilities, four new fields (LogFormat, ApplicationLogLevel, SystemLogLevel, and LogGroup) will be added to the existing LoggingConfig structure. Also, even adding custom log group for Log group section to have the flexibility of changing the name instead of having it as default. This update empowers customers to define their logging behavior and efficiently manage their function logs, reflecting Lambda's commitment to delivering a user-friendly logging solution that addresses the evolving needs of users.

----


kaizencc (approved):

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -848,6 +943,12 @@ export class Function extends FunctionBase {
       resource.tracingConfig = this.buildTracingConfig(props.tracing ?? Tracing.ACTIVE);
     }
 
+    this.loggingConfig = this.getLoggingConfig(props);
```
kaizencc, line 946: why not just add `loggingConfig` directly to the `CfnLambda` resource getting created a few lines above?

kaizencc (approved):

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -385,6 +455,30 @@ export interface FunctionOptions extends EventInvokeConfigOptions {
...
+   * Sets the log group name for the function.
+   * @default `/aws/lambda/${this.functionName}` default log group name created by Lambda
+   */
+  readonly logGroup?: logs.ILogGroup;
+
+  /**
+   * Sets the logFormat for the function.
+   * @default Text format
+   */
+  readonly logFormat?: string;
```
jogold, line 469: Why is this not typed (enum)? Same for log level.

jonife: haa i see.

TastefulElk: A breaking change can be avoided by setting the type to `LogFormat | keyof typeof LogFormat` which would allow using either the enum or type-safe strings as the value.

Suggested change in #28127 which aims to solve this.

tomoish: Thank you for your suggestion.
if LogFormat type is changed to `logFormat?: LogFormat | keyof typeof LogFormat;`, An error like the following occurs in the code: `logFormat: 'Text'`.
`Type '"Text"' is not assignable to type 'LogFormat | "TEXT" | "JSON" | undefined'. Did you mean '"TEXT"'?ts(2820)`
Would you mind telling me a solution that accepts both `lambda.LogFormat.TEXT` and string `'Text'`.

tomoish: I might solve this problem by myself. I suggest changes in #28127.

PR #28048, opened 2023-11-17
chore(lambda): add AWS_DEFAULT_REGION to the list of reserved environment variables

sakurai-ryo:
### Changes
- added `AWS_DEFAULT_REGION` to `reservedEnvironmentVariables`
  - unit tests were not added because there is only one existing test case written.
    - https://github.com/aws/aws-cdk/blob/91a3e8c0fb0b6e91d45146c44c234151c59b3476/packages/aws-cdk-lib/aws-lambda/test/function.test.ts#L3426 

### Description
The Function constructor checks that reserved environment variables are not specified when adding environment variables.
However, when I mistakenly specified `AWS_DEFAULT_REGION` as an environment variable,  `cdk synth` did not generate an error.

The following document contains the list of reserved environment variables.
https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime

Cloudformation gives the following error.
`Resource handler returned message: "Lambda was unable to configure your environment variables because the environment variables you have provided contains reserved keys that are currently not supported for modification. Reserved keys used in this request: AWS_DEFAULT_REGION`

----


vinayak-kukreja (approved):
Hey thank you for contributing this to the repository. Really appreciate it.

PR #28084, opened 2023-11-21
feat(route53): support for scoping down domain names in IHostedZone.grantDelegation()

### LINKED ISSUE #28078: aws-route53: allow to restrict the domain names in cross-account delegation
### Describe the feature

[IHostedZone] has a convenient [grantDelegation()] method for granting an IAM role the necessary permissions for cross-account delegation. I would like to be able to restrict the names of the `NS` records that the grantee role can `UPSERT` and `DELETE`.

[IHostedZone]: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_route53.IHostedZone.html
[grantDelegation()]: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_route53.IHostedZone.html#grantwbrdelegationgrantee

### Use Case

Take the following infrastructure:

- Root account: `example.com`
- Beta account: `beta.example.com`
- Prod account: `prod.example.com`

I want to make sure that the IAM role assumed from the Beta account can only `UPSERT` and `DELETE` the `NS` record `beta.example.com` and not `prod.example.com`. This way I can be certain that if the Beta account is compromised, there will be no impact to the production domain.

### Proposed Solution

Add a backwards compatible parameter to [grantDelegation()] that would allow a user to optionally add this restriction. For example:

```ts
const parentZone = new route53.PublicHostedZone(this, 'HostedZone', {
  zoneName: 'example.com',
});

const betaCrossAccountRole = new iam.Role(this, 'BetaCrossAccountRole', {
  // ...
});
parentZone.grantDelegation(betaCrossAccountRole, route53.DelegationGrantNames.ofEquals({
  names: ['beta.example.com']
}));

const prodCrossAccountRole = new iam.Role(this, 'ProdCrossAccountRole', {
  // ...
});
parentZone.grantDelegation(prodCrossAccountRole, route53.DelegationGrantNames.ofEquals({
  names: ['prod.example.com']
}));
```

Internally, the grant would add the [route53:ChangeResourceRecordSetsNormalizedRecordNames] condition key with the specified names.

The `DelegationGrantNames` is a new class that allows the user to specify exact name matching (i.e. `ForAllValues:StringEquals`) or pattern matching (i.e. `ForAllValues:StringLike`). It's design would be based on the pattern used by [aws-cdk-lib » aws_dynamodb » Billing]:

For example:

```ts
abstract class DelegationGrantNames {
  public static ofEquals(names: string[]): DelegationGrantNames;
  public static ofLike(names: string[]): DelegationGrantNames;
}
```

[grantDelegation()]: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_route53.IHostedZone.html#grantwbrdelegationgrantee
[route53:ChangeResourceRecordSetsNormalizedRecordNames]: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/specifying-rrset-conditions.html#route53_rrset_ConditionKeys
[aws-cdk-lib » aws_dynamodb » Billing]: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_dynamodb.Billing.html

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.110.0

Apple Silicon / macOS 14.1.1 / nodejs 20.9.0

marcogrcr:
Adds a backwards compatible parameter to `IHostedZone.grantDelegation()` in order to restrict the `NS` records with `UPSERT`/`DELETE` access.

Closes #28078.

----


packages/aws-cdk-lib/aws-route53/lib/delegation-grant-names.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Constraints the delegation grant to a set of domain names using the IAM
+ * `route53:ChangeResourceRecordSetsNormalizedRecordNames` context key.
+ */
+export abstract class DelegationGrantNames {
+  /**
+   * Match the domain names using the IAM `equals` operator.
+   *
+   * @param names The allowed names to match using the IAM `equals` operator.
```
lpizzinidev, line 9:
```suggestion
   * Match the domain names using the IAM `StringEquals` condition.
   *
   * @param names List of allowed record names.
```

packages/aws-cdk-lib/aws-route53/lib/delegation-grant-names.ts
```diff
@@ -0,0 +1,45 @@
...
+      public _equals() {
+        return names;
+      }
+    })();
+  }
+
+  /**
+   * Match the domain names using the IAM `like` operator.
+   *
+   * @param names The allowed names to match using the IAM `like` operator.
```
lpizzinidev, line 22:
```suggestion
   * Match the domain names using the IAM `StringLike` condition.
   *
   * @param names List of allowed record names.
```

packages/aws-cdk-lib/aws-route53/lib/delegation-grant-names.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Constraints the delegation grant to a set of domain names using the IAM
```
lpizzinidev, line 2:
```suggestion
 * Limit the delegation grant to a set of domain names using the IAM
```

packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -225,6 +225,21 @@ new route53.CrossAccountZoneDelegationRecord(this, 'delegate', {
 });
 ```
 
+You can optionally restrict the domain names that can be delegated with the IAM role. This allows you to follow the
+minimum privilege principle:
```
lpizzinidev, line 229:
```suggestion
To restrict the domain names that can be delegated with the IAM role use the `DelegationGrantNames` class, 
which enforces the `route53:ChangeResourceRecordSetsNormalizedRecordNames` condition key.

This allows you to follow the minimum privilege principle:
```

packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -225,6 +225,21 @@ new route53.CrossAccountZoneDelegationRecord(this, 'delegate', {
...
+  zoneName: 'someexample.com',
+});
+
+declare const betaCrossAccountRole: iam.Role;
+parentZone.grantDelegation(betaCrossAccountRole, route53.DelegationGrantNames.ofEquals('beta.someexample.com'));
+
+declare const prodCrossAccountRole: iam.Role;
+parentZone.grantDelegation(prodCrossAccountRole, route53.DelegationGrantNames.ofEquals('prod.someexample.com'));
+```
+
```
lpizzinidev, line 242:
```suggestion

> Visit [Using IAM policy conditions for fine-grained access control to manage resource record sets](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/specifying-rrset-conditions.html) for more details.

```

lpizzinidev (changes requested):
Thanks 👍
The implementation is good for me, I left you some comments for minor documentation improvements.

marcogrcr:
Thanks for the review @lpizzinidev! Your suggestions make sense. I have pushed a new commit with these changes.

lpizzinidev (approved):
Thanks 👍

packages/aws-cdk-lib/aws-route53/lib/delegation-grant-names.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Limit the delegation grant to a set of domain names using the IAM
+ * `route53:ChangeResourceRecordSetsNormalizedRecordNames` context key.
+ */
+export abstract class DelegationGrantNames {
```
kaizencc, line 5: Why do we have to do this pattern? Feels like it would be easier to just go with

```ts
parentZone.grantDelegation(prodCrossAccountRole, {
  nameLike: 'beta.someexample.com',
  // or nameEquals, mutually exclusive
});
```

packages/aws-cdk-lib/aws-route53/lib/hosted-zone-ref.ts
```diff
@@ -36,8 +37,11 @@ export interface IHostedZone extends IResource {
 
   /**
    * Grant permissions to add delegation records to this zone
+   *
+   * @param grantee grantee to receive the permissions
+   * @param names specify to restrict the delegation to a specific set of names
    */
-  grantDelegation(grantee: iam.IGrantable): iam.Grant;
+  grantDelegation(grantee: iam.IGrantable, names?: DelegationGrantNames): iam.Grant;
```
kaizencc, line 44: I think we should use a property bag. We only have once chance to add an optional prop to a public API with backwards compatibility, so lets keep the door open for future additions too. It has the additional benefit of forcing users to supply the prop name with their input, which makes things clearer.

`grantDelegation(grantee: iam.IGrantable, { nameLike?: string, nameEquals?: string }): iam.Grant;`

kaizencc (changes requested):
Some thoughts on implementation. Thanks for getting this started @marcogrcr

packages/aws-cdk-lib/aws-route53/lib/delegation-grant-names.ts
marcogrcr: We don't _have_ to. I just followed what I've observed in other instances of `aws-cdk` (e.g. [aws-cdk-lib » aws_dynamodb » Billing](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_dynamodb.Billing.html)). Thought I could maintain consistency that way. I haven't personally come across this type pattern:

```ts
type GrantDelegationProps =
  {
    readonly nameLike: string
  }
  | {
    readonly nameEquals: string
  };
```

Happy to change it though if you consider it's necessary for getting the PR approved.

packages/aws-cdk-lib/aws-route53/lib/hosted-zone-ref.ts
marcogrcr: Acknowledged. Does this look good to you?

```ts
interface GrantDelegationProps {
  readonly name?: {
    readonly equals: string[];
  } | {
    readonly like: string[];
  };
}
```

marcogrcr:
@kaizencc, can we reopen the pull request? I replied to the change request comments.

packages/aws-cdk-lib/aws-route53/lib/hosted-zone-ref.ts
kaizencc: ideally i want

```ts
interface GrantDelegationProps {
  readonly nameEquals?: string[];
  readonly nameLike?: string[];
}
```

And then somewhere else make sure that `nameEquals` and `nameLike` are not both set at once.

packages/aws-cdk-lib/aws-route53/lib/delegation-grant-names.ts
kaizencc: the pattern you are emulating, enum-like classes, i don't think is too relevant here. @marcogrcr, I responded in the other comment thread

TheRealAmazonKendra: So, I'm not sure that I 100% agree with this assessment. I think that it was good to treat this like an enum like class. We should be enforcing in the contract that you can't input both rather than adding a synth time validation. I think, though, I'd actually take this in a slightly different direction.

Instead of altering the current functions, why don't we create a new function that is `grantScopedDelegation` (or something similar) and have those functions take in `GrantDelegationProps` that look something like:
```
interface GrantDelegationProps {
  readonly grantee: IGrantable;
  readonly scope: DelegationScope;
}
```
`DelegationScope` would then have basically the two functions you created, but I would rename them `nameEquals` and `nameLike`.

The above is assuming that these two inputs really should be mutually exclusive as @kaizencc notes in another comment. If they are not, that changes my assessment here.

TheRealAmazonKendra:
Removing the do-not-close label on this as it's been a while since we heard back and there are now conflicts. 

I've also left a comment with a suggested change. If you would like to keep working on this but the bot closes it, I'm happy to discuss design further in the issue prior to you opening another PR so that you're not redoing the same work over and over again.

PR #28099, opened 2023-11-22
feat(cloudfront): CloudFront Function runtime property

### LINKED ISSUE #28163: aws-cloudfront: Support cloudfront-js-2.0 runtime
### Describe the feature

Cloudfront functions support either "cloudfront-js-1.0" or "cloudfront-js-2.0" runtimes

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/functions-javascript-runtime-features.html

### Use Case

AWS provide example functions here: https://github.com/aws-samples/amazon-cloudfront-functions

Not all of them are cloudfront-js-1.0 v1 compatible; specifically I was using the URL rewrites and redirect example.

### Proposed Solution

Add a new enum property to allow the user to choose which runtime is used.

### Other Information

Available in cloudformation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-function-functionconfig.html

Currently Hardcoded in CDK: https://github.com/aws/aws-cdk/blob/85cc0c686049ca40a0bc09589830a909e911fbdf/packages/aws-cdk-lib/aws-cloudfront/lib/function.ts#L168C19-L168C36

### Acknowledgements

### CDK version used

2.111.0

Linux

watany-dev:
fixed #28163
This provides an option to change the runtime for CloudFront Functions. This makes it easier to access the CloudFront KeyValueStore.

https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-function-functionconfig.html

----


watany-dev:
Sorry, I think it's failing to push the snapshot.

mikewrighton (approved):

go-to-k (changes requested):

packages/@aws-cdk-testing/framework-integ/test/aws-cloudfront/test/integ.distribution-function-runtime.ts
```diff
@@ -0,0 +1,41 @@
...
+const app = new cdk.App();
+const stack = new cdk.Stack(app, 'integ-distribution-function', { env: { region: 'eu-west-1' } });
+
+const cfFunctionRequest = new cloudfront.Function(stack, 'FunctionRequest', {
+  code: cloudfront.FunctionCode.fromInline('function handler(event) { return event.request }'),
+});
+
+const cfFunctionResponse = new cloudfront.Function(stack, 'FunctionResponse', {
+  code: cloudfront.FunctionCode.fromInline('function handler(event) { return event.response }'),
+});
```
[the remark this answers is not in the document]
go-to-k: Sorry, It seems that @mikewrighton previously approved. See this comment if you need me to.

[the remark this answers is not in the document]
watany-dev: What you say is true. I am happy to report that I have changed to a meaningful integ test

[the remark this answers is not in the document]
go-to-k: Thanks!

go-to-k (approved):

packages/aws-cdk-lib/aws-cloudfront/lib/function.ts
```diff
@@ -212,3 +218,19 @@ export interface FunctionAssociation {
   /** The type of event which should invoke the function. */
   readonly eventType: FunctionEventType;
 }
+
+/**
+ * The function's runtime environment version.
+ */
+export enum FunctionRuntime {
```
kaizencc, line 225: I think this would be better as an enum-like class, similar to how our other Rumtime types look.

kaizencc (changes requested):
@watany-dev sorry for the delay in getting to this, and just one minor comment. I think we should have a enum-like runtime class with a public constructor so users can specify a custom runtime if necessary.

packages/aws-cdk-lib/aws-cloudfront/lib/function.ts
```diff
@@ -212,3 +232,31 @@ export interface FunctionAssociation {
   /** The type of event which should invoke the function. */
   readonly eventType: FunctionEventType;
 }
+
+/**
+ * The function's runtime environment version.
+ */
+export class FunctionRuntime {
+
```
kaizencc, line 240:
```suggestion
```

packages/aws-cdk-lib/aws-cloudfront/lib/function.ts
```diff
@@ -212,3 +232,31 @@ export interface FunctionAssociation {
...
+   *
+   * Gives full control over the runtime string fragment.
+   */
+  public static custom(runtimeString: string): FunctionRuntime {
+    return new FunctionRuntime(runtimeString);
+  }
+
+  protected constructor(public readonly value: string) {}
+
+}
```
kaizencc, line 262:
```suggestion
  private constructor(public readonly value: string) {}
}

```

kaizencc (dismissed):
Thanks @watany-dev!

packages/aws-cdk-lib/aws-cloudfront/lib/function.ts
```diff
@@ -100,6 +100,12 @@ export interface FunctionAttributes {
    * The ARN of the function.
    */
   readonly functionArn: string;
+
+  /**
+   * The Runtime of the function.
+   */
+  readonly functionRuntime: string;
```
kaizencc, line 107:
```suggestion
  readonly functionRuntime?: string;
```

Build is failing with:

```
err  - IFACE aws-cdk-lib.aws_cloudfront.FunctionAttributes: newly required property 'functionRuntime' added: input to aws-cdk-lib.aws_cloudfront.Function.fromFunctionAttributes [strengthened:aws-cdk-lib.aws_cloudfront.FunctionAttributes]
```

Looks like we cannot require this prop to `fromFunctionAttributes` and instead will have to default to the original runtime, or throw an error if not provided.

kaizencc: @watany-dev

watany-dev: @kaizencc 
We decided to make it an arbitrary property. If not specified, the Default value, runtime1.0, is specified as before.

kaizencc (approved):

PR #28127, opened 2023-11-24
fix(lambda): update the type of LogFormat option

### LINKED ISSUE #28114: aws-lambda: Log format option is not typed
### Describe the bug

The `logFormat` option in the `FunctionOptions` interface is untyped, meaning that it does not accept the `LogFormat` enum.

See line 469 of https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-lambda/lib/function.ts.

It has also been mentioned in https://github.com/aws/aws-cdk/pull/28039#discussion_r1396306546.

### Expected Behavior

It should accept one of the `LogFormat` enum options.

### Current Behavior

It doesn't accept the `LogFormat` enum so you have to specify a string `"JSON"` or `"TEXT"`, which is not type-safe.

### Reproduction Steps

N/A

### Possible Solution

It should be typed as 
```
readonly logFormat?: LogFormat;
```

_No response_

### CDK CLI Version

2.110.1

### Framework Version

_No response_

### Node.js Version

20.9.0

### OS

MacOS

### Language

Python

### Language Version

3.11.2

### Other information

_No response_

tomoish:
The `logFormat` option in the `FunctionOptions` interface should accept one of the `LogFormat` enum options.

This PR updates the type of the `LogFormat` option to `LogFormat` instead of `string`.

Closes #28114.

***

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license

### DIFF
--- a/packages/aws-cdk-lib/aws-lambda/lib/function.ts
+++ b/packages/aws-cdk-lib/aws-lambda/lib/function.ts
@@ -1085,6 +1085,9 @@ export class Function extends FunctionBase {
    * function and undefined if not.
    */
   private getLoggingConfig(props: FunctionProps): CfnFunction.LoggingConfigProperty | undefined {
+    if (props.logFormat !== undefined && !Object.values(LogFormat).includes(props.logFormat as LogFormat)) {
+      throw new TypeError(`'${props.logFormat}' is not a legal LogFormat type.`);
+    }
     if ((props.applicationLogLevel || props.systemLogLevel) && props.logFormat !== LogFormat.JSON) {
       throw new Error(`To use ApplicationLogLevel and/or SystemLogLevel you must set LogFormat to '${LogFormat.JSON}', got '${props.logFormat}'.`);
     }


packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -466,7 +466,7 @@ export interface FunctionOptions extends EventInvokeConfigOptions {
    * Sets the logFormat for the function.
    * @default Text format
    */
-  readonly logFormat?: string;
+  readonly logFormat?: LogFormat;
```
TastefulElk, line 469: I think this would be a breaking change since it's currently possible (arguably encouraged by the type) to set the property to the strings "JSON" or "Text".

I suggest adding `| keyof LogFormat` which would let the user use either the enum or type-safe strings

```suggestion
  readonly logFormat?: LogFormat | keyof typeof LogFormat;
```

tomoish: Thank you for your comment.
I will modify logFormat type to accept both the enum and type-safe strings

tomoish: Thank you for your suggestion.
Your comment realised me that my first change would be a breaking change.
Both of my first change and your suggestion can't accept both of the enum or type-safe strings.
Therefore, I changed the type of `logFormat` as ``readonly logFormat?: LogFormat | `${LogFormat}`;`` due to accept both of `lambda.LogFormat.TEXT` and string `'Text'`.

scanlonp (dismissed):
I believe we add validation that the text passed in is either 'Text' or 'JSON'? This will prevent random strings from being passed in. We could add it in the `getLoggingConfig` function on 1089-1103 of function.ts. While this would technically be breaking for anyone who passed in a bad value, I would assume any LogFormat value other than those two would cause a deployment failure. This would corral the problem fully.

tomoish:
> I believe we add validation that the text passed in is either 'Text' or 'JSON'? This will prevent random strings from being passed in. We could add it in the `getLoggingConfig` function on 1089-1103 of function.ts. While this would technically be breaking for anyone who passed in a bad value, I would assume any LogFormat value other than those two would cause a deployment failure. This would corral the problem fully.

Thank you for your comment.
I will add a validation.

kaizencc (commented):
I don't have a problem with this as is, since the values in which this would be a breaking change would never have worked. That being said, we must sanity check this in our JSII-target languages to make sure that this doesn't somehow break those users. We _cannot_ merge this in without doing that

Unless @tomoish you are willing to do this I'm mostly adding in this note to myself to test this later.

tomoish:
> I don't have a problem with this as is, since the values in which this would be a breaking change would never have worked. That being said, we must sanity check this in our JSII-target languages to make sure that this doesn't somehow break those users. We _cannot_ merge this in without doing that
> 
> Unless @tomoish you are willing to do this I'm mostly adding in this note to myself to test this later.

Hello, @kaizencc.

Thank you for your comment.

I might not be able to perform the sanity check for the JSII-target languages.
Would you be willing to take on this task?

kaizencc (changes requested):
It looks like in python, the ` readonly logFormat?: LogFormat | `${LogFormat}`;` gets converted to:

```py
log_format: typing.Optional[typing.Union[builtins.str, "LogFormat"]] = None,
```

which is not a breaking change. however, it also negates the impact of this PR since it still permits all strings (and the point of this PR is to restrict these strings to 'TEXT' and 'JSON'. 

We want to do a better job of catering to all our languages, not just typescript. So I don't think I want to go down this path. Instead, we should deprecate `logFormat` and create a new variable, `loggingFormat`, which has the correct type. (If you _hate_ the `loggingFormat` name for some reason, I'll also allow `logFormatV2` or feel free to come up with alternatives).

The deprecation will nag users to move to the new prop, where we will successfully restrict typing for all languages, and problem (not-so-elegantly) solved.

kaizencc:
@tomoish ^

tomoish:
Thank you for your consideration, @kaizencc.

I wasn't familiar with jsii, so I was researching how changing the type of logFormat from``readonly logFormat?: LogFormat | \`${LogFormat}`;`` to `readonly logFormat?: LogFormat | "Text" | "JSON";`, thereby explicitly restricting `logFormat` string to only `'Text'` and `'JSON'`, would be translated in other languages. 

If you have any information on this, I would like to know. Also, if this approach does not resolve the issue, I would accept your proposal.

 In this new `loggingFormat`, are we considering a type like `readonly logFormat?: LogFormat;` that does not allow string?

kaizencc:
> In this new loggingFormat, are we considering a type like readonly logFormat?: LogFormat; that does not allow string?

Correct

> If you have any information on this, I would like to know. Also, if this approach does not resolve the issue, I would accept your proposal.

I'm not too familiar with jsii myself unfortunately, but I believe what happens is that any type in typescript like "TEXT" becomes a string type in python and maybe other languages as well.

tomoish:
Thank you for your response, @kaizencc.
I understand the new arguments and how they are converted to other languages.
I will try to implement `loggingFormat`.

PR #28129, opened 2023-11-24
fix(iam): attaching a policy is not idempotent with imported resources

### LINKED ISSUE #28101: (aws_iam): ManagedPolicy.attachToRole should prevent adding to same role multiple times
### Describe the bug

Currently `attachToRole` checks for object equality.

https://github.com/aws/aws-cdk/blob/70f66c7da3db2529e4346cc2f250d892d83474c7/packages/aws-cdk-lib/aws-iam/lib/managed-policy.ts#L293

This does not work if a Role is imported e.g. via `aws_iam.Role.fromRoleArn` as these are unique objects.

### Expected Behavior

`attachToRole` should skip if a role is already added to the list, so that no `Roles: array items are not unique` is encounted on deployments.

### Current Behavior

`attachToRole` adds the same role (different objects) twice, which results in a validation error.

### Reproduction Steps

1. import a role
2. import the same role at another points
3. create a ManagedPolicy
4. attach both roles to the policy

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.102.0

### Framework Version

_No response_

### Node.js Version

18

### OS

Mac

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

lpizzinidev:
Fixes `attachToUser`, `attachToRole`, and `attachToGroup` for `Policy` and `ManagedPolicy` to use ARNs as a discriminant for resource equality to prevent duplicates on imported resources.

Closes #28101.

----


go-to-k (approved):
Looks good!

sumupitchayan (dismissed):

vinayak-kukreja (approved):

PR #28132, opened 2023-11-25
fix(s3-notifications): multiple notifications doesn't work

corymhall:
When multiple bucket notifications are created it creates a race condition where only the last one processed gets applied. All bucket notifications created in a stack are given the same `stackId` prefix. This prefix is then used to filter out the notification created by the custom resource. If there are other notifications created in the same stack, but not by this custom resource, they get filtered out.

This PR fixes that by filtering the notifications by the specific notification id. This ensures that only the notifications created by the individual custom resource are filter out and the rest (included those created by other custom resources) are marked external.

Note - I had to refactor some of the function code to make it fit the inline size limit. This should probably be rewritten in typescript...

----


kaizencc:
👋! `contribution/core` LOL. We gotta get you your real merit badge.

kaizencc (approved):
I don't have a problem with this and it looks like it fixes a pretty egregious bug. Since it touches a custom resource I wanna get @rix0rrr to take a second look before we send it out to the wild

kaizencc (approved):

kaizencc (approved):
Both @rix0rrr and I agree with you that we gotta rewrite this to typescript, but short of getting a volunteer to do that we're just gonna merge this 🤷. Thanks @corymhall

PR #28138, opened 2023-11-26
feat(ec2): support creating key pairs

### LINKED ISSUE #5252: Add ability to add EC2 key-pair
Please add the ability to create EC2 key pairs via CDK.

### Use Case
I would like not to have to interact with AWS Console at all.
Want my keys git-versioned and manageable from CDK.

### Proposed Solution

```python
key = aws_ec2.KeyPair(
    key_name="autoscaled",
    public_key="ssh-rsa AAAABBBBBCCC...",
)
```

### Other
Couldn't find a way how to do that in documentation. I believe this is not doable with CDK right now.

Seems the only way is to manually add key-pair via web Console and use `key_name` string parameter for the EC2 machine:
https://stackoverflow.com/questions/57572065/how-can-i-access-an-ec2-instance-created-by-cdk

laurelmay:
This adds support for creating `AWS::EC2::KeyPair` resources. These are added as a property to `Instance`, `LaunchTemplate`, and `NatInstance` and the older `keyName` prop is deprecated in favor of the new `keyPair: IKeyPair` property.

A getter is added to retrieve the SSM parameter that hold the private key for non-imported keys and checks are added to make sure that ED25519 keys are not used with a Windows instance.

Closes #5252.

----


AllanOricil:
@kylelaker how can I use it locally before they merge it?

kylelaker:
> @kylelaker how can I use it locally before they merge it?

@AllanOricil The `CfnKeyPair` L1 resource already exists. So you can still use that according to the docs and pass its `ref` attribute in as `keyName`. Something like:

```ts
const keyPair = new ec2.CfnKeyPair(this, 'KeyPair', {
  KeyName: "A-NAME-FOR-KEY",
});

const keyName = keyPair.ref;
```

And you can pass `keyName` into anything that accepts that. Since the key material is put into SSM by CloudFormation, you can still access that using the pattern described in the docs.

AllanOricil:
@kylelaker thank you! That is exactly what I used! I love cdk!

sumupitchayan (approved):

PR #28148, opened 2023-11-27
chore: update snapshot after updating the list of metadata regions

otaviomacedo:
After updating the list of regions, it's also necessary to update the test snapshot. New regions will have the `cdkMetadataResourceAvailable` property flipped from `false` to `true`.

----


rix0rrr (approved):

rix0rrr (changes requested):
Momo has a better idea!

PR #28174, opened 2023-11-28
feat(appconfig): add deploy method to configuration constructs

chenjane-dev:
Adding a deploy method to the configuration construct to allow deployments on a config at any time after creation.

----

### DIFF
--- a/packages/@aws-cdk/aws-appconfig-alpha/test/configuration.test.ts
+++ b/packages/@aws-cdk/aws-appconfig-alpha/test/configuration.test.ts
@@ -149,11 +149,7 @@ describe('configuration', () => {
     Template.fromStack(stack).resourceCountIs('AWS::AppConfig::Deployment', 1);
   });
 
-<<<<<<< HEAD
   test('configuration using deploy method and no environment associated', () => {
-=======
-  test('configuration using deploy method', () => {
->>>>>>> 39c360bbc0 (feat(appconfig): add deploy method to configuration constructs)
     const stack = new cdk.Stack();
     const app = new Application(stack, 'MyAppConfig', {
       name: 'MyApplication',


packages/@aws-cdk/aws-appconfig-alpha/lib/configuration.ts
```diff
@@ -302,6 +302,24 @@ abstract class ConfigurationBase extends Construct implements IConfiguration, IE
     this.extensible.addExtension(extension);
   }
 
+  /**
+   * Deploys the configuration to the specified environment.
+   *
+   * @param environment The environment to deploy the configuration to
+   */
+  public deploy(environment: IEnvironment) {
```
lpizzinidev, line 310:
```suggestion
  public deployTo(environment: IEnvironment) {
```
I think we should use this function to create the deployment in [deployConfigToEnvironments](https://github.com/aws/aws-cdk/blob/78b34accfa1cba88cc412b04df42ea5819c2cf4c/packages/%40aws-cdk/aws-appconfig-alpha/lib/configuration.ts#L322-L331) as well to avoid code repetitions.

packages/@aws-cdk/aws-appconfig-alpha/README.md
```diff
@@ -174,6 +174,20 @@ new appconfig.HostedConfiguration(this, 'MyHostedConfiguration', {
 });
 ```
 
+If you want to deploy a configuration to an environment later and not on creation, you can call the `deploy` method on the configuration.
```
lpizzinidev, line 177:
```suggestion
To deploy a configuration to an environment after initialization use the `deployTo` method:
```
More concise.

packages/@aws-cdk/aws-appconfig-alpha/test/configuration.test.ts
```diff
@@ -149,6 +149,45 @@ describe('configuration', () => {
     Template.fromStack(stack).resourceCountIs('AWS::AppConfig::Deployment', 1);
   });
 
+  test('configuration using deploy method', () => {
```
lpizzinidev, line 152: Can you please add unit tests for:
* Configuration with no env and no deployTo that calls method
* Configuration with env and deployTo that calls method

packages/@aws-cdk/aws-appconfig-alpha/lib/configuration.ts
chenjane-dev: Hm yeah I had this change, but it wouldn't allow this naming convention because the `deployTo` object already exists

lpizzinidev:
@chenjane-dev 
Can you please address https://github.com/aws/aws-cdk/pull/28174#discussion_r1410831413?

packages/@aws-cdk/aws-appconfig-alpha/lib/configuration.ts
chenjane-dev: Didn't see the edit on the comment above. `deployConfigToEnvironments` has different behavior and shouldn't be used here. I can however create another method to reference to avoid the duplicate code.

lpizzinidev (approved):
Thanks 👍

PR #28176, opened 2023-11-28
docs(s3): bad type definition for objectsKeyPattern parameter

paulhcsun:
Improve docstrings to explain that parameter objectsKeyPattern of type any should take in string inputs.

Unable to directly change the parameter type because of backwards compatibility concerns (mentioned in https://github.com/aws/aws-cdk/pull/27486)so we are improving documentation as an alternative solution.

Closes https://github.com/aws/aws-cdk/issues/27481.

----


packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -786,7 +786,7 @@ export abstract class BucketBase extends Resource implements IBucket {
    * If encryption is used, permission to use the key to encrypt the contents
    * of written files will also be granted to the same principal.
    * @param identity The principal
-   * @param objectsKeyPattern Restrict the permission to a certain key pattern (default '*')
+   * @param objectsKeyPattern Restrict the permission to a certain key pattern (default '*'). Parameter type is `any` but `string` should be passed in
```
scanlonp, line 789:
```suggestion
   * @param objectsKeyPattern Restrict the permission to a certain key pattern (default '*'). Parameter type is `any` but `string` should be passed in.
```
nit: this one does not have a period.

scanlonp (dismissed):
1 nit, other than that LGTM! Thanks Paul!

scanlonp (approved):

PR #28180, opened 2023-11-28
feat(core): add support for AccountIdFromAlias #27642

### LINKED ISSUE #27642: (aws-cdk-lib): (Add support for this CfnIntrinsic function -- Fn::AccountIdFromAlias)
### Describe the bug

`CfnInclude` fails to import CFN templates if templates contain the usage of this CFN function -- "Fn:: AccountIdFromAlias".  `CfnInclude` fails with error -- "Error: Unsupported CloudFormation function 'Fn::AccountIdFromAlias". `CfnParser.parseIfCfnIntrinsic` reports `Fn::AccountIdFromAlias` as an "unsupported" CFN function.

Example stacktrace
```
    at CfnParser.parseIfCfnIntrinsic (/Volumes/workplace/starship_cdk/src/AWSSCMStarshipServiceCDK/node_modules/aws-cdk-lib/core/lib/helpers-internal/cfn-parse.js:1:12544)
    at CfnParser.parseValue (/Volumes/workplace/starship_cdk/src/AWSSCMStarshipServiceCDK/node_modules/aws-cdk-lib/core/lib/helpers-internal/cfn-parse.js:1:8679)
    ... <truncated for readability> 
    at CfnInclude.getOrCreateResource (/Volumes/workplace/starship_cdk/src/AWSSCMStarshipServiceCDK/node_modules/aws-cdk-lib/cloudformation-include/lib/cfn-include.js:1:13148)
```

### Expected Behavior

`CfnInclude` can properly import the CFN template that uses `AccountIdFromAlias`.

### Current Behavior

`CfnInclude` fails to import the CFN template that uses `AccountIdFromAlias` with error -- "Error: Unsupported CloudFormation function 'Fn::AccountIdFromAlias".

### Reproduction Steps

- Prepare an CFN template that contains the usage of "AccountIdFromAlias".
- Use `aws-cdk-lib/cloudformation-include` to include the CFN template. 

### Possible Solution

This function needs to be updated to support this CFN function. [source code link](https://github.com/aws/aws-cdk/blob/9c0de51c6ddacc2858ae949590229dd4919cbb61/packages/aws-cdk-lib/core/lib/helpers-internal/cfn-parse.ts#L513C3-L513C50). Please add a new "case" statement to support this CFN function.

```
private parseIfCfnIntrinsic(object: any): any {
    const key = this.looksLikeCfnIntrinsic(object);
    switch (key) {
    case 'Fn::GetAtt': {...}
    case 'Ref':'{...}
    // Need to add a new case for `Fn::AccountIdFromAlias`
...
}
```

_No response_

### CDK CLI Version

2.87.0

### Framework Version

_No response_

### Node.js Version

v18.16.0

### OS

Mac, Amazon Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

YYang30:
**reason for this change:**
- Cloud Formation team introduced the support for referencing an AWS account via a new intrinsic function `Fn::AccountIdFromAlias`. This function can be used in cloud formation templates to reference another account(s) using a conventional alias string. We can get rid of hardcoding AWS account ID(s) in cloud formation templates. So it's helpful for automating region build process. 
- However, `cfn-parse` in the `aws-cdk/core` module has not yet added the support for this new CFN function, as a result of which, `cfn-parse` would fail to recognize this function as a valid/supported CFN function. 

fix #27642 

----

### DIFF
--- a/packages/aws-cdk-lib/core/lib/cfn-fn.ts
+++ b/packages/aws-cdk-lib/core/lib/cfn-fn.ts
@@ -178,6 +178,16 @@ export class Fn {
   }
 
   /**
+   * The intrinsic function ``Fn::AccountIdfromAlias`` returns the AWS account ID that
+   * corresponds to the provided AWS account alias.
+   * @param alias The user-provided AWS account alias string.
+   * @returns a token represented as a string
+   */
+  public static accountIdFromAlias(alias: string): string {
+    return Token.asString(new FnAccountIdFromAlias(alias));
+  }
+
+  /**
    * Given an url, parse the domain name
    * @param url the url to parse
    */
@@ -672,6 +682,21 @@ class FnCidr extends FnBase {
       throw new Error(`Fn::Cidr's count attribute must be between 1 and 256, ${count} was provided.`);
     }
     super('Fn::Cidr', [ipBlock, count, sizeMask]);
+  }
+}
+
+/**
+ * The intrinsic function ``Fn::AccountIdFromAlias`` resolves an AWS account
+ * alias to its account id. This function is typically used to reference another
+ * AWS account within your template.
+ */
+class FnAccountIdFromAlias extends FnBase {
+  /**
+   * Creates an ``FnAccountIdFromAlias`` function.
+   * @param alias The user-provided account alias string.
+   */
+  constructor(alias: string) {
+    super('Fn::AccountIdFromAlias', alias);
   }
 }
 
--- a/packages/aws-cdk-lib/core/lib/helpers-internal/cfn-parse.ts
+++ b/packages/aws-cdk-lib/core/lib/helpers-internal/cfn-parse.ts
@@ -657,6 +657,10 @@ export class CfnParser {
         }
         return { Condition: condition.logicalId };
       }
+      case 'Fn::AccountIdFromAlias': {
+        const value = this.parseValue(object[key]);
+        return Fn.accountIdFromAlias(value);
+      }
       default:
         if (this.options.context === CfnParsingContext.RULES) {
           return this.handleRulesIntrinsic(key, object);
--- a/packages/aws-cdk-lib/core/test/fn.test.ts
+++ b/packages/aws-cdk-lib/core/test/fn.test.ts
@@ -303,6 +303,16 @@ test('Fn.len with resolved value', () => {
   expect(Fn.len(Fn.split('|', 'a|b|c'))).toBe(3);
 });
 
+test('Fn.AccountIdFromAlias', () => {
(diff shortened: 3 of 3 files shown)


YYang30:
Exemption Request

- (justification for requesting exemption on integration test) The test does not require a stack deployment. `cfn-parse` is an internal helper class for validating a cloud formation template. Please feel free to suggest. Will appreciate any guidance/suggestion and be happy to adopt them. 

- (need clarification on doc/readme change). Checked this readme file for aws-cdk-lib/core module [link](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/README.md#intrinsic-functions-and-condition-expressions). Is this the right file/place to use?

YYang30:
`Exemption Request`

(justification for requesting exemption on integration test) The test does not require a stack deployment. cfn-parse is an internal helper class for validating a cloud formation template. Please feel free to suggest. Will appreciate any guidance/suggestion and be happy to adopt them.

(need clarification on doc/readme change). Checked this readme file for aws-cdk-lib/core module [link](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/README.md#intrinsic-functions-and-condition-expressions). Is this the right file/place to use?

scanlonp (changes requested):
This change looks great! 

On the exemption, I do not think you need either an integ test or a readme update.

However, I could not find any documentation on the Fn::AccountIdFromAlias cloudformation function. For example it is not here https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html. Can you provide some documentation on this function?

YYang30:
@scanlonp  Thanks for the review.
Please check this document - https://w.amazon.com/bin/view/AWS/CloudFormation/Teams/SkyTrain/Projects/AccountIdFromAlias/

scanlonp:
Hey @YYang30, that is an internal page that specifically says it is an internal only function. First, this is a public github. We should not be posting internal information here. Second, since it is internal, we will not be including the code here.

Please be careful in the future!

YYang30:
Thanks @scanlonp for the notes. Will keep in mind on document sharing.
Wanted to follow up on this code change. 

- What's risk of including this improvement into core/internal-helper module? All it does is just having `cfn-parse` to be able to recognize this internal as a valid CFN function.  
- Is this an "internal" version of `aws-cdk-lib` or another repo that is dedicated for internal use? 
- What's the workaround to have `cfn-parse` support this? 

It will be greatly appreciated if you could shed some lights on above questions. That would be helpful to unblock our progress. Thank you.

PR #28185, opened 2023-11-29
fix(cli): IAM Policy changes not deploying with --hotswap-fallback

mrgrain:
The hotswappable resource detectors failed to correctly identify `AWS::IAM::Policy` resources as not-hotswappable.

When `--hotswap-fallback` was used and the only change to the stack was with `AWS::IAM::Policy`, this caused the deploy command to first report IAM changes, and then report `no changes` on the stack.

image

----


packages/aws-cdk/test/api/hotswap/s3-bucket-hotswap-deployments.test.ts
```diff
@@ -618,6 +618,26 @@ describe.each([HotswapMode.FALL_BACK, HotswapMode.HOTSWAP_ONLY])('%p mode', (hot
...
+      const deploymentLambda2New = {
+        Type: 'AWS::Lambda::Function',
+        Role: {
+          'Fn::GetAtt': [
+            'ServiceRole2',
+            'Arn',
+          ],
+        },
+      };
+
```
mrgrain, line 640: I believe this existing test was unintentionally relying on undefined behavior.

Previously, the stack under would receive a change that changes `Policy2` from being a policy that is used exclusively by S3Deployment, to a "free-floating" policy. This was because `policy2Old -> policy2New` is changing the role from `ServiceRole -> ServiceRole2`. However `ServiceRole2` is not used by the S3Deployment.

Without adjusting the test case, it now would have failed because this new "free-floating" policy would have been detected as non-hotswappable and also not skippable since it is not connected to an S3Deployment anymore.
I believe the intent of the test was to test the skipping (based on test case description and assertions) and this accidentally worked. So I adjusted the test case.

comcalvi: Good catch!

comcalvi (approved):

PR #28194, opened 2023-11-29
feat(ecs): add instance warmup period prop

### LINKED ISSUE #28190: (aws-ecs): AsgCapacityProvider and AsgCapacityProviderProps should expose InstanceWarmupPeriod
### Describe the feature

`aws-ecs.AsgCapacityProviderProps` should expose the [InstanceWarmupPeriod](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-capacityprovider-managedscaling.html#cfn-ecs-capacityprovider-managedscaling-instancewarmupperiod) property from the L1 provider

### Use Case

Allowing this configuration of the underlying `CfnCapacityProvider` L1 object would allow CDK users defining ECS clusters to effectively lower or raise the expected time for instances to become available, without delving into the L1 object themselves.

### Proposed Solution

_No response_

### Other Information

In the alternate, a user can specify their own L1 `CfnCapacityProvider` instead of using `AsgCapacityProvider`

### Acknowledgements

### CDK version used

v2.99.1

MacOS 14.1.1

msambol:
Closes #28190.

----


packages/aws-cdk-lib/aws-ecs/lib/cluster.ts
```diff
@@ -1187,6 +1187,14 @@ export interface AsgCapacityProviderProps extends AddAutoScalingGroupCapacityOpt
...
    */
   readonly targetCapacityPercent?: number;
+
+  /**
+   * The period of time, in seconds, after a newly launched Amazon EC2 instance
+   * can contribute to CloudWatch metrics for Auto Scaling group.
+   *
+   * @default 300
+   */
+  readonly instanceWarmupPeriod?: number;
```
go-to-k, line 1197: It would be user friendly to include validation requirements in the doc, like [this file](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ecs/lib/drain-hook/instance-drain-hook.ts#L33).

```suggestion
  /**
   * The period of time, in seconds, after a newly launched Amazon EC2 instance
   * can contribute to CloudWatch metrics for Auto Scaling group.
   *
   * Must be between 0 and 10000.
   *
   * @default 300
   */
  readonly instanceWarmupPeriod?: number;
```

packages/aws-cdk-lib/aws-ecs/lib/cluster.ts
```diff
@@ -1244,6 +1252,13 @@ export class AsgCapacityProvider extends Construct {
         throw new Error(`Invalid Capacity Provider Name: ${props.capacityProviderName}, If a name is specified, it cannot start with aws, ecs, or fargate.`);
       }
     }
+
+    if (props.instanceWarmupPeriod) {
```
go-to-k, line 1256: Just in case,It would be good to check whether it is Token.

```suggestion
    if (props.instanceWarmupPeriod && !Token.isUnresolved(props.instanceWarmupPeriod)) {
```

go-to-k (changes requested):
Thanks for this PR!

I made only a few minor comments.

msambol:
@go-to-k Updated!

scanlonp (approved):
This looks great! Good tests, good validation, good in-line docs. Thanks!

PR #28197, opened 2023-11-30
chore(kms): prefer new `aliasArn` to `keyArn` for getting arn of an alias

### LINKED ISSUE #28105: aws-cdk-lib/aws-kms: No available option to get the a Alias ARN
### Describe the bug

Currently, there is no available option to retrieve an Alias ARN within the AWS KMS module of the AWS CDK.

### Expected Behavior

I expected to find a method or property within the AWS KMS construct that allows retrieving the Alias ARN associated with a specific key.

### Current Behavior

There isn't a direct method or property available to retrieve the Alias ARN associated with a KMS key in the AWS CDK AWS KMS module.

### Reproduction Steps

```
const key = new Key(this, "app_key")
const alias = new Alias(this, "app_key_alias", {
    aliasName: "app/key",
    targetKey: key
})
```

> Now there should be a way to get the ARN like: `alias.aliasArn`

### Possible Solution

One potential fix for this issue could be to introduce a method or property within the KMS module that exposes the Alias ARN associated with a key. Like there is for the `kms.Key`

Understanding the Alias ARN is crucial for various operational and management tasks involving KMS keys. Having a direct way to retrieve this information within the CDK AWS KMS module would streamline many use cases.

### CDK CLI Version

2.110.0 (build c6471f2)

### Framework Version

_No response_

### Node.js Version

v20.9.0

### OS

Ubuntu 22.04.3 LTS

### Language

TypeScript

### Language Version

Typescript (3.9.10)

### Other information

image
> The ARN is accessible on the AWS Dashboard.

rafaelrcamargo:
**Motivation:**

The current implementation of `keyArn` within the AWS CDK AWS KMS module returns the Key ARN for a key and an alias, which causes confusion for users expecting the Alias ARN. This PR aims to alleviate this confusion by providing clearer access to the Alias ARN.

**Changes:**

Introducing a new attribute `aliasArn` that mirrors the value from `keyArn` specifically for aliases to explicitly retrieve the Alias ARN. 

```typescript
/**
 * The ARN of the alias.
 *
 * @attribute
 * @deprecated use `aliasArn` instead
 */
public get keyArn(): string {
  return Stack.of(this).formatArn({
    service: 'kms',
    // aliasName already contains the '/'
    resource: this.aliasName,
  });
}

/**
 * The ARN of the alias.
 *
 * @attribute
 */
public get aliasArn(): string {
  return this.keyArn;
}
```

**Query:**

Should we deprecate the existing `keyArn` and mirror it in `aliasArn` or change the logic within `keyArn` to `aliasArn` and use the `keyArn` as the mirror?

> Your feedback on the preferred approach would be greatly appreciated!

Closes #28105.

----


daschaa:
@rafaelrcamargo Thanks for opening up a pull request! I think it's good to add the deprecation warning.
One small remark: Could you also add tests for the `aliasArn` getter? The current test just checks if it is the same value as `keyArn`, but if in the future the keyArn logic is changed, we should still test that the `aliasArn` returns a valid ARN.

rafaelrcamargo:
@daschaa for sure, makes sense. Are there any `utils`/`helpers` or recommended ways to test that in this project?

rafaelrcamargo:
Awesome, followed the current implementation in the L2 construct.

If there's anything else I can help with, let me know

daschaa:
Thanks for the contribution. Looks good to me, so a maintainer can have a look :)

daschaa (approved):

rafaelrcamargo:
Exemption request: Integration test and README update not necessary for this PR. No new features have been introduced; the modification solely involves an existing property, which is already well-documented and thoroughly tested with relevant tests.

daschaa (approved):

packages/aws-cdk-lib/aws-kms/lib/alias.ts
```diff
@@ -67,6 +73,15 @@ abstract class AliasBase extends Resource implements IAlias {
...
   }
 
+  /**
+   * The ARN of the alias.
+   *
+   * @attribute
+   */
+  public get aliasArn(): string {
+    return this.keyArn;
+  }
```
scanlonp, line 83: Can we just copy the logic here? No need to make an additional function call to a now-deprecated function.

scanlonp (changes requested):
Looks good, one nit and should be good to go.

I do not think this needs either integ tests or a readme update.

packages/aws-cdk-lib/aws-kms/lib/alias.ts
rafaelrcamargo: @scanlonp was thinking about this, but would be good to have duplicated code? Or should we move the logic to the `aliasArn` and then call it in `keyArn`?

To be honest both should work and be consistent because of the test (`aliasArn and keyArn from alias should match`), but it's something to think about.

I can get to this as soon as I receive feedback on this.

scanlonp: I think having the same code in both places is best here. Minimally better for performance, and also clearer to understand. 
Also if we ever remove the deprecated function in a major version, there does not need to be any mental effort other than simply removing it.

scanlonp (approved):
Thanks!

PR #28199, opened 2023-11-30
fix(cli-lib): deploy fails with "no such file or directory, open 'node_modules/@aws-cdk/integ-runner/lib/workers/db.json.gz'"

mrgrain:
After #27813 the `deploy` action was broken with the above error. This is effectively the same as #27983 .

To ensure these kind of issues are not slipping through again, the PR is adding a basic testing harness for `cli-lib` to `@aws-cdk-testing/cli-integtests`.

----

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/lib/index.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/index.ts
@@ -3,6 +3,7 @@ export * from './corking';
 export * from './integ-test';
 export * from './memoize';
 export * from './resource-pool';
+export * from './with-cli-lib';
 export * from './with-sam';
 export * from './shell';
 export * from './with-aws';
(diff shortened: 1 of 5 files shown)


packages/@aws-cdk/cli-lib-alpha/package.json
```diff
@@ -65,6 +65,7 @@
       "yarn attribute",
       "mkdir -p ./lib/api/bootstrap/ && cp ../../aws-cdk/lib/api/bootstrap/bootstrap-template.yaml ./lib/api/bootstrap/",
       "cp ../../../node_modules/cdk-from-cfn/index_bg.wasm ./lib/",
+      "cp ../../../node_modules/@aws-cdk/aws-service-spec/db.json.gz ./",
```
mrgrain, line 68: This is the fix. We cannot depend on `@aws-cdk/aws-service-spec` since `cli-lib` is a jsii package, and service-spec is not.

rix0rrr (approved):

PR #28251, opened 2023-12-04
feat: implement code generated handler framework

### LINKED ISSUE #27303: (aws-cdk-lib): Internal Lambda Handler Framework
### Describe the feature

Design and implement a series of constructs, and potentially build tools, to standardize how we build and vend lambda handlers within the aws-cdk.

### Use Case

aws-cdk-lib, and the alpha modules, contain a number of lambda handlers for various cases. Most of these are custom resource handlers that provision or mutate resources during cloudformation deploys, but some of them are handlers designed for various runtime integrations, for example, [triggering an aws sdk call on an event bridge event](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-events-targets/lib/aws-api.ts).

All of these lambda handlers are deployed into user's accounts when they instantiate any of the constructs that rely on their specific functionality. Even though these handlers are deployed to customer accounts, how they perform their various responsibilities are implementation details that ideally, a user does not need to know about. However, there are certain details, mainly the runtime version that these lambda functions use, that should be updated as soon as possible to avoid customers having to deploy lambdas with out of date runtime versions into their own accounts. In order to be able to perform these upgrades more quickly and reliably, we should centralize how these handlers are instantiated and packaged.

### Proposed Solution

Create a series of internal constructs, used only within the aws-cdk libraries, for vending lambda handlers that are deployed to user accounts that allow enforcing best practices. This would include a centralized definition of the "default runtime version". Ideally this default runtime version should be the latest version of NodeJS available across aws regions.

I see these constructs looking something like a `CdkHandler` construct, that essentially wraps `lambda.Code`, but has a few niceties built in. Mainly, it should handle building and packaging all of our lambda handlers including all of their dependencies. This will establish a standard for sharing code across handlers which today requires restructuring code if shared functionality is required. Also, all dependencies should be included by default, with only certain handlers excluding specific dependencies explicitly. Notably, `AwsCustomResource` and other handlers that can't list their dependencies very easily (because it's the entire list of `@aws-sdk/*` modules, or can't bundle them all because of size limitations. Additionally this construct should accept a `compatibleRuntimes` property that allows the explicit opting in of new runtime versions for each one. `CdkHandler` would also have a statically defined constant property `defaultRuntime`, which is used to find the intersection of the latest available default runtime, and the compatible runtimes listed for each handler. If a handler does not have the default runtime in its compatible list, it would use the latest defined in that list. Also, the build should break if the latest version in a handlers compatible runtimes list is marked as deprecated.

Additionally, some basic wrappers around `lambda.SingletonFunction` and `lambda.Function` may be required to accept this type as input, or we may be able to just make `CdkHandler` implement `lambda.Code`.

Example:
```ts
const myHandlerCode = new cdk.CdkHandler(this, 'InternalLambdaHandler', {
  compatibleRuntimes: [lambda.Runtime.NODEJS_16_X, lambda.Runtime.NODEJS_18_X],
  entrypoint: path.resolve(__dirname, 'path-to-my-handler.ts'),
});
```

Additionally, we should add linting to prohibit the usage of types such as `lambda.Code`, `lambda.Runtime, and `aws-cdk-lib.CustomResourceProviderRuntime` directly. This will force authors to use our standard tools for vending lambda handler code within the CDK.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.97.0

MacOs

colifran:
This PR introduces an internal handler framework used to code generate constructs that extend a lambda `Function`, lambda `SingletonFunction`, or core `CustomResourceProvider` construct and prohibit the user from directly configuring the `handler`, `runtime`, `code`, and `codeDirectory` properties.  In doing this, we are able to establish best practices, runtime enforcement, and consistency across all handlers we build and vend within the aws-cdk.

As expected, no integ tests were changed as a result of this PR. To verify that the code generated custom resource providers are working correctly I force ran three integ tests all targeted at an individual custom resource provider:
1. integ.global.ts to test replica provider and the code generated construct extending `Function`
2. integ.bucket-auto-delete-objects.ts to test auto delete objects provider and the code generated construct extending `CustomResourceProvider`
3. integ.aws-api.ts to test aws api provider and the code generated construct `SingletonFunction`

All of these integ tests passed successfully.

Closes #27303

----


packages/aws-cdk-lib/aws-certificatemanager/lib/dns-validated-certificate.ts
```diff
@@ -106,10 +105,7 @@ export class DnsValidatedCertificate extends CertificateBase implements ICertifi
       certificateTransparencyLoggingPreference = props.transparencyLoggingEnabled ? 'ENABLED' : 'DISABLED';
     }
 
-    const requestorFunction = new lambda.Function(this, 'CertificateRequestorFunction', {
-      code: lambda.Code.fromAsset(path.resolve(__dirname, '..', '..', 'custom-resource-handlers', 'dist', 'aws-certificatemanager', 'dns-validated-certificate-handler')),
-      handler: 'index.certificateRequestHandler',
-      runtime: lambda.Runtime.NODEJS_18_X,
+    const requestorFunction = new CertificateRequestCertificateRequestProvider(this, 'CertificateRequestorFunction', {
       timeout: cdk.Duration.minutes(15),
```
rix0rrr, line 109: Let's have the `timeout` default to 15 minutes.

I don't think there's a good reason to set it shorter: we trust the code, and we'd rather the code finish (for whatever reason it's taking a long time), than accidentally timeout after 3 seconds or 1 minute or what have you, and us failing the deployment.

packages/aws-cdk-lib/aws-cloudfront/lib/experimental/edge-function.ts
```diff
@@ -199,9 +197,7 @@ export class EdgeFunction extends Resource implements lambda.IVersion {
     });
 
     const resourceType = 'Custom::CrossRegionStringParameterReader';
-    const serviceToken = CustomResourceProvider.getOrCreate(this, resourceType, {
-      codeDirectory: path.join(__dirname, '..', '..', '..', 'custom-resource-handlers', 'dist', 'aws-cloudfront', 'edge-function'),
-      runtime: CustomResourceProviderRuntime.NODEJS_18_X,
+    const serviceToken = CrossRegionStringParamReaderProvider.getOrCreate(this, resourceType, {
```
rix0rrr, line 200: Why 

```ts
const requestorFunction = new CertificateRequestCertificateRequestProvider(this, 'CertificateRequestorFunction', {
```

For the previous CR, but `getOrCreate` for this one?

Wouldn't it make sense to have the same API all the way around?

packages/aws-cdk-lib/aws-eks/lib/cluster-resource-provider.ts
```diff
@@ -66,15 +64,12 @@ export class ClusterResourceProvider extends NestedStack {
...
-      code: lambda.Code.fromAsset(HANDLER_DIR),
+    const onEvent = new ClusterResourceOnEventProvider(this, 'OnEventHandler', {
       description: 'onEvent handler for EKS cluster resource provider',
-      runtime: lambda.Runtime.NODEJS_18_X,
       environment: {
         AWS_STS_REGIONAL_ENDPOINTS: 'regional',
         ...props.environment,
       },
-      handler: 'index.onEvent',
       timeout: Duration.minutes(1),
```
rix0rrr, line 73: Oof! Risky!

packages/aws-cdk-lib/aws-events-targets/lib/aws-api.ts
```diff
@@ -81,12 +81,9 @@ export class AwsApi implements events.IRuleTarget {
    * result from an EventBridge event.
    */
   public bind(rule: events.IRule, id?: string): events.RuleTargetConfig {
-    const handler = new lambda.SingletonFunction(rule as events.Rule, `${rule.node.id}${id}Handler`, {
-      code: lambda.Code.fromAsset(path.join(__dirname, '..', '..', 'custom-resource-handlers', 'dist', 'aws-events-targets', 'aws-api-handler')),
-      runtime: lambda.Runtime.NODEJS_18_X,
+    const handler = new AwsApiProvider(rule as events.Rule, `${rule.node.id}${id}Handler`, {
```
rix0rrr, line 84: This is not actually a custom resource though (so it's also not really a "provider"). I'm not opposed to including it, but aren't we stretching naming conventions here a bit?

packages/aws-cdk-lib/aws-lambda/lib/runtime.ts
```diff
@@ -56,43 +62,64 @@ export class Runtime {
...
+  });
 
   /**
    * The NodeJS 8.10 runtime (nodejs8.10)
    * @deprecated Legacy runtime no longer supported by AWS Lambda. Migrate to the latest NodeJS runtime.
    */
-  public static readonly NODEJS_8_10 = new Runtime('nodejs8.10', RuntimeFamily.NODEJS, { supportsInlineCode: true });
+  public static readonly NODEJS_8_10 = new Runtime('nodejs8.10', RuntimeFamily.NODEJS, {
+    supportsInlineCode: true,
+    isDeprecated: true,
```
rix0rrr, line 94: If you're going to add this, also add a check to `Function` to add a `Annotations.of(this).addWarningV2(...)` that tells people the runtime is deprecated.

packages/aws-cdk-lib/core/lib/custom-resource-provider/custom-resource-provider-base.ts
rix0rrr, line 1: If this class is private, please put it in a folder named `private` or start its file name with a `_`.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/runtime.ts
```diff
@@ -0,0 +1,133 @@
...
+   *
+   * @default false
+   */
+  readonly isDeprecated?: boolean;
+}
+
+/**
+ * Custom resource framework runtimes used for code generation.
+ */
+export class Runtime {
```
rix0rrr, line 23: Oof. This looks like a straight up copy and paste from the lambda module. That's a bit risky. Why do we need this?

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/utils/runtime-determiner.ts
```diff
@@ -0,0 +1,109 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import { Runtime, RuntimeFamily } from '../runtime';
+
+/**
+ * A utility class used to determine the latest runtime for a specific runtime family
+ */
+export class RuntimeDeterminer {
```
rix0rrr, line 7: We are *specifying* the latest runtime now, are we not? It's just always NODE_18 or whatever? (Which we *know* is not deprecated?) So why do we still need this?

packages/@aws-cdk/custom-resource-handlers/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "version": "0.0.0",
   "scripts": {
-    "build": "tsc -b && node scripts/minify-and-bundle-sources.js",
+    "build": "tsc -b && node scripts/generate.js",
```
rix0rrr, line 7: Personally I would `ts-node` the `.ts` file directly.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/config.ts
```diff
@@ -0,0 +1,363 @@
...
+
+export type CustomResourceFrameworkConfig = { [module: string]: { [identifier: string]: ProviderProps[] } };
+
+export const config: CustomResourceFrameworkConfig = {
+  'aws-amplify-alpha': {
+    'asset-deployment-handler': [
+      {
+        type: ProviderType.CDK_NO_OP,
+        sourceCode: path.resolve(__dirname, '..', 'aws-amplify-alpha', 'asset-deployment-handler', 'index.ts'),
+        compatibleRuntimes: [Runtime.NODEJS_18_X],
```
rix0rrr, line 75: Why is this a list? We don't need to do content negotiation, do we? We just stipulate: this Lambda will be using this version. Bam.

Put another way: I see how a list of compatible runtimes makes sense if we say "we support X, Y and Z" and we leave the choice up to the user. But we *don't* leave the choice up to the user!

Extract it to a constant as well.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/config.ts
```diff
@@ -0,0 +1,363 @@
...
+   */
+  CDK_NO_OP = 'CdkNoOp',
+}
+
+/**
+ * Properites used to generate individual custom resource providers.
+ */
+export interface ProviderProps {
+  /**
+   * The custom resource provider type to generate.
```
rix0rrr, line 37: You need to explain the rules of thumb that go into this, otherwise the person after you is going to be mighty confused.

How would you choose between the 4 different alternatives?

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/README.md
```diff
@@ -0,0 +1,87 @@
+# AWS CDK Custom Resource Framework
+
+The CDK custom resource framework is an internal framework developed to establish best practices, runtime enforcement, and consistency for custom resource providers and their associated source code. 
+
+## Custom Resource Framework Concepts
+
+Custom resource providers can be created in one of three forms - `CdkFunction`, `CdkSingletonFunction`, or `CdkCustomResourceProvider`. These three custom resource provider formats are a code generated wrapper around the lambda `Function`, lambda `SingletonFunction`, and `CustomResourceProvider` constructs, respectively. These new CDK prefixed constructs will offer the same behavior and initialization options as their parent minus the ability to configure the following properties during construction:
```
rix0rrr, line 7: This framing around `CdkFunction`, `CdkSingletonFunction` etc is confusing me a little.

As far as I can tell, those exact classes are gone (which is great!) so now these are just **concepts**. That's fine, but as a concept a `CdkFunction` doesn't really mean anything to me.

Can we not give these names after what you would use them for:

* `CdkFunction` -> `BuiltinLambdaFunction` (because I think these are never used as custom resource providers?)
* `SingletonFunction` -> actually confusing to me. Are these usually/always used for custom resource providers? Or no? Are they also built-in Lambda functions that are not actually used as custom resource providers? If true, the naming of this section is confusing, because you're heavily stating it's all about custom resources everywhere, but it might not actually be? Maybe it's just non-Custom Resource Lambdas we provision for you?
* `CdkCustomResourceProvider` -> this should make a distinction between the `core.CustomResourceProvider` and the `custom-resources.CustomResourceProvider`. Maybe name them something like `CoreCustomResource` and `LongStabilizationCustomResource`?

Maybe the flavors of Lambdas we generate here are:

* `RuntimeLambda({ singleton: boolean })`
* `CoreCustomResource`
* `LongStabilizationCustomResource({ onEvent, isComplete })`

rix0rrr: (I'm a little worried we're only adding more layers of indirection, and it looks like we're only adding complexity instead of taking it away...)

rix0rrr: (If `CdkFunction`, `CdkSingletonFunction` etc are still in from the previous commit, then I think we can take them out again, and we can just codegen their implementations)

rix0rrr (dismissed):
Nice! I like where this is going! I think I'd like to see some changes to the level of modeling, and the implementation of the classes we generate.

1. I'd rather see `CdkFunction` etc out of the code base again. Those classes exist because we like to only type code once; well now everything that's common between Lambdas can just go into the codegen.
2. I'd like us to be very clear on when a Lambda is actually a Custom Resource and when it's not. The two classes of "Lambda we bundle" are tossed onto a giant heap here, but I think that's more confusing than illuminating.

On the one hand this framework feels like it's targeting "Lambdas" more than "Custom Resources":

* It's not abstracting the concerns of custom resources away. Instead, it's abstracting the concerns away of "where does the source and runtime of my Lambda come from. How you use those Lambdas is up to you".
* But then on the other hand it's claiming to be a "Custom Resource framework".

And I'm not seeing a clear distinction between `core.CustomResourceProvider` and `custom-resources.Provider` -- both of which exist, and both of which we should target.

I think what's happening is: this framework helps us generate "Lambdas", then the user still needs to use those Lambdas when they set up the `custom-resources.Provider`. But why doesn't this framework create the `custom-resources.Provider` for me? 

----------------

As a matter of fact, why do I still need to think in terms of:

```ts
const provider = new GeneratedProvider(this, 'Provider');
new core.CustomResource(this, 'CR', {
  serviceToken: provider.lambdaArn,
  
  foo: 'foo',
  bar: 'bar',
});
```

That's still showing too much of the internals. Why don't we do:

```ts
new GeneratedCustomResource(this, 'CR', {
  providerScope: this,  // optional?
  providerId: 'Provider',

  foo: 'foo',
  bar: 'bar',  
});
```

Or 

```ts
const provider = new GeneratedProvider(this, 'Provider');
provider.invoke(this, 'CR', {
  foo: 'foo',
  bar: 'bar',  
});
```

Why have the innards poke out? 

(Not the biggest of deals, but there's so much more we could be doing to hide the gory details)

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/README.md
```diff
@@ -0,0 +1,87 @@
...
+        compatibleRuntimes: [Runtime.PYTHON_3_9],
+        // prevent minify and bundle flag is set since the source code is a python file
+        preventMinifyAndBundle: true,
+      },
+    ],
+  },
+};
+```
+
+Code generation for the provider modules is triggered when this package - `@aws-cdk/custom-resource-handlers` - is built. Importantly, this framework is also responsible for minifying and bundling the custom resource providers' source code and dependencies. A flag named `preventMinifyAndBundle` can be configured as part of the `ProviderProps` to prevent minifying and bundling the source code for a specific provider. This flag is `false` by default and is only needed for Python files or for JavaScript/TypeScript files containing require imports.
```
mrgrain, line 61: nit: Swap this around to be the positive: `minifyAndBundle: true` is the default and it can be set to `false`.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/README.md
```diff
@@ -0,0 +1,87 @@
...
+|   |   |   |   |--- replica-provider.generated.ts
+|   |   |   |--- aws-stepfunctions-tasks
+|   |   |   |   |--- eval-nodejs-handler
+|   |   |   |   |   |--- index.js
+|   |   |   |   |--- role-policy-handler
+|   |   |   |   |   |--- index.py
+|   |   |   |   |--- eval-nodejs-provider.generated.ts
+|   |   |   |   |--- role-policy-provider.generated.ts
+
+The code generated custom resource providers are usable in `aws-cdk-lib` once `aws-cdk-lib` is built. The custom resource providers will be consumable from `aws-cdk-lib/custom-resource-handlers/dist` and the file structure therein will match what was generated in `@aws-cdk/custom-resource-handlers/dist` except for providers defined in `core`. To prevent circular dependencies, all custom resource providers defined in `core`and any associated source code will be consumable from `aws-cdk-lib/core/dist/core`.
```
mrgrain, line 80: `aws-cdk-lib/core/dist/core` as a path is a bit confusing. Can it be something like `aws-cdk-lib/core/dist/custom-resource-handlers`

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/config.ts
mrgrain: Looks like we use the same function with different handlers. But I agree, we can make `handler` a list instead.

packages/aws-cdk-lib/aws-dynamodb/lib/replica-provider.ts
```diff
@@ -55,21 +55,13 @@ export class ReplicaProvider extends NestedStack {
...
     super(scope, id);
 
-    const code = lambda.Code.fromAsset(path.join(__dirname, '..', '..', 'custom-resource-handlers', 'dist', 'aws-dynamodb', 'replica-handler'));
-
     // Issues UpdateTable API calls
-    this.onEventHandler = new lambda.Function(this, 'OnEventHandler', {
-      code,
-      runtime: lambda.Runtime.NODEJS_18_X,
-      handler: 'index.onEventHandler',
+    this.onEventHandler = new ReplicaOnEventProvider(this, 'OnEventHandler', {
```
mrgrain, line 59: This used to be a function and now it's a provider?

packages/aws-cdk-lib/aws-lambda/lib/runtime.ts
```diff
@@ -30,6 +30,12 @@ export interface LambdaRuntimeProps {
    * @default false
    */
   readonly supportsSnapStart?: boolean;
+
+  /**
+   * Whether this runtime is deprecated.
```
mrgrain, line 35:
```suggestion
   * Whether this runtime has been deprecated on Lambda.
   * Notably, a particular public version of a runtime version might be EOL, but is still supported on Lambda.
```

mrgrain: Just because someone will get confused by this.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/framework.ts
```diff
@@ -0,0 +1,121 @@
...
+    const name = this.buildProviderName(handler);
+
+    const props: CdkCustomResourceClassProps = {
+      name,
+      handler,
+      codeDirectory,
+      runtime: RuntimeDeterminer.determineLatestRuntime(
+        provider.compatibleRuntimes,
+        CdkCustomResourceModule.DEFAULT_RUNTIME,
+      ),
```
mrgrain, line 57: I would really like this code to be handled at runtime, not at codegen time. Because the next step (out of scope for this PR) would be to add Node20 as a compatible runtime everywhere, but keep the "preferred" runtime on Node18 (since CN and GovCloud don't have Node20 yet). Then we can change the `RuntimeDeterminer` to take a custom fact into account, in case someone wants to run them with node20 already.

I'm concerned that if we don't this check at runtime now, we back ourselves into a corner we cannot get out anymore. I might be wrong though. Thoughts?

mrgrain (commented):
Like @rix0rrr my biggest remaining concern is the naming and intertwining of different concepts. I am ware that one goal of this was to use `Cdk*` prefixed classes everywhere, but what do we really gain from this?

Especially the complexity around picking a "provider type" seems scary. Although I realize this issue has been caused by the current implementation that used to focus on `codeDirectory` as the sole differentiator.

MrArnoldPalmer (approved):
:shipit: Thanks @colifran 

As discussed with @mrgrain and @rix0rrr offline, there are a couple things we want to continue to pursue here.
1. On a case by case basis, move everything possible to generate `CustomResourceProvider` and deprecate supporting for `Function` and `SingletonFunction`. I believe we should still support `Function` as a general case in the future, but the only handler that currently is NOT  a custom resource is the aws-events `AwsApi` handler.
2. Once everything is using a common base, reevaluate the abstraction of `CustomResourceProvider` for our use case more generally. Since we are code generating, there is little cost to us outputting constructs that extend `CfnResource` which would remove all of the concerns around circular dependencies that exist with some of these handlers.
3. Separately pursue moving every handler to use `NODEJS_LATEST`, and make the runtime region availability aware. This would mean making sure all of our handlers are runtime version portable, bundling all of their assets. This PR gives us the ability to change the default runtime in one place and have every handler affected, as well as know where handlers are not using the default requiring separate upgrading and testing. However more consolidation here would mean even faster and more reliable runtime upgrades.

portswigger-tim:
Hmmm.... Did this change break utilities like `CfnJson`?

```sh
jsii.errors.JavaScriptError: 
  Error: Cannot find module '../dist/core/cfn-utils-provider.generated'
  Require stack:
  - /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/core/lib/private/cfn-utils-provider.js
  - /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/core/lib/cfn-json.js
  - /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/core/lib/index.js
  - /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/core/index.js
  - /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/aws-lambda/lib/function-base.js
  - /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/aws-lambda/lib/lambda-augmentations.generated.js
  - /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/aws-lambda/lib/index.js
  - /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/aws-lambda/index.js
  - /tmp/jsii-kernel-JhJ26s/node_modules/@aws-cdk/lambda-layer-kubectl-v28/lib/kubectl-layer.js
  - /tmp/jsii-kernel-JhJ26s/node_modules/@aws-cdk/lambda-layer-kubectl-v28/lib/index.js
  - /tmp/jsii-kernel-JhJ26s
      at Module._resolveFilename (node:internal/modules/cjs/loader:1134:15)
      at Module._load (node:internal/modules/cjs/loader:975:27)
      at Module.require (node:internal/modules/cjs/loader:12[25](https://github.com/portswigger-cloud/scanner/actions/runs/7298299094/job/19889011039#step:10:26):19)
      at require (node:internal/modules/helpers:177:18)
      at cfn_utils_provider_generated_1 (/tmp/jsii-kernel-JhJ[26](https://github.com/portswigger-cloud/scanner/actions/runs/7298299094/job/19889011039#step:10:27)s/node_modules/aws-cdk-lib/core/lib/private/cfn-utils-provider.js:1:436)
      at CfnUtilsProvider.getOrCreate (/tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/core/lib/private/cfn-utils-provider.js:1:629)
      at new CfnJson (/tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/core/lib/cfn-json.js:1:1371)
      at Kernel._Kernel_create (/tmp/tmp8oq0hyj0/lib/program.js:10108:25)
      at Kernel.create (/tmp/tmp8oq0hyj0/lib/program.js:9779:93)
      at KernelHost.processRequest (/tmp/tmp8oq0hyj0/lib/program.js:11699:36)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/runner/work/scanner/scanner/infra/app.py", line 69, in <module>
    eks_cluster_stack = EksClusterStack(
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/jsii/_runtime.py", line 118, in __call__
    inst = super(JSIIMeta, cast(JSIIMeta, cls)).__call__(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/ps_cdk/stacks/eks_cluster_stack.py", line 67, in __init__
    self.karpenter_role = self.create_karpenter_role(cluster_name)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/ps_cdk/stacks/eks_cluster_stack.py", line 565, in create_karpenter_role
    role = self.create_irsa_role(
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/ps_cdk/stacks/eks_cluster_stack.py", line 413, in create_irsa_role
    return ps_iam.IrsaRole(
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/jsii/_runtime.py", line 118, in __call__
    inst = super(JSIIMeta, cast(JSIIMeta, cls)).__call__(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/ps_cdk/constructs/iam/irsa_role.py", line 21, in __init__
    "StringEquals": CfnJson(
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/jsii/_runtime.py", line 118, in __call__
    inst = super(JSIIMeta, cast(JSIIMeta, cls)).__call__(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/aws_cdk/__init__.py", line [27](https://github.com/portswigger-cloud/scanner/actions/runs/7298299094/job/19889011039#step:10:28)574, in __init__
    jsii.create(self.__class__, self, [scope, id, props])
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/jsii/_kernel/__init__.py", line 334, in create
    response = self.provider.create(
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/jsii/_kernel/providers/process.py", line 365, in create
    return self._process.send(request, CreateResponse)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/jsii/_kernel/providers/process.py", line 342, in send
    raise RuntimeError(resp.error) from JavaScriptError(resp.stack)
RuntimeError: Cannot find module '../dist/core/cfn-utils-provider.generated'
Require stack:
- /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/core/lib/private/cfn-utils-provider.js
- /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/core/lib/cfn-json.js
- /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/core/lib/index.js
- /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/core/index.js
- /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/aws-lambda/lib/function-base.js
- /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/aws-lambda/lib/lambda-augmentations.generated.js
- /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/aws-lambda/lib/index.js
- /tmp/jsii-kernel-JhJ26s/node_modules/aws-cdk-lib/aws-lambda/index.js
- /tmp/jsii-kernel-JhJ26s/node_modules/@aws-cdk/lambda-layer-kubectl-v[28](https://github.com/portswigger-cloud/scanner/actions/runs/7298299094/job/19889011039#step:10:29)/lib/kubectl-layer.js
- /tmp/jsii-kernel-JhJ26s/node_modules/@aws-cdk/lambda-layer-kubectl-v28/lib/index.js
- /tmp/jsii-kernel-JhJ26s
```

mrgrain:
@portswigger-tim Thanks for the report. Any chance you could post a code snippet how you use `CfnJson` for me to reprdoduce?

portswigger-tim:
Sure @mrgrain :

```python
import aws_cdk.aws_iam as iam
from aws_cdk import CfnJson


class IrsaRole(iam.Role):
    def __init__(
        self,
        scope: Construct,
        id: str,
        eks_cluster_oidc_provider_arn: str,
        eks_cluster_oidc_provider_issuer: str,
        namespace: str,
        service_account_name: str,
        **kwargs,
    ) -> None:
        assumed_by = iam.FederatedPrincipal(
            eks_cluster_oidc_provider_arn,
            assume_role_action="sts:AssumeRoleWithWebIdentity",
            conditions={
                "StringEquals": CfnJson(
                    scope,
                    f"{id}Conditions",
                    value={
                        f"{eks_cluster_oidc_provider_issuer}:aud": "sts.amazonaws.com",
                        f"{eks_cluster_oidc_provider_issuer}:sub": f"system:serviceaccount:{namespace}:{service_account_name}",
                    },
                )
            },
        )

        super().__init__(
            scope,
            id,
            assumed_by=assumed_by,
            description=f"IAM role for ServiceAccount: {service_account_name}",
            **kwargs,
        )
 ```
 
 I think that this was necessary due to the tokenization of outputs in CDK when used like:
 
 ```python
 
    # pre-canned IRSA roles for various common applications
    def create_irsa_role(
        self, id: str, namespace: str, service_account_name: str, **kwargs
    ):
        return ps_iam.IrsaRole(
            self,
            id,
            eks_cluster_oidc_provider_arn=self.cluster.open_id_connect_provider.open_id_connect_provider_arn,
            eks_cluster_oidc_provider_issuer=self.cluster.open_id_connect_provider.open_id_connect_provider_issuer,
            namespace=namespace,
            service_account_name=service_account_name,
            **kwargs,
        )

    def create_teleport_kube_agent_role(self) -> iam.Role:
        return self.create_irsa_role(
            "TeleportKubeAgentRole",
            namespace="teleport",
            service_account_name="teleport-kube-agent",
            inline_policies={
                "TeleportAgentPolicy": iam.PolicyDocument(
                    statements=[
                        iam.PolicyStatement(
                            actions=[
                                "sts:AssumeRole",
                            ],
                            resources=[
                                "arn:aws:iam::*:role/pipeline-roles/teleport-*-role"
                            ],
                        ),
                    ]
                )
            },
        )

    def create_aws_load_balancer_controller_role(self) -> iam.Role:
        with open(
            f"{os.path.dirname(__file__)}/files/aws-load-balancer-controller-policy.json"
        ) as policy_data:
            return self.create_irsa_role(
                "AwsLoadBalancerControllerRole",
                namespace="kube-system",
                service_account_name="aws-load-balancer-controller",
                inline_policies={
                    "AwsLoadBalancerControllerPolicy": iam.PolicyDocument.from_json(
                        json.load(policy_data)
                    )
                },
            )
 ```

**Comment by mrgrain** [comment]
Thanks @portswigger-tim  I think I've found the issue. Fix underway.

**Comment by Feder1co5oave** [comment]
@mrgrain I'm experiencing the same issue with **aws-cdk-lib@2.116.0**

```
$ node_modules/.bin/cdk diff

Error: Cannot find module '../dist/core/cfn-utils-provider.generated'
Require stack:
- /home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/lib/private/cfn-utils-provider.js
- /home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/lib/cfn-json.js
- /home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/lib/index.js
- /home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/index.js
- /home/federico/MyCdkProject/node_modules/aws-cdk-lib/index.js
- /home/federico/MyCdkProject/bin/cdk.ts
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
    at Function.Module._resolveFilename.sharedData.moduleResolveFilenameHook.installedValue [as _resolveFilename] (/home/federico/MyCdkProject/node_modules/@cspotcode/source-map-support/source-map-support.js:811:30)
    at Function.Module._load (node:internal/modules/cjs/loader:922:27)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:119:18)
    at cfn_utils_provider_generated_1 (/home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/lib/private/cfn-utils-provider.js:1:436)
    at Function.getOrCreate (/home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/lib/private/cfn-utils-provider.js:1:629)
    at new CfnJson (/home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/lib/cfn-json.js:1:1371)
    at new TagmeStack (/home/federico/MyCdkProject/lib/tagme-stack.ts:22:33)
    at Object.<anonymous> (/home/federico/MyCdkProject/bin/cdk.ts:15:15) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/lib/private/cfn-utils-provider.js',
    '/home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/lib/cfn-json.js',
    '/home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/lib/index.js',
    '/home/federico/MyCdkProject/node_modules/aws-cdk-lib/core/index.js',
    '/home/federico/MyCdkProject/node_modules/aws-cdk-lib/index.js',
    '/home/federico/MyCdkProject/bin/cdk.ts'
  ]
}
```

I too am using `CfnJson` to create a IAM trust policy for an OIDC pricipal:
```ts
const trustConditions = new CfnJson(this, 'iam-role-condition', {
    value: {
        [`${oidcProviderArn}:aud`]: oidcAudience,
        [`${oidcProviderArn}:sub`]: `system:serviceaccount:${serviceAccountNamespace}:${serviceAccountName}`,
    }
});
const oidcPrincipal = new iam.OpenIdConnectPrincipal(oidcProvider).withConditions({
    StringEquals: trustConditions
});
const k8sRole = new iam.Role(this, 'iam-role', { assumedBy: oidcPrincipal });
```

mrgrain:
Thanks @Feder1co5oave Fix is on the way: https://github.com/aws/aws-cdk/pull/28467

Issue #6404, opened 2020-02-22
Enumeration value for HttpMethod must be non-empty (Service: AmazonApiGateway; Status Code: 400; Error Code: BadRequestException


joyofdata:
I'm deploying a REST API on API Gateway accessing a web service on Fargate via a VPC Link and an NLB. And this actually works. 

```
[API Gateway] <-> [VPC Link] <-> { [NLB] <-> [Fargate] }

# { ... private subnet ... }
```

The only missing piece is deploying the method:resource ANY:test/. That part is described in below code snippet:

```py
vpc_link = api.VpcLink(
    self,
    "VpcLinkRaffael",
    targets=[fg.load_balancer],
)


rest_api = api.RestApi(
    self,
    "RestApiRaffael",
)

test = rest_api.root.add_resource("test")
test.add_method(
    "ANY",
    # the following integration setting is causing the problem. for the previous deployment
    # I left this specific parameter out and it deployed fine. so my deployment at this point
    # is a rest API with a resource test and method ANY.
    api.Integration(
        type=api.IntegrationType.HTTP_PROXY,
        options=api.IntegrationOptions(
            vpc_link=vpc_link
        ),
        uri="http://example.org/",
    ),
)
```
This causes the error:
```
 0/5 | 4:57:57 PM | UPDATE_IN_PROGRESS   | AWS::ApiGateway::Method                   | RestApiRaffael/Default/test/ANY (RestApiRaffaeltestANY0602E478)
 1/5 | 4:57:58 PM | UPDATE_FAILED        | AWS::ApiGateway::Method                   | RestApiRaffael/Default/test/ANY (RestApiRaffaeltestANY0602E478) Enumeration value for HttpMethod must be non-empty (Service: AmazonApiGateway; Status Code: 400; Error Code: BadRequestException; Request ID: 0a6b9f7d-4501-41e0-8e4a-cbebd0fb1333)
        new Method (/tmp/jsii-kernel-g0iRHB/node_modules/@aws-cdk/aws-apigateway/lib/method.js:45:26)
        \_ Resource.addMethod (/tmp/jsii-kernel-g0iRHB/node_modules/@aws-cdk/aws-apigateway/lib/resource.js:17:16)
        \_ /home/raffael/gitrepos/test-api-elb-fargate/env/lib/python3.8/site-packages/jsii/_embedded/jsii/jsii-runtime.js:7665:51
[...]

PR #28299, opened 2023-12-08
docs(apigateway): set integrationHttpMethod for vpc link integration

tmokmss:
As written in #6404, a VPC link integration fails to deploy unless we set `integrationHttpMethod` prop explicitly.

In this PR we fix the sample code accordingly.
 
----

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,10 @@ This document describes how to set up a development environment and submit your 
 let us know if it's not up-to-date (even better, submit a PR with your  corrections ;-)).
 
 - [Where To Contribute](#where-to-contribute)
+  - [Demonstrating Value](#demonstrating-value)
+  - [Publishing Your Own Package](#publishing-your-own-package)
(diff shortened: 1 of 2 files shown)


lpizzinidev (changes requested):
I think we should also add validation for the property and update the documentation [here](https://github.com/aws/aws-cdk/blob/020bf18c214199cb124a44fee9a3bb4a8f115ebc/packages/aws-cdk-lib/aws-apigateway/lib/integration.ts#L138) and [here](https://github.com/aws/aws-cdk/blob/020bf18c214199cb124a44fee9a3bb4a8f115ebc/packages/aws-cdk-lib/aws-apigateway/lib/integrations/aws.ts#L59) according to the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-method-integration.html#cfn-apigateway-method-integration-integrationhttpmethod).

kaizencc (approved):
@tmokmss thanks! @lpizzinidev feel free to create an issue out of that, can be addressed in a future PR

kaizencc (approved):

PR #28303, opened 2023-12-08
fix(secretsmanager): cannot set hourly rotation

### LINKED ISSUE #28261: aws-secretsmanager: Cannot set rotation schedule to be less than a day
### Describe the bug

AWS [secrets manager documentation](https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotate-secrets_managed.html), says that "You can rotate a secret as often as every four hours". However if I want to do something like:

```typescript
    new Secret(this, 'my-secret').addRotationSchedule('my-secret-rotation-schedule', {
        rotationLambda: myLambdaFunction,
        automaticallyAfter: Duration.hours(4),
    })
```

I will get the following error:

```
Error: '4 hours' cannot be converted into a whole number of days.
    at convert (/Users/wmattei/www/peek/peek-iac/node_modules/aws-cdk-lib/core/lib/duration.js:1:6197)
    at Duration.toDays (/Users/wmattei/www/peek/peek-iac/node_modules/aws-cdk-lib/core/lib/duration.js:1:3641)
    at new RotationSchedule (/Users/wmattei/www/peek/peek-iac/node_modules/aws-cdk-lib/aws-secretsmanager/lib/rotation-schedule.js:1:2264)
    at Secret.addRotationSchedule (/Users/wmattei/www/peek/peek-iac/node_modules/aws-cdk-lib/aws-secretsmanager/lib/secret.js:1:4727)
```

### Expected Behavior

I would expect CDK to allow me to create schedules for 4 hours, as described in the documention.

### Current Behavior

I can only rotation schedules for the minimum of 1 day.

### Reproduction Steps

Try to synth this stack:

```typescript
import { App, Stack, Duration } from "aws-cdk-lib";
import { Secret } from "aws-cdk-lib/aws-secretsmanager";
import { NodejsFunction } from "aws-cdk-lib/aws-lambda-nodejs";
  constructor() {
    const app = new App();

    super(app, "MyStack");

    const myLambdaFunction = new NodejsFunction(this, "my-lambda-function", {
      entry: "path/to/my/lambda",
    });

    new Secret(this, "my-secret").addRotationSchedule(
      "my-secret-rotation-schedule",
      {
        rotationLambda: myLambdaFunction,
        automaticallyAfter: Duration.hours(4),
      }
    );
  }
}

```

### Possible Solution

I suppose this is related to the fact that [when checking if the provided duration is greater than 1000 days](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-secretsmanager/lib/rotation-schedule.ts#L130) we are directly calling `.toDays` which validates [wether the given value can be converted in a whole number of days](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/duration.ts#L329).

_No response_

### CDK CLI Version

2.113.0 (build ccd534a)

### Framework Version

_No response_

### Node.js Version

v21.3.0

### OS

MacOs

### Language

TypeScript

### Language Version

Version 5.1.6

### Other information

_No response_

lpizzinidev:
Allows to set hourly rotation up to 4 hours on secrets as per [official docs](https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotate-secrets_managed.html).

Closes #28261.

----


packages/aws-cdk-lib/aws-secretsmanager/lib/rotation-schedule.ts
```diff
@@ -126,18 +128,31 @@ export class RotationSchedule extends Resource {
...
-    let automaticallyAfterDays: number | undefined = undefined;
-    if (props.automaticallyAfter && props.automaticallyAfter.toDays() > 1000) {
-      throw new Error(`automaticallyAfter must not be greater than 1000 days, got ${props.automaticallyAfter.toDays()} days`);
-    }
-    if (props.automaticallyAfter?.toMilliseconds() !== 0) {
-      automaticallyAfterDays = props.automaticallyAfter?.toDays() || 30;
+    let automaticallyAfterDays: number | undefined;
+    let scheduleExpression: string | undefined;
+    if (props.automaticallyAfter) {
+      if (props.automaticallyAfter.toMilliseconds() > 0) {
```
vinayak-kukreja, line 134: `props.automaticallyAfter.toMilliseconds()` 

I think this is repeating and we can keep this as a variable.

packages/aws-cdk-lib/aws-secretsmanager/lib/rotation-schedule.ts
```diff
@@ -126,18 +128,31 @@ export class RotationSchedule extends Resource {
...
+        if (props.automaticallyAfter.toMilliseconds() < Duration.hours(4).toMilliseconds()) {
+          throw new Error(`automaticallyAfter must not be smaller than 4 hours, got ${props.automaticallyAfter.toHours()} hours`);
+        }
+        if (props.automaticallyAfter.toMilliseconds() > Duration.days(1000).toMilliseconds()) {
+          throw new Error(`automaticallyAfter must not be greater than 1000 days, got ${props.automaticallyAfter.toDays()} days`);
+        }
+        if (props.automaticallyAfter.toHours() > 23) {
```
vinayak-kukreja, line 141: NIT: How about `>=24`, just easier to interpret for me. Thoughts?

packages/@aws-cdk-testing/framework-integ/test/aws-secretsmanager/test/integ.lambda-rotation.ts
```diff
@@ -21,6 +21,7 @@ class TestStack extends cdk.Stack {
         handler: 'index.handler',
         code: lambda.Code.fromInline('NOOP'),
       }),
+      automaticallyAfter: cdk.Duration.hours(4),
```
vinayak-kukreja, line 24: Running the integ test to verify.

vinayak-kukreja (changes requested):
Hey @lpizzinidev , thank you for the contribution. Just some minor comments :)

packages/@aws-cdk-testing/framework-integ/test/aws-secretsmanager/test/integ.lambda-rotation.ts
vinayak-kukreja: Hey, I see the integ test fails due to the schedule not working. I still see the schedule is at 30 days.

lpizzinidev: I didn't update `tree.json`, should be good now.

packages/aws-cdk-lib/aws-secretsmanager/lib/rotation-schedule.ts
```diff
@@ -2,6 +2,7 @@ import { Construct } from 'constructs';
 import { ISecret, Secret } from './secret';
 import { CfnRotationSchedule } from './secretsmanager.generated';
 import * as ec2 from '../../aws-ec2';
+import { Schedule } from '../../aws-events';
```
kaizencc, line 5: AHhhh I suppose this is ok because the schedule is not exposed in anyway, but we STILL need to get this into class into core.

packages/aws-cdk-lib/aws-secretsmanager/lib/rotation-schedule.ts
```diff
@@ -126,18 +128,32 @@ export class RotationSchedule extends Resource {
...
+      automaticallyAfterDays = 30;
     }
 
-    let rotationRules: CfnRotationSchedule.RotationRulesProperty | undefined = undefined;
-    if (automaticallyAfterDays !== undefined) {
+    let rotationRules: CfnRotationSchedule.RotationRulesProperty | undefined;
+    if (automaticallyAfterDays !== undefined || scheduleExpression !== undefined) {
       rotationRules = {
         automaticallyAfterDays,
+        scheduleExpression,
```
kaizencc, line 156: @lpizzinidev I don't like this implementation because we are returning both `automaticallyAfterDays` and `scheduleExpression` knowing full well that one is undefined. We're then relying on this information elsewhere, which is okay for now, but is an assumption we'll have to maintain forever.

Instead, I think we should just localize on `scheduleExpression`. We can keep the `automaticallyAfter` prop, but wire it to `scheduleExpression` all the time. We can specify `rate(5 days)` if we get a unit in days, or `rate(4 hours)` if the unit is in hours.

`scheduleExpression` can do everything `automaticallyAfterDays` can do, and more, so lets standardize on that.

kaizencc (changes requested):
Of course, long term we need to get `schedule` into core so we can use that instead of using the one in aws-events.

kaizencc (approved):
Thanks @lpizzinidev

PR #26819, opened 2023-08-20
feat(cloudwatch): add verticalAnnotations property to GraphWidget

jasonsaruulo:
Adds a verticalAnnotation property to GraphWidget, reference: https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/CloudWatch-Dashboard-Body-Structure.html#CloudWatch-Dashboard-Properties-Annotation-Format

Shoutout to brendo-m for coming up with the solution

Closes #7622.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

Issue #28124, opened 2023-11-23
s3: Transition Date doesn't follow the Cloudformation type


wakeupmh:
### Describe the bug

Our application encountered an **error** during deployment, which occurred suddenly, and when I sought more information about it I saw that [cloud formation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-transition.html#cfn-s3-bucket-transition-transitiondate) requires a string following this pattern:

> TransitionDate
> Indicates when objects are transitioned to the specified storage class. The date value must be in ISO 8601 format. The time is always midnight UTC.
> 
> Required: Conditional
> 
> Type: String
> 
> Pattern: ^([0-2]\d{3})-(0[0-9]|1[0-2])-([0-2]\d|3[01])T([01]\d|2[0-4]):([0-5]\d):([0-6]\d)((\.\d{3})?)Z$

We were following the CDK 2.11.0 docs for creating [Lifecycle Rule](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Transition.html) as follow:

> 
> // The code below shows an example of how to instantiate this type.
> // The values are placeholders you should change.
> import * as cdk from 'aws-cdk-lib';
> import { aws_s3 as s3 } from 'aws-cdk-lib';
> 
> declare const storageClass: s3.StorageClass;
> const transition: s3.Transition = {
>   storageClass: storageClass,
> 
>   // the properties below are optional
>   transitionAfter: cdk.Duration.minutes(30),
>   transitionDate: new Date(),
> };

### Error:

```
2023-11-22T00:28:19.0029491Z  ❌  infrastructure/root-infra (infrastructure) failed: Error: The stack named infrastructure failed to deploy: UPDATE_ROLLBACK_COMPLETE: Properties validation failed for resource rootinfraarchiveassetsbucketarchiveassetsE1767254 with message:
2023-11-22T00:28:19.0033070Z #/LifecycleConfiguration/Rules/0/Transitions/0/TransitionDate: failed validation constraint for keyword [pattern]
2023-11-22T00:28:19.0035318Z #/LifecycleConfiguration/Rules/0/Transitions/1/TransitionDate: failed validation constraint for keyword [pattern]
```

### Expected Behavior

Create a Lifecycle Rule following what cloud formation expects

### Current Behavior

[...]

PR #28306, opened 2023-12-08
feat(scheduler): start and end time for schedule construct

sakurai-ryo:
This PR added support for start and end time of the schedule.

## Description
Currently, users cannot set a start time and an end time for the schedule.
A schedule without a start date will begin as soon as it is created and available, and without an end date, it will continue to invoke its target indefinitely.
With this feature, users can set the start and end dates of a schedule, allowing for more flexible schedule configurations.
https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-resource-scheduler-schedule.html#cfn-scheduler-schedule-startdate
https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-resource-scheduler-schedule.html#cfn-scheduler-schedule-enddate

In CloudFormation, users can use this feature as follows:
```yaml
  TestSchedule:
    Type: AWS::Scheduler::Schedule
    Properties: 
      StartDate: "2024-12-01T13:09:00.000Z"
      EndDate: "2025-12-01T00:00:00.001Z"
      ScheduleExpression: "at(2023-01-01T00:00:00)"
      State: "ENABLED"
      Target: # target
```

### add property to ScheduleProps interface
Added startDate and endDate properties, and typed these values as string based on the following PR comments.
https://github.com/aws/aws-cdk/pull/26819#discussion_r1301532299

It is not necessary to specify both startDate and endDate, they can be set independently.
Validation is performed on the following points.
- Error if not following ISO 8601 format.
  - Must include milliseconds (yyyy-MM-ddTHH:mm:ss.SSSZ)
- Error if startDate is later than endDate

If a time before the current time is specified, the following error occurs in CFn, but no validation is performed because the timing of validation in CDK and the timing of actual deployment are different.
`The StartDate you specify cannot be earlier than 5 minutes ago.`

----


packages/@aws-cdk/aws-scheduler-alpha/lib/schedule.ts
```diff
@@ -213,6 +233,8 @@ export class Schedule extends Resource implements ISchedule {
     return this.metricAll('InvocationsSentToDeadLetterCount_Truncated_MessageSizeExceeded', props);
   }
 
+  private static readonly ISO8601_REGEX = /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])\.[0-9]{3}(Z)?$/;
```
lpizzinidev, line 236:
```suggestion
  private static readonly ISO8601_REGEX = /^([0-2]\d{3})-(0[0-9]|1[0-2])-([0-2]\d|3[01])T([01]\d|2[0-4]):([0-5]\d):([0-6]\d)((\.\d{3})?)Z$/;
```
A [previous issue](https://github.com/aws/aws-cdk/issues/28124) was caused by [TransitionDate](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-transition.html#cfn-s3-bucket-transition-transitiondate) being formatted without the final `Z` so a different regex may be better in this case.

packages/@aws-cdk/aws-scheduler-alpha/lib/schedule.ts
```diff
@@ -306,4 +332,22 @@ export class Schedule extends Resource implements ISchedule {
     const isEmptyPolicy = Object.values(policy).every(value => value === undefined);
     return !isEmptyPolicy ? policy : undefined;
   }
-}
\ No newline at end of file
+
+  private validateTimeFrame(startDate?: string, endDate?: string) {
+    if (startDate && !Schedule.ISO8601_REGEX.test(startDate)) {
+      throw new Error(`startDate needs to follow the format yyyy-MM-ddTHH:mm:ss.SSSZ but got ${startDate}`);
```
lpizzinidev, line 338:
```suggestion
      throw new Error(`startDate must be in ISO 8601 format, got ${startDate}`);
```

packages/@aws-cdk/aws-scheduler-alpha/lib/schedule.ts
```diff
@@ -306,4 +332,22 @@ export class Schedule extends Resource implements ISchedule {
...
+    }
+    if (endDate && !Schedule.ISO8601_REGEX.test(endDate)) {
+      throw new Error(`endDate needs to follow the format yyyy-MM-ddTHH:mm:ss.SSSZ but got ${endDate}`);
```
lpizzinidev, line 341:
```suggestion
      throw new Error(`endDate must be in ISO 8601 format, got ${endDate}`);
```

packages/@aws-cdk/aws-scheduler-alpha/lib/schedule.ts
```diff
@@ -306,4 +332,22 @@ export class Schedule extends Resource implements ISchedule {
...
+    }
+
+    if (startDate && endDate) {
+      const start = new Date(startDate);
+      const end = new Date(endDate);
+
+      if (end <= start) {
+        throw new Error(`startDate must come before the endDate but got startDate: ${startDate}, endDate: ${endDate}`);
+      }
+    }
```
lpizzinidev, line 351:
```suggestion
    if (startDate && endDate && startDate > endDate) {
      throw new Error(`startDate must preceed endDate, got startDate: ${startDate}, endDate: ${endDate}`);
    }
```
It should be fine to compare the strings directly given the format.

packages/@aws-cdk/aws-scheduler-alpha/README.md
```diff
@@ -138,6 +138,21 @@ new Schedule(this, 'Schedule', {
 });
 ```
 
+### Configuring a start and end time of the Schedule
+
+If you choose a recurring schedule, you can set the start and end time of the Schedule by specifying the startDate and endDate.
+These values must follow the format `yyyy-MM-ddTHH:mm:ss.SSSZ`.
```
lpizzinidev, line 144:
```suggestion
These values must follow the ISO 8601 format (`yyyy-MM-ddTHH:mm:ss.SSSZ`).
```

lpizzinidev (changes requested):
Looks good overall 👍
I left some suggestions for minor improvements.

packages/@aws-cdk/aws-scheduler-alpha/lib/schedule.ts
sakurai-ryo: You are absolutely right.
I made some modifications from your suggestion because CFn throws the following error if it does not include milliseconds.
`StartDate needs to follow the format yyyy-MM-ddTHH:mm:ss.SSSZ`

sakurai-ryo:
@lpizzinidev 
Thanks for your review! I fixed them.

lpizzinidev (approved):
Nice 👍

packages/@aws-cdk/aws-scheduler-alpha/README.md
```diff
@@ -138,6 +138,21 @@ new Schedule(this, 'Schedule', {
 });
 ```
 
+### Configuring a start and end time of the Schedule
+
+If you choose a recurring schedule, you can set the start and end time of the Schedule by specifying the startDate and endDate.
```
kaizencc, line 143:
```suggestion
If you choose a recurring schedule, you can set the start and end time of the Schedule by specifying the `startDate` and `endDate`.
```

packages/@aws-cdk/aws-scheduler-alpha/README.md
```diff
@@ -138,6 +138,21 @@ new Schedule(this, 'Schedule', {
...
+These values must follow the ISO 8601 format (`yyyy-MM-ddTHH:mm:ss.SSSZ`).
+
+```ts
+declare const target: targets.LambdaInvoke;
+
+new Schedule(this, 'Schedule', {
+    schedule: ScheduleExpression.rate(cdk.Duration.hours(12)),
+    target: target,
+    startDate: '2023-01-01T00:00:00.000Z',
+    endDate: '2023-02-01T00:00:00.000Z',
```
kaizencc, line 153: I think we standardized on using the `Date` type a while back. Similar to what we are doing [here](https://github.com/aws/aws-cdk/blob/15c7bb2908cff1ab8bebacacc09f26bbab6ec391/packages/aws-cdk-lib/aws-applicationautoscaling/lib/scalable-target.ts#L196). We can then convert to whatever format scheduler expects.

How do you feel about that? I think that's the best course of action here as it allows for more types of inputs, but feel free to push back.

packages/@aws-cdk/aws-scheduler-alpha/README.md
kaizencc: I would also prefer the props to be just named `start` and `end`, or `startTime`/`endTime`.

kaizencc (changes requested):

packages/@aws-cdk/aws-scheduler-alpha/README.md
sakurai-ryo: In CFn, it is StartDate and EndDate, so I shortened it to start and end.

packages/@aws-cdk/aws-scheduler-alpha/README.md
sakurai-ryo: Thanks for the comment!

Using the Date type to allow for a variety of user inputs is certainly appealing.
However, I have some reservations about it, primarily due to potential issues with timezones.
When we initialize a Date object with a string that lacks timezone specification(`new Date("2022-08-04T15:19:46.125")`), the system's timezone is used by default.
For instance, if a local machine is set to Asia/Tokyo and the CI environment to UTC, the same time would be offset by 9 hours between these environments.
Since we are only dealing with ISO 8601 format and exclusively in UTC for this properties, using a string might be a safer option to avoid these timezone pitfalls.

By typing these values as string representation in the ISO 8601 format, we could mitigate these inconsistencies and ensure more predictable behavior across different environments.

This is just my personal opinion and I would love to get feedback from the CDK team.

sakurai-ryo:
@kaizencc 
Thanks for your review!
I fixed it and left some comments.

packages/@aws-cdk/aws-scheduler-alpha/README.md
kaizencc: CFN accepts `startDate` only in UTC time. I think when we do `new Date().toISOString()` the Date is automatically converted to UTC time, specifically ISO 8601, which is what we want.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

sakurai-ryo: Yes, `toISOString` method is used to convert to ISO 8601 format (with UTC timezone), but in the previous comment I was referring to the timezone at the initialization of the Date object.

However, I would like to change these properties to use Date type, since this problem does not occur if the user passes a string with timezone at initialization, and this is not a CDK problem but a JavaScript Date class specification problem.
Sorry for the confusion.

sakurai-ryo:
@kaizencc 
Thanks! Changed property type to use Date type.
https://github.com/aws/aws-cdk/pull/28306#discussion_r1425297957

kaizencc (approved):
Lovely @sakurai-ryo, thanks for your contribution!

PR #28319, opened 2023-12-10
fix(logs): missing KMS key policy statement

### LINKED ISSUE #28304: aws-logs: KMS Policy Not Updated When Using KMS CMK
### Describe the bug

Reading the [docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_logs-readme.html#encrypting-log-groups) for using KMS CMK with a Log Group should just involve passing the key to the Log Group. However, when doing that, KMS lacks the required permission to interact with `Logs`.

### Expected Behavior

For the LogGroup Construct to automatically set the required permissions to interact with KMS when a KMS CMK key is passed to it.

### Current Behavior

```
12:58:26 PM | CREATE_FAILED        | AWS::Logs::LogGroup | logs00112233
Resource handler returned message: "The specified KMS key does not exist or is not allowed to be used with Arn 'arn:aws:logs:us-east-1:112233445566:log-group:PythonStack-logs00112233-AABBCCDDEEFF' (Service: CloudWatchLogs, Status Code: 400, Request <...>. HandlerErrorCode: GeneralServiceException)
```

### Reproduction Steps

```
from aws_cdk import (
    Stack,
    aws_kms as kms,
    aws_logs as logs,
)


class PythonStack(Stack):
    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
        super().__init__(scope, construct_id, **kwargs)

        key = kms.Key(self, "key")

        log_group = logs.LogGroup(self, "logs", encryption_key=key)
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.114.1

### Framework Version

_No response_

### Node.js Version

18.16.0

### OS

Mac

### Language

Python

### Language Version

_No response_

### Other information

_No response_

lpizzinidev:
Fixes [key policy for encrypted log groups](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/encrypt-log-data-kms.html).
The generated log groups are missing a statement causing the deployment to fail.

Closes #28304.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-logs/lib/log-group.ts
+++ b/packages/aws-cdk-lib/aws-logs/lib/log-group.ts
@@ -9,7 +9,7 @@ import { ILogSubscriptionDestination, SubscriptionFilter } from './subscription-
 import * as cloudwatch from '../../aws-cloudwatch';
 import * as iam from '../../aws-iam';
 import * as kms from '../../aws-kms';
-import { Annotations, Arn, ArnFormat, RemovalPolicy, Resource, Stack, Token } from '../../core';
+import { Annotations, Arn, ArnFormat, Lazy, Names, RemovalPolicy, Resource, Stack, Token } from '../../core';
 
 export interface ILogGroup extends iam.IResourceWithPolicy {
   /**
@@ -491,7 +491,9 @@ export class LogGroup extends LogGroupBase {
 
   constructor(scope: Construct, id: string, props: LogGroupProps = {}) {
     super(scope, id, {
-      physicalName: props.logGroupName,
+      physicalName: props.logGroupName ?? Lazy.string({
+        produce: () => Names.uniqueResourceName(this, { maxLength: 512, allowedSpecialCharacters: '-_' }),
+      }),
     });
 
     let retentionInDays = props.retention;
@@ -547,8 +549,8 @@ export class LogGroup extends LogGroupBase {
         ],
         resources: ['*'],
         conditions: {
-          ArnLike: {
-            'kms:EncryptionContext:aws:logs:arn': `arn:${this.stack.partition}:logs:${this.env.region}:${this.env.account}:*`,
+          ArnEquals: {
+            'kms:EncryptionContext:aws:logs:arn': `arn:${this.stack.partition}:logs:${this.env.region}:${this.env.account}:log-group:${this.physicalName}`,
           },
         },
       }));
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-logs/test/integ.log-group-encrypted.js.snapshot/aws-cdk-log-group-encrypted-integ.template.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-logs/test/integ.log-group-encrypted.js.snapshot/aws-cdk-log-group-encrypted-integ.template.json
@@ -37,7 +37,7 @@
         "kms:ReEncrypt*"
        ],
        "Condition": {
-        "ArnLike": {
+        "ArnEquals": {
          "kms:EncryptionContext:aws:logs:arn": {
           "Fn::Join": [
            "",
@@ -54,7 +54,7 @@
             {
              "Ref": "AWS::AccountId"
             },
-            ":*"
+            ":log-group:aws-cdk-log-group-encrypted-integLogGroupDECB5FC9"
            ]
           ]
          }
@@ -87,6 +87,7 @@
   "LogGroupF5B46931": {
    "Type": "AWS::Logs::LogGroup",
    "Properties": {
+    "LogGroupName": "aws-cdk-log-group-encrypted-integLogGroupDECB5FC9",
     "KmsKeyId": {
      "Fn::GetAtt": [
       "Key961B73FD",
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-logs/test/integ.log-group-encrypted.js.snapshot/tree.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-logs/test/integ.log-group-encrypted.js.snapshot/tree.json
@@ -118,6 +118,7 @@
                 "attributes": {
                   "aws:cdk:cloudformation:type": "AWS::Logs::LogGroup",
                   "aws:cdk:cloudformation:props": {
+                    "logGroupName": "aws-cdk-log-group-encrypted-integLogGroupDECB5FC9",
                     "kmsKeyId": {
                       "Fn::GetAtt": [
                         "Key961B73FD",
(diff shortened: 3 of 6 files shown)


packages/aws-cdk-lib/aws-logs/lib/log-group.ts
```diff
@@ -534,6 +534,25 @@ export class LogGroup extends LogGroupBase {
...
+          'kms:Encrypt*',
+          'kms:Decrypt*',
+          'kms:ReEncrypt*',
+          'kms:GenerateDataKey*',
+          'kms:Describe*',
+        ],
+        resources: ['*'],
+        conditions: {
+          ArnLike: {
+            'kms:EncryptionContext:aws:logs:arn': `arn:${this.stack.partition}:logs:${this.env.region}:${this.env.account}:*`,
```
msambol, line 551: should this be `log-group:<log-group-name>` instead of `*` ?

lpizzinidev: I tried to use `ArnEquals` with the log group ARN but it created a circular dependency between the key and the log group.
Not sure if there's a workaround, I agree that this may be too permissive.

msambol: If there is no easy workaround, I am OK with this, as the second example in the docs has `*`.

kaizencc: I'm not sure if this will work @lpizzinidev but have you tried using `log-group:${this.physicalName}`? Instead of using `logGroupArn` or `logGroupName`, which are attributes of the log group.

lpizzinidev: @kaizencc 
I tried `log-group:${this.physicalName}` and it works.
It should be safe to go with it, let me know if you think otherwise.

kaizencc: It works as in you integ tested it and confirmed that the permission boundary works as expected (not just that it deploys successfully)? 

If so, I think that's the best path forward, and once thats in I'm happy to approve.

lpizzinidev: @kaizencc 
I got it wrong.
```
conditions: {
  ArnEquals: {
    'kms:EncryptionContext:aws:logs:arn': `arn:${this.stack.partition}:logs:${this.env.region}:${this.env.account}:log-group:${this.physicalName}`,
  },
},
```
Produces this incorrect policy condition:
```
Condition: {
  ArnEquals: {
    'kms:EncryptionContext:aws:logs:arn': {
      'Fn::Join': [
        '',
        [
          'arn:',
          {
            Ref: 'AWS::Partition',
          },
          ':logs:',
          {
            Ref: 'AWS::Region',
          },
          ':',
          {
            Ref: 'AWS::AccountId',
          },
          ':log-group:',
        ],
      ],
    },
  },
},
```
It deploys successfully but the log group name is left blank.

msambol: If you use `Lazy` does that fix it ?

kaizencc: @lpizzinidev ^. Fingers crossed it does :)

packages/aws-cdk-lib/aws-logs/lib/log-group.ts
```diff
@@ -491,7 +491,9 @@ export class LogGroup extends LogGroupBase {
 
   constructor(scope: Construct, id: string, props: LogGroupProps = {}) {
     super(scope, id, {
-      physicalName: props.logGroupName,
+      physicalName: props.logGroupName ?? Lazy.string({
+        produce: () => Names.uniqueResourceName(this, { maxLength: 512, allowedSpecialCharacters: '-_' }),
+      }),
```
msambol, line 496: this shouldn't change existing log group names will it ?

packages/aws-cdk-lib/aws-logs/lib/log-group.ts
```diff
@@ -534,6 +536,25 @@ export class LogGroup extends LogGroupBase {
...
+          'kms:Encrypt*',
+          'kms:Decrypt*',
+          'kms:ReEncrypt*',
+          'kms:GenerateDataKey*',
+          'kms:Describe*',
+        ],
+        resources: ['*'],
+        conditions: {
+          ArnEquals: {
+            'kms:EncryptionContext:aws:logs:arn': `arn:${this.stack.partition}:logs:${this.env.region}:${this.env.account}:log-group:${this.physicalName}`,
```
msambol, line 553: nice

packages/aws-cdk-lib/aws-logs/lib/log-group.ts
lpizzinidev: @kaizencc 
This is my concern as well.
With this approach, we'll probably trigger a replacement of existing log groups that don't have `logGroupName` specified (unless I'm missing something).
I don't see a workaround, and the previous solution with `ArnLike` / `:log-group:*` may be too loose.
Thoughts?

kaizencc: Workaround as always is a feature flag, and I agree that the previous solution is probably too loose. Give me a bit to think this one through and see what we should od. I think the safest thing is a feature flag but will ask around to make sure.

rix0rrr: Alternative proposal:

If we add a [kms:ViaService](https://docs.aws.amazon.com/kms/latest/developerguide/conditions-kms.html#conditions-kms-via-service) condition to the key, is that not good enough? It will allow us to avoid naming the log group.

lpizzinidev: @rix0rrr 
Isn't it the same as
```
conditions: {
 ArnLike: {
  'kms:EncryptionContext:aws:logs:arn': `arn:${this.stack.partition}:logs:${this.env.region}:${this.env.account}:log-group:*`,
 },
},
```
?
It looks to me like we restrict permissions to the service in both cases (original problem).

lpizzinidev: @rix0rrr 
Isn't it the same as
```
conditions: {
 ArnLike: {
  'kms:EncryptionContext:aws:logs:arn': `arn:${this.stack.partition}:logs:${this.env.region}:${this.env.account}:log-group:*`,
 },
},
```
?
It looks to me like we restrict permissions to the service in both cases (original problem).

lpizzinidev:
@rix0rrr https://github.com/aws/aws-cdk/pull/28319#discussion_r1439482504

PR #28322, opened 2023-12-11
chore: remove Policy property from lambda function hash

mrgrain:
This property was removed from CFN and is therefore not currently in use.

(Needs an upgrade of the spec database before tests will pass)

----


kaizencc (approved):

mrgrain:
Ha, turns out the fix is needed in https://github.com/aws/aws-cdk/pull/28328 to make the PR pass. Closing this one.

PR #28329, opened 2023-12-11
fix(ecs): remove default deployment alarm setting

sumupitchayan:
Originally in this PR https://github.com/aws/aws-cdk/pull/25840, we added default deployment alarm settings to fix an issue where adding deployment alarms, deploying your CFN stack, then removing the deployment alarms from the CFN template, and deploying again WILL NOT remove the deployment alarms from the service.

ECS now already supports default deployment alarm settings. We will remove the default setting of `deploymentAlarms`. Reason for removing this default behaviour is for region build where the deployment alarm service may not be available in new regions but is set by default by CDK.

Internal ticket tracking V1142791950

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-ecs/test/ec2/integ.capacity-provider-managed-draining.js.snapshot/integ-ec2-capacity-provider-managed-draining.assets.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-ecs/test/ec2/integ.capacity-provider-managed-draining.js.snapshot/integ-ec2-capacity-provider-managed-draining.assets.json
@@ -1,7 +1,7 @@
 {
   "version": "36.0.0",
   "files": {
-    "72dd1ca10064654eb4724b599868f5b0d889dc243baed9bb832226e6da869796": {
+    "b34fad7b41eb8f6d5e2b8d503f70a432f882d1605f67f76d700740feed2deeae": {
       "source": {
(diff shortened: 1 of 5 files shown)


packages/aws-cdk-lib/aws-ecs/test/base-service.test.ts
```diff
@@ -82,7 +82,7 @@ describe('For alarm-based rollbacks', () => {
     stack = new cdk.Stack();
   });
 
-  test('deploymentAlarms is set by default for ECS deployment controller', () => {
+  test('deploymentAlarms is NOT set by default for ECS deployment controller', () => {
```
comcalvi, line 85: This is a breaking change. We need a feature flag for this.

comcalvi (changes requested):
Can you update the PR description to indicate why we're doing this? This change needs a feature flag, since it's a breaking change.

GavinZZ:
I discussed with @sumupitchayan and I don't think he'll have the capacity to work on this in the near future. I'll see what I can do to get this merged first.

packages/aws-cdk-lib/aws-ecs/test/base-service.test.ts
GavinZZ: Added a feature flag.

GavinZZ:
Going to close this PR since the original author doesn't have the capacity to work on it. I'll take over and create a new PR with feature flag to guarantee backward compatibility.

PR #28336, opened 2023-12-12
fix(cli): `cdk diff` falsely reports resource replacements on trivial template changes

comcalvi:
Adds a new flag to diff, `--change-set`, that creates a new changeset and uses it to determine resource replacement. This new flag is on by default. 

When the flag is set, the following happens:

* Resource metadata changes are obscured
* Resource changes that do not appear in the changeset are obscured from the diff

When the flag is unset, yaml Fn::GetAtt short-form uses are considered equivalent to their long-form counterpart. 

----

### DIFF
--- a/CHANGELOG.v2.alpha.md
+++ b/CHANGELOG.v2.alpha.md
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.118.0-alpha.0](https://github.com/aws/aws-cdk/compare/v2.117.0-alpha.0...v2.118.0-alpha.0) (2024-01-03)
+
+
+### Features
+
+* **glue:** database description property ([#27744](https://github.com/aws/aws-cdk/issues/27744)) ([cbac240](https://github.com/aws/aws-cdk/commit/cbac24041db7dbc39b4ae1d6da4902b3443528cb)), closes [#27740](https://github.com/aws/aws-cdk/issues/27740)
+* **glue-alpha:** add `cfn-glue-table-tableinput-parameters` to Glue table construct ([#27643](https://github.com/aws/aws-cdk/issues/27643)) ([8e15482](https://github.com/aws/aws-cdk/commit/8e15482295c1324eefea020faeb11e4c686357c6))
+
+
+### Bug Fixes
+
+* **lambda-go:** path with space breaks go build ([#28554](https://github.com/aws/aws-cdk/issues/28554)) ([a8a639e](https://github.com/aws/aws-cdk/commit/a8a639e2a2114162db240361c32c40a596a7a19e)), closes [#28555](https://github.com/aws/aws-cdk/issues/28555)
+
 ## [2.117.0-alpha.0](https://github.com/aws/aws-cdk/compare/v2.116.1-alpha.0...v2.117.0-alpha.0) (2023-12-26)
 
 
--- a/CHANGELOG.v2.md
+++ b/CHANGELOG.v2.md
@@ -2,6 +2,36 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.118.0](https://github.com/aws/aws-cdk/compare/v2.117.0...v2.118.0) (2024-01-03)
+
+
+### Features
+
+* **appsync:** IntrospectionConfig property ([#28500](https://github.com/aws/aws-cdk/issues/28500)) ([98ed6b3](https://github.com/aws/aws-cdk/commit/98ed6b3346e4afd423592296bb1620d20544acad)), closes [#28429](https://github.com/aws/aws-cdk/issues/28429)
+* **autoscaling:** add support for `InstanceRequirements` property ([#28464](https://github.com/aws/aws-cdk/issues/28464)) ([276e3a6](https://github.com/aws/aws-cdk/commit/276e3a66febf32afc65a93d7296ec6be8f6e2126)), closes [#28393](https://github.com/aws/aws-cdk/issues/28393)
+* **cloudfront:** CloudFront Function runtime property ([#28099](https://github.com/aws/aws-cdk/issues/28099)) ([9b466ae](https://github.com/aws/aws-cdk/commit/9b466ae136910d07d4297a31f1010fa26ce864d0)), closes [#28163](https://github.com/aws/aws-cdk/issues/28163)
+* **cloudfront:** Key Value Store L2 ([#28473](https://github.com/aws/aws-cdk/issues/28473)) ([030db42](https://github.com/aws/aws-cdk/commit/030db42ad06a471af3833665cb80dceee034aa02)), closes [#28377](https://github.com/aws/aws-cdk/issues/28377)
+* **codepipeline-actions:** more convenient methods to `CacheControl` ([#28491](https://github.com/aws/aws-cdk/issues/28491)) ([a59dc0c](https://github.com/aws/aws-cdk/commit/a59dc0cb79af98ee601f352a7cf1b5fa9cc4d365)), closes [#25477](https://github.com/aws/aws-cdk/issues/25477)
+* **ecs:** `interactive` option in `ContainerDefinitionOptions` ([#28536](https://github.com/aws/aws-cdk/issues/28536)) ([1f9788f](https://github.com/aws/aws-cdk/commit/1f9788f8ad45c0f159c9b7e6bafb76763e9b0bea)), closes [#24326](https://github.com/aws/aws-cdk/issues/24326)
+* **ecs:** enable cluster to grant task protection API permissions to IAM entities ([#28486](https://github.com/aws/aws-cdk/issues/28486)) ([9bc972b](https://github.com/aws/aws-cdk/commit/9bc972b923b66248fb8392e9cba1b2b5829c6e1b)), closes [#26233](https://github.com/aws/aws-cdk/issues/26233)
+* **ecs:** nvidia support to BottlerocketEcsVariant enum for gpu-accelerated tasks ([#28488](https://github.com/aws/aws-cdk/issues/28488)) ([832e29a](https://github.com/aws/aws-cdk/commit/832e29a47c2cb67558c0148a3bf1fa22e4d5cb82)), closes [#25980](https://github.com/aws/aws-cdk/issues/25980)
+* **iam:** validate roleName ([#28509](https://github.com/aws/aws-cdk/issues/28509)) ([999c01a](https://github.com/aws/aws-cdk/commit/999c01a1250308b482fe33e651d7ee8da1f96956)), closes [#28502](https://github.com/aws/aws-cdk/issues/28502)
+* **opensearchservice:** ip address type for domain ([#28497](https://github.com/aws/aws-cdk/issues/28497)) ([6b80338](https://github.com/aws/aws-cdk/commit/6b80338be59bdfa1d14d81fa1ae79f2eda889b37)), closes [#28436](https://github.com/aws/aws-cdk/issues/28436)
+* **rds:** `timeout` and `timeoutAction` properties to ServerlessCluster ([#28534](https://github.com/aws/aws-cdk/issues/28534)) ([508825b](https://github.com/aws/aws-cdk/commit/508825b14d6ccae7274afc08487329162182e887)), closes [#27183](https://github.com/aws/aws-cdk/issues/27183)
+* **rds:** ClientPasswordAuthType property on DatabaseProxy ([#28540](https://github.com/aws/aws-cdk/issues/28540)) ([669e6ff](https://github.com/aws/aws-cdk/commit/669e6ffc6e9d8c7b06cb0f62348635f933b6f823)), closes [#28415](https://github.com/aws/aws-cdk/issues/28415)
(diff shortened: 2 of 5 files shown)


packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
```diff
@@ -93,7 +104,11 @@ function propagatePropertyReplacement(source: types.ResourceDifference, dest: ty
   }
 }
 
-function calculateTemplateDiff(currentTemplate: { [key: string]: any }, newTemplate: { [key: string]: any }): types.TemplateDiff {
+function calculateTemplateDiff(
+  currentTemplate: { [key: string]: any },
+  newTemplate: { [key: string]: any },
+  replacements?: ResourceReplacements,
```
rix0rrr, line 110: I'm wondering if we can avoid passing this structure all the way down through a bunch of functions.

For example:

- Either stick all of these on a class, so it becomes a class member.
- Or: organize the diffing more into a pipeline. First we calculate all the diffs that derive from the template change itself, and then using the information from change set we remove false positives from the diff. So it would be like:

```ts
function fullDiff(a, b, changeSet) {
  return filterFalsePositivies(changeSet, templateDiff(a,b));
}
```

It might be that that is not possible (tell me if it isn't). But I like keeping concerns separated as much as possible.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -147,10 +158,18 @@ export class CdkToolkit {
         const currentTemplate = templateWithNames.deployedTemplate;
         const nestedStackCount = templateWithNames.nestedStackCount;
 
+        const changeSet = options.changeSet ? await prepareAndCreateChangeSet({
```
rix0rrr, line 161: Just for safety's sake, I think I'd like us to catch any permission errors here, print a warning, and fall back to the "less accurate" diff.

Otherwise, I'm concerned this mandatory new API call is going to break for some people somewhere.

packages/aws-cdk/lib/diff.ts
```diff
@@ -74,8 +76,13 @@ export enum RequireApproval {
...
  * Returns true if the changes are prompt-worthy, false otherwise.
  */
-export function printSecurityDiff(oldTemplate: any, newTemplate: cxapi.CloudFormationStackArtifact, requireApproval: RequireApproval): boolean {
-  const diff = cfnDiff.diffTemplate(oldTemplate, newTemplate.template);
+export function printSecurityDiff(
+  oldTemplate: any,
+  newTemplate: cxapi.CloudFormationStackArtifact,
+  requireApproval: RequireApproval,
+  changeSet?: CloudFormation.DescribeChangeSetOutput,
+): boolean {
```
rix0rrr, line 84: This function is wrong. The input should be a `diff`, not two templates that should be diffed again. Now we're diffing them twice!

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -929,6 +948,13 @@ export interface DiffOptions {
   * @default false
   */
   quiet?: boolean;
+
+  /**
+   * Whether or not to create, analyze, and subsequently delete a changeset
```
rix0rrr, line 953: That's cool, but I don't see the code to delete it :P

packages/aws-cdk/lib/cli.ts
```diff
@@ -262,7 +262,8 @@ async function parseCommandLineArguments(args: string[]) {
       .option('security-only', { type: 'boolean', desc: 'Only diff for broadened security changes', default: false })
       .option('fail', { type: 'boolean', desc: 'Fail with exit code 1 in case of diff' })
       .option('processed', { type: 'boolean', desc: 'Whether to compare against the template with Transforms already processed', default: false })
-      .option('quiet', { type: 'boolean', alias: 'q', desc: 'Do not print stack name and default message when there is no diff to stdout', default: false }))
+      .option('quiet', { type: 'boolean', alias: 'q', desc: 'Do not print stack name and default message when there is no diff to stdout', default: false })
+      .option('change-set', { type: 'boolean', desc: 'Whether to create a changeset to analyze resource replacements', default: true }))
```
rix0rrr, line 266: This uses the deploy role to do it -- did the previous `GetTemplate` also use the deploy role? (If yes, nothing to do, but otherwise you probably need to add that in as a warning)

rix0rrr: Also add in that this is slower, and make sure we print a message while we're waiting for `createchangeset` to happen. It can easily take 10-30s, and previously `cdk diff` was nearly instantaneous.

rix0rrr (changes requested):
I like this change! Got a couple of things for you to have a mull over.

One big thing that came up during a recent discussion, that I'm not sure Hogan talked to you about: what about nested stacks? Does `cdk diff` honor nested stacks today? If yes, make sure that the changeset variant does it as well.

packages/aws-cdk/lib/diff.ts
comcalvi: This function is never called after `diff`, it's invoked directly by the toolkit. Could you elaborate on how the diff is happening twice?

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
comcalvi: The second approach was much easier than my first approach 😅

packages/@aws-cdk/cloudformation-diff/lib/diff/index.ts
```diff
@@ -39,35 +45,67 @@ export function diffResource(oldValue?: types.Resource, newValue?: types.Resourc
...
+  ) {
+    let changeImpact: types.ResourceImpact = types.ResourceImpact.NO_CHANGE;
+    if (!changeSetReplacement) {
+      changeImpact = _resourceSpecImpact(oldV, newV, key, resourceSpec);
+      return new types.PropertyDifference(oldV, newV, { changeImpact });
+    }
+
+    if (changeSetReplacement === RESOURCE_NOT_IN_CHANGE_SET || !(key in changeSetReplacement.propertiesReplaced)) {
+      // The changeset does not contain this property, which means it is not going to be updated. Hide this cosmetic template-only change from the diff
+      return new types.PropertyDifference(1, 1, { changeImpact });
```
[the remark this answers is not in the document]
comcalvi: The 1,1 deliberately obscures the difference between them, since `deepEqual(1,1)` is true.

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
```diff
@@ -184,3 +198,30 @@ function deepCopy(x: any): any {
...
+          // If Evaluation is 'Dynamic', then this may cause replacement, or it may not.
+          // see 'Replacement': https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_ResourceChange.html
+          propertiesReplaced[propertyChange.Target.Name!] = 'Conditionally';
+        } else {
+          propertiesReplaced[propertyChange.Target.Name!] = propertyChange.Target.RequiresRecreation as ChangeSetReplacement;
+        }
+      }
+    }
+    replacements[resourceChange.ResourceChange?.LogicalResourceId!] = {
+      resourceReplaced: resourceChange.ResourceChange?.Replacement === 'True',
```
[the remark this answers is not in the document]
comcalvi: Yep! See `Replacement`: https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_ResourceChange.html

comcalvi:
The changeset fails to create for nested stacks, `CreateChangeSet` returns

```
Template error: Fn::Equals cannot be partially collapsed
```

I'm not sure what's causing this, but now we catch this early and fall back to the regular diff for nested stacks.

packages/@aws-cdk/cloudformation-diff/lib/diff/index.ts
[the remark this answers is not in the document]
rix0rrr: Surely it will not be printed since `changeImpact` is a value that means "don't render this" ?

If your goal was: *"I will put incorrect values into the data model because I know exactly what downstream code will do with these values"*... then congratulations: you've now made a system where different pieces cannot be updated individually anymore, because you need to know how every little piece interacts with every other piece.

I'd rather have pieces that are *locally correct*, and when composed are also *globally correct*.

- It is not *locally correct* to say that the old value and the new value are both `1`. That's lying, and if we try to consume the diff in some other way later on, we'll need to take into account that the model can be lying to us, and we'll need to program in exceptions checking for a value of `1`. (Will we go and refactor later? Fat chance. Nobody dares touching upstream code, especially if it's doing something crazy. Who knows, there might be a reason to do the madness. The poor sap writing a new consumer for this data will dutifully put in a `if (diff.oldV == 1 && diff.newV == 1) { /* do something */ }` and get on with their life, and the madness proliferates).
- If you want to imply certain behavior for downstream systems, and you don't have the fields necessary to express what you want right now, then *add a field*. If changing the interpretation of `changeImpact` doesn't do it for you, it's always better to add `dontRenderThis?: boolean` to the structure than to start putting sentinel values in.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -130,10 +132,20 @@ export class CdkToolkit {
       if (!await fs.pathExists(options.templatePath)) {
         throw new Error(`There is no file at ${options.templatePath}`);
       }
+
+      const changeSet = options.changeSet ? await PrepareAndCreateDiffChangeSet({
```
rix0rrr, line 136:
```suggestion
      const changeSet = options.changeSet ? await prepareAndCreateDiffChangeSet({
```

What is the difference between the verbs `prepare` and `create` ?

packages/aws-cdk/lib/diff.ts
rix0rrr: Hmm, I guess you're right. I must have misread.

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
```diff
@@ -184,3 +208,98 @@ function deepCopy(x: any): any {
...
+            let str = undefined;
+            if (Array.isArray(value.oldValue) && typeof value.newValue === 'string') {
+              array = value.oldValue;
+              str = value.newValue;
+            } else if (typeof value.oldValue === 'string' && Array.isArray(value.newValue)) {
+              str = value.oldValue;
+              array = value.newValue;
+            }
+            if (array && array.length === 1 && str) {
+              change.setOtherChange('DependsOn', new types.Difference<string>(str, array[0]));
```
rix0rrr, line 252: Shouldn't this have been part of `normalize` ?

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
```diff
@@ -184,3 +208,98 @@ function deepCopy(x: any): any {
 
   return x;
 }
+
+function filterFalsePositivies(diff: types.TemplateDiff, changeSet: CloudFormation.DescribeChangeSetOutput) {
+  const replacements = findResourceReplacements(changeSet);
```
rix0rrr, line 213: Looks like for every change we are going to see if it's a replacement or not. 

What happens to additions and deletions?

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
```diff
@@ -184,3 +208,98 @@ function deepCopy(x: any): any {
...
+  diff.resources.forEachDifference((logicalId: string, change: types.ResourceDifference) => {
+    if (!replacements[logicalId] && !diff.resources.get(logicalId).otherChanges.DependsOn) {
```
rix0rrr, line 215: This needs a comment, not sure what's happening here.

packages/aws-cdk/lib/cdk-toolkit.ts
comcalvi: `prepare` means we're preparing an SDK with the correct role, making the body parameter, and checking CFN to see if the stack exists. 

`create` is just calling `CreateChangeSet` and cleaning up the changeset

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
comcalvi: Our schema-based diff is always correct about resource additions / deletions, so I didn't check the change set for those.

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
comcalvi: ahh good catch, this is a leftover

packages/@aws-cdk/cloudformation-diff/lib/diff/index.ts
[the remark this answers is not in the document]
comcalvi: thanks for the breakdown, this makes sense. I've gone with the field approach (turns out we already had one, it just wasn't writable)

packages/aws-cdk/lib/cdk-toolkit.ts
rix0rrr: Prepare is not useful to the caller of this function though, is it?

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
rix0rrr: Do you have a test on that?

rix0rrr (approved):

PR #28337, opened 2023-12-12
feat(eks): Allow atomic flag for helm chart

alexandersperling:
This PR adds the possibilty to use the atomic flag for helm charts.

Basically I took code from an PR which was closed sometime ago, and refactored it.

Add atomic flag for helm chart construct
Update unit test
Update README

Closes https://github.com/aws/aws-cdk/issues/22254


alexandersperling:
Clarification Request - I added an integration test case but when running this test it fails with an error for the 'non ecr oci chart' test case.

Bildschirmfoto 2023-12-13 um 10 32 34

How can this be fixed or could someone run this integration test?

sumupitchayan:
@alexandersperling the integ also failed when I tried running it. Are you sure that this change is supported by EKS / CloudFormation?

alexandersperling:
@sumupitchayan I'm not sure, usually the integ test should show. Unfortunately it fails on a prior test already so that we can not see the result of the new one.

PR #28346, opened 2023-12-13
feat(cloud9-alpha): support image ids for Amazon Linux 2023 and Ubuntu 22.04

go-to-k:
This PR supports new image ids for Amazon Linux 2023 and Ubuntu 22.04.

As for Amazon Linux 2023, it is not yet in the documentation or release notes, but I tried it in CloudFormation with success, so I added it. We can also check it from the Cloud9 console. If we should still wait, I will remove it.

cloud9-al-2023

And I added `@deprecated` to the Ubuntu 18.04, see:

> Since Ubuntu 18.04 has ended standard support as of May 31, 2023, we recommend you choose Ubuntu 22.04.

https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-resource-cloud9-environmentec2.html#cfn-cloud9-environmentec2-imageid

----


go-to-k:
Exemption Request: This PR just adds new enum props, so I exempted them.

kaizencc (approved):

Issue #28170, opened 2023-11-28
aws-cdk: cdk ls throws RuntimeError: Configure 'allowAllOutbound' directly on the supplied SecurityGroups.


badaldavda8:
### Describe the bug

We have the gitlab cicd pipeline which runs a docker based out of python-3.8-alpine image.

We perform *cdk ls* operation and it fails with following error:

```
fn = alambda.DockerImageFunction(
  File "/usr/local/lib/python3.8/site-packages/jsii/_runtime.py", line 118, in __call__
    inst = super(JSIIMeta, cast(JSIIMeta, cls)).__call__(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/aws_cdk/aws_lambda/__init__.py", line 26367, in __init__
    jsii.create(self.__class__, self, [scope, id, props])
  File "/usr/local/lib/python3.8/site-packages/jsii/_kernel/__init__.py", line 334, in create
    response = self.provider.create(
  File "/usr/local/lib/python3.8/site-packages/jsii/_kernel/providers/process.py", line 365, in create
    return self._process.send(request, CreateResponse)
  File "/usr/local/lib/python3.8/site-packages/jsii/_kernel/providers/process.py", line 342, in send
    raise RuntimeError(resp.error) from JavaScriptError(resp.stack)
RuntimeError: Configure 'allowAllOutbound' directly on the supplied SecurityGroups.
Subprocess exited with error 1
```

On removing allowAllOutbound from my SecurityGroups, it works as expected.

[allowAllOutbound](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda/FunctionOptions.html#aws_cdk.aws_lambda.FunctionOptions.allow_all_outbound) with a value of true is supposed to be the default anyway.

Similar Discussion here - https://github.com/aws/aws-cdk/discussions/27669

### Expected Behavior

Since its default value, it should not be failing anyway on explicitly specifying it.

### Current Behavior

```
fn = alambda.DockerImageFunction(
  File "/usr/local/lib/python3.8/site-packages/jsii/_runtime.py", line 118, in __call__
    inst = super(JSIIMeta, cast(JSIIMeta, cls)).__call__(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/aws_cdk/aws_lambda/__init__.py", line 26367, in __init__
    jsii.create(self.__class__, self, [scope, id, props])
[...]

PR #28367, opened 2023-12-14
docs(lambda): clarify FunctionOptions.allowAllOutbound property

sakurai-ryo:
The following PR adds validation for the case when `allowAllOutbound` and `securityGroups` are specified at the same time in `FunctionOptions`.
https://github.com/aws/aws-cdk/pull/26528
(https://github.com/aws/aws-cdk/pull/27157)

According to related issues and discussions, this PR causes existing Lambda deployments to fail.
However, since this change has already been merged and I think it is the correct change, I did not fix the validation process but added documentation to clarify the behavior.

Relates to #28170, #27669 

----


scanlonp (dismissed):
Can we add a guidance sentence? Maybe "Instead configure \'allowAllOutbound\' directly on the security group".
But good addition!

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -260,6 +260,8 @@ export interface FunctionOptions extends EventInvokeConfigOptions {
    * If set to false, you must individually add traffic rules to allow the
    * Lambda to connect to network targets.
    *
+   * Do not specify this property if the `securityGroups` or `securityGroup` property is set.
```
kaizencc, line 263:
```suggestion
   * Do not specify this property if the `securityGroups` or `securityGroup` property is set.
   * Instead, configure `allowAllOutbound` directly on the security group
```

kaizencc (approved):
Agree with @scanlonp, but I just added it in since its simple enough.

kaizencc (approved):

PR #28368, opened 2023-12-14
chore: add $or operator to the EventPattern interface

b-dur:
This PR adds a missing `$or` comparison operator to the EventPattern interface. As describe in the [doc](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html) the `$or` is not limited to only be on the leaf-node.

Before this change the TS compiler would throw `Object literal may only specify known properties, and '$or' does not exist in type 'EventPattern'` Error. By adding `// @ts-ignore` one could get ride of the TS error and the code/rule would work as expected.

With this change the TS compiler no longer raises this error and `$or` is an excepted property on the EventPattern interface.

**Integration test was not run.**

It is uncertain what prefix should have been used for this PR. Is _Fix_. more accurate? 

----

### DIFF
--- a/packages/aws-cdk-lib/aws-events/lib/event-pattern.ts
+++ b/packages/aws-cdk-lib/aws-events/lib/event-pattern.ts
@@ -350,4 +350,12 @@ export interface EventPattern {
    * @default - No filtering on detail
    */
   readonly detail?: { [key: string]: any };
+
+  /**
+   * Comparison operator to use on the root and/or leaf level for multiple 
+   * fields OR comparison.
+   * 
+   * @default - No OR operator
+   */
+  readonly $or?: EventPattern[];
 }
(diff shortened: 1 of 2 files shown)


b-dur:
The error seeing now is: 
```
 error JSII8002: Method and property (unless they are static readonly) names must use camelCase. Rename "aws-cdk-lib.aws_events.EventPattern.$or" to "or"
```
I wonder if there is a way to avoid this? Or can `or` be used instead of `$or` ?

kaizencc (changes requested):
I see the docs that say "Comparison operators only work on leaf nodes, with the exception of $or and anything-but." so I'm willing to entertain this. This PR needs:

- the property can't be named `$or`. This really means 'anyOf' right? Lets name it that
- integ test demonstrating that we can use this property at the root level.

PR #28379, opened 2023-12-15
fix(secretsmanager): hosted rotation with fromSecretNameV2() does not create correct iam policy

### LINKED ISSUE #28308: secrets: addRotationSchedule() with fromSecretNameV2() does not create correct iam policy for lambda
### Describe the bug

Given the code:

```ts
    const alertsServiceCredentials = new secrets.Secret(this, `secret-rotated`, {
      generateSecretString: {
        secretStringTemplate: JSON.stringify({
          username: 'service_rw_',
          engine: 'mysql',
          host: 'dummy-host',
          port: '3306',
          dbname: 'dummy-db'
        }),
        generateStringKey: 'password'
      }
    });

    let adminLogin = secrets.Secret.fromSecretNameV2(this, 'lookup-admin-login', 'MySecret');

    const schedule = alertsServiceCredentials.addRotationSchedule(
      'rotation-schedule',
      {
        automaticallyAfter: Duration.days(30),
        hostedRotation: secrets.HostedRotation.mysqlMultiUser({
          vpc,
          masterSecret: adminLogin
        }),
      }
    );

```

This will create an IAM policy statement for lambda that only allows to `GetSecretValue` on partial secret arn without trailing 6 characters hence permission denied.

```json
{
    "Statement": [
        {
            "Action": [
                "secretsmanager:GetSecretValue"
            ],
            "Resource": "arn:aws:secretsmanager:us-east-1:123456789012:secret:MySecret",
            "Effect": "Allow"
        }
    ]
}
```

### Expected Behavior

should render correct IAM policy statement

### Current Behavior

does not render correct IAM policy statement for Lambda role.

### Reproduction Steps

see the code above

### Possible Solution

1. use `fromSecretCompleteArn()` to generate full secret arn in the iam policy
2. use escape hatches to append `-??????` to the ARN

For example

```ts
    const schedule = alertsServiceCredentials.addRotationSchedule(
      'rotation-schedule',
      {
        automaticallyAfter: Duration.days(30),
        hostedRotation: secrets.HostedRotation.mysqlMultiUser({
          vpc,
          masterSecret: adminLogin
        }),
      }
    );

    const cfnRotationSchedule = schedule.node.tryFindChild('Resource') as secrets.CfnRotationSchedule
    cfnRotationSchedule.addPropertyOverride('HostedRotationLambda.MasterSecretArn', `${adminLogin.secretArn}-??????`)
```
        
And when you run `cdk diff` we should see this

    
```
Resources
[~] AWS::SecretsManager::RotationSchedule secret-rotated/rotation-schedule secretrotatedrotationscheduleFF5ED1F3 
 └─ [~] HostedRotationLambda
     └─ [~] .MasterSecretArn:
         ├─ [-] arn:aws:secretsmanager:us-east-1:123456789012:secret:MySecret
         └─ [+] arn:aws:secretsmanager:us-east-1:123456789012:secret:MySecret-??????
```
    
    

_No response_

### CDK CLI Version

2.113.0

### Framework Version

_No response_

### Node.js Version

v18.16.0

### OS

mac

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

go-to-k:
In the case for a hosted rotation with a master secret by `fromSecretNameV2()`, IAM policy for lambda is not correct.

The secret by the method has a partial ARN, so a generated IAM policy should use an ARN with `'-??????'`, but it will not use the ARN.

Closes #28308.

----


packages/aws-cdk-lib/aws-secretsmanager/lib/rotation-schedule.ts
```diff
@@ -325,14 +325,21 @@ export class HostedRotation implements ec2.IConnectable {
       this.masterSecret.denyAccountRootDelete();
     }
 
+    let masterSecretArn: string | undefined;
+    if (this.masterSecret?.secretFullArn) {
+      masterSecretArn = this.masterSecret.secretArn;
+    } else if (this.masterSecret) { // ISecret as an imported secret with partial ARN
+      masterSecretArn = this.masterSecret.secretArn + '-??????';
+    }
```
go-to-k, line 333: The case where `secretArn` and `secretName` are Token is also taken into account here.

- `if (this.masterSecret?.secretFullArn) {`
  - If we create a secret construct in the stack, here is passed.
  - Otherwise `this.masterSecret` is from `fromSecretArn`, `fromSecretCompleteArn` methods, etc...
    - This `if` statement will be true even if the arg is a token or not.

go-to-k: I'll think about this some more, so I'll draft it once.

go-to-k, line 333: - `if (this.masterSecret?.secretFullArn)`
  - If we generate a secret construct for `this.masterSecret` in the stack, here is true.
  - Otherwise `this.masterSecret` is generated from `fromSecretArn()`, `fromSecretCompleteArn()` methods, etc. They also have a full ARN.
    - This `if` statement will be true even if the arg (so it is the arn) for the methods is a token or not.
- `else if (this.masterSecret)`
  - The secret is `ISecret` as an imported secret with partial ARN. It is generated from `fromSecretNameV2()`.
  - The `secretArn` in this block will be a token. But processing this string is not a problem because the CloudFormation template can use `Fn::Join` to combine them properly. (see the unit tests.)

paulhcsun (approved):
Thanks for this fix @go-to-k!

PR #28387, opened 2023-12-15
feat(cli): option to ignore no stacks

### LINKED ISSUE #28371: (cli): Provide a way to ignore "This app contains no stacks"
### Describe the feature

It would be nice if I could ignore the "This app contains no stacks" error message specifically. 

### Use Case

In many cases, I have a dev and prod AWS account. When I first develop a new CDK project, I often want to only deploy the dev account, while later deploying to dev and prod. I often write this code like this:

```ts
#!/usr/bin/env node
import { TestCdkStack } from '../lib/test-cdk-stack';

const app = new cdk.App();

// Assume 12345678910 is my dev account number
if (process.env.CDK_DEFAULT_ACCOUNT === "12345678910") {
	new TestCdkStack(app, 'TestCdkStack', {
	  /* If you don't specify 'env', this stack will be environment-agnostic.
	   * Account/Region-dependent features and context lookups will not work,
	   * but a single synthesized template can be deployed anywhere. */
	
	  /* Uncomment the next line to specialize this stack for the AWS Account
	   * and Region that are implied by the current CLI configuration. */
	  // env: { account: process.env.CDK_DEFAULT_ACCOUNT, region: process.env.CDK_DEFAULT_REGION },
	
	  /* Uncomment the next line if you know exactly what Account and Region you
	   * want to deploy the stack to. */
	  // env: { account: '123456789012', region: 'us-east-1' },
	
	  /* For more information, see https://docs.aws.amazon.com/cdk/latest/guide/environments.html */
	});
}
```

However, this fails with the error message "This app contains no stacks".

You might think, "why not just set up deployment to not happen until you're ready?". My common use case is a monorepo with many projects. My CI uses `lerna` to run `cdk diff` automatically in projects that changed.

### Proposed Solution

A CLI flag like `--ignore-no-stacks` would be great. This reminds me of `jest`'s `--passWithNoTests` flag that's helpful when you've installed jest but not written tests yet.

### Other Information

I've hacked my way around this by creating this script:

```bash
#! /bin/bash
# Working around this issue: https://github.com/aws/aws-cdk/issues/28371

set -e

if [ "$CI" ]  && [ "$CDK_STAGE" == 'dev' ]; then
  echo "This stack does not deploy to dev"
  exit 0
else
  cdk "$@"
fi

```

and updating the `package.json`'s `cdk` script to call this instead.

### Acknowledgements

### CDK version used

2.115.0

MacOS Latest

msambol:
I'm new to development on this package—any feedback regarding testing is appreciated.

Closes #28371.

----


msambol:
Still working through this, need to add tests.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli-lib.integtest.ts
```diff
@@ -47,6 +47,28 @@ integTest('cli-lib deploy', withCliLibFixture(async (fixture) => {
...
+    });
+    expect(expectedStack.StackResources?.length).toEqual(0);
+  } finally {
+    // delete the stack
+    await fixture.cdk(['destroy', stackName], {
+      captureStderr: false,
+    });
+  }
+}));
+
```
msambol, line 71: should the integ tests go in this file?

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -780,6 +780,31 @@ integTest('deploy stack without resource', withDefaultFixture(async (fixture) =>
...
+    modEnv: {
+      INTEG_STACK_SET: 'stage-with-no-resources',
+    },
+  });
+
+  // verify that we only deployed both stacks (there are 2 ARNs in the output)
+  /* eslint-disable no-console */
+  console.log(stackArn);
+}));
+
```
msambol, line 807: or do the integ tests belong in this file?

msambol:
Exemption Request: I added a CLI integ test instead of framework-integ, because this is a change of CLI.

msambol:
@lpizzinidev When you have time could mind reviewing?

packages/aws-cdk/lib/cli.ts
```diff
@@ -172,7 +172,8 @@ async function parseCommandLineArguments(args: string[]) {
       })
       .option('concurrency', { type: 'number', desc: 'Maximum number of simultaneous deployments (dependency permitting) to execute.', default: 1, requiresArg: true })
       .option('asset-parallelism', { type: 'boolean', desc: 'Whether to build/publish assets in parallel' })
-      .option('asset-prebuild', { type: 'boolean', desc: 'Whether to build all assets before deploying the first stack (useful for failing Docker builds)', default: true }),
+      .option('asset-prebuild', { type: 'boolean', desc: 'Whether to build all assets before deploying the first stack (useful for failing Docker builds)', default: true })
+      .option('ignore-no-stacks', { type: 'boolean', desc: 'Ignore the error message if the app contains no stacks', default: false }),
```
lpizzinidev, line 176:
```suggestion
      .option('ignore-no-stacks', { type: 'boolean', desc: 'Whether to deploy if the app contains no stacks', default: false }),
```

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -175,7 +175,8 @@ export class CdkToolkit {
     }
 
     const startSynthTime = new Date().getTime();
-    const stackCollection = await this.selectStacksForDeploy(options.selector, options.exclusively, options.cacheCloudAssembly);
+    // eslint-disable-next-line max-len
```
lpizzinidev, line 178:
```suggestion
```
We should just take linter's advice and split the next line into multiple lines (same below).

packages/aws-cdk/lib/api/cxapp/cloud-assembly.ts
```diff
@@ -100,7 +107,11 @@ export class CloudAssembly {
     const patterns = sanitizePatterns(selector.patterns);
 
     if (stacks.length === 0) {
-      throw new Error('This app contains no stacks');
+      if (options.ignoreNoStacks) {
+        return new StackCollection(this, []);
+      } else {
+        throw new Error('This app contains no stacks');
+      }
```
lpizzinidev, line 114:
```suggestion
      throw new Error('This app contains no stacks');
```
No need for an `else`

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli-lib.integtest.ts
```diff
@@ -47,6 +47,28 @@ integTest('cli-lib deploy', withCliLibFixture(async (fixture) => {
   }
 }));
 
+integTest('cli-lib deploy no stack', withCliLibFixture(async (fixture) => {
```
lpizzinidev, line 50: We should also add the `throwing` case for coverage.

packages/@aws-cdk-testing/cli-integ/lib/with-cli-no-stacks.ts
```diff
@@ -0,0 +1,134 @@
...
+import { TestContext } from './integ-test';
+import { RESOURCES_DIR } from './resources';
+import { AwsContext, withAws } from './with-aws';
+import { cloneDirectory, installNpmPackages, TestFixture, DEFAULT_TEST_TIMEOUT_S, CdkCliOptions } from './with-cdk-app';
+import { withTimeout } from './with-timeout';
+
+/**
+ * Higher order function to execute a block with a CliLib Integration CDK app fixture
+ */
+export function withCliLibIntegrationCdkApp<A extends TestContext & AwsContext>(block: (context: CliLibIntegrationTestFixture) => Promise<void>) {
```
lpizzinidev, line 12: Should use a different name than `with-cli-lib`

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -780,6 +780,19 @@ integTest('deploy stack without resource', withDefaultFixture(async (fixture) =>
     .rejects.toThrow('conditional-resource does not exist');
 }));
 
+integTest('deploy --ignore-no-stacks', withDefaultFixture(async (fixture) => {
```
lpizzinidev, line 783: Not sure about this.
We should verify that the app deploys when there are no stacks in it.

lpizzinidev (changes requested):
The implementation looks good.
I left some comments for improvements 👍

msambol:
@lpizzinidev I tried to simplify this PR for now. It feels like there are so many options and files around the CLI, and I'm not deep enough in this area. Hopefully this is headed in the right direction.

lpizzinidev (changes requested):
~~Can you please add back the integration tests in [cli-lib.integtest.ts](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli-lib.integtest.ts)? (both happy and failure cases)
Then this will be good imo.~~

lpizzinidev (approved):
On second thought, the integration tests in `cli.integtest.ts` should provide enough coverage here.
Thanks for the changes 👍

kaizencc (approved):

blimmer:
Thank you @msambol !!!

PR #28388, opened 2023-12-15
feat(pipes): EventBridge Pipes alpha module

### LINKED ISSUE #23495: (pipes): Add L2 support for Amazon EventBridge Pipes
### Describe the feature

Add L2 constructs for Amazon EventBridge Pipes

### Use Case

N/A

### Proposed Solution

N/A

### Other Information

- Announcement: https://aws.amazon.com/blogs/aws/new-create-point-to-point-integrations-between-event-producers-and-consumers-with-amazon-eventbridge-pipes/?sc_channel=el
- Landing: https://aws.amazon.com/eventbridge/pipes/
- CF: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-pipes-pipe.html

### Acknowledgements

### CDK version used

N/A

N/A

RaphaelManke:
This PR is the starting point of the implementation for a L2 construct as defined in https://github.com/aws/aws-cdk-rfcs/issues/473

In this PR the basic Pipe class is introduced including the api interfaces for how to define a pipe. 

Closes #23495

----


.gitignore
```diff
@@ -51,3 +51,6 @@ RELEASE_NOTES.md
 
 # Produced by integ tests
 read*lock
+
+# VSCode jest plugin
+.test-output
```
mrgrain, line 56:
```suggestion
.test-output

```

packages/@aws-cdk/aws-pipes-alpha/lib/enrichment.ts
```diff
@@ -0,0 +1,112 @@
...
+export abstract class Enrichment implements IEnrichment {
+  public readonly enrichmentArn: string;
+  public readonly enrichmentParameters: EnrichmentParametersProperty;
+
+  constructor(
+    enrichmentArn: string,
+    props: EnrichmentParametersProperty,
+  ) {
+    this.enrichmentParameters = props;
+    // TODO - validate ARN is a valid enrichment ARN based on regex from cfn
```
mrgrain, line 105: Please remove todo. Either just implement this or create an issue in the backlog with a label for the module. Before stabilization we will go through all issue in the backlog.

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,55 @@
+
+enum reservedVariables {
```
mrgrain, line 2:
```suggestion
enum ReservedVariables {
```

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,55 @@
...
+  PIPES_ARN = '<aws.pipes.pipe-arn>',
+  PIPES_NAME = '<aws.pipes.pipe-name>',
+  PIPES_TARGET_ARN = '<aws.pipes.target-arn>',
+  PIPE_EVENT_INGESTION_TIME = '<aws.pipes.event.ingestion-time>',
+  PIPE_EVENT = '<aws.pipes.event>',
+  PIPE_EVENT_JSON = '<aws.pipes.event.json>',
+}
+
+type StaticString = string;
+// type JsonPath = `<$.${string}>`;
```
mrgrain, line 12: Please remove or use. Create an todo issue if appropriate.

packages/@aws-cdk/aws-pipes-alpha/lib/filterPattern.ts
```diff
@@ -0,0 +1,14 @@
+import { IFilterPattern } from './filter';
+
+/**
+ * Generate a filter pattern from an input.
+ */
+
+export class FilterPattern {
```
mrgrain, line 7:
```suggestion
/**
 * Generate a filter pattern from an input.
 */
export class FilterPattern {
```

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,55 @@
...
+   * @param inputTemplate
+   * @returns
+   */
+  static fromJson(inputTemplate: Record<string, any>): InputTransformation {
+    return new InputTransformation(inputTemplate);
+  }
+
+  inputTemplate: string;
+
+  constructor(inputTemplate: InputTransformationValue) {
```
mrgrain, line 44: Should this be a private constructor? If not, we have a problem with the types further up.

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Builds an input transformation from a JSON object.
+   * @param inputTemplate
+   * @returns
+   */
+  static fromJson(inputTemplate: Record<string, any>): InputTransformation {
+    return new InputTransformation(inputTemplate);
+  }
+
+  inputTemplate: string;
```
mrgrain, line 42:
```suggestion
  public readonly inputTemplate: string;
```

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,55 @@
...
+type InputTransformationValue = StaticString | InputTransformJson;
+
+/**
+ * Transform or replace the input event payload
+ */
+export interface IInputTransformation {
+  /**
+   * Valid JSON text passed to the target.
+   */
+  inputTemplate: string;
```
mrgrain, line 26: pubic readonly inputTemplate: string;
```suggestion
  readonly inputTemplate: string;
```

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,55 @@
...
+type InputTransformJson = Record<string, StaticString | KeyValue | StaticJsonFlat>;
+
+type InputTransformationValue = StaticString | InputTransformJson;
+
+/**
+ * Transform or replace the input event payload
+ */
+export interface IInputTransformation {
+  /**
+   * Valid JSON text passed to the target.
```
mrgrain, line 24: Not sure this is the correct description, since it doesn't have to be valid json. All the `dynamic` json stuff won't be valid json.
https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-input-transformation.html

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.inputTemplate = this.unquoteKeyPlaceholders(inputTemplate);
+  }
+
+  private unquoteKeyPlaceholders(obj: any) {
+    const jsonString = JSON.stringify(obj);
+
+    const result = jsonString.replace(/"(<(?:\$\.|aws\.pipes)[^"]*?)"/g, '$1'); // Retain the "<>" and remove the outer quotes for the values that start with either "<$." or "<aws.pipes"
+
+    return result;
+  }
```
mrgrain, line 54: Don't really like this. I think we need a variable class that can be used as a variable and then manually build the string.

mrgrain (changes requested):
Looks good! Left a bunch of minor comments. The biggest issue is around the uplifted L1 types for `Source` `Target` etc. I think I was lacking some context/understanding when you asked me about this.

Let's use `Source` as an example.

So instead of defining all possible props on `ISource`, we define a method that returns the equivalent L1 interface. Each source will then implement this and return the correct data. Instead of spreading on `source.sourceParameters` you can spread on `source.attach(this)` which is a function that returns the correct L1 interface and also calls `source.grantRead(pipe.pipeRole)` (rather than the other way around. 

This is an example of the pattern: https://github.com/aws/aws-cdk/blob/v2.117.0/packages/aws-cdk-lib/aws-elasticloadbalancingv2-targets/lib/lambda-target.ts

Hope this makes sense.

packages/@aws-cdk/aws-pipes-alpha/lib/targetParameter.ts
```diff
@@ -0,0 +1,59 @@
+/** Reserved pipe variables
+* @see https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-input-transformation.html#input-transform-reserved
+*/
+export enum PipeVariables {
```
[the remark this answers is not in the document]
RaphaelManke: none. Its a dublicate.

packages/@aws-cdk/aws-pipes-alpha/lib/targetParameter.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/**
+ * Helper class to generate dynamic target parameters.
+ */
+export class TargetParameter {
+  /**
+   * Reserved pipe variables
+   */
+  static pipeVariable = PipeVariables;
```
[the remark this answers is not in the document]
RaphaelManke: Yeah, that doesn't make sense that way. I'll remove it completly for now. And just export the enum.

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
RaphaelManke: removed

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
RaphaelManke: I will have another look at it. 
I think the inputTransformation should be build similar to the `RuleTargetInput` class from the EventBridge rules. 
https://github.com/aws/aws-cdk/blob/v2.117.0/packages/aws-cdk-lib/aws-events/lib/input.ts#L10

But I have to get my head around the implementation there first.

packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
```diff
@@ -0,0 +1,284 @@
...
+ * @see https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes.html
+ */
+export class Pipe extends PipeBase {
+  /**
+   * Reference an existing pipe by ARN
+   */
+  public static fromPipeArn(scope: Construct, id: string, pipeArn: string): IPipe {
+    // TODO implement fromPipeArn function
+    throw Error(`Not implemented! ${scope} ${id} ${pipeArn}`);
+  }
```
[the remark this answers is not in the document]
RaphaelManke: will remove it for now, and add it if applicable later. 
Currently I have no concept how the api for "referenced" resources should look like.

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
```diff
@@ -0,0 +1,744 @@
...
+//   readonly batchSize?: Duration;
+
+//   /**
+//      * The maximum length of a time to wait for events.
+//      * Must be between 0 and 300 seconds.
+//      * TODO: check what is the default
+//      * @default - 300
+//      */
+//   readonly maximumBatchingWindow?: Duration;
+// }
```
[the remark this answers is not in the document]
RaphaelManke: There are some properties that are available for all sources. But I think its a premature optimisation and therefore remove it for now.

packages/@aws-cdk/aws-pipes-alpha/README.md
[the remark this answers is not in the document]
RaphaelManke: Do you have a reference or docu how to write these code examples? 
Are they just markdown code snippets? I think they have to be marked somehow so they can be translated to the other programming languages automatically.

packages/@aws-cdk/aws-pipes-alpha/rosetta/default.ts-fixture
```diff
@@ -0,0 +1,14 @@
+// TODO
```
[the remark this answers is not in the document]
RaphaelManke: Can you point me into the right direction how to write an example?

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
RaphaelManke: Renamed to `PipeVariable` and dropped the prefix.

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
RaphaelManke: I will try to refactor it so that it is similar to https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_events.RuleTargetInput.html

packages/@aws-cdk/aws-pipes-alpha/test/pipe.test.ts
```diff
@@ -0,0 +1,543 @@
...
+        grantRead: () => { },
+        sourceArn: 'source-arn',
+        sourceParameters: {},
+
+      },
+      target: {
+        grantPush: () => { },
+        targetArn: 'target-arn',
+        targetParameters: {},
+      },
```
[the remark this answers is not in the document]
RaphaelManke: I introduced test classes, if this is ok we can mark the comment as resolved.

packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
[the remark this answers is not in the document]
RaphaelManke: Ok awslint requires at least one `fromXxx` method, so I implemented a `fromName` for simplicity reasons.

packages/@aws-cdk/aws-pipes-alpha/jest.config.js
```diff
@@ -0,0 +1,10 @@
+const baseConfig = require('@aws-cdk/cdk-build-tools/config/jest.config');
+module.exports = {
+    ...baseConfig,
+    coverageThreshold: {
+        global: {
+            ...baseConfig.coverageThreshold.global,
+            branches: 60,
```
mrgrain, line 7: Can we get this to the default i.e. 80?

packages/@aws-cdk/aws-pipes-alpha/README.md
mrgrain, line 1: Still need to change this focus on CDK usage.

packages/@aws-cdk/aws-pipes-alpha/awslint.json
```diff
@@ -0,0 +1,5 @@
+{
+  "exclude": [
+    "no-unused-type:@aws-cdk/aws-pipes-alpha.PipeVariable"
```
mrgrain, line 3: I'm okay with this in principle, but why is not used?

packages/@aws-cdk/aws-pipes-alpha/lib/pipeVariable.ts
```diff
@@ -0,0 +1,146 @@
+import { IResolvable, IResolveContext, captureStackTrace } from 'aws-cdk-lib';
+
+/** Reserved pipe variables
+* @see https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-input-transformation.html#input-transform-reserved
+*/
+export enum PipeVariable {
```
mrgrain, line 6: Seems like we don't actually want this to be public.

packages/@aws-cdk/aws-pipes-alpha/lib/pipeVariable.ts
```diff
@@ -0,0 +1,146 @@
...
+    return this.value.toString();
+  }
+
+  /**
+   * Return a JSON representation of a dynamic input.
+   */
+  toJSON(): string {
+    return this.value.toString();
+  }
+}
```
mrgrain, line 146:
```suggestion
}

```

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,115 @@
...
+   */
+  private inputTemplate: InputTransformationValue;
+
+  private constructor(inputTemplate: InputTransformationValue) {
+
+    this.inputTemplate = inputTemplate;
+  }
+
+  public bind(pipe: IPipe): InputTransformationConfig {
+    if (typeof this.inputTemplate === 'string') {
```
mrgrain, line 74: We need to use `!Token.isUnresolved(this.inputTemplate)` instead of `typeof`

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,115 @@
...
+    const resolved = Tokenization.resolve(stringifiedJsonWithUnresolvedTokens, {
+      scope: pipe,
+      resolver: new DefaultTokenResolver(new StringConcat()),
+    });
+    return { inputTemplate: this.unquoteDynamicInputs(resolved) };
+  }
+
+  private unquoteDynamicInputs(sub: string) {
+
+    // TODO can we use the Token.asString() here?
```
mrgrain, line 88: `Lazy.uncachedString` is just a wrapper around `Token.asString()`. This is fine.

packages/@aws-cdk/aws-pipes-alpha/lib/logs.ts
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Bind the log destination to the pipe.
+   */
+  bind(pipe: IPipe): LogDestinationConfig;
+
+  /**
+   * Grant the pipe role to push to the log destination.
+   */
+  grantPush(grantee: IRole): void;
+}
```
mrgrain, line 95:
```suggestion
}

```

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
```diff
@@ -0,0 +1,102 @@
...
+   * The ARN of the source resource.
+   */
+  readonly sourceArn: string;
+
+  /**
+   * The parameters required to set up a source for your pipe.
+   *
+   * @default - none
+   */
+  readonly sourceParameters?: SourceParameters;
```
mrgrain, line 86:
```suggestion
  readonly sourceParameters?: CfnPipes.PipeSourceParametersProperty;
```

packages/@aws-cdk/aws-pipes-alpha/lib/logs.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Log destination configuration.
+ */
+export interface LogDestinationConfig {
+  /**
+   * Get the log destination configuration parameters.
+   */
+  readonly parameters: LogDestinationParameters;
```
mrgrain, line 78:
```suggestion
  readonly parameters: CfnPipe.PipeLogConfigurationProperty;
```

packages/@aws-cdk/aws-pipes-alpha/lib/logs.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Log destination base class.
+ */
+export interface ILogDestination {
+
+  /**
+   * Bind the log destination to the pipe.
+   */
+  bind(pipe: IPipe): LogDestinationConfig;
```
mrgrain, line 89:
```suggestion
  bind(pipe: IPipe): CfnPipe.PipeLogConfigurationProperty;
```

packages/@aws-cdk/aws-pipes-alpha/lib/enrichment.ts
```diff
@@ -0,0 +1,40 @@
...
+export interface IEnrichment {
+  /**
+   * Grant the pipes role to invoke the enrichment.
+   */
+  grantInvoke(grantee: IRole): void;
+
+  /**
+   * Bind this enrichment to a pipe.
+   */
+  bind(pipe: IPipe): EnrichmentParametersConfig;
```
mrgrain, line 38:
```suggestion
  bind(pipe: IPipe): CfnPipe.PipeEnrichmentParametersProperty;
```

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/**
+ * Source interface
+ */
+export interface ISource {
+  /**
+   * Bind the source to a pipe.
+   */
+  bind(pipe: IPipe): SourceConfig;
```
mrgrain, line 96:
```suggestion
  bind(pipe: IPipe): CfnPipes.PipeSourceParametersProperty;
```

mrgrain (dismissed):
For the readme, we should have:
- one basic end to end example (source and target)
- one section per feature, including an example:
  - enrichment
  - filter
  - logs

You should be able to build on the examples in the RFC.

packages/@aws-cdk/aws-pipes-alpha/jest.config.js
mrgrain: Looks like you are at 100% anyway :D

packages/@aws-cdk/aws-pipes-alpha/lib/enrichment.ts
RaphaelManke: I am not sure if that's what we want? The cfn interface is just the parameters for the enrichment. But besides the parameters we need a ARN to the enrichment. 
The interface `EnrichmentParametersConfig` has to attributes. The parameters which are referenced by `CfnPipe.PipeEnrichmentParametersProperty` and an attribute for the enrichment ARN. 
If the bind should just return the parameters, where should the ARN be defined?

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
RaphaelManke: Ok alright, I leave it as is.

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,115 @@
...
+    return Lazy.uncachedString({ produce: (ctx: IResolveContext) => Token.asString(deepUnquote(ctx.resolve(sub))) });
+
+    /**
+     * Removes the quotes from the values that are in the keys array
+     *
+     * @param resolved the resolved object containing the dynamic fields with quotes. In cases where a cloudformation intrinsic function is used, the resolved value will be an object.
+     * @returns the resolved object with the dynamic fields without quotes
+     */
+    function deepUnquote(resolved: any): any {
```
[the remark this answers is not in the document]
RaphaelManke: :heart:

packages/@aws-cdk/aws-pipes-alpha/lib/logs.ts
RaphaelManke: The problem here is that the cfn interface includes the attributes like `includeExecutionData` and `level`. These attributes have been lifted to the pipe class as a property there. If the bind method allows returning theses attributes again, there might happen an unintended overwrite? 
So the idea of the custom interface was to make sure only actual log configurations are returned.

packages/@aws-cdk/aws-pipes-alpha/lib/pipeVariable.ts
RaphaelManke: Ok so you suggest to get rid of the enum and just return the string values directly in the return value of for example `get pipeName ()` ?

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
RaphaelManke: Same here as for the enrichment and target. 
A) we need a source ARN from somewhere.
b) The Cfn interface returns more attributes than the `SourceParameters` e.g. the filter pattern. If we use the cfn interface it might be unclear why there are two ways of defining a filter. The custom interface makes sure only parameters for the source are returned.

packages/@aws-cdk/aws-pipes-alpha/lib/target.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface ITarget {
+  /**
+   * Grant the pipe role to push to the target.
+   */
+  grantPush(grantee: IRole): void;
+
+  /**
+   * Bind this target to a pipe.
+   */
+  bind(pipe: IPipe): TargetConfig;
```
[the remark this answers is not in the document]
RaphaelManke: We need to have a way to get a target ARN.

packages/@aws-cdk/aws-pipes-alpha/awslint.json
RaphaelManke: It's used only in the dame file as its declared. The DynamiInput class uses the enum as type. 
Depending on if we drop the enum entirely, the exclude can be reverted.

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
RaphaelManke: This case actually has nothing todo with unresolved values. I added a testcase to make sure tokens are handled correctly. This check tests if the input template is already a string value and should be left as is. 
But I get that this check is not self-explanatory. Therefore I introduced another internal type variable so that we now if we need todo check for unquoting or not. In case of a plain string value there is no need for an unquote.

packages/@aws-cdk/aws-pipes-alpha/lib/enrichment.ts
mrgrain: Isn't that what `IEnchrichment` is for?

In my head the definition of `bind` is strictly "convert this L2 to the respective properties for the L1`>

packages/@aws-cdk/aws-pipes-alpha/lib/target.ts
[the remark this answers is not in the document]
mrgrain: Yes, `ITarget` should be that place.

packages/@aws-cdk/aws-pipes-alpha/lib/enrichment.ts
RaphaelManke: Ok, I will change that.

packages/@aws-cdk/aws-pipes-alpha/lib/target.ts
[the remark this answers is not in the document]
RaphaelManke: Will move it ti ITarget

packages/@aws-cdk/aws-pipes-alpha/lib/pipeVariable.ts
mrgrain: Yes

mrgrain: Seems like this is redundant API surface for now.

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
mrgrain: Unresolved tokens can be strings though.

packages/@aws-cdk/aws-pipes-alpha/lib/enrichment.ts
mrgrain: Makes sense. Let me look into this rule.

packages/@aws-cdk/aws-pipes-alpha/lib/logs.ts
mrgrain: Okay. Accepted.

packages/@aws-cdk/aws-pipes-alpha/lib/pipeVariable.ts
```diff
@@ -0,0 +1,149 @@
...
+  public static get eventJson (): DynamicInput {
+    return this.fromPipeVariable(PipeVariable.EVENT_JSON);
+  }
+
+  /**
+   * Value from one of the provided Pipe variables.
+  */
+  private static fromPipeVariable(variable: PipeVariable): DynamicInput {
+    return new DynamicInput(variable);
+  }
```
mrgrain, line 118: Plus I think we can make this public and accept a string. That way we are not blocking users of a new variable is added.

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
mrgrain: Yep that's fine.

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
```diff
@@ -0,0 +1,121 @@
+import { DefaultTokenResolver, IResolveContext, Lazy, StringConcat, Token, Tokenization } from 'aws-cdk-lib';
+import { IPipe } from './pipe';
+import { DynamicInput, PipeVariable } from './pipeVariable';
+import { unquote } from './unquote';
+
+type StaticString = string;
```
mrgrain, line 6: Can you loose this indirection, seems unnecessary and confusing.

packages/@aws-cdk/aws-pipes-alpha/lib/enrichment.ts
mrgrain: Yes. Alright then. 😂

packages/@aws-cdk/aws-pipes-alpha/awslint.json
mrgrain: Yes, you should be able to remove this.

packages/@aws-cdk/aws-pipes-alpha/lib/inputTransformation.ts
mrgrain: Yes, that's better.

mrgrain (approved):
🎉

mrgrain:
@mergifyio update

mrgrain:
@mergifyio update

PR #28389, opened 2023-12-15
feat(aws-ec2) export NatGatewayProvider for consistency with NatInstanceProvider

### LINKED ISSUE #28372: aws-ec2: Export NatGatewayProvider
### Describe the feature

(originally posted as [a discussion](https://github.com/aws/aws-cdk/discussions/28154))

There are two implementations of [NatProvider](https://github.com/aws/aws-cdk/blob/c6471f2f603e2570e8781d6e1ae7d8252aec0f79/packages/aws-cdk-lib/aws-ec2/lib/nat.ts#L55-L100): [NatGatewayProvider](https://github.com/aws/aws-cdk/blob/c6471f2f603e2570e8781d6e1ae7d8252aec0f79/packages/aws-cdk-lib/aws-ec2/lib/nat.ts#L217-L261) and [NatInstanceProvider](https://github.com/aws/aws-cdk/blob/c6471f2f603e2570e8781d6e1ae7d8252aec0f79/packages/aws-cdk-lib/aws-ec2/lib/nat.ts#L266-L356).

Of the two, only NatInstanceProvider is exported. Both are usable through `aws_ec2.NatProvider.instance()` and `aws_ec2.NatProvider.gateway()` respectively, but since only NatInstanceProvider is exported [the docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.NatProvider.html) indicate that `NatProvider` is only implemented by `NatInstanceProvider`.

I suggest that NatGatewayProvider also be exported, so that both providers are available through both interfaces and so the docs correctly reflect the two available implementations of `NatProvider`.

### Use Case

No real new functionality is exposed, except the ability to instantiate NatGatewayProvider directly instead of using the static method of NatProvider. The main advantage is the implied update to the docs.

### Proposed Solution

I think this is as simple as inserting an `export` at the start of L217 https://github.com/aws/aws-cdk/blob/c6471f2f603e2570e8781d6e1ae7d8252aec0f79/packages/aws-cdk-lib/aws-ec2/lib/nat.ts#L214-L261

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.115.0

Windows 10

crh23:
Export `NatGatewayProvider` for two reasons: to allow instantiation with `new`, and to make docs present it as an implementation of `NatProvider`.
Also added a unit test for the same "functionality" for `NatInstanceProvider`, for symmetry with the added test for `NatGatewayProvider`. 
Added an integration test, because it appears that the docs pull from them for examples (I don't know how to test this).

Closes #28372.

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.nat-gateway-provider.lit.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.nat-gateway-provider.lit.ts
@@ -1,0 +1,41 @@
+import * as cdk from 'aws-cdk-lib';
+import * as ec2 from 'aws-cdk-lib/aws-ec2';
+import { IntegTest } from '@aws-cdk/integ-tests-alpha';
+
+class NatGatewayStack extends cdk.Stack {
+  constructor(scope: cdk.App, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    /// !show
+    // Explicitly declare the NAT provider, so we can later reference the created gateways
+    const natGatewayProvider = new ec2.NatGatewayProvider();
+
+    const vpc = new ec2.Vpc(this, 'MyVpc', {
+      natGatewayProvider,
+      natGateways: 1,
+    });
+
+    // Change one of the created public subnets to have a route to the NAT Gateway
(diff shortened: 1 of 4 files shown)


crh23:
It turns out that I did not successfully run an integration test. Can't get them to work locally for some reason, I'll try again next year.

PR #28414, opened 2023-12-18
fix(core): messages from `annotations.ts` can show up as `[object Object]`

lmammino:
**CDK Version**: 2.115.0 (build 58027ee)
**Os**: macOS 14.2 (BuildVersion: 23C64)

I have observed the following warning showing up in my console today when running `cdk`:

> [Warning at /CdkStack/AuthorizerFunction] [object Object]

I was able to track down where this message was generated and apply a patch to see the error in a more descriptive format. 

For the records the error in my case was:

> addPermission() has no effect on a Lambda Function with region=${Token[TOKEN.23]}, account=${Token[TOKEN.24]}, in a Stack with region=${Token[AWS.Region.12]}, account=${Token[AWS.AccountId.8]}. Suppress this warning if this is is intentional, or pass sameEnvironment=true to fromFunctionAttributes() if you would like to add the permissions. [ack: UnclearLambdaEnvironment]

The fix proposed here makes sure that if

I am not sure this is the best way to fix this issue. The signature of the `addMessage` seems to expect a `string` for the `message` value, so maybe the error needs to be corrected downstream where the `addMessage` call is made (which judging from the stack trace seems to come from `aws-cdk-lib/aws-lambda/lib/function-base.js`).

Thoughts?

----

### DIFF
--- a/packages/aws-cdk-lib/core/lib/annotations.ts
+++ b/packages/aws-cdk-lib/core/lib/annotations.ts
@@ -143,7 +143,7 @@ export class Annotations {
   private addMessage(level: string, message: string) {
     const isNew = !this.scope.node.metadata.find((x) => x.data === message);
     if (isNew) {
-      let normalizedMessage = typeof message === "string" ? message : JSON.stringify(message);
+      let normalizedMessage = typeof message === 'string' ? message : JSON.stringify(message);
       this.scope.node.addMetadata(level, normalizedMessage, { stackTrace: this.stackTraces });
     }
   }


packages/aws-cdk-lib/core/lib/annotations.ts
```diff
@@ -143,7 +143,8 @@ export class Annotations {
   private addMessage(level: string, message: string) {
     const isNew = !this.scope.node.metadata.find((x) => x.data === message);
     if (isNew) {
-      this.scope.node.addMetadata(level, message, { stackTrace: this.stackTraces });
+      let normalizedMessage = typeof message === "string" ? message : JSON.stringify(message);
```
kaizencc, line 146:
```suggestion
      let normalizedMessage = typeof message === 'string' ? message : JSON.stringify(message);
```

kaizencc (approved):
Thanks @lmammino

PR #28419, opened 2023-12-19
chore(eks): document failure conditions for albController

pahud:
`albController` would fail the deployment if there's no defaultCapacity or nodegroup. This PR improves the doc about it.

Documents issue from https://github.com/aws/aws-cdk/issues/22005 but further effort is needed to create a synth-time error.

----


kaizencc (commented):
1) Does AWS have any documentation of this requirement that we can link to? It seems quite weird.

2) Can we account for this in code and throw a synth-time error?

pahud:
> 1. Does AWS have any documentation of this requirement that we can link to? It seems quite weird.


We checked the [doc](https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html), unfortunately it doesn't mention any about that and according to our [test](https://github.com/aws/aws-cdk/issues/22005#issuecomment-1861860396), the node availability is required for a node-based cluster.

> 2. Can we account for this in code and throw a synth-time error?
Yes we probably can do that but this might take a while for further investigation and experiment. Before that, it would be helpful to add a quick note in the README like this to avoid potential errors.

kaizencc (approved):
@pahud ok, I've approved this as it moves us along. But I am stopping it from closing the linked issue as I think the synth-time error is really what we need. If you think the issue can be closed by this doc update, then go ahead and override my thoughts here, I'm not too integrated into this issue.

pahud:
@kaizencc 

Agree. But there is a case if the cluster and nodegroup are in different stacks, the cluster construct would not be able to check the nodegroup availability in another stack.

Another solution off the top of my head is to expose the `wait` property to the service which default is `true` that blocks the deployment. Making it optionally `false` should not block the stack deployment.

https://github.com/aws/aws-cdk/blob/5d6b8eefc66d326325673bda2b86b72aa4579199/packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts#L312

PR #28424, opened 2023-12-19
fix(integ-tests): apply correct IAM policy to waiterProvider

### LINKED ISSUE #27865: @aws-cdk/integ-tests-alpha: (Integ-test's does not have the proper role permission)
### Describe the bug

@aws-cdk/integ-tests-alpha module
https://docs.aws.amazon.com/cdk/api/v2/docs/integ-tests-alpha-readme.html
```
By default, the AwsApiCall construct will automatically add the correct IAM policies to allow the Lambda function to make the API call. It does this based on the service and api that is provided.
```
During the deployment, there are several lambda function and execution role are auto-generated by CDK.

Stack ARN in case helps for troubleshooting: `arn:aws:cloudformation:us-west-2:673604154507:stack/IntegTestdevDefaultTestDeployAssert31A9C52F/5746bdf0-7d00-11ee-9e23-06a38c3b8f8b`

Inside the state machine, we can see the error
```
AccessDeniedException: User: arn:aws:sts::673604154507:assumed-role/IntegTestdevDefaultTestDe-SingletonFunction76b3e830-5BWQ2YQ5iN2A/IntegTestdevDefaultTestDe-SingletonFunction76b3e83-fLl2b7TfQLmk is not authorized to perform: lambda:InvokeFunction on resource: arn:aws:lambda:us-west-2:673604154507:function:CDKTemplate-Microscope-Scheduler-dev because no identity-based policy allows the lambda:InvokeFunction action
```

However, checking the auto-generated lambda execution role,one of the role `IntegTestdevDefaultTestDe-SingletonFunction76b3e830-QIh9L3ag1iz4	
` has the permission **"lambda:Invoke"**, but there is no **"lambda:InvokeFunction"** permission, another role `IntegTestdevDefaultTestDe-SingletonFunction1488541a-PkTEiGn4sLHb` has both **"lambda:Invoke"** and **"lambda:InvokeFunction"** permission automatically generated

### Expected Behavior

inside AWS console, the below action is an invalid action, and console will throw validation error, may I know if this is a bug in CDK that auto-generate **"lambda:Invoke"** in action rather than **"lambda:InvokeFunction"**?

### Current Behavior

We tried to use escape hatch to down to the L2/L1 construct to override the role permission, however, the L1 construct of Role is `CfnResource` not `CfnRole`, which is also a bit different from other construct like `laws-lambda` as I can get `CfnRole` and can be used by `add_property_override`
example of `aws_lambda` construct:
```
        fn = lambda_.Function(self, "MyLambda",
            code=lambda_.InlineCode("exports.handler = async () => { console.log('hello world'); };"),
            handler="index.handler",
            runtime=lambda_.Runtime.PYTHON_3_7
        )
        
        children = fn.node.find_all()
        for child in children:
            print(child, "->", child.node.id, "->", child.node.path)
```
printout with `CfnRole` as the construct
```
<aws_cdk.aws_lambda.Function object at 0x7f1df63e8d90> -> MyLambda -> cdk-1-xx/MyLambda
<aws_cdk.aws_iam.Role object at 0x7f1de43f80d0> -> ServiceRole -> cdk-1-xx/MyLambda/ServiceRole
<aws_cdk.aws_iam.CfnRole object at 0x7f1de43f8310> -> Resource -> cdk-1-xx/MyLambda/ServiceRole/Resource
<aws_cdk.aws_lambda.CfnFunction object at 0x7f1df62a3410> -> Resource -> cdk-1-xx/MyLambda/Resource
```
However, for `@aws-cdk/integ-tests-alpha module`, I can only get `CfnResource` construct (see attachment) and cannot use `add_property_override` to change the policy

### Reproduction Steps

Question1: may I know how does the code intergated in the backend to choose which IAM Role to make the invokefunction action inside statemachine ? and how can we customised this IAM role please?

Question2: inside AWS console, the below action is an invalid action, and console will throw validation error, may I know if this is a bug in CDK that auto-generate **"lambda:Invoke"** in action rather than **"lambda:InvokeFunction"**?

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.104.0

### Framework Version

_No response_

### Node.js Version

n/a

### OS

mac

### Language

Python

### Language Version

_No response_

### Other information

_No response_

sakurai-ryo:
### Description
The following issue describes a bug where the IAM Policy is not correctly set to the calling Lambda when using `invokeFunction` and `waitForAssertions`.

Normally, when the `waitForAssertions` method is invoked, the necessary Policy is granted to the `waiterProvider` using the `adPolicyStatementFromSdkCall` method.
https://github.com/aws/aws-cdk/blob/52a5579aa52c88bb289a7a9677c35385763c8fff/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/sdk.ts#L136

In the case of a Lambda function call, the API name and the Action name of the Policy are different (invoke => invokeFunction), so the `addPolicyStatementFromSdkCall` method cannot grant the correct Policy.
The `LambdaInvokeFunction` is doing the correct Policy assignment to deal with this in the constructor.
https://github.com/aws/aws-cdk/blob/52a5579aa52c88bb289a7a9677c35385763c8fff/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/sdk.ts#L247

However, this is not done for the `waiterProvider`, resulting in an access denied error.
This PR has been modified so that the correct Policy is granted to `waiterProvider`.

fixes #27865

----


packages/@aws-cdk/integ-tests-alpha/lib/assertions/sdk.ts
```diff
@@ -250,6 +250,19 @@ export class LambdaInvokeFunction extends AwsApiCall {
       arnFormat: ArnFormat.COLON_RESOURCE_NAME,
       resourceName: props.functionName,
     })]);
+
+    // If using `waitForAssertions`, do the same for `waiterProvider` as above.
+    Aspects.of(this).add({
```
sakurai-ryo, line 255: Although we have created a Resource-based Policy for Lambda in this part, I have not created it here because it is possible to invoke Lambda from Lambda if the appropriate Policy is granted to the invoking Lambda's execution role.
https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/sdk.ts#L236

If you have any concerns, I would really appreciate it if you could let me know.

kaizencc: I'm fine with this but -- can you document why you've used an Aspect so that others who stumble on your code will understand the decision please

packages/@aws-cdk/integ-tests-alpha/test/assertions/providers/integ.invoke-function-assertions.ts
kaizencc, line 1: Lets document here that what we are trying to verify is that `Lambda.InvokeFunction` exists in the integ assertions stack, not in the actual stack that is getting "tested"

kaizencc (changes requested):
Thanks @sakurai-ryo! Some really minor comments but this looks good!

sakurai-ryo:
@kaizencc 
Thanks again for your review!
As for the comments on the integ-tests part, I'm sorry if I misunderstood what you were saying, but I've added docs to `sdk.ts` and the integ-test file.

kaizencc (approved):
Perfect, thanks @sakurai-ryo

PR #28431, opened 2023-12-20
fix(ecs): empty placement constraints not allowed

paulhcsun:
Removes `omitEmptyArray`from ec2 constructor to support [removing all existing placement constraints](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_UpdateService.html#ECS-UpdateService-request-placementConstraints:~:text=To%20remove%20all%20existing%20placement%20constraints%2C%20specify%20an%20empty%20array).

This PR does not address supporting empty placement strategies because of the following reason: https://github.com/aws/aws-cdk/pull/27572#issuecomment-1766287866

More investigation is required for how to address empty placement strategies. 

Addresses the empty placement constraints half of #27555.

----


paulhcsun:
Removing the `omitEmptyArray: true` is resulting in all newly initialized `Ec2Service` instances to contain an empty array instead of `undefined` for the `PlacementConstraint` unless specified otherwise. 

`PlacementConstraint` is initialized to an empty array if not value is provided and would be removed by the lazy calculation but now there is no way to distinguish between when a user purposely passes in an empty array to clear the PlacementConstraints and when it's initialized. 

Some additional refactoring will be required and perhaps the `PlacementConstraint` will have to be initialized as `undefined` instead of an empty array.

PR #28432, opened 2023-12-20
feat(ecr): tag pattern list for lifecycle policy

go-to-k:
This PR supports `tagPatternList` for the lifecycle policy.

According to the doc, the lifecycle policy has following evaluation rules:

> A lifecycle policy rule may specify either tagPatternList or tagPrefixList, but not both. 

> The tagPatternList or tagPrefixList parameters may only used if the tagStatus is tagged.

> There is a maximum limit of four wildcards (\*) per string. For example, ["\*test\*1\*2\*3", "test\*1\*2\*3\*"] is valid but ["test\*1\*2\*3\*4\*5\*6"] is invalid.

https://docs.aws.amazon.com/AmazonECR/latest/userguide/LifecyclePolicies.html#lp_tag_pattern_list

----


packages/aws-cdk-lib/aws-ecr/lib/repository.ts
```diff
@@ -769,14 +769,28 @@ export class Repository extends RepositoryBase {
...
+      throw new Error('tagPrefixList and tagPatternList can only be specified when tagStatus is set to Tagged');
+    }
+    if (rule.tagPrefixList !== undefined && rule.tagPatternList !== undefined) {
+      throw new Error('Both tagPrefixList and tagPatternList cannot be specified together in a rule');
+    }
+    if (rule.tagPatternList !== undefined) {
+      rule.tagPatternList.forEach((pattern) => {
+        const splittedPatternLength = pattern.split('*').length;
+        if (splittedPatternLength > 5) {
+          throw new Error(`A tag pattern cannot contain more than four wildcard characters (*), pattern: ${pattern}, counts: ${splittedPatternLength - 1}`);
```
kaizencc, line 791:
```suggestion
        const splitPatternLength = pattern.split('*').length;
        if (splitPatternLength > 5) {
          throw new Error(`A tag pattern cannot contain more than four wildcard characters (*), pattern: ${pattern}, counts: ${splitPatternLength - 1}`);
```

kaizencc (approved):
Thanks @go-to-k!

PR #28437, opened 2023-12-20
fix(ecs): EC2 metadata access is blocked when using EC2 capacity provider for autoscaling

### LINKED ISSUE #28270: aws-ecs: Cluster.addAsgCapacityProvider blocks EC2 metadata in instances created by the ASG
### Describe the bug

Calling `Cluster.addAsgCapacityProvider` modifies the AutoScalingGroup to add a userdata script to the launch template which blocks access to 169.254.169.254:
```
"MyEcsClusterAsgLaunchTemplate854B9AAD": {
   "Type": "AWS::EC2::LaunchTemplate",
   "Properties": {
    "LaunchTemplateData": {

     ...

     "UserData": {
      "Fn::Base64": {
       "Fn::Join": [
        "",
        [
         "#!/bin/bash\necho ECS_CLUSTER=",
         {
          "Ref": "MyEcsClusterACF79753"
         },
         " >> /etc/ecs/ecs.config\nsudo iptables --insert FORWARD 1 --in-interface docker+ --destination 169.254.169.254/32 --jump DROP\nsudo service iptables save\necho ECS_AWSVPC_BLOCK_IMDS=true >> /etc/ecs/ecs.config"
        ]
       ]
      }
     }
    },
    
    ...
}
```

In particular: `sudo iptables --insert FORWARD 1 --in-interface docker+ --destination 169.254.169.254/32 --jump DROP`

Without the call to `Cluster.addAsgCapacityProvider` (and possibly other methods do this too), this isn't added.

This behaviour isn't documented anywhere that I can find, and causes the error `No RegionEndpoint or ServiceURL configured` from the SDK when using the default settings and expecting everything to be picked up from the environment (which is how it seems to work when creating the cluster from the console).

### Expected Behavior

Don't block access to EC2 metadata, or at least clearly document it and offer alternatives if someone wants to make use of the TaskRole inside the container

### Current Behavior

The EC2 metadata endpoint is blocked using `sudo iptables --insert FORWARD 1 --in-interface docker+ --destination 169.254.169.254/32 --jump DROP`

### Reproduction Steps

```csharp
var app = new App();
var stack = new Stack(app, "ReproStack", new StackProps() {
    Env = new Amazon.CDK.Environment {
        Account = System.Environment.GetEnvironmentVariable("CDK_DEFAULT_ACCOUNT"),
        Region = System.Environment.GetEnvironmentVariable("CDK_DEFAULT_REGION"),
    },
});

var defaultVpc = Vpc.FromLookup(stack, "defaultVpc", new VpcLookupOptions() {
    IsDefault = true,
});

var asg = new AutoScalingGroup(stack, "AutoScalingGroup", new AutoScalingGroupProps() {
    Vpc = defaultVpc,
    InstanceType = InstanceType.Of(InstanceClass.T3A, InstanceSize.SMALL),
    MachineImage = EcsOptimizedImage.AmazonLinux2(),
});

var cluster = new Cluster(stack, "Cluster", new ClusterProps() {
    Vpc = defaultVpc,
});

var provider = new AsgCapacityProvider(stack, "AsgCapacityProvider", new AsgCapacityProviderProps() {
    AutoScalingGroup = asg,
});

// With this line, the UserData is included in the ASG's launch template; without it, it is not
cluster.AddAsgCapacityProvider(provider);

app.Synth();
```

### Possible Solution

Don't block access to EC2 metadata, or at least clearly document it and offer alternatives if someone wants to make use of the TaskRole inside the container

_No response_

### CDK CLI Version

2.114.0 (build 12879fa)

### Framework Version

_No response_

### Node.js Version

v18.16.1

### OS

MacOS 13.6.2

### Language

.NET

### Language Version

.NET 8

### Other information

_No response_

juinquok:
### Why is this needed?
When adding a auto scaling group as a capacity provider using `Cluster.addAsgCapacityProvider` and when the task definition being run uses the AWS_VPC network mode, it results in the metadata service at `169.254.169.254` being blocked . This is a security best practice as detailed [here](https://docs.aws.amazon.com/AmazonECS/latest/bestpracticesguide/security-iam-roles.html). This practice is implemented [here](https://github.com/aws/aws-cdk/blame/2d9de189e583186f2b77386ae4fcfff42c864568/packages/aws-cdk-lib/aws-ecs/lib/cluster.ts#L502-L504). However by doing this, some applications such as those raised in #28270 as well as the aws-otel package will not be able to source for the AWS region and thus, cause the application to crash and exit. 

### What does it implement?
This PR add an override to the addContainer method when using the Ec2TaskDefinition to add in the AWS_REGION environment variable to the container if the network mode is set as AWS_VPC. The region is sourced by referencing to the stack which includes this construct at synth time.This environment variable is only required in the EC2 Capacity Provider mode and not in Fargate as this issue of not being able to source for the region on startup is only present when using the EC2 Capacity Provider with the AWS_VPC networking mode. The initial issue addresses this during the `addAsgCapacityProvider` action which targets the cluster. However, we cannot mutate the task definition at that point in time thus, this change addresses it when the task definition is actually added to a service that meets all the requirements whereby the failure to source for region will occur.

Updated the relevant integration tests to reflect the new environment variable being created alongside user-defined environment variables.

Closes #28270

----


packages/aws-cdk-lib/aws-ecs/lib/ec2/ec2-task-definition.ts
```diff
@@ -146,4 +147,23 @@ export class Ec2TaskDefinition extends TaskDefinition implements IEc2TaskDefinit
...
+        taskDefinition: this,
+        ...props,
+        environment: {
+          ...props.environment,
+          AWS_REGION: Stack.of(this).region,
+        },
+      });
+    }
+    // If network mode is not AWSVPC, then just add the container as normal
+    return new ContainerDefinition(this, id, { taskDefinition: this, ...props });
```
lpizzinidev, line 167:
```suggestion
    if (this.networkMode === NetworkMode.AWS_VPC) {
      return super.addContainer(id, {
        ...props,
        environment: {
          ...props.environment,
          AWS_REGION: Stack.of(this).region,
        },
      });
    }
    // If network mode is not AWSVPC, then just add the container as normal
    return super.addContainer(id, props);
```
Let's reuse the parent's class method

packages/aws-cdk-lib/aws-ecs/test/ec2/ec2-task-definition.test.ts
```diff
@@ -1110,6 +1110,67 @@ describe('ec2 task definition', () => {
...
+            Name: 'AWS_REGION',
+            Value: {
+              Ref: 'AWS::Region',
+            },
+          }],
+        }],
+      });
+
+      // GIVEN HOST network mode
+      const anotherStack = new cdk.Stack();
```
lpizzinidev, line 1148: This case should be separated into another `test`.

packages/aws-cdk-lib/aws-ecs/test/ec2/ec2-task-definition.test.ts
```diff
@@ -1110,6 +1110,67 @@ describe('ec2 task definition', () => {
         }],
       });
     });
+
+    test('correctly sets env variables when using EC2 capacity provider with AWSVPC mode', () => {
```
lpizzinidev, line 1114: Can you also add another test with `awsvpc` network mode and no added environment variables? (will set only `AWS_REGION`)

lpizzinidev (changes requested):
Thanks 👍
I left some notes for improvements.
Also, the title should be changed to [describe the bug (not the solution)](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request).
Something like `fix(ecs): EC2 metadata access is blocked on auto scaling`.

juinquok:
> Thanks 👍 I left some notes for improvements. Also, the title should be changed to [describe the bug (not the solution)](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request). Something like `fix(ecs): EC2 metadata access is blocked on auto scaling`.

Thank You! I have updated the method and also the tests! :)

lpizzinidev (approved):
Thanks 👍

paulhcsun (approved):
Great work @juinquok! 

And thanks for the review @lpizzinidev!

PR #28443, opened 2023-12-20
fix(lambda-python-alpha): pass architecture property to function

mapk-amazon:
By default Lambda will take x86 architecture, which might get into conflict with layers/code. With this change the architecture information is passed along and the correct architecture for the docker image is chosen. 

----


kaizencc (changes requested):
Thanks @mapk-amazon, we need a unit test to go along with this change to ensure that it works. Are you up for that?

kaizencc (commented):
Superseded by #28449

PR #28475, opened 2023-12-23
fix(aws-lambda): Fix the log format type on logging configurations

### LINKED ISSUE #28114: aws-lambda: Log format option is not typed

AnuragMohapatra:
Updated the logFormat property type from `string` to `LogFormat`

Closes #28114 .

----


AnuragMohapatra:
There is already a PR #28127

Issue #894, opened 2018-10-11
Enabling IPv6 on Resources and VPCs


yandy-r:
Hi,

I'm just starting out to play with the SDK, please forgive if I a missing something. First off, I'd like to probably point out how great this product is and I see a bright future for it. It's in the early stages and I understand things will be missing. This is not a complaint, just something I noticed while attempting to create an IPv6 enabled app.

 It seems like the ability to map an `amazonProvidedIpv6CidrBlock` is missing within the `@aws-cdk/aws-ec2 VpcNetwork` construct as well as within the `SubnetConfiguration` interface. 

Something along the lines of the below would be helpful.

```typescript
import * as ec2 from "@aws-cdk/aws-ec2";

new ec2.VpcNetwork(this, name, {
// other attributes
  amazonProvidedIpv6CidrBlock: true,
// other attributes
}
```

Other values left out on purpose just to keep this concise.

From the subnetConfiguration[] it would be helpful to have.

```typescript
import * as ec2 from "@aws-cdk/aws-ec2";

new ec2.VpcNetwork(this, name, {
// other attributes
  subnetConfiguration: [
    {
      name: "pub",
      cidrMask: 22,
      ipv6Cidr: true,
      mapIPv6OnLaunch: true,
      subnetType: ec2.SubnetType.Public
    }
  ]
// other attributes
}
```

The `ipv6Cidr` could be a `boolean` since all subnet cidrs must be `/64` and then the obvious addition to map IPv6 address to instances within those subnet .

To get the `amazonProvidedIpv6CidrBlock` within the created VpcNetwork I had to do something like this, and in testing, it seems to work. Though going through the documentation, it's advised to use native constructs, which I'd agree completely.

```typescript
this.vpc = new ec2.VpcNetwork(this, `${name}Vpc`, args);
new ec2.cloudformation.VPCCidrBlockResource(
  this,
  `CidrRes`,
  {
    vpcId: this.vpc.vpcId,
    amazonProvidedIpv6CidrBlock: true
  }
);
```

rix0rrr:
We'll definitely take this under consideration. Thanks for the suggestion!

FlorinAsavoaie:
[...]

PR #28480, opened 2023-12-23
feat(ec2): add dual stack vpc support

scanlonp:
Adds parameter to configure a dual stack vpc, `ipProtocol: ec2.IpProtocol.DUAL_STACK`.

By default a dual stack vpc will:

- Create an Amazon provided IPv6 CIDR block (/56) and associate it to the VPC.
- Assign a portion of the block to each of the subnets (/64)
- Enable autoassigning an IPv6 address for each subnet
- **Disable autoassigning public IPv4 addresses** for each subnet
- Create an Egress Only Internet Gateway for private subnets
- Configure IPv6 routes for IGWs and EIGWs

Addresses #894.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-elasticsearch/lib/elasticsearch-access-policy.ts
+++ b/packages/aws-cdk-lib/aws-elasticsearch/lib/elasticsearch-access-policy.ts
@@ -33,6 +33,7 @@ export class ElasticsearchAccessPolicy extends cr.AwsCustomResource {
   constructor(scope: Construct, id: string, props: ElasticsearchAccessPolicyProps) {
     super(scope, id, {
       resourceType: 'Custom::ElasticsearchAccessPolicy',
+      installLatestAwsSdk: false,
       onUpdate: {
         action: 'updateElasticsearchDomainConfig',
         service: 'ES',
--- a/packages/aws-cdk-lib/aws-opensearchservice/lib/opensearch-access-policy.ts
+++ b/packages/aws-cdk-lib/aws-opensearchservice/lib/opensearch-access-policy.ts
@@ -33,6 +33,7 @@ export class OpenSearchAccessPolicy extends cr.AwsCustomResource {
   constructor(scope: Construct, id: string, props: OpenSearchAccessPolicyProps) {
     super(scope, id, {
       resourceType: 'Custom::OpenSearchAccessPolicy',
+      installLatestAwsSdk: false,
       onUpdate: {
         action: 'updateDomainConfig',
         service: 'OpenSearch',
--- a/packages/aws-cdk/lib/api/cxapp/exec.ts
+++ b/packages/aws-cdk/lib/api/cxapp/exec.ts
@@ -118,6 +118,7 @@ export async function execProgram(aws: SdkProvider, config: Configuration): Prom
         if (code === 0) {
           return ok();
         } else {
+          debug('failed command:', commandAndArgs);
           return fail(new Error(`Subprocess exited with error ${code}`));
         }
       });


packages/aws-cdk-lib/aws-ec2/lib/ip-addresses.ts
```diff
@@ -144,9 +152,18 @@ export interface SubnetIpamOptions {
...
+   * Note this is specific to the IPv4 CIDR.
    */
   readonly cidr: string;
+
+  /**
+   * IPv6 Cidr Allocations for a Subnet.
+   *
+   * Note this is specific to the IPv6 CIDR.
+   */
+  ipv6Cidr?: string;
```
sgryphon, line 166: Should this be readonly, like the IPv4 version? (P.S. Nice to see the documentation updates to the IPv4 only fields).

Also, should this be a collection (array/list), because multiple IPv6 subnets (like multiple IPv6 addresses) will be much more common, i.e. make the default with support for multiple IPv6. (Yes, you can have multiple IPv4, but that would be much less common).

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1486,6 +1594,23 @@ export class Vpc extends VpcBase {
...
+        this.ipv6CidrBlock = new CfnVPCCidrBlock(this, 'ipv6cidr', {
+          vpcId: this.vpcId,
+          amazonProvidedIpv6CidrBlock: this.ipv6IpAddresses.amazonProvided,
+        });
+
+        this.ipv6SelectedCidr = Fn.select(0, this.resource.attrIpv6CidrBlocks);
+      }
+    }
+
+    // Add tag here too?
```
sgryphon, line 1612: Tags are added just above here, around line 1565: Tags.of(this).add(NAME_TAG...

I strongly suggest adding a tag, e.g. 'aws-cdk:vpc-protocol' with the selected type, similar to how subnets have a subnet type tag.

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1508,6 +1633,11 @@ export class Vpc extends VpcBase {
 
       (this.publicSubnets as PublicSubnet[]).forEach(publicSubnet => {
         publicSubnet.addDefaultInternetRoute(igw.ref, att);
+
+        // configure IPv6 route if VPC is dual stack
+        if (this.useIpv6) {
+          publicSubnet.addIpv6DefaultInternetRoute(igw.ref);
```
sgryphon, line 1639: Suggest also adding a NAT64 route, to the NAT gateway (used in IPv4). This will allow IPv6 only machines to use DNS64 (e.g. Google) and be able to access IPv4-only resources. Just adding the route has no drawback.

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1517,6 +1647,17 @@ export class Vpc extends VpcBase {
...
     }
 
+    // Create an Egress Only Internet Gateway and attach it if necessary
+    if (this.useIpv6 && this.privateSubnets) {
+      const eigw = new CfnEgressOnlyInternetGateway(this, 'EIGW6', {
+        vpcId: this.vpcId,
+      });
+
+      (this.privateSubnets as PrivateSubnet[]).forEach(privateSubnet => {
+        privateSubnet.addIpv6DefaultEgressOnlyInternetRoute(eigw.ref);
```
sgryphon, line 1657: Also add a NAT64 route, to the NAT gateway (used in IPv4).

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1927,6 +2126,34 @@ export class Subnet extends Resource implements ISubnet {
...
+   * @param gatewayId the logical ID (ref) of the gateway attached to your VPC
+   */
+  public addIpv6DefaultInternetRoute(gatewayId: string) {
+    this.addRoute('DefaultRoute6', {
+      routerType: RouterType.GATEWAY,
+      routerId: gatewayId,
+      destinationIpv6CidrBlock: '::/0',
+      enablesInternetConnectivity: true,
+    });
+  }
```
sgryphon, line 2141: Also add function for addIpv6Nat64Route, with 64:ff9b::/96 needs to route to a NAT gateway (same as for IPv4, i.e. either one per zone, a single gateway, or round robin). Adding NAT64 allows IPv6-only machines to access IPv4-only resources via NAT (works the same as an IPv4 NAT gateway)

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1879,7 +2073,12 @@ export class Subnet extends Resource implements ISubnet {
       cidrBlock: props.cidrBlock,
       availabilityZone: props.availabilityZone,
       mapPublicIpOnLaunch: props.mapPublicIpOnLaunch,
+      ipv6CidrBlock: props.ipv6CidrBlock,
+      assignIpv6AddressOnCreation: props.assignIpv6AddressOnCreation,
```
sgryphon, line 2077: Set enable_dns64 = true (if protocol is dual stack), will simplify creation of IPv6-only machines on the VPC.  A machine can always use an alternative, e.g. Google DNS64, provided there is a NAT64 route (see other comments), but setting here will make it automatic.

NOTE: You need to have NAT64 routes set up before enabling this.

Also (can be in the future) suggest adding an override when creating custom subnets as there could be some cases where you don't want NAT64 turned on.

sgryphon, line 2077: Also set `private_dns_name_options_on_launch`. For IPv6-only subnets, this is automatically enabled, however for dual-stack the default is not, which means they will get an IPv4 internal DNS name but not IPv6. Note this only applies to internal DNS names.

This is not yet fully mapped to CloudFormation, so you just need to pass in an array of any with the known properties.

Set: EnableResourceNameDnsAAAARecord = true, EnableResourceNameDnsARecord = true, HostnameType = 'resource-name'.

This will mean that machines created on the dual stack subnet will have an machine name based on the instance (not the IPv4 address) and be accessible via internal DNS by that name.

By setting for both v4 & v6 that single internal DNS reference (as instance) will work for both IPv4 and IPv6, i.e. a single internal host name.

sgryphon:
Hi, thanks for the tag,

I added comments for the following:
* Adding a tag, e.g. 'aws-cdk:vpc-protocol'
* Add NAT64 route for IPv6 (for both public and egress only routing)
* Enable DNS64 in dual stack subnets
* Enable IPv6 private DNS names in dual stack subnets

See individual comments for details. Happy to discuss if you have any questions.

Those are probably important for default IPv6 behaviour.

I understand some of my other suggestions, like customising subnets (e.g. to create IPv6 only), or multiple IPv6 block assignement, are not part of this (and may be looked at in the future). I've already provided suggestions on these, and will be happy to look at future PRs.

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.vpc-dual-stack-ec2.assets/web-server.py
```diff
@@ -0,0 +1,42 @@
...
+class MyRequestHandler(SimpleHTTPRequestHandler):
+    def do_GET(self):
+        if self.path == '/':
+            self.send_response(200)
+            self.send_header('Content-type', 'application/json')
+            self.end_headers()
+
+            try:
+                # IPv4 request
+                response = urllib.request.urlopen('http://www.google.com')
```
mikewrighton, line 14: Change to `http://ipv4.google.com`

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1757,6 +1940,18 @@ function subnetTypeTagValue(type: SubnetType) {
   }
 }
 
+/**
+ * Wraps the CFN CIDR block which the IPv6 enabled subnets depend on.
+ */
+export interface DependantIpv6CidrBlock {
```
rix0rrr, line 1946: Replace with `IDependable`

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.vpc-dual-stack-ec2.assets/web-server.py
```diff
@@ -0,0 +1,42 @@
...
+                status_code = response.getcode()  
+                if (status_code != 200):
+                  raise Exception(f"Received a non-successful status code: {status_code}")
+                # self.wfile.write(bytes(f"Status code from www.google.com: {status_code}\n\n", 'utf-8'))
```
mikewrighton, line 18: Just need to remove these commented out bits.

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
rix0rrr: Have a look at `internetConnectivityEstablished` as well

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1647,23 +1829,24 @@ export class Vpc extends VpcBase {
...
+      const subnetProps = {
         availabilityZone,
         vpcId: this.vpcId,
         cidrBlock: allocated.cidr,
-        mapPublicIpOnLaunch: mapPublicIpOnLaunch,
-      };
+        mapPublicIpOnLaunch: this.calculateMapPublicIpOnLaunch(subnetConfig),
+        ipv6CidrBlock: allocated.ipv6Cidr,
+        assignIpv6AddressOnCreation: this.useIpv6 ? subnetConfig.ipv6AssignAddressOnCreation ?? true : undefined,
+        dependantIpv6CidrBlock: this.useIpv6 ? { cidrBlock: this.ipv6CidrBlock } : undefined,
```
rix0rrr, line 1848: Is this necessary if, as a caller, you can do `addDependency` instead?

packages/aws-cdk-lib/aws-ec2/lib/ip-addresses.ts
```diff
@@ -144,9 +167,20 @@ export interface SubnetIpamOptions {
  */
 export interface AllocatedSubnet {
   /**
-   * Cidr Allocations for a Subnet
+   * IPv4 Cidr Allocations for a Subnet.
```
mikewrighton, line 170: nit: `IPv4 CIDR allocations for a subnet`?

packages/aws-cdk-lib/aws-ec2/README.md
```diff
@@ -270,6 +270,22 @@ new ec2.Vpc(this, 'TheVPC', {
...
+
+To allocate both IPv4 and IPv6 addresses in your VPC, you can configure your VPC to have a dual stack protocol.
+
+```ts
+new ec2.Vpc(this, 'DualStackVpc', {
+  vpcProtocol: ec2.VpcProtocol.DUAL_STACK,
+})
+```
+
+By default, a dual stack VPC will create an Amazon provided IPv6 /56 CIDR block associted to the VPC. It will then assign /64 portions of the block to each subnet. For each subnet, auto-assigning an IPv6 address will be enabled, and auto-asigning a public IPv4 address will be disabled. An egress only internet gateway will be created for `PRIVATE_WITH_EGRESS` subnets, and IPv6 routes will be added for IGWs and EIGWs.
```
rix0rrr, line 283:
```suggestion
By default, a dual stack VPC will create an Amazon provided IPv6 /56 CIDR block associated to the VPC. It will then assign /64 portions of the block to each subnet. For each subnet, auto-assigning an IPv6 address will be enabled, and auto-asigning a public IPv4 address will be disabled. An egress only internet gateway will be created for `PRIVATE_WITH_EGRESS` subnets, and IPv6 routes will be added for IGWs and EIGWs.
```

rix0rrr: Maybe show an example of a `subnetconfg` block with all the settings, and some comments explaining the defaults and how they hang together?

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -858,12 +879,22 @@ const NAME_TAG: string = 'Name';
...
  */
 export interface VpcProps {
+  /**
+   * The protocol of the vpc.
+   *
+   * Options are IPv4 only or dual stack.
+   *
+   * @default VpcProtocol.IPV4_ONLY
+   */
+  readonly vpcProtocol?: VpcProtocol;
```
rix0rrr, line 889: ipProtocol? 😜

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -159,6 +159,27 @@ export interface IVpc extends IResource {
   addFlowLog(id: string, options?: FlowLogOptions): FlowLog
 }
 
+/**
+ * The types of IP addresses provisioned in the VPC.
+ */
+export enum VpcProtocol {
```
mikewrighton, line 165: Might vote for `VpcMode`?

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1630,7 +1771,48 @@ export class Vpc extends VpcBase {
...
 
-    this.createSubnetResources(requestedSubnets, allocatedSubnets);
+    if (this.useIpv6) {
+      if (this.ipv6SelectedCidr === undefined) {
+        throw new Error('No IPv6 CIDR block associated with this VPC could be found');
+      }
+      if (this.ipv6IpAddresses === undefined) {
+        throw new Error('No IPv6 IpAddresses were found');
+      }
+      // create the IPv6 CIDR block
```
mikewrighton, line 1781: nit: `CIDR block` -> `CIDR blocks`?

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1630,7 +1771,48 @@ export class Vpc extends VpcBase {
...
+      const subnetIpv6Cidrs = Fn.cidr(this.ipv6SelectedCidr, allocatedSubnets.length, (128 - 64).toString());
```
rix0rrr, line 1782: This decision should probably be in `Ipv6Addresses`.

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1630,7 +1771,48 @@ export class Vpc extends VpcBase {
...
+      const allocatedSubnetsIpv6 = this.ipv6IpAddresses.allocateSubnetsIpv6Cidr({
+        allocatedSubnets: allocatedSubnets,
+        ipv6Cidrs: subnetIpv6Cidrs,
+      });
+
+      // call the create function with the updated allocated subnet list
+      this.createSubnetResources(requestedSubnets, allocatedSubnetsIpv6.allocatedSubnets);
+    } else {
+      // keep default behavior without IPv6 CIDRs if not using IPv6
+      this.createSubnetResources(requestedSubnets, allocatedSubnets);
```
rix0rrr, line 1794: Why is this almost the same but not? Can we collapse this a bit nicer?

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1429,6 +1519,24 @@ export class Vpc extends VpcBase {
       throw new Error('supply at most one of ipAddresses or cidr');
     }
 
+    // this property can be set to false if an IPv6_ONLY VPC is implemented in the future
+    this.useIpv4 = props.vpcProtocol === VpcProtocol.IPV4_ONLY || props.vpcProtocol === VpcProtocol.DUAL_STACK;
+    // use property to avoid lint errors
+    if (this.useIpv4) {
+      ; // pass
+    }
```
rix0rrr, line 1527: Use it where it makes sense though, right? Where it allocates IPv4 resources? WHere it adds the IPv4 routes?

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -1486,6 +1594,23 @@ export class Vpc extends VpcBase {
...
+        // create the IPv6 CIDR block and associate it with the VPC
+        // create a dependency of the subnets on the CIDR block
+        this.ipv6CidrBlock = new CfnVPCCidrBlock(this, 'ipv6cidr', {
+          vpcId: this.vpcId,
+          amazonProvidedIpv6CidrBlock: this.ipv6IpAddresses.amazonProvided,
+        });
+
+        this.ipv6SelectedCidr = Fn.select(0, this.resource.attrIpv6CidrBlocks);
+      }
+    }
```
rix0rrr, line 1610: This behavior should be part of the `IIpv6Addresses` protocol. Don't make this boolean property, put the behavior there.

packages/aws-cdk-lib/aws-ec2/README.md
isker, line 283: As a clueless end user 🌞, there's a lot of language both in the description and contents of this PR that says "IPV4 addresses are disabled by default in dual stack VPCs".  But is that really only talking about public IPV4 addresses?  Will I still get private IPV4 addresses in private subnets by default?

I think the language could stand be clarified in this regard.

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.vpc-dual-stack-ec2.assets/web-server.py
sgryphon: :+1: ipv4.google.com would also be useful for testing NAT64

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
sgryphon: It is not really a protocol, as their is V4, V6, and dual stack. I guess you could have flags, but "none" is not really valid. 

There are also different configurations for subnets : right now a VPC can only be IPV4_ONLY or DUAL_STACK, whereas a subnet can be IPV4_ONLY, IPV6_ONLY, or DUAL_STACK.   You could still combine into a single enum, and maybe "ipProtocol" is okay (another suggestion was VpcMode) ... but then only 2 of the values would be valid for a VPC.

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
sgryphon: For standard networks, the interface ID is required to be 64 bits long (RFC 4291), so the subnet will be 64 bits ... but yes, maybe move into the address/subnet provider, to at least be consistent with IPv4

packages/aws-cdk-lib/aws-ec2/README.md
sgryphon: Yes "auto-asigning a public IPv4 address will be disabled". Dual stack, by definition, means both IPv4 & IPv6. 

i.e. In the top level description, change:

"Disable autoassigning IPv4 addresses for each subnet" to "Disable autoassigning public IPv4 addresses for each subnet"

Maybe even add an explicit "Private IPv4 addresses will still be assigned".

packages/aws-cdk-lib/aws-ec2/lib/vpc.ts
```diff
@@ -159,6 +159,27 @@ export interface IVpc extends IResource {
...
+   * The vpc will be configured with only IPv4 addresses.
+   *
+   * This is the default protocol if unset.
+   */
+  IPV4_ONLY = 'Ipv4_Only',
+
+  /**
+   * The vpc will have both IPv4 and IPv6 addresses.
+   *
+   * Unless specified, IPv4 addresses will not be auto assigned,
```
isker, line 176:
```suggestion
   * Unless specified, public IPv4 addresses will not be auto assigned,
```

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.vpc-dual-stack-ec2.assets/web-server.py
mrpackethead: I'd be concerned to see a test that is reliant on any particular service out there, particularly a third party one.

packages/aws-cdk-lib/aws-ec2/lib/ip-addresses.ts
```diff
@@ -1,16 +1,22 @@
...
+import { Construct } from 'constructs';
 
 /**
  * An abstract Provider of IpAddresses
+ *
+ * Note this is specific to the IPv4 CIDR.
  */
 export class IpAddresses {
   /**
    * Used to provide local Ip Address Management services for your VPC
```
mikewrighton, line 15: nit: `IP`- same applies in a few places below.

scanlonp:
@Mergifyio update

mikewrighton (approved):

PR #27978, opened 2023-11-14
feat(elasticloadbalancingv2): add security group support for nlb

msambol:
Closes #26735.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #28494, opened 2023-12-26
feat(elbv2): Implement IConnectable to NLB

### LINKED ISSUE #26735: (aws-elasticloadbalancingv2): (network-load-balancer-supports-security-groups)
### Describe the feature

https://aws.amazon.com/tw/about-aws/whats-new/2023/08/network-load-balancer-supports-security-groups/
network load balancer supports security groups.

### Use Case

Supports create NLB with SG via aws cdk L2.

### Proposed Solution

Support securityGroup into NetworkLoadBalancerProps.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.91.0

macOS Ventura 13.4.1

WinterYukky:
# Summary
Implement an `IConnectable` interface to a NetworkLoadBalancer. 

# Why need this change?
AWS CDK has great features for abstraction. `IConnectable` interface is one of this. `IConnectable` simplifies the management of security groups. AWS CDK add support security group to NLB at https://github.com/aws/aws-cdk/pull/27978. However, Currently NLB not implement `IConnectable`, so customers can't use useful interface in AWS CDK.

# Example use case

```ts
declare const lb: elbv2.NetworkLoadBalancer;
declare const asg: autoscaling.AutoScalingGroup;

asg.connections.allowFrom(lb, ec2.Port.tcp(8088)); // currently can not
```

Closes #26735

----


packages/aws-cdk-lib/aws-elasticloadbalancingv2/test/nlb/load-balancer.test.ts
```diff
@@ -800,34 +827,32 @@ describe('tests', () => {
...
+          account: '123456789012',
+          region: 'us-west-2',
+        },
       });
 
-      // THEN
-      expect(nlb.securityGroups).toEqual([`${sg.securityGroupId}`]);
-    });
-
-    test('can look up with no security groups', () => {
```
WinterYukky, line 818: This test was removed because it was not a test as the test title suggests, and it was a duplicate test where NLB was initialized without a security group.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-load-balancer.ts
```diff
@@ -196,16 +201,19 @@ export class NetworkLoadBalancer extends BaseLoadBalancer implements INetworkLoa
...
   public readonly ipAddressType?: IpAddressType;
+  public readonly connections: ec2.Connections;
 
   constructor(scope: Construct, id: string, props: NetworkLoadBalancerProps) {
     super(scope, id, props, {
       type: 'network',
-      securityGroups: props.securityGroups?.map(sg => sg.securityGroupId),
+      securityGroups: Lazy.list({
+        produce: () => this.connections.securityGroups.length >= 1 ? this.connections.securityGroups.map(sg => sg.securityGroupId) : undefined,
+      }),
```
WinterYukky, line 211: If `securityGroups` becomes an empty array from undefined, an update will be applied and deployment will not be possible, so in the case of an empty array, it is undefined for backwards compatibility.

packages/@aws-cdk-testing/framework-integ/test/aws-elasticloadbalancingv2/test/integ.nlb.ts
```diff
@@ -11,28 +12,41 @@ const vpc = new ec2.Vpc(stack, 'VPC', {
...
+});
+backend.addListener('Listener', {
+  protocol: elbv2.ApplicationProtocol.HTTP,
+  defaultAction: elbv2.ListenerAction.fixedResponse(200, {
+    contentType: 'application/json',
+    messageBody: JSON.stringify({
+      Message: 'I am ALB!',
+    }),
+  }),
+});
```
WinterYukky, line 26: I changed target to ALB from IP to test reachability client -> nlb -> backend. This can test security group settings via http api call.

lpizzinidev (changes requested):
Thanks 👍
I left some suggestions for adjustments, feel free to comment on those.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-load-balancer.ts
```diff
@@ -196,16 +201,19 @@ export class NetworkLoadBalancer extends BaseLoadBalancer implements INetworkLoa
...
   public readonly securityGroups?: string[];
   public readonly ipAddressType?: IpAddressType;
+  public readonly connections: ec2.Connections;
 
   constructor(scope: Construct, id: string, props: NetworkLoadBalancerProps) {
     super(scope, id, props, {
       type: 'network',
-      securityGroups: props.securityGroups?.map(sg => sg.securityGroupId),
+      securityGroups: Lazy.list({
+        produce: () => this.connections.securityGroups.length >= 1 ? this.connections.securityGroups.map(sg => sg.securityGroupId) : undefined,
```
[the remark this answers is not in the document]
WinterYukky: Nice catch!! They are not same. I tested the following changes and I know it's can't change each other undefined and `[]`.

| before | after | error message |
| - | - | - |
| undefined | [] | A security group must be specified |
| undefined | [sg] | You cannot set security groups on a Network Load Balancer which was created without any security groups. |
| [] | undefined | 1 validation error detected: Value null at 'securityGroups' failed to satisfy constraint: Member must not be null |
| [] | [sg] | You cannot set security groups on a Network Load Balancer which was created without any security groups. |
| [sg] | undefined | 1 validation error detected: Value null at 'securityGroups' failed to satisfy constraint: Member must not be null |
| [sg] | [] | A security group must be specified |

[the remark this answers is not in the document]
WinterYukky: If implement feature flag, then I propose to change default security groups value.
If customer does not provide  security groups then AWS CDK creates a security group under the feature flag. If customer would not like to create security group, then customer should provide empty array.

This is similar codes, if under the feature flag.
```ts
new NetworkLoadBalancer(this, "NLB", {
  vpc,
});

declare const sg: ec2.SecurityGroup;
new NetworkLoadBalancer(this, "NLB", {
  vpc,
  securityGroups: [sg],
});
```

When customer is not under the feature flags, then securityGroups are as it is.
This is pseudocode.
```ts
produce: () => this.connections.securityGroups.length > 0 
  ? this.connections.securityGroups.map(sg => sg.securityGroupId) 
  : props.securityGroups?.map(sg => sg.securityGroupId),
```

I'll be implementing this, so please review again.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-load-balancer.ts
```diff
@@ -196,16 +201,19 @@ export class NetworkLoadBalancer extends BaseLoadBalancer implements INetworkLoa
   public readonly metrics: INetworkLoadBalancerMetrics;
   public readonly securityGroups?: string[];
```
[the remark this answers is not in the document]
WinterYukky: I'll mark it as deprecated.

packages/@aws-cdk/cli-lib-alpha/THIRD_PARTY_LICENSES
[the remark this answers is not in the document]
WinterYukky: Reverted.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-load-balancer.ts
```diff
@@ -222,6 +230,13 @@ export class NetworkLoadBalancer extends BaseLoadBalancer implements INetworkLoa
     });
   }
 
+  /**
+   * Add a security group to this load balancer
+   */
+  public addSecurityGroup(securityGroup: ec2.ISecurityGroup) {
+    this.connections.addSecurityGroup(securityGroup);
```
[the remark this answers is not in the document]
WinterYukky: Thanks!

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-load-balancer.ts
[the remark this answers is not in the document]
lpizzinidev: Or, given that we need to keep the `securityGroups` property up-to-date, we could simply use that:
```ts
securityGroups: : Lazy.list({
  produce: () => this.securityGroups?.map(sg => sg.securityGroupId),
}),
```
This won't require the feature flag.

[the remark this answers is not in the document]
WinterYukky: Exactly. However, to keeping the each securityGroups is prone to bugs and difficult for operation. I would like to change securityGroups property to getter that reference `connections`.

```ts
public get securityGroups() {
  /**
   * With the implementation of IConnectable, connections security groups are now used instead of the security group passed in props,
   * however, since connections always return arrays that are not undefined and NLB does not allow mutual changes of undefined and empty array,
   * when the connections security groups is an empty array, the original security group is specified for backwards compatible. (original can be undefined or empty array)
   * https://github.com/aws/aws-cdk/pull/28494
   */
  return this.connections.securityGroups.length > 0
    ? this.connections.securityGroups.map(sg => sg.securityGroupId)
    : this.originalSecurityGroups?.map(sg => sg.securityGroupId);
}
/**
 * The security groups that were passed in Props.
 */
private readonly originalSecurityGroups?: ec2.ISecurityGroup[];

constructor(scope: Construct, id: string, props: NetworkLoadBalancerProps) {
  super(scope, id, props, {
    type: 'network',
    securityGroups: Lazy.list({ produce: () => this.securityGroups }),
    ipAddressType: props.ipAddressType,
  });
```

packages/aws-cdk-lib/aws-elasticloadbalancingv2/test/nlb/load-balancer.test.ts
WinterYukky: Added similar test.
https://github.com/aws/aws-cdk/pull/28494/files#diff-8337948b99d70f9c9a158c9b6c7dc15a6061a239d6d2fcc8f697b05f56ed0868R763

WinterYukky:
@lpizzinidev Thanks your reviewing!!
Your comments are great and I fixed the codes lined your comments. Could you re:review?
(feature flag about create default security group is out of this topic so I'll create another PR)

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-load-balancer.ts
```diff
@@ -429,6 +457,7 @@ export interface INetworkLoadBalancer extends ILoadBalancerV2, ec2.IVpcEndpointS
 
   /**
    * Security groups associated with this load balancer
+   * @deprecated Use `connections.securityGroups`
```
lpizzinidev, line 460:
```suggestion
```
Given the getter, I think we no longer need this deprecation

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-load-balancer.ts
```diff
@@ -194,18 +199,34 @@ export class NetworkLoadBalancer extends BaseLoadBalancer implements INetworkLoa
...
+   * The security groups that were passed in Props.
+   */
+  private readonly originalSecurityGroups?: ec2.ISecurityGroup[];
+  public get securityGroups() {
+    /**
+     * With the implementation of IConnectable, connections security groups are now used instead of the security group passed in props,
+     * however, since connections always return arrays that are not undefined and NLB does not allow mutual changes of undefined and empty array,
+     * when the connections security groups is an empty array, the original security group is specified for backwards compatible. (original can be undefined or empty array)
+     * https://github.com/aws/aws-cdk/pull/28494
+     */
```
lpizzinidev, line 214:
```suggestion
  public readonly connections: ec2.Connections;

  /**
   * The security groups that were passed via the `securityGroups` property.
   */
  private readonly originalSecurityGroups?: ec2.ISecurityGroup[];

  /**
   * After the implementation of `IConnectable` (see https://github.com/aws/aws-cdk/pull/28494), the default
   * value for `securityGroups` is set by the `ec2.Connections` constructor to an empty array.
   * To keep backward compatibility (`securityGroups` is `undefined` if the related property is not specified)
   * a getter has been added.
   */
  public get securityGroups(): string[] | undefined {
```

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-load-balancer.ts
```diff
@@ -194,18 +199,34 @@ export class NetworkLoadBalancer extends BaseLoadBalancer implements INetworkLoa
...
+    return this.connections.securityGroups.length > 0
+      ? this.connections.securityGroups.map(sg => sg.securityGroupId)
+      : this.originalSecurityGroups?.map(sg => sg.securityGroupId);
```
lpizzinidev, line 217:
```suggestion
    return this.connections.securityGroups.length > 0
      ? this.connections.securityGroups.map(sg => sg.securityGroupId)
      : this.isSecurityGroupsPropertyDefined ? [] : undefined;
```
We'll reach the second condition only to return `[]` or `undefined` (correct me if I'm wrong).
I wonder if using a boolean makes for a clearer implementation.
What do you think?

lpizzinidev (changes requested):
Nice 👍
I left some comments for a final cleanup and adjustments.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-load-balancer.ts
WinterYukky: Exactly! this is nice!

lpizzinidev (approved):
Thanks 👍

paulhcsun (approved):
Thanks for the contribution @WinterYukky! This will be a very useful addition for the community. 

As always, thanks for reviewing @lpizzinidev!

PR #1291, opened 2018-12-05
fix(aws-ecs): set permissions for 'awslogs' log driver

rix0rrr:
Make sure that tasks using the 'awslogs' Log Driver have the correct IAM
permissions to actually write logs. Add grant() methods to IAM LogGroups
to make this nicer to write.

Fixes #1279.

----

### Pull Request Checklist

Please check all boxes, including N/A items:

#### Testing

- [x] Unit test and/or integration test added
- [x] __Toolkit change?:__ [integration tests](https://github.com/awslabs/aws-cdk/blob/master/packages/aws-cdk/integ-tests/test.sh) manually executed (paste output to the PR description)
- [x] __Init template change?:__ coordinated update of integration tests (currently maintained in a private repo).

#### Documentation

- [x] __README__: README and/or documentation topic updated
- [x] __jsdocs__: All public APIs documented

### Title and description

- [x] __Change type__: Title is prefixed with change type:
  * `fix(module): <title>` bug fix (_patch_)
  * `feat(module): <title>` feature/capability (_minor_)
  * `chore(module): <title> ` won't appear in changelog
  * `build(module): <title>` won't appear in changelog
- [x] __Title format__: Title uses lower case and doesn't end with a period
- [x] __Breaking change?__: Last paragraph of description is: `BREAKING CHANGE: <describe exactly what changed and how to achieve similar behavior + link to documentation/gist/issue if more details are required>`
- [x] __References__: Indicate issues fixed via: `Fixes #xxx` or `Closes #xxx`

----

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license.

PR #17015, opened 2021-10-16
feat(logs): add support for cloudwatch logs resource policy 

nom3ad:
CloudFormation now supports [Cloudwatch logs Resource policies](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-resourcepolicy.html)
This PR adds L2 support for it.

And now its possible to grant access to service principals as follows. Previously this was throwing an error - see #5343

```ts
const eventsTargetLogs = new logs.LogGroup(this, 'EventsTargetLogGroup');
eventsTargetLogs.grantWrite(new iam.ServicePrincipal('events.amazonaws.com')).assertSuccess();
```

In future, following custom resource implementation of `LogGroupResourcePolicy` could be replaced.

https://github.com/aws/aws-cdk/blob/83b8df8c390a27e10bf362f49babfb24ee425506/packages/@aws-cdk/aws-elasticsearch/lib/log-group-resource-policy.ts#L25
https://github.com/aws/aws-cdk/blob/a872e672f8990fc3879413e5d797533d3916e1fd/packages/@aws-cdk/aws-events-targets/lib/log-group-resource-policy.ts#L26
https://github.com/aws/aws-cdk/blob/a872e672f8990fc3879413e5d797533d3916e1fd/packages/@aws-cdk/aws-events-targets/lib/log-group-resource-policy.ts#L26

closes #5343

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #28495, opened 2023-12-26
fix(ecs): unnecessary CloudWatch logs ResourcePolicy

### LINKED ISSUE #20313: (logs): Policy Limit Reached for LogGroup ResourcePolicies
### Describe the bug

After resolution of #17544 A new issue has turned up if you tend to create multiple ECS tasks referring multiple log groups.

`Resource handler returned message: "Resource limit exceeded. (Service: CloudWatchLogs, Status Code: 400, Request ID: 25bec134-657e-43c3-ae85-810a0ce56fa0)" (RequestToken: 948dab8b-fac6-2903-695d-f9d825eaea90, HandlerErrorCode: ServiceLimitExceeded)
`
This is because Default [quota for resource policies](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html)
<html><head></head><body>

Resource policies | Up to 10 CloudWatch Logs resource policies per Region per account. This quota 								can't be changed.
-- | --

</body></html>

### Expected Behavior

No error after 10th ECS Task/service

### Current Behavior

Each ECS task creates a new log group finally exhausting this limit.

### Reproduction Steps

Create 10 log groups for ecs and you will start to face this.

### Possible Solution

Let us avoid creating not create Resource Policies for CW Logs until this issue is resolved. I understand this defeats the purpose of lowest privilege, but causes issues.

### WorkAround
separate logGroup if created within taskDefinition and add following in the code for now.

`logGroup.node.tryRemoveChild('Policy')`

### CDK CLI Version

2.24.0

### Framework Version

_No response_

### Node.js Version

v17.9.0

### OS

macOS

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

### LINKED ISSUE #22307: aws-ecs: LogDriver automatically creates log-resource-policies, exhausting unmodifiable resource limit
### Describe the bug

We are running 12 Fargate containers, each of them logging to CloudWatch via LogDriver, creare with cdk.

However, the last 2 fails to deploy, due to
` "Resource limit exceeded. (Service: CloudWatchLogs, Status Code: 400,`

which I understand is fixed at 10 and that is it. No way to change.

However, it looks like log-resource-policy is created automaticall by LogDriver, which uses up the amount.

### Expected Behavior

It is not possible a) not to create a policy or b) reuse a policy

### Current Behavior

A log-resource-policy is created with each LogDriver.awsLogs call.

### Reproduction Steps

Deploy 10 Containers with LogDriver

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.40.0 (build 56ba2ab)

### Framework Version

aws-cdk-lib:2.27.0

### Node.js Version

16.15.1

### OS

Win10

### Language

Java

### Language Version

Kotlin

### Other information

_No response_

sakurai-ryo:
This PR modified to avoid creating unnecessary `ResourcePolicy` in CloudWatch Logs.

## issue summary
The related issue reports an error when using the awslogs driver on ECS.
This error is caused by the creation of a ResourcePolicy in CloudWatch Logs that reaches the maximum number of ResourcePolicies.
https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html

## Current behavior
In some cases, this ResourcePolicy will be created and in other cases it will not be created.
Currently, `Grant.addToPrincipalOrResource` is used to grant permissions to ExecutionRole and Log Group in the ECS taskDef.
https://github.com/aws/aws-cdk/blob/607dccb0fd920d25f0fe2613b83c9830322c439e/packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts#L138
https://github.com/aws/aws-cdk/blob/607dccb0fd920d25f0fe2613b83c9830322c439e/packages/aws-cdk-lib/aws-logs/lib/log-group.ts#L194
https://github.com/aws/aws-cdk/blob/607dccb0fd920d25f0fe2613b83c9830322c439e/packages/aws-cdk-lib/aws-iam/lib/grant.ts#L122

`Grant.addToPrincipalOrResource` first grants permissions to the Grantee (ExecutionRole) and creates a resource base policy for cross account access in cases where certain conditions are not met.
This condition is determined by the contents of the `principalAccount` of the ExecutionRole and the accountID in the `env.account` and whether or not these are Tokens, but in this scenario, cross account access is not necessary.
https://github.com/aws/aws-cdk/blob/607dccb0fd920d25f0fe2613b83c9830322c439e/packages/aws-cdk-lib/aws-iam/lib/grant.ts#L141

Also, when the `LogGroup.grantWrite` call was added to `aws-log-driver.ts`, the ResourcePolicy for logs could not be created from CFn and only granted to the ExecutionRole.
https://github.com/aws/aws-cdk/pull/1291
スクリーンショット 2023-12-27 1 08 20
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html

Therefore, the resource base policy should not be necessary when using the awslogs driver.

## Major changes
This PR changed to grant permissions only to ExecutionRole when using the awslogs driver.
With this fix, ResourcePolicy will no longer be created when using the awslogs driver.
I don't consider this a breaking change, as it changes the content of the generated template, but does not change the behavior of forwarding logs to CloudWatch Logs.
However, if this is a breaking change, I think it is necessary to use the feature flag.

fixes #22307, fixes #20313

----


packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -135,7 +136,16 @@ export class AwsLogDriver extends LogDriver {
       ? `${this.props.maxBufferSize.toBytes({ rounding: SizeRoundingBehavior.FLOOR })}b`
       : undefined;
 
-    this.logGroup.grantWrite(containerDefinition.taskDefinition.obtainExecutionRole());
+    // These policies are required for the Execution role to use awslogs driver.
+    // In cases where `addToExecutionRolePolicy` is not implemented in some cases,
+    // for example, when used from aws-batch construct,
```
sakurai-ryo, line 141: for example,
https://github.com/aws/aws-cdk/blob/607dccb0fd920d25f0fe2613b83c9830322c439e/packages/aws-cdk-lib/aws-batch/lib/ecs-container-definition.ts#L610

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -135,7 +136,16 @@ export class AwsLogDriver extends LogDriver {
       ? `${this.props.maxBufferSize.toBytes({ rounding: SizeRoundingBehavior.FLOOR })}b`
       : undefined;
 
-    this.logGroup.grantWrite(containerDefinition.taskDefinition.obtainExecutionRole());
+    // These policies are required for the Execution role to use awslogs driver.
+    // In cases where `addToExecutionRolePolicy` is not implemented in some cases,
```
paulhcsun, line 140:
```suggestion
    // In cases where `addToExecutionRolePolicy` is not implemented,
```

sakurai-ryo:
@paulhcsun
Thanks! fixed it.

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
```diff
@@ -135,7 +136,16 @@ export class AwsLogDriver extends LogDriver {
       ? `${this.props.maxBufferSize.toBytes({ rounding: SizeRoundingBehavior.FLOOR })}b`
       : undefined;
 
-    this.logGroup.grantWrite(containerDefinition.taskDefinition.obtainExecutionRole());
+    // These policies are required for the Execution role to use awslogs driver.
+    // In cases where `addToExecutionRolePolicy` is not implemented,
+    // for example, when used from aws-batch construct,
+    // use `obtainExecutionRole` instead of `addToExecutionRolePolicy` to grant policies to the Execution role.
```
paulhcsun, line 142: Would explicitly changing the call to `addToPrincipalPolicy` throw an error in mentioned case of `aws-batch`?

paulhcsun (commented):
Overall this looks good to me. Your reasoning for why `resource base policy` shouldn't be needed for aws log driver makes sense. I just have one concern about cases where `addToPrincipalPolicy` is not implemented, if that isn't an issue then I'm good to approve.

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
sakurai-ryo: Is it `addToExecutionRolePolicy` instead of `addToPrincipalPolicy`?
If so, the test will fail like this.

スクリーンショット 2024-01-09 10 42 52

sakurai-ryo:
@paulhcsun 
Thanks for your feedback!

In this case, I believe that `addToPrincipalPolicy` is always implemented.
This is because the `obtainExecutionRole` method creates and returns an ExecutionRole using the `Role` construct if one does not exist, and the `Role` construct implements the `addToPrincipalPolicy` method.

https://github.com/aws/aws-cdk/blob/65505f0ed0935ac8748b03ccf8faffa459c37a4a/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts#L712
https://github.com/sakurai-ryo/aws-cdk/blob/10ed1948beb0f83c1b978da9c0a656aa01a382cb/packages/aws-cdk-lib/aws-iam/lib/role.ts#L521

paulhcsun:
Thanks for clarifying @sakurai-ryo, I think I just read the method names incorrectly there for `addToPrincipalPolicy` and `addExecutionRolePolicy`. A couple more questions to help me understand things a bit better as I'm not too familiar with aws-log-driver:

1. Could you clarify how you reached this conclusion from the attached PR that 

> when the LogGroup.grantWrite call was added to aws-log-driver.ts, the ResourcePolicy for logs could not be created from CFn and only granted to the ExecutionRole. https://github.com/aws/aws-cdk/pull/1291

I don't see any explicit mention of that in the PR itself. 

2. The image attached below the above quote mentions that AWS::Logs::ResourcePolicy resource is used to create a IAM policy. Is this the same workflow for which we are no longer creating the resource policy? And if so would this break anything for users that may be using the generated ResourcePolicy to create IAM policies?

Thanks in advance for taking the time to have this discussion with me to better understand the changes.

packages/aws-cdk-lib/aws-ecs/lib/log-drivers/aws-log-driver.ts
paulhcsun: Yes sorry I misread the function name.

sakurai-ryo:
@paulhcsun 
Thank you, too, for digging deeper into this issue.
The change in the IAM Policy is significant, and we should discuss it thoroughly.

1.
The implementation for setting IAM Policy when using AWSLogs driver was added in 2018 (https://github.com/aws/aws-cdk/pull/1291), while `AWS::Logs::ResourcePolicy` was introduced in 2021 and implemented in the CDK (https://github.com/aws/aws-cdk/pull/17015).
Since no other commits have changed AWSLogs driver codes about granting policy, it should have worked without `AWS::Logs::ResourcePolicy` until 2021.

2.
I do not think this PR breaks anything that may be a breaking change.
As per these documents, access is allowed if either the Identity base policy or the Resource base policy is `Allow` for access by the same account.
In addition, IAM's evaluation logic evaluates the Identity base policy after the Resource base policy, so even if log forwarding is denied by `Permissions Boundary` etc, if it is allowed by ExecutionRole, log forwarding will be permitted without any problem.
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic.html#policy-eval-denyallow
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic.html#policies_evaluation_example

Sorry if I misunderstood what you meant.

paulhcsun:
@sakurai-ryo 
Thanks for your response, those answer both of my questions and I'm comfortable moving forward with this without a feature flag. Thanks again for the contribution, this will be a very helpful win for the community!

paulhcsun (approved):

paulhcsun (approved):

PR #28501, opened 2023-12-27
feat(efs): One-Zone filesystem

### LINKED ISSUE #15864: (aws-efs): support one-zone FileSystem
EFS FileSystem should support one-zone storage class
https://aws.amazon.com/blogs/storage/lower-your-amazon-efs-costs-by-47-with-one-zone-storage-classes/

This is done by setting the `AWS::EFS::FileSystem` `AvailabilityZoneName`
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-efs-filesystem.html#cfn-efs-filesystem-availabilityzonename

### Use Case
It's cheaper

### Proposed Solution
Expose `oneZone` property and have that set `AvailabilityZoneName` to one of the subnet availabilityZones for the subnets the FileSystem is in?

---

badmintoncryer:
In this PR, I have added support for creating one zone filesystems. 

[As discussed in the issue](https://github.com/aws/aws-cdk/issues/15864#issuecomment-895483167), I considered including `availabilityZoneName` in the props, but for now, I have added only the `oneZone` argument.

The `efs.FileSystem` construct also generates a mount target simultaneously. However, when creating a one zone file system, both the mount target and the file system should be located in the same Availability Zone. The issue arises because the name of the Availability Zone where the mount target will be placed (obtained from `props.vpc.selectSubnets()`) remains token, making it impossible to compare it with the given `availabilityZoneName`.

Please let me know if there is a smart way to resolve this.

Closes #15864

----


packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -264,6 +264,15 @@ export interface FileSystemProps {
    * or set `@aws-cdk/aws-efs:denyAnonymousAccess` feature flag, otherwise true
    */
   readonly allowAnonymousAccess?: boolean;
+
+  /**
+   * Creating a One Zone file system.
+   * When set to true, only the GENERAL_PURPOSE performance mode is allowed.
```
lpizzinidev, line 270:
```suggestion
   * Whether this is a One Zone file system.
   * If enabled, `performanceMode` must be set to `GENERAL_PURPOSE`.
```

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -478,6 +487,10 @@ export class FileSystem extends FileSystemBase {
   constructor(scope: Construct, id: string, props: FileSystemProps) {
     super(scope, id);
 
+    if (props.performanceMode === PerformanceMode.MAX_IO && props.oneZone) {
+      throw new Error('OneZone mode file systems do not support the MAX_IO performance mode.');
```
lpizzinidev, line 491:
```suggestion
      throw new Error(`performanceMode MAX_IO is not supported for One Zone file systems.`);
```

packages/aws-cdk-lib/aws-efs/test/efs-file-system.test.ts
```diff
@@ -874,4 +874,39 @@ test('anonymous access is prohibited by the @aws-cdk/aws-efs:denyAnonymousAccess
...
+      'Fn::Select': [
+        0,
+        {
+          'Fn::GetAZs': '',
+        },
+      ],
+    },
+  });
+
+  // MountTargetのサブネットIDを取得
```
lpizzinidev, line 899:
```suggestion
  // make sure only one mount target is created.
```

packages/aws-cdk-lib/aws-efs/README.md
```diff
@@ -51,6 +51,24 @@ const importedFileSystem = efs.FileSystem.fromFileSystemAttributes(this, 'existi
 });
 ```
 
+### One Zone file system
+
+One Zone file systems provide continuous availability to data within a single Availability Zone in an AWS Region.
+Here's a basic example:
```
lpizzinidev, line 57:
```suggestion
To initialize a One Zone file system use the `oneZone` property:
```

packages/aws-cdk-lib/aws-efs/README.md
```diff
@@ -51,6 +51,24 @@ const importedFileSystem = efs.FileSystem.fromFileSystemAttributes(this, 'existi
...
+
+new efs.FileSystem(this, 'OneZoneFileSystem', {
+  vpc,
+  oneZone: true,
+})
+```
+
+⚠️ Performance Mode Compatibility: The oneZone option is not compatible with the MAX_IO performance mode. An error will be thrown if these two options are used together.
+⚠️ Availability Zone Selection: When oneZone is enabled, the file system is automatically placed in the first availability zone of the VPC. It is not currently possible to specify a different availability zone.
+⚠️ Mount Target Creation: When using oneZone, mount targets will be created only in the specified availability zone. This is to prevent deployment failures due to cross-AZ configurations.
```
lpizzinidev, line 70:
```suggestion
⚠️ One Zone file systems are not compatible with the MAX_IO performance mode.

⚠️ When `oneZone` is enabled, the file system is automatically placed in the first availability zone of the VPC.
It is not currently possible to specify a different availability zone.

⚠️ When `oneZone` is enabled, mount targets will be created only in the specified availability zone. 
This is to prevent deployment failures due to cross-AZ configurations.
```

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -549,7 +565,22 @@ export class FileSystem extends FileSystemBase {
...
+    let subnetSelection: ec2.SubnetSelection;
+    if (props.oneZone) {
+      subnetSelection = {
+        availabilityZones: [oneZoneAzName],
+      };
+      if (props.vpcSubnets) {
+        subnetSelection = {
+          ...props.vpcSubnets,
+          ...subnetSelection,
+        };
```
lpizzinidev, line 578: Not sure about this, what happens if the user specifies subnets from different AZs?

lpizzinidev (changes requested):
Thanks 👍
I left some comments for minor documentation adjustments and a question.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
badmintoncryer: Probably, the return value from props.vpc.selectSubnets(subnetSelection) is an empty array, and the mountTarget is not created.

Should it be considered an error when the return value of selectSubnets is an empty array with oneZone=true?

badmintoncryer:
@lpizzinidev Thank you for your review!! I have addressed the comments you provided.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
lpizzinidev: Given: 
* [Amazon EFS One Zone file systems support only a single mount target which is located in the same Availability Zone as the file system. You cannot add additional mount targets.](https://docs.aws.amazon.com/efs/latest/ug/mounting-one-zone.html)
* `selectSubnets` [Requires that at least one subnet is matched, throws a descriptive error message otherwise.](https://github.com/aws/aws-cdk/blob/ca91626327ac4a138adc16026d951f21fadf86a3/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts#L129)

I think that the best option here is to validate that the users can specify either `oneZone` or `vpcSubnets`, not both, to prevent deployment errors and change the implementation accordingly.

badmintoncryer:
@lpizzinidev  I have implemented an error for cases where the vpcSubnets argument is specified in a oneZone file system.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -478,6 +487,14 @@ export class FileSystem extends FileSystemBase {
   constructor(scope: Construct, id: string, props: FileSystemProps) {
     super(scope, id);
 
+    if (props.performanceMode === PerformanceMode.MAX_IO && props.oneZone) {
+      throw new Error('performanceMode MAX_IO is not supported for One Zone file systems.');
+    }
+
+    if (props.oneZone && props.vpcSubnets) {
+      throw new Error('vpcSubnets is not supported for One Zone file systems.');
```
lpizzinidev, line 495:
```suggestion
      throw new Error('vpcSubnets cannot be specified when oneZone is enabled.');
```

packages/aws-cdk-lib/aws-efs/test/efs-file-system.test.ts
```diff
@@ -874,4 +874,50 @@ test('anonymous access is prohibited by the @aws-cdk/aws-efs:denyAnonymousAccess
...
+
+test('one zone file system with vpcSubnets is not supported', () => {
+  // THEN
+  expect(() => {
+    new FileSystem(stack, 'EfsFileSystem', {
+      vpc,
+      oneZone: true,
+      vpcSubnets: { subnetType: ec2.SubnetType.PUBLIC },
+    });
+  }).toThrow(/vpcSubnets is not supported for One Zone file systems./);
```
lpizzinidev, line 922:
```suggestion
  }).toThrow(/vpcSubnets cannot be specified when oneZone is enabled./);
```

lpizzinidev (changes requested):
Thanks, just a minor change on the validation message.

badmintoncryer:
@lpizzinidev Thanks. I fixed it!

lpizzinidev (approved):

paulhcsun (approved):
Thanks for adding this support for One-Zone filesystems @badmintoncryer! And thanks for reviewing @lpizzinidev!

PR #28507, opened 2023-12-28
feat(ses): auto setting for DMARC

watany-dev:
Introducing DMARC functionality to Amazon SES Construct to enhance email security by authenticating sender identity. This feature allows users to enable DMARC for their email domains with minimal configuration.

- Changes
Added autoDmarc flag in EmailIdentity for automatic DMARC TXT record creation.
Optional dmarcReportEmail for custom DMARC report destination.

- Example

```ts
new ses.EmailIdentity(this, 'Identity', {
  identity: ses.Identity.publicHostedZone(myHostedZone),
  autoDmarc: true,
  dmarcReportEmail: 'custom-report@domain.com',
});
```

- Impact
Facilitates easy DMARC setup for Amazon SES users, enhancing domain security against email spoofing.
https://support.google.com/a/answer/2466580

----


watany-dev:
The integ test should have passed locally.

packages/aws-cdk-lib/aws-ses/lib/email-identity.ts
```diff
@@ -419,6 +434,16 @@ export class EmailIdentity extends Resource implements IEmailIdentity {
       });
     }
 
+    if (props.autoDmarc && props.identity.hostedZone) {
```
kaizencc, line 437: If `props.dmarcReportEmail` is set but `props.autoDmarc` is not, then the email will be ignored, which I argue is unexpected. Instead, we should assume that `autoDmarc` is true if `dmarcReportEmail` is set.

kaizencc (dismissed):
just one comment @watany-dev, which should also have its own unit test, and then we are good to go. Thanks!

kaizencc (changes requested):
Looks good @watany-dev but looks like the integ test needs updating?

PR #28509, opened 2023-12-28
feat(iam): validate roleName

### LINKED ISSUE #28502: (iam): The naming conventions for IAM role names have not been verified
### Describe the feature

[There are characters that cannot be used in IAM role names.](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#aws-resource-iam-role-properties) (e.g., "/")

I propose a feature for CDK to verify that it adheres to this naming convention.

### Use Case

If a name containing forbidden characters is set for the role name, no error occurs at the stage of ‘cdk synth’, but it fails during deployment. Therefore, I believe it would be preferable to trigger an error at the synth stage.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.116.1

irrelevant

frattallone:
> Validates roleName

Closes #28502 

----


packages/aws-cdk-lib/aws-iam/lib/role.ts
```diff
@@ -413,6 +413,10 @@ export class Role extends Resource implements IRole {
       physicalName: props.roleName,
     });
 
+    if (props.roleName && !/^[\w_+=,.@-]{1,64}$/.test(props.roleName)) {
```
kaizencc, line 416: We need to cover the case where the roleName is a token, and thus we cannot parse it and see if the roleName is valid. We let these go through to CloudFormation, and there's tons of prior art around CDK that handles these situations. Need a unit test for it as well.

kaizencc (changes requested):
Thanks @frattallone. I got you to pass the automated reviewer but I do have one concern re: tokens before I approve this PR.

Also, the build is failing. The build needs to succeed before we take a second look!

kaizencc (changes requested):
Got some unit tests in other submodules failing after this validation. Methinks either the unit tests were using invalid rolenames, or your check is too restrictive. Either way, the unit tests need to pass.

packages/aws-cdk-lib/aws-iam/lib/role.ts
frattallone: I'm going to add this check and relative tests

paulhcsun (approved):
Additional check and tests have been added for roleName is token case. Great work @frattallone, thanks for the contribution!

PR #28510, opened 2023-12-28
fix(iam): withConditions overrides Principal actions

### LINKED ISSUE #28426: IAM: ArnPrincipal withSessionTags depends on order.
### Describe the bug

Order matters if you are using withConditions and withSessionTags on an ArnPrincipal. 

WithConditions will always work but it appears withSessionTags has to go last.

### Expected Behavior

That it would not matter the order in which you call the methods that both modifications would occur.

### Current Behavior

If you use withSessionTags then use withConditions there is no sts:SessionTag.
If you use withConditions then use withConditions there is sts:SessionTag.

### Reproduction Steps

```
// Works as withSessionTags comes after withCondition
new ArnPrincipal(
      cdk.Arn.format({
        service: 'iam',
        account: '123456789012',
        resource: 'role',
        resourceName: roletoassume',
        region: '',
        partition: 'aws',
      }),
    ).withConditions(
      {
        'StringEquals': {
          'aws:PrincipalTag/Tag1': 'value1',
          'aws:PrincipalTag/Tag2': 'value2',
        }
      }
    ).withSessionTags();
```

```
// Does not work (there is no sts:SessionTags in the resulting policy) as withSessionTags comes before withCondition
new ArnPrincipal(
      cdk.Arn.format({
        service: 'iam',
        account: '123456789012',
        resource: 'role',
        resourceName: roletoassume',
        region: '',
        partition: 'aws',
      }),
    ).withSessionTags().withConditions(
      {
        'StringEquals': {
          'aws:PrincipalTag/Tag1': 'value1',
          'aws:PrincipalTag/Tag2': 'value2',
        }
      }
    );
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.115.0

### Framework Version

_No response_

### Node.js Version

18.17.1

### OS

Ubuntu 22.04 - Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

lpizzinidev:
Calling `withConditions` after `withSessionTags` will override the existing `sts:TagSession` action for the statement.
This implementation fixes the issue.

Closes #28426.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/emr/emr-create-cluster.ts
+++ b/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/emr/emr-create-cluster.ts
@@ -684,6 +684,20 @@ export namespace EmrCreateCluster {
      * Capacity-optimized, which launches instances from Spot Instance pools with optimal capacity for the number of instances that are launching.
      */
     CAPACITY_OPTIMIZED = 'capacity-optimized',
+    /**
+     * Price-capacity-optimized, which launches instances from Spot Instance pools with the highest capacity availability for the number of instances that are launching.
+     *
+     * Recommended.
+     */
+    PRICE_CAPACITY_OPTIMIZED = 'price-capacity-optimized',
+    /**
+     * Lowest-price, which launches instances from the lowest priced pool that has available capacity.
+     */
+    LOWEST_PRICE = 'lowest-price',
+    /**
+     * Diversified, which launches instances across all Spot capacity pools.
+     */
+    DIVERSIFIED = 'diversified',
   }
 
   /**
--- a/packages/aws-cdk-lib/aws-stepfunctions-tasks/test/emr/emr-create-cluster.test.ts
+++ b/packages/aws-cdk-lib/aws-stepfunctions-tasks/test/emr/emr-create-cluster.test.ts
@@ -1,4 +1,4 @@
-import { Template, Match } from '../../../assertions';
+import { Template } from '../../../assertions';
 import * as iam from '../../../aws-iam';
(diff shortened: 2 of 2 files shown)


kylelaker (approved):
Very nice fix! Thank you!

packages/aws-cdk-lib/aws-iam/lib/principals.ts
```diff
@@ -264,6 +264,18 @@ export class PrincipalWithConditions extends PrincipalAdapter {
     this.additionalConditions = conditions;
   }
 
+  public addToAssumeRolePolicy(doc: PolicyDocument) {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const adapter: typeof import('./private/policydoc-adapter') = require('./private/policydoc-adapter');
```
kaizencc, line 269: Ok, I'll ask the question: why are we doing this weird import here? I assume something is stopping us from importing `MutatingPolicyDocumentAdapter` the normal way?

lpizzinidev: To prevent a circular dependency on imports (it seems on `PolicyDocument`) as done [here](https://github.com/aws/aws-cdk/blob/9d8b06f6478a98e01e3aaa86c1dbf22d6e861f05/packages/aws-cdk-lib/aws-iam/lib/principals.ts#L368).
I added a comment to clarify.

kaizencc: Perfect thanks. My next comment was going to be that if I'm confused, so will people who look at the code in the future!

kaizencc (dismissed):

PR #28523, opened 2023-12-29
fix(cli): direct deploy method fails when there are no updates

jogold:
This is correctly handled for empty change sets here:

https://github.com/aws/aws-cdk/blob/9d8b06f6478a98e01e3aaa86c1dbf22d6e861f05/packages/aws-cdk/lib/api/util/cloudformation.ts#L290-L296

but not for direct updates.

----


jogold:
Exemption Request

This change doesn't need an integ test.

packages/aws-cdk/lib/api/deploy-stack.ts
```diff
@@ -478,16 +478,24 @@ class FullCloudFormationDeployment {
...
+        }).promise();
+      } catch (err: any) {
+        if (err.message === 'No updates are to be performed.') {
+          debug('No updates are to be performed for stack %s', this.stackName);
+          return { noOp: true, outputs: this.cloudFormationStack.outputs, stackArn: this.cloudFormationStack.stackId };
+        }
+        throw err;
+      }
+
+      return this.monitorDeployment(startTime, undefined);
```
go-to-k, line 498: Why was the `updateTerminationProtection()` method moved before the `cfn.updateStack` call?

go-to-k, line 498: Why was the `cfn.updateStack` call moved after the `updateTerminationProtection()` method?

rix0rrr (approved):

PR #28529, opened 2023-12-30
feat(stepfunctions-tasks): add validations for EmrCreateCluster

go-to-k:
This PR adds validations for `EmrCreateCluster`.

## timeoutDurationMinutes

> Minimum value is 5 and maximum value is 1440. 

https://docs.aws.amazon.com/emr/latest/APIReference/API_SpotProvisioningSpecification.html

## bidPrice and bidPriceAsPercentageOfOnDemandPrice

Both `bidPrice` and `bidPriceAsPercentageOfOnDemandPrice` are specified, the error occurs in Step Functions console.

```
Specify at most one of bidPrice or bidPriceAsPercentageOfOnDemandPrice value for the Spot Instance fleet : Master request. (Service: AmazonElasticMapReduce; Status Code: 400; Error Code: ValidationException; Request ID: xxxxxx; Proxy: null)
```

## targetOnDemandCapacity and targetSpotCapacity

> At least one of TargetSpotCapacity and TargetOnDemandCapacity should be greater than 0. For a master instance fleet, only one of TargetSpotCapacity and TargetOnDemandCapacity can be specified, and its value must be 1.

https://docs.aws.amazon.com/emr/latest/APIReference/API_InstanceFleetConfig.html

----

### DIFF
--- a/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/emr/private/cluster-utils.ts
+++ b/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/emr/private/cluster-utils.ts
@@ -151,7 +151,7 @@ function SpotProvisioningSpecificationPropertyToJson(property?: EmrCreateCluster
     return undefined;
   }
   if (!cdk.Token.isUnresolved(property.timeoutDurationMinutes) && (property.timeoutDurationMinutes < 5 || property.timeoutDurationMinutes > 1440)) {
-    throw new Error(`timeout duration must be between 5 and 1440 minutes, got ${property.timeoutDurationMinutes}`);
+    throw new Error(`timeoutDurationMinutes must be between 5 and 1440, got ${property.timeoutDurationMinutes}`);
   }
   return {
     AllocationStrategy: cdk.stringToCloudFormation(property.allocationStrategy),
--- a/packages/aws-cdk-lib/aws-stepfunctions-tasks/test/emr/emr-create-cluster.test.ts
+++ b/packages/aws-cdk-lib/aws-stepfunctions-tasks/test/emr/emr-create-cluster.test.ts
@@ -1239,7 +1239,7 @@ test('Throws if timeoutDurationMinutes for Spot instances is less than 5', () =>
   // THEN
   expect(() => {
     task.toStateJson();
-  }).toThrow(/timeout duration must be between 5 and 1440 minutes, got 4/);
+  }).toThrow(/timeoutDurationMinutes must be between 5 and 1440, got 4/);
 });
 
 test('Throws if timeoutDurationMinutes for Spot instances is greater than 1440', () => {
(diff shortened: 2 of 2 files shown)


packages/@aws-cdk-testing/framework-integ/test/aws-lambda-nodejs/test/integ.latest.js.snapshot/LambdaNodeJsLatestIntegDefaultTestDeployAssertD40B5C28.assets.json
go-to-k, line 1: Although not related to this PR, CodeBuild was causing errors due to differences caused by other PRs, so I changed it.

go-to-k: I have submitted another PR. If it gets merged, I will pull it in here as well.

https://github.com/aws/aws-cdk/pull/28530

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/emr/private/cluster-utils.ts
```diff
@@ -146,6 +150,9 @@ function SpotProvisioningSpecificationPropertyToJson(property?: EmrCreateCluster
   if (!property) {
     return undefined;
   }
+  if (!cdk.Token.isUnresolved(property.timeoutDurationMinutes) && (property.timeoutDurationMinutes < 5 || property.timeoutDurationMinutes > 1440)) {
```
kaizencc, line 153: Unrelated to your PR, but I can't believe we have a prop called `timeoutDurationMinutes: number`. It should have been typed and named as `timeout: Duration`. Wonder if you're willing to deprecate `timeoutDurationMinutes` in favor of a new prop in a separate PR

kaizencc (approved):
Thanks :)

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/emr/private/cluster-utils.ts
go-to-k: @kaizencc 

> Wonder if you're willing to deprecate timeoutDurationMinutes in favor of a new prop in a separate PR

OK, I will take it.

go-to-k: @kaizencc 

Oh, this param `timeoutDurationMinutes` is required. This means that even if we have a new parameter, we still need to specify this parameter, and making it optional would be a breaking change.

What should we do?

go-to-k: @kaizencc 

Oh, this param `timeoutDurationMinutes` is required in the interface. ~~This means that even if we have a new parameter, we still need to specify this parameter, and making it optional would be a breaking change.~~

I was mistaken: making a required parameter optional for an interface that is a props is not a breaking change.

But making the new parameter require is a breaking change.

What should we do? Should it be added as a new optional parameter, whereas before it was required?

go-to-k: I'll submit a PR anyway.

go-to-k: @kaizencc 

I have submitted the PR.

https://github.com/aws/aws-cdk/pull/28532#issue-2060755266

PR #28535, opened 2023-12-31
docs(core): update old recommendations in core README

laurelmay:
`splitArn` is the correct updated version of `parseArn`. It's now
necessary to use `ArnFormat` not a literal string for the format of the
ARN resource for `splitArn` and `formatArn`.

----


kaizencc (approved):

kylelaker:
@Mergifyio update

kaizencc (dismissed):

kylelaker:
Okay so it looks like I need to make some small changes to the fixtures. But also it seems like `aws-cdk-lib/README.md` and `aws-cdk-lib/core/README.md` are not currently in sync. So I'm going to try to take care of that here too.

I'll probably also work on a separate PR to add a `prlinter` rule for that?

kylelaker:
I am also curious whether they both actually need to exist? It does seem like `aws-cdk-lib/README.md` has gotten more of these kinds of "chore" syntax fixes while `core/README.md` gets actual content updates. But https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib-readme.html#app-context is the "parent" README file.

kaizencc (approved):
Hmm you're right. My best guess is that we want to keep this behavior so that both show up in the generated docs. So I'm happy with this update and also something in the future to ensure they stay in sync.

It shouldn't be in prlint though. I think it can just be a special case of awslint where we check that these two files match.

kaizencc (approved):
Hmm you're right. My best guess is that we want to keep this behavior so that both show up in the generated docs. So I'm happy with this update and also something in the future to ensure they stay in sync.

It shouldn't be in prlint though. I think it can just be a special case of awslint where we check that these two files match.

PR #28540, opened 2023-12-31
feat(rds): ClientPasswordAuthType property on DatabaseProxy

### LINKED ISSUE #28415: (aws-rds): Support configuring `ClientPasswordAuthType` on DBProxy
### Describe the feature

The `AWS::RDS::DBProxy` resource has a [ClientPasswordAuthType](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-dbproxy-authformat.html#cfn-rds-dbproxy-authformat-clientpasswordauthtype) property that is not available in the L2

### Use Case

Need to be able to configure this on the L2.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.115.0

All

lpizzinidev:
Adds support for [`ClientPasswordAuthType`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-dbproxy-authformat.html#cfn-rds-dbproxy-authformat-clientpasswordauthtype) on `DatabaseProxy` construct.

Closes #28415.

----


packages/aws-cdk-lib/aws-rds/lib/proxy.ts
```diff
@@ -259,6 +281,13 @@ export interface DatabaseProxyOptions {
    * The VPC to associate with the new proxy.
    */
   readonly vpc: ec2.IVpc;
+
+  /**
+   * Specifies the details of authentication used by a proxy to log in as a specific database user.
+   *
+   * @default - The default client password authentication type for the database engine is used.
```
kaizencc, line 288: Minor nit: this `@default` tag doesn't provide any useful information. is there a link to a doc that shows what the defaults are for different database engines?

kaizencc (dismissed):
Just one minor nit @lpizzinidev. Everything else looks good :)

packages/aws-cdk-lib/aws-rds/lib/proxy.ts
lpizzinidev: I wasn't able to find one.
I agree that this is unclear, but I'm not sure how we should state this.
Should we just say that CloudFormation handles the default given the specified engine?

kaizencc: Yeah, lest just say CloudFormation handles the default. Should be clear that its not us making the decisions.

kaizencc (approved):
Thanks @lpizzinidev

PR #28547, opened 2024-01-01
feat(ce): cost anomaly monitor and subscription

jogold:
Add L2 constructs for cost anomaly detection.

----


kaizencc:
Hi @jogold, thanks for your contribution -- under our new contribution policy for new L2 alpha constructs, we want to see additional signal from the community before we invest in reviewing, maintaining, and eventually stabilizing this.

There's a new section in our contribution guide that outlines [demonstrating value](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#demonstrating-value), and specifically the best way to demonstrate value here is to have an existing module with [some] userbase. This will signal to us that both the API is ergonomic and that there is a real use case for higher level CDK constructs for the resource.

In that vein, I encourage you to self-publish this as a new module and have it listed on construct hub. From there, we can monitor usage and see if there is value to include within cdk. I am closing this PR because it is not something that we are ready to accept today. I hope you understand.

PR #28548, opened 2024-01-02
fix(appsync): add dependency to logretention for graphql apis log group

### LINKED ISSUE #26564: appsync: Log group for API is not created at deployment leading to "The specified log group does not exist" error when applying metric filter
### Describe the bug

I am deploying a GraphqlApi construct with a `logConfig` prop set with retention days as well. I am also calling `api.logGroup.addMetricFilter` in the same stack. However, I get this error `Resource handler returned message: "The specified log group does not exist`. It appears that the log group is not created even when the `logConfig` prop is provided upon deployment. This is causing newly deployed stacks to fail.

### Expected Behavior

I would have expected the log group to be created when the API is created so that metric filters could be applied.

### Current Behavior

I get a CloudFormation error of `Resource handler returned message: "The specified log group does not exist` and my stack fails to depoy.

### Reproduction Steps

```
const api = new GraphqlApi(this, 'API', {
  name: 'api',
  schema: SchemaFile.fromAsset('schema.graphql'),
  logConfig: {
    fieldLogLevel: FieldLogLevel.ALL,
    retention: RetentionDays.ONE_WEEK,
    excludeVerboseContent: true
  },
});

api.logGroup.addMetricFilter('MetricFilter', {
  filterPattern: {
    logPatternString: '{ $.fieldName = "myQuery" && $.fieldInError IS TRUE }'
  },
  metricName: 'ErrorCount',
  metricNamespace: 'MyNamespace,
  metricValue: '1',
});
```

### Possible Solution

It appears that the code is simply looking for a log group name but not actually generating its own log group when the API is created. See [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts#L568). A fix for this would be to ensure the log group is actually created rather than trying to load it by name.

Some additional context is that I'm using VTL templates for some of my API resolvers. Because they don't natively have their own log groups or metrics, I'm leveraging adding a metric filter to the main API log group which is where the template resolvers log their logs/errors to.

### CDK CLI Version

2.88.0

### Framework Version

_No response_

### Node.js Version

18

### OS

macOS Ventura 13.4

### Language

Typescript

### Language Version

5.0.4

### Other information

_No response_

mrlikl:
This change adds a dependency to log retention if logconfig is set for `GraphqlApi`. This in turn avoids the race condition when subsequent resources refer to the log group property of `GraphqlApi.logGroup`

logGroup is now referring to the custom resource's output log ARN and further resources will depend on the custom resource to be created first.

Closes #26564

----


packages/@aws-cdk-testing/framework-integ/test/aws-appsync/test/integ.appsync-logmetrics.ts
```diff
@@ -0,0 +1,33 @@
...
+    logPatternString: '{ $.fieldName = "myQuery" && $.fieldInError IS TRUE }',
+  },
+  metricName: 'ErrorCount',
+  metricNamespace: 'MyNamespace',
+  metricValue: '1',
+});
+
+new IntegTest(app, 'Integ', { testCases: [stack] });
+
+app.synth();
```
paulhcsun, line 33:
```suggestion
app.synth();

```

paulhcsun (approved):
Great work @mrlikl!

PR #28566, opened 2024-01-03
fix(core): single-file bundling breaks due to left over temp dir

hoegertn:
This change fixes a bad behavior of the asset bundling if we use the SINGLE_FILE asset type with the OUTPUT hash type.
Because only the created file is moved and the temporary bundle dir is left over, subsequent bundling runs fail and create empty asset files.

----

### DIFF
--- a/packages/aws-cdk-lib/core/lib/asset-staging.ts
+++ b/packages/aws-cdk-lib/core/lib/asset-staging.ts
@@ -346,7 +346,7 @@ export class AssetStaging extends Construct {
     // directory after it has been moved to the staging directory if the hash is known before bundling. This way if bundling
     // is skipped because the bundling directory already exists we can still determine
     // the correct packaging type.
-    // If the hash is calculated after bundling we remove the temp dir now.
+    // If the hash is calculated after bundling we remove the temporary directory now.
     if (bundledAsset.packaging === FileAssetPackaging.FILE) {
       if (this.hashType === AssetHashType.OUTPUT || this.hashType === AssetHashType.BUNDLE) {
         fs.removeSync(path.dirname(bundledAsset.path));


hoegertn:
Exemption Request as this does not change anything in CloudFormation. It is a change to the files that are created/removed in the assembly folder

packages/aws-cdk-lib/core/lib/asset-staging.ts
```diff
@@ -343,11 +343,16 @@ export class AssetStaging extends Construct {
     this.stageAsset(bundledAsset.path, stagedPath, 'move');
 
     // If bundling produced a single archive file we "touch" this file in the bundling
-    // directory after it has been moved to the staging directory. This way if bundling
+    // directory after it has been moved to the staging directory if the hash is known before bundling. This way if bundling
     // is skipped because the bundling directory already exists we can still determine
     // the correct packaging type.
+    // If the hash is calculated after bundling we remove the temp dir now
```
paulhcsun, line 349:
```suggestion
    // If the hash is calculated after bundling we remove the temp dir now.
```

packages/aws-cdk-lib/core/lib/asset-staging.ts
```diff
@@ -343,11 +343,16 @@ export class AssetStaging extends Construct {
     this.stageAsset(bundledAsset.path, stagedPath, 'move');
 
     // If bundling produced a single archive file we "touch" this file in the bundling
-    // directory after it has been moved to the staging directory. This way if bundling
+    // directory after it has been moved to the staging directory if the hash is known before bundling. This way if bundling
     // is skipped because the bundling directory already exists we can still determine
     // the correct packaging type.
+    // If the hash is calculated after bundling we remove the temp dir now.
```
paulhcsun, line 349:
```suggestion
    // If the hash is calculated after bundling we remove the temporary directory now.
```

paulhcsun (approved):
Thanks for adding this fix @hoegertn!

PR #28567, opened 2024-01-03
chore(cdk): Update fact tables and tests for lambda layers v0.90.1

vasireddy99:
Add ADOT Lambda Layer ARNs for recent release

- https://aws-otel.github.io/docs/ReleaseBlogs/aws-distro-for-opentelemetry-lambda-layer-v0.36.0
- https://github.com/aws-observability/aws-otel-lambda/pull/818

----


packages/@aws-cdk-testing/framework-integ/test/aws-lambda/test/integ.lambda-adot.ts
```diff
@@ -33,7 +33,7 @@ class StackUnderTest extends Stack {
       code: Code.fromInline(`exports.handler = ${handler.toString()}`),
       architecture: props.architecture,
       adotInstrumentation: {
-        layerVersion: AdotLayerVersion.fromJavaScriptSdkLayerVersion(AdotLambdaLayerJavaScriptSdkVersion.V1_17_1),
+        layerVersion: AdotLayerVersion.fromJavaScriptSdkLayerVersion(AdotLambdaLayerJavaScriptSdkVersion.V1_18_1),
```
lpizzinidev, line 36:
```suggestion
        layerVersion: AdotLayerVersion.fromJavaScriptSdkLayerVersion(AdotLambdaLayerJavaScriptSdkVersion.LATEST),
```
Can you please update all of these to use `LATEST`?

packages/aws-cdk-lib/aws-lambda/lib/adot-layers.ts
```diff
@@ -398,7 +423,7 @@ export class AdotLambdaLayerGenericVersion extends AdotLambdaLayerVersion {
    * The latest layer version available in this CDK version. New versions could
    * introduce incompatible changes. Make sure to test them before deploying to production.
    */
-  public static readonly LATEST = this.V0_88_0;
+  public static readonly LATEST = this.V0_90_1;
   private constructor(protected readonly layerVersion: string) {
```
lpizzinidev, line 427:
```suggestion
  public static readonly LATEST = this.V0_90_1;

  private constructor(protected readonly layerVersion: string) {
```
Spacing

packages/aws-cdk-lib/aws-lambda/test/adot.test.ts
```diff
@@ -21,18 +21,18 @@ describe('ADOT Lambda Layer', () => {
     });
 
     test('is added properly when the region information is available at synthesis time', () => {
-      const layerArn = lambda.AdotLambdaLayerJavaSdkVersion.V1_31_0.layerArn(fn.stack, fn.architecture);
+      const layerArn = lambda.AdotLambdaLayerJavaSdkVersion.V1_32_0.layerArn(fn.stack, fn.architecture);
```
lpizzinidev, line 24:
```suggestion
      const layerArn = lambda.AdotLambdaLayerJavaSdkVersion.LATEST.layerArn(fn.stack, fn.architecture);
```
Same as with integration, we could use `LATEST` for the unit tests as well.

lpizzinidev (changes requested):
Thanks 👍
I left some notes for minor adjustments on tests.

packages/aws-cdk-lib/aws-lambda/test/adot.test.ts
vasireddy99: This test is specific to get with the version tag, there is already a test for the latest tag. so the changes are not required here. but the rest of the places are changed to LATEST.

lpizzinidev (approved):
Thanks for the changes and the clarification 👍

paulhcsun (approved):
Thanks for adding this @vasireddy99, and thank you for reviewing @lpizzinidev!

PR #28571, opened 2024-01-04
feat(cloudfront): associate key value stores to functions

### LINKED ISSUE #28377: aws-cloudfront: support KeyValueStore
### Describe the feature

Hi! Are there any plans of supporting the new Cloudfront [KeyValueStore](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/kvs-with-functions.html) feature in cdk?

### Use Case

I'd like to
1. create a key value store, and
2. associate the store with a function.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.115.0

Ubuntu 22

laurelmay:
CloudFormation now supports specifying [`KeyValueStoreAssocations`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-function-keyvaluestoreassociation.html) on a `AWS::CloudFront::Function`. This adds that support to the CDK. Even though the field is plural (in CloudFormation and the API), [a function can have one value store](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/kvs-with-functions-associate.html) so the implementation here tries to provide a nicer API around that.

If a Key Value Store is specified, the function will default to `js-2.0` instead of `js-1.0`. Behavior for existing functions is retained. Additionally, an error is thrown if 1.0 is used.

Closes #28377.

----


kylelaker:
@mergifyio update

packages/aws-cdk-lib/aws-cloudfront/README.md
```diff
@@ -558,6 +558,19 @@ const store = new cloudfront.KeyValueStore(this, 'KeyValueStore', {
...
+The Key Value Store can then be associated to a function using the `cloudfront-js-2.0` runtime
+or newer:
+
+```ts
+const store = new cloudfront.KeyValueStore(this, 'KeyValueStore');
+new cloudfront.Function(this, 'Function', {
+  code: cloudfront.FunctionCode.fromInline('function handler(event) { return event.request }'),
+  // Note that JS_2_0 must be used for Key Value Store support
+  runtime: cloudfront.FunctionRuntime.JS_2_0,
+  keyValueStores: [store],
```
kylelaker, line 570:
```suggestion
  keyValueStore: store,
```

packages/aws-cdk-lib/aws-cloudfront/test/function.test.ts
```diff
@@ -188,4 +188,89 @@ describe('CloudFront Function', () => {
...
+        keyValueStore,
+      });
+
+      Template.fromStack(stack).hasResourceProperties('AWS::CloudFront::Function', {
+        FunctionConfig: {
+          Runtime: 'cloudfront-js-2.0',
+        },
+      });
+    });
+
```
go-to-k, line 239: I think this is redundant as it is covered by the 'minimal example' test.

```suggestion
```

packages/aws-cdk-lib/aws-cloudfront/test/function.test.ts
```diff
@@ -188,4 +188,89 @@ describe('CloudFront Function', () => {
...
+        keyValueStore: undefined,
+      });
+
+      Template.fromStack(stack).hasResourceProperties('AWS::CloudFront::Function', {
+        FunctionConfig: {
+          KeyValueStoreAssociations: Match.absent(),
+        },
+      });
+    });
+  });
```
go-to-k, line 275: I don't see much need for this, but what do you think? (unless you intend to make sure that it is not generated with an empty array, or something like that.)

```suggestion
```

go-to-k (changes requested):
Thanks for this PR! I made a few comments.

packages/aws-cdk-lib/aws-cloudfront/test/function.test.ts
kylelaker: Yeah, I think I like it being explicit that it should be left absent instead of perhaps an empty list. I agree it's perhaps covered by other cases or less critical but it does feel like an important way to preserve the correct behavior at a CloudFormation layer.

go-to-k: OK, I agree. Let's go ahead with this.

go-to-k (approved):
Looks good.

packages/@aws-cdk-testing/framework-integ/test/aws-cloudfront/test/integ.function-key-value-store-association.ts
```diff
@@ -0,0 +1,23 @@
+import * as path from 'node:path';
+import * as cdk from 'aws-cdk-lib';
+import { IntegTest } from '@aws-cdk/integ-tests-alpha';
+import * as cloudfront from 'aws-cdk-lib/aws-cloudfront';
+
+const app = new cdk.App();
+const stack = new cdk.Stack(app, 'integ-function-key-value-store-assocation');
```
gracelu0, line 7: nit: spelling of `association` (also applies to line 19)

gracelu0 (commented):
Thank you for contributing! Left a tiny comment but otherwise LGTM :)

gracelu0 (approved):

PR #27592, opened 2023-10-18
feat(secretsmanager): validate maximum value of automaticallyAfter in RotationSchedule

go-to-k:
I added a validation for whether `automaticallyAfter` in `RotationSchedule` is not greater than 1000 days. 

We discussed in the following threads.

https://github.com/aws/aws-cdk/pull/27570#pullrequestreview-1682163803

https://github.com/aws/aws-cdk/pull/27570#pullrequestreview-1684311044

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #28575, opened 2024-01-04
feat(core): add validations for export name in CfnOutput

### LINKED ISSUE #28563: (core): CfnOutput does not validate export name during synthesis but can fail during deployment
### Describe the bug

It is possible to create a `CfnOutput` in the CDK that synthesizes without issue, but then fails to deploy due to an invalid export name.

### Expected Behavior

The `CfnOutput` `exportName` property should be validated according to the cloud formation rules at synthesis time to catch this problem sooner in the process.

### Current Behavior

CDK synthesis works just fine, however during deployment Cloud Formation deployments will fail:
```
Deployment failed: Error [ValidationError]: Template format error: Output <SomeOutputNameHere> is malformed. The Name field of every Export member must be specified and consist only of alphanumeric characters, colons, or hyphens.
```

### Reproduction Steps

``` ts
import { CfnOutput, Stack } from 'aws-cdk-lib';

  constructor(...props: ConstructorParameters<typeof Stack>) {
    super(...props);

    new CfnOutput(this, 'id', {
      value: 'SomeValue',
      exportName: 'SOME_INVALID_EXPORT_NAME`,
    });
  }
}
```

### Possible Solution

Validate the `exportName` field of the `CfnOutput` constructor props during synthesis. Possibly via a regex similar to `[A-Za-z\d\-\:]*`.

I filed this under bug even though there isn't anything technically "wrong" with the CDK code because as a user, I would expect the synthesis to fail if something that can be validated to conform to cloud formation rules can be validated at synthesis time (and indeed in this case it appears at least on the surface that it can be).

### CDK CLI Version

2.114.1

### Framework Version

_No response_

### Node.js Version

18

### OS

MacOS 14 + Amazon Linux 2 5.10

### Language

TypeScript

### Language Version

5

### Other information

_No response_

go-to-k:
Closes #28563.

----


go-to-k:
Exemption Request: This PR just adds validations, so unit tests could cover them.

packages/aws-cdk-lib/core/test/output.test.ts
```diff
@@ -140,7 +140,44 @@ describe('output', () => {
...
+  test('throw if export name has invalid strings (space)', () => {
+    const output = new CfnOutput(stack, 'SomeOutput', { value: 'x', exportName: 'SOME INVALID EXPORT NAME' });
+    const errors = output.node.validate();
+
+    expect(errors).toEqual([
+      expect.stringContaining('Export name must only include alphanumeric characters, colons, or hyphens (got \'SOME INVALID EXPORT NAME\''),
+    ]);
+  });
+
+  test('throw if export name has invalid strings (under_bar)', () => {
```
go-to-k, line 165:
```suggestion
  test('throw if export name has invalid strings (underscore)', () => {
```

kylelaker (approved):

kylelaker:
The implementation looks good to me, though I do wonder if this is more of a bugfix along the lines of:

> fix(core): CfnOutput does not validate export name

go-to-k:
@kylelaker 
Thanks for your review!
Recently, I often get the title of PRs for validation additions corrected to `feat`, so I'm using `feat` again. Do you think it should be changed?

https://github.com/aws/aws-cdk/pull/28529
https://github.com/aws/aws-cdk/pull/27592
https://github.com/aws/aws-cdk/pull/27124
https://github.com/aws/aws-cdk/pull/26241

kylelaker:
> @kylelaker Thanks for your review! Recently, I often look to get the title of PRs for validation additions corrected to `feat`, so I'm using `feat` again. Do you think it should be changed?

Oh interesting! Never mind my suggestion then! Thanks for highlighting that; I didn't know that validation additions were being tracked as features. Makes sense to me either way (though I guess it's possibly a little odd that these validation features will basically always need an exemption since there isn't often a README or integration test that makes sense here...) but I won't disagree with the core team when it comes to naming 😄 Thanks for highlighting those previous PRs and I'll try to follow that pattern myself in the future.

Thanks for adding these new validations for `CfnOutput`!

gracelu0:
Thank you for your contribution, we appreciate it! In the future it would be helpful to include a brief description for the PR as well - makes it a bit easier for us to review :)

gracelu0 (approved):

go-to-k:
@gracelu0 

Thanks for the approval. The merge failed with `Mergify / Queue: Embarked in merge queue`. Could you please handle this again?

gracelu0 (approved):

PR #28583, opened 2024-01-05
feat(opensearchservice): TLS security policy for TLS 1.3 and perfect forward secrecy

go-to-k:
This PR supports new TLS security policy 'Policy-Min-TLS-1-2-PFS-2023-10' for TLS 1.3 and perfect forward secrecy.

The description from [CLI reference](https://docs.aws.amazon.com/cli/latest/reference/opensearch/update-domain-config.html):

> Policy-Min-TLS-1-2-PFS-2023-10: TLS security policy that supports TLS version 1.2 to TLS version 1.3 with perfect forward secrecy cipher suites

- Release notes
  - https://aws.amazon.com/jp/about-aws/whats-new/2024/01/amazon-opensearch-service-tls-1-3-perfect-forward-secrecy/
- CloudFormation
  - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-opensearchservice-domain-domainendpointoptions.html#cfn-opensearchservice-domain-domainendpointoptions-tlssecuritypolicy

----


go-to-k:
Exemption Request: This PR just adds the version as enum property.

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
```diff
@@ -281,7 +281,9 @@ export enum TLSSecurityPolicy {
   /** Cipher suite TLS 1.0 */
   TLS_1_0 = 'Policy-Min-TLS-1-0-2019-07',
   /** Cipher suite TLS 1.2 */
-  TLS_1_2 = 'Policy-Min-TLS-1-2-2019-07'
+  TLS_1_2 = 'Policy-Min-TLS-1-2-2019-07',
+  /** Cipher suite TLS 1.2 to 1.3 with perfect forward secrecy (PFS) */
+  TLS_1_2_PFS = 'Policy-Min-TLS-1-2-PFS-2023-10',
```
lpizzinidev, line 286:
```suggestion
  TLS_1_3 = 'Policy-Min-TLS-1-2-PFS-2023-10',
```
I think it's clearer if we specify `v1.3`.

lpizzinidev (changes requested):
Thanks 👍 Just a minor adjustment on naming in my opinion.

packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
go-to-k: Thanks for your review.

I thought about that too, but do we not have to think about the future appearance of a new option that only allows TLS 1.3 (does not include 1.2)?

Also, there is hesitation in key names that are not in line with the value, what do you think?

lpizzinidev (approved):
> do we not have to think about the future appearance of a new option that only allows TLS 1.3

Good point, thanks for the follow-up 👍

kaizencc (approved):

PR #28589, opened 2024-01-05
Update Kinesis Stream buffering interval limits

abogaziah:
According to aws updates of December 2023, Kinesis Stream buffering interval can be as low as five seconds, however that's not yet reflected in the CDK https://aws.amazon.com/about-aws/whats-new/2023/12/amazon-kinesis-data-firehose-zero-buffering/

----


kaizencc (changes requested):
This PR is not in a state where we would be willing to take a look at it. @abogaziah to get a review, you need to:

- make sure your title fits "conventional commits"
- make sure you have a unit test and that the build is succeeding

Otherwise we cannot review this PR and it will eventually be closed

humanzz:
I just realized I created a feature request for supporting zero buffering at #28714 along with a PR at #28716

PR #28597, opened 2024-01-06
feat(certificatemanager): key algorithm support for `PrivateCertificate` and `Certificate`

### LINKED ISSUE #22887: ACM: Add ability when requesting a PrivateCertificate to choose Key Algorithm
### Describe the feature

In the AWS ACM console, when requesting a new private certificate, I can select between the key algorithms: RSA 2048, ECDSA P 384, and ECDSA P 258. 

In CDK, there is not a way to set the key algorithm and when requesting a cert, and it defaults to RSA 2048(which causes the cert to become a failed state because of the private CA being Elliptic Curve based).

### Use Case

We use AWS private certificate authority service and have a CA set as ECDSA P384. To use this private CA we need to be able to generate certs via CDK. Currently, when creating a private cert with CDK, it defaults to RSA 2048 which lands the certificate in a FAILED state since it doesn't match the private certificate authority config.

### Proposed Solution

Add a `keyAlgorithm` property to the PrivateCertificate [contruct](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_certificatemanager.PrivateCertificate.html). 

```
    const cert = new acm.PrivateCertificate(this, 'certificate', {
      domainName: props.certDomain,
      keyAlgorithm: "ECDSAP384"
      certificateAuthority: acmpca.CertificateAuthority.fromCertificateAuthorityArn(this, 'CA',
        props.certificateAuthorityArn),
    });
```

### Other Information

The only solution that works is to use an RSA based private CA and then cdk cert generation will work since the key algorithm is defaulted to RSA. Otherwise we manually create a certificate in the ACM console and select the key algorithm we expect and reference that certificate ARN in our cdk code.

### Acknowledgements

### CDK version used

2.48.0

macOS version 12.5.1

longtv2222:
Add support for key algorithm when requesting a certificate

`keyAlgorithm` support for `DnsValidatedCertificate` is not supported since the construct was _deprecated_. Added a warning if user tries to use `keyAlgorithm` for the construct instead.

CloudFormation docs for [key algorithm](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-certificatemanager-certificate.html#cfn-certificatemanager-certificate-keyalgorithm).

Closes #22887.

----


packages/aws-cdk-lib/aws-certificatemanager/lib/certificate.ts
```diff
@@ -103,6 +103,36 @@ export interface CertificateProps {
   readonly certificateName?: string
 }
 
+/**
+ * Certificate Manager key algorithm
+ *
+ * If you need to use an algorithm that doesn't exist as a static member, you
+ * can instantiate a `KeyAlgorithm` object, e.g: `new KeyAlgorithm('RSA_2048')`.
+ */
+export class KeyAlgorithm {
```
longtv2222, line 112: Allowed values are: `RSA_1024 | RSA_2048 | RSA_3072 | RSA_4096 | EC_prime256v1 | EC_secp384r1 | EC_secp521r1` but I only included `RSA_2048 | EC_prime256v1 | EC_secp384r1` since we can't import a Certificate in CloudFormation

packages/aws-cdk-lib/aws-certificatemanager/lib/private-certificate.ts
```diff
@@ -28,6 +28,13 @@ export interface PrivateCertificateProps {
    * Private certificate authority (CA) that will be used to issue the certificate.
    */
   readonly certificateAuthority: acmpca.ICertificateAuthority;
+
+  /**
+   * Specifies the algorithm of the public and private key pair that your certificate uses to encrypt data.
+   *
+   * @default KeyAlgorithm.RSA_2048
```
longtv2222, line 35: If `keyAlgorithm` is `undefined`, then the default algorithm is `RSA_2048`

packages/aws-cdk-lib/aws-certificatemanager/lib/certificate.ts
```diff
@@ -101,6 +101,43 @@ export interface CertificateProps {
    * @default the full, absolute path of this construct
    */
   readonly certificateName?: string
+
+  /**
+   * Specifies the algorithm of the public and private key pair that your certificate uses to encrypt data.
+   *
+   * @default KeyAlgorithm.RSA_2048
```
kylelaker, line 108:
```suggestion
   *
   * @see https://docs.aws.amazon.com/acm/latest/userguide/acm-certificate.html#algorithms.title
   * @default KeyAlgorithm.RSA_2048
```

packages/aws-cdk-lib/aws-certificatemanager/lib/certificate.ts
kylelaker: Personally I feel like that's the right thing to do for now too. Having the ability to create additional values probably makes up for any edge cases that could come up.

packages/aws-cdk-lib/aws-certificatemanager/lib/private-certificate.ts
```diff
@@ -28,6 +28,13 @@ export interface PrivateCertificateProps {
    * Private certificate authority (CA) that will be used to issue the certificate.
    */
   readonly certificateAuthority: acmpca.ICertificateAuthority;
+
+  /**
+   * Specifies the algorithm of the public and private key pair that your certificate uses to encrypt data.
+   *
```
kylelaker, line 34: Would it be worth mentioning here that the value needs to match the `KeyAlgorithm` "family" of the Private Certificate Authority?

From CloudFormation:

> When you request a private PKI certificate signed by a CA from AWS Private CA, the specified signing algorithm family (RSA or ECDSA) must match the algorithm family of the CA's secret key.

kylelaker (commented):
Thanks for your work on this! I see the PR is still in draft but I happened to be working on something similar so I hope you don't mind getting some comments now 😄 

This is a really great addition for certificates!

longtv2222:
@kylelaker 

Appreciate the feedbacks!!I've added all your PR suggestions

kylelaker (approved):
Thanks @longtv2222! This looks good to me! This will be a super helpful feature!

packages/aws-cdk-lib/aws-certificatemanager/README.md
```diff
@@ -142,6 +142,7 @@ new acm.PrivateCertificate(this, 'PrivateCertificate', {
   subjectAlternativeNames: ['cool.example.com', 'test.example.net'], // optional
   certificateAuthority: acmpca.CertificateAuthority.fromCertificateAuthorityArn(this, 'CA',
     'arn:aws:acm-pca:us-east-1:123456789012:certificate-authority/023077d8-2bfa-4eb0-8f22-05c96deade77'),
+  keyAlgorithm: acm.KeyAlgorithm.RSA_2048 // optional, default algorithm is RSA_2048
```
kaizencc, line 145:
```suggestion
  keyAlgorithm: acm.KeyAlgorithm.RSA_2048, // optional, default algorithm is RSA_2048
```

kaizencc (approved):
Thanks @longtv2222 and @kylelaker!

PR #28604, opened 2024-01-07
feat(route53): `DNSSEC` zone signing

laurelmay:
This makes it possible to enable DNSSEC signing for a Route 53 hosted zone. A new `KeySigningKey` construct is created to enable generally creating KSKs. Additionally, an `enableDnssec` method is added to `HostedZone` that provides an easier interface. The later will create the KSK and further create the necessary `AWS::Route53::DNSSEC` resource to actually enable signing on the resource.

----


packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -236,6 +236,45 @@ new route53.PublicHostedZone(this, 'HostedZone', {
...
+DNSSEC can be enabled for Hosted Zones. For detailed information, see
+[Configuring DNSSEC signing in Amazon Route 53](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/dns-configuring-dnssec.html).
+
+Enabling DNSSEC requires an asymmetric KMS Customer-Managed Key using the `ECC_NIST_P256` key spec.
+Additionally, that KMS key must be in `us-east-1`.
+
+```ts
+const kmsKey = new kms.Key(this, 'KmsCMK', {
+  keySpec: kms.KeySpec.ECC_NIST_P256,
+  keyUsage: kms.keyUsage.SIGN_VERIFY,
```
kylelaker, line 250:
```suggestion
  keyUsage: kms.KeyUsage.SIGN_VERIFY,
```

packages/aws-cdk-lib/aws-route53/lib/key-signing-key.ts
```diff
@@ -0,0 +1,184 @@
...
+    return new Import();
+  }
+
+  public readonly hostedZone: IHostedZone;
+  public readonly keySigningKeyName: string;
+  public readonly keySigningKeyId: string;
+
+  constructor(scope: Construct, id: string, props: KeySigningKeyProps) {
+    super(scope, id, {
+      physicalName: props.keySigningKeyName ?? Lazy.string({
```
lpizzinidev, line 134: We should add validation for `keySigningKeyName` according to the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-route53-keysigningkey.html#cfn-route53-keysigningkey-name).

packages/aws-cdk-lib/aws-route53/lib/hosted-zone.ts
```diff
@@ -202,6 +233,29 @@ export class HostedZone extends Resource implements IHostedZone {
...
+      hostedZone: this,
+      keySigningKeyName: options.keySigningKeyName,
+      kmsKey: options.kmsKey,
+    });
+    const dnssec = new CfnDNSSEC(this, 'DNSSEC', {
+      hostedZoneId: this.hostedZoneId,
+    });
+    // The KSK must exist and be in an 'ACTIVE' status before DNSSEC can be enabled.
+    dnssec.addDependency(this.keySigningKey.node.defaultChild as CfnKeySigningKey);
+    return this.keySigningKey;
```
lpizzinidev, line 257: Do we need to return the key? 🤔

lpizzinidev (changes requested):
Looks good to me 👍
I left some comments for minor improvements.

msambol:
@kylelaker This one looks really close. Want to address @lpizzinidev comments? I can help get it over the finish line if you're swamped.

packages/aws-cdk-lib/aws-route53/lib/key-signing-key.ts
TheRealAmazonKendra: I don't actually love doing this because if the service team changes this, we have to make an additional code change to allow customers to use the new pattern as well. I tend to think that the deploy time validation should be good enough in this case.

Another option I've been considering is trying to add something into our overall validation at synth that takes in any of these patterns and checks against it, but that's much much larger project than this.

TheRealAmazonKendra (approved):
Given that the only sticking point here was really whether or not to add synth time validation, I think this is ready to go. If it is arguable that we really should have that validation, it isn't a breaking change to add it later.

Thank you for this!

TheRealAmazonKendra:
@Mergifyio update

TheRealAmazonKendra:
Get it together Mergify.

TheRealAmazonKendra (dismissed):

TheRealAmazonKendra (approved):

PR #28612, opened 2024-01-08
feat(signer): Notation-OCI-SHA384-ECDSA platform

### LINKED ISSUE #28580: aws-signer: Signer Profile construct should support Notation-OCI-SHA384-ECDSA platform
### Describe the feature

The L2 construct for creating a Signer Profile in AWS Signer doesn't support creating a singer profile with the `Notation-OCI-SHA384-ECDSA` signing platform. This is odd, given that the L1 construct for it does. As such, I'd like to open a PR for the construct to add this capability, assuming that it just involves adding an instantiation to the `Platform` class.

### Use Case

[AWS documentation](https://docs.aws.amazon.com/signer/latest/developerguide/image-signing-steps.html) recommends using this particular signing profile for creating an AWS Signer configuration for signing Docker images that get pushed to an ECR repository. 

### Proposed Solution

Adding a new enum value for the Platform class. https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-signer/lib/signing-profile.ts#L9

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.x

MacOS

lpizzinidev:
Adds support for the `Notation-OCI-SHA384-ECDSA` signing profile platform.
Also, refactors the `Platform` class to an enum-like class to allow custom platforms (and prevent blocking users if added platforms are missing on the `enum` declaration).

Closes #28580.

----

### DIFF
--- a/packages/@aws-cdk/cloudformation-diff/package.json
+++ b/packages/@aws-cdk/cloudformation-diff/package.json
@@ -23,8 +23,8 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@aws-cdk/aws-service-spec": "^0.0.38",
-    "@aws-cdk/service-spec-types": "^0.0.38",
+    "@aws-cdk/aws-service-spec": "^0.0.40",
+    "@aws-cdk/service-spec-types": "^0.0.40",
     "chalk": "^4",
     "diff": "^5.1.0",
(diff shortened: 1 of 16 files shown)


kaizencc (approved):
Thanks @lpizzinidev. It's a feat not a fix because you're adding in a new value that presumably didn't exist before.

PR #28615, opened 2024-01-08
feat(rds): RDS for SQL Server 16 without a specific minor version

go-to-k:
This PR supports version 16 for RDS for SQL Server without a specific minor version.

----


go-to-k:
Exemption Request: This PR just adds a version.

kaizencc (commented):
@go-to-k we can continue to add any random version specification as a static property, but to what end? Why can't users who need this specific configuration just use the `of` API to create it themselves? Is there a large use case for this particular one that makes the syntactic sugar useful?

go-to-k:
@kaizencc 
I think it is also true that users are using them because we contributors have been kind enough to add these so far, rather than because there are specific benefits or use cases.

Of course, one use of this syntactic sugar would be to be able to specify it easily by simple input or selection by completion in an editor. However, I honestly believe that if this is not there, users will be able to use `of` to deal with it. Maybe we could stop these updates and have the existing minor version designations be labeled `@deprecated` and have them use `of` instead so that users don't expect these updates, or at most just provide static methods for the major versions only. What do you think?

(But I do think that one way to do it is to say that we'll continue to follow the minor version as we have in the past.)

kaizencc (approved):
@go-to-k I see that we have similar versions for 15 and 14 with no specific minor version, so I think we should provide the same support for 16. I know you have been very gracious in adding new versions as they come, but it is also the antithesis of what cdk is about, which is that you can unlock whatever you want via this enum-like class. It's exceedingly simple to specify your own version with `of`, my hope is that people go down that route rather than wait for your manual updates. But maybe thats wishful thinking.

go-to-k:
@kaizencc Yes, I really feel the same way. I wish too.

PR #28625, opened 2024-01-09
fix(glue): `PythonRayExecutableProps` has innaccurate properties

### LINKED ISSUE #28570: (aws-glue-alpha): PythonRayExecutableProps are inaccurate
### Describe the bug

The options provided in `PythonRayExecutableProps` are not all useful for the Ray job type - for instance, the `extraPythonFiles` and `extraFiles` keys correspond to parameters on spark and plain python jobs, but not ray jobs.

This could be a source of confusion (as was for me) for people looking to make use of the ray job `--s3-py-modules` parameter, which currently needs to be provided under `defaultArguments`.

### Expected Behavior

I would expect that any Ray specific arguments would be exposed under `PythonRayExecutableProps` and that non-ray parameters would not be.

### Current Behavior

The `extraPythonFiles` key has no effect

### Reproduction Steps

```
new Job(this, 'job', {
      role: myRole,
      jobName: myJobName,
      timeout: Duration.minutes(1),
      workerType: WorkerType.Z_2X,
      workerCount: 2,
      maxConcurrentRuns: 1,
      executionClass: ExecutionClass.STANDARD,
      executable: JobExecutable.pythonRay({
        glueVersion: GlueVersion.V4_0,
        pythonVersion: PythonVersion.THREE_NINE,
        runtime: Runtime.RAY_TWO_FOUR,
        script: Code.fromBucket("<bucketPath>", "<scriptPath>"),
        // does not work
        // extraPythonFiles: [Code.fromBucket("<bucketPath>", "<dependenciesPath>")],
      }),
      defaultArguments: {
       // does work
        '--s3-py-modules': `s3://${props.assetsBucket.bucketName}/${props.dependenciesPath}`,
      }
    });
```

### Possible Solution

https://github.com/aws/aws-cdk/blob/v2.118.0/packages/@aws-cdk/aws-glue-alpha/lib/job-executable.ts#L256 - `PythonRayExecutableProps` is just an alias over the generic python job props and is not differentiated for Ray, which is a misleading experience. Fix would look like refactoring this to omit irrelevant props and include those missing.

_No response_

### CDK CLI Version

2.113.0

### Framework Version

_No response_

### Node.js Version

v18.19.0

### OS

macOS

### Language

TypeScript

### Language Version

5

### Other information

_No response_

moomindani:
Closes #28570.

- Added RayExecutableProps which supports s3PythonModules
- Added check to block extraPythonFiles usage for Ray jobs
- Added unit tests and integ tests
----


lpizzinidev (approved):
Looks good to me 👍

moomindani:
@lpizzinidev Thank you for looking! Could you please approve and merge this if it is okay?

lpizzinidev:
@moomindani Merging requires a core team member, I'm just a community reviewer 🙂

packages/@aws-cdk/aws-glue-alpha/lib/job-executable.ts
```diff
@@ -466,6 +485,15 @@ export interface JobExecutableConfig {
    */
   readonly extraPythonFiles?: Code[];
 
+  /**
+   * Additional Python modules that AWS Glue adds to the Python path before executing your script.
+   *
+   * @default - no extra python files specified.
+   *
+   * @see `--s3-py-modules` in https://docs.aws.amazon.com/glue/latest/dg/author-job-ray-job-parameters.html
```
kaizencc, line 493: @see expects a url after the tag, and nothing else

packages/@aws-cdk/aws-glue-alpha/test/module/utils.zip
kaizencc, line 1: what is this?

packages/@aws-cdk/aws-glue-alpha/lib/job-executable.ts
```diff
@@ -253,7 +269,7 @@ export interface PythonShellExecutableProps extends SharedJobExecutableProps, Py
 /**
  * Props for creating a Python Ray job executable
  */
-export interface PythonRayExecutableProps extends SharedJobExecutableProps, PythonExecutableProps {}
+export interface PythonRayExecutableProps extends SharedJobExecutableProps, RayExecutableProps {}
```
kaizencc, line 272: what happens to `PythonExecutableProps`? a) is it used elsewhere and b) we are not dropping support for any properties right?

kaizencc: if we are dropping support for properties (as the attached issue suggests) it should have been called out in PR description as a `BREAKING CHANGE`.

kaizencc (changes requested):
hi @moomindani, I have a few concerns before we merge this PR

packages/@aws-cdk/aws-glue-alpha/lib/job-executable.ts
moomindani: a) is it used elsewhere
Yes. There are three job types; Spark, Python shell, and Ray. PythonExecutableProps is used for the first two.

 b) we are not dropping support for any properties right?
Right. We are not dropping any or working properties.
Only the property we are dropping is `extraPythonFiles` which was not working for Ray job. This is not breaking change.

packages/@aws-cdk/aws-glue-alpha/lib/job-executable.ts
moomindani: fixed

moomindani:
@kaizencc Can you take a look and merge it if there are no further update needed?

paulhcsun (approved):
The changes look good to me, nice work @moomindani!

PR #28627, opened 2024-01-09
feat(ecs-pattens): pass healthy percent & deregistration delay params

cheruvian:
The `ecs-patterns` for multiple target groups expose a subset of the available properties for each of the lower level constructs form which it is composed. This subset seems to be manually curated today, however it excludes a number of production critical properties.

This change adds 2 such important properties that I have needed recently.
* `minHealthyPercent` from `AWS::ECS::Service`
* `deregistrationDelay` from `AWS::ElasticLoadBalancingV2::TargetGroup`'s `TargetGroupAttributes`

I believe it may be possible/beneficial to operate on a block list rather than allow list of properties and include all the properties of the included child constructs (including docs, etc...) except for those that don't make sense in a multiple target group context. However, it wasn't clear if this change would be approved and as such I have considered this out of scope for this change, if that is acceptable or desired, I propose that as a different / additional scope of work.

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-ecs-patterns/test/ec2/integ.multiple-application-load-balanced-ecs-service.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-ecs-patterns/test/ec2/integ.multiple-application-load-balanced-ecs-service.ts
@@ -19,7 +19,7 @@ new ApplicationMultipleTargetGroupsEc2Service(stack, 'myService', {
     image: ContainerImage.fromRegistry('amazon/amazon-ecs-sample'),
   },
   enableExecuteCommand: true,
-  minimumHealthyPercent: 33,
+  minHealthyPercent: 33,
   targetGroups: [
     {
       containerPort: 80,
--- a/packages/aws-cdk-lib/aws-ecs-patterns/lib/base/application-multiple-target-groups-service-base.ts
+++ b/packages/aws-cdk-lib/aws-ecs-patterns/lib/base/application-multiple-target-groups-service-base.ts
@@ -122,7 +122,7 @@ export interface ApplicationMultipleTargetGroupsServiceBaseProps {
    * @see https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DeploymentConfiguration.html
    * @default - 0 if daemon, otherwise 50
    */
-  readonly minimumHealthyPercent?: number;
+  readonly minHealthyPercent?: number;
 }
 
 /**
--- a/packages/aws-cdk-lib/aws-ecs-patterns/lib/base/network-multiple-target-groups-service-base.ts
+++ b/packages/aws-cdk-lib/aws-ecs-patterns/lib/base/network-multiple-target-groups-service-base.ts
@@ -113,7 +113,7 @@ export interface NetworkMultipleTargetGroupsServiceBaseProps {
    * @see https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DeploymentConfiguration.html
    * @default - 0 if daemon, otherwise 50
    */
-  readonly minimumHealthyPercent?: number;
+  readonly minHealthyPercent?: number;
 }
 
 /**
--- a/packages/aws-cdk-lib/aws-ecs-patterns/lib/ecs/application-multiple-target-groups-ecs-service.ts
+++ b/packages/aws-cdk-lib/aws-ecs-patterns/lib/ecs/application-multiple-target-groups-ecs-service.ts
@@ -171,7 +171,7 @@ export class ApplicationMultipleTargetGroupsEc2Service extends ApplicationMultip
       enableExecuteCommand: props.enableExecuteCommand,
       placementConstraints: props.placementConstraints,
       placementStrategies: props.placementStrategies,
-      minHealthyPercent: props.minimumHealthyPercent,
+      minHealthyPercent: props.minHealthyPercent,
     });
   }
 }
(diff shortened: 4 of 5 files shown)


packages/aws-cdk-lib/aws-ecs-patterns/README.md
```diff
@@ -83,9 +83,11 @@ const loadBalancedEc2Service = new ecsPatterns.ApplicationMultipleTargetGroupsEc
   taskImageOptions: {
     image: ecs.ContainerImage.fromRegistry("amazon/amazon-ecs-sample"),
   },
+  minHealthyPercent: 33,
   targetGroups: [
     {
       containerPort: 80,
+      deregistrationDelay: 300
```
kylelaker, line 90:
```suggestion
      deregistrationDelay: Duration.seconds(300),
```

packages/aws-cdk-lib/aws-ecs-patterns/README.md
```diff
@@ -108,9 +110,11 @@ const loadBalancedFargateService = new ecsPatterns.ApplicationMultipleTargetGrou
...
```
kylelaker, line 117:
```suggestion
      deregistrationDelay: Duration.seconds(300),
```

kylelaker (commented):
Providing a meaningful review is challenging while the build is failing. Reviewing the build logs, applying these changes to the `README` should fix the failing build.

packages/aws-cdk-lib/aws-ecs-patterns/README.md
cheruvian: Apologies, I couldn't get the tests running on my machine even before changes. Various things complained and/or timed out.

Updated with your corrections.

cheruvian: FWIW confirmed the change itself works by using `patch-package`, though obviously understand we need the tests to pass.

cheruvian:
Any thoughts on this now that the build is passing @kylelaker?

cheruvian:
Found another one that I need today

`stickinessCookieDuration`

Can someone from AWS please advise if `ApplicationTargetProps` should just extend or otherwise delegate `AddApplicationTargetsProps` since that seems to be a majority of it's function. This would also allow for udpates to ALB Target Groups to be taken for 'free' ™ by the ecs patterns

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/network-multiple-target-groups-service-base.ts
```diff
@@ -105,6 +105,15 @@ export interface NetworkMultipleTargetGroupsServiceBaseProps {
...
   readonly enableExecuteCommand?: boolean;
+
+  /**
+   * The minimum number of tasks, specified as a percentage of
+   * the Amazon ECS service's DesiredCount value, that must
+   * continue to run and remain healthy during a deployment.
+   *
+   * @default - 0 if daemon, otherwise 50
+   */
+  readonly minHealthyPercent?: number;
```
msambol, line 116: could you add an `@see` here with a link to the docs?

msambol: also, I think I'd prefer this to be `minimumHealthyPercent` to match CloudFormation.

msambol: @cheruvian I didn't see it in `BaseServiceOptions`. Apologies, you can change it to make that.

msambol: Rest LGTM, if you want to change this back.

cheruvian: Any thoughts on whether this should actually just `extend Omit<BaseServiceOptions, ...>`?

msambol: I think this is fine but we'll see what the maintainer's say.

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/application-multiple-target-groups-service-base.ts
```diff
@@ -267,6 +277,26 @@ export interface ApplicationTargetProps {
...
+  readonly deregistrationDelay?: Duration;
+  /**
+   * The stickiness cookie expiration period.
+   *
+   * Setting this value enables load balancer stickiness.
+   *
+   * After this period, the cookie is considered stale. The minimum value is
+   * 1 second and the maximum value is 7 days (604800 seconds).
+   *
+   * @default Duration.days(1)
```
msambol, line 297: This should actually be `Stickiness is disabled`. See https://github.com/aws/aws-cdk/pull/29726

packages/@aws-cdk-testing/framework-integ/test/aws-ecs-patterns/test/ec2/integ.multiple-application-load-balanced-ecs-service.ts
```diff
@@ -19,9 +19,12 @@ new ApplicationMultipleTargetGroupsEc2Service(stack, 'myService', {
     image: ContainerImage.fromRegistry('amazon/amazon-ecs-sample'),
   },
   enableExecuteCommand: true,
+  minHealthyPercent: 33,
   targetGroups: [
     {
       containerPort: 80,
+      deregistrationDelay: Duration.seconds(10),
+      stickinessCookieDuration: Duration.minutes(4),
```
msambol, line 27: can you add an integration test for NLB as well?

TheRealAmazonKendra (changes requested):
A couple things here:

Firstly +1 to the comments inline. 

Secondly, from the amounts of updates in the integ test files, it looks like they were manually updated and that the test wasn't actually run. Please remove the manual updates and run the test with the `--update-on-failed` flag and then add all the resulting changes to this PR.

PR #28630, opened 2024-01-09
feat(codebuild): new CodeBuild lambda compute images

paavovaananen-cgi:
Add new Lambda compute images for both `aarch64` and `x86_64` arhictectures:

- `aws/codebuild/amazonlinux-aarch64-lambda-standard:corretto21`
- `aws/codebuild/amazonlinux-aarch64-lambda-standard:nodejs20`
- `aws/codebuild/amazonlinux-aarch64-lambda-standard:python3.12`

Reference:
https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html#lambda-compute-images
https://github.com/aws/aws-codebuild-docker-images/pull/687

----


packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
```diff
@@ -21,10 +21,16 @@ interface LinuxArmLambdaBuildImageProps {
 export class LinuxArmLambdaBuildImage implements IBuildImage {
   /** The `aws/codebuild/amazonlinux-aarch64-lambda-standard:nodejs18` build image. */
   public static readonly AMAZON_LINUX_2_NODE_18 = LinuxArmLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-aarch64-lambda-standard:nodejs18');
+  /** The `aws/codebuild/amazonlinux-aarch64-lambda-standard:nodejs20` build image. */
+  public static readonly AMAZON_LINUX_2023_NODE_20 = LinuxArmLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-aarch64-lambda-standard:nodejs20');
```
kaizencc, line 25: Why did you decide to name these `AMAZON_LINUX_2023` when the others are named `AMAZON_LINUX_2`? It looks like the codebuild image id is largely the same `amazonlinux-aarch64-lambda-standard` so I'm curious why this decision was made.

paavovaananen-cgi: It seemed sensible to use that as per the `platform` indicated in this [listing](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html#lambda-compute-images). 

Another, although breaking change, would be to rename all with `AMAZON_LINUX` prefix as it would be more in line with the image id as you pointed out.

paavovaananen-cgi: @kaizencc Do you have any suggestions on how to proceed with this?

jokalli2: I agree that it doesn't make any sense to use the `AMAZON_LINUX_2` prefix as the new image is not based on Amazon Linux 2. If the `AMAZON_LINUX_2023` prefix is unacceptable, perhaps just using `AMAZON_LINUX` for the new images would be the best choice. Personally I wouldn't rename the old images to avoid breaking existing setups.

jokalli2 (approved):

skilbjo:
hi, I was going to do a PR to implement the new node version but I see @paavovaananen-cgi has got it. since the PR is approved, when and who is going to merge it ?

packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
kaizencc: Nope. Makes sense now. Just based on the image id it was hard to tell though

kaizencc (approved):

PR #14511, opened 2021-05-03
feat(cloudfront): add L2 support for CloudFront functions

hoegertn:
Add support for CloudFront functions

/cc @njlynch 

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #28641, opened 2024-01-10
fix(cloudfront): function name is non-deterministic (under feature flag)

### LINKED ISSUE #28629: (cloudfront): Function name is created non-deterministically
### Describe the bug

Creating a CloudFront Function without specifying `name` can result in non-deterministic naming when the `Node` ID is relatively long. This results in unexpected function renames.

### Expected Behavior

The method used to create a CloudFront Function name should be deterministic, predicable, and stable.

### Current Behavior

The name of Functions is unstable, resulting in resource renames and snapshot failures.

### Reproduction Steps

Run the following sample integration test several times. See that eventually it bounces between:
 - `AReallyLonIdIsAlsoPrettyLongIGuess25040025`
 - `AReallyLoIdIsAlsoPrettyLongIGuess25040025`
for the name of the function.

```ts
import * as cloudfront from 'aws-cdk-lib/aws-cloudfront';

const app = new cdk.App();
const stack = new cdk.Stack(app, 'A-Really-Long-Stack-Name-Example');

const fn = new cloudfront.Function(stack, 'ExampleFunctionIdIsAlsoPrettyLongIGuess', {
  code: cloudfront.FunctionCode.fromInline('function handler(event) { return event.request }'),
});

new integ.IntegTest(app, 'FunctionNameLong', {
  testCases: [stack],
});
```

### Possible Solution

https://github.com/aws/aws-cdk/blob/3656a4c06539e68a2eda3d3b98f5d92e956dc37d/packages/aws-cdk-lib/aws-cloudfront/lib/function.ts#L198-L205 Should probably be reworked but that might result in yet again renaming functions. `Names.uniqueResourceName` should probably be used instead.

This is also causing the failure on https://github.com/aws/aws-cdk/pull/28571

### CDK CLI Version

2.118.0

### Framework Version

2.118.0

### Node.js Version

v20.2.0

### OS

Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

laurelmay:
This fixes an issue where having a long stack name or resource ID could
result in inconsistent naming of the function (replacing it on each
deployment) because the length of the token that holds the region may
not be consistent. This uses a more deterministic method based on
checking whether the region is resolved and using the
`uniqueResourceName` function.

This change is behind a feature flag, 
`@aws-cdk/aws-cloudfront:useStableGeneratedFunctionName` to avoid issues
with replacing existing functions.

Closes #28629

----

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudfront/lib/function.ts
+++ b/packages/aws-cdk-lib/aws-cloudfront/lib/function.ts
@@ -1,8 +1,8 @@
 import * as fs from 'fs';
 import { Construct } from 'constructs';
 import { CfnFunction } from './cloudfront.generated';
-import { IResource, Lazy, Names, Resource, Stack, Token } from '../../core';
-import { Fact } from '../../region-info';
+import { FeatureFlags, IResource, Lazy, Names, Resource, Stack } from '../../core';
+import { CLOUDFRONT_FUNCTION_STABLE_GENERATED_NAME } from '../../cx-api';
 
 /**
  * Represents the function's source code
@@ -197,17 +197,17 @@ export class Function extends Resource implements IFunction {
   }
 
   private generateName(): string {
-    // Function names are globally unique, even if the stack is deployed in different regions;
-    // therefore, we prefix the region name to the resource name. It's important; however, to make
-    // sure that we don't exceed the service's length limit. We subtract the length of the region
-    // name from the max length to support the concatenation. If the region is currently unresolved,
-    // then we use the name of the longest known region.
-    const serviceMaxNameLength = 64;
-    const region = Stack.of(this).region;
-    const regionNameLength = !Token.isUnresolved(region) ? region.length : Math.max(...Fact.regions.map((r) => r.length));
-    return Lazy.string({
-      produce: () => `${region}${Names.uniqueResourceName(this, { maxLength: serviceMaxNameLength - regionNameLength })}`,
-    });
+    if (FeatureFlags.of(this).isEnabled(CLOUDFRONT_FUNCTION_STABLE_GENERATED_NAME)) {
+      return Lazy.string({
+        produce: () => Names.uniqueResourceName(this, { maxLength: 64, allowedSpecialCharacters: '-_' }),
+      });
+    }
+
+    const name = Stack.of(this).region + Names.uniqueId(this);
+    if (name.length > 64) {
+      return name.substring(0, 32) + name.substring(name.length - 32);
+    }
+    return name;
   }
 }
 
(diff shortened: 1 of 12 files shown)


packages/aws-cdk-lib/aws-cloudfront/lib/function.ts
```diff
@@ -196,11 +197,17 @@ export class Function extends Resource implements IFunction {
...
-    }
-    return name;
+    // Function names are globally unique, even if the stack is deployed in different regions;
+    // therefore, we prefix the region name to the resource name. It's important; however, to make
+    // sure that we don't exceed the service's length limit. We subtract the length of the region
+    // name from the max length to support the concatenation. If the region is currently unresolved,
+    // then we use the name of the longest known region.
+    const serviceMaxNameLength = 64;
+    const region = Stack.of(this).region;
+    const regionNameLength = !Token.isUnresolved(region) ? region.length : Math.max(...Fact.regions.map((r) => r.length));
```
lpizzinidev, line 207: It seems to me that this can potentially trigger re-deploy for those cases where `region` is not resolved.
Given that the issue needs to be fixed and it's probably going to be a breaking change anyway, I think it may be worth considering a refactor/simplification of the name generation function (under a feature flag) to avoid future issues, eg:
```ts
return Lazy.string({
  produce: () => Names.uniqueResourceName(this, { maxLength: 64, allowedSpecialCharacters: '-_' }),
});
```

kylelaker: I think that including the region name is, unfortunately, a very intentional decision so that the same stack can be deployed in multiple regions. https://github.com/aws/aws-cdk/pull/14511#discussion_r625681979

But then again, any callers can always just pass ``functionName: `${this.env.region}Name`,`` if that's really an use case for them under the feature flag.

kylelaker: Added the feature flag and including a note about region inclusion as part of the `detailsMd`. Let me know what you think!

lpizzinidev (approved):

GavinZZ:
> an issue where having a long stack name or resource ID could
result in inconsistent naming of the function (replacing it on each
deployment)

For my own understanding, how does the existing behaviour cause redeployment if stack name and id are not changed?

kylelaker:
> For my own understanding, how does the existing behaviour cause redeployment if stack name and id are not changed?

You _should_ be able to see this if you run the newly added integration tests a couple of times. You will see the function name property change. The CloudFormation documentation for this resource is wrong -- editing the `FunctionName` property _does_ trigger resource replacement (this can be seen by viewing the CloudFormation stack events after the integration test shows a changed name). I believe there is more information in the linked issue.

While replacing a CloudFront function is a minimally disruptive operation, it _is_ still disruptive and a resource _has_ been replaced. We should avoid arbitrarily changing resource physical names within the CDK.

GavinZZ (dismissed):
LGTM, thanks for contributing.

GavinZZ:
@kylelaker would you be able to fix the merge conflicts? This PR looks good and I'm happy to approve once conflicts are resolved.

Leo10Gama:
Closing this as it seems the PR has not been active for quite some time, and [another PR](https://github.com/aws/aws-cdk/pull/30392) has been opened to resolve this issue.

PR #28643, opened 2024-01-10
fix(neptune-alpha): multiple `cloudwatchLogsExports` cannot be set when configuring log retention

### LINKED ISSUE #26295: (neptune-alpha): Unable to set custom retention period for multiple CloudWatch log exports
### Describe the bug

I have created a Neptune cluster with the `DatabaseCluster` construct via `aws-neptune-alpha`. I enabled audit and slowquery logging, and I would like to export these logs to CloudWatch. However, I am unable to set a custom retention period for both sets of logs due to a bug related to string interpolation. The build will fail with the error message "Error: There is already a Construct with name `[object Object]LogRetention` in DatabaseCluster ...".  Since the interpolated LogType object always resolves to `[object Object]`, CDK throws errors stating that such a construct already exists when more than one log type is exported.  

### Expected Behavior

Users should be able to export more than one log type to CloudWatch and set a custom retention period using `cloudwatchLogsExports` and `cloudwatchLogsRetention` like so:

```
this.databaseCluster = new DatabaseCluster(
      this,
      `${name}`,
      {
        ...
        ...
        cloudwatchLogsExports: [new LogType('audit'), new LogType('slowquery')],
        cloudwatchLogsRetention: RetentionDays.ONE_MONTH,
      }
    )
```

### Current Behavior

When a custom retention period is set for more than one log type exported to CloudWatch, the error `There is already a Construct with name '[object Object]LogRetention' in DatabaseCluster {databaseCluster name}` is thrown. 

### Reproduction Steps

Instantiate a new Neptune DatabaseCluster instance using aws-neptune-alpha, and enable audit logging and slow query logging. Then enable exporting these logs to CloudWatch with a custom retention period like so:

```
this.databaseCluster = new DatabaseCluster(
      this,
      `${name}`,
      {
        ...
        ...
        cloudwatchLogsExports: [new LogType('audit'), new LogType('slowquery')],
        cloudwatchLogsRetention: RetentionDays.ONE_MONTH,
      }
    )
```

### Possible Solution

Interpolating the string value of the object instead of the object itself should fix the issue (ie `${logType.value}LogRetention` instead of `${logType}LogRetention`).

_No response_

### CDK CLI Version

2.87.0

### Framework Version

_No response_

### Node.js Version

v19.7.0

### OS

macOS/Linux

### Language

Typescript

### Language Version

3.9.4

### Other information

_No response_

badmintoncryer:
This PR includes a breaking change to the aws-neptune-alpha module.

I have resolved an issue where it was not possible to set multiple `cloudwatchLogsExports` when configuring log retention for a `DatabaseCluster`.

The root cause of the problem was that the LogRetention creation included `[object object]` in the construct's ID. With the current fix, the given `logType` can now be correctly included in the `LogRetention`.
```diff
- new logs.LogRetention(this, `${logType}LogRetention`, {..} // "DatabaseobjectObjectLogRetentionA247AF0C"
+ new logs.LogRetention(this, `${logType.value}LogRetention`, {..} // "DatabaseauditLogRetention8A29F5CC"
```

BREAKING CHANGE: Corrected LogRetention IDs for DatabaseCluster. Previously, regardless of the log type, the string ‘objectObject’ was always included, but after the correction, the log type is now included.

Closes #26295.

----


packages/@aws-cdk/aws-neptune-alpha/test/integ.cluster.ts
```diff
@@ -37,6 +38,7 @@ const cluster = new DatabaseCluster(stack, 'Database', {
   vpc,
   vpcSubnets: { subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS },
   instanceType: InstanceType.R5_LARGE,
+  engineVersion: EngineVersion.V1_2_1_0,
```
badmintoncryer, line 41: If `engineVersion` is not specified, CFN will create the cluster using the latest engine version. Currently, the latest engine is the v1.3 series, which the existing CDK code does not support. 
(If the parameter group is also not set to v1.3, the deployment will fail.)

Therefore, in this integration test, the engine and parameter group have been explicitly set to v1.2.

badmintoncryer: I plan to address the support for the v1.3 in a separate issue.

badmintoncryer:
@paulhcsun I mistakenly included a feature flag in my implementation once, and it seems that a CLI test label got attached at that time. Thank you for removing it!

paulhcsun (dismissed):
The fix itself looks good to me but because this is a breaking change we'll need to add it behind a feature flag. You can refer to the [Feature Flags section](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#feature-flags) of the Contributing guide on how to set that up.

badmintoncryer:
@paulhcsun This is for the alpha module, so I thought breaking changes would be allowed. Is that not the case?
Since I've already implemented feature flags once, I can make the necessary adjustments quickly.

paulhcsun:
@badmintoncryer You're right, sorry I misread the section in the Contributing guide on allowing breaking. A feature flag won't be necessary here and the breaking change callout looks correct to me.

paulhcsun (dismissed):
@badmintoncryer You're right, sorry I misread the section in the Contributing guide on allowing breaking changes. Codewise this looks good to me, could you just update your PR description so that the `BREAKING CHANGE:` callout includes an explanation of the behaviour before and the behaviour now? I'm good to approve after that's been updated.

badmintoncryer:
@paulhcsun Thank you for confirming. I have updated the description. Could you please check it again?

packages/@aws-cdk/aws-neptune-alpha/test/integ.cluster.ts
paulhcsun: Sounds good, feel free to ping me for a review when you get around to that.

paulhcsun (approved):
The breaking change callout looks good to me. Nice work @badmintoncryer!

packages/@aws-cdk/aws-neptune-alpha/test/integ.cluster.ts
badmintoncryer: This problem is resolved by #28647.
Do I need to conduct this integration test with version 1.3, or is it okay to proceed as it is?

badmintoncryer:
Why is the CodeBuild job in a pending state..??

badmintoncryer:
@paulhcsun Could you please re-run the CodeBuild job?

paulhcsun (dismissed):

paulhcsun (dismissed):

paulhcsun (approved):

PR #28649, opened 2024-01-10
docs: spelling grammar and brands for contributing doc

philipws:
When reading through the CONTRIBUTING.md there was a few issues which confused me. These changes correct those so the next new contributor can read it more easily. Also, makes brand names consistent in the file and with the brands public name (E.g. "cloudformation" to "CloudFormation").

----


kaizencc (approved):
Thanks @philipws. Thanks for also reading our contributing guide :)

Issue #22141, opened 2022-09-20
(aws-rds): Can't destroy a stack with a database that has `removalPolicy` property set to `RemovalPolicy.RETAIN`


maxd:
### Describe the bug

I have a stack that creates a database with parameters group:

```ts
const parameterGroup = new ParameterGroup(this, 'ParameterGroup', {
    ...
}

const database = new DatabaseInstance(this, 'DatabaseInstance', {
    parameterGroup: parameterGroup,
    removalPolicy: RemovalPolicy.RETAIN,
    ...
})
```

As you can see the database has `removalPolicy` property set to `RETAIN` because I want to leave the database after destroying this stack.

When I destroy this stack I see the following errors:

```
2:04:24 PM | DELETE_FAILED        | AWS::RDS::DBParameterGroup                  | ParameterGroup5E32DECB
One or more database instances are still members of this parameter group xxx-database-parametergroup5e32decb-daetrwpaqpgw, so the group cannot be deleted (Service: Rd
s, Status Code: 400, Request ID: 389b18db-ea82-482b-a0e6-f64887da6f82)

2:19:21 PM | DELETE_FAILED        | AWS::EC2::SecurityGroup                     | DatabaseInstanceSecurityGroup8BDF0112
resource sg-0bfc8aacb3d3e3d4a has a dependent object (Service: AmazonEC2; Status Code: 400; Error Code: DependencyViolation; Request ID: 1eac5393-83df-48cf-bd75-41f25abb04
7a; Proxy: null)
```

As I understand the CF leaves the database but tries to destroy `ParameterGroup` and `SecurtyGroup`. I suppose this is the wrong behavior and CF should leave related `ParameterGroup` and `SecurityGroup` too.

I can create a custom `SecurityGroup` and change it remove policy use the `applyRemovalPolicy` method:

```
    const securityGroup = new SecurityGroup(this, 'SecurityGroup', {
      vpc: props.vpc,
      allowAllOutbound: true,
    })
    securityGroup.applyRemovalPolicy(removalPolicy(SecurityGroup))
```

BUT I can't change the removal policy for a `ParameterGroup` in the same way because `parameterGroup.applyRemovalPolicy(RemovalPolicy.RETAIN)` raises the following error:

```
Error: Cannot apply RemovalPolicy: no child or not a CfnResource. Apply the removal policy on the CfnResource directly.
[...]

PR #28660, opened 2024-01-10
fix(rds): RDS Parameter Group doesn't support custom removal policy

GavinZZ:
>  Can't destroy a stack that includes a rds database and rds parameter group where the database has removalPolicy property set to RemovalPolicy.RETAIN

### The following is the current behaviour:
```
const parameterGroup = new ParameterGroup(this, 'ParameterGroup', {
    ...
}

const database = new DatabaseInstance(this, 'DatabaseInstance', {
    parameterGroup: parameterGroup,
    removalPolicy: RemovalPolicy.RETAIN,
    ...
})
```

When destroying the stack
```
When I destroy this stack I see the following errors:

2:04:24 PM | DELETE_FAILED        | AWS::RDS::DBParameterGroup                  | ParameterGroup5E32DECB
One or more database instances are still members of this parameter group xxx-database-parametergroup5e32decb-daetrwpaqpgw, so the group cannot be deleted (Service: Rd
s, Status Code: 400, Request ID: 389b18db-ea82-482b-a0e6-f64887da6f82)

2:19:21 PM | DELETE_FAILED        | AWS::EC2::SecurityGroup                     | DatabaseInstanceSecurityGroup8BDF0112
resource sg-0bfc8aacb3d3e3d4a has a dependent object (Service: AmazonEC2; Status Code: 400; Error Code: DependencyViolation; Request ID: 1eac5393-83df-48cf-bd75-41f25abb04
7a; Proxy: null)

```

As pointed out in the issue linked below, we cannot simply use the clusterRds' or instanceRds' removal policy because the parameter group can be simultaneously binded to a cluster and an instance. 

### New behaviour:
Add an optional property `removalPolicy` to the L2 Parameter Group resource and set the deletion policy to the generated L1 Parameter Group (Either cluster or instance) depending on the usage. 

Added unit test and integration test to verify that it works as expected.

Closes https://github.com/aws/aws-cdk/issues/22141

----


packages/@aws-cdk-testing/framework-integ/test/aws-rds/test/integ.instance-with-parameter-group.ts
```diff
@@ -0,0 +1,38 @@
...
+});
+
+new rds.DatabaseInstance(stack, 'Instance', {
+  engine: rds.DatabaseInstanceEngine.postgres({ version: rds.PostgresEngineVersion.VER_15_2 }),
+  vpc,
+  multiAz: false,
+  publiclyAccessible: true,
+  iamAuthentication: true,
+  parameterGroup,
+  removalPolicy: cdk.RemovalPolicy.DESTROY,
```
paulhcsun, line 26: The original issue was not working when `removalPolicy: RemovalPolicy.RETAIN` was set. I know for an integ test we can't leave behind the database but I just want to confirm that either:
1.  The issue wasn't specific to the `RETAIN` removal policy but rather any removal policy set for the databaseInstance or
2. If the issue was specific to the `RETAIN` removal policy that you were able to verify locally that the stack was sucessfully destroyed when the removal policy was `RETAIN`.

If the second case is true maybe we could update this integ test to use `RemovalPolicy.RETAIN` and verify that the stack was destroyed before cleaning up the databaseInstance?

packages/aws-cdk-lib/aws-rds/lib/parameter-group.ts
```diff
@@ -83,6 +83,14 @@ export interface ParameterGroupProps {
    * @default - None
    */
   readonly parameters?: { [key: string]: string };
+
+  /**
+   * The CloudFormation policy to apply when the instance is removed from the
+   * stack or replaced during an update.
```
paulhcsun, line 89: Is this policy always needed or can we update the description to be more specific that it's only needed when the stack also has a `DatabaseInstance`?

paulhcsun (commented):
Overall code looks good to me, just a couple questions about the integ test and policy description.

packages/@aws-cdk-testing/framework-integ/test/aws-rds/test/integ.instance-with-parameter-group.ts
GavinZZ: Yes, this issue is not specific to `RETAIN` rather setting any removal policy. I confirmed <2> by deploying a test stack with setting DBInstance and ParameterGroup to `RETAIN` and deleting the stack. I can confirm that I'm no longer encountering the issue described in the issue #22141.

We can't update the integ test to use `RETAIN` as we won't be able to cleanup the retained resources once the stack is deleted, which will cause leftover resources.

packages/aws-cdk-lib/aws-rds/lib/parameter-group.ts
GavinZZ: This doesn't have to be used in couple of DBInstance. It's technically a standalone removal policy that applies to ParameterGroup.

paulhcsun (approved):

Issue #16253, opened 2021-08-27
(stepfunctions-tasks): LambdaInvoke - explicit null is not passed to lambda when using


TomasChmelik:
When I try to pass `null` as some parameter of a lambda, it gets removed from resulting task's parameters.

It might have same result for javascript lambdas, but in Python there is a difference between key not being in input event and key being there with value `null` (or `None` in Python's case).

This issue can be worked around by adding additional checks in the lambda, but that would add additional step needed for migration to CDK.

### Reproduction Steps
```python
from aws_cdk import aws_lambda
from aws_cdk import aws_stepfunctions
from aws_cdk import aws_stepfunctions_tasks


some_lambda = aws_lambda.Function.from_function_arn(scope, 'SomeLambda', 'arn:aws:lambda:eu-central-1:123456789:function:some-lambda')

aws_stepfunctions_tasks.LambdaInvoke(
    scope,
    'Lambda',
    lambda_function=some_lambda,
    payload=aws_stepfunctions.TaskInput.from_object({
        'nullParameter': None,
        'notNullParameter': 'not null'
    }),
    payload_response_only=True,
    retry_on_service_exceptions=False
)
```

### What did you expect to happen?
```json
{
    "Type": "Task",
    "Resource": "arn:aws:lambda:eu-central-1:123456789:function:some-lambda",
    "Parameters": {
        "nullParameter": null,
        "notNullParameter": "not null"
    }
}
```

### What actually happened?
```json
{
    "Type": "Task",
    "Resource": "arn:aws:lambda:eu-central-1:123456789:function:some-lambda",
    "Parameters": {
        "notNullParameter": "not null"
    }
}
```

### Environment

  - **CDK CLI Version  :1.118.0**
  - **Framework Version:???**
  - **Node.js Version:v15.14.0**
  - **OS               :Ubuntu 18.04.5 LTS**
  - **Language (Version):Python 3.7.4**

### Other

--- 

This is :bug: Bug Report

TomasChmelik:
Typescript has `null` and `undefined` which are not strictly equal, so you could use one of them to specify that property is not there (probably `undefined`, since when you try to access property on object which is not there, you will also get an `undefined`) and other one for `null` valued property.

BenChaimberg:
Ah, my bad, glazed over the value in particular. I believe it is this condition within `renderObject` that needs to be fixed. Only `undefined` should be ignored, `null` should be returned.

[...]

PR #28661, opened 2024-01-10
docs(stepfunctions): use `JsonPath.DISCARD` in place of null for input

paulhcsun:
Update docs for use of `sfn.JsonPath.DISCARD` in place of `null` in `TaskInput.fromObject`s field value for languages that do not support `null` like Python. 

Doc: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions.TaskInput.html#:~:text=to%20a%20task.-,static%20fromObject(obj),-public%20static%20fromObject

Closes https://github.com/aws/aws-cdk/issues/16253.

----


mrgrain:
Does `JsonPath.DISCARD` not work?

paulhcsun:
@mrgrain `JsonPath.DISCARD` does work, I'm just not sure if the name of it is accurate for the purpose of replacing a `null` though, especially because the effect of using a `null` is not consistent across different cases. Taking one example provided in the issue:
```
"ItemResult": {
	"LastEvaluatedKey": JsonPath.DISCARD
     }
```
The name `DISCARD` here isn't the clearest of what we're trying to do but using `NULL_VALUE` is. 

Actually reading the [doc](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions.JsonPath.html#:~:text=Description-,static%20DISCARD,-string) on it, it could make sense (`Special string value to discard state input, output or result.`). Perhaps I'm just thinking of `DISCARD` in a different context. Let me know what you think though.

packages/aws-cdk-lib/aws-stepfunctions/lib/private/json-path.ts
```diff
@@ -142,7 +149,9 @@ export function recurseObject(obj: object | undefined, handlers: FieldHandlers,
 
   const ret: any = {};
   for (const [key, value] of Object.entries(obj)) {
-    if (typeof value === 'string') {
+    if (value === NULL_SUBSTITUTE) {
+      Object.assign(ret, handlers.handleNull(key, null));
```
GavinZZ, line 153: Does the following not work?
```
if (value === null || value === undefined) {
     Object.assign(ret, handlers.handleNull(key, null));
}
```

paulhcsun: I think users want to specifically be able to distinguish `null` from `undefined` in typescript, but at the same time the two are the same in other languages like Python (`None`) so a magic value of some sort will satisfy both cases.

mrgrain:
> Actually reading the [doc](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions.JsonPath.html#:~:text=Description-,static%20DISCARD,-string) on it, it could make sense (`Special string value to discard state input, output or result.`). Perhaps I'm just thinking of `DISCARD` in a different context. Let me know what you think though.

I believe this has been named based on this feature and assuming that the only time someone actually wants `null` in the State Machine definition is for this reason. It's still the only reference to `null` that I can find in the Developer Guide, but that doesn't mean much.

https://docs.aws.amazon.com/step-functions/latest/dg/input-output-resultpath.html

If you look at the PRs and issue involved in `JsonPath.DISCARD` you might get some more context on it. Look out for comments and PRs by @RomainMuller 

> @mrgrain `JsonPath.DISCARD` does work, I'm just not sure if the name of it is accurate for the purpose of replacing a `null` though, especially because the effect of using a `null` is not consistent across different cases. Taking one example provided in the issue:

If we find that the naming/documentation for `JsonPath.DISCARD` isn't working for our users, we can definitely think about improving this. Aside from sprucing up the docs, I could imagine adding an alias for `JsonPath.DISCARD`, e.g. `JsonPath.NULL`.

packages/aws-cdk-lib/aws-stepfunctions/lib/private/json-path.ts
mrgrain: @paulhcsun's right. This isn't an issue in TS (at least not as far as I know), but in jsii (languages) which cannot distinguish between `null` and `undefined`. But ASL (as a JSON-ish schema) sometimes requires users to use `null`

kaizencc (approved):

PR #28669, opened 2024-01-11
fix(cli): running the cli with --debug does not print stack traces

mrgrain:
`--debug` exists for exactly one purpose: Printing source-mapped traces so we can find the code that is going wrong. 
Let's always enabled tracing when debugging.

----


packages/aws-cdk/lib/cli.ts
```diff
@@ -315,16 +315,17 @@ if (!process.stdout.isTTY) {
...
   }
 
-  if (argv.verbose) {
-    setLogLevel(argv.verbose);
-
-    if (argv.verbose > 2) {
-      enableTracing(true);
-    }
+  // Debug should always imply tracing
+  if (argv.debug ||argv.verbose > 2) {
```
kaizencc, line 327:
```suggestion
  if (argv.debug || argv.verbose > 2) {
```

kaizencc (approved):

PR #28671, opened 2024-01-11
chore(appconfig): fix some awslint errors, explicitly exempt others

### LINKED ISSUE #27894: (appconfig): Linter errors must be audited
### Describe the bug

AppConfig L2 in alpha stage was released without the linter checker turned on, so none of these linter exceptions have been audited. These rules likely should be fixed rather than exempted, which could result in breaking changes. Marking this as a `p1` because this issue _should_ block any potential graduation of the AppConfig module. 

As seen in the awslint.json file:

```json
{
        "no-unused-type:@aws-cdk/aws-appconfig-alpha.PredefinedDeploymentStrategyId",
        "ref-via-interface:@aws-cdk/aws-appconfig-alpha.Application.addAgentToEcs.taskDef",
        "props-physical-name:@aws-cdk/aws-appconfig-alpha.ApplicationProps",
        "props-physical-name:@aws-cdk/aws-appconfig-alpha.DeploymentStrategyProps",
        "props-physical-name:@aws-cdk/aws-appconfig-alpha.EnvironmentProps",
        "props-physical-name:@aws-cdk/aws-appconfig-alpha.ExtensionProps",
        "events-in-interface",
        "events-method-signature",
        "events-generic"
}
```

### Expected Behavior

No linter errors

### Current Behavior

Linter rules are being exempted

### Reproduction Steps

n/a

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.100.0

### Framework Version

_No response_

### Node.js Version

20

### OS

Mac

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

chenjane-dev:
Auditing awslint errors.

Reasons for keeping current list of exemptions:

* "no-unused-type:@aws-cdk/aws-appconfig-alpha.PredefinedDeploymentStrategyId" - will be fixed in a separate PR.
* "props-physical-name:@aws-cdk/aws-appconfig-alpha.ApplicationProps" 
 "props-physical-name:@aws-cdk/aws-appconfig-alpha.DeploymentStrategyProps"
 "props-physical-name:@aws-cdk/aws-appconfig-alpha.EnvironmentProps"
"props-physical-name:@aws-cdk/aws-appconfig-alpha.ExtensionProps" - will be fixed in a separate PR.
* "events-method-signature" - exempting because appconfig handles their events without cloudwatch events
* "events-generic" - exempting because appconfig handles their events without cloudwatch events

Closes #27894 alongside #28742 and #28743

----


packages/@aws-cdk/aws-appconfig-alpha/awslint.json
```diff
@@ -26,12 +26,10 @@
         "docs-public-apis:@aws-cdk/aws-appconfig-alpha.IApplication",
 
         "no-unused-type:@aws-cdk/aws-appconfig-alpha.PredefinedDeploymentStrategyId",
-        "ref-via-interface:@aws-cdk/aws-appconfig-alpha.Application.addAgentToEcs.taskDef",
         "props-physical-name:@aws-cdk/aws-appconfig-alpha.ApplicationProps",
         "props-physical-name:@aws-cdk/aws-appconfig-alpha.DeploymentStrategyProps",
         "props-physical-name:@aws-cdk/aws-appconfig-alpha.EnvironmentProps",
         "props-physical-name:@aws-cdk/aws-appconfig-alpha.ExtensionProps",
-        "events-in-interface",
```
paulhcsun, line 34: How is this awslint error fixed in this PR?

chenjane-dev: Fixed by adding these methods to the application interface

https://github.com/aws/aws-cdk/pull/28671/files#diff-95c038ca91ae8a09cd48e458bab9f54f352b47fc300fda6502045c1fbf4e26fb

kaizencc (dismissed):
@chenjane-dev I've done my best to explain the specific linter errors below. There's a few points that require further discussion that I will follow up with later.

---

"no-unused-type:@aws-cdk/aws-appconfig-alpha.PredefinedDeploymentStrategyId" - This is for customers to use in case they want to import an AppConfig predefined deployment strategy

- the reason why this linter error flares up in this use case is because we are specifying an `enum` into a property that allows a `string`. I'm not sure if this even works for non-TS languages, but either way, this is a CDK anti-pattern. The correct path forward imo is to make a `DeploymentStrategyId` enum-like class with a public constructor to specify arbitrary strings and the same static methods for the `Predefined` Ids. And make that the type for `deploymentStrategyId`, which is a breaking change. 

---

"props-physical-name:@aws-cdk/aws-appconfig-alpha.ApplicationProps"
"props-physical-name:@aws-cdk/aws-appconfig-alpha.DeploymentStrategyProps"
"props-physical-name:@aws-cdk/aws-appconfig-alpha.EnvironmentProps"
"props-physical-name:@aws-cdk/aws-appconfig-alpha.ExtensionProps" - These names are not prefixed by [cfnResource]. This is to help distinguish that these parameters are not set at deployment time.

- i think you misunderstand this one. The ask here is to standardize the naming of the resources. Instead of `name` under `ApplicationProps`, it should be `applicationName`. This is to standardize with our other resources i.e. lambda function name is `functionName`. Remediation here is to make breaking changes to all `name` props with `[Resource]Name`.

---

"events-method-signature" - These on methods are not associated with CloudWatch Events

The reason for the linter rule is to _reserve_ `onXxx` APIs for CloudWatch Events. This provides a CDK pattern where any API that begins with `on` is an events API. Therefore, your justification does not really move the needle for me. In this case, I see that the phrasing comes from appConfig itself. I will double check internally. If we do accept the exemption, I would still want you to specify each individual linter exemption rather than turning off the rule entirely (I didn't have time to do this when I was turning on awslint). Feel free to add that in to this PR.

---

"events-generic" - There already is a generic method. This is named on instead of onEvent for clearer naming since the action point will be passed in

The issue is the same as above -- are we ok with `onXxx` methods not being CW event methods. Is there any part of the original RFC that discusses this? And again, if we go through with the exemptions it will have to be each individual exemption.

---

- Finally, though not a blocker to stabilization, all the missing docs present a poor experience for users and should be updated where appropriate.

The first two fixes should come in separate PRs and call out their breaking changes.

chenjane-dev:
Will follow up with the first two in separate PRs. Added the bottom two individual exceptions. Yes, this was designed and approved through the RFC: https://github.com/aws/aws-cdk-rfcs/blob/main/text/0499-appconfig-constructs.md#example-4

kaizencc:
Thanks for the quick link @chenjane-dev . I will double check with @otaviomacedo (bar raiser, i believe) and @rix0rrr (linter rule creator) for a final decision

packages/@aws-cdk/aws-appconfig-alpha/lib/application.ts
```diff
@@ -296,7 +375,7 @@ export class Application extends ApplicationBase {
   /**
    * Adds the AWS AppConfig Agent as a container to the provided ECS task definition.
    *
-   * @param taskDef The ECS task definition
+   * @param taskDef The ECS task definition [disable-awslint:ref-via-interface]
```
kaizencc, line 378: I don't like this here. Why is it not in awslint.json?

kaizencc: nevermind -- i see it as part of the error message.

packages/@aws-cdk/aws-appconfig-alpha/awslint.json
```diff
@@ -26,13 +26,77 @@
...
-        "events-method-signature",
-        "events-generic"
+        "events-method-signature:@aws-cdk/aws-appconfig-alpha.Application.on",
+        "events-method-signature:@aws-cdk/aws-appconfig-alpha.Application.onDeploymentBaking",
```
kaizencc, line 34: @chenjane-dev nit: i know i told you to add each individual linter exemption, because if not, and we stabilized, then we would be exempting `events-method-signature` everywhere. However, I don't think we need the individual `params[0]` and `params[1]` either. Having the `onDeploymentBaking` linter error is enough uniqueness i think

chenjane-dev: I didn't include them at first but then was getting the errors on the `params[0]` and `params[1]`

kaizencc: ok i'll check this out for you. seems weird if true but i believe you :)

kaizencc: used `*` to get it to work

kaizencc (approved):
Sweet!

kaizencc (approved):

PR #28672, opened 2024-01-11
feat(ecs): support for capacity provider managed instance draining

otaviomacedo:
ECS now supports managed instance draining which facilitates graceful termination of Amazon ECS instances for Capacity Providers.

Add a new constructor property, `enableManagedDraining`, to `AsgCapacityProvider`, to allow users to enable this feature.

----


packages/aws-cdk-lib/aws-ecs/lib/cluster.ts
```diff
@@ -1271,6 +1286,12 @@ export class AsgCapacityProvider extends Construct {
     this.machineImageType = props.machineImageType ?? MachineImageType.AMAZON_LINUX_2;
     this.canContainersAccessInstanceRole = props.canContainersAccessInstanceRole;
     this.enableManagedTerminationProtection = props.enableManagedTerminationProtection ?? true;
+    this.enableManagedDraining = props.enableManagedDraining;
+
+    let managedDraining = undefined;
+    if (this.enableManagedDraining != undefined) {
+      managedDraining = this.enableManagedDraining ? 'ENABLED' : 'DISABLED';
+    }
```
xazhao, line 1294: nit: I wonder is there any differences when `managedDraining` is `underfined` or `DISABLED`? Can we set it to `DISABLED` first and change to `ENABLED` when `enableManagedDraining` is set?

xazhao (approved):
LGTM! Left a small comment but it's not a blocker.

PR #28678, opened 2024-01-12
chore(migrate): enable import of resources on apps created from cdk migrate

TheRealAmazonKendra:
Apps generated from cdk migrate with resources that aren't already part of a stack will (soon) create a migrate.json file. This file contains the list of resources that should be imported upon creation of the new app.

If this file is present and the source is either `localfile` or the ARN environment matches the deployment environment, running `cdk deploy` will:
1. Create a new stack and import the resources listed in migrate.json.
2. Apply outputs and CDKMetadata through a normal deployment.

Note: `localfile` is a placeholder value so that we can run integration tests on this change. Once some of the other in-progress work is finished, this will be updated.

----


packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -874,6 +874,63 @@ export class CdkToolkit {
...
+      print('%s: importing resources into stack...', chalk.bold(stack.displayName));
+
+      await this.performResourceMigration(migrateDeployment, resourcesToImport, options);
+
+      fs.rmSync('migrate.json');
+      print('%s: applying CDKMetadata and Outputs to stack (if applicable)...', chalk.bold(stack.displayName));
+    }
+  }
+
+  private async tryGetResources(migrateDeployment: ResourceImporter) {
```
comcalvi, line 900: can we swap the order of `tryGetResources` and `performResourceMigration`? `performResourceMigration` is called by the public function, so it should be above `tryGetResources`

comcalvi, line 900: can we swap the order of `tryGetResources` and `performResourceMigration`? `performResourceMigration` is called by the public function, so it should be above `tryGetResources`

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -874,6 +874,63 @@ export class CdkToolkit {
...
   }
+
+  /**
+   * Checks to see if a migrate.json file exists. If it does and the source is either `filepath` or
+   * is in the same environment as the stack deployment, a new stack is created and the resources are
+   * migrated to the stack using an IMPORT changeset. The normal deployment will resume after this is complete
+   * to add back in any outputs and the CDKMetadata.
+   */
+  private async tryMigrateResources(stacks: StackCollection, options: DeployOptions): Promise<void> {
+    const stack = stacks.stackArtifacts[0];
```
comcalvi, line 885: what about the other stacks?

comcalvi (changes requested):

packages/aws-cdk/lib/cdk-toolkit.ts
TheRealAmazonKendra: CDK migrate only creates one stack. It is not compatible with multiple stacks at this time.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -874,6 +874,63 @@ export class CdkToolkit {
...
+    const migrateDeployment = new ResourceImporter(stack, this.props.deployments);
+    const resourcesToImport = await this.tryGetResources(migrateDeployment);
+
+    if (resourcesToImport) {
+      print('%s: creating stack for resource migration...', chalk.bold(stack.displayName));
+      print('%s: importing resources into stack...', chalk.bold(stack.displayName));
+
+      await this.performResourceMigration(migrateDeployment, resourcesToImport, options);
+
+      fs.rmSync('migrate.json');
```
kaizencc, line 895: 1) i want to check that you mean to remove `migrate.json` after consuming it. 
2) why does this function "deserve" to remove this file? it seems like an unnecessary side effect of a function named `tryMigrateResources`. Is this the place to govern the lifecycle of that file?

This is a non-blocking comment. I don't have the full picture of migrate, just want to call attention to this and if there's no problem, then ignore.

TheRealAmazonKendra: The intent here was to make deployment after using cdk migrate to generate an app zero touch, or as close to it as possible. CDK migrate generates this file that contains all the import data needed. So, the attempt at an import is only made when this file is present. Since the import only needs to happen on the first deployment, it deletes it at the end so that future deployments are not of the import type.

comcalvi (approved):

kaizencc (approved):

PR #28680, opened 2024-01-12
fix(cloudformation-diff): Fix aws-sdk dependency issue

### LINKED ISSUE #28679: cloudformation-diff: Hidden dependency on aws-sdk
### Describe the bug

We have a custom aws-cdk consruct library that exports cloudformation-diff. 

Since `@aws-cdk/cloudformation-diff` for version `2.119.0`, it is throwing the following error:
```
node_modules/@aws-cdk/cloudformation-diff/lib/diff-template.d.ts:1:32 - error TS2307: Cannot find module 'aws-sdk' or its corresponding type declarations.

1 import { CloudFormation } from 'aws-sdk';
                                 ~~~~~~~~~


Found 1 error in node_modules/@aws-cdk/cloudformation-diff/lib/diff-template.d.ts:1
```

We believe the error is coming from changes in this PR: https://github.com/aws/aws-cdk/pull/28336

Specifically in `packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts`, the following change: 
```
// The SDK is only used to reference `DescribeChangeSetOutput`, so the SDK is added as a devDependency.
// The SDK should not make network calls here
// eslint-disable-next-line import/no-extraneous-dependencies
import { CloudFormation } from 'aws-sdk';
```

I have a sample repo to reproduce the exact issue: https://github.com/frankpengau/cdk-lib-cfn-diff-issue-20240112

### Expected Behavior

It should not have a missing dependency. 

### Current Behavior

It currently has an indirect dependency on `aws-sdk` causing our construct library to have a transitive dependency on `aws-sdk`. If it is indeed a dependency, it should be added in as such, instead of being only a devdependency. 

### Reproduction Steps

1. `cdk init lib --language typescript`
2. Install `@aws-cdk/cloudformation-diff` for version `2.119.0`
3. In a new file in lib folder: `export * as cfndiff from @aws-cdk/cloudformation-diff`
4. Compile typescript `tsc`

### Possible Solution

Explicit dependency on `aws-sdk`, so that it shows up in the `package-lock.json` file and installs it as necessary for `@aws-cdk/cloudformation-diff` to work. 

_No response_

### CDK CLI Version

2.114.1

### Framework Version

_No response_

### Node.js Version

18.13.0

### OS

macOS

### Language

TypeScript

### Language Version

TypeScript (5.2.2)

### Other information

_No response_

frankpengau:
Missing dependency when exporting `@aws-cdk/cloudformation-diff` in custom AWS Construct Library. 

Closes #28679

----


frankpengau:
Exemption Request

kylelaker (changes requested):
Can the underlying issue be fixed instead by changing the offending import from

```ts
import { CloudFormation } from 'aws-sdk';
```

to 

```ts
import type { CloudFormation } from 'aws-sdk';
```

The only reason the import is being is used is to get the type not to make any actual API calls. Changing the import should help make that clearer and prevent accidentally needing the `aws-sdk` at runtime.

comcalvi:
Thanks for investigating @kylelaker, I just confirmed that your fix works on the reproduction repo. @frankpengau can you please update the PR accordingly?

kylelaker (approved):
Thank you so much for this fix and for adjusting your PR based on the review feedback! This will definitely be a helpful change. Great catch!

frankpengau:
Exemption Request

frankpengau:
@kylelaker @comcalvi Is there anything else required from my end to get this change in?

comcalvi (approved):

PR #28682, opened 2024-01-12
feat(ecs): attach EBS volume to ECS task

go-to-k:
WIP.

- Releases and blogs
  - https://aws.amazon.com/jp/about-aws/whats-new/2024/01/amazon-ecs-fargate-integrate-ebs/
  - https://aws.amazon.com/jp/blogs/aws/amazon-ecs-supports-a-native-integration-with-amazon-ebs-volumes-for-data-intensive-workloads/
- Developer and User Guide
  - https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ebs-volumes.html
  - https://docs.aws.amazon.com/AmazonECS/latest/userguide/troubleshoot-ebs-volumes.html 
- CloudFormation
  - ConfiguredAtLaunch
    - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-volume.html#cfn-ecs-taskdefinition-volume-configuredatlaunch
  - VolumeConfigurations
    - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html#cfn-ecs-service-volumeconfigurations

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #28691, opened 2024-01-12
feat(ecs): L2 for ebs task attach

KollaAdithya:
To simplify ECS customer experience, customers need the ability to utilize highly available (HA) durable, high-performance, cost effective block storage for their workloads. To address these needs, ECS will provide support for Elastic Block Storage (EBS) task attachments.

This PR adds the L2 constructs for EBS task attach.

----


packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+* Represents the configuration for an ECS Service managed EBS volume.
+*/
+export interface ServiceManagedEBSVolumeConfiguration {
+  /**
+  * An IAM role that allows ECS to make calls to EBS APIs on your behalf.
+  * This role is required to create and manage the Amazon EBS volume.
+  */
+  readonly role: iam.IRole;
```
go-to-k, line 32: I think it would be user friendly to make this optional and create it internally if unspecified, what do you think?

```ts
    // example
    this.role = props.role ?? new iam.Role(this, 'Role', {
      assumedBy: new iam.ServicePrincipal('ecs.amazonaws.com'),
      managedPolicies: [
        iam.ManagedPolicy.fromAwsManagedPolicyName('service-role/AmazonECSInfrastructureRolePolicyForVolumes'),
      ],
    });
```

MrArnoldPalmer:
@go-to-k I put the wrong link in the last PR comment, sorry. Give this one a look.

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,190 @@
...
+  public readonly config?: ServiceManagedEBSVolumeConfiguration;
+
+  /**
+   * configuredAtLaunch indicates volume at launch time, referenced by taskdefinition volume.
+   */
+  public readonly configuredAtLaunch: boolean = true;
+
+  constructor(props: ServiceVolumeProps) {
+    this.name = props.name;
+    this.config = props.managedEBSVolume;
```
go-to-k, line 194: It would be good to have validation for each parameter.

https://github.com/aws/aws-cdk/pull/28682/files#diff-95d9ac7a97b34b5521d19daaf5464f6dc5de93fd10abab0add914683dd430ff7R485-R504

go-to-k: the above (link's) sample code

```ts
    if (this.sizeInGiB === undefined && this.snapshotId === undefined) {
      throw new Error('sizeInGiB or snapshotId must be specified');
    }
    if (this.throughput) {
      if (this.volumeType !== VolumeType.GP3) {
        throw new Error(`throughput can only be thisured with gp3 volume type, got ${this.volumeType}`);
      }
      if (!Token.isUnresolved(this.throughput) && this.throughput > 1000) {
        throw new Error(`throughput must be less than or equal to 1000 MiB/s, got ${this.throughput} MiB/s`);
      }
    }
    if (this.iops) {
      if ([VolumeType.SC1, VolumeType.ST1, VolumeType.STANDARD].some(type => this.volumeType === type)) {
        throw new Error(`iops cannot be specified with sc1, st1, and standard volume types, got ${this.volumeType}`);
      }
    } else {
      if ([VolumeType.IO1, VolumeType.IO2].some(type => this.volumeType === type)) {
        throw new Error(`iops must be specified with io1 or io2 volume types, got ${this.volumeType}`);
      }
    }
```

- Reference
  - size
    - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-service-servicemanagedebsvolumeconfiguration.html#cfn-ecs-service-servicemanagedebsvolumeconfiguration-sizeingib
  - iops
    - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-service-servicemanagedebsvolumeconfiguration.html#cfn-ecs-service-servicemanagedebsvolumeconfiguration-iops
  - throughput
    - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-service-servicemanagedebsvolumeconfiguration.html#cfn-ecs-service-servicemanagedebsvolumeconfiguration-throughput

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,190 @@
...
+  */
+  readonly tags?: {[key: string]: string};
+
+  /**
+  * Specifies whether to propagate the tags from the task definition or the service to the task.
+  * Valid values are: PropagatedTagSource.SERVICE, PropagatedTagSource.TASK_DEFINITION
+  *
+  * @default - undefined
+  */
+  readonly propagateTags?: PropagatedTagSource.SERVICE | PropagatedTagSource.TASK_DEFINITION;
```
go-to-k, line 146: I thought the union type doesn't work for non-TypeScript code.

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,190 @@
...
+}
+
+/**
+ *  Tag Specification for EBS volume.
+ */
+export interface EBSTagSpecification {
+  /**
+  * The tags to apply to the volume.
+  *
+  * @default none
```
go-to-k, line 136:
```suggestion
  * @default - No tags
```

go-to-k (changes requested):
I commented on a few things. If I find something else later, I may comment on it.

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
go-to-k: And the following validations may be added.

- gp3: 3,000 - 16,000 IOPS
- io1: 100 - 64,000 IOPS
- io2: 100 - 256,000 IOPS

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
go-to-k: https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#unions

MrArnoldPalmer: Yeah, this should be an enum as well. @KollaAdithya

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
KollaAdithya: That's a pretty good idea 👍

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
MrArnoldPalmer: We have a lot of this logic [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/volume.ts#L654) already. I wonder if we can abstract this out and reuse it relatively easily. @KollaAdithya @go-to-k what do you think?

go-to-k: I personally feel that it would be good to make it separately, as it would be more complicated with many different points and more conditional branches.

I am also concerned that the details are different in content from the CloudFormation reference on the ECS side. It may be that the same EBS API is used internally, but it seems appropriate for CDK to follow the CloudFormation reference here (not do anything it doesn't say), but what do you think?

- https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/volume.ts#L698C7-L698C91
- https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/volume.ts#L704-L713
- https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/volume.ts#L739
- https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/volume.ts#L759
- etc...

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -721,6 +740,46 @@ export abstract class BaseService extends Resource
...
   }
 
+  /**
+   * Adds a volume to the Service.
+   */
+  public addVolume(volume: ServiceManagedVolume) {
+    this.volumes.push(volume);
+  }
+
+  private renderVolumes(): CfnService.ServiceVolumeConfigurationProperty[] {
```
go-to-k, line 750: Since only one EBS configuration should be set, it might be a good to check here whether the volumes element is one or not.

CloudFormation returns the error `Invalid request provided: UpdateService error: The provided volume configuration is invalid. Only one volume can be configured at launch.`.

go-to-k (changes requested):

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
go-to-k: No, but still, if the same API is used internally, is it good to make it common...?

However, I'm not sure about the cost and impact of modifying the functions in the EC2 module to make it more generic for both EBS and furthermore, I'm not confident...

MrArnoldPalmer: I agree, though the docs call out that "many" of these parameters are 1:1 to the EBS. Which I assumes the validation is 1:1 as well. Hard to have confidence in that though. I'm okay with adding these validations unique to the ECS constructs.

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
KollaAdithya: added a validation check to catch this scenario over here [[link](https://github.com/aws/aws-cdk/pull/28691/files#diff-95d9ac7a97b34b5521d19daaf5464f6dc5de93fd10abab0add914683dd430ff7R747-R750)]

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
KollaAdithya: changed `propagateTags` to an enum over here [[link]](https://github.com/aws/aws-cdk/pull/28691/files#diff-c9cc7970a22bbdf1add32b7adcb50af5f8bc6212db0340715d335b9446ef31e9R170-R183)

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
KollaAdithya: made that role optional and a default role over here [[link]](https://github.com/aws/aws-cdk/pull/28691/files#diff-c9cc7970a22bbdf1add32b7adcb50af5f8bc6212db0340715d335b9446ef31e9R219-R224)

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
KollaAdithya: Added all the validations on `sizeInGiB`, `iops` and `throughput` in this method [[link]](https://github.com/aws/aws-cdk/pull/28691/files#diff-c9cc7970a22bbdf1add32b7adcb50af5f8bc6212db0340715d335b9446ef31e9R243)

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,310 @@
...
+      } else if (throughput > 1000) {
+        throw new Error(`'throughput' must be less than or equal to 1000 MiB/s, got ${throughput} MiB/s`);
+      }
+    }
+
+    // Check if IOPS is not supported for the volume type.
+    // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-service-servicemanagedebsvolumeconfiguration.html#cfn-ecs-service-servicemanagedebsvolumeconfiguration-iops
+    if ([ec2.EbsDeviceVolumeType.SC1, ec2.EbsDeviceVolumeType.ST1, ec2.EbsDeviceVolumeType.STANDARD,
+      ec2.EbsDeviceVolumeType.GP2].includes(volumeType) && iops !== undefined) {
+      throw new Error(`'iops' cannot be specified with '${volumeType}' volume type`);
```
go-to-k, line 290: With this message, the user would first specify STANDARD and encounter an error, then maybe GP2 with iops specified. In that case, they would encounter the same error again.
Therefore, I think it would be more helpful to include all four unsupported volume types in the message. (Fortunately, it's not that many words.)

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    // Check if IOPS is required but not provided.
+    if ([ec2.EbsDeviceVolumeType.IO1, ec2.EbsDeviceVolumeType.IO2].includes(volumeType) && iops === undefined) {
+      throw new Error(`'iops' must be specified with '${volumeType}' volume type`);
```
go-to-k, line 295: Here again, it might be good to include two pieces of information, io1 and io2.

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    // Validate IOPS range if specified.
+    const iopsRanges: { [key: string]: { min: number, max: number } } = {};
+    iopsRanges[ec2.EbsDeviceVolumeType.GP3]= { min: 3000, max: 16000 };
+    iopsRanges[ec2.EbsDeviceVolumeType.IO1]= { min: 100, max: 64000 };
+    iopsRanges[ec2.EbsDeviceVolumeType.IO2]= { min: 100, max: 256000 };
+    if (iops !== undefined) {
```
go-to-k, line 303:
```suggestion
    if (iops !== undefined && !Token.isUnresolved(iops)) {
```

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -721,6 +740,50 @@ export abstract class BaseService extends Resource
     this.node.defaultChild = this.resource;
   }
 
+  /**
+   * Adds a volume to the Service.
+   */
+  public addVolume(volume: ServiceManagedVolume) {
+    // only one ServiceVolumeConfiguration can be specified.
+    if (this.volumes.length >= 1) {
+      throw new Error('Invalid VolumeConfiguration. Only one volume can be configured at launch.');
```
go-to-k, line 749: It is good to count the number and occur an error in the `renderVolumes` instead of making an error the second time it is called in the `addVolume` method.
Since the `renderVolumes` method is called in a lazy method (643 line), which allows for lazy evaluation, it would be kind to indicate with `got: ${}` in the error message how many volumes were finally specified.

go-to-k (changes requested):
Thanks for the changing!

I've added a few more comments that I hope you'll take a look at.

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,310 @@
...
+   * The following are the supported volume size values for each volume type.
+   *   - gp2 and gp3: 1-16,384
+   *   - io1 and io2: 4-16,384
+   *   - st1 and sc1: 125-16,384
+   *   - standard: 1-1,024
+   *
+   * @default - The snapshot size is used for the volume size if you specify `snapshotId`,
+   * otherwise this parameter is required.
+   */
+  readonly sizeInGiB?: number;
```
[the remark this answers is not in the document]
KollaAdithya: Changed `SizeInGiB` to `Size` type in this [commit](https://github.com/aws/aws-cdk/pull/28691/commits/03c888f61acd32941a85323586451afcb084539f)

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,310 @@
...
+  * The name of the volume. This corresponds to the name provided in the ECS TaskDefinition.
+  */
+  readonly name: string;
+
+  /**
+  * Configuration for an Amazon Elastic Block Store (EBS) volume managed by ECS.
+  *
+  * @default - undefined
+  */
+  readonly managedEBSVolume?: ServiceManagedEBSVolumeConfiguration
```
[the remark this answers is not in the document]
KollaAdithya: I want to leave this interface as is right now.
So maintaining the nested structure in ServiceManagedVolumeProps offers future flexibility for supporting non-EBS volumes.

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,310 @@
...
+   * The following are the supported volume size values for each volume type.
+   *   - gp2 and gp3: 1-16,384
+   *   - io1 and io2: 4-16,384
+   *   - st1 and sc1: 125-16,384
+   *   - standard: 1-1,024
+   *
+   * @default - The snapshot size is used for the volume size if you specify `snapshotId`,
+   * otherwise this parameter is required.
+   */
+  readonly sizeInGiB?: Size;
```
go-to-k, line 74: Size type allows specification other than GiB, so

```suggestion
  readonly size?: Size;
```

packages/aws-cdk-lib/aws-ecs/lib/base/base-service.ts
```diff
@@ -721,6 +740,48 @@ export abstract class BaseService extends Resource
...
+  /**
+   * Adds a volume to the Service.
+   */
+  public addVolume(volume: ServiceManagedVolume) {
+    this.volumes.push(volume);
+  }
+
+  private renderVolumes(): CfnService.ServiceVolumeConfigurationProperty[] {
+    if (this.volumes.length > 1) {
+      throw new Error(`Invalid VolumeConfiguration. Only one volume can be configured at launch, got: ${this.volumes.length}`);
```
go-to-k, line 752:
```suggestion
      throw new Error(`Only one EBS volume can be specified for 'volumeConfigurations', got: ${this.volumes.length}`);
```

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,310 @@
...
+      } else if (!Token.isUnresolved(throughput) && throughput > 1000) {
+        throw new Error(`'throughput' must be less than or equal to 1000 MiB/s, got ${throughput} MiB/s`);
+      }
+    }
+
+    // Check if IOPS is not supported for the volume type.
+    // https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html
+    if ([ec2.EbsDeviceVolumeType.SC1, ec2.EbsDeviceVolumeType.ST1, ec2.EbsDeviceVolumeType.STANDARD,
+      ec2.EbsDeviceVolumeType.GP2].includes(volumeType) && iops !== undefined) {
+      throw new Error(`iops cannot be specified with sc1, st1, gp2 and standard volume types, got ${volumeType}`);
```
go-to-k, line 290:
```suggestion
      throw new Error(`'iops' cannot be specified with sc1, st1, gp2 and standard volume types, got ${volumeType}`);
```

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    // Check if IOPS is required but not provided.
+    if ([ec2.EbsDeviceVolumeType.IO1, ec2.EbsDeviceVolumeType.IO2].includes(volumeType) && iops === undefined) {
+      throw new Error(`iops must be specified with io1 or io2 volume types, got ${volumeType}`);
```
go-to-k, line 295:
```suggestion
      throw new Error(`'iops' must be specified with io1 or io2 volume types, got ${volumeType}`);
```

go-to-k (changes requested):
Thanks, I only commented on a minor point at the end. I will approve this when it is completed.

packages/aws-cdk-lib/aws-ecs/lib/base/service-managed-volume.ts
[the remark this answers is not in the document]
go-to-k: OK, let's keep going.

KollaAdithya:
@go-to-k and @MrArnoldPalmer  All the feedback is addressed. Can you take a look and approve it

go-to-k (approved):
Great, thanks @KollaAdithya .
I approve this.

MrArnoldPalmer (dismissed):
:shipit:

MrArnoldPalmer (approved):

PR #28707, opened 2024-01-14
fix(opensearch): always create CloudWatch Logs resource policy when logging is enabled

### LINKED ISSUE #23637: aws-opensearchservice: Domain creates new ResourcePolicy if logging enabled
### Describe the bug

Creating a new OpenSearch domain creates a new ResourcePolicy every time if any of: slowSearchLogEnabled, slowIndexLogEnabled, appLogEnabled, auditLogEnabled is true.

According to the [CloudWatch limits](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html):
```
Up to 10 CloudWatch Logs resource policies per Region per account. This quota can't be changed.
```

We're hitting this quota easily since every domain we create creates a new ResourcePolicy

### Expected Behavior

Not sure what what would be a good expected behavior.

May be: Reuse a previously created ResourcePolicy and add the new log policy statement.

### Current Behavior

A new resource policy is created with every new domain that is created with logging enabled.

### Reproduction Steps

Instantiate a new Domain and set any of the log types as true:

```typescript
                 const domain = new Domain(stack, 'TheDomain', {
                      version: EngineVersion.OPENSEARCH_1_2,
                      logging: {
                          appLogEnabled: true
                      }
                  })
```

### Possible Solution

Reuse a previously created ResourcePolicy and add the new log policy statement.

_No response_

### CDK CLI Version

1.187.0

### Framework Version

1.187.0

### Node.js Version

16.19.0

### OS

Linux

### Language

Typescript

### Language Version

4.4.3

### Other information

_No response_

sakurai-ryo:
This PR adds an option to suppress the creation of logs resource policy when logging is enabled.

### Description
Currently, a CloudWatch Logs resource policy is created by default when the Domain logging is enabled.
However, since only ten resource policies can be created per region, deploying multiple Domains may cause errors.
The `tryRemoveChild` method can be used as a workaround to delete custom resources, but a better user experience is desirable.
```ts
    const domain = new opensearch.Domain(this, 'Domain', domainProps);
    const domainResource = domain.node.defaultChild as opensearch.CfnDomain;
    domainResource.addOverride('DependsOn', undefined); // remove dependency on the custom resource

    domain.node.children
      .filter(child => child instanceof AwsCustomResource)
      .forEach(value => domain.node.tryRemoveChild(value.node.id));
```

So, I add an option to suppress the creation of resource policies.
This option allows users to reuse a broader resource policy and successfully deploy several domains.
https://docs.aws.amazon.com/opensearch-service/latest/developerguide/createdomain-configure-slow-logs.html#:~:text=Resource%22%3A%20%22cw_log_group_arn%3A*%22%7D%5D%7D%27-,Important,-CloudWatch%20Logs%20supports

Closes #23637

----


kylelaker (approved):
Thanks for this! This looks good to me.

packages/aws-cdk-lib/aws-opensearchservice/README.md
```diff
@@ -349,6 +349,36 @@ const domain = new Domain(this, 'Domain', {
 });
 ```
 
+## Suppress creating CloudWatch Logs resource policy
+
+When logging is enabled for the domain, the CloudWatch Logs resource policy is created by default.
+This resource policy is necessary for logging, but since only a maximum of 10 resource policies can be created per region,
+the maximum number of resource policies may be a problem when enabling logging for several domains.
+By setting the `suppressLogsResourcePolicy` option to true, you can suppress the creation of a CloudWatch Logs resource policy.
```
kaizencc, line 357: Can you also hint at the side-effects of adding this prop? What does it mean to not have a resource policy

kaizencc (changes requested):
Basically good to go @sakurai-ryo. Asking for a bit more clarification in the readme.

packages/aws-cdk-lib/aws-opensearchservice/README.md
sakurai-ryo: Right.
I added a description about this to readme and jsdoc.

sakurai-ryo:
@kaizencc 
Thanks for your review! I added docs.

GavinZZ (approved):
Changes LGTM. Thanks for the contribution and clear documentation!

mxthxngx:
i faced a similar issue, and had to create a lambda that makes sdk calls to update the policy. 
i wrote an article on the same:  https://dev.to/math16/solved-resource-limit-exceeded-2np6 -- incase they are trying to update an existing resource policy, as referencing an existing policy was not possible directly through cdk

PR #28712, opened 2024-01-15
fix(cloudwatch-actions): multiple alarms with LambdaAction for the same Lambda fail (under feature flag)

go-to-k:
Multiple alarms with LambdaAction for the same Lambda fail, because of the same logical IDs for generated lambda permissions.

```ts
alarmConcurrentLambdasMaximumReached.addAlarmAction(new awsCloudWatchActions.LambdaAction(lambda));
alarmRdsCpuUtilizationMaximumReached.addAlarmAction(new awsCloudWatchActions.LambdaAction(lambda));
```

```
throw new Error(`There is already a Construct with name '${childName}' in ${typeName}${name.length > 0 ? ' [' + name + ']' : ''}`);
            ^
Error: There is already a Construct with name 'AlarmPermission' in NodejsFunction [...]
```

Since the existing logical IDs would change, I implemented this with a feature flag.

----


packages/aws-cdk-lib/aws-cloudwatch-actions/test/lambda.test.ts
```diff
@@ -111,3 +112,36 @@ def handler(event, context):
   });
 });
 
+test('can create multiple alarms for the same lambda if feature flag is set', () => {
```
GavinZZ, line 115: nit: Can we add another unit test that has feature flag set to false and create multiple alarms for the same lambda? The test should except to throw an exception to guarantee backward compatibility.

packages/aws-cdk-lib/aws-cloudwatch-actions/lib/lambda.ts
```diff
@@ -21,7 +22,8 @@ export class LambdaAction implements cloudwatch.IAlarmAction {
    * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricAlarm.html
    */
   bind(_scope: Construct, _alarm: cloudwatch.IAlarm): cloudwatch.AlarmActionConfig {
```
GavinZZ, line 24: nit: In general, we add `_` to parameter to indicate it's unused. Since we're using it in `idPrefix`, it would be nice to remove the `_`. (I know it's not from you, but would be nice to make the style correct).

GavinZZ (commented):
Changes look good other than some minor feedback. Thanks for contributing and fixing this issue!

go-to-k:
@GavinZZ 

Thanks for your review. I fixed, could you please take a look them?

go-to-k:
@GavinZZ 

Thanks for your review. I fixed, could you please take a look them?

go-to-k:
@GavinZZ I tried CI again and it seems successful. could you please take a look it?

GavinZZ (approved):
LGTM, thanks for the contribution!

Issue #27164, opened 2023-09-15
Amazon.CDK.AWS.Apigatewayv2.Alpha: Support "AWS" as the IntegrationType for WebSockets


JobaDiniz:
### Describe the feature

Currently, `Amazon.CDK.AWS.Apigatewayv2.Alpha` lacks support for `AWS` option as the [IntegrationType](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-integration.html#cfn-apigatewayv2-integration-integrationtype) for **WebSocket** Apigateway.

The quote from the [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-integration.html#cfn-apigatewayv2-integration-integrationtype):
>`AWS`: for integrating the route or method request with an AWS service action, including the Lambda function-invoking action. With the Lambda function-invoking action, this is referred to as the Lambda custom integration. With any other AWS service action, this is known as AWS integration. Supported only for WebSocket APIs.

The important part is:
>With any other AWS service action, this is known as AWS integration. Supported only for WebSocket APIs.

That is, you can create a WebSocket Apigateway that calls **directly** other AWS services **without** a Lambda function middleware.

### Use Case

Ability to create a CDK for WebSocket Apigateway where I can define a direct integration to another AWS service.

For example, WebSocket route integration that triggers a AWS Step Function directly.

This use case is supported using CloudFormation and SAM as follows:
```yaml
  ApiAuthenticateRouteIntegration:
    Type: AWS::ApiGatewayV2::Integration
    Properties:
      ApiId: !Ref BrokerRobotSocketApi
      IntegrationType: AWS
      IntegrationMethod: POST
      IntegrationUri: !Sub "arn:aws:apigateway:${AWS::Region}:states:action/StartExecution"
      CredentialsArn: !Sub "${ApiAuthenticateIntegrationStateMachineExecutionRole.Arn}" 
      TemplateSelectionExpression: \$default
      RequestTemplates: # see: https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-mapping-template-reference.html
        "$default" : 
          Fn::Sub: >
            #set($statesInput=$util.escapeJavaScript($input.json('$.data')).replaceAll("\\'","'"))
            { 
[...]

PR #28718, opened 2024-01-15
feat(apigatewayv2): AWS type websocket api integration in http api

GavinZZ:
Currently, Amazon.CDK.AWS.Apigatewayv2 lacks support for AWS option as the [IntegrationType](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-integration.html#cfn-apigatewayv2-integration-integrationtype) for WebSocket Apigateway.

Added the capability that allows user to create a WebSocket Apigateway that calls directly other AWS services without a Lambda function middleware.

Closes https://github.com/aws/aws-cdk/issues/27164.

----


JobaDiniz:
I'd love to see a facilitator to configure the permission of the Websockets to invoke the service.

GavinZZ:
> I'd love to see a facilitator to configure the permission of the Websockets to invoke the service.

I'm curious how that would be possible. For lambda integration, it's clear to add a lambda permission to allow api to invoke lambda. For AWS integration, depending on the AWS service you want to integrate, the IAM role/policy varies. I couldn't think of a way to configure the permission automatically.

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/websocket/aws.ts
```diff
@@ -0,0 +1,90 @@
...
+   * @param id id of the underlying integration construct
+   */
+  constructor(id: string, private readonly props: WebSocketAwsIntegrationProps) {
+    super(id);
+  }
+
+  bind(options: WebSocketRouteIntegrationBindOptions): WebSocketRouteIntegrationConfig {
+    options;
+
+    if (!this.props.integrationUri) {
```
kaizencc, line 77: (I am not too familiar with apigwv2 integrations so legitimate question)

Why are we checking for `integrationUri` here and not making `integrationUri` required for `WebSocketIntegrationProps`?

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/websocket/aws.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/**
+ * AWS WebSocket AWS Type Integration
+ */
+export class WebSocketAwsIntegration extends WebSocketRouteIntegration {
+  /**
+   * @param id id of the underlying integration construct
+   */
+  constructor(id: string, private readonly props: WebSocketAwsIntegrationProps) {
```
kaizencc, line 70:
```suggestion
  constructor(id: string, private readonly props: WebSocketAwsIntegrationProps = {}) {
```

But only if we stick with having all properties optional

packages/aws-cdk-lib/aws-apigatewayv2/lib/websocket/integration.ts
```diff
@@ -24,7 +25,11 @@ export enum WebSocketIntegrationType {
   /**
    * Mock Integration Type
    */
-  MOCK = 'MOCK'
+  MOCK = 'MOCK',
+  /**
+   * AWS Integration Type
+   */
+  AWS = 'AWS'
```
kaizencc, line 32:
```suggestion
  AWS = 'AWS',
```

With the comma, next time, there's no need to diff the enum above

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/websocket/aws.ts
```diff
@@ -0,0 +1,90 @@
...
+import { IRole } from '../../../aws-iam';
+
+/**
+ * Props for AWS type integration for an HTTP Api.
+ */
+export interface WebSocketAwsIntegrationProps {
+  /**
+   * Integration URI.
+   *
+   * @default None.
```
kaizencc, line 16: Nit: I don't like any of these `@default None.` It doesn't tell me any information for what it means to not provide the property. Moreover, for some of these I don't believe the default truly is "none." We should be describing the default behavior better for all these props.

packages/aws-cdk-lib/aws-apigatewayv2-integrations/README.md
```diff
@@ -210,3 +211,43 @@ webSocketApi.addRoute('sendMessage', {
   integration: new WebSocketLambdaIntegration('SendMessageIntegration', messageHandler),
 });
 ```
+
+### AWS WebSocket Integration
+
+AWS integrations enable integrating with an AWS service action, including the Lambda function-invoking action. With the Lambda function-invoking action, this is referred to as the Lambda custom integration. With any other AWS service action, this is known as AWS integration. Supported only for WebSocket APIs.
```
kaizencc, line 217: This is a direct quote from [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-integration.html#cfn-apigatewayv2-integration-integrationtype). Generally, if we want to link to documentation from elsewhere thats fine. But I think we should express what the CDK construct is doing in our own words.

kaizencc (changes requested):
Nice @GavinZZ! Just a few comments. In addition, 

a) our standard is to have our PR title be mostly lowercase. see contributing [guide](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request)

b) this is mostly me, i dislike using the word 'support' in the title. Should just be the feature added, i.e. 'AWS websocket integration in HTTP API'

packages/aws-cdk-lib/aws-apigatewayv2/lib/websocket/integration.ts
GavinZZ: Is there a formatter that we can use to automatically fix the format issues like trailing comm, whitespace, and etc?

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/websocket/aws.ts
GavinZZ: That's a good call, it should just be a required property.

packages/aws-cdk-lib/aws-apigatewayv2/lib/websocket/integration.ts
kaizencc: there is `prettier` but we don't use it for some reason in cdk. i'm sure it's come up before but i don't have the actual reason why.

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/websocket/aws.ts
```diff
@@ -0,0 +1,82 @@
...
+  WebSocketRouteIntegrationConfig,
+  WebSocketRouteIntegrationBindOptions,
+} from '../../../aws-apigatewayv2';
+import { IRole } from '../../../aws-iam';
+
+/**
+ * Props for AWS type integration for an HTTP Api.
+ */
+export interface WebSocketAwsIntegrationProps {
+  /**
```
kaizencc, line 13: All of your docstrings have the formatting of the last `*/` off by one tab :(

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/websocket/aws.ts
```diff
@@ -0,0 +1,82 @@
...
+export class WebSocketAwsIntegration extends WebSocketRouteIntegration {
+  /**
+   * @param id id of the underlying integration construct
+   */
+  constructor(id: string, private readonly props: WebSocketAwsIntegrationProps) {
+    super(id);
+  }
+
+  bind(options: WebSocketRouteIntegrationBindOptions): WebSocketRouteIntegrationConfig {
+    options;
```
kaizencc, line 71: this line can't be doing anything at all right? If you don't plan on using `options` you can specify `_` before the prop name i.e. `bind(_options: WebSocketRouteIntegrationBindOptions)`. I think that's what you're trying to do here.

packages/aws-cdk-lib/aws-apigatewayv2/lib/websocket/integration.ts
```diff
@@ -45,6 +50,48 @@ export interface WebSocketIntegrationProps {
...
+  /**
+   * A map of Apache Velocity templates that are applied on the request
+   * payload.
+   *
+   * ```
+   *   { "application/json": "{ \"statusCode\": 200 }" }
+   * ```
+   *
+   * @default - No request templates required.
+  */
```
kaizencc, line 86:
```suggestion
   */
```

packages/aws-cdk-lib/aws-apigatewayv2/lib/websocket/integration.ts
```diff
@@ -45,6 +50,48 @@ export interface WebSocketIntegrationProps {
    * Integration URI.
    */
   readonly integrationUri: string;
+
+  /**
+ * Specifies the integration's HTTP method type.
+ *
+ * @default - No HTTP method required.
+ */
```
kaizencc, line 58:
```suggestion
  /**
   * Specifies the integration's HTTP method type.
   *
   * @default - No HTTP method required.
   */
```

kaizencc (changes requested):
Few more comments @GavinZZ

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/websocket/aws.ts
GavinZZ: Generally that's also my preferred way of dealing with unused parameters. But I seem to find that this is CDK preferred way of dealing with unused parameters as I repetitively see this kind of syntax and that's why I used it. 
```
  bind(options: WebSocketRouteIntegrationBindOptions): WebSocketRouteIntegrationConfig {
    options;
```

GavinZZ:
Thanks @kaizencc for the reviews. I've addressed all the feedbacks. Feel free to give it another pass at your convenience.

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/websocket/aws.ts
kaizencc: Where do you see that? Feel free to fix that, it is wrong. Definitely not a CDK preferred way.

GavinZZ: I've seen some in other files under apigatewayv2. I can fix those in a subsequent PR.

kaizencc (approved):

georgechristman:
Hi @GavinZZ and @kaizencc, I was reviewing the code this evening and I'm noticing RequestParameters is not being added to the websocket integration. After reviewing the code, the WebSocketRouteIntegrationConfig within packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/websocket/aws.ts appears to be missing the requestParameters bind. I believe if "requestParameters: this.props.requestParameters" is added to the bind, it will resolve the issue.

FayezX:
@georgechristman thanks for the find, I have added it and waiting for a review. @kaizencc https://github.com/aws/aws-cdk/pull/28859

JobaDiniz:
has this been released to some version yet?

georgechristman:
I think I was able to resolve FayezX's issue. https://github.com/aws/aws-cdk/pull/28905

PR #28728, opened 2024-01-16
feat(ec2): support for the credit configuration mode for burstable instances

### LINKED ISSUE #19166: (aws-ec2): Support configuring credit specification on NAT instance
### Description

Support configuring credit specification on NAT instance.

### Use Case

I want to deploy a NAT instance with standard credit specification. The default for t4g.nano is unlimited.

### Proposed Solution

Pass a creditSpecification property from NatInstanceProvider down to CfnInstance.

### Other information

idea

* Add a creditSpecification property to NatInstanceProps for NatInstanceProvider.
* Add a creditSpecification property to InstanceProps for Instance.
* Use the existing creditSpecification property on CfnInstanceProps for CfnInstance.

links

* https://github.com/aws/aws-cdk/blob/v2.14.0/packages/@aws-cdk/aws-ec2/lib/nat.ts#L303
* https://github.com/aws/aws-cdk/blob/v2.14.0/packages/@aws-cdk/aws-ec2/lib/instance.ts#L371

### Acknowledge

badmintoncryer:
In this PR, I have enabled the setting of the [credit configuration mode](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/burstable-credits-baseline-concepts.html#key-concepts) for EC2 instances and NAT instances in burstable performance instances.

```ts
// for EC2 instance
new Instance(stack, 'Instance', {
  vpc,
  machineImage: new AmazonLinuxImage(),
  instanceType: InstanceType.of(InstanceClass.T3, InstanceSize.LARGE),
  creditSpecification: CpuCredits.STANDARD, // added
});

// for NAT instance
const natInstanceProvider = NatProvider.instance({
  instanceType: InstanceType.of(InstanceClass.T3, InstanceSize.LARGE),
  machineImage: new AmazonLinuxImage(),
  creditSpecification: CpuCredits.STANDARD, // added
});
new Vpc(stack, 'VPC', {
  natGatewayProvider: natInstanceProvider,
});
```

Closes #19166.

----


packages/aws-cdk-lib/aws-ec2/lib/instance.ts
```diff
@@ -295,6 +296,14 @@ export interface InstanceProps {
    * @default - public IP address is automatically assigned based on default behavior
    */
   readonly associatePublicIpAddress?: boolean;
+
+  /**
+   * Specifying the CPU credit type for burstable EC2 instance types (T2, T3, T3a, etc).
+   * T3 instances with `host` tenancy do not support the unlimited CPU credit option.
```
go-to-k, line 302: This document appears to be taken from the L1 or CFn text.
The `host` here refers to the `tenancy` property of `CfnInstanceProps`, which is not visible in the L2 construct. Wouldn't it therefore be more appropriate to use the word `dedicated host`?

Otherwise, users may be get confused (like me).

packages/aws-cdk-lib/aws-ec2/lib/nat.ts
```diff
@@ -218,6 +219,14 @@ export interface NatInstanceProps {
    * @default NatTrafficDirection.INBOUND_AND_OUTBOUND
    */
   readonly defaultAllowedTraffic?: NatTrafficDirection;
+
+  /**
+   * Specifying the CPU credit type for burstable EC2 instance types (T2, T3, T3a, etc).
+   * T3 instances with `host` tenancy do not support the unlimited CPU credit option.
```
go-to-k, line 225: same as above

packages/aws-cdk-lib/aws-ec2/lib/instance.ts
```diff
@@ -370,6 +379,11 @@ export class Instance extends Resource implements IInstance {
       throw new Error('Cannot specify both of \'keyName\' and \'keyPair\'; prefer \'keyPair\'');
     }
 
+    // if credit specification is set, then the instance type must be burstable
+    if (props.creditSpecification && !props.instanceType.isBurstable()) {
+      throw new Error(`creditSpecification is not supported for ${props.instanceType.toString()} instance type`);
```
go-to-k, line 384: How about outputting which instance types are supported?

packages/aws-cdk-lib/aws-ec2/test/nat.test.ts
go-to-k, line 1: Since there are already [tests for Nat instances](https://github.com/aws/aws-cdk/blob/v2.122.0/packages/aws-cdk-lib/aws-ec2/test/vpc.test.ts#L1431), how about putting them in here instead of adding new files? If they are scattered, the contributors will not know which file to put them in when adding tests for Nat instances in the future.

(But if you have a particular intention, I will respect that.)

go-to-k (changes requested):
Thanks for this PR.
The implementation looks good. I made some comments.

packages/aws-cdk-lib/aws-ec2/test/nat.test.ts
badmintoncryer: I hadn't noticed the existing test code. Thank you for your suggestion.

packages/aws-cdk-lib/aws-ec2/lib/instance.ts
badmintoncryer: @go-to-k 
Indeed, you're right. I've updated the sentence, so please check it.

badmintoncryer:
@go-to-k Thank you for your review! I have addressed your comments.

packages/aws-cdk-lib/aws-ec2/lib/instance.ts
```diff
@@ -295,6 +296,14 @@ export interface InstanceProps {
    * @default - public IP address is automatically assigned based on default behavior
    */
   readonly associatePublicIpAddress?: boolean;
+
+  /**
+   * Specifying the CPU credit type for burstable EC2 instance types (T2, T3, T3a, etc).
+   * The unlimited CPU credit option is not supported for T3 instances with dedicated host (`host`) tenancy.
```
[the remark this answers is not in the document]
badmintoncryer: I understand the intention and agree with the proposed modification.

badmintoncryer:
@go-to-k Thank you! I have made the revisions again.

go-to-k (approved):
Thanks!

aaythapa (approved):

PR #28730, opened 2024-01-16
feat(ecs-patterns): cooldown parameter to QueueProcessingServiceBaseProps

### LINKED ISSUE #8298: Support "cooldown" parameter for QueueProcessingServiceBaseProps in aws-ecs-patterns lib
At the moment in the constructor of QueueProcessingFargateService scaling steps can be defined, but `cooldown` parameter from autoscaling cannot:
```
this.service = new ecsPatterns.QueueProcessingFargateService(this, 'Service', {
      scalingSteps: [{...}, {...}],
      ....
    });
```

### Proposed Solution

As far as I can see, the parameter should be provided here: https://github.com/aws/aws-cdk/blob/master/packages/@aws-cdk/aws-ecs-patterns/lib/base/queue-processing-service-base.ts#L270 in a similar fashion as `scalingSteps`.

### Other

* I see that the behavior can be modified by extending QueueProcessingFargateService and overriding `configureAutoscalingForService`, but I believe that if `scalingSteps` is defined in the config, `cooldown` should be defined as well.
* While reading the code of the component QueueProcessingServiceBase, I found out that it has CpuScaling defined which cannot be affected anyhow by configuration, I believe it also should be modifiable. For my use case, scaling based on the queue size is enough.

---

aaythapa:
Add an optional `Cooldown` parameter that allows users to define the grace period after a scaling activity. Default will be `undefined` to maintain compatibility. 

Closes #8298.

----


packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
```diff
@@ -295,6 +308,11 @@ export abstract class QueueProcessingServiceBase extends Construct {
    */
   public readonly scalingSteps: ScalingInterval[];
 
+  /**
+   * Grace period after scaling activity.
+   */
```
paulhcsun, line 313: Is there any reason why the additional information in the docs of the other `cooldown` property is not also added here?

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
```diff
@@ -295,6 +308,11 @@ export abstract class QueueProcessingServiceBase extends Construct {
...
+  readonly cooldown?: Duration;
```
kaizencc, line 314:
```suggestion
  private readonly cooldown?: Duration;
```

Here, we want to be explicit about scope, i.e. public or private. I tend to always default to private unless there is a clear use case for people who want to reference the prop. In this case, there is not. We can always make it public later, but we can't go back and make it private. 

And then, since its private, we don't need any docstring at all :)

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
```diff
@@ -348,6 +366,9 @@ export abstract class QueueProcessingServiceBase extends Construct {
     const defaultScalingSteps = [{ upper: 0, change: -1 }, { lower: 100, change: +1 }, { lower: 500, change: +5 }];
     this.scalingSteps = props.scalingSteps ?? defaultScalingSteps;
 
+    // Setup cooldown
```
kaizencc, line 369: probably can delete this comment :)

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
```diff
@@ -348,6 +366,9 @@ export abstract class QueueProcessingServiceBase extends Construct {
...
+    this.cooldown = props.cooldown ?? undefined;
```
kaizencc, line 370:
```suggestion
    this.cooldown = props.cooldown;
```

Will naturally turn `this.cooldown` into undefined

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
```diff
@@ -149,6 +149,19 @@ export interface QueueProcessingServiceBaseProps {
...
+   *
+   * Subsequent scale outs during the cooldown period are squashed so that only
+   * the biggest scale out happens.
+   *
+   * Subsequent scale ins during the cooldown period are ignored.
+   *
+   * @see https://docs.aws.amazon.com/autoscaling/application/APIReference/API_StepScalingPolicyConfiguration.html
+   * @default No cooldown
+   */
+  readonly cooldown?: Duration;
```
kaizencc, line 163: Are we allowed to supply any duration here? Or is there a min/max? I suggest you do some research to figure that out, and then, if there is a limitation of some sort, we should have a synth-time check for that. And then of course also a unit test.

packages/aws-cdk-lib/aws-ecs-patterns/README.md
```diff
@@ -582,6 +582,19 @@ const queueProcessingFargateService = new ecsPatterns.QueueProcessingFargateServ
...
+### Set Cooldown parameter for QueueProcessingFargateService
+
+```ts
+declare const vpc: ec2.Vpc;
+const queueProcessingFargateService = new ecsPatterns.QueueProcessingFargateService(this, 'Service', {
+  vpc,
+  memoryLimitMiB: 512,
+  image: ecs.ContainerImage.fromRegistry('test'),
+  assignPublicIp: true,
+  cooldown: Duration.seconds(500)
```
kaizencc, line 594:
```suggestion
  cooldown: Duration.seconds(500),
```

packages/aws-cdk-lib/aws-ecs-patterns/README.md
```diff
@@ -582,6 +582,19 @@ const queueProcessingFargateService = new ecsPatterns.QueueProcessingFargateServ
 });
 ```
 
+### Set Cooldown parameter for QueueProcessingFargateService
```
kaizencc, line 585: Missing one sentence explaining what the cooldown parameter is

kaizencc (changes requested):
I took the liberty to review even though it was a draft @aaythapa. Looks mostly good.

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
aaythapa: Couldn't find a hard max in the docs ([here](https://docs.aws.amazon.com/autoscaling/application/APIReference/API_StepScalingPolicyConfiguration.html#API_StepScalingPolicyConfiguration_Contents) and [here](https://docs.aws.amazon.com/autoscaling/application/userguide/step-scaling-policy-overview.html#step-scaling-cooldown)). 

I then tried to make my own scaling policy in the console and see if there's a max for the cooldown. I didn't get an explicit error but there was weird behavior when I went above 999999999 (the policy would have a different number than the one I set). Should that be our max then? 

As for min as long as its above 0 it should work

packages/aws-cdk-lib/aws-ecs-patterns/README.md
aaythapa: added sentence below, let me know if it needs changing

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
aaythapa: No reason, added it

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
paulhcsun: I think it would be reasonable for 999999999 to be the max if there wasn't any other specified value. At the very least it would prevent unexpected behaviour.

aaythapa: Cool, added an error check for the max value

packages/aws-cdk-lib/aws-ecs-patterns/README.md
```diff
@@ -582,6 +582,20 @@ const queueProcessingFargateService = new ecsPatterns.QueueProcessingFargateServ
...
```
kaizencc, line 585:
```suggestion
### Set cooldown for QueueProcessingFargateService
```

packages/aws-cdk-lib/aws-ecs-patterns/README.md
```diff
@@ -582,6 +582,20 @@ const queueProcessingFargateService = new ecsPatterns.QueueProcessingFargateServ
...
+### This will specify the amount of time, in seconds, to wait for a previous scaling activity to take effect.
```
kaizencc, line 586:
```suggestion

The cooldown period is the amount of time to wait for a previous scaling activity to take effect.
To specify something other than the default cooldown period of 300 seconds, use the `cooldown` parameter: 
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
```diff
@@ -149,6 +149,19 @@ export interface QueueProcessingServiceBaseProps {
...
+  /**
+   * Grace period after scaling activity in seconds.
+   *
+   * Subsequent scale outs during the cooldown period are squashed so that only
+   * the biggest scale out happens.
+   *
+   * Subsequent scale ins during the cooldown period are ignored.
+   *
+   * @see https://docs.aws.amazon.com/autoscaling/application/APIReference/API_StepScalingPolicyConfiguration.html
+   * @default No cooldown
```
kaizencc, line 161:
```suggestion
   * @default 300 seconds
```

https://docs.aws.amazon.com/autoscaling/application/APIReference/API_StepScalingPolicyConfiguration.html#API_StepScalingPolicyConfiguration_Contents

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
```diff
@@ -295,6 +308,19 @@ export abstract class QueueProcessingServiceBase extends Construct {
...
+   */
+  private readonly cooldown?: Duration;
```
kaizencc, line 322: docstring not needed for a private prop right? seems like overkill. if you keep it in, the default is not 'No cooldown'

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
kaizencc: This doesn't seem like it was resolved

kaizencc (changes requested):
Some more minor comments @aaythapa

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
aaythapa: Going to keep the docstring but shorten it to match the other private properties.

packages/aws-cdk-lib/aws-ecs-patterns/lib/base/queue-processing-service-base.ts
```diff
@@ -348,6 +367,11 @@ export abstract class QueueProcessingServiceBase extends Construct {
     const defaultScalingSteps = [{ upper: 0, change: -1 }, { lower: 100, change: +1 }, { lower: 500, change: +5 }];
     this.scalingSteps = props.scalingSteps ?? defaultScalingSteps;
 
+    if (props.cooldown && props.cooldown.toSeconds() > Duration.seconds(999999999).toSeconds()) {
```
kaizencc, line 370:
```suggestion
    if (props.cooldown && props.cooldown.toSeconds() > 999999999) {
```

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/queue-processing-fargate-service.test.ts
```diff
@@ -108,6 +108,88 @@ test('test fargate queue worker service construct - with only required props', (
...
+
+  expect(() => {
+    new ecsPatterns.QueueProcessingFargateService(stack, 'Service', {
+      cluster,
+      memoryLimitMiB: 512,
+      image: ecs.ContainerImage.fromRegistry('test'),
+      cooldown: cdk.Duration.seconds(1000000000),
+    });
+  }).toThrow(new Error('cooldown cannot be more than 999999999, found: 1000000000'));
+
```
kaizencc, line 190:
```suggestion
```

kaizencc (approved):

PR #28737, opened 2024-01-17
feat(lambda): deprecate `logRetention` properties in favor of `logGroup`

mrgrain:
https://github.com/aws/aws-cdk/pull/28039 introduced support for custom logging configurations for AWS Lambda Functions.

This change deprecates the `logRetention`, `logRetentionRole` and `logRetentionRetryOptions` properties in favor of using a custom logging configuration.

By default, Lambda functions send logs to an automatically created default log group named `/aws/lambda/<function name>`. However you cannot change the properties of this auto-created log group using the AWS CDK, e.g. you cannot set a different log retention. To overcome the limitation, a custom resource was introduced and configuration exposed via the `logRetention` properties. This is what we are deprecating in this change.

With the introduction of custom logging configuration and the new `logGroup` property, users can now create a fully customizable `LogGroup` ahead of time, and instruct the Lambda function to send logs to it.

Migrating from `logRetention` to `logGroup` will cause the name of the log group to change. Don't attempt to use the name of the auto-created log group, this will cause subtle issue. We recommend using auto-naming for lambda log groups, they can easily be accessed via the Lambda Console. If you want use a well-known name, we recommend using a pattern like `/<your service>/lambda/<function name>`. Be aware that a names log group can prevent a stack from being recreated without manual intervention after it has been deployed (error `Resource already exists`). This is because `LogGroups` are retained by default.

Either way, users will have to adjust and documentation will need to be updated. Any code referencing the old log group name verbatim will have to be changed as well. Keep in mind that in AWS CDK code, you can access the log group name directly from the `LogGroup` construct: 
```ts
declare const myLogGroup: logs.LogGroup;
myLogGroup.logGroupName;
```

----

### DIFF
--- a/packages/aws-cdk-lib/aws-lambda/README.md
+++ b/packages/aws-cdk-lib/aws-lambda/README.md
@@ -999,7 +999,7 @@ This way you can use the full features set supported by Amazon CloudWatch Logs.
 ```ts
 import { LogGroup } from 'aws-cdk-lib/aws-logs';
 
-const myLogGroup = new logs.LogGroup(stack, 'MyLogGroupWithLogGroupName', {
+const myLogGroup = new LogGroup(this, 'MyLogGroupWithLogGroupName', {
   logGroupName: 'customLogGroup',
 });
 


mrgrain:
Exempting tests, as this is a deprecation without changing functionality. Feat in name only to show up in change log.

rix0rrr (approved):

isker:
Question from the peanut gallery: I pulled this change in from today's release, and there are warnings coming from code that is seemingly native to the CDK, in EKS.  Should that be happening?  I can't do anything about it, as far as I can tell.  Here's the stack logged from my tests:

```
    > 95 |     const cluster = new Cluster(this, "Cluster", {
         |                     ^
      96 |       clusterName,
      97 |       version: k8sVersion,
      98 |       kubectlLayer,

      at print (node_modules/aws-cdk-lib/.warnings.jsii.js:5:146)
      at Object.aws_cdk_lib_aws_lambda_FunctionProps (node_modules/aws-cdk-lib/.warnings.jsii.js:1:1044823)
      at new Function (node_modules/aws-cdk-lib/aws-lambda/lib/function.js:1:7775)
      at Provider.createFunction (node_modules/aws-cdk-lib/custom-resources/lib/provider-framework/provider.js:1:2730)
      at new Provider (node_modules/aws-cdk-lib/custom-resources/lib/provider-framework/provider.js:1:1809)
      at new KubectlProvider (node_modules/aws-cdk-lib/aws-eks/lib/kubectl-provider.js:1:4575)
      at Cluster.defineKubectlProvider (node_modules/aws-cdk-lib/aws-eks/lib/cluster.js:1:21288)
      at new Cluster (node_modules/aws-cdk-lib/aws-eks/lib/cluster.js:1:15434)
```

I can't find where this is happening in the source of this repo. There is some magic going on here that I do not understand:
https://github.com/aws/aws-cdk/blob/4d7374ea5b83c4341935f5e5b39429b662c3857d/packages/aws-cdk-lib/aws-eks/lib/kubectl-provider.ts#L5

relm923:
Seeing similar warnings when using CustomResourceProviders with no way to resolve

mrgrain:
Thanks for calling this out folks! The internal usage should have been flagged by our tools, but wasn't. Looking into options right now.

mrgrain:
Fix is coming via https://github.com/aws/aws-cdk/pull/28783. Cluster is now perfectly quiet again. 🤫 

image

isker:
Thanks!

fiserv-plat-eng:
What happens to orphaned log groups? What is the cost implications of having both log groups from older stacks.

mrgrain:
> What happens to orphaned log groups? What is the cost implications of having both log groups from older stacks. 

You can find pricing information here:
https://aws.amazon.com/cloudwatch/pricing/

Logs are charged by data, not log groups.

dude0001:
> > What happens to orphaned log groups? What is the cost implications of having both log groups from older stacks.
> 
> You can find pricing information here: https://aws.amazon.com/cloudwatch/pricing/
> 
> Logs are charged by data, not log groups. Hope that helps.

@mrgrain I converted all Lambdas in a Stack to use `logGroup` and redeployed. The `Custom::LogRetention` custom resource that was created when using `logRetention` is still showing in the Stack Template and is assumed to be orphaned.

Expanding on the first question "What happens to orphaned log groups?" and disregarding the cost implications question:

* Are there manual or other recommended steps to clean up CloudFormation Stacks and AWS environments after refactoring your CDK to use `logGroup` instead of `logRetention` on Lambdas?

* What is the recommended way to remove the `Cutom::LogRetention` Resource from your Stack and clean up its resources? Should this be automated in CDK if this is truly orphaned?

* En lieu of an automated/native solution in CDK, is it okay to manually delete the resources `Custom::LogRetention` creates (LogRetention Lambda, IAM Roles and the CloudWatch Log Groups the LogRetention Lambda creates) and manually delete `Custom::LogRetention` from the CloudFormation Stack Template? Or is this used in some other way by CDK?

fiserv-plat-eng:
[surprised]     Brewer, Michael (US - California) reacted to your message:
________________________________

PR #28742, opened 2024-01-17
fix(appconfig): prefix names with resource name

chenjane-dev:
Linter rules caught that CDK standardizes resource name prop as `[resource]Name`. Previously this module only used `name` for the prop.

Follow up from #28671.

BREAKING CHANGE: `ApplicationProps.name` renamed to `ApplicationProps.applicationName`
  - **appconfig**: `EnvironmentProps.name` renamed to `EnvironmentProps.environmentName`
  - **appconfig**: `DeploymentStrategyProps.name` renamed to `DeploymentStrategyProps.deploymentStrategyName`
  - **appconfig**: `ExtensionProps.name` renamed to `ExtensionProps.extensionName`

----


kaizencc (approved):
Nice!

One overall thing unrelated to this PR (but I know you wrote the module too @chenjane-dev). Everywhere you have a `||` logical OR you really want a `??` nullish coalescer. The subtle difference is this:

```ts
'' ?? 'default'; // evaluates to '', because it takes the right-hand side only if the left-hand side is `null` or `undefined`
'' || 'default'; // evaluates to 'default', because it takes the right-hand side if the left-hand side is falsy
```

Now you can debate the merits for what specifying an empty string truly entails, but in some rare cases, users actually WANT an `''` (in the general sense. i'm not saying that this is a valid input for `name`). On the CDK side we try haphazardly to standardize on `??` for this reason, though we do not do a good job enforcing.

Not a problem for now, and I know you just want to get this in. But food for thought.

PR #28743, opened 2024-01-17
fix(appconfig): fromDeploymentStrategyId takes an enum-like class rather than a string

chenjane-dev:
Previously, we were typing this as a `string` and providing an enum for `PredefinedDeploymentStrategyId`s. This is a CDK anti-pattern because this makes the enum undiscoverable, since users see that it is typed only as a `string`. It also may not work in non-TS languages. Instead, we are moving the type to explicitly be an enum-like class.

Follow up from #28671.

BREAKING CHANGE: `deploymentStrategyId` prop in `fromDeploymentStrategyId` now takes a `DeploymentStrategyId` rather than a `string`. To import a predefined deployment strategy id, use `DeploymentStrategyId.CANARY_10_PERCENT_20_MINUTES`. Otherwise, use `DeploymentStrategyId.fromString('abc123')`.

----


kaizencc (approved):
Looks good. Notice along with the other PR that the `BREAKING CHANGE` syntax needs to be slightly different. It also must be the last thing in the pr description, for stupid parsing reasons.

PR #28658, opened 2024-01-10
fix(amplify): addBranch fails synth with "cannot find entry file..."

kaizencc:
Error:

```bash
Error: Cannot find entry file at /node_modules/@aws-cdk/custom-resource-handlers/dist/
aws-amplify-alpha/asset-deployment-handler/index.js
```

In #27955, we migrated the aws-amplify-alpha custom resource to `@aws-cdk/custom-resource-handlers`, our internal package for storing all custom resources. However, this migration was half-baked. The reason is because we seem to have depended on our `@aws-cdk/custom-resource-handlers` package directly. That works locally, as we have `custom-resource-handlers` readily available. 

```
|-- @aws-cdk
    |-- aws-amplify-alpha
    |-- custom-resource-handlers
```

`aws-amplify-alpha` directly went into `custom-resource-handlers` to grab the necessary entry file. This works locally, because we have access to all folders with impunity.

Of course, when packaged and published, we publish only `aws-amplify-alpha` with no other outside folders:

```
|-- @aws-cdk 
    |-- aws-amplify-alpha
```

Thus, the published module would not have access to the file under `custom-resource-handlers`, and predictably fails at synth time. 

In `aws-cdk-lib`, we airlift the necessary files _into_ the package and release it alongside `aws-cdk-lib`:

```
|-- aws-cdk-lib
    |-- aws-synthetics
    |-- custom-resource-handlers/dist/aws-synthetics/custom-resource-handler // this was airlifted into aws-cdk-lib
```

We are supposed to do the same airlift mechanism for `aws-amplify-alpha` but somehow that was forgotten. This PR adds in the necessary structure, so now `aws-amplify-alpha` looks like this:

```
|-- @aws-cdk 
    |-- aws-amplify-alpha
        |-- custom-resource-handlers/dist/aws-amplify-alpha // airlifted in via this PR
    |-- custom-resource-handlers/dist
        |-- aws-amplify-alpha
```

(please excuse my horrible ascii notation)

Fixes #28633. Fixes #28089. I tested my locally packaged `aws-amplify-alpha` on a local CDK app to confirm that the necessary structure exists in the packaged module.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

Issue #28633, opened 2024-01-09
@aws-cdk\aws-amplify-alpha: branch


PetroLiubynskyi:
### Describe the bug

The [issue #28289](https://github.com/aws/aws-cdk/issues/28289) was not resolved. Adding of **_dev_** dependency to **@aws-cdk/custom-resource-handlers** package inside **@aws-cdk/aws-amplify-alpha** without publishing @aws-cdk/custom-resource-handlers npm package won't have any effect on the described bug.

When running stack with the following code (Python) it is not possible to deploy to AWS Amplify from asset as it is described in [documentation](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_amplify_alpha/README.html#deploying-assets)

The code

```python
amplify_app = amplify.App(
            self, 
            "Amplify",
            app_name="my-app",
            platform=amplify.Platform.WEB
        )

deployed_bundle = assets.Asset(
            self,
            "DeployedBundle",
            path=bundle_path)

amplify_app.add_branch(
            "AmplifyBranch",
            branch_name=env,
            asset=deployed_bundle,
            auto_build=False,
            pull_request_preview=False
        )
```

### Expected Behavior

Amplify branch should be created from an asset

### Current Behavior

Amplify branch is not created from an asset 

The following error appears

![image](https://github.com/aws/aws-cdk/assets/124598808/5b570f98-416c-4d54-ab35-4adbe458ffc7)

### Reproduction Steps

Execute the code

```python
amplify_app = amplify.App(
            self, 
            "Amplify",
            app_name="my-app",
            platform=amplify.Platform.WEB
        )

deployed_bundle = assets.Asset(
            self,
            "DeployedBundle",
            path=bundle_path)

amplify_app.add_branch(
            "AmplifyBranch",
            branch_name=env,
            asset=deployed_bundle,
            auto_build=False,
            pull_request_preview=False
        )
```

### CDK CLI Version

aws-cdk.aws-amplify-alpha==2.118.0a0

### Framework Version

aws-cdk-lib==2.118.0

### Node.js Version

18.14.0

### OS

Windows 11 Pro

### Language

Python

### Language Version

3.10.11

PR #28760, opened 2024-01-18
fix(redshift): enableRebootForParameterChanges fails synth with "cannot find entry file…"

paulhcsun:
Error:
```
Error: Cannot find entry file at /node_modules/@aws-cdk/custom-resource-handlers/dist/
aws-redshift-alpha/asset-deployment-handler/index.js
```

This PR fixes the same issue detailed in https://github.com/aws/aws-cdk/pull/28658 but for `aws-redshift-alpha` as both modules have the same issue with accessing `custom-resource-handlers`. 

This PR uses the same airlift mechanism as `aws-cdk-lib` to move the necessary files into the `aws-redshift-alpha` package so its structure now looks like this:
```
|-- @aws-cdk 
    |-- aws-redshift-alpha
        |-- custom-resource-handlers/dist/aws-redshift-alpha // airlifted in via this PR
    |-- custom-resource-handlers/dist
        |-- aws-redshift-alpha
```

The airlift script only moves the `index.js` file and not the `*/generated.ts` files because imports in alpha module `*/generated.ts` files do not currently work since the import paths were written to only support stable modules in `aws-cdk-lib`. 

I've tested the `aws-redshift-alpha` package locally on a local CDK app and confirmed the necessary structure exists in the packaged module. The local app calls the `cluster.enableRebootForParameterChanges()` method which creates the custom resource and I was able to verify that the cluster deploys properly and is rebootable. 

Related to https://github.com/aws/aws-cdk/issues/28633 but this is the fix for `aws-redshift-alpha`

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts
+++ b/packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts
@@ -338,7 +338,7 @@ export interface LaunchTemplateProps {
   readonly keyName?: string;
 
   /**
-   * The SSK keypair to grant access to the instance.
+   * The SSH keypair to grant access to the instance.
    *
    * @default - No SSH access will be possible.
    */


packages/@aws-cdk/aws-redshift-alpha/scripts/airlift-custom-resource-handlers.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+scriptdir=$(cd $(dirname $0) && pwd)
+customresourcedir=$(node -p "path.dirname(require.resolve('@aws-cdk/custom-resource-handlers/package.json'))")
+awscdklibdir=${scriptdir}/..
+
+function airlift() {
+  # core needs to be airlifted directly to core to prevent circular dependencies
```
kaizencc, line 8: is this comment still relevant?

packages/@aws-cdk/custom-resource-handlers/README.md
```diff
@@ -15,13 +15,16 @@ and included as part of the `aws-cdk-lib` package.
...
 
+- aws-amplify-alpha/asset-deployment-handler
 - aws-amplify-alpha/asset-deployment-handler
 
 These handlers are excluded from `aws-cdk-lib/custom-resource-handlers` and are individually
 copied into their respective `-alpha` packages at build time. When an `-alpha` package is
 stabilized, part of the stabilization process **must** be to remove `-alpha` from the folder
 name, so that it is included in `aws-cdk-lib`.
 
+`*/generated.ts` files are not supported for alpha modules due to import paths that only work for stable modules in `aws-cdk-lib`.
```
kaizencc, line 26: Explain that they must be added into `custom-resources-framework/config.ts` as `ComponentType.NO_OP`

packages/@aws-cdk/custom-resource-handlers/README.md
```diff
@@ -15,13 +15,16 @@ and included as part of the `aws-cdk-lib` package.
 
 ### Experimental:
 
+- aws-amplify-alpha/asset-deployment-handler
```
kaizencc, line 18: wrong custom resource :)

kaizencc (changes requested):
Some documentation changes and then we're all good. Thanks @paulhcsun

Issue #26982, opened 2023-09-01
(aws-ec2): Support scope-local cdk context caching of images


luxaritas:
### Describe the feature

Many of the MachineImage constructs support the cacheInContext parameter to cache the AMI ID in the cdk.context.json. A limitation of this is that then any usage of the image across the entire app will use the same AMI, even though it may be desirable to use the latest image at the time of resource creation.

I would like to be able to specify the CDK context key to be tied to the scope the MachineImage is used in.

### Use Case

I have a number of EC2 instances that I treat as persistent/stateful, namely which host third-party software which expect to be run completely self-contained in a Docker container which maintains its own volumes for databases, file uploads, etc. in addition to maintaining its own update lifecycle through update scripts. Whenever I deploy one of these new stateful instances in my CDK app, I want it to deploy with the latest Amazon Linux API, without the instance being recreated on future deployments.

### Proposed Solution

I currently use the following workaround:
```ts
class CustomImage implements IMachineImage {
  getImage(scope: Construct): MachineImageConfig {
    const parameterName = AmazonLinuxImage.ssmParameterName({
      generation: AmazonLinuxGeneration.AMAZON_LINUX_2,
      edition: AmazonLinuxEdition.STANDARD,
    });
    const imageId = ContextProvider.getValue(scope, {
      provider: ContextProviderOpt.SSM_PARAMETER_PROVIDER,
      props: { parameterName, scope: scope.node.path },
      dummyValue: `dummy-value-for-${scope.node.path}-${parameterName}`,
    }).value;
    return {
      imageId,
      osType: OperatingSystemType.LINUX,
      userData: UserData.forLinux(),
    }
  }
}
```
My anticipated solution would be to support an additional parameter to `ssm.StringParameter.valueFromLookup` called something like `forScope` which provides the scope as an extra prop to `ContextProvider.getValue`, and then `MachineImage` props would either allow `cachedInContext` to support multiple options (say "global" or "scope"), or a new property (eg cachedInContextForScope). This additional forScope option can be useful in general for other string parameters and other "lookup"-type situations as well.

### Other Information

Relevant stale issue: https://github.com/aws/aws-cdk/issues/12355

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.93.0

### Environment details (OS name and version, etc.)

N/A

khushail:
[...]

PR #28766, opened 2024-01-19
feat(core): support adding additional cache key to cdk context values

luxaritas:
The ContextProvider mechanism and various "lookup" functions of a number of constructs support caching resolved values in the cdk.context.json. The context keys are constructed from the parameters of the lookup, which for lookup functions means whenever a resource with the same parameters is resolved, it is resolved as the same value across the entire app. However when a value may change over time, the user may wish to use the latest value when creating creating a new reference to the construct, effectively tying the cached context value to the scope - this patch enables this.

The primary use case is looking up an AMI parameter for a "stateful" EC2 instance. Currently if you specify cachedInContext, any future images created would use the same cached AMI, and updating the value would require updating all usages of the image across the entire app.

Closes https://github.com/aws/aws-cdk/issues/26982

----


luxaritas:
This PR continues from #26984, which was closed as stale while waiting for feedback from a maintainer

TheRealAmazonKendra:
Sorry for the delays. I took a look through the linked PR that was closed and I'm not totally sure I understand @rix0rrr's last comment so I'm going to sync with him before giving this a thorough review. Just wanted to give you a heads up that this is on my radar since I know you've been waiting a while for a new review. 

Thank you for all the work on this! I'll have some real comments either tomorrow or Monday.

luxaritas:
Oops didn't mean to merge in latest, but it looks like everything is good there anyways!

TheRealAmazonKendra:
These two PRs (https://github.com/aws/aws-cdk/pull/29372) and (https://github.com/aws/aws-cdk/pull/28766) are not doing the same thing, but there is definitely overlap. Given that fact, it changes the feedback I was going to give to both a bit.

I absolutely see the value in both these changes but want to make sure we're getting the contract right across both. I know you've both been waiting for feedback for some time but I'm going to tax your patience a tad further so I can discuss with the team tomorrow.

packages/aws-cdk-lib/aws-ec2/README.md
```diff
@@ -810,7 +814,13 @@ If the security group ID is known and configuration details are unknown, use met
...
 ```
 
-The result of `SecurityGroup.fromLookupByName` and `SecurityGroup.fromLookupById` operations will be written to a file called `cdk.context.json`. You must commit this file to source control so that the lookup values are available in non-privileged environments such as CI build steps, and to ensure your template builds are repeatable.
+The result of `SecurityGroup.fromLookupByName` and `SecurityGroup.fromLookupById` operations will be
+written to a file called `cdk.context.json`. 
+You must commit this file to source control so that the lookup values are available in non-privileged
+environments such as CI build steps, and to ensure your template builds are repeatable.
+
+To customize the cache key use the `additionalCacheKey` parameter. 
+This can be useful if you want to scope the context variable to a construct (eg, using `additionalCacheKey: this.node.path`).
```
comcalvi, line 823: Why would a user want to add the scope to the context variable? Why is this meaningfully different from creating multiple keys prefixed with the extra string?

comcalvi: Why is this better than creating multiple context keys? Is it easier to track, easier to use, does it give a way to group related values together?

luxaritas: Ah. If I understand your question correctly now, the upshot is that you otherwise can't! The context keys are determined by the construct (eg, when you call .fromLookup, it creates the key based on the lookup parameters), so this provides an escape hatch to customize them

comcalvi: I'm not understanding the new capability you're proposing. Let's analyze an example. Today, if I want two values from lookup, I need two parameters, like this:

```
ssm.StringParameter.valueFromLookup(this, 'Parameter1');
ssm.StringParameter.valueFromLookup(this, 'Parameter2');
```

With your change, I have this now:

```
ssm.StringParameter.valueFromLookup(this, 'Parameter', 'extraKey1');
ssm.StringParameter.valueFromLookup(this, 'Parameter', 'extraKey2');
```

What does this do?

luxaritas: This allows you to use the _same parameter_, but _separate the cache lifetimes_. The example I gave is with getting the latest Amazon Linux image - the name of the parameter is constant (AWS controls it!), and when spinning up new stateful instances I want the cache to include the value of the parameter at the time the instance was launched (which will be different for each instance).

comcalvi: Ah! Now this is making sense, thank you for patiently explaining this. So when you pass the construct tree path (`this.node.path`) as the extra cache key and synthesize, the result of the parameter lookup is stored under `Parameter` as the top-level key, then `/path/to/my/instance` as the cache key. As you add more instances, you're adding another entry under `Parameter` to be `path/to/my/instance2`.

These all hold the same value. When the parameter changes, those previous context entries continue to retain that value. When you create new instances after the parameter changes though, you get another entry under `Parameter`, this time it's `path/to/my/instance3`, but it has a different value than the first two.

This is a really cool feature! Please update the READMEs to explain some of these details about the cache lifetimes. Also please give an example walkthrough, where you detail the value of a parameter, then show the context after doing a lookup with a certain lifetime key, and then show what happens when the parameter changes and you perform another lookup. This will help users understand how this feature works.

luxaritas: Strictly speaking the path and the parameter name are concatenated to get the cache key, but you get it 🙂

I agree I should work on expanding the documentation, though I need to figure out how to balance descriptiveness and conciseness (I got what I interpreted as conflicting feedback on this previously)

comcalvi: Please lean on the side of overly descriptive initially and then we can trim it down later, as needed.

PR #28770, opened 2024-01-19
chore: run integration tests in javascript only

GavinZZ:
Current situation:
yarn integ allows to run a TS integ file
CI doesn’t allow it, which caused confusing issues.
What we want instead:
Both local integ-runner and CI don’t allow it, so that this issue won’t happen again. To do this, we need to update integ-runner to disable running TS locally.

Closes #<issue number here>.

----


paulhcsun:
Based on the slack discussion this solution makes sense to me but I'd appreciate a second opinion on it. @mrgrain @kaizencc could one of you have a look as well since you had some input on the solution?

mrgrain (approved):

PR #28771, opened 2024-01-19
feat(stepfunctions): support Map ItemSelector

### LINKED ISSUE #23265: (aws_stepfunctions): Depreciated Parameters field in Map
### Describe the feature

As per https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#map-state-inline-deprecated-fields, `Parameters` filed is now depreciated. `ItemSelector` field should be used instead (https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#itemselector)

### Use Case

Depreciated fields may break compatibility

### Proposed Solution

As new `ItemSelector`  filed takes input in exactly the same form as depreciated 'Parameters', the easiest solution would be to rename field and create "alias" with depreciation warning durning synth.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.53.0

MacOS 13.0.1

abdelnn:
Adds support for Map's [ItemSelector](https://docs.aws.amazon.com/step-functions/latest/dg/input-output-itemselector.html) field and deprecates [parameters](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#map-state-inline-deprecated-fields).

With the release of Distributed Maps, AWS StepFunctions has released a few new fields, and deprecated two old ones for Map states. One of them was `Iterator` which was replaced with `ItemProcessor`, the other one is `Parameters` (replaced by `ItemSelector`). A similar PR was made to deprecate `Iterator` (#27913). While they are deprecated, these fields are still supported (PR reflects that)

`ItemProcessor` and `ItemSelector` are both fields that are supported in `DISTRIBUTED` and `INLINE` mode, hence why they were added to the existing Map construct

Closes #23265

----


packages/aws-cdk-lib/aws-stepfunctions/lib/states/map.ts
```diff
@@ -63,12 +63,28 @@ export interface MapProps {
...
    */
   readonly parameters?: { [key: string]: any };
 
+  /**
+   * The JSON that you want to override your default iteration input (mutually exclusive  with `parameters`).
+   *
+   * @see
+   * https://docs.aws.amazon.com/step-functions/latest/dg/input-output-itemselector.html
+   *
+   * @default $
```
scanlonp, line 84: Does $ have a significance as a default? Can this docstring be descriptive?

packages/aws-cdk-lib/aws-stepfunctions/test/map.test.ts
```diff
@@ -209,6 +255,23 @@ describe('Map State', () => {
     app.synth();
   }),
 
+  test('synth is successful with item selector', () => {
```
scanlonp, line 258: This test is covered with the integ test, but is fine.

scanlonp (changes requested):
A couple nits + you should use assertions in the integ test. Otherwise looks good to me!

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map.ts
abdelnn: `$` represents the entire state input in Step Functions. Because this is essentially a replacement of `parameters`, I kept the docstring the same as what it was for `parameters`.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions/test/integ.map-itemselector.ts
```diff
@@ -0,0 +1,37 @@
+import { IntegTest } from '@aws-cdk/integ-tests-alpha';
+import * as cdk from 'aws-cdk-lib';
+import * as sfn from 'aws-cdk-lib/aws-stepfunctions';
+
+/**
+ * Stack verification steps:
+ *
+ * -- aws stepfunctions describe-state-machine --state-machine-arn <stack-output> has a status of `ACTIVE`
+ */
```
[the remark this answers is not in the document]
abdelnn: Thanks for the callout! Should be added now

scanlonp (approved):
Good addition! Thanks!

scanlonp (approved):

PR #28082, opened 2023-11-21
fix(stepfunctions-tasks): mwaa service generates wrong action in role policy

tam0ri:
When we use CallAwsService for Step Functions task, CDK generates IAM policy to grant permission regarding the API call. However, if we specify `mwaa` as service in CallAwsService, CDK generates wrong policy statement such as `mwaa:listEnvironments`. Correct service prefix for MWAA is `airflow`.

https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonmanagedworkflowsforapacheairflow.html
> Amazon Managed Workflows for Apache Airflow (service prefix: airflow) provides the following service-specific resources, actions, and condition context keys for use in IAM permission policies.

This PR solves the issue by adding `mwaa` into iamServiceMap. This is similar with #27623.

Closes #28081

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #28775, opened 2024-01-19
fix(stepfunctions-tasks): mediapackagevod service generates wrong action in role policy

### LINKED ISSUE #28774: stepfunctions-tasks: mediapackagevod service generates wrong action in role policy
### Describe the bug

Following code is properly creates a step in Step Functions

```typescript
new sfn_tasks.CallAwsService(stack, 'DeleteMediaPackageVoDAsset', {
  service: 'mediapackagevod',
  action: 'deleteAsset',
  resultPath: sfn.JsonPath.DISCARD,
  iamResources: ['*'],
});
```
However, it creates wrong statement in the role policy.
```json
        {
            "Action": "mediapackagevod:deleteAsset",
            "Resource": "*",
            "Effect": "Allow"
        }
```

Correct service prefix for `MediaPackageVoD` is `mediapackage-vod`.
https://docs.aws.amazon.com/service-authorization/latest/reference/list_awselementalmediapackagevod.html

This issue is similar with https://github.com/aws/aws-cdk/issues/27573 and https://github.com/aws/aws-cdk/issues/28081.

### Expected Behavior

Action should be mediapackage-vod:deleteAsset

### Current Behavior

Action is mediapackagevod:deleteAsset

### Reproduction Steps

Sample code for reproduction is below:

```typescript
import {
  aws_stepfunctions as sfn,
  aws_stepfunctions_tasks as sfn_tasks,
} from 'aws-cdk-lib';

export class SfnMediaPackageVoDStack extends cdk.Stack {
    super(scope, id, props);

    const deleteAsset = new sfn_tasks.CallAwsService(stack, 'DeleteMediaPackageVoDAsset', {
      service: 'mediapackagevod',
      action: 'deleteAsset',
      resultPath: sfn.JsonPath.DISCARD,
      iamResources: ['*'],
    });
  
    new sfn.StateMachine(this, 'SimpleStateMachine', {
      definitionBody: sfn.DefinitionBody.fromChainable(deleteAsset)
    });
  }
}
```

### Possible Solution

[aws-cdk/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/aws-sdk/call-aws-service.ts](https://github.com/aws/aws-cdk/blob/25ee8ef61d6b33628923a0fee4c042a0f0f2441e/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/aws-sdk/call-aws-service.ts#L92-L96)

_No response_

### CDK CLI Version

2.122.0

### Framework Version

_No response_

### Node.js Version

v18.19.0

### OS

MacOS

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

orekav:
When we use CallAwsService for Step Functions task, CDK generates IAM policy to grant permission regarding the API call. However, if we specify `mediapackagevod` as service in CallAwsService, CDK generates wrong policy statement such as `mediapackagevod:deleteAsset`. Correct service prefix for MediaPackageVOD is `mediapackage-vod`.

https://docs.aws.amazon.com/service-authorization/latest/reference/list_awselementalmediapackagevod.html

> Amazon MediaPackageVOD (service prefix: mediapackage-vod) provides the following service-specific resources, actions, and condition context keys for use in IAM permission policies.

This PR solves the issue by adding mediapackagevod into iamServiceMap. This is similar with https://github.com/aws/aws-cdk/pull/27623 and https://github.com/aws/aws-cdk/pull/28082.

Closes #28774.

----


lpizzinidev (approved):
Thanks 👍

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/aws-sdk/call-aws-service.ts
aaythapa, line 1: Can you add a small NOTE at the end of the [AWS SDK section in the README](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-stepfunctions-tasks/README.md#aws-sdk) stating that we use the service name provided in the policy template unless the prefix is mapped in this file (there is probably a cleaner way to state this). This would save some debugging time in the future if someone encounters a similar error.

aaythapa (commented):
Thank you for the contribution! Everything looks good, just some small documentation nitpicks

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/aws-sdk/call-aws-service.ts
```diff
@@ -92,6 +92,7 @@ export class CallAwsService extends sfn.TaskStateBase {
     const iamServiceMap: Record<string, string> = {
       sfn: 'states',
       cloudwatchlogs: 'logs',
+      mediapackagevod: 'mediapackage-vod',
       mwaa: 'airflow',
     };
```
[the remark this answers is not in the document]
orekav: It is already in the PR/Issue with the link to `https://docs.aws.amazon.com/service-authorization/latest/reference/list_awselementalmediapackagevod.html`

First paragraph!

```
AWS Elemental MediaPackage VOD (service prefix: mediapackage-vod) provides the following service-specific resources, actions, and condition context keys for use in IAM permission policies.
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/aws-sdk/call-aws-service.ts
orekav: The purpose of this PR is to fix the issue, it is not meant to be a `chore` which is the case of your comment for the cases are in that map.

Furthermore, I think the AWS Call Service has more issue, examples:
- S3 `ListObjects` is a valid action but that will create the wrong `IAM policy`
  - Even worse, it doesn't do the permission for the resource `root` + `root/*`
- Dynamo Query, it doesn't give you access in case of using an `index`.

Creating the permission for the `iamAction` it is a "best try attempt"
That's why there is a property call `iamAction` with the following statement

Screenshot 2024-01-27 at 18 35 08

So, once again, if anyone wants to point that related to the documentation, an issue to improve it should be added.

aaythapa (approved):

aaythapa (approved):

PR #28783, opened 2024-01-19
feat(core): custom resources deprecate `logRetention` in favor of `logGroup`

mrgrain:
In #28737 we have deprecated `logRetention` in favor of `logGroup`. Some custom resources have made the `logRetention` property part of their own API and are now emitting deprecation warnings with no way forward to resolve them. So we are now also deprecating `logRetention` for any custom resources.

Migrating log groups for custom resource would follow the same steps as outline in #28737. Given that custom resource logging is for debugging purposes and there are no guarantees about the output format, it should be possible to simply replace `logRetention` with a simple `logGroup` in most cases:

```ts
const awsCustom1 = new cr.AwsCustomResource(this, 'API1', {
  // Replace this
  // with
  logGroup: new logs.LogGroup(this, 'AwsCustomResourceLogs', {
    retention: logs.RetentionDays.ONE_WEEK,
  }),
});
```
Fixes https://github.com/aws/aws-cdk/issues/28806
Fixes https://github.com/aws/aws-cdk/issues/28809
Related to https://github.com/aws/aws-cdk/pull/28737

----


packages/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.ts
```diff
@@ -336,7 +344,10 @@ export class BucketDeployment extends Construct {
         accessPoint,
         mountPath,
       ) : undefined,
-      logRetention: props.logRetention,
+      // props.logRetention is deprecated, make sure we only set it if it is actually provided
+      // otherwise jsii will print warnings even for users that don't use this directly
+      ...(props.logRetention ? { logRetention: props.logRetention } : {}),
+      logGroup: props.logGroup,
```
kaizencc, line 350: what happens if both `logRetention` and `logGroup` are added?

packages/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.ts
```diff
@@ -112,9 +112,17 @@ export interface BucketDeploymentProps {
    * The number of days that the lambda function's log events are kept in CloudWatch Logs.
    *
    * @default logs.RetentionDays.INFINITE
+   * @deprecated Use logGroup for full control over the custom resource log group
```
kaizencc, line 115: ^ does this mean that the default behavior is changed also?

packages/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.ts
```diff
@@ -112,9 +112,17 @@ export interface BucketDeploymentProps {
...
    */
   readonly logRetention?: logs.RetentionDays;
 
+  /**
+   * The Log Group used for logging of events emitted by the custom resource's lambda function.
+   *
+   * @default - a default log group created by AWS Lambda
```
kaizencc, line 122: if `logRetention` defalt was infinite retention, is that true here also?

kaizencc (commented):
I am also surprised by lack of tests -- is the behavior already being tested elsewhere?

packages/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.ts
mrgrain: Yes, the default behavior was to not do anything the log group that the custom resource lambda creates

packages/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.ts
mrgrain: No, the default behavior was to not do anything to the log group that the custom resource lambda creates.

packages/aws-cdk-lib/aws-s3-deployment/lib/bucket-deployment.ts
mrgrain: Throws an error in the Function construct.

mrgrain:
> I am also surprised by lack of tests -- is the behavior already being tested elsewhere?

Yes `Function` has extensive tests. The intermediates (`AwsCustomResourceSingletonFunction` and `BucketDeploymentSingletonFunction`) are codegen'd.

But you're right. I need to update a few tests.

mrgrain:
@kaizencc Added tests. Hope that satisfies your standards 😅

kaizencc (approved):
Yay :)

saurabhmarwaha:
@mrgrain Any solution or workaround to this yet? I am still getting this error. Specific issue that I am getting is given in https://github.com/aws/aws-cdk/issues/28809

I didn't understand what to do based on

[WARNING] aws-cdk-lib.aws_lambda.FunctionOptions#logRetention is deprecated.
instead create a fully customizable log group with logs.LogGroup and use the logGroup property to instruct the Lambda function to send logs to it.
Migrating from logRetention to logGroup will cause the name of the log group to change.
Users and code and referencing the name verbatim will have to adjust.

In AWS CDK code, you can access the log group name directly from the LogGroup construct:

declare const myLogGroup: logs.LogGroup;
myLogGroup.logGroupName;

mrgrain:
@saurabhmarwaha This was originally fixed in [v2.123.0](https://github.com/aws/aws-cdk/releases/tag/v2.123.0) and the deprecation was later reverted in [v2.125.0](https://github.com/aws/aws-cdk/releases/tag/v2.125.0). Please update your version of `aws-cdk-lib`.

l3xcaliburr:
> @saurabhmarwaha This was originally fixed in [v2.123.0](https://github.com/aws/aws-cdk/releases/tag/v2.123.0) and the deprecation was later reverted in [v2.125.0](https://github.com/aws/aws-cdk/releases/tag/v2.125.0). Please update your version of `aws-cdk-lib`.

Just to clarify - you are saying that if we use v2.123.0 the jsii warnings will not be present?

jay-herrera:
Just wanted to say that this is still printing out warnings for me on version 2.220.0

speedhawk21:
Still getting the below warnings as well with the latest.

`    "aws-cdk-lib": "^2.221.1",
`

```
11:36:21 AM [WARNING] aws-cdk-lib.aws_lambda.FunctionOptions#logRetention is deprecated.
11:36:21 AM   use `logGroup` instead
11:36:21 AM   This API will be removed in the next major release.
```

PR #28790, opened 2024-01-20
feat(s3): date-based partitioning for log objects

### LINKED ISSUE #28141: (aws-s3): server access logging support automatic date-based partitioning
### Describe the feature

S3 Bucket Logging Configuration should support automatic date-based partitioning for delivered logs.

### Use Case

See: https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-s3-server-access-logging-date-partitioning/

### Proposed Solution

Probably need to introduce a new property and/or enum class to allow setting the [TargetObjectKeyFormat](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-targetobjectkeyformat.html)

### Other Information

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-partitionedprefix.html
https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerLogs.html

### Acknowledgements

### CDK version used

any

any

sakurai-ryo:
S3 server access logging newly supports date-based partitioning.
This feature allows key formats of log objects to be partitioned by date, which is useful when querying from Athena.

In CloudFormation, `TargetObjectKeyFormat` was added.
This object has the `PartitionedPrefix` and the `SimplePrefix` property, only one of which is allowed.
The `SimplePrefix` is the key format that originally existed.
The `PartitionedPrefix` is the new key format and allows the user to select either the event time or delivery time as the date source.

```yaml
Resources:
  S3BucketServerAccessLogEnabledTest:
    Type: AWS::S3::Bucket
    Properties:
      BucketName: access-log-test
      LoggingConfiguration:
        DestinationBucketName: myDestinationBucket
        LogFilePrefix: 'hello'
        TargetObjectKeyFormat:
          # You can deploy in either of the following formats
          SimplePrefix: {} # 1
          PartitionedPrefix: {} # 2
          PartitionedPrefix: # 3
            PartitionDateSource: EventTime # | DeliveryTime
```

whats-new
https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-s3-server-access-logging-date-partitioning

CloudFormation User Guides
https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-loggingconfiguration.html#cfn-s3-bucket-loggingconfiguration-targetobjectkeyformat

S3 User Guides
https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerLogs.html#server-access-logging-overview

Closes #28141

----


packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1544,6 +1605,13 @@ export interface BucketProps {
    */
   readonly serverAccessLogsPrefix?: string;
 
+  /**
+   * Optional key format for log objects.
+   *
+   * @default - the default key format is: [DestinationPrefix][YYYY]-[MM]-[DD]-[hh]-[mm]-[ss]-[UniqueString]
+   */
+  readonly logObjectKeyFormat?: LogObjectKeyFormat;
+
```
aaythapa, line 1614: Would calling it `targetObjectKeyFormat` be better? Matches with the CFN property name and its essentially a pass through already.

aaythapa (commented):
Thank you for your contribution, looks great overall! Just one question.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
sakurai-ryo: Thanks!
Come to think of this, you are right.
Renamed the property and the enum-like class name.

I was thinking about naming the `log` prefix to the `targetObjectKeyFormat` because the log settings are flat in the `BucketProps` interface and it would be confusing without the `log` prefix.

sakurai-ryo:
Thanks for your review @aaythapa !
I renamed the property.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1336,6 +1336,67 @@ export interface IntelligentTieringConfiguration {
...
+ * The key format for the log object.
+ */
+export abstract class TargetObjectKeyFormat {
+  /**
+   * Use partitioned prefix for log objects.
+   *
+   * The partitioned prefix format as follow:
+   * [DestinationPrefix][SourceAccountId]/​[SourceRegion]/​[SourceBucket]/​[YYYY]/​[MM]/​[DD]/​[YYYY]-[MM]-[DD]-[hh]-[mm]-[ss]-[UniqueString]
+   */
+  public static partitionedPrefix(dateSource?: PartitionDateSource): TargetObjectKeyFormat {
```
aaythapa, line 1364: is the `dataSource` parameter optional here? whats the behaviour if they do 
```
targetObjectKeyFormat: s3.TargetObjectKeyFormat.partitionedPrefix()
```

sakurai-ryo: @aaythapa 
Thanks for clarifying.
If the `dataSource` argument is not specified, `EventTime` is used as a default date source.

> PartitionedPrefix defaults to EventTime delivery when server access logs are delivered.
> https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-partitionedprefix.html

Is it better to explicitly set `EventTime` for `CfnBucket` if no argument is specified?

aaythapa: @sakurai-ryo I think it'll depend. If the user doesn't set `dataSource` then the CFN template will look like the following (let me know if I'm wrong)
```
TargetObjectKeyFormat:
    PartitionedPrefix:
       PartitionDateSource: undefined
```
Will this still deploy and set the delivery to `EventTime`? Or is there a deployment error? If there is a deployment error then I think we should explicitly set it, if not I think it's fine to leave it as is.

Either way we should state somewhere (whether in the doc-string or the readme) that `PartitionedPrefix` defaults to `EventTime` delivery

sakurai-ryo: @aaythapa 
Thanks for digging into this and sorry for my lack of explanation.
In this case, the template would look like this.
I think this is because the undefined property disappears when stringifying from JS to JSON.
So the deployment will succeed.
I added integ-test to verify this.

```json
  "MyBucket43E0A113B": {
   "Type": "AWS::S3::Bucket",
   "Properties": {
    "LoggingConfiguration": {
     "DestinationBucketName": {
      "Ref": "MyAccessLogsBucketF7FE6635"
     },
     "LogFilePrefix": "example4",
     "TargetObjectKeyFormat": {
      "PartitionedPrefix": {}
     }
    }
   },
   "UpdateReplacePolicy": "Delete",
   "DeletionPolicy": "Delete"
  }
```

> Either way we should state somewhere (whether in the doc-string or the readme) that PartitionedPrefix defaults to EventTime delivery

Absolutely right.
I added the comment in doc-string.

aaythapa: Awesome, thank you

aaythapa (approved):
🚀🚀

PR #27138, opened 2023-09-14
feat(rds): support CA certificate for cluster instances

lpizzinidev:
Exposes the `caCertificate` property for an RDS cluster instance to allow specifying a custom CA identifier using the `CaCertificate` class.

Usage:
```
new DatabaseCluster(this, 'Database', {
  engine: rds.DatabaseClusterEngine.auroraMysql({ version: rds.AuroraMysqlEngineVersion.VER_3_01_0 }),
  writer: rds.ClusterInstance.provisioned('writer', {
    caCertificate: rds.CaCertificate.RDS_CA_RDS2048_G1,
  }),
  readers: [
    rds.ClusterInstance.serverlessV2('reader', {
      caCertificate: rds.CaCertificate.of('custom-ca'),
    }),
  ],
  vpc,
});
```

Closes #26865.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #28791, opened 2024-01-20
feat(docdb): support CA certificate for cluster instances

### LINKED ISSUE #28356: (documentDB): (Expose CACertificateIdentifier in L2 Construct)
### Describe the feature

Pls expose the CloudFormation CACertificateIdentifier property on the L2 construct `docDb.DatabaseCluster`

### Use Case

With that one does not have to use `Aspects` to set the ca certificate

### Proposed Solution

If one can guide me I might be able to submit a PR myself.

### Other Information

Right now one can use `Aspects` as a workaround:

```ts
import * as docDb from 'aws-cdk-lib/aws-docdb';

let database: docDb.DatabaseCluster;

// ... 

cdk.Aspects.of(database).add({
    visit: (node) => {
        if (node instanceof docDb.CfnDBInstance) {
            node.addPropertyOverride('CACertificateIdentifier', 'rds-ca-rsa2048-g1');
        }
    }
});
```

### Acknowledgements

### CDK version used

2.100.0

macOS 14.2

brokad:
Exposes the [CaCertificateIdentifier](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-docdb-dbinstance.html#cfn-docdb-dbinstance-cacertificateidentifier) property of [AWS::DocDB::DBInstance](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-docdb-dbinstance.html) in the L2 constructs `DatabaseCluster` and `DatabaseInstance` of `aws_docdb`. This allows specifying a custom CA identifier using the CaCertificate class.

Usage with `DatabaseCluster`:

```typescript
new DatabaseCluster(stack, 'Database', {
  // ...
  instanceType: InstanceType.of(InstanceClass.R5, InstanceSize.LARGE),
  instanceCaCertificate: CaCertificate.RDS_CA_RSA4096_G1,
  // ...
});
```

Usage with `DatabaseInstance`:

```typescript
new DatabaseInstance(stack, 'Instance', {
  cluster: databaseCluster,
  instanceType: InstanceType.of(InstanceClass.R5, InstanceSize.LARGE),
  caCertificate: CaCertificate.RDS_CA_RSA4096_G1,
});
```

This is modelled on #27138.

Closes #28356.

----


packages/aws-cdk-lib/aws-docdb/lib/cluster.ts
```diff
@@ -88,6 +89,17 @@ export interface DatabaseClusterProps {
...
+  /**
+   * The identifier of the CA certificate used for the instances.
+   *
+   * Specifying or updating this property triggers a reboot.
+   *
+   * @see https://docs.aws.amazon.com/documentdb/latest/developerguide/ca_cert_rotation.html
+   *
+   * @default - DocumentDB will choose a certificate authority
+   */
+  readonly instanceCaCertificate?: CaCertificate;
```
msambol, line 101: could you change this to just `caCertificate` ?

packages/aws-cdk-lib/aws-docdb/lib/ca-certificate.ts
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * rds-ca-rsa4096-g1 certificate authority
+   */
+  public static readonly RDS_CA_RSA4096_G1 = new CaCertificate('rds-ca-rsa4096-g1');
+
+  /**
+   * @param identifier - CA certificate identifier
+   */
+  private constructor(private readonly identifier: string) { }
+
```
msambol, line 26: I see RDS has the following:

```
  /**
   * Custom CA certificate
   *
   * @param identifier - CA certificate identifier
   */
  public static of(identifier: string) {
    return new CaCertificate(identifier);
  }
```

Do we need to support this? If not, this could be an `enum`.

msambol (commented):
Great stuff! See inline comments.

packages/aws-cdk-lib/aws-docdb/lib/ca-certificate.ts
brokad: That's a good question!

#27138 allows for it, so we probably should too for consistency. Anyway, in case a new CA gets added, it is probably better to have this instead of relying on a property override?

I've added the method.

brokad:
> Great stuff! See inline comments.

Thanks for the review!

packages/aws-cdk-lib/aws-docdb/lib/ca-certificate.ts
msambol: Possible to add a unit and integration test for this?

msambol (approved):
LGTM

scanlonp (changes requested):
Hey @brokad, thanks for this PR! I think this looks good overall, but I was wondering if we can just use the `CaCertificate` class defined in RDS. It looks like the two classes are identical, and the class seems more tied to RDS than anything.

I am not an expert on certificates, but unless there is a reason not to, I think we can import `CaCertificate` from RDS, and use it the same. This will keep the certificate list in one place for updates and maintainability.

vumdao:
Hope this ticket will be released soon, we'd like to update the CaCertificate through CDK instead of customizing or manually before Aug

brokad:
> Hey @brokad, thanks for this PR! I think this looks good overall, but I was wondering if we can just use the `CaCertificate` class defined in RDS. It looks like the two classes are identical, and the class seems more tied to RDS than anything.
> 
> I am not an expert on certificates, but unless there is a reason not to, I think we can import `CaCertificate` from RDS, and use it the same. This will keep the certificate list in one place for updates and maintainability.

Thanks for the review!

I have updated this PR to import `CaCertificate` from RDS. I noticed there probably is a typo in the variants introduced by f5a5a08a. The RSA CAs are labeled [`RDS_CA_RDS*_G1`](https://github.com/aws/aws-cdk/commit/f5a5a08af8563052f874c61cbe18f61064c00191#diff-908a12331ed40124af045fa94cc2aab6be56fb72c7a3ff3bc1ea37af7cf42625R20), but I'm guessing the intention was `RDS_CA_RSA*_G1`. I elected to not change this in RDS as I am assuming it would be a breaking change for folks already using it.

The update otherwise introduces no change to the previous state of this PR.

scanlonp (approved):
This looks great. I think the re-export is very nice for ease of use!

Looks like you are exactly right on the typo in the certificate enum, but probably does not need to be included in this PR. I can do a small change to update the name.

Thanks!

scanlonp:
@brokad, enum values updated with #30135, good catch.

scanlonp:
@vumdao, should be out in the next release 2.142.0, hopefully mid-next week.

PR #28792, opened 2024-01-20
feat(autoscaling):  `datapointsToAlarm` property for step scaling policy

### LINKED ISSUE #28749: AutoScalingGroup: `DatapointsToAlarm` property support for `BasicStepScalingPolicyProps`
### Describe the feature

Currently, the [scale_on_metric](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_autoscaling.AutoScalingGroup.html#scalewbronwbrmetricid-props) method for the [AutoScalingGroup](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_autoscaling.AutoScalingGroup.html) construct does not support explicitly specifying the `DatapointsToAlarm` property but supports setting `EvaluationPeriods`. The `scale_on_metric` implements `BasicStepScalingPolicyProps`. Hence support `DatapointsToAlarm` in `BasicStepScalingPolicyProps` interface.

### Use Case

This feature would be helpful instead of having to use the workaround of Escape Hatches to modify the underlying `DatapointsToAlarm` property in the CloudWatch alarm generated.

### Proposed Solution

Add property `DatapointsToAlarm` property for `BasicStepScalingPolicyProps` construct.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.114.1

Linux/Windows

go-to-k:
This PR adds a property `datapointsToAlarm` for `StepScalingPolicy`.

Closes #28749.

----


packages/aws-cdk-lib/aws-autoscaling/test/scaling.test.ts
```diff
@@ -327,6 +327,62 @@ test('step scaling with evaluation period configured', () => {
   });
 });
 
+test('step scaling with evaluation period and data points to alarm configured', () => {
```
lpizzinidev, line 330:
```suggestion
describe('datapointsToAlarm', () => {
  test('step scaling with evaluation period and data points to alarm configured', () => {
    ...
```
Let's group these tests to keep it organized.

packages/aws-cdk-lib/aws-autoscaling/lib/step-scaling-policy.ts
```diff
@@ -57,10 +57,26 @@ export interface BasicStepScalingPolicyProps {
...
+  /**
+   * The number of data points out of the evaluation periods that must be breaching to
+   * trigger a scaling action
+   *
+   * Creates an "M out of N" alarm, where this property is the M and the value set for
+   * `evaluationPeriods` is the N value.
+   *
+   * Only has meaning if `evaluationPeriods != 1`.
+   *
+   * @default `evaluationPeriods`
```
lpizzinidev, line 76:
```suggestion
   * @default Same as `evaluationPeriods`
```

packages/aws-cdk-lib/aws-autoscaling/lib/step-scaling-policy.ts
```diff
@@ -102,6 +118,10 @@ export class StepScalingPolicy extends Construct {
       throw new Error(`'scalingSteps' can have at most 40 steps, got ${props.scalingSteps.length}`);
     }
 
+    if (props.datapointsToAlarm !== undefined && !Token.isUnresolved(props.datapointsToAlarm) && props.datapointsToAlarm < 1) {
```
lpizzinidev, line 121: Can we set `datapointsToAlarm` to a value greater than `evaluationPeriods`?
If not, I think we should validate that case as well (and update the documentation for the property).

lpizzinidev (changes requested):
Thanks 👍
I left some notes for minor adjustments.

packages/aws-cdk-lib/aws-autoscaling/lib/step-scaling-policy.ts
go-to-k: CFn returns an error `DatapointsToAlarm must be less than or equal to EvaluationPeriods`. So I will add the validation and the documentation the property.

go-to-k:
@lpizzinidev 

Thanks for your review. I fixed.

packages/aws-cdk-lib/aws-autoscaling/lib/step-scaling-policy.ts
```diff
@@ -102,6 +119,18 @@ export class StepScalingPolicy extends Construct {
...
     }
 
+    if (props.datapointsToAlarm !== undefined && !Token.isUnresolved(props.datapointsToAlarm)) {
+      if (props.datapointsToAlarm < 1) {
+        throw new Error(`datapointsToAlarm cannot be less than 1, got: ${props.datapointsToAlarm}`);
+      }
+      if (props.evaluationPeriods !== undefined
+        && !Token.isUnresolved(props.evaluationPeriods)
+        && props.evaluationPeriods < props.datapointsToAlarm
+      ) {
```
go-to-k, line 129: The condition `if (props.evaluationPeriods < 1)` is not included. This is because `props.datapointsToAlarm < 1` and `props.evaluationPeriods < props.datapointsToAlarm` will cover it.

lpizzinidev (approved):
Thanks 👍

packages/aws-cdk-lib/aws-autoscaling/lib/step-scaling-policy.ts
aaythapa: what about if `evaluationPeriod` is set but `datapointsToAlarm` isn't?

packages/aws-cdk-lib/aws-autoscaling/lib/step-scaling-policy.ts
```diff
@@ -102,6 +119,18 @@ export class StepScalingPolicy extends Construct {
       throw new Error(`'scalingSteps' can have at most 40 steps, got ${props.scalingSteps.length}`);
     }
 
+    if (props.datapointsToAlarm !== undefined && !Token.isUnresolved(props.datapointsToAlarm)) {
+      if (props.datapointsToAlarm < 1) {
+        throw new Error(`datapointsToAlarm cannot be less than 1, got: ${props.datapointsToAlarm}`);
+      }
+      if (props.evaluationPeriods !== undefined
+        && !Token.isUnresolved(props.evaluationPeriods)
+        && props.evaluationPeriods < props.datapointsToAlarm
```
aaythapa, line 128: Should we throw an error if `datapointsToAlarm` is set without setting `evaluationPeriods`. I know `evaluationPeriods` can be set by itself and changing that behavior would break compatibility but we should probably add that check for `datapointsToAlarm`. Can't find a use case for setting `datapointsToAlarm` by itself.

aaythapa (commented):
looks good overall, some minor comments/questions

packages/aws-cdk-lib/aws-autoscaling/lib/step-scaling-policy.ts
go-to-k: That's going to behave the same as it did before, so I'm not going to specifically check that.

packages/aws-cdk-lib/aws-autoscaling/lib/step-scaling-policy.ts
go-to-k: I agree. I will add the validation for only `datapointsToAlarm`.

packages/aws-cdk-lib/aws-autoscaling/lib/step-scaling-policy.ts
aaythapa: Did a quick check and the old behavior is a deployment failure saying `evaluationPeriod` can't be less than 1. A synth-time check would help avoid that and allow people to debug faster.

go-to-k: OK, I will add the validation for `props.evaluationPeriod < 1`

go-to-k: The above validations with the conditions include if `props.evaluationPeriod < 1`, so we could catch `props.evaluationPeriod < 1` in this code. Still, do we need to do anything?

go-to-k: For example, evaluationPeriod for an input is 0:

- `datapoints >= 1` is guaranteed by the value pass `if (props.datapointsToAlarm < 1)` check.
- So `evaluationPeriods >= 1` is guaranteed by the value pass `evaluationPeriods < datapointsToAlarm` check and the above result (`datapoints >= 1`).

go-to-k: Sorry, I misunderstood. The above conditions are in `props.datapointsToAlarm !== undefined`. So we should check regardless of whether datapointsToAlarm is undefined or not. I will add the validation!

go-to-k: @aaythapa 

Changed it again and added its validation.

aaythapa (dismissed):
🚀

aaythapa (approved):

Issue #27188, opened 2023-09-19
(docdb): (L3 constructs permit creating a cluster from snapshot)


jordiredondo:
### Describe the feature

Hi,

Unlike RDS CDK constructs, for docdb there is no choice to create a docdb cluster from scratch using an snapshot unless it is used L1 constructs. The problem if we create the cluster from L1 constructs, we automatically loose the choice to work with L3/L4 constructs like password rotation feature

### Use Case

I want to keep docdb L3/L4 power with the choice of creating a cluster from snapshot

### Proposed Solution

bring parity to what we can do already for RDS being possible to create with L3/L4 clusters from snapshots

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.83.1

### Environment details (OS name and version, etc.)

macos

pahud:
Thanks for the feature request. Yes, we should consider to create a L2 construct that allows users to create clusters from snapshots. We welcome all pull requests from the community and please help us prioritize with upvotes.

PR #28798, opened 2024-01-21
feat(docdb): support snapshot removal policy

### LINKED ISSUE #28773: (aws-docdb): setting up cluster with retention policy  == snapshot breaks internal security group
### Describe the bug

Setting up a cluster with `removalPolicy: cdk.RemovalPolicy.SNAPSHOT` fails to synth because `ec2.SecurityGroup` does not supoort said `RemovalPolicy`

### Expected Behavior

`cdk synth` executed correctly.

### Current Behavior

`cdk synth` fails with `Error: AWS::EC2::SecurityGroup does not support snapshot removal policy`

### Reproduction Steps

```typescript
#!/usr/bin/env node
import * as docDb from 'aws-cdk-lib/aws-docdb';

const app = new cdk.App();
const env = {
  account: 'your account',
  region: 'your region',
};
const stack = new cdk.Stack(app, 'DocDB', {
  env,
});

new docDb.DatabaseCluster(stack, 'Cluster', {
  removalPolicy: cdk.RemovalPolicy.SNAPSHOT,
  masterUser: {
    username: 'test',
    password: cdk.SecretValue.unsafePlainText('test'),
  },
  instanceType: ec2.InstanceType.of(
    ec2.InstanceClass.T4G,
    ec2.InstanceSize.MEDIUM,
  ),
  vpc: ec2.Vpc.fromLookup(stack, 'VPC', { isDefault: true }),
});

```

### Possible Solution

1. Fallback to retaining security group if document db has snapshot policy set
2. Do not set retention policy inside construct

Personally I prefer the 2nd idea, because setting a retention policy inside Construct does seem to be a bit of an odd idea.
I think setting up one with CDK defaults make more sense and if user wants to change the default it can always create a security group before hand and pass it as `prop`

_No response_

### CDK CLI Version

2.121.1

### Framework Version

_No response_

### Node.js Version

20.8

### OS

MacOS sierra

### Language

TypeScript

### Language Version

5.3

### Other information

_No response_

lpizzinidev:
Adds support for `removalPolicy: RemovalPolicy.SNAPSHOT` for DocumentDB clusters as specified in the [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html).

To allow users to specify custom policies for the cluster's instances and security group the following properties have been added:
* `instanceRemovalPolicy`
* `securityGroupRemovalPolicy`

Closes #28773.
Closes https://github.com/aws/aws-cdk/issues/28861

----


lpizzinidev:
Exemption Request.
The feature is currently not documented.
Let me know if we should add some information about it.

kornicameister:
@lpizzinidev but is it a good idea to silently fallback to another removal policy here? Personally I find it better to know that something is not supported instead of learning sometime later in production that there is no snapshot although my infrastructure code says there should be.

lpizzinidev:
@kornicameister 
From my understanding, the snapshot removal policy [is supported](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html) for `AWS::DocDB::DBCluster`.
The fallback to delete is used on `AWS::DocDB::DBInstance` and `AWS::EC2::SecurityGroup` resources since they don't support a snapshot removal policy.
The cluster's snapshot is created.
I think this is a valid approach for handling this feature, please let me know if you have suggestions for changing the current implementation.

kornicameister:
@lpizzinidev suggestion remains.
I would maintain creating security group with defaults which excludes setting up any retention policies inside of a document DB construct. 

If user wants to retain w security group it may create one before creating a cluster or an instance. That way, at least, caller is aware of a setting instead of it being magically transformed between incompatible resources.

What it user wishes to remove security group but retain cluster. What is there is a need to remove cluster but retain a group. All in all there's more than one combination and I find it better to let caller decide.

Finally... isn't what you are proposing, potentially dangerous, precedence? I know CDK comes with reasonable defaults but, at least to my knowledge, dealing with security group is always exposed for caller to customize. It is like that here but there is this one extra bit that requires extra documentation.

Are there more places hacking their way through escape hatches like that? Or is it really a precedence?

kornicameister:
@lpizzinidev your link leads to RDS.

Also I tried setting up snapshot policy for cluster but `cdk synth` failed for me with unsupported policy error.

lpizzinidev:
@kornicameister 
If you [scroll down](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html#aws-attribute-deletionpolicy-options) you can see the full list of resources supporting a snapshot removal policy.
If you synth using the current CDK version it won't work as the [list of supported resources](https://github.com/aws/aws-cdk/blob/df8fbc48c2eaf59c3380d3ffa46d0024ef88ada5/packages/aws-cdk-lib/core/lib/cfn-resource.ts#L143) if outdated.
We can discuss what's the best fallback option for the resources not supporting a snapshot removal policy.
My approach is based on the idea that the user probably doesn't want to retain resources when specifying a snapshot policy, so I simply fall back to delete.

kornicameister:
@lpizzinidev I rest my case about it being unsupported...my bad. However that begs a question where is this list and how we can update it. That's a bit of a separate issue I am happy to report.

Nonetheless I still do not agree with this fallback. It still sounds a bit like hidden hack here.

Perhaps it's time to get other people here to express opinion because I am just one guy and it feels wrong just for me which is hardly an argument if general attitude in cdk is to let those hacks happen. In that case I have to live with it 😝

lpizzinidev:
@kornicameister 
Yep, it's up to the core team to discuss the intended behavior, mine is just a proposal.
Anyway, I've updated the documentation as this might be confusing for users as well.
Thanks for following up with your suggestions.

packages/aws-cdk-lib/rosetta/aws_docdb/default.ts-fixture
```diff
@@ -1,5 +1,5 @@
 // Fixture with packages imported, but nothing else
-import { Stack } from 'aws-cdk-lib';
+import { RemovalPolicy, Stack } from 'aws-cdk-lib';
```
GavinZZ, line 2: Curious why is this change needed?

GavinZZ (commented):
I am partly agreeing with this change. 

To be fair, I agree with the security group change for the following reason:
1. The default removal policy for security group is`DELETE`. It makes sense to me to fall back to the default policy.
2. Users wasn't able to use `SNAPSHOT` policy on DocDB cluster resource prior to this fix, so it won't cause breaking change.
3.  Users can always provide their own security group if they want other removal policy (and we should add an example to readme to set a custom deletion policy on security group with `SNAPSHOT` policy on docdb cluster).

However, I'm a bit hesitant on the DocDB instance part. I appreciate the fact that this PR now adds support for `SNAPSHOT` deletion policy for cluster. Correct me if wrong, but there's no way to specify `RETAIN` instance while using `SNAPSHOT` on cluster. 


### Edit
> but there's no way to specify `RETAIN` instance while using `SNAPSHOT` on cluster.

I guess this is not entirely true, as there's a pretty ugly workaround like the following
```
cluster.node.children.forEach(child => {
  if (child instanceof docdb.CfnDBInstance) {
    child.applyRemovalPolicy(cdk.RemovalPolicy.RETAIN, { applyToUpdateReplacePolicy: true } as cdk.RemovalPolicyOptions);
  }
});
```

I personally don't like this workaround but I also can't think of a better way other than creating another property (which I hate even more). I'll bring it up for team discussion for a second thought.

GavinZZ:
> but there's no way to specify `RETAIN` instance while using `SNAPSHOT` on cluster.

I guess this is not entirely true, as there's a pretty ugly workaround like the following
```
cluster.node.children.forEach(child => {
  if (child instanceof docdb.CfnDBInstance) {
    child.applyRemovalPolicy(cdk.RemovalPolicy.RETAIN, { applyToUpdateReplacePolicy: true } as cdk.RemovalPolicyOptions);
  }
});
```

I'll bring it up for team discussion for a second thought.

packages/aws-cdk-lib/rosetta/aws_docdb/default.ts-fixture
lpizzinidev: I've added an example using `RemovalPolicy.SNAPSHOT` in the DocDB README file.
It's required for rosetta to resolve the dependency.

lpizzinidev:
@GavinZZ 

> Correct me if wrong, but there's no way to specify RETAIN instance while using SNAPSHOT on cluster.

That's correct.
However, I think that if we want to allow users to specify a different policy when the cluster is the to SNAPSHOT, then I think the clearer approach is just providing a separate prop for that.
What if there are users who need to apply the DELETE policy in that case?

kornicameister:
On there other hand.
I haven't checked that but assuming one can restore from snapshot. Should we either handle this here or create separate issue?

kornicameister:
@lpizzinidev @GavinZZ ; friendly ping here.
I am quite interested in being able to do restoration from snapshot, so bumping up a question.

GavinZZ:
I would think that as a separate issue and PR. 

@lpizzinidev I do agree with you that adding a separate props for the removal policy is the best way to go here. Would you be available to work on it? If not I can help pushing commits to this PR.

kornicameister:
@GavinZZ later in the evening I will submit another ticket for passing sourceSnapshot into cluster to make restoration possible.

lpizzinidev:
@GavinZZ 
Sure, I'll work on it during the weekend 👍

kornicameister:
@lpizzinidev @GavinZZ issue is already created by someone else: #27188.

packages/@aws-cdk-testing/framework-integ/test/aws-docdb/test/integ.cluster-removal-policy-snapshot.ts
```diff
@@ -0,0 +1,59 @@
...
+        username: 'docdb',
+        password: cdk.SecretValue.unsafePlainText('7959866cacc02c2d243ecfe177464fe6'),
+      },
+      instanceType: ec2.InstanceType.of(ec2.InstanceClass.R5, ec2.InstanceSize.LARGE),
+      vpcSubnets: { subnetType: ec2.SubnetType.PUBLIC },
+      vpc,
+      parameterGroup: params,
+      kmsKey,
+      removalPolicy: cdk.RemovalPolicy.SNAPSHOT,
+      instanceRemovalPolicy: cdk.RemovalPolicy.RETAIN, // Remember to cleanup after running this test
```
GavinZZ, line 45: For integ test, I would recommend to keep it as `DESTROY`.

packages/aws-cdk-lib/aws-docdb/test/cluster.test.ts
```diff
@@ -956,6 +956,138 @@ describe('DatabaseCluster', () => {
...
+    // WHEN
+    new DatabaseCluster(stack, 'Database', {
+      instances: 1,
+      masterUser: {
+        username: 'admin',
+        password: cdk.SecretValue.unsafePlainText('tooshort'),
+      },
+      instanceType: ec2.InstanceType.of(ec2.InstanceClass.BURSTABLE2, ec2.InstanceSize.SMALL),
+      vpc,
+      instanceRemovalPolicy: cdk.RemovalPolicy.DESTROY,
```
GavinZZ, line 1016: Can we have this test instead as `SNAPSHOT` removal policy with instance removal policy `RETAIN`?

GavinZZ (commented):
Love it, thanks for making the changes. Just two minor comments on test cases.

GavinZZ (approved):
LGTM thanks for contributing!

GavinZZ:
@mergify update

GavinZZ (approved):

PR #28800, opened 2024-01-21
fix(cognito): getting user pool client secret creates unnecessary custom resource

### LINKED ISSUE #28785: UserPoolClient - Retrieving the client secret requires an unnecessary custom resource
### Describe the bug

A change was merged previously to CDK to support a workaround for getting the client secret. This custom resource requires someone deploying a stack with only cognito to have a VPC attachment to the custom resource for accounts with SCPs limiting lambdas to have to run in a VPC.

### Expected Behavior

Native CFN is used to get Cognito Client Secrets

### Current Behavior

As implemented here, a custom resource is used to get the client secret:
https://github.com/aws/aws-cdk/pull/21262/files#diff-9713362aa6af827d0bf2a8c68319b5bb9c74f888f9ab417266ff1b98aa121ae2R429

### Reproduction Steps

create a cognito app client w/ secret

```python
self.client = user_pool.add_client(
            "service-client",
            user_pool_client_name="service-client",
            supported_identity_providers=[
                aws_cognito.UserPoolClientIdentityProvider.COGNITO
            ],
            o_auth=self.o_auth_settings,
            auth_flows=aws_cognito.AuthFlow(
                user_srp=True,
            ),
            refresh_token_validity=Duration.days(1),
            generate_secret=True,
        )

self.my_secret = SecretValue.unsafe_unwrap(self.client.user_pool_client_secret)
```

### Possible Solution

Remove the custom resource and instead generate the following CFN during synth to access the client secret:

```json
UserPoolClientIdSecret:{
    Value: {
        "Fn::GetAtt": ["CognitoUserPoolClient", "ClientSecret"]
    }
}
```

_No response_

### CDK CLI Version

2.117.0

### Framework Version

_No response_

### Node.js Version

v18.17.1

### OS

OSX Sonoma

### Language

Python

### Language Version

3.10.11

### Other information

_No response_

laurelmay:
This updates the `UserPoolClient` construct to not use a custom resource to fetch the secret. I thought about gating this behind a feature flag (and am still happy to do so) but this should functionally behave the same (the underlying value doesn't change, just the token) and it doesn't require any code changes for end users, it just removes a now-unneeded construct.

Without `generateSecret: true` the `ClientSecret` attribute is not available so those checks are still retained (as are the related tests).

The `ClientSecret` attribute seems to have been available in the resource schema for quite awhile but it isn't documented on the CloudFormation resource page.

Closes #28785.

----


kylelaker:
Woo! Not sure why the integration testing didn't replace the hosted zone and domain correctly... I guess I'll try again.

PR #28802, opened 2024-01-21
feat(cognito): validate oidc provider name

### LINKED ISSUE #28667: aws-cognito: Add PCRE Regex Validation
### Describe the feature

When creating a userpool using an OICD identity provider, UserPoolIdentityProviderOidc(), The provider name must follow a PCRE convention, which can throw an error in CFN if not matched. CFN displays: /[^_\p{Z}][\p{L}\p{M}\p{S}\p{N}\p{P}][^_\p{Z}]+/ (basically anything, but no spaces, no underscores).
Feature would be to add a type check, or a check during a local build, that shows if the name matches this. 

### Use Case

Saves time, if a name is passed with "_" or " " (In my case _) then you need to wait for rollback, and depending on how much of the other cognito portions were created, may have to manually delete the user pool that was half created after rollback.

### Proposed Solution

Add a Type check, maybe something like 
```
type RegexString<P extends RegExp> = string extends P ? string : string;
const provrule = /[^_\p{Z}][\p{L}\p{M}\p{S}\p{N}\p{P}][^_\p{Z}]+/;
type provname = RegexString<typeof provrule>;

//from cdk /aws-cdk-lib/aws-cognito/lib/user-pool-idps/oidc.d.ts
export interface UserPoolIdentityProviderOidcProps extends UserPoolIdentityProviderProps {
//
// ...
//
     * @default - the unique ID of the construct
     */
    readonly name?: provname;
```
But this doesn't work with PCRE, if I have time later to get something working I will try, but opening this anyways.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.117.0

Linux / Centos 7, x86_64

msambol:
Closes #28667.

----


packages/@aws-cdk-testing/framework-integ/test/aws-cognito/test/integ.user-pool-idp.oidc.ts
```diff
@@ -14,7 +14,7 @@ const userpool = new UserPool(stack, 'pool', {
 
 new UserPoolIdentityProviderOidc(stack, 'cdk', {
   userPool: userpool,
-  name: 'cdk',
+  name: 'my-provider',
```
msambol, line 17: I think the pattern is right, but `cdk` is no longer working. Perhaps this is a new check from CloudFormation?

packages/aws-cdk-lib/aws-cognito/lib/user-pool-idps/oidc.ts
```diff
@@ -140,6 +136,9 @@ export class UserPoolIdentityProviderOidc extends UserPoolIdentityProviderBase {
       if (!Token.isUnresolved(name) && (name.length < 3 || name.length > 32)) {
         throw new Error(`Expected provider name to be between 3 and 32 characters, received ${name} (${name.length} characters)`);
       }
+      if (!name.match(/[^_\p{Z}][\p{L}\p{M}\p{S}\p{N}\p{P}][^_\p{Z}]+/)) {
```
kylelaker, line 139: Because this regex is long and uses less-common character classes, what do you think about linking to the docs that reference the pattern?

```suggestion
      // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpoolidentityprovider.html#cfn-cognito-userpoolidentityprovider-providername
      if (!name.match(/[^_\p{Z}][\p{L}\p{M}\p{S}\p{N}\p{P}][^_\p{Z}]+/)) {
```

kylelaker (approved):
Thanks! I do think I've seen a few recent validation additions get marked as `feat` instead of `chore`; not sure whether that applies here.

packages/aws-cdk-lib/aws-cognito/lib/user-pool-idps/oidc.ts
msambol: good shout 👍🏼

msambol:
Exemption Request: I only added validation, I don't believe a README update is needed.

IkeNefcy:
Does this function check work? From what I'm reading PCRE is not natural in js/ts. The string CFN uses to check is in PCRE but .match() uses ECMAScript right? I think the string would need to be translated to this regex dialect first for this to work.

kylelaker:
> Does this function check work? From what I'm reading PCRE is not natural in js/ts. The string CFN uses to check is in PCRE but .match() uses ECMAScript right? I think the string would need to be translated to this regex dialect first for this to work.

Good catch! I believe that per [the specification](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp-regular-expression-objects) the correct thing to do here to enable the `u` flag on the regex for unicode support. In this situation, both sites you linked to (regex101 with the JS engine selected and regexr) show a proper parsing of the regex.

packages/aws-cdk-lib/aws-cognito/lib/user-pool-idps/oidc.ts
kylelaker:
```suggestion
      if (!name.match(/^[^_\p{Z}][\p{L}\p{M}\p{S}\p{N}\p{P}][^_\p{Z}]+$/u)) {
```

This will ensure the full string matches and that unicode mode is enabled.

packages/aws-cdk-lib/aws-cognito/lib/user-pool-idps/oidc.ts
```diff
@@ -140,6 +136,10 @@ export class UserPoolIdentityProviderOidc extends UserPoolIdentityProviderBase {
       if (!Token.isUnresolved(name) && (name.length < 3 || name.length > 32)) {
         throw new Error(`Expected provider name to be between 3 and 32 characters, received ${name} (${name.length} characters)`);
       }
+      // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpoolidentityprovider.html#cfn-cognito-userpoolidentityprovider-providername
+      if (!name.match(/[^_\p{Z}][\p{L}\p{M}\p{S}\p{N}\p{P}][^_\p{Z}]+/)) {
```
kylelaker, line 140:
```suggestion
      if (!name.match(/^[^_\p{Z}][\p{L}\p{M}\p{S}\p{N}\p{P}][^_\p{Z}]+$/u)) {
```

This will ensure the full string matches and that unicode mode is enabled.

msambol:
@kylelaker / @IkeNefcy I updated the pattern with the `u`. Thanks!

packages/@aws-cdk-testing/framework-integ/test/aws-cognito/test/integ.user-pool-idp.oidc.ts
GavinZZ: This is a bit concerning, is this caused by the regex change?

kylelaker: Have you tried `cdk` since the regex has been fixed? According to regex101, `cdk` should match.

Trying running the integration test again using the original name. 

If it works with the original name then great! We can keep the name `cdk` and simplify this PR.

If it fails with the original name, then determine if the failure is in CloudFormation or thrown by the CDK. If thrown by the CDK, see if running the same test from the `main` branch succeeds or fails in CloudFormation.

IkeNefcy: Agreed, this was run before the update with the PCRE regex, cdh would not have worked before since it was checking for literal characters, p, Z etc, so if it didn't start with p/Z/_ it wouldn't have worked. But with the fix the start of the string should be saying that the first character cannot be a space or underscore.

msambol: My bad, y'all. I ran it again but I had forgotten to build. We should be good now.

packages/@aws-cdk-testing/framework-integ/test/aws-cognito/test/integ.user-pool-idp.oidc.js.snapshot/integ-user-pool-idp-google.template.json
```diff
@@ -34,9 +34,6 @@
   "poolclient2623294C": {
    "Type": "AWS::Cognito::UserPoolClient",
    "Properties": {
-    "UserPoolId": {
```
GavinZZ, line 37: looks like just order change in this integ test snapshot. We could probably remove this change to make the PR clearer.

msambol: @GavinZZ It is a little messier but it caught an error with the regex. I'd prefer to leave it in for peace of mind. 🙂

GavinZZ: Maybe I missed something, but I don't see any integ test changed only snapshots changed. What error are you referring to here?

msambol: See convo here: https://github.com/aws/aws-cdk/pull/28802#discussion_r1461173898.

GavinZZ: I'm not following, was that change in provider name a mistaken? I don't see any integration test file (not snapshot) change now.

msambol: I'll delete the snap changes. I just wanted validation the stack still deployed with the change.

GavinZZ:
We've had some issue with CodeBuild which is why the CI is not passing. We're actively working on a fix, sorry for any inconvenience.

GavinZZ:
@Mergify update

GavinZZ (approved):
LGTM, thanks for contributing!

PR #28810, opened 2024-01-22
fix(ec2): export NatGatewayProvider for consistency with NatInstanceProvider

### LINKED ISSUE #28372: aws-ec2: Export NatGatewayProvider

crh23:
Export `NatGatewayProvider` for two reasons: to allow instantiation with `new`, and to make docs present it as an implementation of `NatProvider`.
Also added a unit test for the same "functionality" for `NatInstanceProvider`, for symmetry with the added test for `NatGatewayProvider`. 
Unit test and integration test for new functionality added.

Closes #28372.

----


crh23:
Clarification Request: this change is somewhere between a feature and a bug fix ([actual change is tiny](https://github.com/aws/aws-cdk/pull/28810/files#diff-1228e5eb29751308416d9af171ef17872ef911c3b2f1249a78e120a1f22c3565)).

1. Should I relabel it as a bug fix?
2. If not, is a README change required? The enabled functionality is very minor, and the previously-available version isn't in the README currently (closest is "[Using NAT instances](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2-readme.html#using-nat-instances)")

crh23:
Applying a bit more thought, I think this is a bug fix more than a feature - the change to actual functionality is the addition of a single `export`, and it seems unintentional that that wasn't already present,

packages/aws-cdk-lib/aws-ec2/test/vpc.test.ts
```diff
@@ -1641,6 +1651,25 @@ describe('vpc', () => {
       Template.fromStack(stack).resourceCountIs('AWS::EC2::Instance', 1);
     });
 
+    test('Can instantiate NatGatewayProvider directly with new', () => {
```
msambol, line 1654: `Can instantiate NatInstanceProvider directly with new`

packages/aws-cdk-lib/aws-ec2/test/vpc.test.ts
```diff
@@ -1256,6 +1258,14 @@ describe('vpc', () => {
       expect(natGatewayProvider.configuredGateways.length).toBeGreaterThan(0);
     });
 
+    test('Default NAT gateway provider instantiated directly with new', () => {
```
msambol, line 1261: nit: `Can instantiate NatGatewayProvider directly with new`

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.nat-gateway-provider.ts
```diff
@@ -0,0 +1,36 @@
+import * as cdk from 'aws-cdk-lib';
+import * as ec2 from 'aws-cdk-lib/aws-ec2';
+import { IntegTest } from '@aws-cdk/integ-tests-alpha';
+
+class NatGatewayStack extends cdk.Stack {
+  constructor(scope: cdk.App, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    /// !show
```
msambol, line 9: what's `/// !show` ?

msambol (changes requested):
Few nits but looks good to me.

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.nat-gateway-provider.ts
TheRealAmazonKendra, line 9: We don't use pragma anymore.

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.nat-gateway-provider.ts
```diff
@@ -0,0 +1,36 @@
...
+    // Change one of the created public subnets to have a route to the NAT Gateway
+    new ec2.CfnRoute(this, 'NewRoute', {
+      routeTableId: vpc.publicSubnets[0].routeTable.routeTableId,
+      destinationCidrBlock: '1.2.3.4/32',
+      natGatewayId: natGatewayProvider.configuredGateways.find(x => (x.az === vpc.publicSubnets[0].availabilityZone))!.gatewayId,
+    });
+    /// !hide
+
+    Array.isArray(vpc);
+    Array.isArray(natGatewayProvider.configuredGateways);
```
TheRealAmazonKendra, line 27: What are these for?

TheRealAmazonKendra:
Hmm, not totally sure how I feel about this one, not because anything you've done here is invalid but because it looks to me like the error here was exporting the other function, not the lack of export on this one. We can't undo the export because that would be a breaking change, but I'm not sure if it's better to have it consistently wrong in both places or inconsistent.

crh23:
> Hmm, not totally sure how I feel about this one, not because anything you've done here is invalid but because it looks to me like the error here was exporting the other function, not the lack of export on this one. We can't undo the export because that would be a breaking change, but I'm not sure if it's better to have it consistently wrong in both places or inconsistent.

I've seen folks be confused by the current inconsistency. In the design of the CDK docs, the natural way (for me) to find the implementations of an abstract class (such as `NatProvider`) is to view the classes listed in _Implemented by_. Currently this misses `NatGatewayProvider`, but includes `NatInstanceProvider`. The new `NatInstanceProviderV2`, which was added since I raised this CR, is also `export`ed.

I think I agree that it would be best to `export` none of them (and add a usage note to the docs of `NatProvider`), but that's not an option now, and I think `export`ing one but not the other (current situation) is worse than `export`ing all (this CR).

TheRealAmazonKendra (approved):
> I've seen folks be confused by the current inconsistency. In the design of the CDK docs, the natural way (for me) to find the implementations of an abstract class (such as NatProvider) is to view the classes listed in Implemented by. Currently this misses NatGatewayProvider, but includes NatInstanceProvider. The new NatInstanceProviderV2, which was added since I raised this CR, is also exported.
>
> I think I agree that it would be best to export none of them (and add a usage note to the docs of NatProvider), but that's not an option now, and I think exporting one but not the other (current situation) is worse than exporting all (this CR).

Fair point. 

YOLO

TheRealAmazonKendra (dismissed):

TheRealAmazonKendra:
Drat. This failing test has nothing to do with your change and everything to do with how long your waiting on this approval. I'll fix it.

crh23:
> Drat. This failing test has nothing to do with your change and everything to do with how long your waiting on this approval. I'll fix it.

Ack - I recall getting integ tests working in my environment was a bit of a chore, so happy to let you do it this time 😁 
(assuming that is what the issue is)

TheRealAmazonKendra (approved):
YOLO Again. I should probably stop saying that in official PR reviews... ¯\_(ツ)_/¯

PR #24331, opened 2023-02-24
feat(stepfunctions): distributed map state

beck3905:
Adds support for Step Functions Map state in Distributed mode. I decided to implement this as a new state type rather than as properties on the existing Map state. I thought that would make for a simpler user experience rather than having a bunch of complicated combinations of configuration properties.

Closes aws#23265.
Closes aws#23216.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #28821, opened 2024-01-23
feat(stepfunctions): distributed map construct

### LINKED ISSUE #23216: (stepfunctions): Add native support for DistributedMap
### Describe the feature

https://docs.aws.amazon.com/step-functions/latest/dg/concepts-orchestrate-large-scale-parallel-workloads.html

Currently there is a workaround this this but we should have inbuilt support to appropriately define `ItemReader`, `ResultWriter` etc.

### Use Case

Ability to use the new feature.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.53.0

mac

abdelnn:
Adds support for Step Functions Map state in Distributed mode. Currently, in order to create a Distributed Map in CDK, CDK users have to define a Custom State containing their Amazon States Language definition. 

This solution consists of the creation of a new L2 construct, `DistributedMap`. This design decision was made due to the fact that some fields are exclusive to Distributed Maps, such as `ItemReader`. Adding support for it through the existing `Map` L2 construct would lead to some fields being conditionally available.

Some design decisions that were made:
- I created an abstract class `MapBase` that encapsulates all fields currently supported by both `inline` and `distributed` maps. This includes all currently supported fields in the CDK except for `iterator` and `parameters` (deprecated fields). Those are now part of the Map subclass which extends `MapBase`. All new Distributed Maps fields are part of the new `DistributedMap` construct (also a subclass of `MapBase`)
- Permissions specific to Distributed Maps are added as part of this new construct

Thanks to @beck3905 and their PR https://github.com/aws/aws-cdk/pull/24331 for inspiration. A lot of the ideas here are re-used from the PR cited.

Closes #23216 

----


packages/aws-cdk-lib/aws-stepfunctions/lib/states/map.ts
```diff
@@ -272,14 +97,36 @@ export class Map extends State implements INextable {
...
+  /**
+   * Define item processor in Map.
+   *
+   * A Map must either have a non-empty iterator or a non-empty item processor (mutually exclusive  with `iterator`).
+   */
+  public itemProcessor(processor: IChainable, config: ProcessorConfig = {}): Map {
+    const name = `Map ${this.stateId} Item Processor`;
+    const stateGraph = new StateGraph(processor.startState, name);
+    super.addItemProcessor(stateGraph, config);
+    return this;
```
abdelnn, line 130: I had to add those methods to the individual subclasses (vs. the parent `MapBase`) because of backwards incompatibility concerns

```
err  - METHOD aws-cdk-lib.aws_stepfunctions.Map.addCatch: returns aws-cdk-lib.aws_stepfunctions.MapBase (formerly aws-cdk-lib.aws_stepfunctions.Map): aws-cdk-lib.aws_stepfunctions.MapBase does not extend aws-cdk-lib.aws_stepfunctions.Map [change-return-type:aws-cdk-lib.aws_stepfunctions.Map.addCatch]
err  - METHOD aws-cdk-lib.aws_stepfunctions.Map.addRetry: returns aws-cdk-lib.aws_stepfunctions.MapBase (formerly aws-cdk-lib.aws_stepfunctions.Map): aws-cdk-lib.aws_stepfunctions.MapBase does not extend aws-cdk-lib.aws_stepfunctions.Map [change-return-type:aws-cdk-lib.aws_stepfunctions.Map.addRetry]
err  - METHOD aws-cdk-lib.aws_stepfunctions.Map.itemProcessor: returns aws-cdk-lib.aws_stepfunctions.MapBase (formerly aws-cdk-lib.aws_stepfunctions.Map): aws-cdk-lib.aws_stepfunctions.MapBase does not extend aws-cdk-lib.aws_stepfunctions.Map [change-return-type:aws-cdk-lib.aws_stepfunctions.Map.itemProcessor]
```

dominhhai:
IMO, the `ItemBatcher` should be configurable

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map-base.ts
```diff
@@ -0,0 +1,196 @@
...
+   * An upper bound on the number of iterations you want running at once.
+   *
+   * @default - full concurrency
+   */
+  readonly maxConcurrency?: number;
+}
+
+/**
+ * Returns true if the value passed is a positive integer
+ * @param value the value ti validate
```
moltar, line 100:
```suggestion
 * @param value the value to validate
```

abdelnn:
> IMO, the `ItemBatcher` should be configurable

I agree, I've changed it so that it is. Thanks for the feedback!

kaizencc:
I'm going to continue looking at this tomorrow. I haven't finished my review but there were some style changes that I wanted to get in first.

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+export interface IItemReader {
+  /**
+   * S3 Bucket containing objects to iterate over or a file with a list to iterate over
+   */
+  readonly bucket: IBucket;
+
+  /**
+   * Limits the number of items passed to the Distributed Map state
+   *
+   * @default - Distribute Map state will iterate over all items provided by the ItemReader
```
kaizencc, line 25:
```suggestion
   * @default - Distributed Map state will iterate over all items provided by the ItemReader
```

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+interface ItemReaderProps {
+  /**
+   * S3 Bucket containing objects to iterate over or a file with a list to iterate over
+   */
+  readonly bucket: IBucket;
+
+  /**
+   * Limits the number of items passed to the Distributed Map state
+   *
+   * @default - Distribute Map state will iterate over all items provided by the ItemReader
```
kaizencc, line 52:
```suggestion
   * @default - Distributed Map state will iterate over all items provided by the ItemReader
```

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+   *
+   * @default - No prefix
+   */
+  readonly prefix?: string;
+}
+
+/**
+ * Item Reader configuration for iterating over objects in an S3 bucket
+ */
+export class S3ObjectsItemReader implements IItemReader {
```
kaizencc, line 72: should this just be `S3ObjectsReader` or is it important to have the word `Item` in there

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+  /**
+   * Compile policy statements to provide relevent permissions to the state machine
+   */
+  providePolicyStatements(): iam.PolicyStatement[];
+}
+
+/**
+ * Base interface for Item Reader configuration properties
+ */
+interface ItemReaderProps {
```
kaizencc, line 43: why are we not exporting this?

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+        resources: [this.bucket.bucketArn],
+      }),
+    ];
+  }
+}
+
+/**
+ * Base interface for Item Reader configuration properties the iterate over entries in a S3 file
+ */
+export interface S3ItemReaderProps extends ItemReaderProps {
```
kaizencc, line 137: Should this be `S3FileItemReader` or `S3FileReader`?

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+  readonly key: string;
+
+  /**
+   * Limits the number of items passed to the Distributed Map state
+   *
+   * @default - No maxItems
+   */
+  readonly maxItems?: number;
+
+  protected readonly resource: string = `arn:${Aws.PARTITION}:states:::s3:getObject`
```
kaizencc, line 165:
```suggestion
  protected readonly resource: string = `arn:${Aws.PARTITION}:states:::s3:getObject`;
```

We shouldn't have to hardcode arns this way though. We have a method for these things, `Arn.format`. But before you go and change stuff up, I'm curious to know why the implementations of `IItemReader` have this protected property but its not actually a part of the `IItemReader` contract.

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+import { StateGraph } from '../state-graph';
+import { StateMachineType } from '../state-machine';
+import { CatchProps, IChainable, INextable, ProcessorConfig, ProcessorMode, RetryProps } from '../types';
+
+const DISTRIBUTED_MAP_SYMBOL = Symbol.for('@aws-cdk/aws-stepfunctions.DistributedMap');
+
+/**
+ * Base interface for Item Reader configurations
+ */
+export interface IItemReader {
```
kaizencc, line 16: I think anything to do with ItemReader should be in its own file

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+
+  /**
+   * S3 key of a file with a list to iterate over
+   */
+  readonly key: string;
+
+  /**
+   * Limits the number of items passed to the Distributed Map state
+   *
+   * @default - No maxItems
```
kaizencc, line 161: this default does not actually tell me the default

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+export class S3CsvItemReader extends S3ItemReader {
+  /**
+   * CSV headers configuration
+   */
+  readonly csvHeaders: CsvHeaders;
+  protected readonly inputType: string = 'CSV';
+
+  constructor(props: S3CsvItemReaderProps) {
+    super(props);
+    this.csvHeaders = props.csvHeaders || CsvHeaders.useFirstRow();
```
kaizencc, line 292:
```suggestion
    this.csvHeaders = props.csvHeaders ?? CsvHeaders.useFirstRow();
```

Anywhere else too. Use nullish coalescing

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+        ...(this.prefix && { Prefix: this.prefix }),
+      },
+    });
+  }
+
+  /**
+   * Compile policy statements to provide relevent permissions to the state machine
+   */
+  public providePolicyStatements(): iam.PolicyStatement[] {
+
```
kaizencc, line 372:
```suggestion
```

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+    const resource = `arn:${Aws.PARTITION}:s3:::${this.bucket.bucketName}/*`;
```
kaizencc, line 373: All of these need to be using `Arn.format`

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+/**
+ * Interface for ItemBatcher configuration properties
+ */
+export interface ItemBatcherProps {
+  /**
+   * MaxItemsPerBatch
+   *
+   * Specifies the maximum number of items that each child workflow execution processes, as static number
+   *
+   * @default - No maxItemsPerBatch
```
kaizencc, line 398: this does not tell me the actual default. same for the rest of the defaults here

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+   * Render ResultWriter in ASL JSON format
+   */
+  public render(): any {
+    if (
+      !this.props.maxItemsPerBatch &&
+      !this.props.maxItemsPerBatchPath &&
+      !this.props.maxInputBytesPerBatch &&
+      !this.props.maxInputBytesPerBatchPath &&
+      !this.props.batchInput) {
+      return undefined;
```
kaizencc, line 459: We should throw an error here if nothing is specified right? Why is this not a condition of the `validate` function

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+
+    return errors;
+  }
+}
+
+/**
+ * Properties for configuring a Distribute Map state
+ */
+export interface DistributedMapProps extends BaseMapProps {
+  /**
```
kaizencc, line 492: formatting of the docstrings is indented incorrectly here and below

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
```diff
@@ -0,0 +1,770 @@
...
+        resources: [resource],
+      }),
+    ];
+  }
+}
+
+/**
+ * Interface for ItemBatcher configuration properties
+ */
+export interface ItemBatcherProps {
```
kaizencc, line 392: ItemBatcher should be its own file too

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map-base.ts
```diff
@@ -0,0 +1,195 @@
...
+ * @see https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-map-state.html
+ */
+export abstract class MapBase extends State implements INextable {
+  public readonly endStates: INextable[];
+
+  private readonly maxConcurrency: number | undefined;
+  protected readonly itemsPath?: string;
+  protected readonly itemSelector?: { [key: string]: any };
+
+  constructor(scope: Construct, id: string, props: BaseMapProps = {}) {
```
kaizencc, line 128: should be `MapBaseProps`

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map-base.ts
```diff
@@ -0,0 +1,195 @@
...
+ *
+ * While the Parallel state executes multiple branches of steps using the same input, a Map state
+ * will execute the same steps for multiple entries of an array in the state input.
+ *
+ * @see https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-map-state.html
+ */
+export abstract class MapBase extends State implements INextable {
+  public readonly endStates: INextable[];
+
+  private readonly maxConcurrency: number | undefined;
```
kaizencc, line 124:
```suggestion
  private readonly maxConcurrency?: number;
```

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map-base.ts
```diff
@@ -0,0 +1,195 @@
...
+ *
+ * A `Map` state can be used to run a set of steps for each element of an input array.
+ * A Map state will execute the same steps for multiple entries of an array in the state input.
+ *
+ * While the Parallel state executes multiple branches of steps using the same input, a Map state
+ * will execute the same steps for multiple entries of an array in the state input.
+ *
+ * @see https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-map-state.html
+ */
+export abstract class MapBase extends State implements INextable {
```
kaizencc, line 121: Whats the motivation behind moving `MapBase` out of `Map`?

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
kaizencc, line 1: The changes in this file are unrelated and don't change behavior right

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map.ts
kaizencc: Yes doing this will have some backwards compatibility concerns we have to skirt

kaizencc (dismissed):
Ok. I started on a preliminary review. This is going to take a while though, since it is quite involved.

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
abdelnn: This was named this way to represent the fact that it's a type of item reader

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
abdelnn: I'll rename to `S3FileItemReader`

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map-base.ts
abdelnn: Agreed

packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts
abdelnn: Yes

packages/aws-cdk-lib/aws-stepfunctions/lib/states/map-base.ts
abdelnn: With this PR, I wanted to do two things:
- Implement a new Distributed Maps construct
- Ensure that this new construct does not support deprecated ASL fields.

The way I decided to go about this was to abstract the common properties of both distributed and inline maps, and only support deprecated ASL fields (`parameters` and `iterator`) in the existing Map construct for backwards compatibility concerns.

Two other alternatives that I briefly considered but ultimately rejected:
- Adding to the existing Map construct the required Distributed Map fields. This was rejected as having fields be conditionally available based on what kind of map it is seemed unnecessarily convoluted
- Extending the existing Map construct to create the Distributed Map one. This was rejected as the new construct would contain deprecated ASL fields

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
abdelnn: Didn't feel it was necessary as other item props are implementing this interface

abdelnn:
> Yes doing this will have some backwards compatibility concerns we have to skirt

That shouldn't be the case anymore though, right? We're now correctly returning the Map type just like we were before

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
abdelnn: Yes, I'll change that up

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
abdelnn: Miss on my end

packages/aws-cdk-lib/aws-stepfunctions/lib/states/distributed-map.ts
abdelnn: It should be part of the `ItemReader` contract, that's a miss on my end. The resource field is a required field of the `ItemReader`

kaizencc (approved):
Nice! Thanks for the hefty PR @abdelnn. lets give it a try...

packages/aws-cdk-lib/aws-stepfunctions/lib/state-machine.ts
```diff
@@ -444,13 +443,22 @@ export class StateMachine extends StateMachineBase {
 
     this.stateMachineType = props.stateMachineType ?? StateMachineType.STANDARD;
 
+    let graph: StateGraph | undefined = undefined;
+    if (definitionBody instanceof ChainDefinitionBody) {
+      graph = new StateGraph(definitionBody.chainable.startState, 'State Machine definition');
+      graph.timeout = props.timeout;
+      for (const statement of graph.policyStatements) {
+        this.addToRolePolicy(statement);
```
orekav, line 451: @abdelnn is it possible that this is adding the role policies on the wrong place?

Looking at what it was doing before, I would say that it should be
```typescript
this.role.addToRolePolicy(statement);
```

https://github.com/aws/aws-cdk/issues/29203

orekav: As soon as I remove this for and put it back where it was before (Only the for).
The bug disappears.

abdelnn: I believe you're right, this should be 

`this.role.addToPrincipalPolicy(statement)`

PR #28772, opened 2024-01-19
fix(amplify): addBranch fails synth with "cannot find entry file..."

brettstack:
#28658 missed changing one of the paths (i.e it didn't remove the extra '..') and I was still getting the "cannot find entry file" error. I can confirm that my CDK stack works as expected after making the change locally.

This is my first CDK PR. I did the following:

```
yarn install
npx lerna run build --scope=@aws-cdk/aws-amplify-alpha
cd packages/@aws-cdk/aws-amplify-alpha
yarn build
yarn test
yarn integ-runner --update-on-failed
npx lerna run build --scope=aws-cdk-lib
```

Closes #28764

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #28828, opened 2024-01-23
chore(eslint-plugin): rule against invalid paths

kaizencc:
This is a follow up to #28658, #28772, and #28760. We had to fix multiple places where that file path extended beyond the package itself into other areas of the local repository (that would not be available after packaging). This caused myriad issues at synth time with `file not found` errors.

This PR introduces a linter rule with the following specifications:
- no inefficient paths, i.e. no going backwards multiple times. Ex. `path.join(__dirname, '..', 'folder', '..', 'another-folder')`. This should and can be easily simplified
- no paths that go backwards past a `package.json` file. This should catch the instances we faced next time.

The `yarn lint` command on `aws-cdk-lib` took 51.47s seconds without this new rule and 53.32s seconds with the rule enabled. The difference of ~2 seconds shouldn't be a hindrance in this case but I am happy to look for additional efficiencies in the rule I've written.

----

### DIFF
--- a/packages/@aws-cdk/integ-runner/README.md
+++ b/packages/@aws-cdk/integ-runner/README.md
@@ -97,6 +97,12 @@ If not, changes cannot be compared across systems.
   - `go`:
     - File RegExp: `^integ_.*\.go$`
     - App run command: `go run {filePath}`
+- `--list` (default=`false`)
+  List tests instead of running them.
+- `--inspect-failures` (default=`false`)
+  Keep generated snapshots when differences exist in snapshot comparisons.
+- `--max-workers` (default=`16`)
+  The max number of workerpool workers to use when running integration tests concurrently.
 
 Example:
 
--- a/packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
+++ b/packages/aws-cdk-lib/aws-codebuild/lib/linux-arm-lambda-build-image.ts
@@ -21,10 +21,16 @@ interface LinuxArmLambdaBuildImageProps {
 export class LinuxArmLambdaBuildImage implements IBuildImage {
   /** The `aws/codebuild/amazonlinux-aarch64-lambda-standard:nodejs18` build image. */
   public static readonly AMAZON_LINUX_2_NODE_18 = LinuxArmLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-aarch64-lambda-standard:nodejs18');
+  /** The `aws/codebuild/amazonlinux-aarch64-lambda-standard:nodejs20` build image. */
+  public static readonly AMAZON_LINUX_2023_NODE_20 = LinuxArmLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-aarch64-lambda-standard:nodejs20');
   /** The `aws/codebuild/amazonlinux-aarch64-lambda-standard:python3.11` build image. */
   public static readonly AMAZON_LINUX_2_PYTHON_3_11 = LinuxArmLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-aarch64-lambda-standard:python3.11');
+  /** The `aws/codebuild/amazonlinux-aarch64-lambda-standard:python3.12` build image. */
+  public static readonly AMAZON_LINUX_2023_PYTHON_3_12 = LinuxArmLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-aarch64-lambda-standard:python3.12');
   /** The `aws/codebuild/amazonlinux-aarch64-lambda-standard:ruby3.2` build image. */
   public static readonly AMAZON_LINUX_2_RUBY_3_2 = LinuxArmLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-aarch64-lambda-standard:ruby3.2');
+  /** The `aws/codebuild/amazonlinux-aarch64-lambda-standard:corretto21` build image. */
+  public static readonly AMAZON_LINUX_2023_CORRETTO_21 = LinuxArmLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-aarch64-lambda-standard:corretto21');
   /** The `aws/codebuild/amazonlinux-aarch64-lambda-standard:corretto17` build image. */
   public static readonly AMAZON_LINUX_2_CORRETTO_17 = LinuxArmLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-aarch64-lambda-standard:corretto17');
   /** The `aws/codebuild/amazonlinux-aarch64-lambda-standard:corretto11` build image. */
--- a/packages/aws-cdk-lib/aws-codebuild/lib/linux-lambda-build-image.ts
+++ b/packages/aws-cdk-lib/aws-codebuild/lib/linux-lambda-build-image.ts
@@ -21,10 +21,16 @@ interface LinuxLambdaBuildImageProps {
 export class LinuxLambdaBuildImage implements IBuildImage {
   /** The `aws/codebuild/amazonlinux-x86_64-lambda-standard:nodejs18` build image. */
   public static readonly AMAZON_LINUX_2_NODE_18 = LinuxLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-x86_64-lambda-standard:nodejs18');
+  /** The `aws/codebuild/amazonlinux-x86_64-lambda-standard:nodejs20` build image. */
+  public static readonly AMAZON_LINUX_2023_NODE_20 = LinuxLambdaBuildImage.fromCodeBuildImageId('aws/codebuild/amazonlinux-x86_64-lambda-standard:nodejs20');
   /** The `aws/codebuild/amazonlinux-x86_64-lambda-standard:python3.11` build image. */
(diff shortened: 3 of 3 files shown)


tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
```diff
@@ -0,0 +1,79 @@
...
+          // If path is going forward, we cannot have '..' anymore
+          if (forward && path === pathToParentDir) {
+            validPath = false;
+            // I could break here but then I can't build the simulated path for the error message,
+            // and in practice this should be a trivial amount of extra work.
+          }
+          if (path !== pathToParentDir) {
+            forward = true;
+          }
+          paths.push(node.arguments[i].value);
```
paulhcsun, line 41: any reason not to use `paths.push(path);` here?

paulhcsun (commented):
Looks good, just one minor comment.

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
```diff
@@ -0,0 +1,79 @@
+import { Rule } from 'eslint';
+import { isProdFile } from '../private/is-prod-file';
+import * as path from 'path';
+import * as fs from 'fs';
+
+function isPathJoinFuncCall(node: any): boolean {
+  return (
+    node.callee &&
+    node.callee.property &&
+    node.callee.property.name === 'join' &&
```
rix0rrr, line 10:
```suggestion
   node?.callee?.property == 'join' &&
```

?

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export function create(context: Rule.RuleContext): Rule.NodeListener {
+  return {
+    CallExpression(node: any) {
+      if (!isProdFile(context.getFilename())) {
+        return;
+      }
+
+      if (isPathJoinFuncCall(node)) {
+        // Confirm path does not have unnecessary '..'
```
rix0rrr, line 25: First: is this check against oscillating worth the squeeze? (Or is this just to make the subsequent checks easier? If so, I can dig that. Might want to write that down.)

Second: can this not be simplified in the following way:

- `..` may only occur in the path prefix.
- So any `..` whose index is after the first non-`..` is incorrect:

```ts
const components = confirmFirstArgumentIsDirnameAndRemove(node.arguments);

const firstDownDir = components.findIndex((p) => p !== '..');
if (components.some((p, i) => p === '..' && i > firstDownDir) {
  // ERROR: Oscillating paths
  return;
}
```

Then we can observe that `package.json` may only exist in the very last directory that we ended up with by doing `..`: if it occurs anywhere earlier that's an error:

```ts
// Note the - 1 to exclude the last directory from the check
// Exclude the case where there are no '..' at all in the path -- those are never invalid
if (firstDownDir > 0) {
  for (let i = 0; i < firstDownDir - 1; i++) {
    const pjFile = path.join(...[path.dirname(currentFile), ...components.slice(0, i), 'package.json']);
    if (existsSync(pjFile)) {
      // ERROR: this path will end up going out of the package.json directory
      return;
    }
  }
}
```

rix0rrr (changes requested):

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
rix0rrr: Oh you should also add a check that `/` doesn't occur anywhere inside any of the components, otherwise our checks are going to fail:

```
path.join(__dirname, '../../../..', 'banana')
```

Is bad style but works, and anything that accidentally works will occur in our code base 😅. We need to guard against it.

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
kaizencc: 😔

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
kaizencc: nope, i forgot to replace it

kaizencc: good catch

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
```diff
@@ -0,0 +1,79 @@
...
+        const pathToParentDir = '..';
+        const paths: string[] = [];
+        let forward = false; // Determines if the path is officially going forward
+        let validPath = true;
+        for (let i=1; i<node.arguments.length; i++) {
+          const path = node.arguments[i].value;
```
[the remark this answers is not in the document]
kaizencc: Yes, in `isPathJoinFuncCall`:

```ts
function isPathJoinFuncCall(node: any): boolean {
  return (
    node.callee?.property?.name === 'join' &&
    node.arguments.length > 1 &&
    node.arguments[0].name &&
    node.arguments[0].name === '__dirname'
  )
}
```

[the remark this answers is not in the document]
kaizencc: I assume that we don't have any rule that says `path.join` must have `__dirname` as the first argument

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
```diff
@@ -0,0 +1,94 @@
...
+          return;
+        }
+
+        // Note the - 1 to exclude the last directory from the check
+        // Exclude the case where there are no '..' at all in the path -- those are never invalid
+        const currentFile = context.getFilename();
+        if (firstDownDir > 0) {
+          for (let i = 0; i < firstDownDir - 1; i++) {
+            // Note i + 1 here to include the current '..' in the path, since Array.slice excludes the end index.
+            const pjFile = path.join(...[path.dirname(currentFile), ...args.slice(0, i + 1), 'package.json']);
```
kaizencc, line 83: I think this is the correct implementation of your algo @rix0rrr

tools/@aws-cdk/eslint-plugin/test/rules/fixtures/no-invalid-path/variable-path.ts
```diff
@@ -0,0 +1,4 @@
+import * as path from 'path';
+
+const myPath = [__dirname, '..', '../..', 'fixtures'];
+path.join(...myPath); // Should fail but does not currently because myPath is treated as one 'SpreadElement' argument.
```
kaizencc, line 4: What shall we do here? Disallow it? What about a situation like

```ts
export function packageJsonDir(dir: string) {
  fs.readFileSync(path.join(dir, 'package.json'));
}
```

I'm thinking about this too, just ran out of time today.

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
```diff
@@ -0,0 +1,94 @@
...
+        if (!allArgumentsExpected(node)) {
+          // ERROR: unexpected non-string in the argument list
+          context.report({ node, message: 'Part of the arguments to \'path.join()\' is not a string. Only the first argument can be \'__dirname\'.'})
+          return;
+        }
+
+        // We currently do not lint any path.join without '__dirname' as the first argument
+        if (!firstArgIsDirname(node)) {
+          return;
+        }
```
kaizencc, line 57: Can discuss what we want to do in this thread

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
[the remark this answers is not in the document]
rix0rrr: No, but just saying that we don't need to do any checks on `path.join`s where the first argument ISN'T `__dirname`.

tools/@aws-cdk/eslint-plugin/test/rules/fixtures/no-invalid-path/variable-path.ts
rix0rrr: Interesting. I think we can flag ANY array or sequence of arguments in which `__dirname` is followed by strings as suspicious. Doesn't even matter if it's in a `path.join`.

Does this actually occur? Or are you being safe?

As for the `packageJsonDir` -- well those are the limits of static checking, aren't they? We're not trying to catch everything, we're just trying to catch common thoughtless mistakes. How likely is it that people would actually write  that?

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
rix0rrr: This is perfect!

tools/@aws-cdk/eslint-plugin/lib/rules/no-invalid-path.ts
rix0rrr: Yeah, you are right. Good catch! 

I suppose we could also do `i` in [`[1..firstDownDir)`](https://www.varsitytutors.com/hotmath/hotmath_help/topics/interval-notation) as opposed to `[0..firstDownDir - 1)`, and that would have avoided this `i + 1`.

rix0rrr (approved):
Conditionally approved! I'll leave you to decide what to do with the `i + 1`/`i - 1` business

mrgrain:
@mergifyio update

PR #28830, opened 2024-01-23
fix(eks): helm uninstall in custom resource handler does not respect `Wait`

### LINKED ISSUE #28831: (eks): Helm custom handler uninstall does not respect --wait
### Describe the bug

When I use `--wait` in the HelmChart construct, helm does not correctly wait during object deletion.

### Expected Behavior

Helm should pass `--wait` on uninstall.

### Current Behavior

Dependency ordering is not respected.

### Reproduction Steps

* Create helm chart that contains objects with finalizers
* Delete helm chart 
* Helm chart should block deletion until the objects are successfully deleted

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.117.0

### Framework Version

_No response_

### Node.js Version

14

### OS

osx

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

ellistarn:
I have a stack that installs and uninstalls helm charts, where the underlying resources use [Kubernetes Finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/). CDK's helm construct currently ignores `--wait`, which means that any object finalization in those helm charts is not respected in my dependency ordering.

## Compatibility

I could see some debate around whether or not this is a breaking change. I'm currently viewing it as a bug fix. I'm a bit cautious about how valuable it would be to gate this behavior compared to the additional complexity.

## Testing

Tested on a local fork and it worked like a charm

Closes #28831

----


ellistarn:
Exemption Request: https://github.com/aws/aws-cdk/pull/28830#pullrequestreview-1839859109. This code does not have testing built out yet AFAICT.

GavinZZ (approved):
I would classify this issue as a bug and accept the minor breaking changes, so going to approve this PR.

My reasoning for the approval is that prior to this proposed modification, setting `wait=True` resulted in identical behavior to `wait=False` or not set at all. Consequently, this update is reasonable. However, it will alter the existing behavior for users who have previously set `wait=True`, as they will now experience a delay while waiting for the uninstallation process to complete. The change is confined to CLI change and its primary impact is merely extending the duration users will need to wait for the command’s execution to finish.

ellistarn:
Anything I need to do to dismiss the testing linter errors?

Issue #23828, opened 2023-01-25
aws-cognito: Cannot customise "Verify email" string in emailBody of UserPool userVerification


fdansey-ostmodern:
### Describe the bug

When I change the text in the "{##Verify Email##}" placeholder, e.g. to "{##verify your email##}", `cdk synth` and `cdk deploy` commands yield the error,

> Error: Verification email body must contain the template string '{##Verify Email##}'

This is incorrect as that documentation states that this string may be customised.
https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pool-settings-email-verification-message-customization.html

Help text in the AWS Cognito console reads:
> You can customize this message with HTML. "Verify email" is the text that will be displayed over the clickable link in the message. You can customize the "Verify email" string, but the variable - some text enclosed by "{##" and "##}" - must be kept in the message.

### Expected Behavior

I am able to change the `emailBody` property of a UserPool to include the placeholder of format, "{##Verify Your Email##}" **with any custom string** allowed by AWS Cognito.

### Current Behavior

`cdk synth` and `cdk deploy` commands yield the error,

> Error: Verification email body must contain the template string '{##Verify Email##}'
  at UserPool.verificationMessageConfiguration (<PATH>/node_modules/aws-cdk-lib/aws-cognito/lib/user-pool.js:1:10661)
  at new UserPool (<PATH>/node_modules/aws-cdk-lib/aws-cognito/lib/user-pool.js:1:4946)
  at new BlarnStack (<PATH>/blarn-stack/blarn-stack.ts:75:22)
  at Object.<anonymous> (<PATH>/stack-blarn.ts:12:1)
  at Module._compile (node:internal/modules/cjs/loader:1218:14)
  at Module.m._compile (<PATH>/node_modules/ts-node/src/index.ts:1455:23)
  at Module._extensions..js (node:internal/modules/cjs/loader:1272:10)
  at Object.require.extensions.<computed> [as .ts] (<PATH>/node_modules/ts-node/src/index.ts:1458:12)
  at Module.load (node:internal/modules/cjs/loader:1081:32)
  at Function.Module._load (node:internal/modules/cjs/loader:922:12)

### Reproduction Steps

Create a stack containing a cognito.UserPool resource with property,
```
userVerification: {
  emailSubject: "Please verify your email",
  emailBody: `<p>Hello Hottie Pingi!</p>
<p>Please {##verify your email address##}</p>`,
[...]

PR #28832, opened 2024-01-23
fix(cognito): allow custom email msg placeholder

GavinZZ:
When I change the text in the "{##Verify Email##}" placeholder, e.g. to "{##verify your email##}", cdk synth and cdk deploy commands yield the error,
```
Error: Verification email body must contain the template string '{##Verify Email##}'
```
This is incorrect as that documentation states that this string may be customised.
https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pool-settings-email-verification-message-customization.html

Help text in the AWS Cognito console reads:

    You can customize this message with HTML. "Verify email" is the text that will be displayed over the clickable link in the message. You can customize the "Verify email" string, but the variable - some text enclosed by "{##" and "##}" - must be kept in the message.

Expected Behavior

Be able to change the emailBody property of a UserPool to include the placeholder of format, "{##Verify Your Email##}" with any custom string allowed by AWS Cognito.

Closes https://github.com/aws/aws-cdk/issues/23828

----


packages/aws-cdk-lib/aws-cognito/lib/user-pool.ts
```diff
@@ -1069,7 +1070,7 @@
     } else {
       const emailMessage = props.userVerification?.emailBody ??
         `Verify your account by clicking on ${VERIFY_EMAIL_TEMPLATE}`;
-      if (!Token.isUnresolved(emailMessage) && emailMessage.indexOf(VERIFY_EMAIL_TEMPLATE) < 0) {
+      if (!Token.isUnresolved(emailMessage) && !VERIFY_EMAIL_REGEX.test(emailMessage)) {
```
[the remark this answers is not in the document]
GavinZZ: Fixed using lazy qualifier `?`, which will return for the fist `##` found.

packages/@aws-cdk-testing/framework-integ/test/aws-cognito/test/integ.user-pool-link-custom-placeholder.ts
```diff
@@ -0,0 +1,32 @@
...
+  },
+});
+
+new CfnOutput(stack, 'user-pool-id', {
+  value: userpool.userPoolId,
+});
+
+new IntegTest(app, 'user-pool-link-custom-placeholder-integ-test', {
+  testCases: [stack],
+  cdkCommandOptions: {
```
ConnorRobertson, line 25: What's this?

GavinZZ: As far as I understand, it enables options to certain CDK CLI commands. In this case, it allows the `cdk destroy` command to forcefully delete the user pool.

packages/aws-cdk-lib/aws-cognito/test/user-pool.test.ts
```diff
@@ -161,14 +161,35 @@ describe('User Pool', () => {
...
+      userVerification: {
+        emailStyle: VerificationEmailStyle.LINK,
+        emailBody: 'invalid email body {##Verify !! Email##}',
+      },
+    })).toThrow(/Verification email body/);
+
+    expect(() => new UserPool(stack, 'Pool9', {
+      userVerification: {
+        emailStyle: VerificationEmailStyle.LINK,
+        emailBody: 'invalid email body {##Click here to verify##}',
```
gracelu0, line 190: this `emailBody` is valid right? if so it's a bit confusing that it says 'invalid email body'

GavinZZ: good catch.

packages/aws-cdk-lib/aws-cognito/test/user-pool.test.ts
```diff
@@ -159,14 +159,35 @@ describe('User Pool', () => {
...
+      userVerification: {
+        emailStyle: VerificationEmailStyle.LINK,
+        emailBody: 'invalid email body ##Verify Email##',
+      },
+    })).toThrow(/Verification email body/);
+
+    expect(() => new UserPool(stack, 'Pool8', {
+      userVerification: {
+        emailStyle: VerificationEmailStyle.LINK,
+        emailBody: 'invalid email body {##Verify !! Email##}',
```
gracelu0, line 183: from the [docs](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pool-settings-email-verification-message-customization.html), it says that `The maximum length for the message, including the verification code (if present), is 20,000 UTF-8 characters. You can use HTML tags in this message to format the contents.`. so I believe some special characters (like `!`) are allowed? if so we might want to update the regex accordingly

GavinZZ: good catch, looking at the [cfn documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cognito-userpool-verificationmessagetemplate.html#cfn-cognito-userpool-verificationmessagetemplate-emailmessagebylink), the regex they used is `\{##[\p{L}\p{M}\p{S}\p{N}\p{P}\s*]*##\}` which includes symbol, punctuation and accent characters. Let me update it.

gracelu0 (approved):
thanks for making the changes! lgtm :)

PR #28833, opened 2024-01-23
feat(route53): ip-based routing

### LINKED ISSUE #28801: route53: support IP-based routing
### Describe the feature

I want to enable the configuration of [IP-based routing](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy-ipbased.html) in Route 53 through the L2 construct.

### Use Case

Currently, to configure IP-based routing in Route 53, it is necessary to use L1 constructs.

```ts
    // Create a CIDR collection
    const cidrCollection = new route53.CfnCidrCollection(this, 'MyCidrCollection', {
      name: 'MyCidrCollection',
      locations: [
        {
          cidrList: ['192.0.2.0/24', '203.0.113.0/24'],
          locationName: 'LocationName'
        }
      ]
    });

    // Create a RecordSet that uses the CIDR collection
    new route53.CfnRecordSet(this, 'MyRecordSet', {
      // ... other properties ...
      // Reference the CIDR collection
      cidrRoutingConfig: {
        collectionId: cidrCollection.ref,
        locationName: 'locationName',
      },
    });
```

### Proposed Solution

I believe that generating CfnCidrCollection within the recordSet constructor allows for intuitive configuration. 
However, additional handling is required when [sharing the same collection across different RecordSets. ](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy-ipbased.html)

`IP-based routing resource record sets reference a location in a collection, and all resource record sets for the same record set name and type must reference the same collection. For example, if you create websites in two Regions and want to direct DNS queries from two different CIDR locations to a specific website based on the originating IP addresses, then both of those locations must be listed in the same CIDR collection.`

```ts
interface CidrRoutingConfig {
  collectionName?: string;
  locations: Location[];
  // Define only if sharing the same collection with other RecordSets
  collection?: CfnCidrCollection:
}

interface Location {
  cidrList: string[];
  locationName?: string;
}

const record1 = new route53.RecordSet(this, 'IPBasedRoutingRecordSet1', {
  // ... other properties ...
  cidrRoutingConfig: {
    collectionName: 'MyCollection',
    locations: [
      {
        locationName: 'Location1',
        cidrList: ['192.0.2.0/24', '203.0.113.0/24'],
      },
    ],
  },
});

// By adding Location2 to the CfnCidrCollection created in record1, the same CidrCollection can be shared.
const record2 = new route53.RecordSet(this, 'IPBasedRoutingRecordSet2', {
  // ... other properties ...
  cidrRoutingConfig: {
    collection: record1.getCidrCollection(),
    locations: [
      {
        locationName: 'Location2',
        cidrList: ['192.0.3.0/24', '203.0.114.0/24'],
      },
    ],
  },
});
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.116.0

irrelevant

badmintoncryer:
In this PR, I have implemented support for IP-based routing. 
We can configure it from the `cidrRoutingConfig` as shown in the following example:

```ts
declare const myZone: route53.HostedZone;

const record1 = new route53.ARecord(this, 'ARecordIpBased1', {
  zone: myZone,
  recordName: 'test',
  target: route53.RecordTarget.fromIpAddresses('1.2.3.4'),
  cidrRoutingConfig: {
    cidrList: ['192.168.1.0/24', '192.168.16.0/20'],
    locationName: 'TokyoServer',
    collectionName: 'myCollection',
  },
});

// You can also add a new Location to an existing CidrCollection
// Ensure all locations for the same record set name and type are part of the same CIDR collection to guarantee consistent routing.
const record2 = new route53.ARecord(this, 'ARecordIpBased2', {
  zone: myZone,
  recordName: 'test',
  target: route53.RecordTarget.fromIpAddresses('2.3.4.5'),
  cidrRoutingConfig: {
    cidrList: ['192.168.2.0/24', '192.168.48.0/20'],
    locationName: 'LondonServer',
    collection: record1.cidrCollection,
  },
});

// To define a zero bit CIDR block (0.0.0.0/0 or ::/0), use the default ("*") location.
const record3 = new route53.ARecord(this, 'ARecordIpBased3', {
  zone: myZone,
  recordName: 'default',
  target: route53.RecordTarget.fromIpAddresses('1.2.3.4'),
  cidrRoutingConfig: {
    locationName: '*',
  },
});
```

## Question
As mentioned above, the existing `cidrCollection` can be passed as `collection` argument. 
During this process, a new `Location` is added to the existing `Locations` in the `cidrCollection`.
This code works, but I am not confident about the implementation approach. I would greatly appreciate any candid feedback.
```ts
if (cidrRoutingConfig.collection) {
      this._cidrCollection = cidrRoutingConfig.collection;
      const currentLocations = this._cidrCollection.locations ?? [];
      const locationsAsArray = Array.isArray(currentLocations) ? currentLocations : [currentLocations];
      this._cidrCollection.addPropertyOverride('Locations', [...locationsAsArray.map((location) => {
        // Since the location is either CfnCidrCollection.LocationProperty or IResolvable,
        // use a type guard function to ascertain its exact type.
        if ('cidrList' in location && 'locationName' in location) {
          return {
            CidrList: location.cidrList,
            LocationName: location.locationName,
          };
        }
        return location;
      }), {
        CidrList: cidrList,
        LocationName: locationName,
      }]);
    }
```

Closes #28801.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-route53/lib/record-set.ts
+++ b/packages/aws-cdk-lib/aws-route53/lib/record-set.ts
@@ -5,7 +5,7 @@ import { IHostedZone } from './hosted-zone-ref';
 import { CfnCidrCollection, CfnRecordSet } from './route53.generated';
 import { determineFullyQualifiedDomainName } from './util';
 import * as iam from '../../aws-iam';
-import { CustomResource, Duration, IResource, Lazy, Names, RemovalPolicy, Resource, Token } from '../../core';
+import { CustomResource, Duration, IResource, Names, RemovalPolicy, Resource, Token } from '../../core';
 import { CrossAccountZoneDelegationProvider } from '../../custom-resource-handlers/dist/aws-route53/cross-account-zone-delegation-provider.generated';
 import { DeleteExistingRecordSetProvider } from '../../custom-resource-handlers/dist/aws-route53/delete-existing-record-set-provider.generated';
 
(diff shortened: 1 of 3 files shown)


packages/aws-cdk-lib/aws-route53/lib/record-set.ts
```diff
@@ -226,6 +236,46 @@ export interface RecordSetOptions {
...
+   * Existing Cidr Collection.
+   *
+   * Use this to add a new Location to an existing Cidr Collection.
+   * Note that for IP-based routing, all resource record sets for the same record set name and type must reference the same CIDR collection.
+   *
+   * @see https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy-ipbased.html
+   *
+   * @default - Create a new CIDR Collection
+   */
+  collection?: CfnCidrCollection;
```
badmintoncryer, line 276: Since there is no L2 construct for CidrCollection, I am using CfnCidrCollection as an argument.

packages/aws-cdk-lib/aws-route53/lib/record-set.ts
```diff
@@ -233,6 +243,46 @@ export interface RecordSetOptions {
...
+   * @default - zero bit CIDR block (0.0.0.0/0 or ::/0) for the default location
+   */
+  cidrList?: string[];
+  /**
+   * The name of the location.
+   *
+   * When '*' is specified, it is treated as the default location.
+   * In this case, the cidrList cannot be specified.
+   */
+  locationName: string;
```
TheRealAmazonKendra, line 264: We should't have two props that are reliant on the value of the other to work or not work. Since these two seem to be tightly coupled, we want the contract to enforce that.

packages/aws-cdk-lib/aws-route53/lib/record-set.ts
```diff
@@ -233,6 +243,46 @@ export interface RecordSetOptions {
...
+   * Existing Cidr Collection.
+   *
+   * Use this to add a new Location to an existing Cidr Collection.
+   * Note that for IP-based routing, all resource record sets for the same record set name and type must reference the same CIDR collection.
+   *
+   * @see https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy-ipbased.html
+   *
+   * @default - Create a new CIDR Collection
+   */
+  collection?: CfnCidrCollection;
```
TheRealAmazonKendra, line 283: Same comment as above. The peculiarities of how these four props interact should be enforced in the contract.

packages/aws-cdk-lib/aws-route53/lib/record-set.ts
TheRealAmazonKendra: Yes, that's fine. Though, it looks like the CidrCollection would just be the first two props here.

packages/aws-cdk-lib/aws-route53/lib/record-set.ts
```diff
@@ -302,7 +354,14 @@ export class RecordSet extends Resource implements IRecordSet {
     if (props.setIdentifier && (props.setIdentifier.length < 1 || props.setIdentifier.length > 128)) {
       throw new Error(`setIdentifier must be between 1 and 128 characters long, got: ${props.setIdentifier.length}`);
     }
-    if (props.setIdentifier && props.weight === undefined && !props.geoLocation && !props.region && !props.multiValueAnswer) {
+    if (
```
TheRealAmazonKendra, line 357: We have a problem with the contract here already. I see that a lot of these were newly added. We might need to deprecate a bunch of these and restructure the props here. We shouldn't have so many mutually exclusive options.

TheRealAmazonKendra (dismissed):
Thank you for your contribution. I see that you've been making progressive changes in this package and we very much appreciate that work. I do see, however, that a lot of props have been added that interact with each other. I think I'd like to know what else you plan to add so that we can work on an overall contract for all those features. Just adding additional optional props that don't always work together is a pattern we avoid now because it isn't a very good user experience.

badmintoncryer:
oh.. I've forgotten this.

Issue #28292, opened 2023-12-07
parseDeletionPolicy(module name): Reference based deletion policy


stuburg:
### Describe the bug

There is no ability to use reference based deletion policies in the code. 

### Expected Behavior

The code should be able to use the reference to the deletion policy. 

### Current Behavior

Error: Unrecognized DeletionPolicy '[object Object]' is what is output

### Reproduction Steps

 'DeletionPolicy':
    'Default': 'RetainExceptOnCreate'
    'Description': 'Default DeletionPolicy. This parameter sources from AWSWaypointPipelineDeployableCfn.'
    'Type': 'String'
    
    'DeletionPolicy':
      'Ref': 'DeletionPolicy'

### CDK CLI Version

2.114.1 

### Node.js Version

v21.3.0

### OS

Ventura 13.5.1

### Language

TypeScript

### Language Version

TS 5.0.4

tim-finnigan:
Hi @stuburg thanks for reaching out. Can you provide the full steps to reproduce? There was another issue (https://github.com/aws/aws-cdk/issues/26863) a few months ago involving `parseDeletionPolicy`, but that was fixed in [2.9.4](https://github.com/aws/aws-cdk/releases/tag/v2.94.0).

stuburg:
Yea, we are using lpt-to-cdk in a Typescript CDK project. We have the template.yml defined like this in the build: 
`  "DeletionPolicy":
    "Default": "RetainExceptOnCreate"
    "Description": "Default DeletionPolicy."
    "Type": "String"`
    
    and we have resources defined in the template.yml defined like this:
    
   `  "SampleResource":
    "DeletionPolicy":
      "Ref": "DeletionPolicy"`

upon build, that error is generated given that a reference value is passed and not a string.

mrog:
I'm seeing the same error when I use an intrinsic function to provide a deletion policy, like this:

```
Resources:
  ServiceHostedZone:
    Type: "AWS::Route53::HostedZone"
    Properties:
      Name: "MyZoneName" 
    DeletionPolicy:
      !If
        - AllowUndelegateDNS
        - 'Delete'
        - 'Retain'
```

[...]

PR #28834, opened 2024-01-23
fix(iam): allow intrinsic functions in deletion policy

ConnorRobertson:
> # Issue
> > When using `CfnInclude` it was not possible to have an intrinsic function in the `DeletionPolicy`. It only allowed the DeletionPolicy to be explicitly defined. 
> # Solution
> > Check if policy looks like an intrinsic. Check if it is an explicitly defined DeletionPolicy. Then Default to a case that checks if this is an intrinsic if it is return the policy after using parseValue. Else throw the same error it was previously.
> # Important Design Decisions
> > Unsure if any have been made please let me know if there are any I didn't think were Important Design Decisions. 
>
> code you submit.
>

Closes https://github.com/aws/aws-cdk/issues/28292 .

----

### DIFF
--- a/CHANGELOG.v2.alpha.md
+++ b/CHANGELOG.v2.alpha.md
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.123.0-alpha.0](https://github.com/aws/aws-cdk/compare/v2.122.0-alpha.0...v2.123.0-alpha.0) (2024-01-24)
+
+
+### Features
+
+* **iot-actions-alpha:** open search action in IoT topic rule ([#28748](https://github.com/aws/aws-cdk/issues/28748)) ([84b23cb](https://github.com/aws/aws-cdk/commit/84b23cb07d421ec004e412bf48e837ae0d87f7fe))
+
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


packages/aws-cdk-lib/cloudformation-include/test/test-templates/intrinsic-deletion-policy.json
```diff
@@ -0,0 +1,53 @@
...
+            "Fn::Equals": [
+                {
+                    "Ref": "Stage"
+                },
+                "Prod"
+            ]
+        }
+    },
+    "Resources": {
+        "Table": {
```
GavinZZ, line 25: Can we add another resource called i.e. `Table2` that uses `!Ref` intrinsic? The reason being is that `looksLikeCfnIntrinsic` function handles `Ref` different from `Fn::` intrinsics.

GavinZZ (changes requested):

packages/aws-cdk-lib/core/lib/helpers-internal/cfn-parse.ts
```diff
@@ -466,14 +466,21 @@ export class CfnParser {
   }
 
   private parseDeletionPolicy(policy: any): CfnDeletionPolicy | undefined {
+    if (policy === undefined || policy === null) {
+      return undefined;
+    }
+    const isIntrinsic = this.looksLikeCfnIntrinsic(policy) !== undefined;
```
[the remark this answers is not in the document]
ConnorRobertson: This is just to make it a boolean. Instead of checking in the if statement. `looksLikeCfnIntrinsic' returns a string or undefined not a boolean.

packages/aws-cdk-lib/cloudformation-include/test/test-templates/intrinsic-deletion-policy.json
ConnorRobertson: Done.

ConnorRobertson:
Changes requested by Gavin have been done.

GavinZZ:
Changes look good to me. Would love to get a second eye on it before approving it.

paulhcsun (approved):

PR #28841, opened 2024-01-24
feat(batch): add fargate Runtime Platform properties to ECS Fargate C…

### LINKED ISSUE #26484: (aws-batch-alpha): Construct Props RuntimePlatform

xazhao:
The property [RuntimePlatform](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-containerproperties.html#cfn-batch-jobdefinition-containerproperties-runtimeplatform) is not present in the AWS Batch ECS Fargate Job Definition. This PR adds flatten properties fargateCpuArchitecture and fargateOperatingSystemFamily to the ECS Fargate Job Definition in AWS Batch.

Closes #26484.

----


packages/@aws-cdk-testing/framework-integ/test/aws-batch/test/integ.ecs-job-definition.js.snapshot/tree.json
```diff
@@ -1616,7 +1620,7 @@
               }
             },
             "constructInfo": {
-              "fqn": "@aws-cdk/aws-batch-alpha.EcsJobDefinition",
+              "fqn": "aws-cdk-lib.aws_batch.EcsJobDefinition",
```
jfuss, line 1623: Do we know why this changed?

packages/aws-cdk-lib/aws-batch/lib/ecs-container-definition.ts
```diff
@@ -1009,6 +1023,20 @@ export interface EcsFargateContainerDefinitionProps extends EcsContainerDefiniti
    * @default - 20 GiB
    */
   readonly ephemeralStorageSize?: Size;
+
+  /**
+   * The vCPU architecture of Fargate Runtime.
+   *
+   * @default - X86_64
```
jfuss, line 1030: For my own understanding, why set defaults for this and `fargateOperatingSystemFamily`?

packages/@aws-cdk-testing/framework-integ/test/aws-batch/test/integ.ecs-job-definition.ts
```diff
@@ -60,6 +60,8 @@ new batch.EcsJobDefinition(stack, 'ECSFargateJobDefn', {
     memory: Size.mebibytes(32768),
     ephemeralStorageSize: Size.gibibytes(100),
     fargatePlatformVersion: FargatePlatformVersion.LATEST,
+    fargateCpuArchitecture: ecs.CpuArchitecture.ARM64,
+    fargateOperatingSystemFamily: ecs.OperatingSystemFamily.LINUX,
```
jfuss, line 64: Is it guaranteed that the `batch` `EcsFaragetContainerDefintion` will always match `ecs`? I am wondering if there is a case where ECS needs some update and then batch needs to follow. Meaning there would be a time where batch doesn't support all types ECS supports.

I might be looking at this in the wrong way or have an miss understanding on how things connect together within CDK.

jfuss (commented):
All comments are mostly for my understanding of the change. Otherwise, seems like a straight forward change.

packages/@aws-cdk-testing/framework-integ/test/aws-batch/test/integ.ecs-job-definition.js.snapshot/tree.json
xazhao: The last change of this file was to stabilizing (convert alpha to regular module). So the snapshot would be before stabilizing which was still using `aws-batch-alpha`

packages/aws-cdk-lib/aws-batch/lib/ecs-container-definition.ts
xazhao: My understanding is the default value should be same as the default value in [CFN resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-runtimeplatform.html).

packages/@aws-cdk-testing/framework-integ/test/aws-batch/test/integ.ecs-job-definition.ts
xazhao: I saw it was already using enum values from ECS definition so I didn't really think about it... Looking at the current values they're the same but I'm not sure if that's guaranteed or it's documented somewhere.
If we can generate all these values from the schema in the future, then we can be confident about the values here.

packages/aws-cdk-lib/aws-batch/README.md
```diff
@@ -479,6 +479,21 @@ jobDefn.container.addVolume(batch.EcsVolume.efs({
 }));
 ```
 
+### Running a ECS workflow with Fargate container
```
paulhcsun, line 482:
```suggestion
### Running an ECS workflow with Fargate container
```

paulhcsun (approved):

PR #28856, opened 2024-01-25
fix(redshift): Fix Redshift User Secret Multi-User Rotation

### LINKED ISSUE #28852: aws-redshift-alpha: User DatabaseSecret created without masterarn in Secret Text causing multi-user secret rotation to fail
### Describe the bug

Multi-User Secret Rotation on Redshift User Secret does not work without a `masterarn` field present in Redshift User Secret serialized JSON text.

The User Construct does not plumb in the Cluster's secret (i.e. the master secret) to the DatabaseSecret in the User constructor.

A [similar change](https://github.com/aws/aws-cdk/commit/a9e5b609b20eff9edaf5775cfbe0802177852475#diff-f2da017a5ffcbbdabed063fde26c976af3a2b9734233149af9da5287a2aec903) was applied to RDS's [DatabaseSecret](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.DatabaseSecret.html) and Cluster to enable Multi-User Secret Rotation

### Expected Behavior

After enabling Multi-User Secret Rotation:
```
        cluster.addRotationMultiUser(`${id}CredentialsRotation`, {
            secret: redshiftUser.secret,
            automaticallyAfter: secureProps.rotationPeriod,
        });
```

The Redshift User Secret has a `masterarn` so that Multi-User Secret Rotation works

### Current Behavior

The Secret Rotation Lambda fails with

```
2024-01-10T00:23:45.982-08:00
[ERROR] KeyError: 'masterarn'
Traceback (most recent call last):
  File "/var/task/lambda_function.py", line 81, in lambda_handler
    set_secret(service_client, arn, token)
  File "/var/task/lambda_function.py", line 188, in set_secret
    admin_arn = current_dict['masterarn']
```

### Reproduction Steps

```
      const cluster = // cluster props

       const user = return new User(this, 'User', {
            cluster: cluster,
            databaseName: databaseName,
            username: username,
            adminUser: cluster.secret,
        });

        cluster.addRotationMultiUser(`CredentialsRotation`, {
            secret: user.secret
        });
```

### Possible Solution

- add masterSecret to [Redshift's DatabaseSecret](https://github.com/aws/aws-cdk/blob/v2.122.0/packages/%40aws-cdk/aws-redshift-alpha/lib/database-secret.ts) similar to how it as done in [RDS's DatabaseSecret](https://github.com/aws/aws-cdk/commit/a9e5b609b20eff9edaf5775cfbe0802177852475#diff-f2da017a5ffcbbdabed063fde26c976af3a2b9734233149af9da5287a2aec903)
- When [creating a new Redshift User](https://github.com/aws/aws-cdk/blob/v2.122.0/packages/%40aws-cdk/aws-redshift-alpha/lib/user.ts#L153), plumb in admin user secret `props.adminUser` to DatabaseSecret's new masterSecret prop

_No response_

### CDK CLI Version

2.70.0

### Framework Version

_No response_

### Node.js Version

18

### OS

AL2

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

penniman26:
Fixes Redshift User Secret Multi-User Rotation for new Users by including `masterarn` in the Secret's Serialized JSON Object Text.

Note: This doesn't affect existing users (nor fixes roation for them) since the secret string template is only used when the secret is first created. For those existing secrets, the secret text will need to be updated to include `masterarn` using the GetSecretValue and UpdateSecret SecretManager APIs.

Closes #28852


packages/@aws-cdk/aws-redshift-alpha/test/user.test.ts
```diff
@@ -11,11 +11,12 @@ describe('cluster user', () => {
   let cluster: redshift.ICluster;
   const databaseName = 'databaseName';
   let databaseOptions: redshift.DatabaseOptions;
+  let databaseOptionsWithAdminUser: redshift.DatabaseOptions;
 
   beforeEach(() => {
     stack = new cdk.Stack();
     vpc = new ec2.Vpc(stack, 'VPC');
-    cluster = new redshift.Cluster(stack, 'Cluster', {
+    const clusterImpl = new redshift.Cluster(stack, 'Cluster', {
```
lpizzinidev, line 19: Why can't we use the global variable? 🤔

packages/@aws-cdk/aws-redshift-alpha/lib/user.ts
```diff
@@ -153,6 +153,7 @@ export class User extends UserBase {
     const secret = new DatabaseSecret(this, 'Secret', {
       username,
       encryptionKey: props.encryptionKey,
+      masterSecret: props.adminUser,
```
lpizzinidev, line 156: Not sure about this.
Shouldn't we simply pass a `DatabaseSecret` with `masterSecret` set to the `addRotationMultiUser` method?
It seems like we might be assigning unnecessary permissions to the user here.
Also, I think we should validate that the secret for `addRotationMultiUser` has the required `masterSecret` value set [here](https://github.com/aws/aws-cdk/blob/95d187e4aabf69b7210ffdf53e53eda077b46d6e/packages/%40aws-cdk/aws-redshift-alpha/lib/cluster.ts#L653).

lpizzinidev (changes requested):
Thanks 👍
Looks good overall, I left a couple of comments on the implementation.
Also, can you please update the title to [describe the bug (not the solution)](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request)?

packages/@aws-cdk/aws-redshift-alpha/test/user.test.ts
penniman26: `cluster` is ICluster which does not have a `.secret` property. I figured that the rest of the tests wanted to continue to use the ICluster interface so I have this awkward code that has both an ICluster global variable and a Cluster local variable pointing to the same object but different types.

https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-redshift-alpha.ICluster.html
https://github.com/aws/aws-cdk/blob/v2.122.0/packages/%40aws-cdk/aws-redshift-alpha/lib/cluster.ts

packages/@aws-cdk/aws-redshift-alpha/lib/user.ts
penniman26: Absolutely! I was having the same conversation with a teammate this morning about that. The secret itself shouldn't need to encode the masterSecret reference especially since there is nothing secret about that reference (nor anything bad that could happen if an incorrect or different arn was set - the rotation lambda would just fetch wrong cluster admin credentials and fail to make the update!).

I think we may be bound by the implementation of [addRotationMultiUser](https://github.com/aws/aws-cdk/blob/v2.122.0/packages/%40aws-cdk/aws-redshift-alpha/lib/cluster.ts#L654C31-L662) and the [redshift multi-user rotation lambda](https://github.com/aws/aws-cdk/blob/v2.122.0/packages/%40aws-cdk/aws-redshift-alpha/lib/cluster.ts#L534) secrets manager provides.

Re unassigning unecessary permissions, I see your point about sending `masterSecret` into DatabaseSecret but at impl time its just encoding the master secret ARN w/o additional permissions. Its the rotation lambda that has permission to the masterSecret and secret to perform the rotation.

On the validation, I don't know if we can perform that validation at build time since `options.secret` is a reference to ISecret and not DatbaseSecret :/ - thoughts?

penniman26:
> Thanks 👍 Looks good overall, I left a couple of comments on the implementation. Also, can you please update the title to [describe the bug (not the solution)](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#step-4-pull-request)?

Thank you and will do!

packages/@aws-cdk/aws-redshift-alpha/lib/user.ts
lpizzinidev: Makes sense and I agree with your analysis of the current implementation.
Thanks for the detailed follow-up 👍
On validation, we could expose the `masterSecret` property publicly in `DatabaseSecret` and perform a type-check in `addRotationMultiUser` for the `secret`.
I think it might be worth adding to provide clearer feedback to users invoking the method without providing the necessary parameters.

lpizzinidev (changes requested):

comcalvi:
@penniman26 This is a great change! Thank you for your patience as get to reviewing PRs. I'll merge this one once the changes from @lpizzinidev have been addressed.

Issue #28861, opened 2024-01-25
(aws-docdb): unsupported snapshot policy although documentation says it should be


kornicameister:
### Describe the bug

#28773 addressed the fact that it is impossible to spin up L2 DocumentDbCluster with retention policy said to Snaphost.
Although that is correct for security group inside (there's an escape hatch in DocumentDbCluster that sets cluster's retention policy onto the group) it turned out that `cdk synth` against `DocumentDbCluster` with a policy == Snaphost and seperately created security group also fails with an error that this time snaphost policy is not possible for `DocumentDbCluster` even though [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html) says so

### Expected Behavior

It is possible to configure DocumentDBCluster with Snapshot removal policy.

### Current Behavior

It is not possible to configure said cluster with Snapshot removal policy.

### Reproduction Steps

```typescript
#!/usr/bin/env node

import 'source-map-support/register';

import * as cdk from 'aws-cdk-lib';
import * as docdb from 'aws-cdk-lib/aws-docdb';
import * as ec2 from 'aws-cdk-lib/aws-ec2';

const app = new cdk.App();
const stack = new cdk.Stack(app, 'Stack');

const vpc = new ec2.Vpc(stack, 'VPC');
const securityGroup = new ec2.SecurityGroup(stack, 'SG', {
  vpc,
});
new docdb.DatabaseCluster(stack, 'Cluster', {
  vpc,
  removalPolicy: cdk.RemovalPolicy.SNAPSHOT,
  securityGroup,
  instanceType: ec2.InstanceType.of(
    ec2.InstanceClass.A1,
    ec2.InstanceSize.SMALL,
  ),
  masterUser: {
    username: 'admin',
    password: cdk.SecretValue.unsafePlainText('password'),
  },
});

app.synth();
```

### Possible Solution

@lpizzinidev mentioned that there might some sort of a list that needs to be updated to allow *sscce* to synthetize but I have no idea what that is and where that is.

### CDK CLI Version

2.121.1

### Node.js Version

20.8

### OS

MacOS sierra

### Language

TypeScript

### Language Version

5.3

GavinZZ:
[...]

PR #28868, opened 2024-01-25
chore(core): add doc db cluster as supported snapshot

GavinZZ:
According to [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html#aws-attribute-deletionpolicy-options), doc db cluster should support snapshot as deletion policy.

Add it. Not breaking change because users won't be able to use snapshot at all before the change (fail at synthesize step).

Closes https://github.com/aws/aws-cdk/issues/28861

----


GavinZZ:
Closing this PR as it's included https://github.com/aws/aws-cdk/pull/28798

PR #24313, opened 2023-02-23
fix(cloudwatch): math expressions incorrectly warn about search and metrics

ParkerVR:
Closes [#20136](https://github.com/aws/aws-cdk/issues/20136).

It is intended that all metric identifiers referenced in a MathExpression are included in the usingMetrics map and we will raise warnings if the customer does not follow this contract. 
However for SEARCH and METRICS queries, we can refer directly to metrics attribute values inside the query. Therefore we should not raise warnings.

Change made based on work done in https://github.com/aws/aws-cdk/commit/55108b969a94f671a492b4536d2ad9d13d11cf9d with regex extended to a few other expressions. Looks like integ requests will not be required based on that commit.

I had some firsthand experience getting thousands of this warning message after upgrading to CDK 2 and decided it would be easier to fix than suppress the excessive warnings.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #28870, opened 2024-01-25
fix(cloudwatch): Math:UnknownIdentifier warning for INSIGHT_RULE_METRIC

aws-sde:
The CloudWatch `MathExpression` class warns about identifiers missing from `usingMetrics` when `INSIGHT_RULE_METRIC` is used in the expression. It incorrectly parses the arguments to `INSIGHT_RULE_METRIC` as identifiers. When using `INSIGHT_RULE_METRIC`, I don't believe there is anything that needs to be added to `usingMetrics`.

This implementation follows a similar fix done for some other expressions here: https://github.com/aws/aws-cdk/pull/24313

----


aws-sde:
Exemption Request

kaizencc (approved):
Thanks, looks good to me

PR #28871, opened 2024-01-25
docs(lambda): log group docs not being rendered properly

longtv2222:
TSDoc for `logGroup` is not being rendered properly due to the use of `<` and `>`.

This affects Vscode:

Screenshot 2024-01-25 164442

And `aws-cdk-lib` docs as well: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.Function.html#loggroup

Screenshot 2024-01-25 164752

Fixed by escaping `<` and `>` properly

This is probably caused by `<` is interpreted as opening `JSX` tags by `MDX`.

After fix:
Screenshot 2024-01-25 165425

----


mrgrain (approved):
❤️

PR #28880, opened 2024-01-26
feat(applicationautoscaling): validate `evaluationPeriods` and `datapointsToAlarm` for step scaling policy

go-to-k:
This PR adds validations for `evaluationPeriods` and `datapointsToAlarm` for step scaling policy. It is based on [the PR for datapointsToAlarm in autoscaling step scaling policy](https://github.com/aws/aws-cdk/pull/28792).

Check the following cases:
- `evaluationPeriods` < 1
- `datapointsToAlarm` is set, and
  - `evaluationPeriods` is not set
  - `datapointsToAlarm` < 1
  - `evaluationPeriods` < `datapointsToAlarm`

These validations also consider whether those parameters are tokens or not.

----


go-to-k:
Exemption Request: This PR just adds validations, so the unit tests could cover.

packages/aws-cdk-lib/aws-applicationautoscaling/lib/step-scaling-policy.ts
```diff
@@ -117,8 +117,22 @@ export class StepScalingPolicy extends Construct {
...
     }
 
-    if (props.datapointsToAlarm !== undefined && props.datapointsToAlarm < 1) {
-      throw new RangeError(`datapointsToAlarm cannot be less than 1, got: ${props.datapointsToAlarm}`);
+    if (props.evaluationPeriods !== undefined && !cdk.Token.isUnresolved(props.evaluationPeriods) && props.evaluationPeriods < 1) {
+      throw new RangeError(`evaluationPeriods cannot be less than 1, got: ${props.evaluationPeriods}`);
+    }
+    if (props.datapointsToAlarm !== undefined) {
+      if (props.evaluationPeriods === undefined) {
+        throw new RangeError('evaluationPeriods must be set if datapointsToAlarm is set');
```
lpizzinidev, line 125:
```suggestion
        throw new Error('evaluationPeriods must be set if datapointsToAlarm is set');
```
Should be a simple `Error`.
In general, I don't see `RangeError`s being used much in the repo.
I'm wondering if it wouldn't just be better to use the `Error` class for all guards in this file as well 🤔

lpizzinidev (changes requested):
Thanks 👍
Just a small nit.
Also, I think we can consider this a `chore`.

packages/aws-cdk-lib/aws-applicationautoscaling/lib/step-scaling-policy.ts
go-to-k: OK, I have adapted the existing code as it is not the essence of this PR, but I will change everything to `Error`.

go-to-k:
@lpizzinidev 

Thanks for your review. I changed the `RangeError` and PR title.

lpizzinidev (approved):
Thanks 👍

kaizencc (approved):

PR #28887, opened 2024-01-26
chore: correct installLatestAwsSdkDefault flag docs

joshlartz:
Correcting docs about this flag as the implementation still defaults to true. 

https://github.com/aws/aws-cdk/blob/v2.123.0/packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/aws-custom-resource.ts#L471-L473

----

### DIFF
--- a/packages/aws-cdk-lib/cx-api/lib/features.ts
+++ b/packages/aws-cdk-lib/cx-api/lib/features.ts
@@ -647,7 +647,7 @@ export const FLAGS: Record<string, FlagInfo> = {
     compatibilityWithOldBehaviorMd: 'Set installLatestAwsSdk: true on all resources that need it.',
     introducedIn: { v2: '2.60.0' },
     recommendedValue: false,
-    defaults: { v2: true }
+    defaults: { v2: true },
   },
 
   //////////////////////////////////////////////////////////////////////


kylelaker (changes requested):
That markdown file will get overwritten the next time the package is built. If this documentation change is correct, I believe this needs to be changed by specifying `defaults: { v2: true }` in `features.ts`. I am not sure if that will have any other downstream effects.

The change would be somewhere around here:
https://github.com/aws/aws-cdk/blob/6a7a24afcc1ebebf71c267b890732a455e865cc8/packages/aws-cdk-lib/cx-api/lib/features.ts#L647-L649

This is the code that performs the generation of that markdown file:
https://github.com/aws/aws-cdk/blob/6a7a24afcc1ebebf71c267b890732a455e865cc8/packages/aws-cdk-lib/cx-api/build-tools/flag-report.ts#L95-L99

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -647,6 +647,7 @@ export const FLAGS: Record<string, FlagInfo> = {
     compatibilityWithOldBehaviorMd: 'Set installLatestAwsSdk: true on all resources that need it.',
     introducedIn: { v2: '2.60.0' },
     recommendedValue: false,
+    defaults: { v2: true }
```
kylelaker, line 650:
```suggestion
    defaults: { v2: true },
```

To fix the build error

kylelaker (commented):
Thanks! I think there should be some other files that are modified as a consequence of this if you were to build the `aws-cdk-lib` package. `CONTRIBUTING.md` has details on how to do that, but you probably would want to do the following:

1. In the base of the repo run `yarn install` to install all the CDK dependencies
2. Move into the `aws-cdk-lib` directory with `cd packages/aws-cdk-lib`
3. Build that package and its dependencies with `../../scripts/buildup`

From there, you can run tests locally with `yarn test`. The CodeBuild check on the PR will also run the full test suite after you push.

joshlartz:
@kylelaker You were right that there were unintended changes from this. It looks like there was an assumption built in that flags would always end up being true in v2. It is causing the "Flags with a different default in v2" section to be updated incorrectly. 

Example:
| Flag | Summary | Type | Since | v1 default | v2 default |
| ----- | ----- | ----- | ----- | ----- | ----- |
| [@aws-cdk/customresources:installLatestAwsSdkDefault](#aws-cdkcustomresourcesinstalllatestawssdkdefault) | Whether to install the latest SDK by default in AwsCustomResource | (default) |  | `false` | `true` |

Here is an example of a `cdk.json` file that restores v1 behavior for these flags:

```json
{
  "context": {
    "@aws-cdk/core:newStyleStackSynthesis": false,
    "@aws-cdk/core:stackRelativeExports": false,
    "@aws-cdk/aws-rds:lowercaseDbIdentifier": false,
    "@aws-cdk/aws-apigateway:usagePlanKeyOrderInsensitiveId": false,
    "@aws-cdk/aws-lambda:recognizeVersionProps": false,
    "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021": false,
    "@aws-cdk/customresources:installLatestAwsSdkDefault": false
  }
}
```

joshlartz:
Not stale, just waiting on an answer still :( 
@kylelaker

PR #28898, opened 2024-01-27
fix(events-targets): ecs:TagResource permission

### LINKED ISSUE #28854: (aws-events-targets): EcsTask target with tags does not get ecs:TagResource permission added to role
### Describe the bug

When using the [`EcsTask` EventBridge target](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_events_targets-readme.html#run-an-ecs-task) and providing `tags`, the auto-generated role does not get the `ecs:TagResource` IAM permission.

This causes problems when the AWS account has the [`tagResourceAuthorization` setting](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#tag-resources-setting) enabled. According to an email I recently received from Amazon, this setting will be enabled by default on 29 March 2024:

> March 29, 2024 - Tagging Authorization will be turned on for all AWS accounts. The account level setting will no longer be
used and will be removed from the ECS Account Settings page in the AWS Console.

Once the setting is enabled by default on all accounts, these event targets will start failing.

### Expected Behavior

When I provide a list of `tags` on the EcsTask event target, it should automatically grant the `ecs:TagResource` permission for the specified tags.

### Current Behavior

The role does not get the `ecs:TagResource` permission, which causes the event invocation to fail.

The event fails with:

> User: arn:aws:sts::ACCOUNT:assumed-role/TravelBlogDeploymentPipel-StaticWordpressEcsTaskTas-MSjLKMJsnohL/b07e937039c03003adf70077b02829a0 is not authorized to perform: ecs:TagResource on resource: arn:aws:ecs:us-west-2:ACCOUNT:task/CLUSTER/* because no identity-based policy allows the ecs:TagResource action

### Reproduction Steps

Consider a basic event rule, as specified [in the CDK documentation](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_events_targets-readme.html#tagging-tasks):

```ts
import * as ecs from 'aws-cdk-lib/aws-ecs';

declare const cluster: ecs.ICluster;
declare const taskDefinition: ecs.TaskDefinition;

const rule = new events.Rule(this, 'Rule', {
  schedule: events.Schedule.rate(cdk.Duration.hours(1)),
});

rule.addTarget(
  new targets.EcsTask({
    cluster: cluster,
    taskDefinition: taskDefinition,
    propagateTags: ecs.PropagatedTagSource.TASK_DEFINITION,
    tags: [
      {
        key: 'my-tag',
        value: 'my-tag-value',
      },
    ],
  }),
);
```

This code produces an IAM role that does not allow `ecs:TagResource`, as you can see in the source code: https://github.com/aws/aws-cdk/blob/280135567956b78fedd6a528cff0adda6171ec33/packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts#L268-L296

The event triggers without errors when `tagResourceAuthorization` is disabled. However, when you enable `tagResourceAuthorization` via:

```shell
aws ecs put-account-setting-default --name tagResourceAuthorization --value on
```

The task will start to fail because of the missing permission.

### Possible Solution

The `createEventRolePolicyStatements` method should be updated. If `tags` are present the IAM policy should include `ecs:TagResource` for the specified tags. The [docs](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonelasticcontainerservice.html) should be reviewed and the appropriate restrictive conditions should be applied.

_No response_

### CDK CLI Version

2.123.0

### Framework Version

_No response_

### Node.js Version

20.11.0

### OS

MacOS

### Language

TypeScript

### Language Version

_No response_

### Other information

I'm unsure if this is needed if `propagateTags` is specified. More digging in the docs on `tagResourceAuthorization` is needed.

msambol:
I enabled the following:

`aws ecs put-account-setting-default --name tagResourceAuthorization --value on`

And then confirmed the task completes successfully.

Closes #28854.

----


packages/@aws-cdk-testing/framework-integ/test/aws-events-targets/test/ecs/image-simple/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM --platform=linux/amd64 public.ecr.aws/docker/library/python:3.12
+ADD index.py .
+CMD [ "python3", "./index.py" ]
```
msambol, line 3: this image needed to be changed. it was using a lambda image which resulted in the following:
`entrypoint requires the handler name to be the first argument`.

packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts
```diff
@@ -266,13 +266,20 @@ export class EcsTask implements events.IRuleTarget {
...
+        actions: ['ecs:RunTask'],
+        resources: [this.taskDefinition.taskDefinitionArn],
+        conditions: {
+          ArnEquals: { 'ecs:cluster': this.cluster.clusterArn },
+        },
+      }),
+      new iam.PolicyStatement({
+        actions: ['ecs:TagResource'],
+        // eslint-disable-next-line @aws-cdk/no-literal-partition
+        resources: [`arn:aws:ecs:*:*:task/${this.cluster.clusterName}/*`],
```
msambol, line 280: open to ideas on this. I don't think there is a good way to get region? hence the `*` for region in arn.

msambol:
@ConnorRobertson Any feedback on this one?

packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts
msambol: I actually think no Region may be the way to go here.. since it's a role/policy.

paulhcsun: `region` seems to be accessible from `this.cluster.env.region`

image

paulhcsun (commented):
Hi @msambol, apologies for the delay to get around to this. The changes look good to me, just left one comment about accessing `region`. 

Btw I just realized you're the one behind the Data structures and algorithms in X minutes videos, I'm a big fan!

msambol:
> Hi @msambol, apologies for the delay to get around to this. The changes look good to me, just left one comment about accessing `region`.
> 
> Btw I just realized you're the one behind the Data structures and algorithms in X minutes videos, I'm a big fan!

Thank you for the kind words! 😄 Much appreciated.

Good catch on cluster region. I updated the PR.

packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts
```diff
@@ -277,13 +277,20 @@ export class EcsTask implements events.IRuleTarget {
...
+        resources: [this.taskDefinition.taskDefinitionArn],
+        conditions: {
+          ArnEquals: { 'ecs:cluster': this.cluster.clusterArn },
+        },
+      }),
+      new iam.PolicyStatement({
+        actions: ['ecs:TagResource'],
+        // eslint-disable-next-line @aws-cdk/no-literal-partition
+        resources: [`arn:aws:ecs:${this.cluster.env.region}:*:task/${this.cluster.clusterName}/*`],
+      }),
```
aaythapa, line 292: I'm wondering if this breaks compatibility? I saw in the issue that `March 29, 2024 - Tagging Authorization will be turned on for all AWS accounts.` so maybe after March 29th this should be standard but till then should there be condition here that checks if the tag property is set then adds the policy?

aaythapa (commented):
Hi @msambol, thanks for the code change (also a big fan 😁), just a small question but otherwise looks good.

packages/aws-cdk-lib/aws-events-targets/lib/ecs-task.ts
msambol: Thank you! 😄 

This won't break compatibility, only adding a permission to a role.

aaythapa: Talked it over with paul and we're ok with merging, thanks again for the change

aaythapa (approved):

paulhcsun (approved):

PR #28899, opened 2024-01-27
feat(cognito): specifying the trigger event version in UserPool.addTrigger()

### LINKED ISSUE #28683: aws_cognito: lambd trigger version
### Describe the feature

Now that cognito officially supports modification of access tokens by using lambda version 2 it would be great if the addTrigger functionality could support this. If this is already supported but undocumented maybe just an update to the docs are needed

https://aws.amazon.com/about-aws/whats-new/2023/12/amazon-cognito-user-pools-customize-access-tokens/

https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html

From the second link:
To support access token customization in a pre token generation Lambda trigger

Generate a [CreateUserPool](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateUserPool.html) or [UpdateUserPool](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_UpdateUserPool.html) API request. You must specify a value for all parameters that you don't want set to a default value. For more information, see [Updating user pool configuration](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pool-updating.html).

Include the following content in the LambdaVersion parameter of your request. A LambdaVersion value of V2_0 causes your user pool to add parameters for access token customization. To invoke a specific function version, use a Lambda function ARN with a function version as the value of LambdaArn.

```json
"PreTokenGenerationConfig": { 
   "LambdaArn": "arn:aws:lambda:us-west-2:123456789012:function:MyFunction",
   "LambdaVersion": "V2_0"
}
```

this is also supported in cloudformation

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cognito-userpool-pretokengenerationconfig.html

```yaml
  LambdaArn: String
  LambdaVersion: String
```

### Use Case

Instead of writing custom code to update this or call the cfn resource and add logic for the updates it would make a lot of sense if this was supported natively with cdk addTrigger function

### Proposed Solution

Update add trigger to support input of version string like
```javascript
(method) UserPool.addTrigger(operation: cdk.aws_cognito.UserPoolOperation, fn: cdk.aws_lambda.IFunction, lambdaVersion: string): void
```
It should default to V1_0 if not set and it should only be evaluated if the operation ins PRE_TOKEN_GENERATION

```javascript
userPool.addTrigger(cognito.UserPoolOperation.PRE_TOKEN_GENERATION, lambda.Function.fromFunctionArn(
      this,
      'PreTokenGenerationLambda',
      `arn:aws:lambda:${props.env?.region}:${props.env?.account}:function:MyFunction`
    ),"V2_0");
```

An alternate approach would be to do this on the userpool trigger config in the userpool
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cognito.UserPoolTriggers.html
ie change the
preTokenGeneration? from iFunction to be a construct with iFunction and version string

```javascript
(property) UserPoolTriggers.preTokenGenerationConfig?: cdk.aws_cognito.PreTokenGenerationConfig | undefined 
```
This seem somewhat prepared because you can set the value to preTokenGenerationConfig but it only expects a IFunction
```javascript
(property) preTokenGenerationConfig: cdk.aws_lambda.IFunction
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.119.0

macOS 14.2.1 (23C71)

badmintoncryer:
I have added a `lambdaVersion` to the `UserPool.addTrigger()`. 
This is in response to the [support for V2.0 trigger event in preTokenGeneration](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html).

```ts
declare const userpool: cognito.UserPool;
declare const preTokenGenerationFn: lambda.Function;

userpool.addTrigger(cognito.UserPoolOperation.PRE_TOKEN_GENERATION_CONFIG, preTokenGenerationFn, LambdaVersion.V2_0);
```

In #28683, apart from the current implementation approach, there was also a proposal to add `lambdaVersion` to `UserPoolProps.lambdaTrigger`. However, it was not adopted as it would result in a breaking change.

Closes #28683

----


packages/aws-cdk-lib/aws-cognito/lib/user-pool.ts
```diff
@@ -994,10 +1020,13 @@ export class UserPool extends UserPoolBase {
    * Add a lambda trigger to a user pool operation
    * @see https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools-working-with-aws-lambda-triggers.html
    */
-  public addTrigger(operation: UserPoolOperation, fn: lambda.IFunction): void {
+  public addTrigger(operation: UserPoolOperation, fn: lambda.IFunction, lambdaVersion?: LambdaVersion): void {
     if (operation.operationName in this.triggers) {
       throw new Error(`A trigger for the operation ${operation.operationName} already exists.`);
     }
+    if (operation !== UserPoolOperation.PRE_TOKEN_GENERATION_CONFIG && lambdaVersion === LambdaVersion.V2_0) {
```
aaythapa, line 1027: Do we have to check if `lambdaVersion` exists as well?

packages/aws-cdk-lib/aws-cognito/README.md
```diff
@@ -493,6 +493,15 @@ userpool.addTrigger(cognito.UserPoolOperation.USER_MIGRATION, new lambda.Functio
 }));
 ```
 
+Additionally, only the pre token generation Lambda trigger supports the trigger event V2.0:
```
aaythapa, line 496:
```suggestion
Additionally, only the pre token generation Lambda trigger supports trigger events with lambda version V2.0:
```

packages/aws-cdk-lib/aws-cognito/lib/user-pool.ts
```diff
@@ -994,10 +1020,13 @@ export class UserPool extends UserPoolBase {
...
+      throw new Error('Only the `preTokenGenerationConfig` operation supports V2_0 lambda version.');
```
aaythapa, line 1028:
```suggestion
      throw new Error('Only the `PRE_TOKEN_GENERATION_CONFIG` operation supports V2_0 lambda version.');
```
IMO, having the actual enum value in the error msg will help user know right away which operation to use. Open to discussion if you think different.

aaythapa (commented):
Thank you for your contribution! Overall looks great, just some nitpicks

packages/aws-cdk-lib/aws-cognito/lib/user-pool.ts
badmintoncryer: I believe the only scenario where I want to raise an error is when the lambdaVersion is set to V2 and the operation is PRE_TOKEN_GENERATION. 
Since lambdaVersion defaults to V1 when undefined, I think checking for that is unnecessary. What do you think?

aaythapa: my only concern was the check would throw an error saying the lambdaVersion doesn't exist (if the user left it empty). Did a quick check myself and realized thats not the case so i think its fine to leave as is. Might be good to have a test for it tho (if it doesn't exist already)

badmintoncryer: Sure! I've added unit test.
https://github.com/aws/aws-cdk/pull/28899/files#diff-2c630fbb5d50f4f6d047a873debf0ab809aae98b8e00d9ab8d86c8ee4f3cdaf2R511

aaythapa: beautiful

aaythapa (commented):
🚀

aaythapa (approved):

PR #28902, opened 2024-01-28
fix(ec2): deploying an isolated subnet fails when `ipv6AssignAddressOnCreation` is set to true

### LINKED ISSUE #28843: (ec2): ipv6AssignAddressOnCreation(true) on isolated subnet configuration produces an error
### Describe the bug

Creating isolated subnet configuration for a new VPC with `ipv6AssignAddressOnCreation(true)`, causes an error during first deployment. 

### Expected Behavior

I'm not entirely sure what is expected behaviour here. I would probably expect this to work, even though IPV6 addresses are public by default, there still won't be a route created to internet. My understanding of this property, would be that I won't have to set similar property when creating EC2 instances that need IPV6 address on isolated subnet.
Curiously, if you deploy without this property first, then add it later, there's no difference in the stack template. Therefore, I'm not sure what this property actually supposed to do.

### Current Behavior

The error:
```
6:39:48 PM | CREATE_FAILED        | AWS::EC2::Subnet                      | vpcisolatedSubnet1Subnet06BBE51F
Template error: Fn::Select  cannot select nonexistent value at index 0
```

### Reproduction Steps

Below code causes deployment error on first deploy:
```
var subnet = SubnetConfiguration
  .builder()
  .subnetType(SubnetType.PRIVATE_ISOLATED)
  .name("isolated")
  .cidrMask(21)
  .ipv6AssignAddressOnCreation(true)
  .build();
var vpc = Vpc
  .Builder
  .create(stack, "vpc")
  .vpcName("some-vpc")
  .ipProtocol(IpProtocol.DUAL_STACK)
  .ipv6Addresses(Ipv6Addresses.amazonProvided())
  .ipAddresses(IpAddresses.cidr("172.16.0.0/16"))
  .natGateways(0)
  .subnetConfiguration(Collections.singletonList(subnet))
  .build();
```

### Possible Solution

It works if the property is not specified. I believe the cause is a race condition between subnet and ipv6 cidr creation, since dependency is created for other types, but not for isolated subnet: https://github.com/aws/aws-cdk/blob/v2.122.0/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts#L1792

_No response_

### CDK CLI Version

2.121.1 (build d86bb1a)

### Framework Version

2.121.1

### Node.js Version

18.17.0

### OS

OSX 14.2.1

### Language

Java

### Language Version

OpenJDK Runtime Environment Temurin-17.0.10+7 (build 17.0.10+7)

### Other information

_No response_

badmintoncryer:
This PR resolves the issue where deploying an isolated subnet with `ipv6AssignAddressOnCreation` enabled fails. 

### example
```ts
new Vpc(stack, 'TheVPC', {
        ipProtocol: IpProtocol.DUAL_STACK,
        subnetConfiguration: [
          {
            subnetType: testData.subnetType,
            name: 'subnetName',
            ipv6AssignAddressOnCreation: true,
          },
        ],
      });
```

### error
```sh
6:39:48 PM | CREATE_FAILED        | AWS::EC2::Subnet                      | vpcisolatedSubnet1Subnet06BBE51F
Template error: Fn::Select  cannot select nonexistent value at index 0
```

### solution
A dependency on the CidrBlock has been added [as discussed in issue](https://github.com/aws/aws-cdk/issues/28843#issuecomment-1909006632).

```ts
      (this.isolatedSubnets as PrivateSubnet[]).forEach((isolatedSubnet) => {
        if (this.ipv6CidrBlock !== undefined) {
          isolatedSubnet.node.addDependency(this.ipv6CidrBlock);
        }
      });
```

## Question
This modification results in the failure of existing integration tests.
I don't consider this change to be a breaking one, so I went ahead and updated the snapshot. Is that okay?
```sh
  CHANGED    aws-ec2/test/integ.vpc-dual-stack-ec2 0.776s
      Resources
[~] AWS::EC2::Subnet Ip6VpcDualStackPrivateSubnet1Subnet842B7F4C 
 └─ [+] DependsOn
     └─ ["Ip6VpcDualStackipv6cidr40BE830A"]
[~] AWS::EC2::RouteTable Ip6VpcDualStackPrivateSubnet1RouteTable5326D239 
 └─ [+] DependsOn
     └─ ["Ip6VpcDualStackipv6cidr40BE830A"]
[~] AWS::EC2::SubnetRouteTableAssociation Ip6VpcDualStackPrivateSubnet1RouteTableAssociationF1C10B6A 
 └─ [+] DependsOn
     └─ ["Ip6VpcDualStackipv6cidr40BE830A"]
[~] AWS::EC2::Subnet Ip6VpcDualStackPrivateSubnet2SubnetEB493489 
 └─ [+] DependsOn
     └─ ["Ip6VpcDualStackipv6cidr40BE830A"]
[~] AWS::EC2::RouteTable Ip6VpcDualStackPrivateSubnet2RouteTable56BF517C 
 └─ [+] DependsOn
     └─ ["Ip6VpcDualStackipv6cidr40BE830A"]
[~] AWS::EC2::SubnetRouteTableAssociation Ip6VpcDualStackPrivateSubnet2RouteTableAssociationD37A3D3D 
 └─ [+] DependsOn
     └─ ["Ip6VpcDualStackipv6cidr40BE830A"]
```

Closes #28843

----


ConnorRobertson:
Yeah I think this looks good to me. The added dependency shouldn't be a breaking change in this case I think as well.

ConnorRobertson (approved):

PR #28903, opened 2024-01-28
feat(app-staging-synthesizer-alpha): encryption type for staging bucket

### LINKED ISSUE #28815: (app-staging-synthesizer-alpha): Allow using S3 Managed KMS Key for Bucket
### Describe the feature

Currently, the `AppStagingSythesizer` always creates a KMS key to encrypt the staging bucket: https://github.com/aws/aws-cdk/blob/1b6be8b94b8620911cf8418f0627227045c47bf9/packages/@aws-cdk/app-staging-synthesizer-alpha/lib/default-staging-stack.ts#L360-L378

It would be nice if we could opt into using the `SSE-S3` keys instead.

### Use Case

I'd like to start making more frequent use of the `AppStagingSynthesizer`. However, by forcing the use of a custom KMS key, each app using this synthesizer incurs a $1/month fee for the key.

### Proposed Solution

The `DefaultStackSynthesizer` does not specify an encryption key and, thus, uses the `SSE-S3` managed key by default. It feels like `AppStagingSynthesizer` should do the same thing by default. IMO, the custom KMS key feels like it should be an opt-in behavior for those who want it.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.122.0

MacOS

msambol:
Closes #28815.

----

### DIFF
--- a/.github/workflows/repo-metrics.yml
+++ b/.github/workflows/repo-metrics.yml
@@ -36,7 +36,7 @@ jobs:
         SEARCH_QUERY: 'repo:aws/aws-cdk is:issue created:${{ env.last_month }} -reason:"not planned"'
 
     - name: Create report for issues
-      uses: peter-evans/create-issue-from-file@v4
+      uses: peter-evans/create-issue-from-file@v5
       with:
         title: Monthly issue metrics report
         token: ${{ secrets.GITHUB_TOKEN }}
@@ -50,7 +50,7 @@ jobs:
         SEARCH_QUERY: 'repo:aws/aws-cdk is:pr created:${{ env.last_month }} -is:draft'
 
     - name: Create report for PRs
-      uses: peter-evans/create-issue-from-file@v4
+      uses: peter-evans/create-issue-from-file@v5
       with:
         title: Monthly PR metrics report
         token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/request-cli-integ-test.yml
+++ b/.github/workflows/request-cli-integ-test.yml
@@ -19,7 +19,7 @@ jobs:
           persist-credentials: false
       - name: Find changed cli files
         id: changed-cli-files
-        uses: tj-actions/changed-files@ae82ed4ae04587b665efad2f206578aa6f0e8539
+        uses: tj-actions/changed-files@90a06d6ba9543371ab4df8eeca0be07ca6054959
         with:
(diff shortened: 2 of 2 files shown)


packages/@aws-cdk/app-staging-synthesizer-alpha/lib/default-staging-stack.ts
```diff
@@ -358,7 +367,15 @@ export class DefaultStagingStack extends Stack implements IStagingResources {
     }
 
     this.ensureFileRole();
-    const key = this.createBucketKey();
+
+    let key = undefined;
+    if (this.stagingBucketEncryption === s3.BucketEncryption.KMS || this.stagingBucketEncryption === undefined) {
+      if (this.stagingBucketEncryption === undefined) {
+        // default is KMS for backwards compatibility
+        this.stagingBucketEncryption = s3.BucketEncryption.KMS;
```
kaizencc, line 375: Thanks @msambol. Give me a sec to figure this out, we don't necessarily have to honor backwards compatibility but I'm not recalling why we went with KMS as the default in the first place. Let me ask around

kaizencc: Ok -- we are using KMS because that is officially AWS's best practice. So we're going to keep that as default as well.

packages/@aws-cdk/app-staging-synthesizer-alpha/lib/default-staging-stack.ts
```diff
@@ -358,7 +367,15 @@ export class DefaultStagingStack extends Stack implements IStagingResources {
     }
 
     this.ensureFileRole();
-    const key = this.createBucketKey();
+
+    let key = undefined;
+    if (this.stagingBucketEncryption === s3.BucketEncryption.KMS || this.stagingBucketEncryption === undefined) {
+      if (this.stagingBucketEncryption === undefined) {
+        // default is KMS for backwards compatibility
```
kaizencc, line 374:
```suggestion
        // default is KMS as an AWS best practice, and for backwards compatibility
```

kaizencc:
@Mergifyio refresh

kaizencc (approved):

blimmer:
@kaizencc from your comment above, it sounds like the default should be managed encryption, allowing an opt-in to KMS. Should we take advantage of this package's Alpha status and change the default now?

PR #28913, opened 2024-01-29
feat(efs): create read-only file system for use as a destination of replication

### LINKED ISSUE #28912: efs: support for FileSystemProtection property for file system
### Describe the feature

There is `FileSystemProtection.ReplicationOverwriteProtection` parameter for replication configuration in the EFS file system. [CloudFormation supports this setting](https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-properties-efs-filesystem-filesystemprotection.html), so I want to enable its configuration through the L2 constructor.

### Use Case

When creating a replication destination filesystem, it is necessary to set the `ReplicationOverwriteProtection` configuration to `DISABLED`, but this cannot be executed from the L2 construct.

### Proposed Solution

Since this parameter can only take two values, 'ENABLED' or 'DISABLED', I plan to convert a boolean type argument into the appropriate string within the L2 construct.

```ts
new FileSystem(stack, 'EfsFileSystem', {
    vpc,
    replicationOverwriteProtection: false, // 'DISABLED'
  });
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.123.0

irrelevant

badmintoncryer:
In this PR, I have added the `replicationOverwriteProtection` argument to the L2 construct. This enables the creation of a filesystem that can be used as a destination for replication.

```ts
const fileSystem = new efs.FileSystem(this, 'MyEfsFileSystem', {
  vpc: new ec2.Vpc(this, 'VPC'),
  replicationOverwriteProtection: true, // Set to false if you want to create a read-only file system for use as a replication destination
});
```

Closes #28912.

----


packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -548,6 +558,10 @@ export class FileSystem extends FileSystemBase {
 
     const oneZoneAzName = props.vpc.availabilityZones[0];
 
+    const fileSystemProtection = props.replicationOverwriteProtection !== undefined ? {
+      replicationOverwriteProtection: props.replicationOverwriteProtection ? 'ENABLED' : 'DISABLED',
+    } : undefined;
```
aaythapa, line 563: with enums theres no need for this check and it can be a simple passthrough 👍

aaythapa (commented):
Thank you once again for your contribution! 

I'd suggest we use enums in this case. I understand that there are other CFN properties that take either ENABLED or DISABLED but in CDK they're boolean so this change would be consistent with that. 

But in the future if CFN adds more valid values then we'll regret making this a boolean. Plus if its enums we can do a simple pass through and don't have to do any extra meddling in the code.

(I talked this over with the team beforehand so I won't make you revert it this time 😅)

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
badmintoncryer: Unfortunately, it seems necessary to continue checks because an object or undefined must be passed here. 
However, the ternary operator that replaces with the strings ENABLED, DISABLED seems to be removable!

badmintoncryer:
@aaythapa 

Thank you for the review. As you suggested, I have revised it to use the enum format. 
Could you please check it again?

aaythapa (approved):

badmintoncryer:
@aaythapa 
I have resolved the failure in the CI. Please check again!

aaythapa (approved):

PR #28928, opened 2024-01-30
feat(lambda):  Add `Ipv6AllowedforDualStack` support for Lambda function

Vandita2020:
This PR adds support to configure IPv6 connectivity for lambda function by adding a single field to the VPC configuration in CDK’s Lambda Console to enable IPv6 dual stack support.

The `Ipv6AllowedForDualStack` field is a new property that allows Lambda functions to specify whether the IPv6 traffic should be allowed when using dual-stack VPCs.

- Support IPv6 connectivity for Lambda function by setting "ipv6AllowedForDualStack : true". This value will be converted to VpcConfig: {Ipv6AllowedForDualStack: true,} in template.


packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -1075,6 +1075,49 @@ const fn = new lambda.Function(this, 'MyLambda', {
...
+import * as ec2 from 'aws-cdk-lib/aws-ec2';
+
+const natProvider = ec2.NatProvider.gateway();
+
+// create dual-stack VPC
+const vpc = new ec2.Vpc(this, 'DualStackProtocolVpc', {
+  ipProtocol: ec2.IpProtocol.DUAL_STACK,
+  subnetConfiguration: [
+    {
+      name: 'Pub1',
```
xazhao, line 1094: nit: can we use `Pub1` and `Private1` in this case? They're more clear to customers.

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -1497,10 +1510,17 @@ Environment variables can be marked for removal when used in Lambda@Edge by sett
     // List can't be empty here, if we got this far you intended to put your Lambda
     // in subnets. We're going to guarantee that we get the nice error message by
     // making VpcNetwork do the selection again.
-
-    return {
-      subnetIds: selectedSubnets.subnetIds,
-      securityGroupIds: securityGroups.map(sg => sg.securityGroupId),
+    if (props.ipv6AllowedForDualStack) {
+      return {
+        ipv6AllowedForDualStack: ipv6AllowedForDualStack,
```
xazhao, line 1515: What happens if we return `ipv6AllowedForDualStack: false`? Just thinking if we need this if statement.

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3832,6 +3832,58 @@ describe('VPC configuration', () => {
...
+      vpc,
+      allowAllOutbound: true,
+      allowAllIpv6Outbound: true,
+    });
+    new lambda.Function(stack, 'MyLambda', {
+      vpc: vpc,
+      code: new lambda.InlineCode('foo'),
+      handler: 'index.handler',
+      runtime: lambda.Runtime.PYTHON_3_9,
+      ipv6AllowedForDualStack: true,
```
xazhao, line 3862: Could we add a tiny test with ipv6AllowedForDualStack as false to make sure it works as expect?

packages/@aws-cdk-testing/framework-integ/test/aws-lambda/test/integ.lambda_with_dualStack_vpc.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const app = new App();
+const stack = new Stack(app, 'DualStackVpc');
+
+const natProvider = ec2.NatProvider.gateway();
+const vpc = new ec2.Vpc(stack, 'DualStackProtocolVpc', {
+  ipProtocol: ec2.IpProtocol.DUAL_STACK,
+  subnetConfiguration: [
+    {
+      name: 'Pub1',
```
xazhao, line 14: Same as readme, let's make it `Public` or `Private1` so it's more clear to customers.

packages/aws-cdk-lib/aws-lambda/lib/function.ts
Vandita2020: The default value of `ipvAllowedForDualStack` is false, so even if we don't set it to the construct, it'll stay false. We have made the decision of adding `ipv6AllowedForDualStack` to return config only if it is set to true, because when it is not set, it was breaking other integration test, where VpcConfig is being used. As this property is set for dualStack VPC only and the other test might not be using dualStack VPC.

packages/aws-cdk-lib/aws-lambda/README.md
Vandita2020: Done

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
Vandita2020: Added new test to check when `ipv6AllowedForDualStack` is set to false

packages/@aws-cdk-testing/framework-integ/test/aws-lambda/test/integ.lambda_with_dualStack_vpc.ts
Vandita2020: Done

xazhao (approved):

PR #28942, opened 2024-01-31
feat(lambda): support enums in loggingFormat 

ConnorRobertson:
# Issue
The issue is that LogFormat is a String so it doesn't allow the enum LogFormat. 
# Solution
Created a new enum for the LoggingFormat and added testing. 
So the solution sets these values as potential environment variables.
The main difference is LoggingFormat is assigned to an enum instead of a string.
# Important Design Decisions
This is so that an enum could be used for LoggingFormat without breaking JSII target languages. Some background information is in this pr https://github.com/aws/aws-cdk/pull/28127. Was a recommended solution here. https://github.com/aws/aws-cdk/pull/28127

Remember to follow the [CONTRIBUTING GUIDE] and [DESIGN GUIDELINES] for any
code you submit.

[CONTRIBUTING GUIDE]: https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md
[DESIGN GUIDELINES]: https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md

Closes https://github.com/aws/aws-cdk/issues/28114.

----


packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -501,6 +519,12 @@ export interface FunctionOptions extends EventInvokeConfigOptions {
    */
   readonly logFormat?: string;
```
ConnorRobertson, line 520: Here is where it is not an enum.

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -501,6 +519,12 @@ export interface FunctionOptions extends EventInvokeConfigOptions {
...
 
+  /**
+   * Sets the loggingFormat for the function.
+   * @default Text format
+   */
+  readonly loggingFormat?: LoggingFormat;
```
ConnorRobertson, line 526: Here is the enum change.

paulhcsun (approved):

PR #28973, opened 2024-02-02
fix(lambda): LoggingConfig default error

ConnorRobertson:
### Issue # (if applicable)

Fixes bug in LoggingFormat `@default`

### Reason for this change
Incorrect
https://github.com/aws/aws-cdk/issues/28114 related issue

### Description of changes
Not much to describe here. 

Shouldn't need any as there is integ tests from this pr
https://github.com/aws/aws-cdk/pull/28942

### Checklist

BREAKING CHANGE: changing the default type of `lambda.loggingFormat`. Previously it was a string `"Text format"`, not it is an enum `LoggingFormat.TEXT`.


packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -521,7 +521,7 @@ export interface FunctionOptions extends EventInvokeConfigOptions {
 
   /**
    * Sets the loggingFormat for the function.
-   * @default Text format
+   * @default "Text"
```
kaizencc, line 524:
```suggestion
   * @default LoggingFormat.TEXT
```

kaizencc: wouldn't this be the default?

ConnorRobertson: I think you are right.

paulhcsun (approved):

PR #28977, opened 2024-02-04
fix(app-staging-synthesizer-alpha): destroy KMS key on stack delete

blimmer:
### Reason for this change

@moltar called out that the KMS encryption keys used for the App Staging Synthesizer are retained when the stack is deleted. Since KMS keys are $1/month, we should clean these up when the stack is deleted.

### Description of changes

Updated the retention policy from the default (`RETAIN`) to `DESTROY`.

Updated integration tests.

### Checklist

----


blimmer:
Exemption Request - since this is just changing the retention policy, I don't think we need another test for this. The existing integration test snapshot update should suffice.

moltar (approved):

blimmer:
Exemption request

PR #28978, opened 2024-02-04
feat(app-staging-synthesizer-alpha): require passing `stagingBucketEncryption` and note that we intend to default to `S3_MANAGED` in the future

blimmer:
### Issue # (if applicable)

Relates to #28815

### Reason for this change

The App Staging Synthesizer is great - I've moved to using it for most of my stacks. However, the current default uses a Customer-Managed KMS key, which costs $1/month.

The default synthesizer bucket uses SSE-S3 encryption by default. This is nice because users do not incur additional fees for a KMS key.

In my opinion, SSE-S3 is good enough for most people. If folks need additional security, they should opt-in to SSE-KMS, which they can do via the `stagingBucketEncryption` property @msambol introduced with #28903.

### Description of changes

With guidance from @kaizencc [below](https://github.com/aws/aws-cdk/pull/28978#issuecomment-1930007176), this PR makes `stagingBucketEncryption` a required property, with a user-facing note that we intend to default to `S3_MANAGED` as the module is stablized.

Updated unit tests.

### Checklist

BREAKING CHANGE: `stagingBucketEncryption` property is now required. For existing apps, specify `BucketEncryption.KMS` to retain existing behavior. For new apps, choose the bucket encryption that makes most sense for your use case. `BucketEncryption.S3_MANAGED` is available and is intended to be the default when this module is stabilized.

----


kaizencc (commented):
Hi @blimmer! The reason why we didn't do this in #28815 is:

- Apparently, AWS best practice is to use KMS keys
- The migration story for people who previously used the default KMS style is murky. It looks like if you change the encryption of a bucket, only new items in the bucket will have the new encryption type, leaving the other objects reliant on the KMS key we just removed.

For the second reason, I think the best we can do is a feature flag. For the first reason, I'm not sure if we even want to update the default and go against "best practice"

blimmer:
Hey @kaizencc - thanks for your response.

> Apparently, AWS best practice is to use KMS keys

Could you tell me more about this? On the [S3 Security Best Practices page](https://docs.aws.amazon.com/AmazonS3/latest/userguide/security-best-practices.html). I see them discussing the different options, but I don't see KMS as a strong preference.

I only use a separate KMS key for a bucket if the data is really sensitive. CDK assets don't feel particularly sensitive to me, and it's $12/year/app staging application.

If it really was an important best practice, shouldn't the default CDK Synthesizer also use a custom KMS key?

> The migration story for people who previously used the default KMS style is murky. It looks like if you change the encryption of a bucket, only new items in the bucket will have the new encryption type, leaving the other objects reliant on the KMS key we just removed.

I agree that this isn't ideal, and had trouble writing a concise fix for this in my `BREAKING CHANGES` section.

That said, I feel fairly strongly that this is the direction we should go. The extra cost for the KMS key doesn't feel worth it to me. And it feels like the best time to make this breaking change is during the Alpha phase, where we warn people that we can and will break things at any point.

If they don't want to deal with the re-encryption of the bucket, they can simply set `stagingBucketEncryption` to `BucketEncryption.KMS` and everything remains the same.

What do you think?

kaizencc:
Hi @blimmer, here's what we talked about internally.

While we do not promise stability in alpha modules, users still expect it. Therefore, its a nonstarter to flip the default of a property with no clear migration path.

We don't want to add a feature flag to an alpha module, because that is a can of worms we don't want to open. Feature flags are for stable modules that require backwards compatibility.

So, here's the proposal I have for you.

We make the `stagingBucketEncryption` property required. A breaking change, but one that does not involve behavioral changes. Most people can upgrade by simply supplying `stagingBucketEncryption: s3.Encryption.KMS` and see no diff. New users can grapple with which encryption method best suits their needs. We then add a giant `FIXME` to the source code that details that when stabilizing this module, we want to make the property optional again with `S3_MANAGED` as default.

Let me know if that sounds reasonable to you!

packages/@aws-cdk/app-staging-synthesizer-alpha/test/app-staging-synthesizer.test.ts
```diff
@@ -16,7 +16,7 @@ describe(AppStagingSynthesizer, () => {
 
   beforeEach(() => {
     app = new App({
-      defaultStackSynthesizer: AppStagingSynthesizer.defaultResources({ appId: APP_ID }),
+      defaultStackSynthesizer: AppStagingSynthesizer.defaultResources({ appId: APP_ID, stagingBucketEncryption: BucketEncryption.S3_MANAGED }),
```
blimmer, line 19: For most of these tests, I used `S3_MANAGED`, since we intend to make this the default in the future.

packages/@aws-cdk/app-staging-synthesizer-alpha/test/integ.synth-default-resources.ts
```diff
@@ -17,6 +18,7 @@ const app = new App({
 const stack = new Stack(app, 'synthesize-default-resources', {
   synthesizer: AppStagingSynthesizer.defaultResources({
     appId: APP_ID_MAX, // this has implications on the overall template size
+    stagingBucketEncryption: BucketEncryption.KMS,
```
blimmer, line 21: I left this as KMS to avoid changing the integration test. @msambol already added an S3_MANAGED integration test with #28903.

blimmer:
@kaizencc - thanks for the guidance. I've made the requested changes. This is ready for re-review.

packages/@aws-cdk/app-staging-synthesizer-alpha/README.md
```diff
@@ -117,8 +123,13 @@ source code. As part of the `DefaultStagingStack`, an S3 bucket and IAM role wil
 used to upload the asset to S3.
 
 ```ts
+import { BucketEncryption } from 'aws-cdk-lib/aws-s3';
+
 const app = new App({
-  defaultStackSynthesizer: AppStagingSynthesizer.defaultResources({ appId: 'my-app-id' }),
+  defaultStackSynthesizer: AppStagingSynthesizer.defaultResources({
+    appId: 'my-app-id',
+    stagingBucketEncryption: BucketEncryption.S3_MANAGED
```
kaizencc, line 131:
```suggestion
    stagingBucketEncryption: BucketEncryption.S3_MANAGED,
```

kaizencc (approved):
Thanks @blimmer! Hope this helps the usability of this package

kaizencc (approved):

Issue #28862, opened 2024-01-25
(aws-dms): Tags.of(CfnReplicationConfig construct).add({...}) does not work


gossandr:
### Describe the bug

When attempting to add Tags to the CfnReplicationConfig construct, no tags are actually added. Furthermore, removing the Tags.of() and re-deploying the stack results in an error (presumably from the DMS API?).  Tags.of() does work for other constructs in aws-dms. For example, for the CfnEndpoint construct Tags.of() works as expected.

### Expected Behavior

I would expect the tags to be added to the resource. 

I would expect after removing Tags.of from the resource construct that I would  not get any deployment errors.

### Current Behavior

after adding Tags.of() and deploying, no tags are deployed on the replication config resource.

after then removing the Tags.of() and re-deploying, the deploy fails with this error, perhaps related to the dms API itself.

```
Resource handler returned message: "No modifications requested for replication
config with arn, 'arn:aws:dms:eu-west-1:259363291751:replication-config:RTQ7JQX
ZPNBOBJFNHAVMWAJXLWS35IWZOID6VAI' (Service: DatabaseMigration, Status Code: 400
, Request ID: 2248b0dc-8583-4b56-8f20-5381c630c8f0)" (RequestToken: c316768f-94
3b-0a97-d2ed-550977a3729d, HandlerErrorCode: GeneralServiceException)
```

### Reproduction Steps

```
export class DmsReplication extends Construct {

  constructor(scope: Construct, id: string, props: DmsReplicationProps) {
    super(scope, id);

    const replication = new CfnReplicationConfig(this, 'rc01', {

      replicationType: 'full-load',


      sourceEndpointArn: 'arn:of:source:endpoint',
      targetEndpointArn: 'arn:of:target:endpoint',

    });
    
    
    Tags.of(replication).add('test', 'value');
  }
```

### Possible Solution

I suspect that this is related to the relative new-ness of the serverless replication support in cloudformation. But I have honestly no idea. :swea

### CDK CLI Version

2.123.0 (build a594112)

### Node.js Version

v18.2.0

### OS

macos Ventura 13.6.3 (22G436)

### Language

TypeScript

### Language Version

TypeScript 5.2.2

pahud:
I think the `tags` of `CfnReplicationConfig` is probably mis-typed. It should be `tags?: cdk.TagManager;` instead.

```
    /**
     * One or more optional tags associated with resources used by the AWS DMS Serverless replication.
     */
[...]

PR #28989, opened 2024-02-05
fix(core): a number of resources are not taggable with `Tags.of()`

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/28862

### Reason for this change

CFN resources that has array format `tags` cannot use `Tags.of()` to add tags.

### Description of changes

Enable modern style tags `ITaggableV2`

The generated looks correct and shouldn't cause breaking changes.


rix0rrr:
What we usually do is phrase the PR title as user-visible changes (because the PR titles make it into the CHANGELOG). Since this fixes a bug, I would do something like:


```
fix(core): a number of resources are not taggable with `Tags.of()`
```

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
```diff
@@ -184,7 +185,7 @@ export class ResourceDecider {
 
   private handleTagPropertyModern(cfnName: string, prop: Property, variant: TagVariant) {
     const originalName = propertyNameFromCloudFormation(cfnName);
-    const originalType = this.converter.makeTypeResolvable(this.converter.typeFromProperty(prop));
```
rix0rrr, line 187: Why this change?

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
```diff
@@ -221,7 +222,7 @@ export class ResourceDecider {
             expr.object({ tagPropertyName: expr.lit(originalName) }),
           ),
         cfnValueToRender: {
-          [originalName]: $this.tags.renderTags(...(HAS_25610 ? [$this[originalName]] : [])),
```
rix0rrr, line 224: Why this change?

GavinZZ: This is necessary because the `tags` type is `public tags?: Array<cdk.CfnTag>;` instead of `TagManager`. 

So the original code `this.tags.renderTags(...)` won't work here since `tags` is an array which doesn't have `renderTags` method. However, since it inherits `ITaggableV2`, it has direct access to `cdkTagManager` to access `renderTags`.

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
GavinZZ: This line will the the tags type from `Array<cdk.CfnTag>` into `IResolvable | Array<cdk.CfnTags | IResolvable>`. This is not ideal as it will cause exception [here](https://github.com/aws/aws-cdk/blob/dd7efb11ba846fba2fd63db57c27cd3f210705c6/tools/@aws-cdk/spec2cdk/lib/cdk/cloudformation-mapping.ts#L187) since `originalType` doesn't find `Array<cdk.CfnTags | IResolvable>` type.

Removing the resolvable is ideal since this feature was just enabled and guaranteed to not cause any breaking changes. Fixing it at the exception location is less ideal since the code is used in other generation steps as well.

GavinZZ:
Exemption Request
No existing test suite for `spec2cdk` tool.

GavinZZ:
@mergify update

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
```diff
@@ -217,11 +218,11 @@ export class ResourceDecider {
           new CDK_CORE.TagManager(
             this.tagManagerVariant(variant),
             expr.lit(this.resource.cloudFormationType),
-            HAS_25610 ? expr.UNDEFINED : $E(props)[originalName],
+            $E(props)[originalName],
```
GavinZZ, line 221: This is to set the initial tag value to be `props.tags` value.

gossandr:
Any update on this? I am looking forward to a new release ;-)

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
rix0rrr: We still need to pass the directly-passed tags in though, to combine them with the aspect-applied tags?

Should this not look like:

```ts
 [originalName]: $this.cdkTagManager.renderTags($this[originalName]),
```

Also, since we effectively seem to have removed all reference to `HAS_25610`, we can as well remove the boolean constant I think?

GavinZZ: Good point. Will update

GavinZZ:
@gossandr Will try to get this merged this week.

gossandr:
@GavinZZ if you need any assistance, i might be able to help

tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
GavinZZ: The reason that `$this[originalName]` is not needed is because I provided initial tag in the initizer. My understanding is that both serves the same purpose. Correct me if wrong.
```
initializer: (props: Expression) =>
          new CDK_CORE.TagManager(
            this.tagManagerVariant(variant),
            expr.lit(this.resource.cloudFormationType),
            $E(props)[originalName],
            expr.object({ tagPropertyName: expr.lit(originalName) }),
          ),
```

rix0rrr (approved):

PR #28995, opened 2024-02-06
feat: list stack dependencies

vinayak-kukreja:
### Reason for this change

Existing `cdk list` functionality does not display stack dependencies. This PR introduces that functionality. 

For instance,
Existing functionality:
```
❯ cdk list
producer
consumer
```

Feature functionality:
```
❯ cdk list --show-dependencies
- id: producer
  dependencies: []
- id: consumer
  dependencies:
    - id: producer
      dependencies: []
```

### Description of changes

Changes are based on internal team design discussions.
* A new flag `--show-dependencies` is being introduced for `list` cli command.
* A new file `list-stacks.ts` is being added.
    * Adding `listStacks` function within the file for listing stack and their dependencies using cloud assembly from the cdk toolkit. 

* Unit and Integration testing

### Checklist

### Co-Author

Co-authored-by: @SankyRed 

-----

> NOTE: We are currently getting it reviewed by UX too so the final display output might change.

----


packages/aws-cdk/README.md
```diff
@@ -74,7 +74,7 @@ $ cdk init lib --language=typescript
 
 ### `cdk list`
 
-Lists the stacks modeled in the CDK app.
+Lists the stacks and their dependencies modeled in the CDK app.
 
 ```console
 $ # List all stacks in the CDK app 'node bin/main.js'
```
vinayak-kukreja, line 80: We need to update these examples once CLI team decides on the UX for this command.

packages/aws-cdk/test/workflows.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        'Test-Stack-A': { Foo: 'Bar' },
+        'Test-Stack-B': { Baz: 'Zinga!' },
+      }),
+    });
+
+    // WHEN
+    const workflow = await listWorkflow( toolkit, { selectors: ['Test-Stack-A', 'Test-Stack-B'] });
+
+    // THEN
+    expect(JSON.parse(workflow)).toEqual([{
```
vinayak-kukreja, line 101: This should fail. The output does not have dependencies within them.

packages/aws-cdk/test/workflows.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  constructor(
+    expectedTags: { [stackName: string]: { [key: string]: string } } = {},
+  ) {
+    super({ sdkProvider: new MockSdkProvider() });
+
+    for (const [stackName, tags] of Object.entries(expectedTags)) {
+      this.expectedTags[stackName] =
+        Object.entries(tags).map(([Key, Value]) => ({ Key, Value }))
+          .sort((l, r) => l.Key.localeCompare(r.Key));
+    }
```
vinayak-kukreja, line 173: Why do we need this?

packages/aws-cdk/test/workflows.test.ts
vinayak-kukreja: Maybe `toEqual` is not doing a deep equal. Maybe this helps: https://jestjs.io/docs/expect#tomatchobjectobject

vinayak-kukreja:
Exemption request:

Since this is a new flow path, the integration test for CLI would not make sense here. Integ tests would be added once we have the other flow path ready.

vinayak-kukreja:
We are emitting warnings as stack [here](https://github.com/aws/aws-cdk/blob/eca1bcf0b8a449d07692e6363cefa149b1fe0ce4/packages/%40aws-cdk/integ-runner/lib/runner/integ-test-suite.ts#L140-L142)

Which is causing integ test runs to fail with,
```
@aws-cdk-testing/framework-integ:   ERROR      aws-apigatewayv2-integrations/test/http/integ.nlb 1.793s
@aws-cdk-testing/framework-integ:       "cdk-integ" can only operate on apps with a single stack.
@aws-cdk-testing/framework-integ:   If your app has multiple stacks, specify which stack to select by adding this to your test source:
@aws-cdk-testing/framework-integ:       /// !cdk-integ STACK ...
@aws-cdk-testing/framework-integ:   Available stacks: [Warning at /integ-nlb-integration/lb/listener/targetGroup] When creating an empty TargetGroup, you should specify a 'targetType' (this warning may become an error in the future). [ack: @aws-cdk/aws-elbv2:targetGroupSpecifyTargetTypeForEmptyTargetGroup] ---NEW STACK--- integ-nlb-integration (wildcards are also supported) and Stack Count: 2
```

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -844,6 +844,139 @@ integTest('cdk ls', withDefaultFixture(async (fixture) => {
   }
 }));
 
+/**
+ * Type to store stack dependencies recursively
+ */
+export type DependencyDetails = {
```
vinayak-kukreja, line 850: Do we need to export these?

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -844,6 +844,139 @@ integTest('cdk ls', withDefaultFixture(async (fixture) => {
...
+  id: string;
+  dependencies: DependencyDetails[];
+};
+
+export type StackDetails = {
```
vinayak-kukreja, line 855: Similar comment for this.

vinayak-kukreja:
Exemption Request:

We are cli integ testing this change which would not generate a snapshot.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
SankyRed: Not needed to export. Nice catch, will update!

packages/aws-cdk/lib/list-stacks.ts
```diff
@@ -0,0 +1,98 @@
...
+          const stackWithDeps = calculateStackDependencies(depStack);
+
+          for (const stackDetail of stackWithDeps) {
+            data.dependencies.push({
+              id: stackDetail.id,
+              dependencies: stackDetail.dependencies,
+            });
+          }
+        } else {
+          data.dependencies?.push({
```
colifran, line 84: nit: should this be optional? Doesn't `data` always have a `dependencies` array even if it is just empty?

packages/aws-cdk/lib/list-stacks.ts
```diff
@@ -0,0 +1,98 @@
...
+      };
+
+      for (const dependencyId of stack.dependencies.map(x => x.id)) {
+        if (dependencyId.includes('.assets')) {
+          continue;
+        }
+
+        const depStack = assembly.stackById(dependencyId);
+
+        if (depStack.stackArtifacts[0].dependencies.length > 0 &&
```
colifran, line 72: Just curious -- why are we only getting the first `stackArtifact` here?

colifran (changes requested):
Looks great! I just have a few small comments.

packages/aws-cdk/lib/list-stacks.ts
vinayak-kukreja: So we are getting the dependent stack by ID: `assembly.stackById(dependencyId);`, so there must be just one unique stack by id. 

I can add a comment explaining this or/and initially I had an error if it was greater than one but I believed that validation should be done before this (I can check if that is present). Thougths?

packages/aws-cdk/lib/list-stacks.ts
colifran: To be clear I just mean I think this could just be `data.dependencies.push(...)`

packages/aws-cdk/lib/list-stacks.ts
vinayak-kukreja: Ok, so [stackById](https://github.com/aws/aws-cdk/blob/7205143d563449ab211406da04d99bbbcdc7c8cc/packages/aws-cdk/lib/api/cxapp/cloud-assembly.ts#L206-L212) uses [getStackArtifact](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/cx-api/lib/cloud-assembly.ts#L110) which would always return one artifact but `stackById` wraps it in an array.

packages/aws-cdk/lib/list-stacks.ts
vinayak-kukreja: You are correct. Removing this. Good catch 💯

packages/aws-cdk/lib/list-stacks.ts
colifran: Thanks for the clarification. Not now, unrelated, and also not sure what the impact would be, but maybe this is something we can update in a future PR. It feels like `stackById` should return a single stack, not an array containing a single stack.

vinayak-kukreja: I agree :)

colifran (approved):

w-oh:
Approved by UXD 3/4/24 (woh@)

Issue #28907, opened 2024-01-29
ec2: Upgrade NatInstanceProvider


tmokmss:
### Describe the feature

The current `NatInstanceProvider` [uses](https://github.com/aws/aws-cdk/blob/6a7a24afcc1ebebf71c267b890732a455e865cc8/packages/aws-cdk-lib/aws-ec2/lib/nat.ts#L407-L414) the AWS official NAT instance AMI, which has already reached EOL on Dec 31 2023.

> NAT AMI is built on the last version of the Amazon Linux AMI, 2018.03, which reached the end of standard support on December 31, 2020 and end of maintenance support on December 31, 2023. For more information, see the following blog post: [Amazon Linux AMI end of life](https://aws.amazon.com/blogs/aws/update-on-amazon-linux-ami-end-of-life/).
> 
> If NAT instances are a better match for your use case than NAT gateways, you can create your own NAT AMI from a current version of Amazon Linux as described in [Create a NAT AMI](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_NAT_Instance.html#create-nat-ami).
> 
> https://docs.aws.amazon.com/vpc/latest/userguide/VPC_NAT_Instance.html

It'd be great if CDK provides an easy way to configure a modern NAT instance.

### Use Case

NAT Instances are still popular since they can be much cheaper than NAT Gateways.

### Proposed Solution

We can create a NAT instance from AL2023 AMI by using these commands in [this doc](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_NAT_Instance.html#create-nat-ami).

I think CDK can internally use ec2 userdata to configure them and abstract all the complexities away from users.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.124.0

### Environment details (OS name and version, etc.)

macOS

pahud:
Yes this is very important. Thank you @tmokmss

GavinZZ:
I would like to work on this issue if it's not been worked on yet. But I have some questions before I can get started with the implementation. Reading the description, it seems that the default `nat instance machine image` is using a EOL machine image. Are you suggesting that this should be update to a new&supported version?

> We can create a NAT instance from AL2023 AMI by using these commands in [this doc](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_NAT_Instance.html#create-nat-ami)

I'm a bit confused on how it relates to the commands you linked in this doc.

[...]

PR #29013, opened 2024-02-07
feat(ec2): add NAT instance V2 support using AL2023

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/28907

### Reason for this change

Current NAT instance image has reached EOL on Dec 31 2023.

### Description of changes

If NAT instances are a better match for your use case than NAT gateways, you can create your own NAT AMI from a current version of Amazon Linux as described in [Create a NAT AMI](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_NAT_Instance.html#create-nat-ami).

New unit and integration tests


packages/aws-cdk-lib/aws-ec2/lib/nat.ts
```diff
@@ -399,6 +419,115 @@ class PrefSet<A> {
...
+      throw new Error('Cannot specify both of \'keyName\' and \'keyPair\'; prefer \'keyPair\'');
+    }
+  }
+
+  public configureNat(options: ConfigureNatOptions) {
+    const defaultDirection = this.props.defaultAllowedTraffic ??
+      (this.props.allowAllTraffic ?? true ? NatTrafficDirection.INBOUND_AND_OUTBOUND : NatTrafficDirection.OUTBOUND_ONLY);
+
+    // Create the NAT instances. They can share a security group and a Role.
+    const machineImage = this.props.machineImage || new AmazonLinuxImage({ generation: AmazonLinuxGeneration.AMAZON_LINUX_2 });
```
tmokmss, line 448: Is there any reason not to use AL2023 instead of AL2 by default? AL2 will [reach EOL](https://aws.amazon.com/amazon-linux-2/faqs/) next year, whereas AL2023 [will be supported](https://aws.amazon.com/linux/amazon-linux-2023/faqs/) until 2028.

tmokmss, line 448: It'll be more useful if it automatically sets an AMI [cpu type](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.AmazonLinuxImageProps.html#cputype) according to the instance type [architecture](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.InstanceType.html#architecture). 

I believe lots of people will use t4g.nano because it's the cheapest instance type.

GavinZZ: Good point, I'll update it to AL2023

GavinZZ: I don't want to make the default value based on assumption. I haven't found any support evidence that this is the most wanted instance type, so I'll keep the instance type as is.

tmokmss: Thanks. I didn't mean make arm architecture default. Instead we can infer the appropriate architecture from an instance type passed from a user. This way users will not need to be aware of arch type of the machine image. Currently if a user choose a graviton instance type, it fails during CFn deployment because of architecture mismatch, resulting in bad UX.

packages/aws-cdk-lib/aws-ec2/README.md
```diff
@@ -189,9 +189,9 @@ are managed by AWS. If you would prefer to use your own managed NAT
 *instances* instead, specify a different value for the `natGatewayProvider`
 property, as follows:
 
-[using NAT instances](test/integ.nat-instances.lit.ts)
+[using NAT instances](test/integ.nat-instances-v2.ts)
```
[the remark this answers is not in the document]
GavinZZ: Thanks Kaizen for the clarification. I've updated the README accordingly.

packages/aws-cdk-lib/aws-ec2/lib/nat.ts
```diff
@@ -399,6 +419,115 @@ class PrefSet<A> {
...
+    }
+  }
+
+  public configureNat(options: ConfigureNatOptions) {
+    const defaultDirection = this.props.defaultAllowedTraffic ??
+      (this.props.allowAllTraffic ?? true ? NatTrafficDirection.INBOUND_AND_OUTBOUND : NatTrafficDirection.OUTBOUND_ONLY);
+
+    // Create the NAT instances. They can share a security group and a Role.
+    const machineImage = this.props.machineImage || new AmazonLinuxImage({ generation: AmazonLinuxGeneration.AMAZON_LINUX_2023 });
+    this._securityGroup = this.props.securityGroup ?? new SecurityGroup(options.vpc, 'NatSecurityGroup', {
```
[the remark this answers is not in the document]
kaizencc: If your answer is "that is what was done in `NatInstanceProvider`, well, someone there probably didn't realize our convention either

packages/aws-cdk-lib/aws-ec2/lib/nat.ts
```diff
@@ -271,6 +289,8 @@ class NatGatewayProvider extends NatProvider {
 
 /**
  * NAT provider which uses NAT Instances
+ *
+ * @deprecated use NatInstanceProviderV2
```
kaizencc, line 293: can we say why this is deprecated too

packages/aws-cdk-lib/aws-ec2/lib/nat.ts
```diff
@@ -399,6 +419,115 @@ class PrefSet<A> {
...
+      throw new Error('Cannot specify both of \'keyName\' and \'keyPair\'; prefer \'keyPair\'');
+    }
+  }
+
+  public configureNat(options: ConfigureNatOptions) {
+    const defaultDirection = this.props.defaultAllowedTraffic ??
+      (this.props.allowAllTraffic ?? true ? NatTrafficDirection.INBOUND_AND_OUTBOUND : NatTrafficDirection.OUTBOUND_ONLY);
+
+    // Create the NAT instances. They can share a security group and a Role.
+    const machineImage = this.props.machineImage || new AmazonLinuxImage({ generation: AmazonLinuxGeneration.AMAZON_LINUX_2023 });
```
kaizencc, line 448: Can we document why we are doing an `InstanceV2` here? I don't quite understand. This looks like the important change between the two. Does that mean that everything under `NatInstanceImage()` is old-generation and deprecated?

Either way, lets document clearly why we went down this path

kaizencc (commented):
A few more comments @GavinZZ.

packages/aws-cdk-lib/aws-ec2/lib/nat.ts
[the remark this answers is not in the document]
GavinZZ: Agree, I copied and pasted and didn't realize this inconsistency. Will update.

GavinZZ:
@kaizencc Can I get another review? I would love to have this PR merged.

kaizencc (approved):

PR #29021, opened 2024-02-07
fix(appconfig): deprecate deploy method

chenjane-dev:
### Reason for this change

Want to recommend against creating AppConfig resources across multiple stacks.

### Checklist

----


kaizencc (commented):
Couple of minor things

- wouldn't call this a refactor. we want this to show up in the changelog, so lets call it a fix
- part of the stabilization process if/when we get there, will be to remove deprecated properties. so what this PR means is that we plan on removing this API in the future. just want to confirm.

chenjane-dev:
Updated as a fix. Yes we plan on removing this once we stabilize

kaizencc (approved):

PR #16190, opened 2021-08-24
feat(cognito): identity pools

smguggen:
Adds Identity Pool L2 Construct which to date has not been implemented. Since Identity Pool's can't be used without default auth and unauth roles, also incorporated the L1 CfnIdentityPoolRoleAttachment into the Construct. Contains unit and integration tests as well as fully updated ReadMe.

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #29025, opened 2024-02-08
fix(cognito-identitypool-alpha): inconvenient IdentityPoolProviderUrl.userPool()

Tietew:
### Reason for this change

`IdentityPoolProviderUrl.userPool()` requires a string `url` currently.
The description is "User Pool Provider Url".
It should be ``` `${userPool.userPoolProviderName}:${userPoolClient.userPoolClientId}` ```.

`UserPool` has an attribute `userPoolProviderUrl` which description is "User Pool Provider Url", but confusingly, it cannot be specified to `IdentityPoolProviderUrl.userPool()`.

The format of the identity provider identifier isn't well documented.
See [SetIdentityPoolRoles API reference](https://docs.aws.amazon.com/cognitoidentity/latest/APIReference/API_SetIdentityPoolRoles.html) for example of User Pool's identity provider identifier.

### Description of changes

This PR fixes `IdentityPoolProviderUrl.userPool()` to accept `UserPool` and `UserPoolClient` instead of a string `url`.
It generates a correct identifier described above.

Existing integration test generates an identifier as described above.
The snapshot won't be changed by this PR.

### Checklist

BREAKING CHANGE: The argument of `IdentityPoolProviderUrl.userPool()` has been changed from `url: string` to `userPool: UserPool, userPoolClient: UserPoolClient`. If you want to specify custom identifier string, use `IdentityPoolProviderUrl.custom()` instead.

----


Tietew:
Exemption Request
The snapshop should not be changed.

packages/@aws-cdk/aws-cognito-identitypool-alpha/lib/identitypool.ts
```diff
@@ -155,7 +157,8 @@ export class IdentityPoolProviderUrl {
   }
 
   /** User Pool Provider Url */
-  public static userPool(url: string): IdentityPoolProviderUrl {
+  public static userPool(userPool: UserPool, userPoolClient: UserPoolClient): IdentityPoolProviderUrl {
```
kaizencc, line 160: I know we're ok with breaking changes, but this seems like it would ruin anyone who was simply supplying a url here that was not connected to a specific `UserPool` or `UserPoolCliet` construct. I'm not sure what the difference would be if the `IdentityPoolProviderType` turns from `USER_POOL` to `CUSTOM`

packages/@aws-cdk/aws-cognito-identitypool-alpha/test/identitypool.test.ts
```diff
@@ -440,7 +440,7 @@ describe('role mappings', () => {
     const providerUrl = Fn.importValue('ProviderUrl');
     expect(() => new IdentityPool(stack, 'TestIdentityPoolRoleMappingErrors', {
       roleMappings: [{
-        providerUrl: IdentityPoolProviderUrl.userPool(providerUrl),
+        providerUrl: IdentityPoolProviderUrl.custom(providerUrl),
```
kaizencc, line 443: This would change the `IdentityPoolProviderTyle` right? How does that affect things?

kaizencc (commented):
Couple probing questions here to start @Tietew

packages/@aws-cdk/aws-cognito-identitypool-alpha/lib/identitypool.ts
Tietew: `IdentityPoolProviderType` (`IdentityPoolProviderUrl.type`) is not used since #16190 (implemented PR).
Changing `.userPool(somthing)` to `.custom(something)` makes no differences in CloudFormation template.

packages/@aws-cdk/aws-cognito-identitypool-alpha/test/identitypool.test.ts
Tietew: Same as above. No affects in CloudFormation template.

Tietew:
Updated BREAKING CHANGE in description

TheRealAmazonKendra (approved):

TheRealAmazonKendra:
@Mergifyio update

PR #29031, opened 2024-02-08
feat(cloudfront): `reportOnly` flag on `ResponseHeadersContentSecurityPolicy`

### LINKED ISSUE #29006: `cloudfront`: Support `Content-Security-Policy-Report-Only` header
### Describe the feature

Support setting 'report-only' mode for CSP in the `ResponseHeadersPolicy`
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only

### Use Case

> The HTTP Content-Security-Policy-Report-Only response header allows web developers to experiment with policies by monitoring (but not enforcing) their effects. These violation reports consist of [JSON](https://developer.mozilla.org/en-US/docs/Glossary/JSON) documents sent via an HTTP POST request to the specified URI.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.126.0

n/a

dillonstreator:
### Issue #29006 

Closes #29006.

### Reason for this change

Improve ergonomics/discoverability/DX of CSP Report Only configuration

### Description of changes

Added an optional `reportOnly` field on `ResponseHeadersContentSecurityPolicy` that maps the CSP to a custom header.

Added a unit test

### Checklist

----


dillonstreator:
Exemption Request:
I may be missing something but I don't think the following check applies to this change. I couldn't find a relevant integration test or snapshot.
```
❌ Features must contain a change to an integration test file and the resulting snapshot.
```

packages/aws-cdk-lib/aws-cloudfront/lib/response-headers-policy.ts
```diff
@@ -114,13 +114,36 @@ export class ResponseHeadersPolicy extends Resource implements IResponseHeadersP
...
+        ...securityHeadersBehavior,
+        contentSecurityPolicy: undefined,
+      };
+
+      if (!customHeadersBehavior) {
+        customHeadersBehavior = {
+          customHeaders: [],
+        };
+      }
+      // TODO: log a warning if custom headers already contains CSP-Report-Only header?
```
TheRealAmazonKendra, line 136: Looks like this PR is not ready?

packages/aws-cdk-lib/aws-cloudfront/lib/response-headers-policy.ts
```diff
@@ -114,13 +114,36 @@ export class ResponseHeadersPolicy extends Resource implements IResponseHeadersP
...
+      };
+      securityHeadersBehavior = {
+        ...securityHeadersBehavior,
+        contentSecurityPolicy: undefined,
+      };
+
+      if (!customHeadersBehavior) {
+        customHeadersBehavior = {
+          customHeaders: [],
+        };
```
TheRealAmazonKendra, line 134: Why are you making this an empty array only if `reportOnly` is set?

packages/aws-cdk-lib/aws-cloudfront/test/response-headers-policy.test.ts
```diff
@@ -180,4 +180,26 @@ describe('ResponseHeadersPolicy', () => {
       },
     });
   });
+
+  test('it respects CSP `reportOnly` flag by mapping to custom header', () => {
```
TheRealAmazonKendra, line 184: This is only testing a single logic path. We need to test each of the paths added by this PR.

TheRealAmazonKendra (changes requested):
Thank you for your contribution. I notice that this still has a TODO in it so we would want that completed before we take another look. In regard to your exemption request, I'd like to see an integ test that uses the settings you're adding so that we know it deploys just fine.

packages/aws-cdk-lib/aws-cloudfront/lib/response-headers-policy.ts
dillonstreator: This is mainly a question mark. I didn't notice a pattern for logging at this layer of the application but thought it could be valuable to surface a warning in the event of conflicting configuration. If someone sets `reportOnly`, the expectation would be that there is not already a `Content-Security-Policy-Report-Only` set in the custom headers. However, this cannot be enforced pre-runtime.

packages/aws-cdk-lib/aws-cloudfront/lib/response-headers-policy.ts
dillonstreator: The `Content-Security-Policy-Report-Only` header is only respected as a custom header configuration. The following line is pushing the report only csp header into this `customHeaders` array. If the `customHeadersBehavior` object is undefined (which it's allowed to be), we wont have anything to push into.
In the non-report-only configuration case, we don't need to worry about touching the `customHeaders` array.

packages/aws-cdk-lib/aws-cloudfront/test/response-headers-policy.test.ts
dillonstreator: This test hits both new paths introduced by this PR
- `reportOnly` being set
- `customHeadersBehavior` not existing

dillonstreator:
> In regard to your exemption request, I'd like to see an integ test that uses the settings you're adding so that we know it deploys just fine.

@TheRealAmazonKendra - could you point me the right direction here? I couldn't find an existing integration test that covered `ResponseHeadersPolicy`, leading me to believe this should be exempt.

Issue #16954, opened 2021-10-13
RDS: support passing PreferredMaintenanceWindow to Cluster database instances


cpaton:
### Description

When creating a DatabaseCluster it should be possible to specify PreferredMaintenanceWindow property for the database instances as per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-preferredmaintenancewindow

### Use Case

Control maintenance window for database instances so maintenance is performed outside of production use hours e.g. minor version upgrades

### Proposed Solution

Add preferredMaintenanceWindow property to InstanceProps when creating a DatabaseCluster.  This value should be copied onto the underlying CfnDatabaseInstance object

### Acknowledge

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

skinny85:
Thanks for opening the issue @cpaton. This should be a relatively simple feature to implement - it means adding the `preferredMaintenanceWindow` property to the [`InstanceProps` interface](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-rds.InstanceProps.html), and then passing it [here](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-rds.InstanceProps.html).

Would you consider opening us a PR adding this feature? Here's our "Contributing guide": https://github.com/aws/aws-cdk/blob/master/CONTRIBUTING.md.

Thanks,
Adam

ronocod:
If anyone else is running into this, you can synchronize the DBInstance maintenance windows with the DBCluster window using this:
```typescript
function setInstanceMaintenanceWindows(cluster: rds.DatabaseCluster) {
  const clusterNode = cluster.node.children.find(
    (c) => c instanceof rds.CfnDBCluster
  ) as rds.CfnDBCluster;
  cluster.node.children.forEach((c) => {
    if (c instanceof rds.CfnDBInstance) {
      const temp = c;
      temp.preferredMaintenanceWindow = clusterNode.preferredMaintenanceWindow;
    }
  });
}

//  setInstanceMaintenanceWindows(this.cluster);

keisukekomeda:
Can also use cdk.Aspects to set node property.

```ts
cdk.Aspects.of(cluster).add({
  visit: (node) => {
    if (node instanceof rds.CfnDBInstance) {
      node.addPropertyOverride(
        "PreferredMaintenanceWindow",
[...]

PR #29033, opened 2024-02-08
feat(rds): add ability to specify PreferredMaintenanceWindow to RDS cluster database instances

rjjaisim-at-amazon:
### Issue # (if applicable)

Closes [#16954](https://github.com/aws/aws-cdk/issues/16954)

### Reason for this change
Noticed that we were able to specify preferredMaintenanceWindow for a cluster, but unable to do so for the instances created under the cluster. Instead, AWS (semi-)randomly assigns a maintenance window ([doc](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Maintenance.html#Concepts.DBMaintenance)) for the instances, which leads to things being out of sync b/w the cluster and its child instances

There are some workarounds as mentioned in the issue above, but those are a little hacky (imo) and I figured adding the preferredMaintenanceWindow as an instance prop is a better long-term solution. Also, it might be hard for other developers to find the workarounds as they are only mentioned in the above issue and aren't available through normal channels (Stack overflow/official CDK docs) 

### Description of changes
Added optional preferredMaintenanceWindow field under `InstanceProps`, and passed that field in during the creation of the `CfnDBInstance`. Also added a quick unit test

Added a unit test, did not add integ tests. Ran `yarn build` and `yarn test`

Callout: I was unable to run integration tests locally, kept getting errors with `yarn integ --directory packages/aws-cdk-lib/aws-rds` and `yarn integ-runner --directory packages/aws-cdk-lib/aws-rds` - `Error: Cannot find module './integ-runner.js'`, not sure if I'm missing something

### Checklist

No breaking changes


packages/aws-cdk-lib/aws-rds/lib/props.ts
```diff
@@ -101,6 +101,17 @@ export interface InstanceProps {
...
+  /**
+   * A preferred maintenance window day/time range. Should be specified as a range ddd:hh24:mi-ddd:hh24:mi (24H Clock UTC).
+   *
+   * Example: 'Sun:23:45-Mon:00:15'
+   *
+   * @default - 30-minute window selected at random from an 8-hour block of time for
+   * each AWS Region, occurring on a random day of the week.
+   * @see https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/USER_UpgradeDBInstance.Maintenance.html#Concepts.DBMaintenance
+   */
+  readonly preferredMaintenanceWindow?: string;
```
xazhao, line 114: Because it's a string requires special format, what errors will customer see if the format is not right?

xazhao: Just thinking maybe we should add a regex check.

packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
```diff
@@ -209,6 +209,30 @@ describe('cluster new api', () => {
         ],
       });
     });
+
+    test('preferredMaintenanceWindow provided in InstanceProps', () => {
+      // GIVEN
+      const stack = testStack();
+      const vpc = new ec2.Vpc(stack, 'VPC');
+
+      const PREFERRED_MAINTENANCE_WINDOW: string = 'Sun:12:00-Sun:13:00';
```
xazhao, line 218: We should test when the property is not using the format. It will likely to happen because of a typo.

xazhao:
I'm okay with no integration changes. But IMO we should add this property to `README` to give customers a better idea how to use it. We can add a small section `specify preferred maintenance window` and provide a simple example.

rjjaisim-at-amazon:
Added details to README.md. 

As for validation of the string, I believe that is done at runtime when the stacks are deployed. This has always been the case for CfnDBInstances - the preferredMaintenanceWindow prop has always existed for CfnDBInstances, but has always been set to undefined when these instances were created under a cluster (and therefore AWS would semi-randomly set a value for it on creation). My PR only allows the ability to pass in preferredMaintenanceWindow prop when creating CfnDBInstances under a cluster instead of it being undefined. We should keep the validation as is IMO and it doesn't fall under the purview of my PR

xazhao (dismissed):
Looks good to me

xazhao (approved):

Konosh93:
@rjjaisim-at-amazon Thank you for this fix. I was wondering why this change utilized `instanceProps` and not `writer` and `readers` props instead, given that `instanceProps` is deprecated?

rjjaisim-at-amazon:
@Konosh93 Thanks for pointing that out, I wasn't aware of the writer and reader props - thanks for adding it

PR #29043, opened 2024-02-09
feat(stepfunctions-tasks): bedrock createModelCustomizationJob integration

### LINKED ISSUE #29042: stepfunctions-tasks: support for bedrock createModelCustomizationJob task
### Describe the feature

AWS stepfunction supports bedrock optimized integrations. 
Currently, only `invokeModel` is supported by CDK, but I would like `createModelCustomizationJob` to be supported in the same manner.

### Use Case

This would be helpful when wanting to execute `createModelCustomizationJob` as a single task in Step Functions.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.123.0

irrelevant

badmintoncryer:
### Issue

Closes #29042

### Reason for this change

AWS stepfunctions support optimized integration with AWS bedrock.
Currently, only `invokeModel` is supported by CDK, but I would like `createModelCustomizationJob` to be supported in the same manner.

### Description of changes

I've added CreatemodelCustomizationJob class. 

```ts
const taskConfig = {
  baseModel: model,
  clientRequestToken: 'MyToken',
  customizationType: CustomizationType.FINE_TUNING,
  kmsKey,
  customModelName: 'MyCustomModel',
  customModelTags: [{ key: 'key1', value: 'value1' }],
  hyperParameters: {
    batchSize: '10',
  },
  jobName: 'MyCustomizationJob',
  jobTags: [{ key: 'key2', value: 'value2' }],
  outputDataS3Uri: outputBucket.s3UrlForObject(),
  trainingDataS3Uri: trainingBucket.s3UrlForObject(),
  validationDataS3Uri: [validationBucket.s3UrlForObject()],
  vpcConfig: {
    securityGroups: [new ec2.SecurityGroup(stack, 'SecurityGroup', { vpc })],
    subnets: vpc.isolatedSubnets,
  },
};

const task1 = new BedrockCreateModelCustomizationJob(stack, 'CreateModelCustomizationJob1', taskConfig);

const chain = sfn.Chain
  .start(new sfn.Pass(stack, 'Start'))
  .next(task1)
  .next(new sfn.Pass(stack, 'Done'));

new sfn.StateMachine(stack, 'StateMachine', {
  definitionBody: sfn.DefinitionBody.fromChainable(chain),
  timeout: cdk.Duration.seconds(30),
});
```

I've added  both unit and integ tests.

### Checklist

----


packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,396 @@
...
+   *
+   * @default - no tags
+   */
+  readonly jobTags?: ITag[];
+  /**
+   * The S3 URI where the output data is stored.
+   *
+   * @see https://docs.aws.amazon.com/bedrock/latest/APIReference/API_OutputDataConfig.html
+   */
+  readonly outputDataS3Uri: string;
```
lpizzinidev, line 120:
```suggestion
  readonly outputDataS3: s3.IBucket;
```
What about passing a bucket and using the [`s3UrlForObject`](https://github.com/aws/aws-cdk/blob/df48fd781a476ada79b6b49f006a60f33f73d409/packages/aws-cdk-lib/aws-s3/lib/bucket.ts#L169) method to retrieve the S3 URI?
Same for other `...S3Uri` props. 
We can then remove the `s3UriToArn` private method.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,396 @@
...
+   */
+  readonly kmsKey?: kms.IKey;
+  /**
+   * A name for the resulting custom model.
+   */
+  readonly customModelName: string;
+  /**
+   * Tags to attach to the resulting custom model.
+   *
+   * @default - no tags
```
lpizzinidev, line 94:
```suggestion
   * Tags to attach to the resulting custom model.
   *
   * The maximum number of tags is 200.
   *
   * @default - no tags
```
Can you please add constraints documentation for validated props?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,396 @@
...
+  }
+
+  /**
+   * The IAM role for the bedrock create model customization job
+   */
+  public get role(): iam.IRole {
+    return this._role;
+  }
+
+  private renderBedrockCreateModelCustomizationJobRole(): iam.IRole {
```
lpizzinidev, line 196: Can you add a link to the related documentation for this role?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,396 @@
...
+        ],
+      }),
+      new iam.PolicyStatement({
+        actions: ['iam:PassRole'],
+        resources: [this._role.roleArn],
+      }),
+      ...(this.props.kmsKey
+        ? [
+          new iam.PolicyStatement({
+            // TODO - this should be more specific
```
lpizzinidev, line 315: Yes, can you share some documentation about the policy?

lpizzinidev (changes requested):
Thanks 👍
Left some comments for adjustments

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: I've added the link of the documentation touching about S3 access. 
I am referring to the display on the management console for the IAM policy regarding the VPC config settings and there may be no documentation about it.

スクリーンショット 2024-04-01 18 06 18

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: It sounds nice! 

My concern is it may be difficult to specify S3 path.
Ex. To specify s3://someBucket/train as training data and s3://someBucket/validate as validation data.

What is your thought about this? I think it is acceptable to receive these props as IBucket.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: I could only find this [docs](https://docs.aws.amazon.com/step-functions/latest/dg/bedrock-iam.html) which has no information about kms...

packages/aws-cdk-lib/aws-stepfunctions-tasks/README.md
```diff
@@ -355,6 +355,10 @@ const task = new tasks.BedrockInvokeModel(this, 'Prompt Model', {
 });
 ```
 
+### createModelCustomizationJob
+
+WIP
```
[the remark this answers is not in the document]
badmintoncryer: Sorry for easy mistakes..

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
lpizzinidev: Please check [this](https://docs.aws.amazon.com/bedrock/latest/userguide/encryption-custom-job.html).

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
lpizzinidev: Good point!
However, I'm not a fan of the `s3UriToArn` method, I find it a bit too complicated.
Maybe adding an `outputDataS3Prefix?: string` property to specify the S3 path could be a solution (same for the other 2 buckets).
What's your take?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,428 @@
...
+            ] : []),
+            new iam.PolicyStatement({
+              actions: ['s3:GetObject'],
+              resources: [
+                this.s3UriToArn(this.props.trainingDataS3Uri),
+                ...(this.props.validationDataS3Uri.map((s3Uri) => this.s3UriToArn(s3Uri))),
+              ],
+            }),
+            new iam.PolicyStatement({
+              actions: ['s3:PutObject'],
```
lpizzinidev, line 308: From the linked policy, we also need `s3:GetObject` and `s3:ListBucket` for the output bucket  (and the required bucket-level resource).

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,428 @@
...
+                      ...(this.props.vpcConfig
+                        ? this.props.vpcConfig.subnets.map((subnet) => subnet.subnetId)
+                        : []),
+                    ],
+                  },
+                },
+              }),
+            ] : []),
+            new iam.PolicyStatement({
+              actions: ['s3:GetObject'],
```
lpizzinidev, line 301: From the linked policy, we also need `s3:ListBucket` for the training and validation buckets (and the required bucket-level resource).

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,428 @@
...
+    if (this.props.role) {
+      return this.props.role;
+    }
+    const stack = Stack.of(this);
+    const role = new iam.Role(this, 'BedrockRole', {
+      assumedBy: new iam.ServicePrincipal('bedrock.amazonaws.com'),
+      inlinePolicies: {
+        BedrockCreateModelCustomizationJob: new iam.PolicyDocument({
+          statements: [
+            ...(this.props.vpcConfig ? [
```
lpizzinidev, line 238: Can you please create a separate private method for returning the VPC-related statements?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,428 @@
...
+                actions: [
+                  'ec2:CreateNetworkInterfacePermission',
+                  'ec2:DeleteNetworkInterface',
+                  'ec2:DeleteNetworkInterfacePermission',
+                ],
+                resources: ['*'],
+                conditions: {
+                  StringEquals: {
+                    'ec2:Subnet': [
+                      ...(this.props.vpcConfig
```
lpizzinidev, line 292: Do we need this conditional check?
We already have a check at line 238

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
lpizzinidev: This [doc](https://docs.aws.amazon.com/bedrock/latest/userguide/vpc-model-customization.html) seems more related. Can you please update the documentation and the policy as necessary?

lpizzinidev (changes requested):
Thanks for implementing the changes!
I left some suggestions for further improvements

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: I agree with your suggestion and will try to modify the code to accept `IBucket` and `prefix` as arguments respectively.

I have a question for consultation. If `validationDataS3` is received as an array of `IBucket`, should I also receive prefix as an array of strings and associate elements of the same index?

Ex. 
```ts
{
  ...,
  validationDataS3: [bucket1, bucket2],
  validationDataPrefix: ['prefix1', 'prefix2']
}
```

lpizzinidev: 🤔 Yeah, that would be annoying to manage.
What about adding an `interface` for the bucket + prefix pair ([example](https://github.com/aws/aws-cdk/blob/bf2cf51bf1e8458097b23e7f580ad0978b7e4cca/packages/aws-cdk-lib/aws-synthetics/lib/canary.ts#L80))?

badmintoncryer: Thank you for your nice suggestion.

I've defined `BucketConfiguration` and updated training, validation and output data properties for receiving `BucketConfiguration`.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: Thanks a lot.

I've updated the policy for the custom model builder.
```
    "Statement": [
        {
            "Sid": "Permissions for custom model builders",
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::account-id:user/role"
            },
            "Action": [
                "kms:Decrypt",
                "kms:GenerateDataKey",
                "kms:DescribeKey",
                "kms:CreateGrant"
            ],
            "Resource": "*"
        },
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: Thanks.
I've updated the policy description.

badmintoncryer:
@lpizzinidev 
Thank you for your review. I've addressed your comments.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
lpizzinidev: Following, point 1.
If no VPC or a VPC with public internet access is provided, this is correct.
Otherwise, if a VPC with no public internet access is provided, we should validate that the VPC has a gateway endpoint to S3 attached, and provide the bucket policy, VPC endpoint policy, and role permissions specified [here](https://docs.aws.amazon.com/bedrock/latest/userguide/vpc-model-customization.html).

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,486 @@
...
+            resource: 'model-customization-job',
+            resourceName: '*',
+          }),
+        ],
+      }),
+      new iam.PolicyStatement({
+        actions: ['iam:PassRole'],
+        resources: [this._role.roleArn],
+      }),
+      ...(this.props.customModelKmsKey
```
lpizzinidev, line 419: Following point 2, this should be attached to the `customModelKmsKey` key policy.

lpizzinidev (changes requested):
🤔 I don't think the current IAM policies setup is correct.
From the documentation, we should create:
1. If the task is in a private VPC with no internet access, a VPC endpoint policy granting S3 access via the private network ([docs](https://docs.aws.amazon.com/bedrock/latest/userguide/vpc-model-customization.html)). In this case, I think we should also validate that a VPC gateway endpoint for S3 must is provided.
2. If `customModelKmsKey` is specified, a key policy to grant KMS permissions to the Bedrock role ([docs](https://docs.aws.amazon.com/bedrock/latest/userguide/encryption-custom-job.html)

Let me know if you think I missed something or want to discuss it further.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: When dealing with subnets that cannot access the Internet (isolated subnet), I've come to think that it might be best not to handle the following within AWS CDK:
- Verification of endpoint existence
- Configuration of endpoint policies
- Configuration of bucket policies

Starting with the verification of endpoint existence, the [documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/vpc-model-customization.html) mentions the use of S3 gateway endpoints, but I believe S3 interface endpoints might also be used in practice. Indeed, the documentation refers to private access via PrivateLink, i.e., interface endpoints, at the beginning. It seems more appropriate for users to take responsibility for ensuring access routes to S3 from isolated subnets, including setting up appropriate security groups as arguments.

As for endpoint policies, I view these as optional settings to enhance security rather than mandatory configurations. The [documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/vpc-model-customization.html#train-vpc-policy) explicitly permits certain job-related permissions, but such configurations implicitly deny all other S3 operations. Implicitly adding this policy through CDK might affect other uses of the same VPC endpoint.

Similarly, for bucket policies, there could be implications, such as denying operations from outside the VPC for buckets intended for output data, which might be accessed externally.

I apologize for the lengthy discussion but would greatly appreciate any thoughts on this matter. Thank you for your continued support.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: You are right. I added resource based policy to `customModelKmsKey`

```ts
if (this.props.customModelKmsKey) {
      this.props.customModelKmsKey.addToResourcePolicy(new iam.PolicyStatement({
        actions: ['kms:Decrypt', 'kms:GenerateDataKey', 'kms:DescribeKey', 'kms:CreateGrant'],
        resources: ['*'],
        principals: [new iam.ArnPrincipal(this._role.roleArn)],
      }));
    }
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
lpizzinidev: > the documentation mentions the use of S3 gateway endpoints, but I believe S3 interface endpoints might also be used in practice.

Agree, thanks for making the point.

> I think it's valid to verify the existence of either an Interface or Gateway type endpoint

Yes, only if the VPC cannot access the internet.

> Implicitly adding this policy through CDK might affect other uses of the same VPC endpoint.

🤔 
I think we can add the bucket policy restricting access to the VPC without issues.
For the VPC endpoint policy, what about adding a condition that enforces it only on the task role? ([docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-principalarn))

lpizzinidev:
Thanks for updating the implementation 👍
I left https://github.com/aws/aws-cdk/pull/29043#discussion_r1557032110 to be discussed.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: > Yes, only if the VPC cannot access the internet.
Sure. I'll add verification when all `vpcConfig.subnets` are isolated.

How can I retrieve the endpoint resource on this construct?
My only solution is to use `App` or `Stack.node.findAll()` because `IVpc` does not have `endpoints` properties.

```ts
    if (this.props.vpcConfig) {
      if (this.props.vpcConfig.subnets) {
        this.subnets = this.props.vpcConfig.subnets;
      } else {
        // use all subnets in the VPC
        this.subnets = this.props.vpcConfig.vpc.selectSubnets().subnets;
      }

      const isOnlyIsolatedSubnets = this.subnets.every((subnet) => this.props.vpcConfig?.vpc.isolatedSubnets.includes(subnet));
      const hasS3VpcEndpoint = App.of(this)?.node.findAll().some((resource) => {
        return resource instanceof ec2.CfnVPCEndpoint
          && resource.serviceName === `com.amazonaws.${Stack.of(this).region}.s3`
          && resource.vpcId === this.props.vpcConfig?.vpc.vpcId;
      });
      if (isOnlyIsolatedSubnets && !hasS3VpcEndpoint) {
        throw new Error('VPC configuration must include at least one subnet that is not an isolated subnet or a VPC endpoint to S3');
      };
    }
```

> I think we can add the bucket policy restricting access to the VPC without issues.
I'm sorry. You are right.

> what about adding a condition that enforces it only on the task role?
Nice! I'll implement it.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,515 @@
...
+   * @see https://docs.aws.amazon.com/bedrock/latest/APIReference/API_Validator.html
+   */
+  readonly validationData: BucketConfiguration[];
+
+  /**
+   * Configuration parameters for the private Virtual Private Cloud (VPC) that contains the resources you are using for this job.
+   *
+   * @default - no VPC configuration
+   */
+  readonly vpcConfig?: IBedrockCreateModelCustomizationJobVpcConfig;
```
lpizzinidev, line 199:
```suggestion
   * The minimum number of security groups for the VPC is 1.
   * The maximum number of security groups for the VPC is 5.
   *
   * The maximum number of subnets for the VPC is 1.
   * The maximum number of subnets for the VPC is 16.
   *
   * @default - no VPC configuration
   */
  readonly vpc?: ec2.IVpc;
```
This might simplify things in case we want to add the optional VPC endpoint policy and bucket policy (requires the VPC ID) in the future. Also, it provides a cleaner interface.
What do you think?

lpizzinidev (changes requested):
Thanks for implementing the changes 👍
I left comments for adjusting the `vpc` interface (https://github.com/aws/aws-cdk/pull/29043#discussion_r1562973486) and other nits.
Thank you for your patience, I should have thought about it before 😅 (of course let me know if you think that the implementation is better as is)

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: Thank you for your suggestion! I'm agree with your suggestion.
While this approach eliminates the ability to specify specific subnets and security groups explicitly, I believe it would be appropriate to address any requests as they arise over time.

My only concern is regarding the settings for the security group. If only the `vpc` is specified as an argument, there will be a need to automatically generate a security group for the ENI used by bedrock. For the outbound rules of this security group, is it acceptable to allow access to S3 via HTTPS to 0.0.0.0/0?

If the strictest outbound rules are required, the settings would be as follows:
- When using an interface endpoint: Allow HTTPS to the security group attached to the S3 interface endpoint.
- When using a gateway endpoint, NAT gateway or assigning public IP to the ENI: Allow HTTPS to the S3 prefix list.

(But I don't want to implement it...)

badmintoncryer:
@lpizzinidev Thank you for your continuous review! 
I have updated to use only vpc as props and I will fix CI failure later.

I have 1 question about security groups.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,479 @@
...
+  private readonly integrationPattern: sfn.IntegrationPattern;
+  private _role: iam.IRole;
+  private subnets: ec2.ISubnet[];
+  private securityGroup?: ec2.ISecurityGroup;
+
+  constructor(scope: Construct, id: string, private readonly props: BedrockCreateModelCustomizationJobProps) {
+    super(scope, id, props);
+    this.integrationPattern = props.integrationPattern ?? sfn.IntegrationPattern.REQUEST_RESPONSE;
+    this.subnets = props.vpc?.selectSubnets().subnets ?? [];
+    this.securityGroup = props.vpc ? new ec2.SecurityGroup(this, 'SecurityGroup', {
```
lpizzinidev, line 210: 🤔 Now I see the problem of just using the `vpc` property.
We don't allow much control over security groups to the users.
I guess we need to switch back using the `IBedrockCreateModelCustomizationJobVpcConfig` interface (please also add validation on `securityGroups` and `subnets` array lengths).

lpizzinidev (changes requested):
https://github.com/aws/aws-cdk/pull/29043#discussion_r1563818708

> My only concern is regarding the settings for the security group. If only the vpc is specified as an argument, there will be a need to automatically generate a security group for the ENI used by bedrock. For the outbound rules of this security group, is it acceptable to allow access to S3 via HTTPS to 0.0.0.0/0?

We should allow users control over `securityGroups` configuration.
Which is why I think we should revert to using the previous interface.

> If the strictest outbound rules are required

Let's keep it simple for now as adding stricter rules will probably require adding more props.

The rest of the changes look good 👍

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,497 @@
...
+   * @see https://docs.aws.amazon.com/bedrock/latest/userguide/encryption-custom-job.html
+   *
+   * @default - no encryption
+   */
+  readonly customModelKmsKey?: kms.IKey;
+
+  /**
+   * A name for the resulting custom model.
+   *
+   * The maximum length is 63 characters and it needs to satisfy the regular expression ^([0-9a-zA-Z][_-]?)+$.
```
lpizzinidev, line 128:
```suggestion
   * The maximum length is 63 characters.
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,497 @@
...
+   * @see https://docs.aws.amazon.com/bedrock/latest/userguide/custom-models-hp.html
+   *
+   * @default - use default hyperparameters
+   */
+  readonly hyperParameters?: { [key: string]: string };
+
+  /**
+   * A name for the fine-tuning job.
+   *
+   * The maximum length is 63 characters and it needs to satisfy the regular expression ^[a-zA-Z0-9](-*[a-zA-Z0-9\+\-\.])*$.
```
lpizzinidev, line 153:
```suggestion
   * The maximum length is 63 characters.
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,497 @@
...
+}
+
+/**
+ * VPC configuration
+ */
+export interface IBedrockCreateModelCustomizationJobVpcConfig {
+  /**
+   * VPC configuration security groups
+   *
+   * The maximum number of security groups is 5.
```
lpizzinidev, line 70:
```suggestion
   * The minimum number of security groups is 1.
   * The maximum number of security groups is 5.
```
Let's specify both boundaries.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,497 @@
...
+   */
+  readonly securityGroups: ec2.ISecurityGroup[];
+
+  /**
+   * VPC configuration subnets
+   *
+   * The maximum number of subnets is 16.
```
lpizzinidev, line 77:
```suggestion
   * The minimum number of subnets is 1.
   * The maximum number of subnets is 16.
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/README.md
```diff
@@ -355,6 +358,61 @@ const task = new tasks.BedrockInvokeModel(this, 'Prompt Model', {
...
+    prefix: 'training-data', // optional
+  },
+  validationData: [
+    {
+      bucket: validationBucket,
+      prefix: 'validation-data1', // optional
+    },
+  ],
+  vpcConfig: {
+    vpc,
```
lpizzinidev, line 409:
```suggestion
```

lpizzinidev (approved):
Thanks 👍 Left comments for some final minor adjustments.
Otherwise, this looks good to me 💪

badmintoncryer:
@lpizzinidev I'm sorry for my misunderstanding. I've removed `vpc` props from `vpcConfig`. 
I apologize for the time it took to make the corrections in this PR. I look forward to continuing to work with you.

lpizzinidev:
@badmintoncryer 
No worries, you're pretty quick in implementing changes 😄 
Hopefully, this will get merged soon 💪

comcalvi (changes requested):
This is a fantastic change! Shoutout to @lpizzinidev for reviewing this PR!

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,491 @@
...
+              actions: ['s3:GetObject', 's3:PutObject', 's3:ListBucket'],
+              resources: [
+                this.props.outputData.bucket.bucketArn,
+                `${this.props.outputData.bucket.bucketArn}/*`,
+              ],
+            }),
+          ],
+        }),
+      },
+    });
```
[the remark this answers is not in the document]
badmintoncryer: The `grant` method of the `BucketBase` class is private, and we need to use the `grantXxx` methods such as `grantRead`. However, all of the `grantXxx` methods allow for more actions than necessary, making it impossible to replicate the existing code.
What should be done?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,491 @@
...
+    return role;
+  }
+
+  private createVpcConfigPolicyStatement(): iam.PolicyStatement[] {
+    const vpcConfig = this.props.vpcConfig;
+    if (!vpcConfig) {
+      return [];
+    }
+
+    return [
```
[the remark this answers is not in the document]
badmintoncryer: I've added doc links.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,491 @@
...
+              resource: 'security-group',
+              resourceName: securityGroup.securityGroupId,
+            }),
+          ),
+          ...vpcConfig.subnets.map(
+            (subnet) => Stack.of(this).formatArn({
+              service: 'ec2',
+              resource: 'subnet',
+              resourceName: subnet.subnetId,
+            }),
```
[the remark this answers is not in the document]
badmintoncryer: I cannot explain the specific reasons, but I am following the official documentation. https://docs.aws.amazon.com/bedrock/latest/userguide/vpc-model-customization.html#vpc-data-access-role

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,491 @@
...
+          'ec2:DeleteNetworkInterfacePermission',
+        ],
+        resources: ['*'],
+        conditions: {
+          ArnEquals: {
+            'ec2:Subnet': vpcConfig.subnets.map(
+              (subnet) => Stack.of(this).formatArn({
+                service: 'ec2',
+                resource: 'subnet',
+                resourceName: subnet.subnetId,
```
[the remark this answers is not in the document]
badmintoncryer: This implementation also follows [the official documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/vpc-model-customization.html#vpc-data-access-role), but is there a need for any modifications?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,491 @@
...
+          },
+          'ForAllValues:StringEquals': {
+            'aws:TagKeys': [
+              'BedrockManaged',
+              'BedrockModelCustomizationJobArn',
+            ],
+          },
+        },
+      }),
+    ];
```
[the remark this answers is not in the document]
badmintoncryer: My understanding is that it means to allow tag creation only at the time of ENI creation. This is also in accordance with an example from [the official documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/vpc-model-customization.html#vpc-data-access-role).

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: I think Bedrock is executed outside of a VPC. Therefore, Bedrock does not create VPC resources if a VPC configuration is not provided. 
If we want to more strictly manage access to S3 resources, I believe that specifying a VPC configuration will result in the creation of an ENI within the VPC for Bedrock to access. 

Please feel free to correct me if my understanding is incorrect.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,491 @@
...
+            resourceName: '*',
+          }),
+        ],
+      }),
+      new iam.PolicyStatement({
+        actions: ['iam:PassRole'],
+        resources: [this._role.roleArn],
+      }),
+    ];
+    return policyStatements;
```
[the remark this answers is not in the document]
badmintoncryer: In implementing sfn.TaskStateBase, it is necessary to generate iam.PolicyStatement[] and assign it to this.taskPolicies. In this case, I think it would be difficult to utilize the grant methods. What do you think?

badmintoncryer:
@comcalvi Thank you for your review and I'm sorry for my confusing implementation. I've replied your comments.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,491 @@
...
+    this.validateArrayLength('securityGroups', 1, 5, props.vpcConfig?.securityGroups);
+    this.validateArrayLength('subnets', 1, 16, props.vpcConfig?.subnets);
+
+    validatePatternSupported(this.integrationPattern, BedrockCreateModelCustomizationJob.SUPPORTED_INTEGRATION_PATTERNS);
+
+    this._role = this.renderBedrockCreateModelCustomizationJobRole();
+    this.taskPolicies = this.renderPolicyStatements();
+
+    if (this.props.customModelKmsKey) {
+      this.props.customModelKmsKey.addToResourcePolicy(new iam.PolicyStatement({
```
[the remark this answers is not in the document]
badmintoncryer: If we use grant method like below, it will add resource policy to BedrockRole instead of key policy.
In [this document](https://docs.aws.amazon.com/bedrock/latest/userguide/encryption-custom-job.html), I think it would be preferable to implement it according to the documentation. What do you think?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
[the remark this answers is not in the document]
comcalvi: Could you put that doc link here above these values, and also a quick summary of them like you gave here? This helps future contributors understand why these permissions are here.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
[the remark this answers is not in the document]
comcalvi: Please link those docs here, it will help make this clearer

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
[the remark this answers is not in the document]
comcalvi: Can we instead refactor this long list of policy statements to a private method, something like `grantBedrockPermissions`, and then call that in the constructor?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
comcalvi: >Bedrock does not create VPC resources if a VPC configuration is not provided.

So if you provide `vpcConfig` Bedrock create a new VPC?

>Please feel free to correct me if my understanding is incorrect.

I'm trying to understand how it works myself, the documentation doesn't make it clear. Can you test it out and see if providing `vpcConfig` creates a new VPC?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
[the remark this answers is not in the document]
comcalvi: unfortunately I think the grants aren't going to support these permissions, so we can do it like this.

comcalvi (commented):
I think the policy statements are in good shape, the last piece for me is the VPC; does Bedrock create one for you if you provide `VpcConfig`?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: @comcalvi Sorry for the unclear sentence. 
I believe Bedrock does not create a new VPC but instead creates a new ENI within the existing VPC. When a vpcConfig is provided, Bedrock searches for an existing VPC and creates an ENI in the specified subnet, then attaches the provided security group.

badmintoncryer: @comcalvi Could you confirm above message?

[Here is a reference document](https://docs.aws.amazon.com/bedrock/latest/userguide/vpc-model-customization.html#vpc-config) for creating an ENI.

> When you specify the VPC subnets and security groups for a job, Amazon Bedrock creates elastic network interfaces (ENIs) that are associated with your security groups in one of the subnets. ENIs allow the Amazon Bedrock job to connect to resources in your VPC.

comcalvi: got it, thanks.

comcalvi (commented):
Sorry for the lateness of this review! This looks almost ready to merge.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
comcalvi: @badmintoncryer what do you think of implementing the change @lpizzinidev suggested; turning `VpcConfig` into an `IVpc`? Since Bedrock doesn't create one for you, I don't see the harm.

comcalvi: @badmintoncryer what do you think of implementing the change @lpizzinidev suggested; turning VpcConfig into an IVpc?

comcalvi (changes requested):

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
```diff
@@ -0,0 +1,491 @@
...
+      throw new Error(`${name} must be between ${min} and ${max} items long, got: ${value.length}`);
+    }
+  }
+
+  /**
+   * Provides the Bedrock CreateModelCustomizationJob service integration task configuration
+   *
+   * @internal
+   */
+  protected _renderTask(): any {
```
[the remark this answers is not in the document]
badmintoncryer: I couldn't find Cfn type corresponding to this return value. 
This is the [pararmeter of the AWS API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_CreateModelCustomizationJob.html) and I think there is no CFN type.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/create-model-customization-job.ts
badmintoncryer: @comcalvi 

Changing the argument to `IVpc` is not an issue!

As mentioned above, there is only concern about the security group attached to the ENI.
When using `IVpc` as an argument, a security group needs to be generated automatically. In this case, is it acceptable to allow all outbound rules?

comcalvi (dismissed):

comcalvi:
approved but putting do-not-merge on it until the p0 esbuild issue is fixed

badmintoncryer:
@comcalvi I've resolved conflicts. Could you approve again after esbuild issue is fixed?

PR #29064, opened 2024-02-10
feat(appsync): `environmentVariables` property for GraphqlApi

go-to-k:
### Reason for this change

AppSync now supports environment variables in GraphQL resolvers and functions. It would be good for `GraphqlApi` construct to have the property.

- https://aws.amazon.com/jp/about-aws/whats-new/2024/02/aws-appsync-environment-variables-graph-ql-resolvers-functions/
- https://docs.aws.amazon.com/en_en/appsync/latest/devguide/environmental-variables.html
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-graphqlapi.html#cfn-appsync-graphqlapi-environmentvariables

### Description of changes

The `environmentVariables` property is added to `GraphqlApi` construct. To add environment variables after the initiation, we can use `addEnvironmentVariables` method.

Both unit and integ tests.

### Checklist

----


packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
```diff
@@ -442,6 +442,20 @@ export interface GraphqlApiProps {
    * @default IntrospectionConfig.ENABLED
    */
   readonly introspectionConfig?: IntrospectionConfig;
+
+  /**
+   * A map containing the list of resources with their properties and environment variables.
+   *
+   * There are a few rules you must follow when creating keys and values:
+   *   - Keys must begin with a letter.
+   *   - Keys must be between 2 and 64 characters long.
```
go-to-k, line 451: These requirements are based on [the reference](https://docs.aws.amazon.com/en_en/appsync/latest/devguide/environmental-variables.html#configuring-environmental-variables-console) and the CFn error messages.

- Map keys must satisfy constraint:
  - Member must have length less than or equal to 64
  - Member must have length greater than or equal to 2
  - Member must satisfy regular expression pattern: ^[A-Za-z]+\w*$]

```
 AppSyncEnvironmentVariables failed: Error: The stack named AppSyncEnvironmentVariables failed creation, it may need to be manually deleted from the AWS console: ROLLBACK_COMPLETE: 1 validation error detected: Value '{1|2|3=non-empty-2, EnvKey1=non-empty-1}' at 'environmentVariables' failed to satisfy constraint: Map keys must satisfy constraint: [Member must have length less than or equal to 64, Member must have length greater than or equal to 2, Member must satisfy regular expression pattern: ^[A-Za-z]+\w*$]
```

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
```diff
@@ -442,6 +442,20 @@ export interface GraphqlApiProps {
...
+   *   - Keys can only contain letters, numbers, and the underscore character (_).
+   *   - Values can be up to 512 characters long.
+   *   - You can configure up to 50 key-value pairs in a GraphQL API.
```
go-to-k, line 454: These requirements are based on [the reference](https://docs.aws.amazon.com/en_en/appsync/latest/devguide/environmental-variables.html#configuring-environmental-variables-console) and the CFn error messages.

- Map keys must satisfy constraint:
  - Member must have length less than or equal to 64
  - Member must have length greater than or equal to 2
  - Member must satisfy regular expression pattern: ^[A-Za-z]+\w*$]

```
 AppSyncEnvironmentVariables failed: Error: The stack named AppSyncEnvironmentVariables failed creation, it may need to be manually deleted from the AWS console: ROLLBACK_COMPLETE: 1 validation error detected: Value '{1|2|3=non-empty-2, EnvKey1=non-empty-1}' at 'environmentVariables' failed to satisfy constraint: Map keys must satisfy constraint: [Member must have length less than or equal to 64, Member must have length greater than or equal to 2, Member must satisfy regular expression pattern: ^[A-Za-z]+\w*$]
```

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
```diff
@@ -847,6 +870,40 @@ export class GraphqlApi extends GraphqlApiBase {
     return true;
   }
 
+  /**
+   * This method adds an environment variable.
```
go-to-k, line 874:
```suggestion
   * Add an environment variable to the construct.
```

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
```diff
@@ -847,6 +870,40 @@ export class GraphqlApi extends GraphqlApiBase {
...
+    }
+    if (!Token.isUnresolved(value) && value.length > 512) {
+      throw new Error(`Value for '${key}' is too long. Values can be up to 512 characters long, got ${value.length}`);
+    }
+
+    this.environmentVariables[key] = value;
+  }
+
+  private validateEnvironmentVariables() {
+    this.node.addValidation( {
```
kaizencc, line 891: why add validation to the node rather than just throwing an error here? I'm not too familiar with this method but usually we want to default to fail as early as possible and i wonder if this violates that. did you see this done elsewhere?

kaizencc (commented):
Hi @go-to-k! I just have one question really. the rest looks good

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
go-to-k: The `addValidation` (see [DESIGN_GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#attaching-lazy-validators)) method is a lazy way to delay and make decisions about processing that cannot be determined during initialization by the constructor.

For example, in this case, we get an array-type property from props and initialize an instance variable in the constructor. However, there are cases where we want to pass nothing in props and add it later with the `addEnvironmentVariable` method. In that case, if we check the number of the array in the constructor, it will be assumed to be zero.
Because of this case, we should use this method for lazy checking instead of early checking at the constructor.

Also, if we were to check the number of the array in the `addEnvironmentVariable` method, it would throw an error the moment the maximum number is reached, so even if we specify 60 elements for the array, we would get a message `Only 50 environment variables can be set, got 51`. The `addValidation` method is also necessary to check exactly how many elements the user has specified for the array.

Elsewhere, I saw this in CodePipeline.

https://github.com/go-to-k/aws-cdk/blob/a21731c9b1f34525433fbee668ebb2321debf5b8/packages/aws-cdk-lib/aws-codepipeline/lib/pipeline.ts#L542

go-to-k:
@kaizencc 

Thanks for your review. I have returned your comment and would like you to confirm it.

onlybakam:
This looks great. can you add a check to make sure that environment variables are not set on a MERGED api?

go-to-k:
@onlybakam 

Thanks, nice catch. I will add the change not to set the environment variables on the Merged API.

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
go-to-k: I made it a little clearer.

https://github.com/aws/aws-cdk/pull/29064/commits/1a100b5318e3d0a828cfaa41467745ad5502d3a9

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
```diff
@@ -847,6 +870,39 @@ export class GraphqlApi extends GraphqlApiBase {
     return true;
   }
 
+  /**
+   * Add an environment variable to the construct.
+   */
+  public addEnvironmentVariable(key: string, value: string) {
+    if (this.definition.sourceApiOptions) {
+      throw new Error('Environment variables are not supported for merged APIs');
+    }
```
go-to-k, line 879: Because Environment variables are not supported for merged API. The Cfn errors: 

```
MergedAPI (MergedAPI08D3EAD1) Operation not allowed, Merged API resources are read-only (Service: AmazonDeepdish; Status Code: 400; Error Code: BadRequestException; Request ID: ...
```

packages/aws-cdk-lib/aws-appsync/lib/graphqlapi.ts
kaizencc: thanks for the clarification. i agree that this is the right thing to do here now

kaizencc (approved):

go-to-k:
@kaizencc 

Thanks for your approval, but the merge has failed because of `Queue: Embarked in merge queue`.

Could you please trigger again?

kaizencc (approved):

PR #29073, opened 2024-02-11
fix(defaultStackSynthesizer): prevent App level synthesizer from propagating assets across stacks

### LINKED ISSUE #29071: DefaultStackSynthesizer: App level synthesizer leaks assets between Stacks when specified
### Describe the bug

I am working on a multi-stack App with multiple specified account targets (via the env property on each Stack). When specifying the property defaultStackSynthesizer on the App I started facing deployment issues I had never faced before. The output implied that the specific Stack I was trying to deploy wanted to publish assets belonging to other stacks.

By looking at the manifest for each Stack I was able to verify this. This issue does not present itself if you have all Stacks configured to deploy to the same account. I assume the same goes if they are all account agnostic. While the deployment failures are not irrelevant I think the underlying issue of the wrong assets being added to manifests is the key here.

### Expected Behavior

The Stack manifests only contain relevant files.

### Current Behavior

The Stack manifests contains assets belonging to other stacks. Screenshot of manifest for Stack attached to App after another Stack:
image

Note how it not only contains its own template but also the template for StackA.

### Reproduction Steps

1. Create a new project using "npx cdk init app --language=typescript"
2. Modify the app to instantiate a DefaultStackSynthesizer, no props needed
3. Add two Stacks that use this App as scope
4. Synth using "npx cdk synth"
5. Check Stack manifests under cdk.out

I have also taken the liberty of creating a repository which highlights the issue: [CDK Assets leak](https://github.com/castodius/cdk-assets-leak). The cdk.out directory is commited and this [file](https://github.com/castodius/cdk-assets-leak/blob/main/cdk.out/StackB.assets.json) highlights the issue with more than one template being present.

### Possible Solution

I have located the source of the issue in DefaultStackSynthesizer. It stems from reusableBind
image

It copies the DefaultStackSynthesizer, however this only partially works since the class instance contains another class instance and the reference is kept. The reference to the assetManifest is kept and all Stacks end up sharing the same AssetManifestBuilder.
image

This snippet of code highlights why this is an issue:
```javscript
const helper = {
  value: 1
}

const wrapper = {
  foo: helper
}

// prints 1
console.log(wrapper.foo.value)

const copy1 = Object.create(wrapper)
copy1.foo.value = 2

// prints 2 since the reference to helper was kept
console.log(wrapper.foo.value)
```

I will create a PR soon that should fix this.

### CDK CLI Version

2.127.0 (build 6c90efc)

### Framework Version

_No response_

### Node.js Version

v20.10.0

### OS

Sonoma 14.2.1

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

castodius:
Issue 29071
Closes #29071

Reason for this change
The method reusableBind on DefaultStackSynthesizer which is intended to copy the current DefaultStackSynthesizer ends up returning an instance with a reference to the AssetManifestBuilder instance. All stacks end up sharing the same asset manager and additional assets end up in the wrong stacks.

Description of changes
The method reusableBind now instantiates a new AssetManifestBuild eliminating the problem.

Description of how you validated changes
I have added a unit test which fails if my changes are not included. In addition I took the questionable cowboy approach of verifying the code change by changing minified code for default-stack-synthesizer.ts in a test project.

Checklist
[ x ] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)


castodius:
I was not sure where to place the unit test, so I simply placed it under the only describe available. Perhaps it belongs in a new describe together with [this test](https://github.com/aws/aws-cdk/pull/29073/files#diff-cb705f974596f7c9bd3feb213cee2b3ee0205b47ef8c8ff49d0d17a85b209904R503) specifically for "app level synthesizer".

PR #29074, opened 2024-02-11
fix(autoscaling-hooktargets): passing parent construct id to the t…

### LINKED ISSUE #29010: aws-autoscaling-hooktargets: "Topic" already exists (can only use FunctionHook once per stack)
### Describe the bug

If you try to use `FunctionHook` more than once in a stack, or you use it once but try to attach it to more than one lifecycle event, you get this error:

> Error: A subscription with id "Topic" already exists under the scope mystack/lambdaAutoScaling

### Expected Behavior

I expected all lifecycle events to be attached to the Lambda function hook without any errors.

### Current Behavior

The stack fails to synth with an error saying the name `Topic` is used more than once.

### Reproduction Steps

```
const lambdaAutoScaling = new NodejsFunction(this, 'lambdaAutoScaling', {});
const autoScalingGroup = new AutoScalingGroup(this, 'example', {});

autoScalingGroup.addLifecycleHook('lambda-onLaunch', {
  lifecycleTransition: LifecycleTransition.INSTANCE_LAUNCHING,
  notificationTarget: new FunctionHook(lambdaAutoScaling),
});

autoScalingGroup.addLifecycleHook('lambda-onTerminate', {
  lifecycleTransition: LifecycleTransition.INSTANCE_TERMINATING,
  notificationTarget: new FunctionHook(lambdaAutoScaling),
});
```
The problem still exists if you create only one `FunctionHook` and pass it to both `addLifecycleHook()` calls.

### Possible Solution

The problem appears to be in [lambda-hook.ts:28](https://github.com/aws/aws-cdk/blob/0e6e37e2bb7c4e014e0ae8e0b61f523477af5dcf/packages/aws-cdk-lib/aws-autoscaling-hooktargets/lib/lambda-hook.ts#L28) where it creates an SNS topic with the hard-coded name `Topic`.  So you can only create a single one or the name conflicts.

I guess using something like `cdk.Names.nodeUniqueId()` is needed here?

I'm using a loop to create four hooks - launch and terminate, for AMD and ARM architectures.  So even if the events were somehow combined into an array, I'd still have to call `addLifecycleHook()` once for each architecture's auto scaling group.

### CDK CLI Version

2.126.0 (build fb74c41)

### Framework Version

_No response_

### Node.js Version

v21.6.0

### OS

Arch Linux

### Language

TypeScript

### Language Version

TypeScript (5.3.3)

### Other information

_No response_

ahmetsoykan:
…opic which will be created with addLifecycleHook function of ASG.

### 29010

Closes #29010.

### Description of changes

passing parent construct id to the topic which can be created after using addLifecycleHook function of ASG. If this function called twice, this is causing resource id collision in the same stack.

### Checklist

----


ahmetsoykan:
Clarification Request 
about these pipeline checks: 
- request-cli-integ-test / cli-changes : this step is failing but i dont know it's related to changes in here.
- AWS CodeBuild us-east-1: two of the integration tests can't be updated locally, because of it's dependency of public r53 hosted zone with the name of example.com. it fails with this error: `example.com is reserved by AWS!`

PR #29075, opened 2024-02-12
feat(elasticloadbalancingv2): health check interval greater than timeout

### LINKED ISSUE #29062: aws-ecs-patterns: Health Check Interval > Timeout Fails On Deploy
### Describe the bug

When configuring ```ApplicationLoadBalancedFargateService targetGroup.configureHealthCheck```, it is possible to have deployment fail with "Health check interval must be greater
than the timeout.". This should fail during synthesis as the values are known at that time.

### Expected Behavior

I expected my deployment to succeed, or the error to be caught during synthesis.

### Current Behavior

Stack was synthesized successfully, but the deployment failed.

### Reproduction Steps

Create a new ApplicationLoadBalancedFargateService:
```
import * as ecsp from "aws-cdk-lib/aws-ecs-patterns";
const imageProps = {
            image: ecs.ContainerImage.fromRegistry(ADDSOMEIMAGEHERE),
            containerPort: 3000,
        };
        const fargateProps = {
            assignPublicIp: true,
            cpu: 512,
            desiredCount: 1,
            memoryLimitMiB: 1024,
            publicLoadBalancer: true,
            taskImageOptions: imageProps,
        };
const ecsService = new ecsp.ApplicationLoadBalancedFargateService(
            this,
            parentId + "Server",
            fargateProps,
        );
```

Then apply the following:
```
ecsService.targetGroup.configureHealthCheck({
            timeout: cdk.Duration.seconds(60),
        })
```

### Possible Solution

Synthesis should check the interval of the health check is larger than the timeout.

_No response_

### CDK CLI Version

2.126.0

### Framework Version

_No response_

### Node.js Version

18.18.2

### OS

macOS 14.3.1

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

msambol:
Closes #29062.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-rds/lib/proxy.ts
+++ b/packages/aws-cdk-lib/aws-rds/lib/proxy.ts
@@ -104,7 +104,8 @@ export class ProxyTarget {
 
     const engineFamily = engine.engineFamily;
     if (!engineFamily) {
-      throw new Error(`Engine '${engineDescription(engine)}' does not support proxies`);
+      throw new Error('RDS proxies require an engine family to be specified on the database cluster or instance. ' +
+        `No family specified for engine '${engineDescription(engine)}'`);
     }
 
     // allow connecting to the Cluster/Instance from the Proxy
--- a/packages/aws-cdk-lib/aws-rds/test/proxy.test.ts
+++ b/packages/aws-cdk-lib/aws-rds/test/proxy.test.ts
@@ -203,7 +203,7 @@ describe('proxy', () => {
         vpc,
         secrets: [new secretsmanager.Secret(stack, 'Secret')],
       });
-    }).toThrow(/Engine 'mariadb-10\.0\.24' does not support proxies/);
+    }).toThrow(/RDS proxies require an engine family to be specified on the database cluster or instance. No family specified for engine 'mariadb-10\.0\.24'/);
   });
 
   test('correctly creates a proxy for an imported Cluster if its engine is known', () => {


msambol:
Exemption request: I'm adding a validation for parameters. Shouldn't need an integration test here.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-target-group.ts
```diff
@@ -351,6 +352,21 @@ export abstract class TargetGroupBase extends Construct implements ITargetGroup
...
+
+  protected validateHealthCheck(): string[] {
+    const ret = new Array<string>();
+
+    const intervalSeconds = this.healthCheck.interval?.toSeconds();
+    const timeoutSeconds = this.healthCheck.timeout?.toSeconds();
+
+    if (intervalSeconds && timeoutSeconds) {
+      if (intervalSeconds < timeoutSeconds) {
+        ret.push('Health check interval must be greater than the timeout; received interval ' +
```
tmokmss, line 364: > must be greater than 

shouldn't the validation be `intervalSeconds <= timeoutSeconds` then? A link to the doc (in PR comment) stating this constraint will help.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-target-group.ts
```diff
@@ -351,6 +352,21 @@ export abstract class TargetGroupBase extends Construct implements ITargetGroup
...
     return ret;
   }
+
+  protected validateHealthCheck(): string[] {
+    const ret = new Array<string>();
+
+    const intervalSeconds = this.healthCheck.interval?.toSeconds();
+    const timeoutSeconds = this.healthCheck.timeout?.toSeconds();
+
+    if (intervalSeconds && timeoutSeconds) {
```
tmokmss, line 362: Will this work as expected if either intervalSeconds or timeoutSeconds equals to zero?

tmokmss (commented):
LGTM overall, with a few minor comment.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-target-group.ts
msambol: added a link to the doc. should be `<`.

tmokmss: Thanks for the link.

> Timeout
> This value must be less than the Interval value.

So we shouldn't allow the case when `intervalSeconds == timeoutSeconds`. Then the condition should be `intervalSeconds <= timeoutSeconds`

tmokmss: The PR introduced the validation https://github.com/aws/aws-cdk/pull/26031 also described the requirement as below:

> Ensure that HealthCheckTimeoutSeconds is not greater than HealthCheckIntervalSeconds.

So I guess it was just a mistake?

Ideally we should try the configuration (interval==timeout) both for ALB and NLB and it emits the same error.

msambol: Take a look here: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-elasticloadbalancingv2/test/nlb/target-group.test.ts#L386-L409. We have a specific test to not throw an error when they're the same. Confusing!

@pahud Is this something you could ask the AWS team about? It would be nice to have this added to the AWS documentation, and then we can decide which way the code needs to be.

tmokmss: There was actually a similar discussion in the PR: https://github.com/aws/aws-cdk/pull/26031#discussion_r1237324560

So I guess we should keep the validation `intervalSeconds < timeoutSeconds`, and update the comment to e.g. `must be greater than or equal to`.

msambol: I changed back to `<` and added a note. Thanks, @tmokmss 👍🏼

tmokmss (approved):
Thanks, LGTM

msambol:
@kaizencc could I have a review here please ?

packages/aws-cdk-lib/aws-elasticloadbalancingv2/test/nlb/target-group.test.ts
```diff
@@ -144,6 +144,25 @@ describe('tests', () => {
     }).toThrow(/Health check interval '3' not supported. Must be between 5 and 300./);
   });
 
+  test('Throws error for health check interval less than timeout', () => {
```
paulhcsun, line 147: Could you add an edge case test for `interval == timeout`?

paulhcsun (commented):
Code changes look good, just left one comment about adding an additional test for the edge case.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/test/nlb/target-group.test.ts
msambol: I added a test for this. See [discussion above](https://github.com/aws/aws-cdk/pull/29075#discussion_r1501803750) on why we chose to allow them to be equal.

paulhcsun (approved):
Nice work @msambol! Thanks for the contribution. And thank you @tmokmss for the review!

PR #29076, opened 2024-02-12
fix(amplify): add the default amplify policy to the auto-generated role

### LINKED ISSUE #28986: (amplify-alpha): The default created role is empty
### Describe the bug

When I create an amplify app, the associated role has no policies.

### Expected Behavior

If I do the same thing in the web interface it creates a role with the managed policy `AdministratorAccess-Amplify` attached

### Current Behavior

The role has no policies

### Reproduction Steps

```
    const app = new amplify.App(this, 'WebApp', {
      appName: 'web',
      platform: amplify.Platform.WEB_COMPUTE,
      sourceCodeProvider: new amplify.GitHubSourceCodeProvider({
        owner,
        repository,
        oauthToken,
      }),
});
```

### Possible Solution

I've worked around this by manually creating and passing in a role

_No response_

### CDK CLI Version

2.125.0

### Framework Version

_No response_

### Node.js Version

20.11.0

### OS

Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

johnf:
### Issue # (if applicable)

Closes #28986 

### Reason for this change

The default created role for the Amplify App has no managed policy applied. So it can't do things like pull details of the backend config.

### Description of changes

I've attached the Managed policy ```AdministratorAccess-Amplify``` which is what happens when you create an Amplify application in the console.

I've added unit tests.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/aws-amplify-alpha/test/integ.app.js.snapshot/cdk-amplify-app.template.json
+++ b/packages/@aws-cdk/aws-amplify-alpha/test/integ.app.js.snapshot/cdk-amplify-app.template.json
@@ -84,6 +84,7 @@
     "CustomRules": [
      {
       "Source": "/source",
+      "Status": "302",
       "Target": "/target"
      }
     ],
--- a/packages/@aws-cdk/aws-amplify-alpha/test/integ.app.js.snapshot/tree.json
+++ b/packages/@aws-cdk/aws-amplify-alpha/test/integ.app.js.snapshot/tree.json
@@ -141,7 +141,8 @@
                     "customRules": [
                       {
                         "source": "/source",
-                        "target": "/target"
+                        "target": "/target",
+                        "status": "302"
                       }
                     ],
                     "iamServiceRole": {
--- a/packages/@aws-cdk/aws-amplify-alpha/test/integ.app.ts
+++ b/packages/@aws-cdk/aws-amplify-alpha/test/integ.app.ts
@@ -31,6 +31,9 @@ class TestStack extends Stack {
     amplifyApp.addCustomRule({
       source: '/source',
       target: '/target',
+      // NOTE: This is optional according to the API but Cloudformation is breaking without it
+      // Resource handler returned message: "Invalid request provided: Status field in rewrite custom rules should not be empty
+      status: amplify.RedirectStatus.TEMPORARY_REDIRECT,
     });
 
     const mainBranch = amplifyApp.addBranch('main');


johnf:
I've also updated the customRule integ to set a status.
Even though according to the docs status isn't required, cloudformation errors if it isn't provided

packages/@aws-cdk/aws-amplify-alpha/lib/app.ts
```diff
@@ -224,6 +224,7 @@ export class App extends Resource implements IApp, iam.IGrantable {
 
     const role = props.role || new iam.Role(this, 'Role', {
       assumedBy: new iam.ServicePrincipal('amplify.amazonaws.com'),
+      managedPolicies: [iam.ManagedPolicy.fromAwsManagedPolicyName('AdministratorAccess-Amplify')],
```
kaizencc, line 227: I'm down to add common sense permissions to this role, but I balk at the idea of adding this managed policy. We typically don't like adding managed policies to our roles in CDK, and instead opt for fine grained controls (so our users know exactly what policy they are adding.

As always, if _you_ want to use AdministratorAccess-Amplify, you can easily add it in as your role. But I don't want to push that on to everyone.

kaizencc:
Hmm I've had an instant change of heart and will source another opinion :). Stay with me @johnf

packages/@aws-cdk/aws-amplify-alpha/lib/app.ts
johnf: That's precisely the conversation I wanted to have in the issue before creating a PR :)
I can try to develop a smaller set of permissions, i.e., the minimum required to deploy a basic app. But I might miss some key things I'm not using myself.
The main thing missing for me was permission to talk to the backend to grab the amplify configuration.

johnf:
> I'm not sure if we want to add this set of permissions to the role.

I think it's key that something is added to support amplify apps. Even if we define some prebaked permissions that someone can use or at a minimum add something to the documentation.

packages/@aws-cdk/aws-amplify-alpha/lib/app.ts
kaizencc: That makes sense and I'm wary of my initial disdain towards the black box of the managed policy. Just need someone to confirm w me.

johnf:
@kaizencc Any more thoughts on this?
I need to update for conflicts but only want to spend the time if this or something similar is going to be approved

PR #29085, opened 2024-02-13
feat(ecs): credentialSpecs in ContainerDefinitionOptions

cresvi:
### Issue # (if applicable)

Closes #N/A

### Reason for this change

v2.127.0 updated the L1 construct for AWS::ECS::TaskDefinition, adding support for the property ContainerDefinitions.CredentialSpecs, [see](https://github.com/aws/aws-cdk/pull/29053). This PR adds support for CredentialSpecs property in the L2 construct used by `Ec2TaskDefinition.addContainer` method.

### Description of changes

Added property in L2 construct, updated unit test and added integration test.

- [x] Unit test updated and validated
- [x] Integration test added and validated

### Checklist

----


packages/aws-cdk-lib/aws-ecs/README.md
```diff
@@ -623,6 +623,26 @@ taskDefinition.addContainer('windowsservercore', {
...
+### Using Windows authentication with gMSA
+
+Amazon ECS supports Active Directory authentication for Linux containers through a special kind of service account called a group Managed Service Account (gMSA). For more details, please see the [product documentation on how to implement on Windows containers](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/windows-gmsa.html), or this [blog post on how to implement on  Linux containers](https://aws.amazon.com/blogs/containers/using-windows-authentication-with-gmsa-on-linux-containers-on-amazon-ecs/).
+
+```ts
+declare const taskExecutionRole: iam.Role;
+declare const taskDefinition: ecs.TaskDefinition;
+
+// A task execution role is needed, and it should have permissions to read from the S3 bucket or SSM parameter where the CredSpec file is stored.
+taskDefinition.executionRole = taskExecutionRole
```
kaizencc, line 635:
```suggestion
taskDefinition.executionRole = taskExecutionRole;
```

packages/aws-cdk-lib/aws-ecs/README.md
```diff
@@ -623,6 +623,26 @@ taskDefinition.addContainer('windowsservercore', {
...
+
+taskDefinition.addContainer('gmsa-container', {
+  image: ecs.ContainerImage.fromRegistry("amazon/amazon-ecs-sample"),
+  cpu: 128,
+  memoryLimitMiB: 256,
+  credentialSpecs: ['credentialspecdomainless:arn:aws:s3:::bucket_name/key_name'], 
+  // Valid values are: 'credentialspecdomainless:ARN' or 'credentialspec:ARN'
```
kaizencc, line 642: An idea here. What if we exposed two properties, `credentailSpecs` and `credentialSpecsDomainless`, both which just takes the `ARN` component (to be clearly documented in the docstring). And then we can append the correct prefix before sending it to ecs. This will save users from having to do the appending themselves, right?

kaizencc (changes requested):
Hi @cresvi, thanks for opening this PR. I have a slight change in the API that I think would help. Lemme know what you think

cresvi:
> ### AWS CodeBuild CI Report
> * CodeBuild project: AutoBuildv2Project1C6BFA3F-wQm2hXv2jqQv
> * Commit ID: [7edc755](https://github.com/aws/aws-cdk/commit/7edc7550c7cc2fb2fa33a01abfc8ee531c1ac84c)
> * Result: FAILED
> * [Build Logs](https://2cl995f30a.execute-api.us-east-1.amazonaws.com/Prod/buildlogs?key=c62083fc-33fa-49e4-b477-5e98d6b9652c%2Fbuild.log) (available for 30 days)
> 
> _Powered by [github-codebuild-logs](https://serverlessrepo.aws.amazon.com/applications/arn:aws:serverlessrepo:us-east-1:277187709615:applications~github-codebuild-logs), available on the [AWS Serverless Application Repository](https://aws.amazon.com/serverless/serverlessrepo/)_


I'm not sure why the integration test failed, as when I run it locally it succeeds. Am I missing something?

CleanShot 2024-02-14 at 14 10 53@2x

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
```diff
@@ -0,0 +1,139 @@
...
+   * Get the ARN for a SSM parameter.
+   */
+  protected static arnForSsmParameter(parameter: IParameter) {
+    return parameter.parameterArn;
+  }
+
+  /**
+   * Prefix string based on the type of CredSpec.
+   */
+  public prefixId: string;
```
kaizencc, line 29:
```suggestion
  public readonly prefixId: string;
```

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Location or ARN from where to retrieve the CredSpec file.
+   */
+  public fileLocation: string;
```
kaizencc, line 34:
```suggestion
  public readonly fileLocation: string;
```

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * Credential specification (CredSpec) file.
+ */
+export class DomainJoinedCredentialSpec extends CredentialSpec {
+  /**
+   * Prefix Id for this type of CredSpec.
+   */
+  public static readonly PREFIX_ID = 'credentialspec';
```
kaizencc, line 63: do we need this as a public variable? don't think it needs to be

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * Credential specification for domainless gMSA.
+ */
+export class DomainlessCredentialSpec extends CredentialSpec {
+  /**
+   * Prefix Id for this type of CredSpec.
+   */
+  public static readonly PREFIX_ID = 'credentialspecdomainless';
```
kaizencc, line 98: same here. I think we can just put the raw string in the constructor

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    return bucket.arnForObjects(key);
+  }
+
+  /**
+   * Get the ARN for a SSM parameter.
+   */
+  protected static arnForSsmParameter(parameter: IParameter) {
+    return parameter.parameterArn;
+  }
```
kaizencc, line 24: I was initially confused by this section. Can we document both clearly that they are only intended to be used by extensions of `CredentialSpec` as helper methods?

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
```diff
@@ -0,0 +1,139 @@
...
+   * Loads the CredSpec from a SSM parameter.
+   *
+   * @param parameter The SSM parameter
+   * @returns CredSpec with it's locations set to the SSM parameter's ARN.
+   */
+  public static fromSsmParameter(parameter: IParameter) {
+    return new DomainlessCredentialSpec(CredentialSpec.arnForSsmParameter(parameter));
+  }
+
+  constructor(fileLocation: string) {
```
kaizencc, line 121:
```suggestion
  public constructor(fileLocation: string) {
```

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
```diff
@@ -0,0 +1,139 @@
...
+   * Loads the CredSpec from a SSM parameter.
+   *
+   * @param parameter The SSM parameter
+   * @returns CredSpec with it's locations set to the SSM parameter's ARN.
+   */
+  public static fromSsmParameter(parameter: IParameter) {
+    return new DomainJoinedCredentialSpec(CredentialSpec.arnForSsmParameter(parameter));
+  }
+
+  constructor(fileLocation: string) {
```
kaizencc, line 86:
```suggestion
  public constructor(fileLocation: string) {
```

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Location or ARN from where to retrieve the CredSpec file.
+   */
+  public fileLocation: string;
+
+  /**
+   * @param fileLocation Location or ARN from where to retrieve the CredSpec file
+   */
+  constructor(prefixId: string, fileLocation: string) {
```
kaizencc, line 39:
```suggestion
  public constructor(prefixId: string, fileLocation: string) {
```

packages/aws-cdk-lib/aws-ecs/README.md
kaizencc: I think this is exactly what we want and this exact care to concoct purpose-driven APIs is what makes CDK L2s powerful. People who want the raw `string` format + other bare bones types use L1s.

packages/aws-cdk-lib/aws-ecs/README.md
```diff
@@ -623,6 +623,23 @@ taskDefinition.addContainer('windowsservercore', {
...
+```ts
+// Make sure the task definition's execution role has permissions to read from the S3 bucket or SSM parameter where the CredSpec file is stored.
+declare const bucket: s3.Bucket;
+declare const taskDefinition: ecs.TaskDefinition;
+
+taskDefinition.addContainer('gmsa-container', {
+  image: ecs.ContainerImage.fromRegistry("amazon/amazon-ecs-sample"),
+  cpu: 128,
+  memoryLimitMiB: 256,
+  credentialSpecs: [ecs.DomainlessCredentialSpec.fromS3Bucket(bucket, 'credSpec')],
```
kaizencc, line 639: can we mention somewhere that there are options for `DomainJoinedCredentialSpec` and also `fromSssmParameter`. Don't need examples for all, but just document their availability after this example

packages/aws-cdk-lib/aws-ecs/lib/container-definition.ts
```diff
@@ -126,6 +127,15 @@ export interface ContainerDefinitionOptions {
    */
   readonly command?: string[];
 
+  /**
+   * A list of ARNs in SSM or Amazon S3 to a credential spec (`CredSpec`) file that configures the container for Active Directory authentication.
+   *
+   * We recommend that you use this parameter instead of the `dockerSecurityOptions`. The maximum number of ARNs is 1.
```
kaizencc, line 133: > The maximum number of ARNs is 1

I know you're pulling this directly from the docs themselves, but this is confusing to me. It sounds like the max number of ARNs this `credentialSpecs` property takes is 1, and then why type it as a string at all?

I see two scenarios:

a) I misunderstand. In that case, lets make the docs a bit clearer.
b) The max number of ARNs for now is 1, but might be more in the future. Lets document that if thats the decision behind the array type.

But given that we aren't doing any synth time checks for this limit of 1, I think we are in scenario a)...

kaizencc (dismissed):
Hi @cresvi, I think what you've got now is much, much better than typing things as `string[]`. Just a few more comments here and there.

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
cresvi: done

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
cresvi: done, moved to constructor

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
cresvi: done

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
cresvi: done, moved to constructor

packages/aws-cdk-lib/aws-ecs/README.md
cresvi: Expanded documentation to give more information and show sample of the four use cases

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
cresvi: done

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
cresvi: done

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
cresvi: done

packages/aws-cdk-lib/aws-ecs/lib/credential-spec.ts
cresvi: Updated a more detailed description

packages/aws-cdk-lib/aws-ecs/lib/container-definition.ts
cresvi: I also think the way the engineer team implemented this, but in the API the `credentialSpecs` accepts an array of strings. Still they only use the first one...
I updated the comment a little to make it clearer. I don't want to limit the L2 construct to only 1 CredSpec because if later the service team decide to use the rest of the entries, it will be a breaking change for customers.

packages/aws-cdk-lib/aws-ecs/lib/container-definition.ts
```diff
@@ -538,6 +553,14 @@ export class ContainerDefinition extends Construct {
       }
     }
 
+    if (props.credentialSpecs) {
```
kaizencc, line 556: Based off of a previous comment, we should have a synth-time check (+ unit test) for ensuring that `credentialSpecs.length === 1`. We never want to silently ignore additional info

packages/aws-cdk-lib/aws-ecs/lib/container-definition.ts
```diff
@@ -126,6 +127,15 @@ export interface ContainerDefinitionOptions {
    */
   readonly command?: string[];
 
+  /**
+   * A list of ARNs in SSM or Amazon S3 to a credential spec (`CredSpec`) file that configures the container for Active Directory authentication.
+   *
+   * We recommend that you use this parameter instead of the `dockerSecurityOptions`. Only the first entry on this array is used. This may be expanded in the future.
```
kaizencc, line 133:
```suggestion
   * We recommend that you use this parameter instead of the `dockerSecurityOptions`.
   * Currently, only one credential spec is allowed per container definition.
```

kaizencc (changes requested):
One last comment @cresvi :)

packages/aws-cdk-lib/aws-ecs/lib/container-definition.ts
cresvi: done

packages/aws-cdk-lib/aws-ecs/lib/container-definition.ts
cresvi: I wanted to avoid the need to create a new PR if a new feature starts using more than one CredSpec in the future. But if you really insist, I made the change and added the appropriate unit test.

kaizencc (approved):

PR #29087, opened 2024-02-13
feat(cli): support CloudFormation simplified resource import

### LINKED ISSUE #28060: cli: support CloudFormation simplified resource import
### Describe the feature

On Nov 17 2023, a new feature was announced: AWS CloudFormation simplifies resource import with a new parameter for ChangeSets.

https://aws.amazon.com/about-aws/whats-new/2023/11/aws-cloudformation-import-parameter-changesets/

Summary:

AWS CloudFormation's new ImportExistingResources parameter for CreateChangeSet API automatically imports existing resources in the template during deployments. It identifies resources using custom names, avoiding errors from naming conflicts. This reduces manual effort as a separate resource name document is no longer needed. ImportExistingResources allows importing existing resources and deploying changes in a single ChangeSet.

### Use Case

Automatically import exsting resources with the same physical name, such as S3 bucket, DDB table, etc, during a CFn deployment. 

Because resource import is a vital feature for CDK users e.g. to refactor a construct tree, it would benefit many potential users if cdk natively support it.

### Proposed Solution

This feature introduced a new request parameter [`ImportExstingResources`](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateChangeSet.html#:~:text=ImportExistingResources) to the `CreateChangeSet` API. So I guess we can add a CLI option like `--import-exsting-resources: boolean` to `cdk deploy` command to enable the new feature.

The API is called here:
https://github.com/aws/aws-cdk/blob/c66e197f6f8840da6475383dbf2421c3b06ea417/packages/aws-cdk/lib/api/deploy-stack.ts#L410-L418

### Other Information

_No response_

### Acknowledgements

- [x] I may be able to implement this feature request

### CDK version used

2.110.0

macOS

tmokmss:
### Issue # (if applicable)

Closes #28060.

### Reason for this change

This feature allows to automatically import exsting resources with the same physical name, such as S3 bucket, DDB table, etc, during a CFn deployment.

Because resource import is a vital feature for CDK users e.g. to refactor a construct tree, cdk migrate, etc, it would benefit many potential users if cdk natively support it.

### Description of changes

This PR adds a CLI option --import-exsting-resources: boolean to cdk deploy command and pass it to createChangeSet API call.

Added a cli integ test.

### Checklist

----

### DIFF
--- a/packages/aws-cdk/README.md
+++ b/packages/aws-cdk/README.md
@@ -401,20 +401,20 @@ unmanaged resources in your account.
 Reduces the manual effort of import operations and avoids
 deployment failures due to naming conflicts with unmanaged resources in your account.
 
-Use `--method=prepare-change-set` flag to review which resources are imported or not before deploying
+Use `--method=prepare-change-set` flag to review which resources are imported or not before deploying a changeset.
 You can inspect the change set created by CDK from the management console or other external tools.
 
 ```console
 $ cdk deploy --import-existing-resources --method=prepare-change-set
 ```
 
-Use the `--exclusively` flag to enable this feature for a specific stack
+Use the `--exclusively` flag to enable this feature for a specific stack.
 
 ```console
 $ cdk deploy --import-existing-resources --exclusively StackName
 ```
 
-Only resources that have custom names can be imported using `--import-existing-resources`
+Only resources that have custom names can be imported using `--import-existing-resources`.
 For more information, see [name type](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-name.html). 
 To import resources that do not accept custom names, such as EC2 instances,
 use the `cdk import` instead. 


tmokmss:
Exemption Request: I added a CLI integ test instead of framework-integ, because this is a change of CLI.

TheRealAmazonKendra:
I've been thinking about this one a LOT and been back and forth about what I think the right solution for it is. One issue I have with just adding this flag to `deploy` is the amount of flags we now have on this command that only work with some/none of the other flags. I'm worried that it's getting too messy.

I've also been thinking that with the interplay between `cdk migrate`, `cdk import`, `cfn-include`, and this feature, there's a lot of weird messy overlap that I'd like to potentially clean up a bit. Maaaaaaybe this should actually be added as a flag on `cdk import` instead? 

Right now `cdk import` has some kind of weird limitations that were due to unresolved issues on CloudFormation's end that I got resolved for `cdk migrate` so maybe the right path forward is fixing those oddities and then adding this flag there.

tmokmss:
@TheRealAmazonKendra Thanks. I agree with that both `cdk migrate` and `cdk import` commands can internally use the CloudFormation simplified import feature for better UX. But I also think we should definitely add this flag to `cdk deploy` command as the lowest level API to access the feature. 

The fact is that users currently cannot use the feature at all, even though the feature has been available since last year, because there is no way to add argument from outside to `createChangeSet` API called by `deploy` command.

The feature itself is quite simple, and in most use cases, users should not need fancy API around it.

tmokmss:
@SankyRed Thanks! I guess we also need to add `pr-linter/exempt-integ-test` label?

packages/aws-cdk/lib/api/deploy-stack.ts
```diff
@@ -416,6 +424,7 @@ class FullCloudFormationDeployment {
       ResourcesToImport: this.options.resourcesToImport,
       Description: `CDK Changeset for execution ${this.uuid}`,
       ClientToken: `create${this.uuid}`,
+      ImportExistingResources: importExistingResources,
```
comcalvi, line 427: what happens if they specify both `ResourcesToImport` and `ImportExistingResources`?

comcalvi (dismissed):
we probably want this feature on diff changesets as well... @TheRealAmazonKendra what do you think?

packages/aws-cdk/lib/api/deploy-stack.ts
tmokmss: I guess it will not work, but [the spec](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateChangeSet.html) says nothing about such restriction, so I would like to keep it as-is. 

Actually users cannot set both ImportExistingResources and ResourcesToImport from the CDK CLI, because the former is only used in cdk deploy, and cdk deploy does not have any command argument to directly set ResourcesToImport.

rachit-shah:
@comcalvi @tmokmss @TheRealAmazonKendra 
We use CDK deployment and face resource conflicts often. We'd like to avoid calling CFN APIs directly since we are using CDK app. 

What exactly is this PR blocked on? It has been months since the PR was raised. If it's adding this feature in a different cdk operation (cdk diff), can that be a separate PR/issue?

tmokmss:
@comcalvi Hi, I resolved the merge conflict. Can you take a look at this once again?

packages/aws-cdk/lib/api/deploy-stack.ts
HBobertz: The spec isn't really the source of truth we all wish it was so it's not uncommon we enforce logical things the spec does not. However in this specific instance I believe this is correct because there is a difference between the resources supported by `importExistingResources` and `ResourcesToImport`. Auto-import is a subset of resources supported by `ResourcesToImport` so it is entirely possible to have both. I believe this is fine to leave as is

packages/aws-cdk/README.md
```diff
@@ -388,6 +388,40 @@ $ cdk deploy --method=prepare-change-set --change-set-name MyChangeSetName
...
+You can pass the `--import-existing-resources` flag to the `deploy` command:
+
+```console
+$ cdk deploy --import-existing-resources
+```
+
+This allows you to automatically import the resources in your template that already
+exist in your AWS account during CloudFormation deployments.
+With this feature, you can reduce the manual effort of import operations and avoid
+deployment failures because of naming conflicts.
```
HBobertz, line 402:
```suggestion
Automatically import resources in your CDK application which represent
unmanaged resources in your account. 
Reduces the manual effort of import operations and avoids
deployment failures due to naming conflicts with unmanaged resources in your account
```

HBobertz: adding some active voice to be more similar to how our other command readme's are written.

packages/aws-cdk/README.md
```diff
@@ -388,6 +388,40 @@ $ cdk deploy --method=prepare-change-set --change-set-name MyChangeSetName
...
+
+To review which resources are imported or not before actually executing a change set,
+use `--method=prepare-change-set` flag.
```
HBobertz, line 405:
```suggestion
use `--method=prepare-change-set` flag to review which resources are imported or not before deploying
```

packages/aws-cdk/README.md
```diff
@@ -388,6 +388,40 @@ $ cdk deploy --method=prepare-change-set --change-set-name MyChangeSetName
...
+You can inspect the change set created by CDK from the management console or other external tools.
+
+```console
+$ cdk deploy --import-existing-resources --method=prepare-change-set
+```
+
+To enable this feature only for a specific stack, use `--exclusively` flag.
```
HBobertz, line 412:
```suggestion
Use the `--exclusively` flag to enable this feature for a specific stack
```

packages/aws-cdk/README.md
```diff
@@ -388,6 +388,40 @@ $ cdk deploy --method=prepare-change-set --change-set-name MyChangeSetName
...
+
+```console
+$ cdk deploy --import-existing-resources --exclusively StackName
+```
+
+This parameter can only import resources that have custom names in templates.
```
HBobertz, line 418:
```suggestion
Only resources that have custom names can be imported using `--import-existing-resources`
```

packages/aws-cdk/README.md
```diff
@@ -579,6 +613,11 @@ To import an existing resource to a CDK stack, follow the following steps:
 5. When `cdk import` reports success, the resource is managed by CDK. Any subsequent
    changes in the construct configuration will be reflected on the resource.
 
+NOTE: You can also import existing resources by `--import-existing-resources` option of `cdk deploy` command.
```
HBobertz, line 616:
```suggestion
NOTE: You can also import existing resources by using `--import-existing-resources` flag of the `cdk deploy` command.
```

packages/aws-cdk/README.md
```diff
@@ -579,6 +613,11 @@ To import an existing resource to a CDK stack, follow the following steps:
...
+This parameter only works for resources that you can set custom physical names,
```
HBobertz, line 617:
```suggestion
Only works for resources that allow custom physical names
```

packages/aws-cdk/README.md
```diff
@@ -579,6 +613,11 @@ To import an existing resource to a CDK stack, follow the following steps:
...
+such as S3 bucket, DynamoDB table, etc.
```
HBobertz, line 618:
```suggestion
such as S3 bucket, DynamoDB table, etc...
```

HBobertz (changes requested):

packages/aws-cdk/README.md
HBobertz: Love that you're pushing this change and really appreciate the contribution. I've wanted this functionality for a while so it's great to see it being implemented. Just have some doc changes and it's g2g imo

packages/aws-cdk/test/api/deploy-stack.test.ts
```diff
@@ -880,6 +880,41 @@ describe('disable rollback', () => {
 
 });
 
+describe('import-existing-resources', () => {
```
comcalvi, line 883: Can we get a CLI integ test that creates a single resource and then imports it with this new flag?

comcalvi (dismissed):

packages/aws-cdk/test/api/deploy-stack.test.ts
tmokmss: @comcalvi 
Thanks, but I don't fully agree with implementing such test on CDK side. Because:

1. Such test would more or less complicate the test code, not noly the success case, but also ensuring there is no orphaned resource when test failed halfway through
2. It is not CDK's responsibility to ensure that a resource is successfully imported when the flag is set. It is CFn's responsibility to guarantee the functionality. CDK should just make sure the flag is set properly (as tested in the current code.)

What do you think?

tmokmss:
@HBobertz @comcalvi 

Thanks! I addressed all the comments from you :)

> It's unclear if this flag works with IMPORT type changesets. Those changesets should not create anything, but autoimport likely needs a CREATE to import the resource...maybe IMPORT changesets recognize that?

As far as I tested, CFn decides whether a resource is created or imported when we create a changeset. If a resource will be imported, the changeset type for the resource will become `IMPORT`, and if not, it wil become `CREATE`.

packages/aws-cdk/test/api/deploy-stack.test.ts
HBobertz: > It is not CDK's responsibility to ensure that a resource is successfully imported when the flag is set. It is CFn's responsibility to guarantee the functionality. CDK should just make sure the flag is set properly (as tested in the current code.)

This is true, however ultimately any test involving any cloudformation api usage (such as most CDK deploy tests) also falls under that same umbrella. Even though we don't own cloudformation nor are we responsible to fix any issue blocking our service in cloudformation, it does allow us to report the impact our customer's impact to appropriate CFN teams. Regardless of all that though it's already a just line in the sand we've crossed ages ago, so I believe it's appropriate to add that test

> Such test would more or less complicate the test code, not noly the success case, but also ensuring there is no orphaned resource when test failed halfway through. This possibly increases maintenance cost of the cli integ test code.

This is true, it would be far from the most complex integ test we've written but any integ tests which are verifying deployments tend to require some setup/cleanup so it does get a bit complex. Orphaning is technically possible but as long as you set up a `finally` to manually cleanup you manually created, and the stack rolls back, it should be g2g. You can look at other tests in CDK Deploy or perhaps CDK Migrate for some ideas of how we've done this in the past.

HBobertz: This is ultimately just a very long winded way of me saying I believe the test is appropriate 😅.

tmokmss: @HBobertz Thanks, hmmm but what is the actual value to adding the test? As long as we set the importExistingResources flag, it should result in a successfull deployment, unless CFn is not working properly. If there happens to be the case when CDK fails although CFn is working as expected, that is when we should add a new test for the very case. At this moment, however, I think such test does not add any reliability to the behavior of CDK; it just adds unneessary complexity to the cli integ test code base.

comcalvi (commented):
LGTM, pending a CLI test run.

tmokmss:
@comcalvi Thanks! Just for confirmation, have you triggered a cli test? I see the request is still pending. (not really sure how it works though)
image

tmokmss:
@comcalvi Hi, let me know the integ-test result when you get a chance :) I guess I don't have access to it.

kaizencc:
@tmokmss I mean to help you get this PR across the finish line. first step was to fix the merge conflicts, which are a little bit difficult. I did it locally, but I'm not able to push to this branch. Is there any reason that is the case on your end? It makes me sad because I spent some time trying to fix up the branch for you :(.

i can also start a new branch and we can go from there but im first wondering if there's something quick you can do on your end to ensure that i have push access to your fork.

tmokmss:
Hi @kaizencc, thanks for working on this and sorry I didn't realize it was in conflict.

I confirmed that `Allow edits and access to secrets by maintainers' is checked, so I think you should be able to push commits. Not sure about the error, but if this is a blocker, it's okay to create a new branch and PR. I'm excited to see this finally merged :)

kaizencc:
@tmokmss i can't figure it out :/ so i just went ahead and created a new PR to main from your PR. I think you'll still get credit, but i also want to add a coauthor message to the PR description just in case. I'll need an email address associated to your github account to make that happen though.

tmokmss:
@kaizencc thanks for your kindness! You can use this address:
tmokmss@users.noreply.github.com

kaizencc:
https://github.com/aws/aws-cdk/pull/32676 was merged

PR #29101, opened 2024-02-14
feat(cloudfront-origin): adding FunctionUrlOrigin

watany-dev:
### Reason for this change

`origins.httpOrigin` requires a trick to specify the Lambda Functions URL. 

### Description of changes

This eases that pain and simplifies integration.

### Checklist

----


packages/aws-cdk-lib/aws-cloudfront-origins/lib/function-url-origin.ts
```diff
@@ -0,0 +1,49 @@
...
+import * as lambda from '../../aws-lambda';
+import * as cdk from '../../core';
+
+/**
+ * Properties for a Lambda Function URL Origin.
+ */
+export interface FunctionUrlOriginProps extends cloudfront.OriginProps {
+  /**
+   * Specifies how long, in seconds, CloudFront waits for a response from the origin.
+   * The valid range is from 1 to 180 seconds, inclusive.
```
colifran, line 12: I'm seeing the max read timeout as 60 seconds https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-customoriginconfig.html#cfn-cloudfront-distribution-customoriginconfig-originreadtimeout. Where does the 180 seconds come from?

packages/aws-cdk-lib/aws-cloudfront-origins/lib/function-url-origin.ts
```diff
@@ -0,0 +1,49 @@
...
+   *
+   * @default cdk.Duration.seconds(30)
+   */
+  readonly readTimeout?: cdk.Duration;
+
+  /**
+   * Specifies how long, in seconds, CloudFront persists its connection to the origin.
+   * The valid range is from 1 to 180 seconds, inclusive.
```
colifran, line 20: Same as above.

colifran (changes requested):
This looks great! I just have one small question.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/function-url-origin.ts
watany-dev: This is in reference to this code; I added this because Note was missing.

https://github.com/aws/aws-cdk/blob/b4daf84cf8a90ef38e3a2328df6271bdaf19de01/packages/aws-cdk-lib/aws-cloudfront-origins/lib/http-origin.ts#L37-L58

packages/aws-cdk-lib/aws-cloudfront-origins/lib/function-url-origin.ts
watany-dev: same.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/function-url-origin.ts
colifran: Thanks for the clarification!

colifran (approved):
Nice job!

PR #29134, opened 2024-02-16
feat(codepipeline): pipeline type v2 option for CDK codepipeline (#29134)

vumdao:
### Issue # (if applicable)

Add CDK codepipeline V2 option of pipeline type

### Reason for this change

Some case pipelineType "V2" is more cost effective.


gossandr:
I think the title here may be wrong. This is a feature of pipelines (the "opinionated construct") and not aws-codepipeline.

PR #29144, opened 2024-02-17
feat(sns): enforce ssl with topic policy

### LINKED ISSUE #29142: SNS: Enforce SSL with Prop
### Describe the feature

I'd like the ability to enforce SSL on an SNS topic in a similar fashion as the Bucket construct. The Bucket construct has a property, `enforceSSL`, that will automatically update the bucket policy and enforce SSL. I'd like something similar that will automatically update the topic policy.

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Bucket.html#enforcessl

### Use Case

In the `AwsSolutionsChecks` within cdk-nag, there is a rule that requires SSL on SNS Topics, `AwsSolutions-SNS3`.  Given that is a recommended practice, I believe setting an SNS Topic's policy so that it requires SSL would be a typical scenario. Since it's a typical scenario, I'd like a property to do it for me rather than having to write up a TopicPolicy every time.

https://github.com/cdklabs/cdk-nag/blob/main/RULES.md

### Proposed Solution

```
new Topic(this, 'MyTopic', {
    enforceSSL: true,
});
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.126.0

macOS 14.3

msambol:
Adds a statement to match the document in the [docs](https://docs.aws.amazon.com/sns/latest/dg/sns-security-best-practices.html#enforce-encryption-data-in-transit):

```
{
  "Id": "ExamplePolicy",
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "AllowPublishThroughSSLOnly",
      "Action": "SNS:Publish",
      "Effect": "Deny",
      "Resource": [
        "arn:aws:sns:us-east-1:1234567890:test-topic"
      ],
      "Condition": {
        "Bool": {
          "aws:SecureTransport": "false"
        }
      },
      "Principal": "*"
    }
  ]
}
```

Closes #29142.

----

### DIFF
--- a/packages/aws-cdk-lib/aws-sns/test/sns.test.ts
+++ b/packages/aws-cdk-lib/aws-sns/test/sns.test.ts
@@ -186,7 +186,7 @@ describe('Topic', () => {
 
   });
 
-  test('can add a policy to the topic enforcing ssl', () => {
+  test('can enforce ssl when creating the topic', () => {
     // GIVEN
     const stack = new cdk.Stack();
     const topic = new sns.Topic(stack, 'Topic', {


packages/aws-cdk-lib/aws-sns/lib/policy.ts
```diff
@@ -51,9 +60,31 @@ export class TopicPolicy extends Resource {
...
+  protected createSslPolicyDocument(topicArn: string): PolicyStatement {
+    return new PolicyStatement ({
+      sid: 'AllowPublishThroughSSLOnly',
+      actions: ['sns:Publish'],
+      effect: Effect.DENY,
+      resources: [topicArn],
+      conditions: {
+        Bool: { 'aws:SecureTransport': 'false' },
+      },
+      principals: [new AnyPrincipal()],
```
msambol, line 87: The SNS docs go back and forth but I _think_ this should be `AnyPrincipal` not `StarPrincipal`.

packages/aws-cdk-lib/aws-sns/lib/topic-base.ts
```diff
@@ -122,21 +122,43 @@ export abstract class TopicBase extends Resource implements ITopic {
...
+  protected createSslPolicyDocument(): iam.PolicyStatement {
+    return new iam.PolicyStatement ({
+      sid: 'AllowPublishThroughSSLOnly',
+      actions: ['sns:Publish'],
+      effect: iam.Effect.DENY,
+      resources: [this.topicArn],
+      conditions: {
+        Bool: { 'aws:SecureTransport': 'false' },
+      },
+      principals: [new iam.AnyPrincipal()],
```
msambol, line 158: Likewise here.. The SNS docs go back and forth but I _think_ this should be `AnyPrincipal` not `StarPrincipal`.

packages/aws-cdk-lib/aws-sns/lib/policy.ts
msambol: I changed this to star to match this specific [doc](https://docs.aws.amazon.com/sns/latest/dg/sns-security-best-practices.html#enforce-encryption-data-in-transit).

packages/aws-cdk-lib/aws-sns/lib/topic-base.ts
```diff
@@ -122,21 +136,43 @@ export abstract class TopicBase extends Resource implements ITopic {
    * Adds a statement to the IAM resource policy associated with this topic.
    *
    * If this topic was created in this stack (`new Topic`), a topic policy
-   * will be automatically created upon the first call to `addToPolicy`. If
+   * will be automatically created upon the first call to `addToResourcePolicy`. If
    * the topic is imported (`Topic.import`), then this is a no-op.
    */
-  public addToResourcePolicy(statement: iam.PolicyStatement): iam.AddToResourcePolicyResult {
+  public addToResourcePolicy(statement: iam.PolicyStatement, addToResourcePolicyProps?: AddToResourcePolicyProps): iam.AddToResourcePolicyResult {
```
colifran, line 142: I don't think adding a new function argument here is the right way to do this. The `addToResourcePolicy(statement: iam.PolicyStatement)` is a pattern used throughout the code base. I think a better solution is to just make `enforceSSL` a protected attribute of the base class:
`protected enforceSSL?: boolean`. We can then add `enforceSSL` as a property on `TopicProps` and set it in the constructor. So:

```ts
export interface TopicProps {
	// ...
	enforceSSL?: boolean;
}

export class Topic extends TopicBase {
	// ...
	constructor(scope: Construct, id: string, props: TopicProps = {}) {
		// ...
		this.enforceSSL = props.enforceSSL;
	}
}
```

Then in `addToResourcePolicy` we just check `this.enforceSSL` and reuse the logic you've added with `createSSLPolicyDocument`.

Thoughts?

colifran (dismissed):
I like where this is going! I left a comment about what I think might be a cleaner solution.

msambol:
@colifran updated. Much cleaner

colifran (approved):
Great job!

PR #29167, opened 2024-02-19
fix(lambda-nodejs): fixing esbuildArgs to take in account re-specified keys

### LINKED ISSUE #25385: @aws-cdk/aws-lambda-nodejs: esbuildArgs alias must use : instead of =
### Describe the bug

I want to reference libs in my layers with `import { somefunc } from '@layer/somelib';`.
When setting the alias build argument for esbuild, it throws an error :

```typescript
const opts = {
            bundling: {
                externalModules: ['@aws-sdk/*', 'aws-lambda', '@layer/*'],
                esbuildArgs: {
                    '--alias': '@layer=',
                },
            },
        };
```

It will generate a bundling step like this: 
```
esbuild --bundle src/index.js  --outfile=dist/index.js --alias=@layer=
✘ [ERROR] Invalid build flag: "--alias=@layer="

  Use "--alias:@layer=" instead of "--alias=@layer=". Flags that can be re-specified multiple times
  use ":" instead of "=".
```

By replacing the `=` with `:` it solves the issue

### Expected Behavior

esbuildArgs key/value needs `:` when it can  be re-specified.

### Current Behavior

esbuildArgs['--alias'] is using `=` which leads to a bundle error.

Full stack:
```
    at AssetStaging.bundle (node_modules/aws-cdk-lib/core/lib/asset-staging.js:2:603)
    at AssetStaging.stageByBundling (node_modules/aws-cdk-lib/core/lib/asset-staging.js:1:4540)
    at stageThisAsset (node_modules/aws-cdk-lib/core/lib/asset-staging.js:1:1901)
    at Cache.obtain (node_modules/aws-cdk-lib/core/lib/private/cache.js:1:242)
    at new AssetStaging (node_modules/aws-cdk-lib/core/lib/asset-staging.js:1:2296)
    at new Asset (node_modules/aws-cdk-lib/aws-s3-assets/lib/asset.js:1:736)
    at AssetCode.bind (node_modules/aws-cdk-lib/aws-lambda/lib/code.js:1:4628)
    at new Function (node_modules/aws-cdk-lib/aws-lambda/lib/function.js:1:2829)
    at new NodejsFunction (node_modules/aws-cdk-lib/aws-lambda-nodejs/lib/function.js:1:1171)
    at new MyConstruct (stack/MyConstruct .ts:18:9)
```

## CDK:

```typescript
new NodejsFunction(this, 'test-function', {
            entry: 'src/functions/test/index.ts',
            architecture: Architecture.ARM_64,
            runtime: Runtime.NODEJS_18_X,
            memorySize: 128,
            tracing: Tracing.ACTIVE,
            handler: 'index.handler',
            timeout: Duration.seconds(30),
            retryAttempts: 0,
            logRetention: RetentionDays.ONE_DAY,
            bundling: {
                minify: process.env.NODE_ENV === 'production',
                banner: 'import { createRequire } from \'module\'; const require = createRequire(import.meta.url);',
                mainFields: ['module', 'main'],
                target: 'node18',
                externalModules: ['@aws-sdk/*', 'aws-lambda', '@layer/*'],
                esbuildArgs: {
                    '--alias': '@layer=',
                },
                format: OutputFormat.ESM,
            },
        });
```

## Lambda 

```
// src/functions/test/index.ts
import { ping } from '@layer/echo';
```

## Layer

```
// dist/layers/echo/index.js
export{console.log as ping};
```

### Possible Solution

esbuildArgs key/value needs `:` when it can  be re-specified; from [documentation](https://esbuild.github.io/api/):

> CLI: If you are using the command-line API, it may be helpful to know that the flags come in one of three forms: --foo, --foo=bar, or --foo:bar. The form --foo is used for enabling boolean flags such as [--minify](https://esbuild.github.io/api/#minify), the form --foo=bar is used for flags that have a single value and are only specified once such as [--platform=](https://esbuild.github.io/api/#platform), and the form --foo:bar is used for flags that have multiple values and can be re-specified multiple times such as [--external:](https://esbuild.github.io/api/#external).

_No response_

### CDK CLI Version

2.76.0 (build 78c411b)

### Framework Version

_No response_

### Node.js Version

Node.js v19.8.1

### OS

WSL2 Ubuntu

### Language

Typescript

### Language Version

Typescript (5.0.4)

### Other information

_No response_

Vandita2020:
### Issue # (if applicable)
Closes #25385 

### Reason for this change
This PR fixes a bug in CDK where CDK does not take into account re-specified keys while doing bundling. The CLI supports flags in one of three forms: `--foo`, `--foo=bar`, or `--foo:bar`. However, the `--foo:bar` form was not initially supported. 
With `--foo:bar`, users can now specify flags that have multiple values and can be re-specified multiple times.

### Description of changes
The code has a list of keys that can be re-specified multiple times. While assigning the flags it checks whether the key is among the list of re-specified keys, if yes, it specifies the flag with `:`.

The PR includes unit test and integration test both to validate the changes.

### Checklist

----


packages/aws-cdk-lib/aws-lambda-nodejs/test/bundling.test.ts
```diff
@@ -245,6 +245,7 @@ test('esbuild bundling with esbuild options', () => {
       '--resolve-extensions': '.ts,.js',
       '--splitting': true,
       '--keep-names': '',
+      '--out-extension': '.js=.mjs',
```
jonife, line 248: why do we need this?

Vandita2020: I have added this to check if the code is working fine. `out-extension` is one of the fields that uses `:`

jonife: so why didn't we add the others here?

Vandita2020: There is a list of esbuildArgs that uses ':' to assign the flags. Here is the [list](https://github.com/aws/aws-cdk/pull/29167/files#diff-d3a78437ce44ac758a33052c1ce663976f465681199c2c85cc42b8b57832b08eR421) with all of them. The reason I chose `--out-extension` to test the code is because it was the one among the list that could be set without modifying much of the test code or without breaking any existing functionality.

jonife:
Great job on fixing the integration test. However it seems you need to rebase your changes on the latest changes in the main branch.

Also can you change the title to:
fix(lambda-nodejs): fixing esbuildArgs to consider re-specified keys. 
Or
fix(lambda-nodejs): fixing Bundling command to take account re-specified keys

This should be clear to the reviewers what problems you are trying to fix.

packages/aws-cdk-lib/aws-lambda-nodejs/lib/bundling.ts
```diff
@@ -418,10 +418,13 @@ function toTarget(runtime: Runtime): string {
 
 function toCliArgs(esbuildArgs: { [key: string]: string | boolean }): string {
   const args = new Array<string>();
+  const ls = ['--alias', '--drop', '--pure', '--log-override', '--out-extension'];
```
valerena, line 421: what does `ls` stand for here? Should this variable have a better/clearer name? 

Also, does it make sense maybe to have this as a top level constant instead of just being a local variable here? 
(I'm not sure about this because I'm not familiar with other functionalities or the structure of this file)

Vandita2020: I have changed the name to `reSpecifiedKeys`, making it more specific to what the list is about. And I have kept the list inside `toCliAgs` function as it is being used inside this function only and for other functions too they have their constants declared inside the Function only.

GavinZZ:
I'm a bit hesitant to approve this PR because I'm not sure if this change will cause a breaking change as technically these new args will come in effect after this PR. And there may be users who used these args in their CDK stack before the PR (and CDK falsely ignore these args, now adding effect to these args cause a behavioural change, although very unlikely).

Will get another team member to review this for a second opinion.

GavinZZ:
> I'm a bit hesitant to approve this PR because I'm not sure if this change will cause a breaking change as technically these new args will come in effect after this PR. And there may be users who used these args in their CDK stack before the PR (and CDK falsely ignore these args, now adding effect to these args cause a behavioural change, although very unlikely).
> 
> Will get another team member to review this for a second opinion.

Dismissing my concern, I didn't notice that before the code change, it will cause an error during deployment. Happy to approve.

GavinZZ:
@mergify update

paulhcsun (approved):

PR #29172, opened 2024-02-19
feat(CLI): improved nested stack diff

comcalvi:
### Reason for this change

The existing nested stack diff places a fake property, `NestedTemplate`, in the templates to be diffed. This prevents displaying resource replacement information in the diff, like we do for top level stacks. This PR does *not* add changeset replacement information from changesets, but it does add replacement information from the spec. 

### Description of changes

Reworked nested stack diff to treat nested stacks as top level stacks. This improves the visual UX and sets us up for using changesets with nested stacks. 

#### Before

Screenshot 2024-02-19 at 1 47 59 PM

#### After
Screenshot 2024-02-19 at 1 48 48 PM

Unit tests + manual tests.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/aws-msk-alpha/README.md
+++ b/packages/@aws-cdk/aws-msk-alpha/README.md
@@ -217,7 +217,8 @@ You can configure an MSK cluster storage mode using the `storageMode` property.
 Tiered storage is a low-cost storage tier for Amazon MSK that scales to virtually unlimited storage,
 making it cost-effective to build streaming data applications.
 
-> Visit [Tiered storage](https://docs.aws.amazon.com/msk/latest/developerguide/msk-tiered-storage.html) for more details.
+> Visit [Tiered storage](https://docs.aws.amazon.com/msk/latest/developerguide/msk-tiered-storage.html)
+to see the list of compatible Kafka versions and for more details.
 
 ```ts
 declare const vpc: ec2.Vpc;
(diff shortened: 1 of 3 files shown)


colifran (approved):
LGTM -- nice job!

packages/aws-cdk/lib/api/nested-stack-helpers.ts
```diff
@@ -5,59 +5,38 @@ import { ISDK } from './aws-auth';
...
 
-export interface NestedStackNames {
-  readonly nestedStackPhysicalName: string | undefined;
-  readonly nestedChildStackNames: { [logicalId: string]: NestedStackNames };
-}
-
-export interface TemplateWithNestedStackCount {
-  readonly deployedTemplate: Template;
-  readonly nestedStackCount: number;
+export interface RootTemplateWithNestedStacks {
```
rix0rrr, line 15: Can `RootTemplateWithNestedStacks` and `NestedStackTemplates` not be the same structure?

comcalvi: I decided not to because the functions that use this interface as a return type only get the generated templates of nested stacks, not the top level root stack. The functions that take these objects as input have a different way of getting the top-level generated template (namely, they take a `StackArtifact` that contains it). They need something special that the `StackArtifact` provides, so I decided to keep the generated template out of this data structure, but only for the top-level artifacts.

We could mark `generatedTemplate` as optional, but that's less clear than using the separate type.

PR #29187, opened 2024-02-20
chore(codebuild): add support for Windows Server Core 2019 3.0 CodeBuild image

awschristou:
### Issue # N/A

Closes # N/A

### Reason for this change

CodeBuild now has a Windows Server Core 2019 3.0 image (reference: [Images](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html#ec2-compute-images)), but this is not available from CDK.

Additionally:
- a missing validation for `WindowsBuildImage.WIN_SERVER_CORE_2019_BASE` was added to the tests

### Description of changes

Add definition for WIN_SERVER_CORE_2019_BASE_3_0 to the CDK.
This change is similar to #22259, which added the Base 2.0 image.

I added a test into `packages/aws-cdk-lib/aws-codebuild/test/project.test.ts`, and ran it using the commands:
```
cd packages/aws-cdk-lib/
yarn test aws-codebuild/
```

The tests passed.

I also validated the change by attempting a deployment that uses it. I created a file in `packages/@aws-cdk-testing/framework-integ/test/aws-codebuild/test/sample.ts` with the contents:

```typescript
import { App, Stack } from 'aws-cdk-lib'
import { BuildSpec, Project, WindowsBuildImage } from 'aws-cdk-lib/aws-codebuild'

const app = new App()
const env = { region: 'us-west-2', account: 'XXXXXX' }

const stack = new Stack(app, 'cc2-cdk-dev-test', { env: env })

new Project(stack, 'sample-win-3-image', {
    buildSpec: BuildSpec.fromObject({
        version: '0.2',
        phases: {
            build: { commands: ['echo "hello world!"'] }
        }
    }),
    environment: { buildImage: WindowsBuildImage.WIN_SERVER_CORE_2019_BASE_3_0 }
})
```

And then I ran

```
cd packages/@aws-cdk-testing/framework-integ
yarn watch
npx cdk -a test/aws-codebuild/test/sample.js deploy
```

This produced a stack containing a CodeBuild project, which successfully runs using the new Base 3.0 image.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-codebuild/test/project.test.ts
+++ b/packages/aws-cdk-lib/aws-codebuild/test/project.test.ts
@@ -797,6 +797,7 @@ describe('Environment', () => {
     ['Standard 6.0', codebuild.LinuxBuildImage.STANDARD_6_0, 'aws/codebuild/standard:6.0'],
     ['Amazon Linux 4.0', codebuild.LinuxBuildImage.AMAZON_LINUX_2_4, 'aws/codebuild/amazonlinux2-x86_64-standard:4.0'],
     ['Amazon Linux 5.0', codebuild.LinuxBuildImage.AMAZON_LINUX_2_5, 'aws/codebuild/amazonlinux2-x86_64-standard:5.0'],
+    ['Windows Server Core 2019 1.0', codebuild.WindowsBuildImage.WIN_SERVER_CORE_2019_BASE, 'aws/codebuild/windows-base:2019-1.0'],
     ['Windows Server Core 2019 2.0', codebuild.WindowsBuildImage.WIN_SERVER_CORE_2019_BASE_2_0, 'aws/codebuild/windows-base:2019-2.0'],
     ['Windows Server Core 2019 3.0', codebuild.WindowsBuildImage.WIN_SERVER_CORE_2019_BASE_3_0, 'aws/codebuild/windows-base:2019-3.0'],
   ])('has build image for %s', (_, buildImage, expected) => {


packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -2033,6 +2033,16 @@ export class WindowsBuildImage implements IBuildImage {
...
+  /**
+   * The standard CodeBuild image `aws/codebuild/windows-base:2019-3.0`, which is
+   * based off Windows Server Core 2019.
+   */
+  public static readonly WIN_SERVER_CORE_2019_BASE_3_0: IBuildImage = new WindowsBuildImage({
+    imageId: 'aws/codebuild/windows-base:2019-3.0',
+    imagePullPrincipalType: ImagePullPrincipalType.CODEBUILD,
+    imageType: WindowsImageType.SERVER_2019,
+  });
+
```
lpizzinidev, line 2045: Can you please also add support for `aws/codebuild/windows-base:2022-1.0`?

packages/aws-cdk-lib/aws-codebuild/test/project.test.ts
```diff
@@ -798,6 +798,7 @@ describe('Environment', () => {
     ['Amazon Linux 4.0', codebuild.LinuxBuildImage.AMAZON_LINUX_2_4, 'aws/codebuild/amazonlinux2-x86_64-standard:4.0'],
     ['Amazon Linux 5.0', codebuild.LinuxBuildImage.AMAZON_LINUX_2_5, 'aws/codebuild/amazonlinux2-x86_64-standard:5.0'],
     ['Windows Server Core 2019 2.0', codebuild.WindowsBuildImage.WIN_SERVER_CORE_2019_BASE_2_0, 'aws/codebuild/windows-base:2019-2.0'],
+    ['Windows Server Core 2019 3.0', codebuild.WindowsBuildImage.WIN_SERVER_CORE_2019_BASE_3_0, 'aws/codebuild/windows-base:2019-3.0'],
```
lpizzinidev, line 801: Can you please also add `aws/codebuild/windows-base:2019-1.0` and `aws/codebuild/windows-base:2022-1.0` to this test?

lpizzinidev (changes requested):
Thanks 👍
I left some comments for adjustments

TheRealAmazonKendra:
Apologies, I didn't look for an existing PR before also making this change. It just merged, but it did not include the Windows 22 version because I didn't see it listed [here](https://docs.aws.amazon.com/codebuild/latest/userguide/available-runtimes.html) so if you would still like to make that change we would definitely welcome it.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
awschristou: `aws/codebuild/windows-base:2022-1.0` requires an alternate configuration -- this requires reserved capacity. This should be added as a dedicated change.

I don't have time to validate that adding `aws/codebuild/windows-base:2022-1.0` works correctly, so I'm not inclined to add it to this change.

PR #29199, opened 2024-02-21
fix(pipelines): generates warning since CDK 2.128.0 due to addition of v2 pipeline support in aws-codepipeline

### LINKED ISSUE #29190: (pipelines): generates warning since CDK 2.128.0 due to addition of v2 pipeline support in aws-codepipeline
### Describe the bug

pipelines is the opinionated construct library for creating codepipelines. while not labeled as such, i think it is a L3 Pattern Construct.

Since CDK 2.128.0, this construct generates a warning because `PipelineType` is not being provided when pipelines creates the aws-codepipeline.

This warning should have been caught before the release of 2.128.0, but instead the test that caught it was changed to effectively check nothing. See https://github.com/aws/aws-cdk/commit/40ffe2ba8c3cd46d4096bbb93210d4aa6952953c#diff-8e15d23929b5ff6c128d73523da6a4322f975856fea30df4f969a21edf5f4eb2

I'm calling this an introduced bug because it breaks when running cdk synth --strict

### Expected Behavior

I expect pipelines to not generate this warning by properly providing `PipelineType` when creating the underlying code pipeline resource. 

### Current Behavior

Currently it generates a warning like this: [Warning at example] V1 pipeline type is implicitly selected when `pipelineType` is not set. If you want to use V2 type, set `PipelineType.V2`. [ack: @aws-cdk/aws-codepipeline:unspecifiedPipelineType]

This warning will cause cdk synth strict to fail, which breaks numerous pipleines of ours based on the pipelines library.

### Reproduction Steps

see example code from the [docs](https://docs.aws.amazon.com/cdk/v2/guide/cdk_pipeline.html)
```ts

import { CodePipeline, CodePipelineSource, ShellStep } from 'aws-cdk-lib/pipelines';

export class MyPipelineStack extends cdk.Stack {
    super(scope, id, props);

    const pipeline = new CodePipeline(this, 'Pipeline', {
      pipelineName: 'MyPipeline',
      synth: new ShellStep('Synth', {
        input: CodePipelineSource.gitHub('OWNER/REPO', 'main'),
        commands: ['npm ci', 'npm run build', 'npx cdk synth']
      })
    });
  }
}
```

### Possible Solution

For example, in this code block: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/pipelines/lib/codepipeline/codepipeline.ts#L467-L480

change it to look something like this:
```ts
    } else {
      this._pipeline = new cp.Pipeline(this, 'Pipeline', {
        pipelineName: this.props.pipelineName,
        pipelineType: cp.PipelineType.V1,
        crossAccountKeys: this.props.crossAccountKeys ?? false,
        crossRegionReplicationBuckets: this.props.crossRegionReplicationBuckets,
        reuseCrossRegionSupportStacks: this.props.reuseCrossRegionSupportStacks,
        // This is necessary to make self-mutation work (deployments are guaranteed
        // to happen only after the builds of the latest pipeline definition).
        restartExecutionOnUpdate: true,
        role: this.props.role,
        enableKeyRotation: this.props.enableKeyRotation,
        artifactBucket: this.props.artifactBucket,
      });
    }
```

_No response_

### CDK CLI Version

2.128.0

### Framework Version

_No response_

### Node.js Version

18.2.0

### OS

macos

### Language

TypeScript

### Language Version

_No response_

### Other information

This was also mentioned in #29146 but this issue is looking to add v2 support. I haven't dug that deep to see how much other things would have to change to support that. This issue is merely looking to resolve the warning introduced. While this warning is suppressible, it isn't intuitive. Pipelines should at a minimum provide the `PipelineType` when it creates the pipeline to prevent this warning from being geneerated. v2 support as mentioned in the other issue would also resolve this, but I assume it takes longer than simply fixing this bug

gossandr:
### Issue # (if applicable)

Closes #29190 

### Reason for this change

CDK v2.128.0 introduced a warning in aws-codepipeline to warn users of the implicit behavior now that v2 pipelines are supported in CDK. This warning can cause established pipelines to fail if they are using cdk synth --strict. The warning can be suppressed, but the better fix is to have this module supply the `PipelineType` added in CDK v128.0 and set it to `v1`. A future change would have to address adding v2 pipeline support to this module, this only resolves the introduced warning. 

### Description of changes

- added the optional (and new) `PipelineType` property in the creation of the codepipeline in https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/pipelines/lib/codepipeline/codepipeline.ts#L467-L480
- slightly modified the README where it mentions using aws-codepipeline to include specifically an mention of v2 as a reason to use aws-codepipeline (until of course v2 is added properly in this lib)

- reverted a change in https://github.com/aws/aws-cdk/commit/40ffe2ba8c3cd46d4096bbb93210d4aa6952953c as I believe this previously caught the new warning and was changed to fix the test rather than looking into the new warning. This would catch the warning if `PipelineType` is not supplied

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
+++ b/packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
@@ -172,6 +172,7 @@ export interface LoggingOptions {
   /**
    * Specify if slow search logging should be set up.
    * Requires Elasticsearch version 5.1 or later or OpenSearch version 1.0 or later.
+   * An explicit `false` is required when disabling it from `true`.
    *
    * @default - false
    */
@@ -187,6 +188,7 @@ export interface LoggingOptions {
   /**
    * Specify if slow index logging should be set up.
    * Requires Elasticsearch version 5.1 or later or OpenSearch version 1.0 or later.
+   * An explicit `false` is required when disabling it from `true`.
    *
    * @default - false
    */
@@ -202,6 +204,7 @@ export interface LoggingOptions {
   /**
    * Specify if Amazon OpenSearch Service application logging should be set up.
    * Requires Elasticsearch version 5.1 or later or OpenSearch version 1.0 or later.
+   * An explicit `false` is required when disabling it from `true`.
    *
    * @default - false
    */


gossandr:
Exemption Request
I don't think updated integration tests are necessary here.  The modification in the props that are supplied to creating the new CodePipeline (from the aws-codepipeline lib) will not produce a change in output. Current behavior was to assume v1 and add warning annotation, now with the property set on creation no warning is generated. Otherwise exactly the same pipeline is created.

gossandr:
> Exemption Request I don't think updated integration tests are necessary here. The modification in the props that are supplied to creating the new CodePipeline (from the aws-codepipeline lib) will not produce a change in output. Current behavior was to assume v1 and add warning annotation, now with the property set on creation no warning is generated. Otherwise exactly the same pipeline is created.

So i think I was wrong and they are required...but i could use some guidance perhaps 😅

kaizencc (dismissed):
Yes @gossandr that would be the correct way to update those tests. it may be quite time-intensive. I see that the snapshot diff is simply explicitly specifying pipeline v1. That is to be expected, and in this case I am happy to assume that CFN accepts the property happily since it is exactly as documented.

In this case, I am happy with you doing a shortcut: 

run the integration tests with `--update-on-failed --dry-run`. This will simply update the snapshots without deploying the integ test. Not great, but I don't think this is a high risk change to the snapshot.

kaizencc (approved):

PR #29207, opened 2024-02-21
fix(lambda-nodejs): support bundling aws-sdk as part of the bundled code asset

jonife:
### Issue # (if applicable)
#25492.

Closes #<issue number here>.
#25492.

### Reason for this change
The BundlingOptions in NodejsFunction construct removes AWS SDK dependencies by default.

This uses Lambda Provided SDK in the resulting function. This has higher cold start than a bundled function with AWS SDK dependencies included.

This happens, because the Node.js runtime has to do module resolution and go through multiple files while reading dependency code in the bundled function which uses Lambda Provided SDK. When SDK in bundled with the function code, the cold starts are lower as the as Node.js runtime has to read single file without any module resolution.

Result from reproduction:

{
'NodejsFunction default (uses Lambda Provided SDK)': 1227.1435,
'NodejsFunction custom (uses Customer Deployed SDK)': 929.441
}

related to this issue: https://github.com/aws/aws-cdk/issues/25492

### Description of changes
While maintaining backward compatibility, an new option `useAwsSDK` was introduced to include the sdk in the code asset

yes kindly refer to the above

Added both unit and integration test 

yes 

```
Running integration tests for failed tests...

Running test /Users/jonife/Documents/dev/lambda-tooling/cdk/aws-cdk/packages/@aws-cdk-testing/framework-integ/test/aws-lambda-nodejs/test/integ.dependencies.js in us-east-1
  SUCCESS    aws-lambda-nodejs/test/integ.dependencies-LambdaDependencies/DefaultTest 329.553s
       AssertionResultsLambdaInvoke5050b1f640cc49956b59f2a71febe95c - success
      AssertionResultsLambdaInvokee35a5227846e334cb95a90bacfbfb877 - success
      AssertionResultsLambdaInvoke7d0602e4b9f40ae057f935d874b5f971 - success

Test Results: 

Tests:    1 passed, 1 total
✨  Done in 337.42s.
```

### Checklist

----


garysassano:
I like the proposal, but I think calling the option `bundleAwsSDK` instead of `useAwsSDK` would make it clearer. The name `useAwsSDK` as a `true`/`false` value doesn't clearly indicate whether you want to use the Lambda Provided SDK or supply your own.

garysassano:
I like the proposal, but I think calling the option `bundleAwsSDK` instead of `useAwsSDK` would make it clearer. The name `useAwsSDK` as a `true`/`false` value doesn't clearly convey whether you want to use the Lambda Provided SDK or supply your own.

packages/aws-cdk-lib/aws-lambda-nodejs/lib/bundling.ts
```diff
@@ -263,7 +265,7 @@ export class Bundling implements cdk.BundlingOptions {
 
       // Create dummy package.json, copy lock file if any and then install
       depsCommand = chain([
-        isPnpm ? osCommand.write(pathJoin(options.outputDir, 'pnpm-workspace.yaml'), ''): '', // Ensure node_modules directory is installed locally by creating local 'pnpm-workspace.yaml' file
+        isPnpm ? osCommand.write(pathJoin(options.outputDir, 'pnpm-workspace.yaml'), '') : '', // Ensure node_modules directory is installed locally by creating local 'pnpm-workspace.yaml' file
```
Vandita2020, line 268: [nit] : Is the space intentional? It is there at couple of places in this file.

GavinZZ (commented):
Same comment as the other PR on the title, changes look good. But we should add the new property to [README](https://github.com/aws/aws-cdk/blob/8d07b857aec64211054e409a098a841ac57d9e09/packages/aws-cdk-lib/aws-lambda-nodejs/README.md#L1)

GavinZZ (approved):
Looks good, thanks for contributing!

gshpychka:
@jonife Wouldn't this be a breaking change for those who included the AWS SDK by setting `externalModules` to an empty array?

PR #29211, opened 2024-02-22
fix: correct .NET namespace naming

vlesierse:
### Issue # (if applicable)

Closes #<issue number here>.

### Reason for this change

Currently the naming of the API Gateway V2 NuGet packages are very different from what is expected as you can see blow in the global using example.
```
global using Amazon.CDK.AWS.APIGateway;
global using Amazon.CDK.AWS.Apigatewayv2;
global using Amazon.CDK.AwsApigatewayv2Authorizers;
global using Amazon.CDK.AwsApigatewayv2Integrations;
```

### Description of changes

Adding and correcting `.jsiirc.json` with the right package names. The new global using will be reflected like this.

```
global using Amazon.CDK.AWS.APIGateway;
global using Amazon.CDK.AWS.APIGatewayV2;
global using Amazon.CDK.AWS.APIGatewayV2.Authorizers;
global using Amazon.CDK.AWS.APIGatewayV2.Integrations;
```

Package CDK and see the right NuGet package names.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/scripts/submodules/index.ts
+++ b/packages/aws-cdk-lib/scripts/submodules/index.ts
@@ -65,7 +65,7 @@ async function ensureSubmodule(submodule: ModuleMapEntry, modulePath: string) {
           package: submodule.definition.javaPackage,
         },
         dotnet: {
-          package: submodule.definition.dotnetPackage,
+          namespace: submodule.definition.dotnetNamespace,
         },
         python: {
           module: submodule.definition.pythonModuleName,
--- a/tools/@aws-cdk/pkglint/lib/library-creation.ts
+++ b/tools/@aws-cdk/pkglint/lib/library-creation.ts
@@ -9,7 +9,7 @@ export interface ModuleDefinition {
   readonly moduleBaseName: string;
   readonly packageName: string;
 
-  readonly dotnetPackage: string;
+  readonly dotnetNamespace: string;
   readonly javaGroupId: string;
   readonly javaPackage: string;
   readonly javaArtifactId: string;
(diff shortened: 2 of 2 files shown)


vlesierse:
> `Exemption Request`

This PR doesn't contain any functional code changes and only covers package naming.

mrgrain:
@kaizencc @xazhao Is this something that's missing in the stabilization docs?

vlesierse:
There are many more packages which aren't correct.
```
Amazon.CDK.AwsArczonealshift
Amazon.CDK.Aws
Amazon.CDK.Aws
Amazon.CDK.Aws
Amazon.CDK.Aws
```

I can expand the PR to include those as well?

vlesierse:
@mrgrain I have pushed the fix for the other packages are well. Apparently in the `_jsiirc.json` files the `dotnet` sections are set to `package` instead of `namespace`

packages/aws-cdk-lib/aws-workspacesthinclient/.jsiirc.json
```diff
@@ -4,7 +4,7 @@
       "package": "software.amazon.awscdk.services.workspacesthinclient"
     },
     "dotnet": {
-      "package": "Amazon.CDK.AWS.WorkSpacesThinClient"
+      "namespace": "Amazon.CDK.AWS.WorkSpacesThinClient"
```
kaizencc, line 7: @vlesierse the best fix for all of these is to actually fix the source that codegens new packages. I assume `package` and `namespace` are configs that do the same thing right?

The file gets generated here: https://github.com/aws/aws-cdk/blob/704f596539372364408b6bff83e21221b6bb0970/packages/aws-cdk-lib/scripts/submodules/index.ts#L54-L75

kaizencc (changes requested):

packages/aws-cdk-lib/aws-workspacesthinclient/.jsiirc.json
vlesierse: Good idea to change this at the source, was missing that. So a package it the NuGet package name, which is correct, but because CDK v2 put everything in the AWS.CDK.Lib package, the module needs to be specified on the namespace.

vlesierse: If you look at the other modules, it is done correctly.

Issue #29165, opened 2024-02-19
wafv2: add_override does not obey Action key


Rondineli:
### Describe the bug

Trying to add a rule to an specific key breaks as Action is not added:

```python
waf_acl = wafv2.CfnWebACL(self, 'id', **params)
# Assuming above we would have only one rule
new_rule = {
    "Action": {
        "Block": {}
    },
    "Name": "Deny_regex",
    "Priority": 1,
    "Statement": {
        "RegexPatternSetReferenceStatement": {
            "Arn": {
                "Fn::ImportValue": "my-regex-arn"
            },
            "FieldToMatch": {
                "SingleHeader": {
                    "Name": "my-header"
                }
            },
            "TextTransformations": [
                {
                    "Priority": 0,
                    "Type": "NONE"
                }
            ]
        }
    },
    "VisibilityConfig": {
        "CloudWatchMetricsEnabled": true,
        "MetricName": "Deny_regex",
        "SampledRequestsEnabled": true
    }
}
waf_acl.add_override('Properties.Rules.1', new_rule)

```

At the final template, the `Action` block is not present:
```json
{
    "Name": "Deny_regex",
    "Priority": 1,
    "Statement": {
        "RegexPatternSetReferenceStatement": {
            "Arn": {
                "Fn::ImportValue": "my-regex-arn"
            },
            "FieldToMatch": {
                "SingleHeader": {
                    "Name": "my-header"
                }
            },
            "TextTransformations": [
                {
                    "Priority": 0,
                    "Type": "NONE"
                }
            ]
        }
    },
    "VisibilityConfig": {
        "CloudWatchMetricsEnabled": true,
        "MetricName": "Deny_regex",
        "SampledRequestsEnabled": true
    }
}
```

### Expected Behavior

Expected the rule json output to have the `Action` block

### Current Behavior

Somehow, the Action json block is not being added to the final template

### Reproduction Steps

Create a waf, and try to use add_override with a plain json rule on wafv2

### Possible Solution

Add Action to the jsii interface mappings? Somehow Action is missing.

### CDK CLI Version

2.127.0 (build 6c90efc)

### Node.js Version

v18.17.0

### OS

MacOs

### Language

Python

### Language Version

Python 3.9.6

pahud:
I guess you should just use `Rules`

This works for me in ts

```ts

export class DummyStack extends Stack {
  constructor(scope: Construct, id: string, props: StackProps) {
    super(scope, id, props);

[...]

PR #29217, opened 2024-02-22
fix(aws-cdk-lib): allow create explicitly empty trees on aws-cdk-lib core

Rondineli:
### Issue # (if applicable)

Closes #[29165](https://github.com/aws/aws-cdk/issues/29165).

### Reason for this change

Using `addOverride` calls `deepMerge`, and few edge cases like wafv2 rules needs Actions and other keys with objects and empty objects.

### Description of changes

Changing the code to be more realistic with the comment, where only null data is deleted

No tests, just making sure the current change would be compatible with current tests, should not break anything

### Checklist

----


jeffersonfrazer:
+1

packages/aws-cdk-lib/core/lib/cfn-resource.ts
```diff
@@ -654,7 +664,6 @@ function deepMerge(target: any, ...sources: any[]) {
       if (key === '__proto__' || key === 'constructor') {
         continue;
       }
-
```
paulhcsun, line 657: can you add back the linebreak here

packages/aws-cdk-lib/core/lib/cfn-resource.ts
```diff
@@ -640,6 +640,16 @@ const MERGE_EXCLUDE_KEYS: string[] = [
   'Fn::Or',
 ];
 
+// Availabe wafv2 empty objects
+// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-wafv2-webacl-ruleaction.html
```
paulhcsun, line 644:
```suggestion
/**
 * Availabe WAFv2 empty objects
 *
 * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-wafv2-webacl-ruleaction.html
 */
```

packages/aws-cdk-lib/core/lib/cfn-resource.ts
```diff
@@ -640,6 +640,16 @@ const MERGE_EXCLUDE_KEYS: string[] = [
   'Fn::Or',
 ];
 
+// Availabe wafv2 empty objects
```
paulhcsun, line 643: This description is vague and doesn't explain what these are used for, nor how "empty objects" relates to `DELETE_MERGE_EXCLUDE_KEYS`. Maybe provide a short snippet of context, you can reference the description for `MERGE_EXCLUDE_KEYS` on line 617

packages/aws-cdk-lib/core/test/resource.test.ts
```diff
@@ -541,6 +541,44 @@ describe('resource', () => {
       });
     });
 
+    test('addOverride(p, {object: {object: {}}}) will respect empty trees', () => {
```
paulhcsun, line 544: prefer different wording like `support` over `respect` here

packages/aws-cdk-lib/core/lib/cfn-resource.ts
```diff
@@ -725,7 +734,20 @@ function deepMerge(target: any, ...sources: any[]) {
         // sibling concrete values alongside, so we can delete this tree.
         const output = target[key];
         if (typeof(output) === 'object' && Object.keys(output).length === 0) {
-          delete target[key];
+          /*
+          * some empty trees are needed in some edge cases
```
paulhcsun, line 738:
```suggestion
          * Some empty trees are needed in some edge cases
```

packages/aws-cdk-lib/core/lib/cfn-resource.ts
```diff
@@ -725,7 +734,20 @@ function deepMerge(target: any, ...sources: any[]) {
...
+          * ie: wafv2 Rules (for Action, OverrideAction, etc...)
```
paulhcsun, line 739: Please clarify which `some empty trees` and maybe combine the `ie` into the `some edge cases` like `... are needed when WAFv2 rule action is one of 'Action', 'Block', 'Captcha', 'Challenge' or 'Count'.`

paulhcsun (changes requested):
Left some comments

paulhcsun:
@Rondineli I have some concerns about this change breaking existing customers so please add it behind a [feature flag](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md?rgh-link-date=2024-03-12T16%3A45%3A17Z#feature-flags).

Rondineli:
> @Rondineli I have some concerns about this change breaking existing customers even behind a feature flag. Please have a look at the comments on the original issue.

Sorry, but I really dont get the backwards compatibility bit? As I am explicitly setting the keys to it ignore. And trust me, I tried the way it is right now and cloudformation broke because those missing keys on waf rules, so, my guess is if someone tried to use it, it did not work.

I would be happy to change the PR for another approach, but this is definitely something broken and wont cause regression, quite the opposite.

PR #29235, opened 2024-02-23
fix(ecs): stack name can result in noncompliant capacity provider name

### LINKED ISSUE #29151: aws_ecs.AsgCapacityProvider: CDK should be smart enough to auto-name ASG capacity provider something valid if stack name starts with illegal keyword
### Describe the bug

If your stack name begins with "aws", "ecs", or "fargate", you will receive the following error when creating an ASG capacity provider if you don't specify a capacity provider name.

**"Invalid request provided: CreateCapacityProvider error: The specified capacity provider name is invalid. Up to 255 characters are allowed, including letters (
upper and lowercase), numbers, underscores, and hyphens. The name cannot be prefixed with "aws", "ecs", or "fargate". Specify a valid name and try again."**

### Expected Behavior

I expected the ASG capacity provider to be created without issue no matter what the stack name is. 

### Current Behavior

I described it pretty well above, but if you have any specific questions. I can't think of anything more relevant that would apply here.

### Reproduction Steps

Init a new stack with the CDK in a directory that starts with "aws-"

Create an ASG capacity provider, but don't specify the "capacityProviderName" property so the CDK will autogenerate a name for you

Deploy it

### Possible Solution

Maybe don't let a project be initialized with illegal keywords. You could also do a simple regex check when creating the ASG capacity provider to see if the autogenerated name contains those illegal keywords and either remove them or add some other word that is valid.

If we are going to ban the words aws across the board then the CDK should specify that so a linter can catch things like that. 

### CDK CLI Version

2.128.0 (build d995261)

### Framework Version

_No response_

### Node.js Version

v20.10.0

### OS

Linux fedora 6.7.4-200.fc39.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Feb  5 22:21:14 UTC 2024 x86_64 GNU/Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

msambol:
Closes #29151.

----


msambol:
Exemption request: adding a name check, I don't think an integration test is required.

packages/aws-cdk-lib/aws-ecs/lib/cluster.ts
```diff
@@ -1300,10 +1300,15 @@ export class AsgCapacityProvider extends Construct {
...
 
+    const capacityProviderNameRegex = /^(?!aws|ecs|fargate).+/gm;
     if (props.capacityProviderName) {
-      if (!(/^(?!aws|ecs|fargate).+/gm.test(props.capacityProviderName))) {
+      if (!(capacityProviderNameRegex.test(props.capacityProviderName))) {
         throw new Error(`Invalid Capacity Provider Name: ${props.capacityProviderName}, If a name is specified, it cannot start with aws, ecs, or fargate.`);
       }
+    } else {
+      if (!(capacityProviderNameRegex.test(Stack.of(this).stackName))) {
+        throw new Error(`Invalid Capacity Provider Name: ${Stack.of(this).stackName}, No name was specified, the stack name cannot start with aws, ecs, or fargate.`);
```
tmokmss, line 1310: Should we throw an error here, or create a valid name automatically as suggested in https://github.com/aws/aws-cdk/issues/29151#issuecomment-1954661302 ?

Sometimes it's difficult to rename a stack, so maybe the latter is better. wdyt?

msambol: I agree it would be better to generate the name but that has backwards compatibility concerns, I think ?

tmokmss: I see. In what kind of situation do you think backwards compatibility will become a problem?

msambol: If we add code that does this and someone redeploys a stack, CDK will generate a name and it will be different than the name already assigned by CFN.

tmokmss: But previously it should have resulted in a deployment error, right?

msambol: @tmokmss I updated this to generate a capacity provider name when the stack name isn't compliant.

msambol:
Exemption Request: There is already an integration test for capacity provider. The unit test should suffice here.

tmokmss (approved):

packages/aws-cdk-lib/aws-ecs/test/cluster.test.ts
```diff
@@ -2945,6 +2945,31 @@ test('throws when ASG Capacity Provider with capacityProviderName starting with
   }).toThrow(/Invalid Capacity Provider Name: ecscp, If a name is specified, it cannot start with aws, ecs, or fargate./);
 });
 
+test('throws when ASG Capacity Provider with no capacityProviderName but stack name starting with aws, ecs or faragte', () => {
```
gracelu0, line 2948:
```suggestion
test('throws when ASG Capacity Provider with no capacityProviderName but stack name starting with aws, ecs or fargate', () => {
```

gracelu0 (commented):
can we update the description for `capacityProviderName` [here](https://github.com/aws/aws-cdk/pull/29235/files#diff-1658b3219cbeb615eb4e61d96ccbadd7b7b1bb8ea6a8ec025d7ca8038c7599b3R1164) to include the added format with `cp-` prefix?

msambol:
> can we update the description for `capacityProviderName` [here](https://github.com/aws/aws-cdk/pull/29235/files#diff-1658b3219cbeb615eb4e61d96ccbadd7b7b1bb8ea6a8ec025d7ca8038c7599b3R1164) to include the added format with `cp-` prefix?

I added a note for this. 👍🏼

gracelu0 (approved):
LGTM, thanks for contributing!

Issue #22267, opened 2022-09-28
(aws-cloudtrail): isOrganizationTrail attaches insufficient permissions to bucket


danilobuerger:
### Describe the bug

When using a Trail with `isOrganizationTrail: true`, the bucket policy contains insufficient permissions and the stack creation fails with:

```
Invalid request provided: Incorrect S3 bucket policy is detected for bucket: ...
```

The Trail attaches the following permissions:

https://github.com/aws/aws-cdk/blob/c425e8ca1a3d296eb6a7fd7e005d07c1eadd16aa/packages/%40aws-cdk/aws-cloudtrail/lib/cloudtrail.ts#L226-L241

However, it is missing:

```
"Action": "s3:PutObject",
"Resource": "arn:aws:s3:::myOrganizationBucket/AWSLogs/o-exampleorgid/*"
```

as described here: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/create-s3-bucket-policy-for-cloudtrail.html#org-trail-bucket-policy

### Expected Behavior

It creates an organization trail.

### Current Behavior

It fails with
```
Invalid request provided: Incorrect S3 bucket policy is detected for bucket: ...
```

### Reproduction Steps

```
const bucket = new Bucket(this, "Bucket");
new Trail(this, 'Trail', { bucket, isOrganizationTrail: true });
```

### Possible Solution

Adding the permission as outlined above.

### CDK CLI Version

2.43.1

### Node.js Version

18.9.0

### OS

macOS

### Language

Typescript

peterwoodworth:
Thanks for reporting this @danilobuerger,

I can reproduce this just with this snippet

```ts
    new Trail(this, 'Trail', {
      isOrganizationTrail: true
    });
```

@daschaa did you ever run into this error? It's suspicious to me that the integration test you submitted with this PR works if I can't get it to deploy now here

daschaa:
@peterwoodworth That's very strange. The integration tests should have failed in the pull request.
I'm sorry that I forgot to adjust the permissions!

The challenge that we will now face is, that we have to pass/retrieve the belonging organization id to construct the correct permission statement. @peterwoodworth Afaik there is no way to get the organization id from the props that are passed when instantiating a trail. Do you know if this is possible?

danilobuerger:
@daschaa a few ideas:
1) pass the organization id as a prop
[...]

PR #29242, opened 2024-02-24
fix(cloudtrail): isOrganizationTrail attaches insufficient permissions to bucket

paulhcsun:
### Issue https://github.com/aws/aws-cdk/issues/22267 (if applicable)

Closes https://github.com/aws/aws-cdk/issues/22267.

### Reason for this change

Setting the `isOrganizationTrail` property to `true` attaches insufficient permissions to the bucket thus failing to deploy the `Trail` with:
```
Invalid request provided: Incorrect S3 bucket policy is detected for bucket: ...
```

### Description of changes

This PR adds a new property `orgId` to `TrailProps` that will be used to attach the [missing permission](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/create-s3-bucket-policy-for-cloudtrail.html#org-trail-bucket-policy) 
```
"Action": "s3:PutObject",
"Resource": "arn:aws:s3:::myOrganizationBucket/AWSLogs/o-organizationID/*",
"Condition": {
    "StringEquals": {
          "s3:x-amz-acl": "bucket-owner-full-control",
          "aws:SourceArn": "arn:aws:cloudtrail:region:managementAccountID:trail/trailName"
}
```
when `isOrganizationTrail: true`. 

Validated locally that I can deploy when `Trail.isOrganizationTrail: true` with a valid `orgId` + added unit test case.

I can't think of a way to test this with an integ test as it requires a valid `orgId` to deploy but any suggestions are welcome. 

### Checklist

----


packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
```diff
@@ -262,6 +270,22 @@ export class Trail extends Resource {
       },
     }));
 
+    if (props.isOrganizationTrail) {
```
paulhcsun, line 273: Could also add a warning in the case that `isOrganizationTrai: true` and `orgId` is not provided. Probably not an error even though it would fail to deploy since some users may have manually added the missing policy already and won't need to use the new `orgId` prop. 

Warning could be something like this but any alternative suggestions are welcome:
`Organization Trails require passing in an organization id using the orgId property to attach a missing policy.`

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
```diff
@@ -128,6 +128,14 @@ export interface TrailProps {
    */
   readonly isOrganizationTrail?: boolean;
 
+  /** The orgId.
```
GavinZZ, line 131: Why does it call `orgId`? Reaching this [link](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/create-s3-bucket-policy-for-cloudtrail.html), it seems to be `The name of the folder where the log files are stored, including the bucket name, a prefix (if you specified one), and your AWS account ID`. So more like account id?

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
```diff
@@ -262,6 +270,22 @@ export class Trail extends Resource {
       },
     }));
 
+    if (props.isOrganizationTrail) {
+      this.s3bucket.addToResourcePolicy(new iam.PolicyStatement({
+        resources: [this.s3bucket.arnForObjects(
```
GavinZZ, line 275: In the doc, it mentions `"arn:aws:s3:::myBucketName/[optionalPrefix]/AWSLogs/myAccountID/*"`. I haven't looked deep into it, but is the optional prefix part of the bucket and it would be handled by `this.s3bucket.arnForObjects`?

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
```diff
@@ -262,6 +270,22 @@ export class Trail extends Resource {
...
+          `AWSLogs/${props.orgId}/*`,
```
GavinZZ, line 276: Looking at the below snippet, wouldn't the `orgId` be `this.s3bucket.stack.account`?
```
Resource": "arn:aws:s3:::myBucketName/[optionalPrefix]/AWSLogs/myAccountID/*",
            "Condition": {
                "StringEquals": {
                    "s3:x-amz-acl": "bucket-owner-full-control",
                    "aws:SourceArn": "arn:aws:cloudtrail:region:myAccountID:trail/trailName"
                }
            }
```

GavinZZ (commented):
Some questions

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
paulhcsun: The missing permission is specified under [this section for organization trails](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/create-s3-bucket-policy-for-cloudtrail.html#org-trail-bucket-policy). Specifically [this policy](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/create-s3-bucket-policy-for-cloudtrail.html#:~:text=%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%22Sid%22%3A%20%22AWSCloudTrailOrganizationWrite20150319,%3Atrail/trailName%22%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%7D) which requires the `o-organizationID`:

```
{
            "Sid": "AWSCloudTrailOrganizationWrite20150319",
            "Effect": "Allow",
            "Principal": {
                "Service": [
                    "cloudtrail.amazonaws.com"
                ]
            },
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::myOrganizationBucket/AWSLogs/o-organizationID/*",
            "Condition": {
                "StringEquals": {
                    "s3:x-amz-acl": "bucket-owner-full-control",
                    "aws:SourceArn": "arn:aws:cloudtrail:region:managementAccountID:trail/trailName"
                }
            }
        }
```

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
paulhcsun: It would be this value for the resources: `"arn:aws:s3:::myOrganizationBucket/AWSLogs/o-organizationID/*"` where `arn:aws:s3:::myOrganizationBucket/` is handled by `this.s3bucket.arnForObjects`. But for the case you're talking about it would not handle it, [the optional prefix part is passed in from props](https://github.com/aws/aws-cdk/blob/d30881882f0742362d171340d2f94be8b7707acc/packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts#L254-L257):
```
    this.s3bucket.addToResourcePolicy(new iam.PolicyStatement({
      resources: [this.s3bucket.arnForObjects(
        `${props.s3KeyPrefix ? `${props.s3KeyPrefix}/` : ''}AWSLogs/${Stack.of(this).account}/*`,
      )],
```

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
paulhcsun: That's a different permission from that one we are adding here:
```
"Resource": "arn:aws:s3:::myOrganizationBucket/AWSLogs/o-organizationID/*",
            "Condition": {
                "StringEquals": {
                    "s3:x-amz-acl": "bucket-owner-full-control",
                    "aws:SourceArn": "arn:aws:cloudtrail:region:managementAccountID:trail/trailName"
                }
            }
```

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
GavinZZ: Oh ok that makes a lot more sense!

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
GavinZZ: This makes a lot more sense! Thanks!

GavinZZ (approved):
LGTM

PR #29243, opened 2024-02-24
fix(ecr): incorrect format for rule pattern

### LINKED ISSUE #29225: (aws-ecr): ECR EventBridge rule has incorrect event pattern
### Describe the bug

Using the CDK, creating an EventBridge rule on an ECR repository creates a rule with the following event pattern:
```json
{
  "resources": ["arn:aws:ecr:ap-southeast-2:<account-id>:repository/my-repository-name"],
  "source": ["aws.ecr"]
}
```

However when ECR sends the event, `"resources"` is empty, so the rule never matches (example from https://docs.aws.amazon.com/AmazonECR/latest/userguide/ecr-eventbridge.html#ecr-eventbridge-bus):
```json
{
    "version": "0",
    "id": "13cde686-328b-6117-af20-0e5566167482",
    "detail-type": "ECR Image Action",
    "source": "aws.ecr",
    "account": "123456789012",
    "time": "2019-11-16T01:54:34Z",
    "region": "us-west-2",
    "resources": [], // <--- empty, rule fails
    "detail": {
        "result": "SUCCESS",
        "repository-name": "my-repository-name",
        "image-digest": "sha256:7f5b2640fe6fb4f46592dfd3410c4a79dac4f89e4782432e0378abcd1234",
        "action-type": "PUSH",
        "image-tag": "latest"
    }
}
```

### Expected Behavior

Use the `detail.repository-name` field as a filter instead of `resources`:
```json
{
  "detail": {
    "repository-name": ["my-repository-name"]
  },
  "source": ["aws.ecr"]
}
```

### Current Behavior

Uses `resources` as a filter, which is empty in the real event

### Reproduction Steps

```csharp
var ecrRepo = new EcrRepository(this, "CiTestEcrRepository", new EcrRepositoryProps());
ecrRepo.OnEvent("EcrRepoEventRule", new OnEventOptions() {
    Target = new LambdaFunction(myLambdaHandler),
});
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.127.0 (build 6c90efc)

### Framework Version

.NET 8

### Node.js Version

v18.16.1

### OS

MacOS

### Language

.NET

### Language Version

.NET 8

### Other information

_No response_

msambol:
Closes #29225. 

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ecr/README.md
+++ b/packages/aws-cdk-lib/aws-ecr/README.md
@@ -215,6 +215,7 @@ You can publish repository events to a CloudWatch event rule with `onEvent`:
 
 ```ts
 import * as lambda from 'aws-cdk-lib/aws-lambda';
+import { LambdaFunction } from 'aws-cdk-lib/aws-events-targets';
 
 const repo = new ecr.Repository(this, 'Repo');
 const lambdaHandler = new lambda.Function(this, 'LambdaFunction', {


lpizzinidev (approved):
Looks good to me 👍

scanlonp:
Hey @msambol, just checking if this PR is ready to go, or the rule pattern is still giving you problems with the trigger on the Lambda. 

No worries if not, just wanted to check in.

msambol:
> Hey @msambol, just checking if this PR is ready to go, or the rule pattern is still giving you problems with the trigger on the Lambda.
> 
> No worries if not, just wanted to check in.

The trigger isn't being created but I think we should address that in a separate PR.

scanlonp:
Maybe I am missing some context. I see from the issue that the rule never matched. Does that mean the lambda trigger was never created, even before this PR?

msambol:
Let me test it out again. When I first worked on this, it didn't. I'll follow up.

msambol:
@scanlonp I just redeployed my integration test again. The EventBridge rule gets created, but the trigger on the Lambda does not get created. Once I manually add the trigger, I can get the Lambda to fire.

shikha372:
@msambol Thank you for your contribution, would like to understand this issue better before reviewing , in the reference [docs ](https://docs.aws.amazon.com/AmazonECR/latest/userguide/ecr-eventbridge.html#ecr-eventbridge-bus)resources[] field is not empty always eg. I see it is required for event type `Event for a pull through cache action`  or `Event for a completed image scan (basic scanning)` .. how are we handling these cases ? let me know if my understanding is not correct

shikha372:
Thank you @msambol , would also like to confirm if there are no issues with lambda trigger when we define the rule using aws-events-targets, I would also test at my end to understand the lambda trigger problem to see we are not impacting any existing functionality for the targets with this change.

shikha372:
hey @msambol ,
had a chance to reproduce this today, curious to know is there any specific reason to use bind here, I tried with the following code to register lambda function but yes i cannot see the trigger above cases, but it is getting registered if we specify resources.
```
const myLambdaHandler = new lambda.Function( stack, 'test', {
  runtime: lambda.Runtime.PYTHON_3_12,
  code: lambda.Code.fromInline('# dummy func'),
  handler: 'index.handler',
});

repoOnEvent.onEvent('OnEventTargetLambda', {
  target: new LambdaFunction(myLambdaHandler),
});
```

shikha372:
hey @msambol  ,
had a chance to reproduce this today, curious to know is there any specific reason to use bind here? I tried with the following code to register lambda function but yes i cannot see the trigger in above case which might indicate this is not the correct pattern to replace here.  It is getting registered if we specify resources under eventpattern.

```
const myLambdaHandler = new lambda.Function( stack, 'test', {
  runtime: lambda.Runtime.PYTHON_3_12,
  code: lambda.Code.fromInline('# dummy func'),
  handler: 'index.handler',
});

repoOnEvent.onEvent('OnEventTargetLambda', {
  target: new LambdaFunction(myLambdaHandler),
});
```

msambol:
@shikha372 It was so long ago I'm not sure why I did that 🤦🏼‍♂️  Fixed per your recommendation—thanks!

shikha372:
> @shikha372 It was so long ago I'm not sure why I did that 🤦🏼‍♂️ Fixed per your recommendation—thanks!

Thanks to confirm just ran the integration tests on my local and  I can the trigger too with details as a filter

msambol:
> > @shikha372 It was so long ago I'm not sure why I did that 🤦🏼‍♂️ Fixed per your recommendation—thanks!
> 
> Thanks, to confirm just ran the integration tests on my local and I can see the lambda trigger too with details as a filter

I see it as well: https://github.com/aws/aws-cdk/pull/29243/commits/e4579427e582f4fd5d1704486300a734015ede8b. Thanks!

shikha372 (approved):

PR #29268, opened 2024-02-26
fix(CLI): change-set diff not required for new stack diffs

### LINKED ISSUE #29265: (cli): change-set diff not required for new stack diffs
### Describe the bug

### Expected Behavior

### Current Behavior

n/a

### Reproduction Steps

- create a new stack
- run `cdk diff`
- check CloudTrail

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.130.0

### Framework Version

_No response_

### Node.js Version

any

### OS

any

### Language

### Language Version

_No response_

### Other information

_No response_

sumupitchayan:
Closes #29265 

Adds a check in the `fullDiff` function to make sure that change-set diff is not run for new stack diffs.

----


packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
```diff
@@ -1231,4 +1231,22 @@ describe('changeset', () => {
     expect(differences.resources.differenceCount).toBe(1);
     expect(differences.resources.get('BucketResource').changeImpact === ResourceImpact.WILL_IMPORT);
   });
+
+  test('does not do changeset checks for new stack', async () => {
```
TheRealAmazonKendra, line 1235: I might be missing something but I'm not sure how this tests that we do not do a changeset for new stacks. Could you clarify that?

packages/aws-cdk/test/diff.test.ts
```diff
@@ -93,13 +93,12 @@ describe('imports', () => {
...
-    const plainTextOutput = buffer.data.replace(/\x1B\[[0-?]*[ -/]*[@-~]/g, '');
-    expect(plainTextOutput).toContain(`Stack A
-Resources
-[←] AWS::SQS::Queue Queue import
-[←] AWS::SQS::Queue Queue2 import
-[←] AWS::S3::Bucket Bucket import
-`);
+    let plainTextOutput = buffer.data.replace(/\x1B\[[0-?]*[ -/]*[@-~]/g, '');
+    expect(plainTextOutput).toContain('Stack A');
+    expect(plainTextOutput).toContain('Resources');
```
TheRealAmazonKendra, line 98: In the case of imports, this should still specify that an import is happening so the change to this test tells me something is missing in the additional logic.

TheRealAmazonKendra (changes requested):
Comments inline. Also did you push this into the testing pipeline yet?

packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
sumupitchayan: Was confused on this too - my logic was that calling `fullDiff` with the current stack being empty (new) should mean that there should be only 1 difference count, and the change impact should be `ResourceImpact.WILL_CREATE`; I'm not sure how else to test that we don't do changeset for new stacks

scanlonp:
Closed by #29394.

Issue #29260, opened 2024-02-26
@aws-cdk/integ-tests-alpha: SingletonFunctions used for test assertions should set a CloudWatch log group retention policy


jvalore-medallia:
### Describe the feature

Currently if you perform a test like

```
import { ExpectedResult, IntegTest } from '@aws-cdk/integ-tests-alpha';

const integ = new IntegTest(app, 'Example', {
  testCases: [testStack],
  assertionStack: new cdk.Stack(app, 'assertions', { synthesizer }),
});

const invoke = integ.assertions.invokeFunction({
  functionName: 'some:lambda:arn',
  payload: '{}'
});
```

The stack deletes itself after a test, but CloudWatch log groups named like
```
/aws/lambda/assertions-SingletonFunction{random id}
```
and are set to never expire, so end up cluttering up your aws account and need manual cleanup.

### Use Case

Test results are not needed forever and should have some default retention policy, or at least a way to set them.

### Proposed Solution

Include a default retention policy for `SingletonFunction` CloudWatch log groups.

### Other Information

nodejs dependencies:
```
    "@aws-cdk/integ-runner": "^2.129.0-alpha.0",
    "@aws-cdk/integ-tests-alpha": "2.129.0-alpha.0",
    "aws-cdk": "2.129.0",
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.129.0

### Environment details (OS name and version, etc.)

Mac. Node 20. cdk 2.129.0

pahud:
I guess this is because by default lambda.Function will create a built-in log group that never expires unless a custom loggroup is specified [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-loggingconfig.html#cfn-lambda-function-loggingconfig-loggroup).

Yes it would be great to have the log retention policy support for that.

PR #29277, opened 2024-02-27
chore(integ-test-alpha): add log group retention days to integ test stack

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29260

### Reason for this change

CloudWatch log groups created by the integ tests are set to never expire, so end up cluttering up your aws account and need manual cleanup.

### Description of changes

Added custom log group with log group retention

All existing tests and new tests pass.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-rds/test/integ.cluster-data-api.js.snapshot/cluster-data-api.assets.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-rds/test/integ.cluster-data-api.js.snapshot/cluster-data-api.assets.json
@@ -14,15 +14,15 @@
         }
       }
     },
-    "30c06211c853a3fd7c30a7376cbd63b2155519a67ba90ed0a731aa67dcd179e2": {
+    "b2fb40833cc33946a3656aa5e2cdb19e4259b9a84dbabe7a6559ad319344359d": {
       "source": {
-        "path": "cluster-kerberos.template.json",
+        "path": "cluster-data-api.template.json",
         "packaging": "file"
       },
       "destinations": {
         "current_account-current_region": {
           "bucketName": "cdk-hnb659fds-assets-${AWS::AccountId}-${AWS::Region}",
-          "objectKey": "30c06211c853a3fd7c30a7376cbd63b2155519a67ba90ed0a731aa67dcd179e2.json",
+          "objectKey": "b2fb40833cc33946a3656aa5e2cdb19e4259b9a84dbabe7a6559ad319344359d.json",
           "assumeRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-file-publishing-role-${AWS::AccountId}-${AWS::Region}"
         }
       }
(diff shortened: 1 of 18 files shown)


TheRealAmazonKendra (dismissed):
I would prefer that we make this a configurable field rather than just assigning an arbitrary time period.

packages/@aws-cdk/integ-tests-alpha/README.md
```diff
@@ -439,6 +439,28 @@ invoke.expect(ExpectedResult.objectLike({
 }));
 ```
 
+The above example will by default create a CloudWatch log group that's never
+expired. If you want to configure it with custom log retention days, you need
+to specify the `logRenteion` property.
```
shikha372, line 444: small nit: `logRetention`

packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/provider.ts
```diff
@@ -13,6 +14,13 @@ export interface LambdaFunctionProviderProps {
    * @default index.handler
    */
   readonly handler?: string;
+
+  /**
+   * How long, in days, the log contents will be retained.
+   *
+   * @default - no retention days specified
```
shikha372, line 21: should we set some days as default ?

packages/@aws-cdk/integ-tests-alpha/README.md
GavinZZ: Good catch. Will update

packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/provider.ts
GavinZZ: Not in this case, since the default behaviour is no log retention. Once users provide any value, we would then add a LogGroup resource.

packages/@aws-cdk/integ-tests-alpha/lib/assertions/providers/provider.ts
```diff
@@ -118,6 +131,13 @@ interface SingletonFunctionProps extends LambdaFunctionProviderProps {
    * We recommend generating a UUID per provider.
    */
   readonly uuid: string;
+
+  /**
+   * How long, in days, the log contents will be retained.
+   *
+   * @default - no retention days specified
+   */
+  readonly logRetention?: RetentionDays;
```
shikha372, line 140: to confirm do we need this here if its already defined in `LambdaFunctionProviderProps`

GavinZZ: They're duplicated, removing this one.

shikha372 (approved):

paulhcsun:
@Mergifyio update

mokele:
This PR has a bug that if you use this feature then you cannot have more than 1 long running assertion stack co-exist in the same aws account.

It's common to have shared aws accounts, and this feature uses the hardcode singleton uuid value for the log group `/aws/logs/{id}`, making it the same across the whole account not just the current stack, making it not possible to have multiple longer running assertion stacks (with `--no-clean`).

I'm not sure yet if there is a way to circumvent this and get to override the uuid that is baked in here that results in part of the singleton functions's id 😢

I'll work on writing up a proper bug issue when I have a moment

Issue #26640, opened 2023-08-04
Apprunner: After pushing a Docker image to ECR, isn't triggered.


watany-dev:
### Describe the bug

Deployed AppRunner using AWS CDK.
The expected behavior is for AppRunner to automatically deploy upon each change to the container image in ECR.

### Expected Behavior

After pushing a Docker image to ECR, the AppRunner deployment is triggered.

### Current Behavior

However, even after pushing a Docker image to ECR, the AppRunner deployment isn't triggered.

### Reproduction Steps

import * as apprunner from '@aws-cdk/aws-apprunner-alpha'
import { Cpu, Memory } from '@aws-cdk/aws-apprunner-alpha'
import * as cdk from 'aws-cdk-lib'
import * as iam from 'aws-cdk-lib/aws-iam'
import { Construct } from 'constructs'
import { EcrStack } from './ecr'

export class AppRunnerStack extends cdk.Stack {
  constructor(
    scope: Construct,
    id: string,
    ecrStack: EcrStack,
    props?: cdk.StackProps,
  ) {
    super(scope, id, props)

    new apprunner.Service(this, 'SampleAppRunnerService', {
      serviceName: 'sample-app',
      cpu: Cpu.ONE_VCPU,
      memory: Memory.TWO_GB,
      autoDeploymentsEnabled: true,
      source: apprunner.Source.fromEcr({
        imageConfiguration: {
          port: 3000,
          startCommand: 'npm run start --workspace=app',
        },
        repository: ecrStack.repository,
        tagOrDigest: 'latest',
      }),
    })
  }
}

## Cause:

The access role for AppRunner’s service is automatically generated. However, it lacks permission for the ecr:DescribeImages action, preventing it from detecting image changes.

## Necessary Permissions:
Actions required for ECR access:
- ecr:BatchCheckLayerAvailability
- ecr:GetDownloadUrlForLayer
- ecr:BatchGetImage
- ecr:GetAuthorizationToken

## Actions that AppRunner needs for ECR access:
- ecr:GetDownloadUrlForLayer
- ecr:BatchCheckLayerAvailability
- ecr:BatchGetImage
- ecr:DescribeImages
- ecr:GetAuthorizationToken

## Proposed Solution:
1. Add the ecr:DescribeImages action to the access role of AppRunner's service.
2. Attach the IAM policy that includes the above actions to AppRunner.
3. Make the AccessRole public to allow users to access it.

### Additional Information/Context

[...]

PR #29285, opened 2024-02-27
fix: unable to trigger ECR image auto deployment

Mickeypeng:
### Issue # (if applicable)

Closes [#26640 ](https://github.com/aws/aws-cdk/issues/26640)

### Reason for this change

ECR image auto deployment can not be triggered due to insufficient permission on access role

### Description of changes

Hi CDK team, this is mingran@ from AppRunner service team. I took a look at this bug and found that when access role is not specified for ECR, CDK create a new role, but lacking necessary permission to trigger auto deployment. Hence adding it.

Can be tested by reproduce the bug, or simpler, create an app runner service and verify that access role is attached with correct permissions.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
+++ b/packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
@@ -1318,7 +1318,13 @@ export class Service extends cdk.Resource implements iam.IGrantable {
       assumedBy: new iam.ServicePrincipal('build.apprunner.amazonaws.com'),
     });
     accessRole.addToPrincipalPolicy(new iam.PolicyStatement({
-      actions: ['ecr:GetAuthorizationToken'],
+      actions: [        
+        'ecr:BatchCheckLayerAvailability',
+        'ecr:BatchGetImage',
+        'ecr:DescribeImages',
+        'ecr:GetAuthorizationToken',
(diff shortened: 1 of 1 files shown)


Mickeypeng:
Exemption Request: We don't need unit test to verify if an access role have attach the correct permission. It's too straight forward.

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -1318,7 +1318,13 @@ export class Service extends cdk.Resource implements iam.IGrantable {
       assumedBy: new iam.ServicePrincipal('build.apprunner.amazonaws.com'),
     });
     accessRole.addToPrincipalPolicy(new iam.PolicyStatement({
-      actions: ['ecr:GetAuthorizationToken'],
+      actions: [        
+        'ecr:BatchCheckLayerAvailability',
+        'ecr:BatchGetImage',
+        'ecr:DescribeImages',
+        'ecr:GetAuthorizationToken',
+        'ecr:GetDownloadUrlForLayer',
```
msambol, line 1326: can you comment on the other permissions added? In https://github.com/aws/aws-cdk/pull/29426, it looks like only `ecr:DescribeImages` is needed?

PR #29288, opened 2024-02-28
feat(stepfunctions-tasks): add optional role property to EvaluateExpressionProps

lagroujl:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29350.

### Reason for this change

This change enables customers who need more control over IAM roles and whose environments do not allow them to use auto-generated roles to use EvaluateExpression.

### Description of changes

Added a `role` property to `EvaluateExpressionProps` that allows customers to bring their own role. This would replace the role that is auto-generated by SingletonFunction when creating the Lambda Function used inside EvaluateExpression.

Added new unit test that validates created Function uses role created externally

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/cloudformation-diff/package.json
+++ b/packages/@aws-cdk/cloudformation-diff/package.json
@@ -23,8 +23,8 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@aws-cdk/aws-service-spec": "^0.1.6",
-    "@aws-cdk/service-spec-types": "^0.0.74",
+    "@aws-cdk/aws-service-spec": "^0.1.7",
+    "@aws-cdk/service-spec-types": "^0.0.75",
     "chalk": "^4",
     "diff": "^5.2.0",
     "fast-deep-equal": "^3.1.3",
--- a/packages/@aws-cdk/integ-runner/package.json
+++ b/packages/@aws-cdk/integ-runner/package.json
@@ -74,7 +74,7 @@
     "@aws-cdk/cloud-assembly-schema": "0.0.0",
     "@aws-cdk/cloudformation-diff": "0.0.0",
     "@aws-cdk/cx-api": "0.0.0",
-    "@aws-cdk/aws-service-spec": "^0.1.6",
+    "@aws-cdk/aws-service-spec": "^0.1.7",
     "cdk-assets": "0.0.0",
     "@aws-cdk/cdk-cli-wrapper": "0.0.0",
     "aws-cdk": "0.0.0",
--- a/packages/aws-cdk-lib/aws-applicationsignals/.jsiirc.json
+++ b/packages/aws-cdk-lib/aws-applicationsignals/.jsiirc.json
@@ -1,0 +1,13 @@
+{
+  "targets": {
+    "java": {
+      "package": "software.amazon.awscdk.services.applicationsignals"
+    },
+    "dotnet": {
+      "package": "Amazon.CDK.AWS.ApplicationSignals"
+    },
+    "python": {
+      "module": "aws_cdk.aws_applicationsignals"
+    }
+  }
+}
--- a/packages/aws-cdk-lib/aws-applicationsignals/README.md
+++ b/packages/aws-cdk-lib/aws-applicationsignals/README.md
@@ -1,0 +1,39 @@
+# AWS::ApplicationSignals Construct Library
+<!--BEGIN STABILITY BANNER-->
+
+---
+
+![cfn-resources: Stable](https://img.shields.io/badge/cfn--resources-stable-success.svg?style=for-the-badge)
+
+> All classes with the `Cfn` prefix in this module ([CFN Resources]) are always stable and safe to use.
+>
+> [CFN Resources]: https://docs.aws.amazon.com/cdk/latest/guide/constructs.html#constructs_lib
+
+---
+
+<!--END STABILITY BANNER-->
+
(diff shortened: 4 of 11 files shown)


packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
```diff
@@ -111,13 +118,18 @@ function createEvalFn(runtime: lambda.Runtime | undefined, scope: Construct) {
...
 
-  const uuid = nodeJsGuids[runtime?.name ?? NO_RUNTIME];
+  let uuid = nodeJsGuids[props.runtime?.name ?? NO_RUNTIME];
   if (!uuid) {
-    throw new Error(`The runtime ${runtime?.name} is currently not supported.`);
+    throw new Error(`The runtime ${props.runtime?.name} is currently not supported.`);
+  }
+
+  if (props.role) {
+    uuid += props.role.node.addr;
```
lagroujl, line 127: I am not totally sure if this is the right way to go about this. But the point is, I want the role to be part of the signature of the function created, so if I want to use different roles in one part of the stack or don't set a role at all, it will create different functions, instead of the first occurrence blocking all of the others.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
```diff
@@ -22,6 +22,13 @@ export interface EvaluateExpressionProps extends sfn.TaskStateBaseProps {
    * @default lambda.Runtime.NODEJS_18_X
    */
   readonly runtime?: lambda.Runtime;
+
+  /**
+   * This is the role that will be assumed by the function upon execution.
+   *
+   * @default - a role will be automatically created
```
msambol, line 29: nit: could you remove the hyphen to match the other `@default`.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
msambol: Is there a risk of this uuid being too long?

lagroujl: coI could not find one documented. Also, addr should have a fixed length of 42. SingletonFunction eventually passes it to `tryFindChild`. So that is where a length constraint could potentially come up, but otherwise, I think the risk is very low/none.

lagroujl: More likely, what I'm worried about is this addr value changing between deployments and deleting/recreating the function. I'm not sure I understand how the value of addr is calculated well enough to guarantee that won't happen

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
```diff
@@ -22,6 +22,13 @@ export interface EvaluateExpressionProps extends sfn.TaskStateBaseProps {
    * @default lambda.Runtime.NODEJS_18_X
    */
   readonly runtime?: lambda.Runtime;
+
+  /**
+   * This is the role that will be assumed by the function upon execution.
+   *
+   * @default a role will be automatically created
```
paulhcsun, line 29:
```suggestion
   * @default - a role will be automatically created
```

We need to add a `-` after the `@default` tag if we're using a sentence after instead of specifying the default value.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
paulhcsun: What even is the value of `props.role.node.addr`? Could we use the `props.role.roleName/roleArn` instead?

paulhcsun (commented):
Just one question and docstring change.

lagroujl:
> What is the value of props.role.node.addr? 

https://docs.aws.amazon.com/cdk/api/v2/docs/constructs.Node.html#addr

tl:dr; SHA-1 of the components of the construct path.

> Could we use the props.role.roleName/roleArn instead?

No. this could be a token and change between deployments.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
lagroujl: > What is the value of props.role.node.addr?

https://docs.aws.amazon.com/cdk/api/v2/docs/constructs.Node.html#addr

tl:dr; SHA-1 of the components of the construct path.

> Could we use the props.role.roleName/roleArn instead?

No. this could be a token and change between deployments.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
lagroujl: https://github.com/aws/aws-cdk/pull/29288#discussion_r1580277065

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
paulhcsun: Were you able to test if the value of `props.role.node.addr` changes between deployments? From the description, `Addresses are calculated using a SHA-1 of the components of the construct path.`,  I'm not confident that it would be the same between deployments.

PR #29133, opened 2024-02-16
chore(rds): add clusterArn property to DatabaseCluster

BenFradet:
### Reason for this change

This is to bring parity with #10741, we need the cluster arn to enable data api manually waiting for #28574.

### Description of changes

This adds a clusterArn property to IDatabaseCluster

### Description of how you validated changes

The changes have been validated through a unit test

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #29289, opened 2024-02-28
feat(rds): add clusterArn property to DatabaseCluster

blimmer:
### Reason for this change

It wasn't clear to me how to get the `clusterArn` from `DatabaseCluster`. This PR adds that property.

### Description of changes

Add `clusterArn` to DatabaseCluster, just like `ServerlessCluster` exposes.

Added unit test similar to `ServerlessCluster`.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-rds/lib/cluster.ts
+++ b/packages/aws-cdk-lib/aws-rds/lib/cluster.ts
@@ -686,6 +686,18 @@ abstract class DatabaseClusterNew extends DatabaseClusterBase {
   }
 
   /**
+   * The ARN of the cluster
+   */
+  public get clusterArn(): string {
+    return Stack.of(this).formatArn({
+      service: 'rds',
+      resource: 'cluster',
+      arnFormat: ArnFormat.COLON_RESOURCE_NAME,
+      resourceName: this.clusterIdentifier,
+    });
+  }
+
+  /**
    * Create cluster instances
    *
    * @internal
--- a/packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
+++ b/packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
@@ -4052,6 +4052,34 @@ describe('cluster', () => {
       ],
     });
   });
+
+  test('check that clusterArn property works', () => {
+    // GIVEN
+    const stack = testStack();
+    const vpc = new ec2.Vpc(stack, 'VPC');
+    const exportName = 'DbCluterArn';
+
+    // WHEN
+    const cluster = new DatabaseCluster(stack, 'Database', {
+      engine: DatabaseClusterEngine.AURORA,
+      writer: ClusterInstance.provisioned('writer'),
+      vpc,
+    });
+    new cdk.CfnOutput(stack, exportName, {
+      exportName,
+      value: cluster.clusterArn,
+    });
+
+    // THEN
+    expect(stack.resolve(cluster.clusterArn)).toEqual({
+      'Fn::Join': ['', [
+        'arn:',
+        { Ref: 'AWS::Partition' },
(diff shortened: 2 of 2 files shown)


packages/aws-cdk-lib/aws-rds/lib/cluster.ts
```diff
@@ -685,6 +685,18 @@ abstract class DatabaseClusterNew extends DatabaseClusterBase {
...
+   */
+  public get clusterArn(): string {
+    return Stack.of(this).formatArn({
+      service: 'rds',
+      resource: 'cluster',
+      arnFormat: ArnFormat.COLON_RESOURCE_NAME,
+      resourceName: this.clusterIdentifier,
+    });
+  }
+
```
blimmer, line 699: Just like `ServerlessCluster`

https://github.com/aws/aws-cdk/blob/1f30b5d873b7a1b11733211fa9ef47792fa4c678/packages/aws-cdk-lib/aws-rds/lib/serverless-cluster.ts#L345-L355

packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
```diff
@@ -4052,6 +4052,34 @@ describe('cluster', () => {
       ],
     });
   });
+
+  test('check that clusterArn property works', () => {
```
blimmer, line 4056: https://github.com/aws/aws-cdk/blob/1f30b5d873b7a1b11733211fa9ef47792fa4c678/packages/aws-cdk-lib/aws-rds/test/serverless-cluster.test.ts#L602-L627

blimmer:
Exemption request. The JSDoc should be sufficient and I added a unit test.

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
jfuss: @blimmer Is this how CDK normally handles this? I ask because the output becomes a `Join` instead of doing a `GetAtt` on the resource. Trying to wrap my head around why we should do this over the `GetAtt` path. I tried looking at the Lambda Function L2, since there is a `readonly` property for the Arn.

Honestly not sure if it matters (likely not) but trying to learn different parts of this as I go. To me seems like we have two patterns going on, the one you have here and the `readonly` property for the Arn.

blimmer: I'm not sure - I just followed the pattern in the linked code. One of the core contributors might be able to tell us.

jfuss: I asked the team directly. I did find SQS which does this differently. https://github.com/aws/aws-cdk/blob/v2.130.0/packages/aws-cdk-lib/aws-sqs/lib/queue.ts#L437-L440

```
const queue = new sqs.Queue(this, 'CdkExampleQueue', {
      visibilityTimeout: Duration.seconds(300)
    });

new CfnOutput(this, "arn", {
      exportName: "arn",
      value: queue.queueArn,
    })
```


CFN Synth
```
Outputs:
  arn:
    Value:
      Fn::GetAtt:
        - CdkExampleQueue7618E31B
        - Arn
    Export:
      Name: arn
```

Which fits my mental model better. I am not sure if there is a big difference or a reason it was done this way in RDS. Maybe one is a "legacy" way of doing things. 

Generally, this looks fine. I just want to get some deeper clarity on what is typical.

go-to-k:
Is there a duplication with this PR?

https://github.com/aws/aws-cdk/pull/29133

jfuss:
@blimmer Thanks for spending the time. I missed there was another PR already for this and was merged yesterday: https://github.com/aws/aws-cdk/pull/29133.

Closing this as it is a duplicate

Issue #29308, opened 2024-02-29
aws-kms: `Key.grant*` methods misidentify Key region when enclosing Stack is different region; leads to wildcard resource ARNs


ryderben:
### Describe the bug

This is a possible cause of https://github.com/aws/aws-cdk/issues/23991.

This problem is `grant()` determines the region of a `Key` using `Stack.of(key).region`, however the enclosing `Stack`'s region may differ to that of the actual resource.

https://github.com/aws/aws-cdk/blob/8d07b857aec64211054e409a098a841ac57d9e09/packages/aws-cdk-lib/aws-kms/lib/key.ts#L259-L266

One example of where these differ is when the `Key` was imported as a replicaKey within a  Global DynamoDB `TableV2` - in these contexts, resources under several regions must be scoped under a single-region `Stack`. 

```typescript
import * as cdk from "aws-cdk-lib";
import * as kms from "aws-cdk-lib/aws-kms";
import * as ddb from "aws-cdk-lib/aws-dynamodb";
import * as iam from "aws-cdk-lib/aws-iam";

const account = "123412341234";
const primaryRegion = "us-west-2";
const replicaRegion = "eu-north-1";
const kmsKeys = {
    "us-west-2":
        "arn:aws:kms:us-west-2:123412341234:key/5a84f9d3-7a10-4b37-afe7-95a496db8b79",
    "eu-north-1":
        "arn:aws:kms:eu-north-1:123412341234:key/e3ab59e5-3dc3-4bc4-9c3f-c790231d2287",
};

const app = new cdk.App();

const dbStack = new cdk.Stack(app, "DbStack", {
    env: { account, region: primaryRegion },
});
const table = new ddb.TableV2(dbStack, "Table", {
    partitionKey: { name: "id", type: ddb.AttributeType.STRING },
    replicas: [{ region: replicaRegion }],
    encryption: ddb.TableEncryptionV2.customerManagedKey(
        kms.Key.fromKeyArn(dbStack, "PrimaryTableKey", kmsKeys["us-west-2"]),
        { "eu-north-1": kmsKeys["eu-north-1"] },
    ),
});

for (const region of [primaryRegion, replicaRegion]) {
    const iamStack = new cdk.Stack(app, `IamStack-${region}`, {
        env: { region, account },
    });
    const role = new iam.Role(iamStack, "Role", {
        assumedBy: new iam.AccountPrincipal("123456789012"),
    });
    const replica = table.replica(region);
[...]

PR #29315, opened 2024-02-29
fix(kms): kms key grant methods misidentify region when enclosing stack is different region

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29308

### Reason for this change

This problem is grant() determines the region of a Key using Stack.of(key).region, however the enclosing Stack's region may differ to that of the actual resource. When this happens, the IAM policy generated allows a `*` resource which is against the least privilege rule.

### Description of changes

KMS key already has `env` value on account and region, use this first. If not exist, use stack account and region.

New unit test

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
+++ b/packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
@@ -37,7 +37,6 @@ const DIFF_HANDLERS: HandlerRegistry = {
  * @param currentTemplate the current state of the stack.
  * @param newTemplate     the target state of the stack.
  * @param changeSet       the change set for this stack.
- * @param isImport        if the stack is importing resources (a migrate stack).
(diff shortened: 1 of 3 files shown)


GavinZZ:
This change could technically be considered as breaking change although very unlikely. After the change, the new IAM policy would have the `Resource` field set to a specific KMS key instead of `*` when `key.grant` is used to a grantee from a different stack in a different region.

Because it originally grants `*` to `Resource` in the policy, users could rely on a different KMS key (which is quite unlikely but possible), this change cause issue. Perhaps I should wrap this using feature flag, open to suggestions.

paulhcsun:
We should add this behind a feature flag to avoid any possibility of introducing a breaking change.

paulhcsun (approved):
noice

PR #29332, opened 2024-03-01
chore: add some debugging to regression test skipping

rix0rrr:
The regression test skipping mechanism doesn't seem to be working. Investigate what's going on by adding print logging.

----


packages/@aws-cdk-testing/cli-integ/bin/apply-patches
```diff
@@ -14,6 +14,11 @@ fi
 
 candidate_dir="${scriptdir}/../resources/cli-regression-patches/v${version}"
 
-if [[ -f "$candidate_dir" ]]; then
-  cp -R "${candidate_dir}/"* "$2"
-fi
\ No newline at end of file
+if [[ -d "$candidate_dir" ]]; then
```
rix0rrr, line 17: Turning `-f` into `-d` was the load bearing fix

sumupitchayan (approved):

PR #29340, opened 2024-03-01
feat(autoscaling): support custom termination policy with lambda

### LINKED ISSUE #19750: AutoScaling: Support Custom Termination Policies (Lambda)
### Describe the feature

AWS Autoscaling supports custom termination policies by passing a lambda function ARN as a policy, however CDK only supports standard policies by means of an enum.

Ref: https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-termination-policies.html

### Use Case

In order to implement this functionality currently, I need to use escape hatches to insert the desired state, which is undesirable and error prone.

### Proposed Solution

It would be useful to be able to either pass a string value of the lambda arn (i.e. import value), or to pass a the reference of a lambda function in the same stack supporting its use of lambda versioning attributes.

### Other Information

_No response_

### Acknowledgements

### CDK version used

CDK 2.10.0, Python aws-cdk==1.143.0

Debian Bullseye 11.1

godwingrs22:
### Issue # (if applicable)

Closes #19750.

### Reason for this change

Amazon EC2 autoscaling supports [custom termination policy](https://docs.aws.amazon.com/autoscaling/ec2/userguide/lambda-custom-termination-policy.html) using lambda to terminate instances, but it is not supported yet by the `aws-autoscaling` package.

### Description of changes

This code change adds an enum `TerminationPolicy.CUSTOM_LAMBDA_FUNCTION` and an optional `terminationPolicyCustomLambdaFunctionArn` property to pass the lambda arn.

The change involves the logic to check If there are multiple termination policies specified since lamba termination policy has to be [first in order](https://docs.aws.amazon.com/autoscaling/ec2/userguide/lambda-custom-termination-policy.html#lambda-custom-termination-policy-limitations).

Added unit tests and integration test.
Created a simple CDK application with the updated constructs and verified that it synthesized and deployed correctly.

### Checklist

----


packages/aws-cdk-lib/aws-autoscaling/lib/auto-scaling-group.ts
```diff
@@ -1471,6 +1482,27 @@ export class AutoScalingGroup extends AutoScalingGroupBase implements
     }
 
     const { subnetIds, hasPublic } = props.vpc.selectSubnets(props.vpcSubnets);
+
+    const terminationPolicies: string[] = [];
+    if (props.terminationPolicies) {
+      props.terminationPolicies.forEach((terminationPolicy, index) => {
+        if (terminationPolicy === TerminationPolicy.CUSTOM_LAMBDA_FUNCTION) {
+          if (index != 0) {
```
GavinZZ, line 1490: nit: let's use `!==` instead. `==` is an operator that checks equality after type conversion. It doesn't make a difference here but it's good practice to use the strict equality operator `===` instead.

GavinZZ:
Overall looks good, just two minor comments.

packages/aws-cdk-lib/aws-autoscaling/README.md
```diff
@@ -539,10 +539,20 @@ to determine which instances it terminates first during scale-in events. You
 can specify one or more termination policies with the `terminationPolicies`
 property:
 
+[Custom termination policy](https://docs.aws.amazon.com/autoscaling/ec2/userguide/lambda-custom-termination-policy.html) with lambda
+can be used to determine which instances to terminate based on custom logic.
+The custom termination policy can be specified using `TerminationPolicy.CUSTOM_LAMBDA_FUNCTION`. If this is
+specified, you must also supply a value of lambda arn in the `terminationPolicyCustomLambdaFunctionArn` property.
+
+If there are multiple termination policies specified,
```
GavinZZ, line 547: I would also add some note here saying that to allow the lambda function to terminate ASG, it would needs appropriate permission, like example provided [here](https://docs.aws.amazon.com/autoscaling/ec2/userguide/lambda-custom-termination-policy.html#lambda-custom-termination-policy-create-function)

packages/aws-cdk-lib/aws-autoscaling/lib/auto-scaling-group.ts
godwingrs22: Good catch. Will update it

packages/aws-cdk-lib/aws-autoscaling/lib/auto-scaling-group.ts
```diff
@@ -1471,6 +1482,27 @@ export class AutoScalingGroup extends AutoScalingGroupBase implements
...
+            throw new Error('TerminationPolicy.CUSTOM_LAMBDA_FUNCTION must be specified first in the termination policies');
+          }
+
+          if (!props.terminationPolicyCustomLambdaFunctionArn ||
+            props.terminationPolicyCustomLambdaFunctionArn.length == 0) {
```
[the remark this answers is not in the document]
godwingrs22: Updated in the latest build

packages/aws-cdk-lib/aws-autoscaling/README.md
godwingrs22: Updated in the latest build

GavinZZ (approved):

PR #29347, opened 2024-03-03
feat(efs): replicating file systems

### LINKED ISSUE #21455: EFS: Create replication configuration
### Describe the feature

By the moment the aws cdk doesn't offer support to create a replication rule for the EFS. It would be great have this functionality out of the box and don't need to run a lambda to create the replication rule.

### Use Case

Create a replication configuration for EFS directly from the cdk code.

### Proposed Solution

_No response_

### Other Information

- https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_efs/README.html
- https://docs.aws.amazon.com/cli/latest/reference/efs/create-replication-configuration.html

### Acknowledgements

### CDK version used

2.31.0

Python cdk

badmintoncryer:
### Issue # (if applicable)

Closes #21455.

### Reason for this change

EFS supports [replicating file systems](https://docs.aws.amazon.com/efs/latest/ug/efs-replication.html) but AWS CDK cannot configure it.

### Description of changes

Add `replicationConfiguration` to `FileSystemProps`

```ts
declare const vpc: ec2.Vpc;
declare const kmsKey: kms.Key;

// auto generate a replication destination file system
new efs.FileSystem(this, 'ReplicationSourceFileSystem1', {
  vpc,
  replicationConfiguration: {
    kmsKey, // optional
    region: 'us-east-1', // optional
    availabilityZone: 'us-east-1a', // optional, Specifing the AZ means creating a One Zone file system as the replication destination
  }
});

// specify the replication destination file system
const destinationFileSystem = new efs.FileSystem(this, 'DestinationFileSystem', {
  vpc,
  // set as the read-only file system for use as a replication destination
  replicationOverwriteProtection: efs.ReplicationOverwriteProtection.DISABLED,
});

new efs.FileSystem(this, 'ReplicationSourceFileSystem2', {
  vpc,
  replicationConfiguration: {
    destinationFileSystem,
    // cannot configure other properties when destinationFileSystem is specified
  }
});
```

I have added both unit and integ tests.

### Checklist

----


packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -324,6 +324,13 @@ export interface FileSystemProps {
    * @default ReplicationOverwriteProtection.ENABLED
    */
   readonly replicationOverwriteProtection?: ReplicationOverwriteProtection;
+
+  /**
+   * Replication configuration for the file system.
+   *
+   * @default - no replication
+   */
+  readonly replicationConfiguration?: ReplicationConfiguration;
```
lpizzinidev, line 333: Why aren't we allowing multiple destinations? ([CFN declaration](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-efs-filesystem-replicationconfiguration.html))

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -350,6 +357,51 @@ export interface FileSystemAttributes {
...
+/**
+ * Replication configuration for the file system.
+ */
+export interface ReplicationConfiguration {
+  /**
+   * Whether to enable automatic replication.
+   *
+   * Other replication settings(`destinationFileSystem`, `kmsKey`, `region`, `az`) cannot be set if this is set to false.
+   */
+  readonly enable: boolean;
```
lpizzinidev, line 369: Is this necessary? 🤔
We don't have a corresponding flag in CFN and it's probably gonna work the same by leaving `replicationConfiguration: undefined`

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -579,6 +651,19 @@ export class FileSystem extends FileSystemBase {
...
     } : undefined;
 
+    const replicationConfiguration = enable ? {
+      destinations: [
+        {
+          fileSystemId: destinationFileSystem?.fileSystemId,
+          kmsKeyId: kmsKey?.keyArn,
+          region: region ??
+            // if destinationFileSystem is set, specify the region of the destination file system
+            (destinationFileSystem ? destinationFileSystem.env.region : Stack.of(this).region),
```
lpizzinidev, line 661:
```suggestion
          region: destinationFileSystem ? destinationFileSystem.env.region : (region ?? Stack.of(this).region)
```
Should work the same, but the logic it's closer to the above conditions.

lpizzinidev (changes requested):
Thanks 👍 I left some initial comments for some refactoring.
Please let me know if you have suggestions for further improvements.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
badmintoncryer: CloudFormation supports an array of destinations, but the maximum number of elements is restricted to 1. Therefore, I did not implement support for multiple destinations. 

Should I add support for it to accommodate future expansion?

lpizzinidev: Tricky question.
I guess we can keep it like this for now and deprecate it in the future in favor of an array of configurations if necessary.

badmintoncryer: I've updated to receive `replicationConfiguration` as array.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
badmintoncryer: You're right. What am I even doing...

One concern is that configuring replicationConfiguration as `[{}]` can still result in successful replication creation.
```ts
new efs.FileSystem(this, 'ReplicationSourceFileSystem1', {
  vpc,
  replicationConfiguration: [{}] // This means specify region as stack's one
});
```

This won't be a problem, but it does look somewhat odd.
What are your thoughts on this?

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
badmintoncryer: I'm sorry that I misunderstood your opinion. Which implementation will you adopt?

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
lpizzinidev: I think we should leave `readonly replicationConfiguration?: ReplicationConfiguration` for now since we can only define one configuration atm.
It would be cumbersome to define an array for which you can only pass one element.
Btw, can you provide some documentation on `replicationConfiguration` accepting only one configuration?

badmintoncryer: I've reverted the changes to accept `replicationConfiguration` as an object not an array.
And, I've added some notes to README.md. 

Could you please check it?

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -553,6 +598,26 @@ export class FileSystem extends FileSystemBase {
     if (props.throughputMode === ThroughputMode.ELASTIC && props.performanceMode === PerformanceMode.MAX_IO) {
       throw new Error('ThroughputMode ELASTIC is not supported for file systems with performanceMode MAX_IO');
     }
+
+    const { destinationFileSystem, region, availabilityZone, kmsKey } = props.replicationConfiguration ?? {};
+    if (props.replicationConfiguration) {
```
lpizzinidev, line 603:
```suggestion
    if (props.replicationConfiguration) {
        const { destinationFileSystem, region, availabilityZone, kmsKey } = props.replicationConfiguration;
```
No need to default.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -553,6 +598,26 @@ export class FileSystem extends FileSystemBase {
...
+      if (props.replicationOverwriteProtection === ReplicationOverwriteProtection.DISABLED) {
+        throw new Error('Cannot configure `replicationConfiguration` when `replicationOverwriteProtection` is set to `DISABLED`');
```
lpizzinidev, line 605:
```suggestion
        throw new Error('Cannot configure \'replicationConfiguration\' when \'replicationOverwriteProtection\' is set to \'DISABLED\'');
```
We usually use single quotes for parameter names in error messages.

packages/aws-cdk-lib/aws-efs/test/efs-file-system.test.ts
```diff
@@ -958,3 +958,162 @@ test.each([
     },
   });
 });
+
+describe('replication configuration', () => {
+  test('default settings', () => {
+    // WHEN
+    new FileSystem(stack, 'EfsFileSystem', {
+      vpc,
+      replicationConfiguration: {},
```
lpizzinidev, line 967: Not sure about this, it provides an unclear API.
We should validate that either `destinationFileSystem` or `region` are defined, and update documentation and test accordingly.

packages/aws-cdk-lib/aws-efs/README.md
```diff
@@ -76,6 +76,44 @@ This is to prevent deployment failures due to cross-AZ configurations.
 
 ⚠️ When `oneZone` is enabled, `vpcSubnets` cannot be specified.
 
+### [Replicating file systems](https://docs.aws.amazon.com/efs/latest/ug/efs-replication.html)
```
lpizzinidev, line 79:
```suggestion
### Replicating file systems
```
Let's add the link as a note below.

packages/aws-cdk-lib/aws-efs/README.md
```diff
@@ -76,6 +76,44 @@ This is to prevent deployment failures due to cross-AZ configurations.
...
+  vpc,
+  replicationConfiguration: {
+    destinationFileSystem,
+    // cannot configure other properties when destinationFileSystem is specified
+  }
+});
+```
+
+**Note**: EFS now supports only one replication destination and thus allows specifying just one `replicationConfiguration` for each file system.
+
```
lpizzinidev, line 116:
```suggestion

> Visit [Replicating file systems](https://docs.aws.amazon.com/efs/latest/ug/efs-replication.html) for more details.

```

lpizzinidev (changes requested):
Looks good overall 👍
I left some comments for further refinement.

packages/aws-cdk-lib/aws-efs/README.md
```diff
@@ -76,6 +76,48 @@ This is to prevent deployment failures due to cross-AZ configurations.
...
+declare const vpc: ec2.Vpc;
+declare const kmsKey: kms.Key;
+
+// auto generate a replication destination file system
+new efs.FileSystem(this, 'ReplicationSourceFileSystem1', {
+  vpc,
+  replicationConfiguration: {
+    kmsKey, // optional
+    region: 'us-east-1', // optional
```
lpizzinidev, line 94:
```suggestion
    region: 'us-east-1',
```
Now required if `destinationFileSystem` is not defined.

packages/aws-cdk-lib/aws-efs/README.md
```diff
@@ -76,6 +76,48 @@ This is to prevent deployment failures due to cross-AZ configurations.
...
+    // cannot configure other properties when destinationFileSystem is specified
+  }
+});
+```
+
+**Note**: EFS now supports only one replication destination and thus allows specifying just one `replicationConfiguration` for each file system.
+
+> Visit [Replicating file systems](https://docs.aws.amazon.com/efs/latest/ug/efs-replication.html) for more details.
+
+**Note**: You have to specify either `region` or `destinationFileSystem` when creating a replication destination file system.
```
lpizzinidev, line 119: It should be added with the above note.

lpizzinidev (approved):
Thanks 👍

badmintoncryer:
@lpizzinidev Thank you very much!

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -350,6 +357,44 @@ export interface FileSystemAttributes {
...
 }
 
+/**
+ * Replication configuration for the file system.
+ */
+export interface ReplicationConfiguration {
+  /**
+   * The existing destination file system for the replication.
+   *
+   * You cannot configure `kmsKey`, `region` and `az` when `destinationFileSystem` is set.
```
gracelu0, line 367:
```suggestion
   * You cannot configure `kmsKey`, `region` and `availabilityZone` when `destinationFileSystem` is set.
```

packages/aws-cdk-lib/aws-efs/test/efs-file-system.test.ts
```diff
@@ -958,3 +958,168 @@ test.each([
...
+        vpc,
+        replicationConfiguration: {
+          region: 'us-east-1',
+        },
+        replicationOverwriteProtection: ReplicationOverwriteProtection.DISABLED,
+      });
+    }).toThrow('Cannot configure \'replicationConfiguration\' when \'replicationOverwriteProtection\' is set to \'DISABLED\'');
+  });
+
+  test('throw error for specifing both destinationFileSystem and other parameters', () => {
```
gracelu0, line 1054:
```suggestion
  test('throw error for specifying both destinationFileSystem and other parameters', () => {
```

packages/aws-cdk-lib/aws-efs/test/efs-file-system.test.ts
```diff
@@ -958,3 +958,168 @@ test.each([
...
+        vpc,
+        replicationConfiguration: {
+          destinationFileSystem: destination,
+          region: 'us-east-1',
+        },
+      });
+    }).toThrow('Cannot configure \'replicationConfiguration.region\', \'replicationConfiguration.availabilityZone\' or \'replicationConfiguration.kmsKey\' when \'replicationConfiguration.destinationFileSystem\' is set');
+  });
+
+  test('throw error for specifing both destinationFileSystem and kmsKey', () => {
```
gracelu0, line 1073:
```suggestion
  test('throw error for specifying both destinationFileSystem and kmsKey', () => {
```

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -553,6 +598,30 @@ export class FileSystem extends FileSystemBase {
     if (props.throughputMode === ThroughputMode.ELASTIC && props.performanceMode === PerformanceMode.MAX_IO) {
       throw new Error('ThroughputMode ELASTIC is not supported for file systems with performanceMode MAX_IO');
     }
+
+    if (props.replicationConfiguration) {
+      const { destinationFileSystem, region, availabilityZone, kmsKey } = props.replicationConfiguration;
+      if (props.replicationOverwriteProtection === ReplicationOverwriteProtection.DISABLED) {
```
gracelu0, line 604: do we need to add validation to check that if `destinationFileSystem` is being passed in then it should be read-only (i.e.  replicationOverwriteProtection set to `DISABLED`)?

gracelu0 (commented):
some small typo fix suggestions and one question, overall looks good though

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
badmintoncryer: I understand your point, but the IFileSystem interface does not include a replicationOverwriteProtection property. Could we instead use the FileSystem class for the destinationFileSystem parameter?
(I generally think it's common to receive an ISomeClass as an argument.)

badmintoncryer:
@gracelu0 Thank you for your review! I've fixed typo and returned comment.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -350,6 +357,44 @@ export interface FileSystemAttributes {
   readonly fileSystemArn?: string;
 }
 
+/**
+ * Replication configuration for the file system.
+ */
+export interface ReplicationConfiguration {
```
gracelu0, line 363: Since there are multiple ways to configure the destination file system in `ReplicationConfiguration`, I'd suggest making this an enum-like class with static functions to enforce stronger typing and coupling between required properties. For example:

```ts
export abstract class ReplicationConfiguration {
    
    public static destinationFileSystem(fileSystem: IFileSystem): ReplicationConfiguration {
        // configure properties and return 
    }

    public static oneZoneFileSystem(availabilityZone: string, region: string, kmsKey?: kms.IKey): ReplicationConfiguration {
        // configure properties and return 
    }

    public static regionalFileSystem(region: string, kmsKey?: kms.IKey): ReplicationConfiguration {
        // configure properties and return  
    }
}
```

This should also reduce the need for some of the validation/error-handling below which is ideal.

gracelu0 (changes requested):
Hi @badmintoncryer , thank you so much for your work and patience in getting this PR reviewed. I asked for some second opinions from the team and have left some suggestions for how we can improve the user experience and reduce the amount of validation/error-handling needed. Let me know if you have any questions :)

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
badmintoncryer: Sounds good! I'll try to implement it.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -350,6 +357,84 @@ export interface FileSystemAttributes {
   readonly fileSystemArn?: string;
 }
 
+export class ReplicationConfiguration {
```
badmintoncryer, line 360: @gracelu0 I've implemented the `ReplicationConfiguration` class along with three factory methods. However, I have a couple of questions:

- In your example, `ReplicationConfiguration` is defined as an abstract class. What is the purpose of that? I've defined it as a regular class, but feel free to suggest modifications to my code.
- Even with the factory methods in place, users can still create instances of the `ReplicationConfiguration` class directly using `new ReplicationConfiguration({...})`. Therefore, I believe all the validation code is necessary, and I couldn't simplify it. What are your thoughts on this?

badmintoncryer:
@gracelu0 I've updated my implementation and have a [question](https://github.com/aws/aws-cdk/pull/29347#discussion_r1545550179). Could you please confirm it again?

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
gracelu0: Hi @badmintoncryer ! Thanks again for your patience in getting this PR reviewed :) I think an abstract class might make the most sense here still so users cannot create instances of `ReplicationConfiguration` class. As you pointed out, as long as users can create a new instance then we would still need the validation code. My updated suggestion is to create subclasses for each type of destination file system configuration (Existing, Regional, One Zone) along with corresponding interfaces to enforce the correct group of properties being set for each type:
```typescript

/**
 * Properties for the ReplicationConfiguration.
 */
export interface ReplicationConfigurationProps {
  /**
   * The existing destination file system for the replication.
   */
  readonly destinationFileSystem?: IFileSystem;

  /**
   * AWS KMS key used to protect the encrypted file system.
   *
   * @default - use service-managed KMS key for Amazon EFS
   */
  readonly kmsKey?: kms.IKey;

  /**
   * The AWS Region in which the destination file system is located.
   *
   * @default - the region of the stack
   */
  readonly region?: string;

  /**
   * The availability zone name of the destination file system.
   * One Zone file system is used as the destination file system when this property is set.
   *
   * @default - no availability zone is set
   */
  readonly availabilityZone?: string;
}

/**
 * Properties for configuring ReplicationConfiguration to replicate
 * to a new One Zone file system.
 */
export interface OneZoneFileSystemProps {
  /**
   * AWS KMS key used to protect the encrypted file system.
   *
   * @default - use service-managed KMS key for Amazon EFS
   */
  readonly kmsKey?: kms.IKey;

  /**
   * The AWS Region in which the destination file system is located.
   *
   * @default - the region of the stack
   */
  readonly region?: string;

  /**
   * The availability zone name of the destination file system.
   * One zone file system is used as the destination file system when this property is set.
   */
  readonly availabilityZone?: string;
}

/**
 * Properties for configuring ReplicationConfiguration to replicate
 * to a new Regional file system.
 */
export interface RegionalFileSystemProps {
  /**
   * AWS KMS key used to protect the encrypted file system.
   *
   * @default - use service-managed KMS key for Amazon EFS
   */
  readonly kmsKey?: kms.IKey;

  /**
   * The AWS Region in which the destination file system is located.
   *
   * @default - the region of the stack
   */
  readonly region?: string;
}

/**
 * Properties for configuring ReplicationConfiguration to replicate
 * to an existing file system.
 */
export interface ExistingFileSystemProps {
  /**
   * The existing destination file system for the replication.
   */
  readonly destinationFileSystem?: IFileSystem;
}

/**
 * EFS Replication Configuration
 */
export abstract class ReplicationConfiguration {
  /**
   * Specify the existing destination file system for the replication.
   *
   * @param destinationFileSystem The existing destination file system for the replication
   */
  public static existingFileSystem(destinationFileSystem: IFileSystem): ReplicationConfiguration {
    return new ExistingFileSystem({ destinationFileSystem });
  }

  /**
   * Create a new regional destination file system for the replication.
   *
   * @param region The AWS Region in which the destination file system is located. Default is the region of the stack.
   * @param kmsKey  AWS KMS key used to protect the encrypted file system. Default is service-managed KMS key for Amazon EFS.
   */
  public static regionalFileSystem(region?: string, kmsKey?: kms.IKey): ReplicationConfiguration {
    return new RegionalFileSystem({ region, kmsKey });
  }

  /**
   * Create a new one zone destination file system for the replication.
   *
   * @param region The AWS Region in which the specified availability zone belongs to.
   * @param availabilityZone The availability zone name of the destination file system.
   * @param kmsKey AWS KMS key used to protect the encrypted file system. Default is service-managed KMS key for Amazon EFS.
   */
  public static oneZoneFileSystem(region: string, availabilityZone: string, kmsKey?: kms.IKey): ReplicationConfiguration {
    return new OneZoneFileSystem({ region, availabilityZone, kmsKey });
  }

  /**
  * The existing destination file system for the replication.
  */
  public readonly destinationFileSystem?: IFileSystem;

  /**
   * AWS KMS key used to protect the encrypted file system.
   */
  public readonly kmsKey?: kms.IKey;

  /**
   * The AWS Region in which the destination file system is located.
   */
  public readonly region?: string;

  /**
   * The availability zone name of the destination file system.
   * One zone file system is used as the destination file system when this property is set.
   */
  public readonly availabilityZone?: string;

  constructor(props: ReplicationConfigurationProps) {
    this.availabilityZone = props.availabilityZone;
    this.destinationFileSystem = props.destinationFileSystem;
    this.kmsKey = props.kmsKey;
    this.region = props.region;
  }
}

/**
 * Represents an existing file system used as the destination file system
 * for ReplicationConfiguration.
 */
export class ExistingFileSystem extends ReplicationConfiguration {
  constructor(props: ExistingFileSystemProps) {
    super(props);
  }
}

/**
 * Represents a new Regional file system used as the
 * destination file system for ReplicationConfiguration.
 */
export class RegionalFileSystem extends ReplicationConfiguration {
  constructor(props: RegionalFileSystemProps) {
    super(props);
  }
}

/**
 * Represents a new One Zone file system used as the
 * destination file system for ReplicationConfiguration.
 */
export class OneZoneFileSystem extends ReplicationConfiguration {
  constructor(props: OneZoneFileSystemProps) {
    super(props);
  }
}
```
In the snippet above I've also tweaked some of the docstrings to be more specific. I think the above would allow us to remove the property validation statements. Let me know your thoughts on this.

TheRealAmazonKendra (commented):
Well shit. I have a bunch of review comments in pending. They're from a hot minute ago so some may be outdated.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
badmintoncryer: @gracelu0 Thank you for your suggestion. I have understand your point and agree with you.
I've updated my implementation to use abstract class.

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -350,6 +357,188 @@ export interface FileSystemAttributes {
...
+   *
+   * @default - the region of the stack
+   */
+  readonly region?: string;
+
+  /**
+   * The availability zone name of the destination file system.
+   * One zone file system is used as the destination file system when this property is set.
+   *
+   * @default - create regional file system when `destinationFileSystem` is not set
```
gracelu0, line 389:
```suggestion
   * @default - no availability zone is set
```

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
```diff
@@ -350,6 +357,188 @@ export interface FileSystemAttributes {
...
 }
 
+/**
+ * Properties for the ReplicationConfiguration.
+ */
+export interface ReplicationConfigurationProps {
+  /**
+   * The existing destination file system for the replication.
+   *
+   * @default - create a new file system for the replication
```
gracelu0, line 367:
```suggestion
```

gracelu0 (commented):
Final tiny docstring comments and then I'm good to approve :)

packages/aws-cdk-lib/aws-efs/lib/efs-file-system.ts
badmintoncryer: @gracelu0 
I believe the @default description is essential because the `destinationFileSystem` is optional. Removing this description could cause the CI to fail.

What do you think?

gracelu0: Ohh right, thanks for pointing that out! perhaps “None” or “No destinationFileSystem set” instead then?

badmintoncryer: Thanks! I've updated it.

gracelu0 (approved):
LGTM, thank you for contributing!

PR #29357, opened 2024-03-04
fix(glue): s3 path specified in --spark-event-logs-path needs to end with slash

### LINKED ISSUE #29356: (aws-glue-alpha): S3 path specified in `--spark-event-logs-path` needs to end with slash
### Describe the bug

When only bucket is provided for spark event log path in job, the s3 path does not end with slash.
It is accepted in job API, but it does not work on actual Spark UI.

### Expected Behavior

S3 path specified in `--spark-event-logs-path` ends with slash

### Current Behavior

S3 path specified in `--spark-event-logs-path` does not end with slash in case only bucket is provided but prefix is not provided

### Reproduction Steps

```
      sparkUI: {
        enabled: true,
      },
```

### Possible Solution

Add trailing slash when it does not end with slash

_No response_

### CDK CLI Version

2.131.0

### Framework Version

_No response_

### Node.js Version

20.11.1

### OS

macOS Ventura 13.6.4

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

moomindani:
### Issue # (if applicable)

Closes #29356.

### Reason for this change

Currently S3 path specified in `--spark-event-logs-path` does not end with slash in case only bucket is provided but prefix is not provided. This parameter causes errors when viewing the event log through Spark UI / Spark history server.

### Description of changes

Add trailing slash when it does not end with slash.

Completed unit test and integ test.

### Checklist

----


scanlonp (dismissed):
Hey @moomindani, this looks great! Have you confirmed that this works in Spark UI manually?

One thought: adding the trailing slash to the url seems like it could be used in more places than this. If so, we could add this functionality to the `s3UrlForObject()` function itself. Something along the lines of
`public s3UrlForObject(key?: string, trailingSlash?: boolean): string {`
in [bucket.ts](https://github.com/moomindani/aws-cdk/blob/9de77bb5e43971d615abb34b77b997f5143f775d/packages/aws-cdk-lib/aws-s3/lib/bucket.ts#L737).

Then this line would be `bucket.s3UrlForObject(props.prefix, true)`.

Lastly, this is changing current templates. Are we sure that 
1) This is not breaking
2) if it is, that old templates were not working in the first place

We have some leeway since this is an alpha package, but we want to be clear on the PR if this is breaking.

moomindani:
Hi @scanlonp, 
Thanks for quick feedback and comment

> Have you confirmed that this works in Spark UI manually?

Yes.

> Lastly, this is changing current templates. Are we sure that
> 1. This is not breaking
> 2. if it is, that old templates were not working in the first place

Yes, actually it introduces change to the template but as you noticed the old template without trailing slash did not work at all in all the places.


Let me try to make changes on `bucket.s3UrlForObject` then update this PR.

scanlonp:
Hey @moomindani, I fear I may have led you down the wrong path. I talked with some other team members, and my suggestion to change the function in the s3 package was the wrong call. We should keep this change scoped to glue. I feel pretty good approving your original PR if you would not mind going back to your first commit. Maybe adding a small comment on the purpose of the regex, if you would like.

Apologies again, but hope we can get this in smoothly from here!

moomindani:
@scanlonp Thanks for your comment, I understand it.
I fixed the PR with reverting S3 side change. Can you take another look?

packages/@aws-cdk/aws-glue-alpha/lib/job.ts
```diff
@@ -826,7 +826,7 @@ export class Job extends JobBase {
     bucket.grantReadWrite(role, this.cleanPrefixForGrant(props.prefix));
     const args = {
       '--enable-spark-ui': 'true',
-      '--spark-event-logs-path': bucket.s3UrlForObject(props.prefix),
+      '--spark-event-logs-path': bucket.s3UrlForObject(props.prefix).replace(/\/?$/, '/'),
```
scanlonp, line 829:
```suggestion
      '--spark-event-logs-path': bucket.s3UrlForObject(props.prefix).replace(/\/?$/, '/'), // path will always end with a slash
```

scanlonp (approved):

scanlonp (approved):

Issue #29331, opened 2024-03-01
cloudwatch: Use of EC2 action with Multiple dimension set in metric results into error


rtejwani1309:
### Describe the bug

While trying to create a Custom Metric with multiple dimension, and adding EC2 action, the CDK synth fails with the error below.

```
/Users/Desktop/tscdk/node_modules/aws-cdk-lib/aws-cloudwatch/lib/alarm.js:1
[...]

PR #29364, opened 2024-03-06
fix(cloudwatch): cloudwatch ec2 alarm action with multiple dimension results in error

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29331

### Reason for this change

While trying to create a Custom Metric with multiple dimension, and adding EC2 action, the CDK synth fails.

### Description of changes

As long as there's instance id in dimension, we should accept it instead of raising exception.

new tests and existing tests pass.

### Checklist

----


GavinZZ:
Exemption Request: No integration test needed as this changes is a bug fix and only affect when we raise exception.

shikha372 (approved):

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -267,7 +267,7 @@ export class Alarm extends AlarmBase {
     if (ec2ActionsRegexp.test(actionArn)) {
       // Check per-instance metric
       const metricConfig = this.metric.toMetricConfig();
-      if (metricConfig.metricStat?.dimensions?.length != 1 || metricConfig.metricStat?.dimensions![0].name != 'InstanceId') {
+      if (metricConfig.metricStat?.dimensions?.length != 1 || !metricConfig.metricStat?.dimensions?.some(dimension => dimension.name === 'InstanceId')) {
```
sebastian-kucharzyk-tcl, line 270: I've just stumbled upon this PR and I'm wondering if there shouldn't be `&&` operator instead of `||`? We want to throw the exception when there ale multiple dimensions AND none of them is `InstanceId`. Is that right?

PR #29372, opened 2024-03-06
feat(ssm): allow disabling context caching in valueFromLookup

### LINKED ISSUE #12366: (aws-ssm): valueFromLookup should have an option to skip context caching
When using `ssm.StringParameter.valueFromLookup()` I would like to have the option to not store the value of the parameter in the `cdk context` so the parameter is resolved every time that I `synth`. 

### Use Case
In particular, I store the value of which ECR Image Tag should be deployed in my Task Definition for my Fargate Service. Currently, if I forget to `cdk context --reset ...` then I can potentially deploy a previous version of the application. If I was able to skip storing the value in the context I would not have to manage this risk.

### Proposed Solution

```typescript
ssm.StringParameter.valueFromLookup(this, '/my/ssm/parameter', { skipContext: true });
```

---

andreialecu:
### Issue # (if applicable)

Closes #12366

### Reason for this change

There are cases when the ssm value should not be cached because it can accidentally roll back certain resources. See discussion in #12366

In my case, I'm using the AWS CLI to update a Cloudfront origin, but it's also part of a CDK stack. I want to avoid deploying the CDK stack and accidentally reverting the origin (which is also persisted to SSM, so the CDK stack can read it if needed).

### Description of changes

I added an optional parameter to `valueFromLookup(this, 'param', { disableContextCaching: true })`

It works by setting an adjacent context key of `key + TRANSIENT_CONTEXT_KEY`, piggybacking on similar behavior in the CDK. 

This seemed like the most straightforward implementation because `value` is usually a primitive value, and there was no way to attach this metadata. 

I didn't want to grow the scope of the change, but ideally, `value` should've been an object so that `{ [TRANSIENT_CONTEXT_KEY]: true, value }` could've been used directly. I wasn't sure of that behavior's implications and potential breaking changes.

Added tests.

### Checklist

----


andreialecu:
Exemption Request: I don't think an integration test is required here because there's no change to the synthesis. This is just a change relating to what gets saved in `cdk.context.json`

Also, not sure if a readme change is necessary?

packages/aws-cdk-lib/core/lib/context-provider.ts
```diff
@@ -27,8 +27,20 @@ export interface GetContextKeyOptions {
 }
 
 /**
+ * Configuration options for context providers.
  */
-export interface GetContextValueOptions extends GetContextKeyOptions {
+export interface ContextProviderConfig {
+  /**
+   * Whether the context provider is allowed to cache the value.
```
andreialecu, line 34:
```suggestion
   * Whether caching the value to the context is disabled.
```

packages/aws-cdk-lib/cloud-assembly-schema/schema/cloud-assembly.schema.json
```diff
@@ -482,6 +482,11 @@
                 "props": {
                     "$ref": "#/definitions/ContextQueryProperties",
                     "description": "A set of provider-specific options."
+                },
+                "disableContextCaching": {
+                    "description": "Whether to disable persisting this to the context file.",
+                    "default": false,
+                    "type": "boolean"
```
aaythapa, line 489: Curious about the schema changes, did you add these manually?

packages/aws-cdk-lib/aws-ssm/test/parameter.test.ts
```diff
@@ -652,6 +656,100 @@ test('fromLookup will use the SSM context provider to read value during synthesi
...
+
+  await mockCloudExecutable.configuration.load();
+
+  await cdkToolkit.assembly();
+
+  await mockCloudExecutable.configuration.saveContext();
+  const config = await new Configuration({ readUserContext: false }).load();
+
+  // THEN
+  expect(value).toEqual('dummy-value-for-my-param-name');
```
aaythapa, line 701: Confused by why this is true, did you set value to `dummy-value-for-my-param-name` somewhere?

packages/aws-cdk-lib/aws-ssm/test/parameter.test.ts
```diff
@@ -652,6 +656,100 @@ test('fromLookup will use the SSM context provider to read value during synthesi
...
+  expect(requestedParameterName).toEqual('my-param-name');
+  expect(config.context.keys).toEqual(['ssm:account=12344:parameterName=my-param-name:region=us-east-1']);
+});
+
+test('fromLookup will not persist value in context if requested', async () => {
```
aaythapa, line 706: With `disableContextCaching` would the value in `cdk context` get overwritten when the parameter is resolved or is it the value never gets saved in `cdk context`?

If its the former I'd prefer if this test showed that with `disableContextCaching` the context will have a new value everytime it resolves the parameter

packages/aws-cdk/lib/context-providers/index.ts
```diff
@@ -69,6 +69,11 @@ export async function provideContextValues(
       }, resolvedEnvironment, sdk);
 
       value = await provider.getValue({ ...missingContext.props, lookupRoleArn: arns.lookupRoleArn });
+
+      if (missingContext.disableContextCaching) {
+        debug(`Skipping context caching for ${key}`);
+        context.set(key + TRANSIENT_CONTEXT_KEY, { [TRANSIENT_CONTEXT_KEY]: true });
```
aaythapa, line 75: Does this property work for any context lookup? Like it was suggested in [this comment](https://github.com/aws/aws-cdk/issues/12366#issuecomment-755093975)

packages/aws-cdk/test/context.test.ts
```diff
@@ -137,3 +137,18 @@ test('transient values arent saved to disk', async () => {
   // THEN
   expect(config2.context.get('some_key')).toEqual(undefined);
 });
+
+test('transient keys arent saved to disk', async () => {
```
aaythapa, line 141: Not sure about the purpose of this test since it doesn't use `disableContextCaching`?

aaythapa (commented):
Thank you for the PR!

A couple of comments, mostly from my own lack of understanding for this part of the code base

packages/aws-cdk-lib/cloud-assembly-schema/schema/cloud-assembly.schema.json
andreialecu: Not manual, no. If I remember correctly this was the result of running `yarn update-schema` which I think was required for getting tests to pass.

packages/aws-cdk-lib/aws-ssm/test/parameter.test.ts
andreialecu: See here: https://github.com/aws/aws-cdk/blob/ab940701ce3b423243c657bf10cf1f2aaa9553de/packages/aws-cdk-lib/aws-ssm/README.md?plain=1#L61-L65

packages/aws-cdk/lib/context-providers/index.ts
andreialecu: It should be able to be used it in other context lookups, but this PR only implements it for SSM. I'm not sure what other context lookups would benefit from this, but with the underlaying changes in this PR, it would be easy to make it work elsewhere.

packages/aws-cdk/test/context.test.ts
andreialecu: This is a copy of the test above it ("transient values arent saved to disk") but it checks that the specified keys/values don't get saved to the context, regardless of what the value is.

packages/aws-cdk-lib/aws-ssm/test/parameter.test.ts
andreialecu: Also, if you look at the test before this one, a similar assertion is made (line 641)

packages/aws-cdk-lib/aws-ssm/test/parameter.test.ts
andreialecu: If there's a previous value in the context, `valueFromLookup()` will use that one and not do another lookup.

This PR only allows disabling persisting to the context cache if the value needs to be looked up.

andreialecu: I added a new test to confirm what I mentioned in my comment above.

andreialecu: I'm open to renaming this to something like skipContextPersistance or some other name that would be more descriptive.

TheRealAmazonKendra:
These two PRs (https://github.com/aws/aws-cdk/pull/29372) and (https://github.com/aws/aws-cdk/pull/28766) are not doing the same thing, but there is definitely overlap. Given that fact, it changes the feedback I was going to give to both a bit.

I absolutely see the value in both these changes but want to make sure we're getting the contract right across both. I know you've both been waiting for feedback for some time but I'm going to tax your patience a tad further so I can discuss with the team tomorrow.

aros3rg3d:
Just came across this feature while searching how to disable caching for ssm lookup. Thanks @andreialecu, can't wait to use it !

sakurai-ryo:
@andreialecu @aaythapa @TheRealAmazonKendra 
I’m really interested in using this feature it will introduces.
Could I ask if there are any updates or blockers on this PR?
I would be happy to help and potentially take over the work to get it across the finish line.

GavinZZ:
@andreialecu I apologize that this PR was just sitting here since Kendra reviewed and we did not make any new progress. Are you still interested in contributing this feature? If yes, please address the conflicts and ping me for another review. I'll make sure to prioritize reviewing this PR once the conflicts are addressed.

andreialecu:
> @andreialecu I apologize that this PR was just sitting here since Kendra reviewed and we did not make any new progress. Are you still interested in contributing this feature? If yes, please address the conflicts and ping me for another review. I'll make sure to prioritize reviewing this PR once the conflicts are addressed.

The changes merged while this PR was overlooked have introduced some awkwardness in usage. Someone added functionality to return a default value, which could have been integrated into the options object I proposed in this PR under a `defaultValue` field.

I must note my disappointment with the turnaround on this.

I'll try to fix the conflicts.

andreialecu:
@GavinZZ this PR introduced [changes to the `MissingContext` type](https://github.com/aws/aws-cdk/pull/29372/files#diff-f5fa3de06b950d4c005a2524973b85e69e7ccca1a2a1ee7b0c5af8683ee3863bR63-R69) which now appears to have been moved elsewhere:
https://github.com/cdklabs/cloud-assembly-schema/blob/main/lib/cloud-assembly/schema.ts

I'm not sure how to proceed, should I open a PR there first and wait for that to be merged before doing this?

It won't work without the change.

GavinZZ:
@andreialecu Thanks for getting back to me. I had an internal discussion with my colleague and we have some questions on why the disable context caching is needed given that CloudFormation supports dynamic reference. My colleague will post more detail on this discussion shortly in this issue https://github.com/aws/aws-cdk/issues/12366 and let's continue the discussion from there.

PR #29375, opened 2024-03-06
fix(cli): `diff` allow specifying `changeSet` via `cdk.json`

### LINKED ISSUE #29232: cli: diff --change-set flag not configurable via cdk.json
### Describe the bug

A new flag was introduced as part of #28336

This flag doesn't seem to be set to wanted default value using the cdk.json.

There's also missing documentation about this flag in the [CLI reference](https://docs.aws.amazon.com/cdk/v2/guide/cli.html). Arguably a different issue since it'll be a different repo? I'm not sure how to 

### Expected Behavior

Ability to override default value in `cdk.json`

### Current Behavior

Anything set in in the `cdk.json` will be ignored

### Reproduction Steps

1. Use any stack setup
2. Run a diff
3. You'll see the following output on the stack:

```
```

4. Try to override the `change-set` flag in `cdk.json` to false
5. Run diff again

expected: warning message above dissapears
actual: still seeing the warning (behaviour hasn't changed)

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.128.0

### Framework Version

_No response_

### Node.js Version

v20.11.1

### OS

darwin AMD64

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

andreialecu:
### Issue # (if applicable)

Closes aws/aws-cdk-cli#755

### Reason for this change

Allows specifying whether to use changesets via cdk.json:

```json
"diff": {
  "changeSet": false
}
```

It also moves this informational message to the `debug()` log level:

```
```

There was a `--quiet` CLI argument that was not squelching it, which seemed like an oversight in the initial PR. 

With many stacks, the message can be quite verbose.`--verbose` will reveal it, but otherwise it will be hidden

### Description of changes

The default value of the change-set flag is removed and instead is handled in the diff call.

The value for `changeSet` in the diff call is processed in this way: setting the flag gives the value > if the flag is un-set, check the context in cdk.json > if both are un-set, default to true.

This poses a question about what flags we allow to be set in cdk.json. Here, "diff" is new in cdk.json, and "changeSet" is the first and only flag able to be set. The team can have a discussion and come to a conclusion.

Added tests.

Manual testing.

### Checklist

----


andreialecu:
Exemption Request: There are integration tests here but there should be no need to touch any snapshots for this change.

TheRealAmazonKendra:
@Mergifyio update

TheRealAmazonKendra:
Thank you for picking up this change. I'm running this through our test pipeline before doing a full review. I've also added the integ test exemption because, you are correct, this does not require a change to snapshots.

scanlonp:
@Mergifyio update

TheRealAmazonKendra:
@Mergifyio update

andreialecu:
Thanks for taking a look at this @TheRealAmazonKendra 

I have another PR open at #29372, which is a bit more important to me; if there was any chance someone could take a look.

packages/aws-cdk/lib/api/util/cloudformation.ts
```diff
@@ -341,7 +341,7 @@ async function uploadBodyParameterAndCreateChangeSet(options: PrepareChangeSetOp
     const exists = (await CloudFormationStack.lookup(cfn, options.stack.stackName, false)).exists;
 
     const executionRoleArn = preparedSdk.cloudFormationRoleArn;
-    options.stream.write('Hold on while we create a read-only change set to get a diff with accurate replacement information (use --no-change-set to use a less accurate but faster template-only diff)\n');
+    debug('Hold on while we create a read-only change set to get a diff with accurate replacement information (use --no-change-set to use a less accurate but faster template-only diff)');
```
blimmer, line 344: 💯 - thank you! This message appears too often now.

packages/aws-cdk/lib/cli.ts
```diff
@@ -514,7 +514,7 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
           stream: args.ci ? process.stdout : undefined,
           compareAgainstProcessedTemplate: args.processed,
           quiet: args.quiet,
-          changeSet: args['change-set'],
+          changeSet: configuration.settings.get(['diff', 'changeSet']) ?? args['change-set'],
```
andreialecu, line 517: Perhaps this needs to be swapped so that the cdk.json setting can be overridden via the cli argument?

andreialecu, line 517:
```suggestion
          changeSet: args['change-set'] ?? configuration.settings.get(['diff', 'changeSet'],
```

scanlonp:
Hey @andreialecu, I have eyes on this. Will try to give a full review tomorrow, but at a glance this looks helpful and straightforward!

packages/aws-cdk/lib/cli.ts
scanlonp: I think you are right about this order of operations. Only question is if there is a difference between no flag being provided and the default. 

If the flag is always given a value regardless of being set, we may have to put the context first (but I will ask some others on the team about this).

scanlonp:
I pushed some changes so the flag and context are both respected. Will put a full explanation in the PR description.

TheRealAmazonKendra:
Shushing up the PR bot that says changes have been requested for too long. That's on us. The bot will still want us to run the PR through our test pipeline, but now it won't close it.

comcalvi:
One of the CLI integ tests is failing for this PR:

```
  ✕ cdk diff should only show verbose messages in --verbose mode
Expected substring: "Hold on while we create a read-only change set"
Received string:    "[17:46:04] CDK toolkit version: 2.139.0-rc.0 (build 9cd365e)
```

Could you fix this one?

kaizencc:
Hi @andreialecu, I see the following outstanding issues on this PR 

-- merge conflicts 
-- cli integ test failing

I know a lot of these are on us :( and for that I apologize. Are you still interested in this PR or should I close it? Knowing that there will be quite a bit of migration to do, in addition to some changes the maintainer team is planning on achieving this week.

kaizencc:
yes @mrgrain. i'm going to close this PR as the cli code is no longer in this repository but I think the better thing to do to support this is to eventually pick back up #32760

PR #22031, opened 2022-09-14
fix: throw an error when using an encrypted s3 bucket for capturing alb logs

devnotfound:
Currently if one tries to capture alb logs into an encrypted S3 bucket that uses a KMS key, they get an `Access Denied for bucket` error. This association is currently unsupported, however it delays the error at the time of the bucket association. 

This PR now introduces an upfront check that would error saying `Encryption key detected. Bucket encryption using KMS keys is unsupported` before the alb to s3 bucket log association is established. 

This closes #21947 

----

### All Submissions:

* [✔️] Have you followed the guidelines in our [Contributing guide?](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md)

### Adding new Unconventional Dependencies:

* [✖️ ] This PR adds new unconventional dependencies following the process described [here](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md/#adding-new-unconventional-dependencies)

### New Features

* [ ✖️ ] Have you added the new feature to an [integration test](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md)?
	* [ ] Did you use `yarn integ` to deploy the infrastructure and generate the snapshot (i.e. `yarn integ` without `--dry-run`)?

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

Issue #21947, opened 2022-09-07
elbv2: cannot use load balancer access logs when bucket is encrypted with KMS key 


josefaidt:
### Describe the bug

I am running into almost exactly what is described in [this previous issue](https://github.com/aws/aws-cdk/issues/8113) and similar to [this StackOverflow post](https://stackoverflow.com/questions/72009403/how-do-i-enable-access-log-for-load-balancer-using-applicationloadbalancedfargat) where I have:

```ts
albFargateService.loadBalancer.logAccessLogs(bucket, 'alb-access')
```

And am receiving 

```
8:58:34 AM | UPDATE_FAILED        | AWS::ElasticLoadBalancingV2::LoadBalancer     | BotAlbFargateServiceLB34
7310A1
Access Denied for bucket: <my-fake-bucket>. Please check S3bucket permission (Service: AmazonElasticL
oadBalancing; Status Code: 400; Error Code: InvalidConfigurationRequest; Request ID: e95cece4-c4af-4b8f-a6f1
-d1eb0d2e2778; Proxy: null)
```

Am I missing an additional piece of configuration for the bucket to enable access logs?

### Expected Behavior

Access logs are enabled and emitted to bucket

### Current Behavior

Running `cdk deploy` with the snippet shown in the section above we are presented with what appears to be the correct permissions

```
IAM Statement Changes
┌───┬──────────────────────┬────────┬──────────────────────┬──────────────────────┬────────────────────────┐
│   │ Resource             │ Effect │ Action               │ Principal            │ Condition              │
├───┼──────────────────────┼────────┼──────────────────────┼──────────────────────┼────────────────────────┤
│ + │ ${Bucket.Arn}        │ Allow  │ s3:GetBucketAcl      │ Service:delivery.log │                        │
│   │                      │        │                      │ s.amazonaws.com      │                        │
├───┼──────────────────────┼────────┼──────────────────────┼──────────────────────┼────────────────────────┤
│ + │ ${Bucket.Arn}/elb-ac │ Allow  │ s3:Abort*            │ AWS:arn:${AWS::Parti │                        │
│   │ cess/AWSLogs/1234567 │        │ s3:PutObject         │ tion}:iam::123456789 │                        │
│   │ 53096/*              │        │ s3:PutObjectLegalHol │ 021:root             │                        │
│   │                      │        │ d                    │                      │                        │
│   │                      │        │ s3:PutObjectRetentio │                      │                        │
[...]

PR #29382, opened 2024-03-06
fix: add validation for ALB access log bucket when KMS key is provided

shikha372:
### Issue # (if applicable)

Closes #22031, #21947.

### Reason for this change

Adds a validation with correct error indicating ALB Access log bucket does not support KMS encryption

### Description of changes

Currently access logs bucket encryption with KMS is not supported in case of ALB but while deploying it throws an error indicating the failure with bucket permissions. 
This validation introduces an upfront check to throw an error if `bucket.encryptionKey `is defined.
Documentation: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html

Added unit tests for validation.

### Checklist

----


packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-load-balancer.ts
```diff
@@ -151,6 +154,66 @@ export class ApplicationLoadBalancer extends BaseLoadBalancer implements IApplic
...
+  public logAccessLogs(bucket: s3.IBucket, prefix?: string) {
+
+    /**
+    * KMS key encryption is not supported on Access Log bucket for ALB, the bucket must use Amazon S3-managed keys (SSE-S3).
+    * See https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html#bucket-permissions-troubleshooting
+    */
+
+    if (bucket.encryptionKey) {
+      throw new Error('Encryption key detected. Bucket encryption using KMS keys is unsupported');
+    }
```
GavinZZ, line 172: Is this the only different in the `logAccessLogs` call for base load balancer vs application load balancer class?

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-load-balancer.ts
```diff
@@ -151,6 +154,66 @@ export class ApplicationLoadBalancer extends BaseLoadBalancer implements IApplic
...
+   * Enable access logging for this load balancer.
+   *
+   * A region must be specified on the stack containing the load balancer; you cannot enable logging on
+   * environment-agnostic stacks. See https://docs.aws.amazon.com/cdk/latest/guide/environments.html
+   */
+  public logAccessLogs(bucket: s3.IBucket, prefix?: string) {
+
+    /**
+    * KMS key encryption is not supported on Access Log bucket for ALB, the bucket must use Amazon S3-managed keys (SSE-S3).
+    * See https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html#bucket-permissions-troubleshooting
```
GavinZZ, line 167: Can we also add this to `README.md`

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-load-balancer.ts
shikha372: yes only this part to check whether encryption key is enabled

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-load-balancer.ts
shikha372: Didn't find any prior section to enable access logging on ALB in Readme, added it with above doc reference.

GavinZZ (approved):

shikha372:
@Mergifyio update

shikha372:
@Mergifyio update

Issue #29388, opened 2024-03-06
(xray): Invalid template synthesis for X-Ray resources when using Tags.of()


GavinZZ:
### Describe the bug

When using Tags() methods for aws_xray .CfnGroup class and aws_xray .CfnSamplingRule , the tag's Key and Value values are not passcal-cased.

However, in 2.130.0, when using Tags() method behavior for CfnGroup class and CfnSamplingRule, the tag's are not added.

I find that Tags() methods changed in release 2.131.0 https://github.com/aws/aws-cdk/pull/28989/files

### Expected Behavior

cdk synth should run successfully

### Current Behavior

cdk synth fails due to 
```
Tags:
        - key: Environment # expected is Key: Environment
          value: Prod # expected is Value: Prod

```

### Reproduction Steps

```
import { Construct } from 'constructs';
import { aws_xray as xray } from 'aws-cdk-lib';

export class AppStack extends cdk.Stack {
  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
    super(scope, id, props);

    const xRay = new xray.CfnGroup(this, 'XRayGroup', {
      groupName: 'ECSSample',
      filterExpression: `ok`,
      insightsConfiguration: {
        insightsEnabled: false,
        notificationsEnabled: false,
      },
    });

    new xray.CfnSamplingRule(this, 'ApplicationRule', {
      samplingRule: {
        ruleName: 'app',
        resourceArn: '*',
        priority: 10,
        fixedRate: 0,
        reservoirSize: 10,
        serviceName: '*',
        serviceType: '*',
        host: '*',
        httpMethod: '*',
        urlPath: '*',
        version: 1,
      },
    });

    cdk.Tags.of(this).add("Environment", "Prod");
  }
}
```

### Possible Solution

Use latest type instead of legacy type for this `Tags` property.

### CDK CLI Version

2.131.0

### Node.js Version

18

### OS

MacOs

### Language

TypeScript

PR #29389, opened 2024-03-06
fix(spec2cdk): use modern type when building tag type

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29388

### Reason for this change

Some of the modern tags failed to run `cdk synth` due to type misconfiguration.

### Description of changes

Always default to use the latest type for modern tags.

Fixed for failed resources.

### Checklist

----


tools/@aws-cdk/spec2cdk/lib/cdk/resource-decider.ts
```diff
@@ -179,7 +179,7 @@ export class ResourceDecider {
 
   private handleTagPropertyModern(cfnName: string, prop: Property, variant: TagVariant) {
     const originalName = propertyNameFromCloudFormation(cfnName);
-    const originalType = this.converter.typeFromProperty(prop);
+    const originalType = this.converter.typeFromPropertyForModernTags(prop);
```
GavinZZ, line 182: This should be safe because `handleTagPropertyModern` is only enabled in version 2.131.0 which is the current CDK version, see this PR https://github.com/aws/aws-cdk/pull/28989/files.

Before version 2.131.0 there would be no way to reach this function call due to `HAS_25610` is always set to `false` (see the PR I linked). And for 2.131.0 there wouldn't be a deployable template since it will fails at deployment step, see description.

rix0rrr (changes requested):
I think your reasoning makes sense. But please add a unit test for the type this was reported against.

xazhao (approved):

PR #29394, opened 2024-03-07
fix(cli): prevent changeset diff for non-deployed stacks

### LINKED ISSUE #29265: (cli): change-set diff not required for new stack diffs

scanlonp:
### Reason for this change

When a stack does not exist in CloudFormation, creating a changeset makes an empty `REVIEW_IN_PROGRESS` stack. We then call `delete-stack` to clean up the empty stack. However, this can cause a race condition with a deploy call. 

### Description of changes

This change prevents changeset diffs for stacks that do not yet exist in CloudFormation. This overrides the changeset diff flag. This change also adds logic for migrate stacks in the old diff logic to represent resource imports without needing the changeset present.

Testing with new stacks only uses changeset diffs once the stack is deployed. Testing with new migrate stacks only uses changeset diffs once deployed. Pre-deployment the resources correctly show as imports.  

Note: the deleted test assumes the diff will be calculated using the mocked changeset. The new logic avoids the changeset, so the test is no longer relevant.

Closes #29265.

----


TheRealAmazonKendra (approved):

scanlonp:
Changes succeeded in the cli test pipeline.

PR #29395, opened 2024-03-07
chore(rds): `IO2` instance storage

### LINKED ISSUE #29396: rds: support for IO2 instance storage
### Describe the feature

Support for IO2 instance storage type.

https://aws.amazon.com/jp/blogs/aws/amazon-rds-now-supports-io2-block-express-volumes-for-mission-critical-database-workloads/

### Use Case

AWS CDK cannot configure IO2 storage type for RDS instance .

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.131.0

irrelevant

badmintoncryer:
### Issue # (if applicable)

Closes #29396.

### Reason for this change

[AWS RDS now supports IO2 instance storage](https://aws.amazon.com/jp/blogs/aws/amazon-rds-now-supports-io2-block-express-volumes-for-mission-critical-database-workloads/) but CDK cannot configure it yet.

### Description of changes

- Added IO2 storage type to the `StorageType` enum.
- Set default IOPS to 1000, which is minimum value in the allowed IOPS range.

Added both unit and integ test

### Checklist

----


badmintoncryer:
Exemption Request
I think it is unnecessary to update README.md.

packages/aws-cdk-lib/aws-rds/test/instance.test.ts
```diff
@@ -194,6 +194,22 @@ describe('instance', () => {
...
+      storageType: rds.StorageType.IO2,
+    });
+
+    // THEN
+    Template.fromStack(stack).hasResourceProperties('AWS::RDS::DBInstance', {
+      Iops: 1000,
+      StorageType: 'io2',
+    });
+  });
+
```
go-to-k, line 212: There is already a similar test here, so it might be better to write it there.

https://github.com/go-to-k/aws-cdk/blob/0605a8b869aac6bccc9631fcea719895ca397f73/packages/aws-cdk-lib/aws-rds/test/instance.test.ts#L2011

Or maybe we could create tests that are separate from this test (`'gp3 storage type'`) to check at once if the `defaultIops` method outputs iops for each storage type. (Leave the test (`'gp3 storage type'`) to check the `storageThroughput`, and change the test name to another name related to `gp3` and `storageThroughput`.)

For example:
```ts
  test.each([
    [rds.StorageType.STANDARD, 'standard', 2000, undefined],
    [rds.StorageType.GP2, 'gp2', 2000, undefined],
    [rds.StorageType.GP3, 'gp3', 2000, 2000],
    [rds.StorageType.IO1, 'io1', 2000, 2000],
    [rds.StorageType.IO1, 'io1', undefined, 1000],
    [rds.StorageType.IO2, 'io2', 2000, 2000],
    [rds.StorageType.IO2, 'io2', undefined, 1000],
  ])('storage type and IOPS for %s storage type', (inStorageType, outStorageType, inIops, outIops) => {
    new rds.DatabaseInstance(stack, 'Instance', {
      engine: rds.DatabaseInstanceEngine.mysql({ version: rds.MysqlEngineVersion.VER_8_0_30 }),
      vpc,
      storageType: inStorageType,
      iops: inIops,
    });

    Template.fromStack(stack).hasResourceProperties('AWS::RDS::DBInstance', {
      StorageType: outStorageType,
      Iops: outIops ?? Match.absent(),
    });
  });
```

It may not be to your taste, which do you think would be better? Of course, the above code can be modified.

go-to-k (changes requested):
Looks good. I made one minor comment.

go-to-k:
> Set default IOPS to 1000, which is minimum value in the allowed IOPS range.

By the way, what happens if we specify a value lower than 1000?
If CloudFormation fails, should we do that validation?

packages/aws-cdk-lib/aws-rds/test/instance.test.ts
badmintoncryer: @go-to-k Thank you for the excellent suggestion!! I replaced unit test and renamed existing `gp3` test.

go-to-k:
@badmintoncryer Yes! I think so. I will approve this PR!

go-to-k (approved):

badmintoncryer:
@go-to-k Thank you Hero!!

aaythapa (approved):
Thanks for the PR and to @go-to-k for a great review! I do agree we need validation for the different storage types, approving this for now

PR #29406, opened 2024-03-08
fix(custom-resources): log statement exposes information prohibited by security guideline

GavinZZ:
### Reason for this change

Current log statement will log too much content, including response URL which may not be ideal to be logged according to AWS Security Guideline. Removing the `input event` from the log statement.

### Description of changes

Remove the log statement.

N/A

### Checklist

----


packages/aws-cdk-lib/custom-resources/lib/provider-framework/runtime/framework.ts
```diff
@@ -50,8 +50,6 @@ async function onEvent(cfnRequest: AWSLambda.CloudFormationCustomResourceEvent)
     input: JSON.stringify(resourceEvent),
   };
 
-  log('starting waiter', waiter);
```
scanlonp, line 53: Maybe some log line that still has waiter info as a stop gap.
Like `log('starting waiter', { stateMachineArn: waiter.stateMachineArn, name: waiter.name);`?

GavinZZ: Yes, sounds good since the issue is mainly from the `resourceEvent`.

GavinZZ:
This seems to be a breaking change, will wrap it with a feature flag.

TheRealAmazonKendra:
Changes to the hash doesn't necessarily indicate a breaking change. I would pull this PR down and run those tests with `--force --update-on-failed` to see if there is actually an impact to the functionality or if it just changes the contents of those lambdas, which we would expect on a change like this.

packages/@aws-cdk-testing/framework-integ/test/lambda-layer-awscli/test/integ.awscli-layer.js.snapshot/lambda-layer-awscli-integ-stack.template.json
```diff
@@ -12,7 +12,7 @@
     "Description": "/opt/awscli/aws"
    }
   },
-  "Lambdapython37ServiceRoleB5A704D4": {
+  "Lambdapython39ServiceRoleE2CFED77": {
```
GavinZZ, line 15: Expected change, since we updated python runtime to 3.9 and 3.10 for the test. 3.7 is deprecated and results in a deployment failure.

paulhcsun (approved):

PR #29419, opened 2024-03-09
feat(cloudfront): adding support for inline KeyValueStore sources

### LINKED ISSUE #29204: (cloudfront): Support inline data as KeyValueStore import source
### Describe the feature

It would be nice to do something like this:

```ts
const store = new cloudfront.KeyValueStore(this, 'KeyValueStore', {
  keyValueStoreName: 'KeyValueStore',
  source: cloudfront.ImportSource.fromInline({
    "data":[
      {
        "key": "key1",
        "value": "value"
      },
      {
        "key": "key2",
        "value": "value"
      }
    ]
  }),
});
```

### Use Case

I'd like to store a hashing key, which don't want to be versioned in git.
It is saved in Systems Manager Parameter Store, and I retrieve it before initializing the stacks, so it is a parameter to the stack which should create the `KeyValueStore`.

I know I could create a temp JSON file, but this has already been made for Lambda's code, so it could be in here too.
https://github.com/aws/aws-cdk/blob/v2.129.0/packages/aws-cdk-lib/aws-lambda/lib/code.ts#L36
https://github.com/aws/aws-cdk/blob/v2.129.0/packages/aws-cdk-lib/aws-lambda/lib/code.ts#L243

### Proposed Solution

Add an `InlineImportSource` class aswell as an static method `ImportSource.fromInline()`, as it is done with Lambda code.

### Other Information

_No response_

### Acknowledgements

### CDK version used

1.129.0

ssennettau:
### Issue # (if applicable)

Closes #29204.

### Reason for this change

CloudFront's `KeyValueStore` has the option to initialize the store using a file, imported with an `ImportSource`. The CloudFormation resource does not support inline declarations, so the current L2 construct does not support inline either. To provide easier 

### Description of changes

Added the `InlineImportSource`, along with the companion static method to `cloudfront.KeyValueStore`.

#### 📄 Temporary Files

Since inline declarations are not supported by the underlying L1 construct, I have resorted to creating a temporary file which will be handled as a CDK asset. This is done utilizing elements of the cdk core, and follows the pattern used by the `s3-deployment`. This does also couple the KeyValueStore to the assets in S3, though the only other option would seem to be creating its own S3 bucket.

This is the only case I'm aware of in CDK of handling a temporary file like this. As far as I'm aware, all design principles are met, but this seems something of an edge case.

Open to suggestions and reviews.

New Unit Tests and Integration Test implemented

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
+++ b/packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
@@ -1,11 +1,13 @@
 // The SDK is only used to reference `DescribeChangeSetOutput`, so the SDK is added as a devDependency.
 // The SDK should not make network calls here
-import type { CloudFormation } from 'aws-sdk';
+import type { DescribeChangeSetOutput as DescribeChangeSet } from '@aws-sdk/client-cloudformation';
 import * as impl from './diff';
 import * as types from './diff/types';
 import { deepEqual, diffKeyedEntities, unionOf } from './diff/util';
 
 export * from './diff/types';
+
+export type DescribeChangeSetOutput = DescribeChangeSet;
 
 type DiffHandler = (diff: types.ITemplateDiff, oldValue: any, newValue: any) => void;
 type HandlerRegistry = { [section: string]: DiffHandler };
@@ -45,7 +47,7 @@ const DIFF_HANDLERS: HandlerRegistry = {
 export function fullDiff(
   currentTemplate: { [key: string]: any },
   newTemplate: { [key: string]: any },
-  changeSet?: CloudFormation.DescribeChangeSetOutput,
+  changeSet?: DescribeChangeSetOutput,
   isImport?: boolean,
 ): types.TemplateDiff {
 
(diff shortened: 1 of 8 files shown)


packages/aws-cdk-lib/aws-cloudfront/README.md
```diff
@@ -548,14 +548,30 @@ To create an empty Key Value Store:
 const store = new cloudfront.KeyValueStore(this, 'KeyValueStore');
 ```
 
-To also include an initial set of value, the `source` property can be specified. For the
-structure of this file, see [Creating a file of key value pairs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/kvs-with-functions-create-s3-kvp.html).
+To also include an initial set of value, the `source` property can be specified, either from a
```
lpizzinidev, line 551:
```suggestion
To also include an initial set of values, the `source` property can be specified, either from a 
```

lpizzinidev (approved):
Thanks 👍 Just a minor documentation adjustment

GavinZZ (approved):
Amazing, thank you for the contribution.

PR #29425, opened 2024-03-10
fix(s3-assets): throw if path property is empty

### LINKED ISSUE #29410: (software.amazon.awscdk.services.lambda.Function): Recursive folder creation in cdk.out when lambda jar path is mentioned as ""
### Describe the bug

When creating a lambda function using CDK Java, if the lambda jar file path mentioned as "" in the Code parameter, the synth/diff/deploy process initiates an infinite recursive creation of directories "cdk.out/asset-xxx/cdk.out/asset-xxx/,,,".
It throws an error at the end when the file path is too deep. There is no warning if the asset folder is not mentioned.
Also this leads to a blue screen when the project is opened in Eclipse.

Please fix this issue.

### Expected Behavior

Throw an error that the asset path is not provided.

### Current Behavior

Creation of infinite nested folders.

### Reproduction Steps

   	FunctionProps props = FunctionProps.builder()
	    	.code(Code.fromAsset(""))
	    	.handler("com.example.action.handler.MyAction")
	    	.functionName("MyFunction")
	    	.runtime(Runtime.JAVA_11)
	    	.logRetention(RetentionDays.ONE_WEEK)
	    	.role(f1Role)
	    	.build();
    	
    	Function f1 = new Function(this,"Fun", props);

### Possible Solution

Stop the synthesis and throw an error message that the asset path cannot be a zero-lenth string.

_No response_

### CDK CLI Version

2.102.0

### Framework Version

2.102.0

### Node.js Version

v18.16.0

### OS

Windows 10

### Language

Java

### Language Version

11

### Other information

_No response_

nmussy:
### Issue # (if applicable)

Closes #29410.

### Reason for this change

It was reported that a `Code.fromAsset('')` was creating an infinite loop by including itself through `cdk.out`. This is caused by the following line:

https://github.com/aws/aws-cdk/blob/730fe63efc461c14f6e2b4aa9206c10f9b0f4cd9/packages/aws-cdk-lib/aws-s3-assets/lib/asset.ts#L145

If an empty string is given to `path.resolve()`, the current working directory is returned.

### Description of changes

I've added a check that verifies that the given `path` property is not empty.

I've added a test for both the `aws-lambda` package, where the issue was originally reported, and `aws-s3-assets`, where the fix was implemented

### Checklist

----


packages/aws-cdk-lib/aws-s3-assets/lib/asset.ts
```diff
@@ -137,6 +137,10 @@ export class Asset extends Construct implements cdk.IAsset {
   constructor(scope: Construct, id: string, props: AssetProps) {
     super(scope, id);
 
+    if (!props.path) {
+      throw new Error('Asset path cannot be empty');
+    }
```
nmussy, line 142: It might also be a good idea to check that `cdk.out` is excluded from the asset source, as `Code.fromAsset('.')` would cause the same reported issue.

nmussy:
Exemption Request: This is just an additional property constraint, a unit test should be sufficient

packages/aws-cdk-lib/aws-s3-assets/test/asset.test.ts
```diff
@@ -143,11 +143,18 @@ test('"readers" or "grantRead" can be used to grant read permissions on the asse
...
+    path: '',
+  })).toThrow(/Asset path cannot be empty/);
+});
+
 test('fails if directory not found', () => {
   const stack = new cdk.Stack();
   expect(() => new Asset(stack, 'MyDirectory', {
     path: '/path/not/found/' + Math.random() * 999999,
-  })).toThrow();
+  })).toThrow(/Cannot find asset/);
```
msambol, line 157: this just throws inherently, correct? Possible to catch this in `asset.ts` and add the path: `Cannot find asset at path: <path>` ?

packages/aws-cdk-lib/aws-s3-assets/lib/asset.ts
msambol: +1 on this

packages/aws-cdk-lib/aws-s3-assets/test/asset.test.ts
msambol: similar to https://github.com/aws/aws-cdk/blob/e63c7770ea0a7ec12353973cdf64015f9a4121f1/packages/aws-cdk-lib/core/lib/asset-staging.ts#L174-L176

nmussy: I just wanted to refine the existing test to help differentiate between the two errors, since I was adding a new one.

msambol: I see it now 👍🏼

msambol (approved):
LGTM. Thanks @nmussy

paulhcsun (approved):
Thanks for the contribution @nmussy! Apologies for the delay on reviewing this.

And appreciate the eyes on this @msambol.

paulhcsun:
@Mergifyio update

PR #29444, opened 2024-03-11
chore: add weekly workflow for github issues/pr metrics

paulhcsun:
### Issue # (if applicable)

Closes #<issue number here>.

### Reason for this change

Generate weekly version of the [PR/issues metrics](https://github.com/aws/aws-cdk/issues/28952). 

### Description of changes

Add new workflow that runs weekly on Mondays @ 9 AM to generate a PR/Issue metrics report for the past week (Monday - Sunday). 

### Checklist

----


.github/workflows/repo-metrics-weekly.yml
```diff
@@ -0,0 +1,58 @@
...
+      run: |
+        # Calculate the previous Monday's date
+        PREVIOUS_MONDAY=$(date -d "last monday" "+%Y-%m-%d")
+
+        # Calculate the current Sunday's date
+        CURRENT_SUNDAY=$(date -d "next Sunday" "+%Y-%m-%d")
+
+        # Set an environment variable with the date range
+        echo "$PREVIOUS_MONDAY..$CURRENT_SUNDAY"
+        echo "last_week=$first_day..$last_day" >> "$GITHUB_ENV"
```
godwingrs22, line 30: Seems the variable needs to be updated with previous monday and current sunday

.github/workflows/repo-metrics-weekly.yml
```diff
@@ -0,0 +1,58 @@
...
+
+    steps:
+    - name: Get dates for last week
+      shell: bash
+      run: |
+        # Calculate the previous Monday's date
+        PREVIOUS_MONDAY=$(date -d "last monday" "+%Y-%m-%d")
+
+        # Calculate the current Sunday's date
+        CURRENT_SUNDAY=$(date -d "next Sunday" "+%Y-%m-%d")
```
godwingrs22, line 26: QQ: Just would like to know if "next sunday" provides the current sunday or the upcoming sunday? Since the script runs every monday, does this gives the current sunday? We could also get the current sunday may be using "yesterday"?

godwingrs22 (commented):
Overall looks good. Just few minor comments.

.github/workflows/repo-metrics-weekly.yml
paulhcsun: Good catch!

.github/workflows/repo-metrics-weekly.yml
paulhcsun: Good point as this would be run every monday, I'll update the naming.

.github/workflows/repo-metrics-weekly.yml
```diff
@@ -0,0 +1,58 @@
...
+
+    steps:
+    - name: Get dates for last week
+      shell: bash
+      run: |
+        # Calculate the previous Monday's date
+        PREVIOUS_MONDAY=$(date -d "last monday" "+%Y-%m-%d")
+
+        # Calculate the current Sunday's date
+        CURRENT_SUNDAY=$(date -d "current Sunday" "+%Y-%m-%d")
```
godwingrs22, line 26: Just wondering if this command returns the correct previous monday and current sunday? Because to me seems both identical `+%Y-%m-%d`. Do you have any reference doc or any sample?

godwingrs22: nit: consider renaming `last monday` to `previous monday` or vice versa to be consistent.

paulhcsun: Oh wait you're right that's not actually doing anything right now, let me have a look

paulhcsun: Pushed updated shell script. Here's some docs to reference: https://ioflood.com/blog/bash-date-format/

.github/workflows/repo-metrics-weekly.yml
```diff
@@ -0,0 +1,64 @@
...
+        CURRENT_DAY_OF_WEEK=$(date +%u)
+
+        # Calculate the number of days to subtract to get to the previous Monday
+        DAYS_TO_MONDAY=$((CURRENT_DAY_OF_WEEK - 1))
+
+        # Calculate the date of the previous Monday
+        PREVIOUS_MONDAY=$(date -d "$DAYS_TO_MONDAY days ago" "+%Y-%m-%d")
+
+        # Calculate the date of the current Sunday (assuming today is not Sunday)
+        CURRENT_SUNDAY=$(date -d "$DAYS_TO_MONDAY days ago +6 day" "+%Y-%m-%d")
```
godwingrs22, line 32: I think this may not work since the `DAYS_TO_MONDAY` will return `0` if the script is running on monday. 
Moreover, since we already know the script will be executed only every monday based on the cron job scheduled, we can simplify as below to get last week (last monday to last sunday)
```
LAST_MONDAY=$(date -d "7 days ago" "+%Y-%m-%d")
LAST_SUNDAY=$(date -d "1 day ago" "+%Y-%m-%d")
```

paulhcsun: Updated

.github/workflows/repo-metrics-weekly.yml
```diff
@@ -0,0 +1,58 @@
...
+        assignees: paulhcsun
+        
+    - name: Report on PRs
+      uses: github/issue-metrics@v2
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        SEARCH_QUERY: 'repo:aws/aws-cdk is:pr created:${{ env.last_week }} -is:draft'
+
+    - name: Create report for PRs
+      uses: peter-evans/create-issue-from-file@v5
```
godwingrs22, line 53: Not a blocker. Would like to know is it normal and safe to use third party github actions for aws owned repos? As I can see this action is not from a `Verified creator` label which usually GITHUB does.

peter-evans/create-issue-from-file@v5 -> https://github.com/marketplace/actions/create-issue-from-file

Sample verified action: https://github.com/marketplace/actions/checkout
image

godwingrs22 (approved):
LGTM. Left a comment

PR #29445, opened 2024-03-11
fix(elasticloadbalancingv2): allow alb slow start duration of 0 seconds

### LINKED ISSUE #29437: [aws_elasticloadbalancingv2]: Target Group 'slowStart' option can't be disabled
### Describe the bug

When you set **slowStart** attribute on a Target Group it sets this property correctly in the CFN template:

```
const targetGroup = httpListener.addTargets('myTG', {
      protocol: elbv2.ApplicationProtocol.HTTP,
      slowStart: cdk.Duration.seconds(40),
    });
```

Template:
```
"TargetGroup52DC4313": {
   "Type": "AWS::ElasticLoadBalancingV2::TargetGroup",
   "Properties": {
   .
   .
   .
    "TargetGroupAttributes": [
     {
      "Key": "slow_start.duration_seconds",
      "Value": "40"
     }
```

But when trying to remove this setting by just removing **slowStart** attribute, CDK removes the synthesized attribute from the CFN template but the actual ElbV2 Target Group resource is not changed. 

[In the docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_elasticloadbalancingv2.AddApplicationTargetsProps.html#slowstart) the default value would be set to "0" (disabled) but it doesn't accept anything outside the 30-900 range:

`Error: Slow start duration value must be between 30 and 900 seconds.`

The only way to disable it is by directly changing the Target Group Attributes from AWS Console/CLI.

### Expected Behavior

_Slow Start_ attribute removed from the ElbV2 Target Group by either removing 'slowStart' option from the construct's properties or by setting it to "0" (`slowStart: cdk.Duration.seconds(0)`)

### Current Behavior

Slow Start attribute can not be removed from CDK. Not possible to set it to "0".

### Reproduction Steps

1 - Create ALB, Listener and Target Group (with `slowStart` property set)

```
const vpc = new ec2.Vpc(this, 'myVPC', {
      maxAzs: 2,
    });

    const alb = new elbv2.ApplicationLoadBalancer(this, 'myALB', {
      vpc: vpc,
    });

    const httpListener = alb.addListener('myListener-http', { 
      protocol: elbv2.ApplicationProtocol.HTTP,
    });

    const targetGroup = httpListener.addTargets('myTG', {
      protocol: elbv2.ApplicationProtocol.HTTP,
      slowStart: cdk.Duration.seconds(40),
    });
```

2 - Try removing 'slowStart' property and it deploys successfully, but Target Group resource keeps it enabled. Or try to set it to "0" and CDK does not allow it.

### Possible Solution

Allow "0" in the range for `cdk.Duration.seconds()` values and implicitly synthesize the Target Group resource with this as default when **slowStart** attribute is omited (if previously enabled):

`   "TargetGroupAttributes": [
     {
      "Key": "slow_start.duration_seconds",
      "Value": "0"
     }
`

_No response_

### CDK CLI Version

All supported versions, including latest 2.132.0

### Framework Version

_No response_

### Node.js Version

16

### OS

Mac

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

msambol:
Closes #29437.

----


msambol:
Exemption request: Unit test should be fine here for this fix.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-target-group.ts
```diff
@@ -330,8 +330,9 @@ export class ApplicationTargetGroup extends TargetGroupBase implements IApplicat
 
     if (props) {
       if (props.slowStart !== undefined) {
```
godwingrs22, line 332: Wondering if we need to handle where `slowStart` is not specified/undefined, to explicity set `slow_start.duration_seconds` attribute as default `0`.  since I saw in the issue reported, it was mentioned if it is not specified, it deploys successfully but Target Group resource still has it enabled(if previously enabled).

godwingrs22 (commented):
left a minor comment.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-target-group.ts
msambol: hmm.. I think if they want it disabled they should have to set it explicitly ?  My concern is if we set to `0`, it will cause changes with stacks that are already deployed.

godwingrs22: Yeah seems valid. This looks good to me.

msambol: awesome, thanks!

godwingrs22 (approved):

msambol:
@godwingrs22 Do you mind merging `main` into this?

PR #29452, opened 2024-03-11
fix(iam): grantAssumeRole silently fails with service and account principals

kishiel:
### Reason for this change

grantAssumeRole silently fails if a Service Principal or Account Principal is used which led me to a false assumption about the correctness of a role's permission scope

### Description of changes

This change will throw an error if a Service Principal is used.  I was unable to find a way to accomplish the same behavior for Account Principals.

Documentation was updated to help guide a user to the appropriate function usage for Service and Account Principals.

* Added a unit test
* This change required me to re-run two unrelated snapshot tests which were throwing errors outside of the scope of this change.

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-cloudwatch-actions/test/integ.lambda-alarm-action.js.snapshot/LambdaAlarmActionIntegrationTestStack.assets.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-cloudwatch-actions/test/integ.lambda-alarm-action.js.snapshot/LambdaAlarmActionIntegrationTestStack.assets.json
@@ -1,7 +1,7 @@
 {
   "version": "36.0.0",
   "files": {
-    "443cd498fca620868cfc5e66c6d03b3dcd2cc9d88b6eba24927a4e8d06ae7984": {
+    "db34b72b17af5b7d83e311e6552b5f6cc3c82caf16943655fcbabc28a5c9b826": {
       "source": {
         "path": "LambdaAlarmActionIntegrationTestStack.template.json",
         "packaging": "file"
@@ -9,7 +9,7 @@
       "destinations": {
         "current_account-current_region": {
           "bucketName": "cdk-hnb659fds-assets-${AWS::AccountId}-${AWS::Region}",
-          "objectKey": "443cd498fca620868cfc5e66c6d03b3dcd2cc9d88b6eba24927a4e8d06ae7984.json",
+          "objectKey": "db34b72b17af5b7d83e311e6552b5f6cc3c82caf16943655fcbabc28a5c9b826.json",
           "assumeRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-file-publishing-role-${AWS::AccountId}-${AWS::Region}"
         }
       }
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-cloudwatch-actions/test/integ.lambda-alarm-action.js.snapshot/LambdaAlarmActionIntegrationTestStack.template.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-cloudwatch-actions/test/integ.lambda-alarm-action.js.snapshot/LambdaAlarmActionIntegrationTestStack.template.json
@@ -79,8 +79,36 @@
      }
     ],
     "EvaluationPeriods": 1,
+    "InsufficientDataActions": [
+     {
+      "Ref": "alarmLambdaFeatureCurrentVersionCF39751979501d2f67eaf906b2ef0c378303873b"
+     },
+     {
+      "Ref": "alarmLambdaFeatureAliasaliasName16F91D34"
+     },
+     {
+      "Fn::GetAtt": [
+       "alarmLambdaFeatureD560800F",
+       "Arn"
+      ]
+     }
+    ],
     "MetricName": "Errors",
     "Namespace": "AWS/Lambda",
+    "OKActions": [
+     {
+      "Ref": "alarmLambdaFeatureCurrentVersionCF39751979501d2f67eaf906b2ef0c378303873b"
+     },
+     {
+      "Ref": "alarmLambdaFeatureAliasaliasName16F91D34"
+     },
+     {
+      "Fn::GetAtt": [
+       "alarmLambdaFeatureD560800F",
+       "Arn"
+      ]
+     }
+    ],
     "Period": 60,
     "Statistic": "Sum",
     "Threshold": 1,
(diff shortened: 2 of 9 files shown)


packages/aws-cdk-lib/aws-iam/README.md
```diff
@@ -369,6 +369,37 @@ new iam.Role(this, 'Role', {
 });
 ```
 
+### Granting assume role permission from a role
+
+Principals can be granted permission to assume a role using `grantAssumeRole`.  Note that this does not apply to Service Principals or Account Principals as they must be added to the role Trust Policy.
```
msambol, line 374: Nit: lowercase `service principals` [1]
Nit: lowercase `account principals [2]

Also move the note to a newline. 

[1] https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html#principal-services
[2] https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html#principal-accounts

packages/aws-cdk-lib/aws-iam/README.md
```diff
@@ -369,6 +369,37 @@ new iam.Role(this, 'Role', {
 });
 ```
 
+### Granting assume role permission from a role
```
msambol, line 372: "From a role" is confusing here. Can you change this to "Granting principals permission to assume a role"

packages/aws-cdk-lib/aws-iam/README.md
```diff
@@ -369,6 +369,37 @@ new iam.Role(this, 'Role', {
...
+```ts
+const user = new iam.User(this, 'user')
+const role = new iam.Role(this, 'role', {
+  assumedBy: new iam.AccountPrincipal(this.account)
+});
+
+role.grantAssumeRole(user);
+```
+
+### Giving Service Principals and Account Principals assume role permission from a role
```
msambol, line 385: "Granting service and account principals permission to assume a role"

packages/aws-cdk-lib/aws-iam/README.md
```diff
@@ -369,6 +369,37 @@ new iam.Role(this, 'Role', {
...
+
+Service Principals and Account Principals can be granted permission to assume a role using `assumeRolePolicy` which modifies the role Trust Policy.
```
msambol, line 387: lowercase

msambol: `Trust Policy` should be lowercase too.

packages/aws-cdk-lib/aws-iam/lib/role.ts
```diff
@@ -594,6 +594,10 @@ export class Role extends Resource implements IRole {
    * Grant permissions to the given principal to assume this role.
    */
   public grantAssumeRole(identity: IPrincipal) {
+    // Service Principals must use assumeRolePolicy
+    if (identity.policyFragment.principalJson.Service) {
+      throw new Error('Cannot use a Service Principal with grantAssumeRole, use assumeRolePolicy instead.');
+    }
```
msambol, line 600: I think this would work?
```
    if (identity instanceof ServicePrincipal || identity instanceof AccountPrincipal) {
      throw new Error('Cannot use a service or account principal with grantAssumeRole, use assumeRolePolicy instead.');
    }
```

msambol:
Can you change the title? It should reflect the bug. Maybe:

`fix(iam): grantAssumeRole silently fails with service and account principals`

msambol (changes requested):
See inline. Great start!

packages/aws-cdk-lib/aws-iam/lib/role.ts
kishiel: Good call, this appears to have worked.  Added another test for account principal.

kishiel:
Incorporated all feedback and updated title

packages/aws-cdk-lib/aws-iam/README.md
```diff
@@ -369,6 +369,39 @@ new iam.Role(this, 'Role', {
 });
 ```
 
+### Granting an identity permission to assume a role
```
msambol, line 372: Can you change this to "Granting a principal...". That's the more common verbiage.

packages/aws-cdk-lib/aws-iam/README.md
```diff
@@ -369,6 +369,39 @@ new iam.Role(this, 'Role', {
...
+
+An identity can be granted permission to assume a role using `grantAssumeRole`.
```
msambol, line 374: Likewise here, "a principal."

msambol (commented):
Few more comments then we're good to go.

packages/aws-cdk-lib/aws-iam/README.md
kishiel: Ah, I forgot I'd modified this.  I changed this to identity based on [this section](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-iam/README.md#principals-vs-identities) of the readme but I think I'm just misreading it.  Will change

msambol (approved):
Looks good. Thanks!

packages/@aws-cdk-testing/framework-integ/test/aws-iam/test/integ.managed-policy.ts
```diff
@@ -24,7 +24,8 @@ const policy3 = ManagedPolicy.fromAwsManagedPolicyName('SecurityAudit');
 user.addManagedPolicy(policy3);
 
 const role = new Role(stack, 'Role', { assumedBy: new AccountRootPrincipal() });
-role.grantAssumeRole(policy.grantPrincipal);
+role.grantAssumeRole(user);
```
GavinZZ, line 27: Why is this change needed in the test file?

packages/aws-cdk-lib/aws-iam/README.md
```diff
@@ -369,6 +369,39 @@ new iam.Role(this, 'Role', {
 });
 ```
 
+### Granting a principal permission to assume a role
```
GavinZZ, line 372: Thanks for this clear documentation update.

GavinZZ (commented):
Generally LGTM, just a question for clarification before approving.

packages/@aws-cdk-testing/framework-integ/test/aws-iam/test/integ.managed-policy.ts
msambol: I think to prove you can still use `grantAssumeRole` on an IAM user.

kishiel: This was a preexisting fault that bubbled up when I made this change.  The error was something to the effect that you cannot grant a policy assume permission.  Since the apparent intent of this portion was to prove that grantAssume could be given to the policy's principal (the user) I changed it to that.

kishiel: I'm taking another look at this to see if I can reproduce it

kishiel: I can't get this to reproduce.  I've reverted the change and re-run the snapshot.  Strange.  Pushing an update.

GavinZZ (dismissed):

GavinZZ (approved):

PR #29454, opened 2024-03-12
feat(eks): add atomic flag for aws-eks Helm Chart

### LINKED ISSUE #22254: aws-eks: support `--atomic` flag for helm commands

shikha372:
### Issue # (if applicable)

Closes #22254.

### Reason for this change

Currently, if chart is installed to the EKS cluster with wait timeout period set and fails initialization, helm will fail to send a response back to custom resource and it will be stuck in pending upgrade state (expected state is failed) . 
Subsequent attempts to update the stack will result in failure while chart is stuck in `pending upgrade` state until manually rolled back or deleted from the cluster. 

### Description of changes

Added feature flag `--atomic` supported by helm currently to mark the operation as atomic which will automatically rolls back the changes in case of upgrade/installation failure.
Reference doc: https://helm.sh/docs/helm/helm_install/#options

Added unit tests to check if flag is set as per user input in the template.

### Checklist

----


packages/aws-cdk-lib/aws-eks/lib/helm-chart.ts
```diff
@@ -73,6 +74,13 @@ export interface HelmChartOptions {
    */
   readonly timeout?: Duration;
 
+  /**
+   * Whether or not Helm should treat this operation as atomic; if set, upgrade process rolls back changes
+   * made in case of failed upgrade. The --wait flag will be set automatically if --atomic is used.
+   * @default false
+   */
+  readonly atomic?: boolean;
```
godwingrs22, line 82: Let us also update the prop in [README.md](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-eks/README.md#helm-charts) helm-charts section

godwingrs22 (changes requested):

packages/aws-cdk-lib/aws-eks/lib/helm-chart.ts
shikha372: don't think its needed since there is no dedicated section in Readme that points to what all flags are supported in helm, its generally described as
`Helm charts will be installed and updated using helm upgrade --install, where a few parameters are being passed down (such as repo, values, version, namespace, wait, timeout, etc).` which covers this case

packages/aws-cdk-lib/aws-eks/lib/helm-chart.ts
```diff
@@ -168,6 +179,7 @@ export class HelmChart extends Construct {
         Repository: this.repository,
         CreateNamespace: createNamespace || undefined,
         SkipCrds: skipCrds || undefined,
+        Atomic: atomic || undefined, // props are stringified so we encode “false” as undefined
```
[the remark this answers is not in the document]
shikha372: Good catch.. updated in latest commit

packages/aws-cdk-lib/aws-eks/test/helm-chart.test.ts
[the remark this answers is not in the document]
shikha372: I see no specific integration tests to check for helm flag values except for `--skip-crds` , moreover this flag useful in the cases of failed helm chart installation. But as a good practice, updating existing integ test to verify that helm chart installation is successful with this flag enabled.

packages/@aws-cdk/custom-resource-handlers/lib/aws-eks/kubectl-handler/helm/__init__.py
```diff
@@ -157,7 +158,7 @@ def get_chart_from_oci(tmpdir, repository = None, version = None):
     raise Exception(f'Operation failed after {maxAttempts} attempts: {output}')
 
 
-def helm(verb, release, chart = None, repo = None, file = None, namespace = None, version = None, wait = False, timeout = None, create_namespace = None, skip_crds = False):
+def helm(verb, release, chart = None, repo = None, file = None, namespace = None, version = None, wait = False, timeout = None, create_namespace = None, skip_crds = False, atomic = False):
```
godwingrs22, line 161: Thanks for handling it in the helm definition. However, seems like the property is not passed as parameter to the calling helm method from the handler [here](https://github.com/aws/aws-cdk/blob/211c98e3a5ddc06f15643979097cfed1a5c02350/packages/%40aws-cdk/custom-resource-handlers/lib/aws-eks/kubectl-handler/helm/__init__.py#L94).

godwingrs22 (changes requested):
One small comment. Other than that LGTM.

shikha372:
> One small comment. Other than that LGTM.

Thank you @godwingrs22  for the review, addressed in latest rev for your final review

packages/@aws-cdk/custom-resource-handlers/lib/aws-eks/kubectl-handler/helm/__init__.py
```diff
@@ -90,7 +91,7 @@ def helm_handler(event, context):
             chart_dir = get_chart_from_oci(tmpdir.name, repository, version)
             chart = chart_dir
 
-        helm('upgrade', release, chart, repository, values_file, namespace, version, wait, timeout, create_namespace)
+        helm('upgrade', release, chart, repository, values_file, namespace, version, wait, timeout, create_namespace, atomic)
```
godwingrs22, line 94: Actually this will be passed to `helm` definition as `skip_crds` as that is the next parameter. I think
since `atomic` is a 12th parameter in the helm defintion, this needs to come after `skip_crds` is passed or you can pass the parameter explicitly like below

```helm('upgrade', release, chart, repository, values_file, namespace, version, wait, timeout, create_namespace, atomic=atomic)```

But not sure why the existing `skip_crds` is not passed to the defintion. Current code looks like it always be false and doesn't get from the resouce props. But this is different issue and just want to highlight it.

godwingrs22 (changes requested):

packages/@aws-cdk/custom-resource-handlers/lib/aws-eks/kubectl-handler/helm/__init__.py
shikha372: yes it is interesting since both integration test and unit tests are passing for this too, imo it could be related to that these two flags don't need any additional handling in regards to values and will be just appended to command if set to true, somehow if this is propogating from props.atomic to underlying function, but yes this is weird that this is working for skip_crds and also have a successful tests passed

godwingrs22: Yeah, seems like the unit test for `skip_crds` to [disable by default](https://github.com/aws/aws-cdk/blob/eb82bc51b7b9c895a356575cbfeaa0aca42154cc/packages/aws-cdk-lib/aws-eks/test/helm-chart.test.ts#L262) is not correct I feel. Instead it needs to find the resource with `{ SkipCrds: true }` which will be expected to 0 like how you have handled for `atomic` or `wait` property.

godwingrs22 (approved):
LGTM. Thanks @shikha372 for your contribution.

godwingrs22:
@Mergifyio update

godwingrs22:
@/Mergifyio update

godwingrs22:
@Mergifyio update

PR #29476, opened 2024-03-14
feat(pipes-sources): add Kinesis and DynamoDB

### LINKED ISSUE #29378: (pipes-sources): Add Kinesis stream source for EventBridge pipes
### Describe the feature

Add a implementation for the Kinesis stream source for EventBridge pipes in the `pipes-sources` module.

### Use Case

For feature completeness of the L2 pipes module this source is required.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

-

-

msambol:
Closes #29378, #29377.

----

### DIFF
--- a/packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
+++ b/packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
@@ -233,6 +233,12 @@ export class Pipe extends PipeBase {
      */
     const source = props.source.bind(this);
     props.source.grantRead(this.pipeRole);
+
+    /**
+     * An optional dead-letter queue stores any events that are not successfully delivered to
+     * a target after all retry attempts are exhausted. The IAM role needs permission to write
+     * events to the dead-letter queue, either an SQS queue or SNS topic.
+     */
     if (SourceWithDeadLetterTarget.isSourceWithDeadLetterTarget(props.source)) {
       props.source.grantPush(this.pipeRole, props.source.deadLetterTarget);
     }
--- a/packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts
+++ b/packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts
@@ -17,7 +17,11 @@ export interface StreamSourceParameters {
   readonly batchSize?: number;
 
   /**
-    * Define the target queue to send dead-letter queue events to.
+    * Define the target to send dead-letter queue events to.
+    *
+    * The dead-letter queue stores any events that are not successfully delivered to a Pipes target after all retry attempts are exhausted.
+    * You can then resolve the issue that caused the failed invocations and replay the events at a later time.
+    * In some cases, such as when access is denied to a resource, events are sent directly to the dead-letter queue and are not retried.
     *
     * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcekinesisstreamparameters.html#cfn-pipes-pipe-pipesourcekinesisstreamparameters-deadletterconfig
     * @default - no dead-letter queue or topic


packages/@aws-cdk/aws-pipes-sources-alpha/lib/deadLetterConfig.ts
```diff
@@ -0,0 +1,14 @@
...
+  /**
+   * The ARN of the specified target for the dead-letter queue,
+   *
+   * For Amazon Kinesis stream and Amazon DynamoDB stream sources, specify either an Amazon SNS topic or Amazon SQS queue ARN.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-deadletterconfig.html#cfn-pipes-pipe-deadletterconfig-arn
+   * @default undefined
+   */
+  readonly arn?: string;
+}
```
msambol, line 14: this will be used by DynamoDB too, hence adding it in a separate file.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
```diff
@@ -0,0 +1,146 @@
...
+    return {
+      sourceParameters: {
+        kinesisStreamParameters: {
+          batchSize: this.sourceParameters.batchSize,
+          deadLetterConfig: this.sourceParameters.deadLetterConfig,
+          maximumBatchingWindowInSeconds: this.sourceParameters.maximumBatchingWindow?.toSeconds(),
+          maximumRecordAgeInSeconds: this.sourceParameters.maximumRecordAge?.toSeconds() ?? -1,
+          maximumRetryAttempts: this.sourceParameters.maximumRetryAttempts ?? -1,
+          onPartialBatchItemFailure: this.sourceParameters.onPartialBatchItemFailure,
+          parallelizationFactor: this.sourceParameters.parallelizationFactor ?? 1,
```
msambol, line 135: do we prefer to set defaults like this? or let CFN do it?

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
```diff
@@ -0,0 +1,183 @@
...
+   */
+  readonly maximumBatchingWindow?: Duration;
+
+  /**
+   * (Streams only) Discard records older than the specified age. The default value is -1, which sets the maximum age to infinite. When the value is set to infinite, EventBridge never discards old records.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcekinesisstreamparameters.html#cfn-pipes-pipe-pipesourcekinesisstreamparameters-maximumrecordageinseconds
+   * @default -1 (infinite)
+   */
+  readonly maximumRecordAgeInSeconds?: number;
```
msambol, line 41: Using `number` here instead of `Duration` because -1 is an allowed value.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
```diff
@@ -0,0 +1,186 @@
...
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcekinesisstreamparameters.html#cfn-pipes-pipe-pipesourcekinesisstreamparameters-maximumbatchingwindowinseconds
+   * @default no batching window
+   */
+  readonly maximumBatchingWindow?: Duration;
+
+  /**
+   * (Streams only) Discard records older than the specified age. The default value is -1, which sets the maximum age to infinite. When the value is set to infinite, EventBridge never discards old records.
+   *
+   * Leave undefined to set the maximum record age to infinite.
```
msambol, line 38: The linter wants this a `Duration`, not a `number`, which makes sense. However, `-1` is an allowed value, which doesn't work with `Duration`. I added this note but open to ideas if there is a better way.

packages/@aws-cdk/aws-pipes-sources-alpha/README.md
```diff
@@ -61,3 +61,22 @@ const pipe = new pipes.Pipe(this, 'Pipe', {
 });
 ```
 
+### Amazon Kinesis
+
+A Kinesis stream can be used as a source for a pipe. The stream will be polled for new messages and the messages will be sent to the pipe.
+
+```ts
+import * as kinesis from 'aws-cdk-lib/aws-kinesis';
```
msambol, line 69: I'm not sure why this is required but an import for SQS isn't needed ?

RaphaelManke: They are declared in the default rosetta file 
https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-pipes-sources-alpha/rosetta/default.ts-fixture

packages/@aws-cdk/aws-pipes-sources-alpha/test/integ.dynamodb.ts
```diff
@@ -0,0 +1,69 @@
...
+});
+
+const message = test.assertions.awsApiCall('SQS', 'receiveMessage', {
+  QueueUrl: targetQueue.queueUrl,
+});
+
+message.assertAtPath('Messages.0.Body.dynamodb.Keys.id.S', ExpectedResult.stringLikeRegexp('"1"')).waitForAssertions({
+  totalTimeout: cdk.Duration.minutes(2),
+  interval: cdk.Duration.seconds(15),
+});
```
msambol, line 67: Screenshot 2024-03-15 at 2 21 49 PM

This is failing with this error ^ Still trying to debug it.

msambol:
@nmussy Feel like reviewing a big one? 😄

nmussy:
I'll have a look on Monday 👍

packages/@aws-cdk/aws-pipes-sources-alpha/lib/deadLetterConfig.ts
```diff
@@ -0,0 +1,14 @@
...
+ * A DeadLetterConfig object that contains information about a dead-letter queue configuration.
+ */
+export interface DeadLetterConfigParameters {
+  /**
+   * The ARN of the specified target for the dead-letter queue,
+   *
+   * For Amazon Kinesis stream and Amazon DynamoDB stream sources, specify either an Amazon SNS topic or Amazon SQS queue ARN.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-deadletterconfig.html#cfn-pipes-pipe-deadletterconfig-arn
+   * @default no dead letter queue
```
nmussy, line 11:
```suggestion
   * @default - no dead letter queue
```

Otherwise it's interpreted as `no` being the default value

I won't comment on all of them but there are quite a few after this

packages/@aws-cdk/aws-pipes-sources-alpha/lib/deadLetterConfig.ts
```diff
@@ -0,0 +1,14 @@
...
+   */
+  readonly arn?: string;
```
nmussy, line 13: It might be more explicit to have construct interfaces instead of the `arn` there, especially given it accepts both SQS and SNS? From a user perspective, they can just either use the queue construct they just created, or retrieve it with `.fromQueue/TopicArn()`:

```suggestion
  readonly target?: IQueue | ITopic;
```

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,158 @@
...
+  /**
+    * Define the target queue to send dead-letter queue events to.
+    *
+    * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-deadletterconfig
+    * @default no dead letter queue
+    */
+  readonly deadLetterConfig?: DeadLetterConfigParameters;
+
+  /**
+   * The maximum length of a time to wait for events.
```
nmussy, line 29: Could you add the min/max values there?

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,158 @@
...
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-maximumbatchingwindowinseconds
+   * @default no batching window
+   */
+  readonly maximumBatchingWindow?: Duration;
+
+  /**
+   * (Streams only) Discard records older than the specified age. The default value is -1, which sets the maximum age to infinite. When the value is set to infinite, EventBridge never discards old records.
```
nmussy, line 37: Min/max values there as well, I'll stop tagging them from then on

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * (Streams only) Discard records after the specified number of retries. The default value is -1, which sets the maximum number of retries to infinite. When MaximumRetryAttempts is infinite, EventBridge retries failed records until the record expires in the event source.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-maximumretryattempts
+   * @default -1 (infinite)
+   */
+  readonly maximumRetryAttempts?: number;
+
+  /**
+   * (Streams only) Define how to handle item process failures. AUTOMATIC_BISECT halves each batch and retry each half until all the records are processed or there is one failed message left in the batch.
```
nmussy, line 55:
```suggestion
   * (Streams only) Define how to handle item process failures. {@link OnPartialBatchItemFailure.AUTOMATIC_BISECT} halves each batch and retry each half until all the records are processed or there is one failed message left in the batch.
```

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  /**
+   * (Streams only) Discard records older than the specified age. The default value is -1, which sets the maximum age to infinite. When the value is set to infinite, EventBridge never discards old records.
+   *
+   * Leave undefined to set the maximum record age to infinite.
+   *
+   *  @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-maximumrecordageinseconds
+   * @default -1 (infinite)
+   */
+  readonly maximumRecordAge?: Duration;
```
nmussy, line 44: `@default undefined` here as well, see below

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
nmussy: I can't find a better solution than:

`@default undefined - If <field> is not set, the duration will be infinite`

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,158 @@
...
+      if (this.maximumBatchingWindowInSeconds > 300) {
+        throw new Error(`Maximum batching window must be between 0 and 300, received ${this.maximumBatchingWindowInSeconds}`);
+      }
+    }
+    if (this.maximumRecordAgeInSeconds !== undefined) {
+      // only need to check upper bound since Duration amounts cannot be negative
+      if (this.maximumRecordAgeInSeconds > 604800) {
+        throw new Error(`Maximum record age in seconds must be between -1 and 604800, received ${this.maximumRecordAgeInSeconds}`);
+      }
+    }
```
nmussy, line 124: Is `Duration.seconds(0)` a valid value? Might need to check with an integ

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * (Streams only) Define how to handle item process failures. AUTOMATIC_BISECT halves each batch and retry each half until all the records are processed or there is one failed message left in the batch.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcekinesisstreamparameters.html#cfn-pipes-pipe-pipesourcekinesisstreamparameters-onpartialbatchitemfailure
+   * @default off
+   */
+  readonly onPartialBatchItemFailure?: OnPartialBatchItemFailure;
+
+  /**
+   * (Streams only) The number of batches to process concurrently from each shard. The default value is 1.
```
nmussy, line 63:
```suggestion
   * (Streams only) The number of batches to process concurrently from each shard.
```

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
```diff
@@ -0,0 +1,161 @@
...
+   */
+  readonly startingPosition: KinesisStartingPosition;
+
+  /**
+   * With StartingPosition set to AT_TIMESTAMP, the time from which to start reading, in Unix time seconds.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcekinesisstreamparameters.html#cfn-pipes-pipe-pipesourcekinesisstreamparameters-startingpositiontimestamp
+   * @default no starting position timestamp
+   */
+  readonly startingPositionTimestamp?: string;
```
nmussy, line 83: Pretty sure this can be a number. I'd also add `{@link}`s to refer to the other field/value

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
```diff
@@ -0,0 +1,161 @@
...
+      if (this.maximumBatchingWindowInSeconds > 300) {
+        throw new Error(`Maximum batching window must be between 0 and 300, received ${this.maximumBatchingWindowInSeconds}`);
+      }
+    }
+    if (this.maximumRecordAgeInSeconds !== undefined) {
+      // only need to check upper bound since Duration amounts cannot be negative
+      if (this.maximumRecordAgeInSeconds > 604800) {
+        throw new Error(`Maximum record age in seconds must be between -1 and 604800, received ${this.maximumRecordAgeInSeconds}`);
+      }
+    }
```
nmussy, line 127: Same here, is 0 seconds valid?

nmussy (changes requested):
Still need to take a look at the tests, but I already went through the code and docs 👍

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
msambol: I had `undefined` originally but the linter didn't like it. It wants the value CFN will assign, even if we don't assign -1 below.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
msambol: Good shout. Doesn't say it in the docs but needs to be min 60 if defined.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/deadLetterConfig.ts
msambol: @nmussy I made this change, but now there is a dependency between the DLQ and the pipe (see below). I'm not quite sure where to add the dependency. Maybe in `aws-pipes-alpha/pipes.ts` ? Let me know if you have any ideas. 

```
Deployment failed: Error: The stack named aws-cdk-pipes-sources-dynamodb-stream failed creation, it may need to be manually deleted from the AWS console: ROLLBACK_COMPLETE: Resource handler returned message: "Resource of type 'AWS::Pipes::Pipe' with identifier 'Pipe7793F8A1-B5m3oRxh5ZKt' did not stabilize. Status Reason is Input parameter is invalid from the request due to : Error occurred while sending message to SQS queue: arn:aws:sqs:us-east-1:REDACED:aws-cdk-pipes-sources-dynamodb-stream-DlqQueueFDA42DA7-LKdmBtvZnzrn" (RequestToken: 63c085a3-c417-a234-e570-343c179f1333, HandlerErrorCode: NotStabilized)
```

msambol: hmm, something is wrong with my code. The permissions aren't being added to the role. Calling it for the night but I'll try to figure this out.

nmussy: If it's an issue caused by interface constructs or the union type, a good compromise might be to go back to an interface and separate the ARN props:
```ts
export interface DeadLetterConfig {
  /**
   * The ARN of the target Amazon SQS queue.
   * 
   * At least one of {@link queueArn} and {@link topicArn} must be provided.
   * 
   * @default - not an SQS dead letter queue
   */
  readonly queueArn?: string;

  /**
   * The ARN of the target Amazon SNS topic.
   * 
   * At least one of {@link queueArn} and {@link topicArn} must be provided.
   * 
   * @default - not an SNS dead letter queue
   */
  readonly topicArn?: string;
}
```

msambol: What are your thoughts on granting permissions to the queue/topic outside of this construct? i.e., specify in the README and the integration test the role must have permission to push to them.

Here's what I have now... It's not working but I'll keep investigating.
https://github.com/aws/aws-cdk/pull/29476/files#diff-4bb7f8b01c48466253fab15c1b3492a1e240a2fc1a272b48a90c01534636431eR180-R184

nmussy: > What are your thoughts on granting permissions to the queue/topic outside of this construct? i.e., specify in the README and the integration test the role must have permission to push to them.

That would be less than ideal, I would expect an L2 construct to handle that for me

msambol: yeah... let me figure out why these permissions aren't being applied.

nmussy: Let me know if you want me to take a look 👍

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
```diff
@@ -99,4 +99,9 @@ export interface ISource {
    * Grant the pipe role read access to the source.
    */
   grantRead(grantee: IRole): void;
+
+  /**
+   * Grant the pipe role write access to the dead-letter target.
+   */
+  grantDlqPush?(grantee: IRole): void;
```
msambol, line 106: @nmussy Making this optional solved a whole lot of problems.

packages/@aws-cdk/aws-pipes-sources-alpha/test/integ.kinesis.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const sourceUnderTest = new KinesisSource(sourceKinesisStream, {
+  batchSize: 1,
+  maximumBatchingWindow: cdk.Duration.seconds(20),
+  maximumRecordAge: cdk.Duration.seconds(60),
+  maximumRetryAttempts: 3,
+  onPartialBatchItemFailure: OnPartialBatchItemFailure.AUTOMATIC_BISECT,
+  parallelizationFactor: 1,
+  startingPosition: KinesisStartingPosition.LATEST,
+});
```
msambol, line 39: @nmussy I couldn't get this to deploy with a dead-letter target either. But given this is `alpha`, and you can't even specify dead-letter in the Console, I think we're good here. I did add unit tests to make sure the permissions get added.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/deadLetterConfig.ts
msambol: I had forgotten to build `aws-pipes-alpha/lib/source.ts` 🤦🏼‍♂️

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
msambol: CDK seems to go back and forth on this, depending what CFN expects..

nmussy: Up to you, but I think you can tell that a `number` timestamp field is an epoch without looking at the JSDoc. A `string` timestamp, I wouldn't be sure if it's an epoch or an ISO 8601 date/datetime at first glance

packages/@aws-cdk/aws-pipes-sources-alpha/test/integ.kinesis.ts
nmussy: Might be a good idea to report the issue and not to expose the Kinesis DLQ parameter to the user in the meantime then

msambol: @pahud Is this something you can raise with the CFN/Pipes team? When you create a Kinesis source with a dead-letter config, it doesn't appear to stabilize in time.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,187 @@
...
+   * @default 1
+   */
+  readonly parallelizationFactor?: number;
+
+  /**
+   * (Streams only) The position in a stream from which to start reading.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-startingposition
+   */
+  readonly startingPosition: DynamoDBStartingPosition;
```
acjay, line 89: Every other option has a default. Might it make sense to give this one a default behavior?

msambol: This one is required by CFN and has no default

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
```diff
@@ -0,0 +1,198 @@
...
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcekinesisstreamparameters.html#cfn-pipes-pipe-pipesourcekinesisstreamparameters-startingposition
+   */
+  readonly startingPosition: KinesisStartingPosition;
+
+  /**
+   * With StartingPosition set to AT_TIMESTAMP, the time from which to start reading, in Unix time seconds.
+   *
+   * @example
+   * '2025-01-01T00:00:00Z'
```
msambol, line 92: @nmussy decided to keep this a string but I added an example here.

packages/@aws-cdk/aws-pipes-sources-alpha/test/integ.kinesis.ts
```diff
@@ -0,0 +1,70 @@
...
+
+const sourceUnderTest = new KinesisSource(sourceKinesisStream, {
+  batchSize: 1,
+  maximumBatchingWindow: cdk.Duration.seconds(20),
+  maximumRecordAge: cdk.Duration.seconds(60),
+  maximumRetryAttempts: 3,
+  onPartialBatchItemFailure: OnPartialBatchItemFailure.AUTOMATIC_BISECT,
+  parallelizationFactor: 1,
+  startingPosition: KinesisStartingPosition.AT_TIMESTAMP,
+  startingPositionTimestamp: '2024-01-01T00:00:00Z',
```
msambol, line 39: @nmussy I made sure this is covered by an integration test.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
nmussy: So just to be clear, `StartingPositionTimestamp` supports both epochs and ISO strings? We should probably amend the documentation above as well, since in only mentions "in Unix time seconds"

msambol: This is the error if I try epoch:

```[#/SourceParameters/KinesisStreamParameters/StartingPositionTimestamp: [1711726773] is not a valid date-time. Expected [yyyy-MM-dd'T'HH:mm:ssZ, yyyy-MM-dd'T'HH:mm:ss.[0-9]{1,9}Z, yyyy-MM-dd'T'HH:mm:ss[+-]HH:mm, yyyy-MM-dd'T'HH:mm:ss.[0-9]{1,9}[+-]HH:mm]]```

I'm going to open a bug on their documentation.

Comment (msambol) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts:92]: Unless they are considering that "Unix time seconds."

Comment (nmussy) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts:92]: I wouldn't think so, it's probably a documentation error yeah. Could you change the JSDoc above then as well?

Comment (msambol) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts:92]: @nmussy I changed this to `in ISO 8601 format`. Submitting AWS doc tickets too.

**Comment by nmussy** [comment]
I'll give it another overall look tomorrow, I should be able to give it community approval  now

**Comment by msambol** [comment]
> I'll give it another overall look tomorrow, I should be able to give it community approval now

Nice! And thank you. Will be good to get this in, there are ~10 other sources/targets I want to add.

File: packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts (hunk 19 above)
Comment (acjay) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts:89]: If the idea is to match CFN's behavior, that makes sense to me 👍

File: packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts [hunk 22]
```diff
@@ -0,0 +1,187 @@
...
+    * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-batchsize
+    * @default 1
+    */
+  readonly batchSize?: number;
+
+  /**
+    * Define the target SQS queue or SNS topic to send dead-letter queue events to.
+    *
+    * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-deadletterconfig
+    * @default - no dead letter queue
```
nmussy, line 28:
```suggestion
    * @default - no dead letter queue or topic
```

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,187 @@
...
+  readonly deadLetterTarget?: IQueue | ITopic;
+
+  /**
+   * The maximum length of a time to wait for events.
+   *
+   * Minumum: Duration.seconds(0)
+   * Maxiumum: Duration.seconds(300)
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-maximumbatchingwindowinseconds
+   * @default - no batching window
```
nmussy, line 39: It might be better to make this more explicit, something like "the events will be treated immediately"

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,187 @@
...
+  readonly maximumBatchingWindow?: Duration;
+
+  /**
+   * (Streams only) Discard records older than the specified age. The default value is -1, which sets the maximum age to infinite. When the value is set to infinite, EventBridge never discards old records.
+   *
+   * Minumum: Duration.seconds(60) (leave undefined to set the maximum age to -1)
+   * Maxiumum: Duration.seconds(604800)
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-maximumrecordageinseconds
+   * @default -1 - maximum age is infinite
```
nmussy, line 50: Same here, saying explicitly what this value does in the `@default` is probably a good idea, even if it's described above

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,187 @@
...
+  readonly maximumRecordAge?: Duration;
+
+  /**
+   * (Streams only) Discard records after the specified number of retries. The default value is -1, which sets the maximum number of retries to infinite. When MaximumRetryAttempts is infinite, EventBridge retries failed records until the record expires in the event source.
+   *
+   * Minumum: -1
+   * Maxiumum: 10000
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-maximumretryattempts
+   * @default -1 (infinite)
```
nmussy, line 61: Here as well

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,187 @@
...
+   */
+  readonly maximumRetryAttempts?: number;
+
+  /**
+   * (Streams only) Define how to handle item process failures. {@link OnPartialBatchItemFailure.AUTOMATIC_BISECT} halves each batch and retry each half until all the records are processed or there is one failed message left in the batch.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcedynamodbstreamparameters.html#cfn-pipes-pipe-pipesourcedynamodbstreamparameters-onpartialbatchitemfailure
+   * @default off
```
nmussy, line 69: Again, unclear on what happens when there is no behavior set

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -0,0 +1,187 @@
...
+  readonly sourceArn;
+  private sourceParameters;
+
+  private batchSize;
+  private maximumBatchingWindowInSeconds;
+  private maximumRecordAgeInSeconds;
+  private maximumRetryAttempts;
+  private parallelizationFactor;
+  private deadLetterTarget;
+  private deadLetterTargetArn;
```
nmussy, line 106: It would be ideal to type all of these, both for safety and readability

packages/@aws-cdk/aws-pipes-sources-alpha/lib/enums.ts
```diff
@@ -0,0 +1,41 @@
...
+export enum DynamoDBStartingPosition {
+  /**
+   * TRIM_HORIZON
+   */
+  TRIM_HORIZON = 'TRIM_HORIZON',
+  /**
+   * LATEST
+   */
+  LATEST = 'LATEST',
+}
```
nmussy, line 41: The enum value comments should describe the behavior, not just repeat the Cfn value

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
nmussy, line 1: See comments in `dynamodb.ts`. Might also be a good idea to refactor some of this into a common `Props` interface and `StreamSource`, there's a lot of duplicated code there

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
nmussy: I don't know if I'm a fan of having this be an optional method, because you want sources that will have a DLQ property to implement `grantDlqPush`, and those that don't not to have that method. You're not ensuring either with that pattern. I was thinking of doing something like this instead:

```ts
interface ISource {
  /**
   * The ARN of the source resource.
   */
  readonly sourceArn: string;

  /**
   * Bind the source to a pipe.
   */
  bind(pipe: IPipe): SourceConfig;

  /**
   * Grant the pipe role read access to the source.
   */
  grantRead(grantee: IRole): void;
}

interface ISourceWithDlq extends ISource {
  grantDlqPush?(grantee: IRole): void;
}

abstract class SourceWithDlq implements ISourceWithDlq {
  readonly sourceArn: string;
  private readonly deadLetterTargetArn: string;

  abstract grantDlqPush?(grantee: IRole): void;
  abstract bind(pipe: IPipe): SourceConfig;
  abstract grantRead(grantee: IRole): void;
}

class DynamoDBSource extends SourceWithDlq {}
```

This way, you do not need to check for whether or not the method is implemented directly there:
https://github.com/msambol/aws-cdk/blob/8f5df5c453fe60409f2636d90362a22033f3e060/packages/%40aws-cdk/aws-pipes-alpha/lib/pipe.ts#L236-L238

You can instead make a guard to ensure that you have a `SourceWithDlq`:

```ts
const isSourceWithDlq = (source: ISource): source is SourceWithDlq => !!(source as SourceWithDlq).grantDlqPush;
```

nmussy (changes requested):
Tests and integs LGTM, most of my requests are docs/refactor 👍

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
msambol: API and CFN docs don't say 😂

nmussy: The user guide should help: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-dynamodb.html#pipes-ddb-batch-failures

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -29,22 +30,19 @@ export interface DynamoDBSourceParameters extends StreamSourceParameters {
 /**
  * A source that reads from an DynamoDB stream.
  */
-export class DynamoDBSource implements ISource {
+export class DynamoDBSource extends SourceWithDlq {
   private readonly table: ITableV2;
-  readonly sourceArn: string;
```
msambol: @nmussy this goes away, correct? since it's now defined in `SourceWithDlq`.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/dynamodb.ts
```diff
@@ -29,22 +30,19 @@ export interface DynamoDBSourceParameters extends StreamSourceParameters {
...
   private sourceParameters: DynamoDBSourceParameters;
-
   private deadLetterTarget?: IQueue | ITopic;
-  private deadLetterTargetArn?: string;
```
msambol: likewise here, this is gone now too?

packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
```diff
@@ -233,6 +234,10 @@ export class Pipe extends PipeBase {
      */
     const source = props.source.bind(this);
     props.source.grantRead(this.pipeRole);
+    if (props.source instanceof SourceWithDlq) {
+      props.source.grantDlqPush(this.pipeRole);
+    }
```
msambol, line 239: @nmussy this should work as well?

packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts
```diff
@@ -0,0 +1,136 @@
...
+  return undefined;
+}
+
+export function grantDlqPush(grantee: IRole, deadLetterTarget?: IQueue | ITopic) {
+  if (deadLetterTarget instanceof Queue) {
+    deadLetterTarget.grantSendMessages(grantee);
+  } else if (deadLetterTarget instanceof Topic) {
+    deadLetterTarget.grantPublish(grantee);
+  }
+}
```
msambol, line 136: @nmussy do we like these functions here in this file?

packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
nmussy: Yes, definitely 👍

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
```diff
@@ -100,3 +103,42 @@ export interface ISource {
...
 }
+
+/**
+ * Source interface with dead-letter target
+ */
+interface ISourceWithDlq extends ISource {
+  grantDlqPush(grantee: IRole, deadLetterTarget?: IQueue | ITopic): void;
+}
+
+export abstract class SourceWithDlq implements ISourceWithDlq {
```
nmussy, line 114: You can combine these and remove `ISourceWithDlq`, since `grantDlqPush` is implemented below:
```suggestion
export abstract class SourceWithDlq implements ISource {
```

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
```diff
@@ -0,0 +1,82 @@
...
+   */
+  readonly startingPositionTimestamp?: string;
+}
+
+/**
+ * A source that reads from Kinesis.
+ */
+export class KinesisSource extends SourceWithDlq {
+  private readonly stream: IStream;
+  private sourceParameters;
```
nmussy, line 42:
```suggestion
  private sourceParameters: KinesisSourceParameters;
```

packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts
nmussy, line 1: You could create a `StreamSource` class that extends `SourceWithDlq`, extend that in your Dynamo and Kinesis sources, and do the validation of the common props in the `StreamSource` constructor

msambol:
@nmussy This should be cleaner now? I appreciate the thorough review, thank you!

nmussy (approved):
LGTM 👍

packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
```diff
@@ -233,6 +233,10 @@ export class Pipe extends PipeBase {
      */
     const source = props.source.bind(this);
     props.source.grantRead(this.pipeRole);
+    if (props.source instanceof SourceWithDlq) {
```
TheRealAmazonKendra, line 236: `instanceof` doesn't always play nice with jsii. I can't remember the specifics of why off the top of my head, but I'd avoid using it here.

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
```diff
@@ -100,3 +103,48 @@ export interface ISource {
    */
   grantRead(grantee: IRole): void;
 }
+
+/**
+ * Sources that support a dead-letter target.
+ */
+export abstract class SourceWithDlq implements ISource {
```
TheRealAmazonKendra, line 110: My only comment in this file, besides a repeat on `instanceof` is the lack of consistency between using `Dlq` and `DeadLetter` in the contract. I'd prefer one or the other be chose. Usually I think it's better to write it out.

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
```diff
@@ -100,3 +103,48 @@ export interface ISource {
...
+    this.deadLetterTarget = deadLetterTarget;
+  }
+
+  abstract bind(pipe: IPipe): SourceConfig;
+  abstract grantRead(grantee: IRole): void;
+
+  /**
+   * Grants the pipe role permission to publish to the dead-letter target.
+   */
+  public grantDlqPush(grantee: IRole, deadLetterTarget?: IQueue | ITopic) {
```
TheRealAmazonKendra, line 131: I would just name this `grantPush`. Since it's in this class, it's already established that it's acting on the DLQ.

packages/@aws-cdk/aws-pipes-alpha/test/pipe.test.ts
```diff
@@ -186,6 +187,36 @@ describe('Pipe', () => {
...
+      const sourceWithDlq = new TestSourceWithDlq(topic);
+
+      new Pipe(stack, 'TestPipe', {
+        pipeName: 'TestPipe',
+        source: sourceWithDlq,
+        target,
+      });
+
+      // THEN
+      expect(sourceWithDlq.grantDlqPush).toHaveBeenCalled();
```
TheRealAmazonKendra, line 203: Fine to check for this but these tests should also check that the right grant has been added to the right role using our template matchers in our assert library.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/enums.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Define how to handle item process failures.
+ */
+export enum OnPartialBatchItemFailure {
```
TheRealAmazonKendra, line 4: Do we expect this to have more values in the future? I don't see the benefit of using an enum here as is.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * With StartingPosition set to AT_TIMESTAMP, the time from which to start reading, in ISO 8601 format.
+   *
+   * @example
+   * '2025-01-01T00:00:00Z'
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcekinesisstreamparameters.html#cfn-pipes-pipe-pipesourcekinesisstreamparameters-startingpositiontimestamp
+   * @default - no starting position timestamp
+   */
+  readonly startingPositionTimestamp?: string;
```
TheRealAmazonKendra, line 27: Do we not have any type for this? I guess there would be very little to add but it seems like a thing we might have in core.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts
```diff
@@ -0,0 +1,145 @@
...
+   * Minumum: Duration.seconds(0)
+   * Maxiumum: Duration.seconds(300)
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipesourcekinesisstreamparameters.html#cfn-pipes-pipe-pipesourcekinesisstreamparameters-maximumbatchingwindowinseconds
+   * @default - the events will be handled immediately
+   */
+  readonly maximumBatchingWindow?: Duration;
+
+  /**
+   * (Streams only) Discard records older than the specified age. The default value is -1, which sets the maximum age to infinite. When the value is set to infinite, EventBridge never discards old records.
```
TheRealAmazonKendra, line 42: The several times that I'm seeing `(Streams only)` makes me think we may have the wrong level of abstraction here. If this was scoped only to where it is relevant, this comment wouldn't be necessary.

TheRealAmazonKendra (changes requested):
Thank you for your contribution! Please see my comments inline.

packages/@aws-cdk/aws-pipes-sources-alpha/lib/enums.ts
msambol: I'm not sure on this. Possible AWS decides to add another method in the future?

packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
msambol: What's the preferred route here?

packages/@aws-cdk/aws-pipes-sources-alpha/lib/kinesis.ts
msambol: I updated this to `Date`.

packages/@aws-cdk/aws-pipes-alpha/test/pipe.test.ts
msambol: Done.

packages/@aws-cdk/aws-pipes-alpha/lib/source.ts
msambol: I changed everything to `...DeadLetterTarget`.

packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
nmussy: A static guard method should work well:

```ts
class SourceWithDlq implements ISource {
  //...
  static isSourceWithDlq(source: ISource): source is SourceWithDlq {
    return (source as SourceWithDlq).sourceArn !== undefined;
  }
}

**Comment by msambol** [comment]
@TheRealAmazonKendra I think this is ready for another look... thank you for your feedback.

File: packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts (hunk 41 above)
Comment (TheRealAmazonKendra) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts:42]: Ah, so doc weirdness rather than abstraction weirdness. I'm going to check in with the service team on this to make sure there's no actual weirdness that I'm missing here. Gimme a day or two to chat with them and I'll do the next round of reviews.

**Comment by msambol** [comment]
@TheRealAmazonKendra Any further thoughts on this one?

**Comment by msambol** [comment]
@TheRealAmazonKendra when you have time, I'd love another pass on this so I can work on the other sources. Thanks!

File: packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts (hunk 41 above)
Comment (redwheeler3) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts:42]: I'm Jeff from the EventBridge Pipes service team. I'll have a chat with our doc writer about making this clearer. It's an internal implementation issue that probably shouldn't be exposed. The "streams pollers" are the ones that poll from DynamoDB and Kinesis. When it says "streams only" it's synonymous with "DynanmoDB and Kinesis sources only". 

As a concrete example, let's consider "MaximumRecordAgeInSeconds". That applies to DynamoDB and Kinesis sources (it's in the PipeSourceKinesisStreamParameters and PipeSourceDynamoDBStreamParameters data type) but not for SQS / Kafka / MQ. 

Do you have a suggestion on what I can change the docs to? I'm thinking of just killing "(streams only)".

Comment (msambol) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts:42]: I agree with the killing that verbiage. But good to know they are grouped together, I think I implemented this _mostly_ correct (thanks you @nmussy). @redwheeler3 Once this PR is in, I'll knock out the other sources & targets.

Comment (redwheeler3) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts:42]: OK. I filed a ticket with our doc team to get that verbiage removed. It may take a while, but eventually it will be gone.

Comment (msambol) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts]: @TheRealAmazonKendra I removed `(Streams only)` from the documentation.

File: packages/@aws-cdk/aws-pipes-sources-alpha/lib/enums.ts (hunk 39 above)
Comment (redwheeler3) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/enums.ts:4]: We did it this way for future-proofing. We don't have plans to expand the functionality beyond "automatic_bisect" but wanted to leave the option open to permit it. Once an API is delivered, it's extremely hard to change it. In the CDK world, if changing things in the future is also considered difficult, go with the enum... if it's easy to change it to an enum later, you are probably good with something simpler.

Comment (msambol) [reply packages/@aws-cdk/aws-pipes-sources-alpha/lib/enums.ts:4]: Ok I'd prefer to leave this as an enum too.

**Comment by msambol** [comment]
@TheRealAmazonKendra Have time for a review this week? I'd love to push forward on the Pipes features.

**Comment by RaphaelManke** [comment]
> @TheRealAmazonKendra Have time for a review this week? I'd love to push forward on the Pipes features.

Or maybe @GavinZZ can have a look at this one?

**Comment by GavinZZ** [comment]
I will take a look but since @TheRealAmazonKendra has already reviewed and left feedback, I would love her to have the final approval. I can help reaching out to her internally and asking for her attention on this PR.

**Comment by vvo** [comment]
I would love to use this module, thanks a lot!

**Comment by davidmjames** [comment]
Is this still just awaiting a final sign-off before a merge, or is there more to it? It would be nice to have official support for these sources.

**Comment by garysassano** [comment]
I would also like to see this merged.

**Comment by msambol** [comment]
@TheRealAmazonKendra Gentle poke here. Possible to give this another pass?

**Comment by GavinZZ** [comment]
@msambol I've reached out to Kendra just now internally. She's on leave right now, but once I hear back from her, if she has capacity will let me do the final review; if she doesn't, I'll review and approve it.

File: packages/@aws-cdk/aws-pipes-sources-alpha/lib/streamSource.ts [hunk 42]
```diff
@@ -0,0 +1,143 @@
...
+function validateBatchSize(batchSize?: number) {
+  if (batchSize !== undefined && !Token.isUnresolved(batchSize)) {
+    if (batchSize < 1 || batchSize > 10000) {
+      throw new Error(`Batch size must be between 1 and 10000, received ${batchSize}`);
+    }
+  }
+}
+
+function validateMaximumBatchingWindow(window?: number) {
+  if (window !== undefined) {
```
msambol, line 113: since `maximumBatchingWindow` is a `Duration`, is there a good way to check for a `Token` here? Or can this check be skipped?

GavinZZ: I think we can skip it given the `window` is from `Duration.toSeconds`

packages/@aws-cdk/aws-pipes-sources-alpha/test/__snapshots__/dynamodb.test.ts.snap
```diff
@@ -0,0 +1,68 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
GavinZZ, line 1: What is this snapshot file?

packages/@aws-cdk/aws-pipes-sources-alpha/test/integ.dynamodb.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const message = test.assertions.awsApiCall('SQS', 'receiveMessage', {
+  QueueUrl: targetQueue.queueUrl,
+});
+
+message.assertAtPath('Messages.0.Body.dynamodb.Keys.id.S', ExpectedResult.stringLikeRegexp('myPK')).waitForAssertions({
+  totalTimeout: cdk.Duration.minutes(2),
+  interval: cdk.Duration.seconds(15),
+});
```
GavinZZ, line 75: Nice, I like to see integ tests with assertions to validate the results.

packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
```diff
@@ -233,6 +233,10 @@ export class Pipe extends PipeBase {
      */
     const source = props.source.bind(this);
     props.source.grantRead(this.pipeRole);
+    if (SourceWithDeadLetterTarget.isSourceWithDeadLetterTarget(props.source)) {
+      props.source.grantPush(this.pipeRole, props.source.deadLetterTarget);
+    }
```
GavinZZ, line 238: Can you add some comments to explain why this grantPush permission is needed?

GavinZZ (commented):
Left some minor feedback. I think overall this looks good to me and I'm happy to approve once the comments are addressed.

Also sorry for the extra long wait on this PR review. Kendra is on vacation. I see her feedbacks are all addressed so I'll review the PR on behalf for her. Given this is alpha modules, if she has more outstanding comments, we can address it in a future PR.

packages/@aws-cdk/aws-pipes-sources-alpha/test/__snapshots__/dynamodb.test.ts.snap
msambol: It's a result of this test: https://github.com/aws/aws-cdk/pull/29476/files#diff-1196955de9d57d66bf83a476e71951d39e5e2735a0eda43541d7842e80bef677R187-R212.

packages/@aws-cdk/aws-pipes-alpha/lib/pipe.ts
msambol: I added some notes in https://github.com/aws/aws-cdk/pull/29476/commits/ecba4fa2c25d57cfd8d0877f7573c198e0993d47.

msambol:
@GavinZZ Thanks for the review! I know this is a large one.

GavinZZ (approved):
LGTM, thank you for the effort and patience!

PR #29486, opened 2024-03-14
feat(api-gateway): deploy RestApi to existing stage

### LINKED ISSUE #25582: api-gateway: Allow to use an existent stage for deployments
### Describe the feature

Ability to specify existent environments name when creating a new deployment for an API Gateway.

Currently is possible to do that bypassing cdk lib type definition using the suggestion specified here https://stackoverflow.com/questions/63950199/how-to-use-an-existing-stage-in-api-gateway-deployments-in-aws-cdk

``` typescript

const deployment = new apigateway.Deployment(this, "Deployment", {
      api: restApi,
    });
    //deploys an existent stage (dev, stage, prod) - 
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    (deployment as any).resource.stageName = "myStage";
    
```

See more information here: https://stackoverflow.com/questions/63950199/how-to-use-an-existing-stage-in-api-gateway-deployments-in-aws-cdkenvironment

### Use Case

When you have a central API Gateway and multiple other APIs/resources defined in other stacks/repo, we want to be able to deploy it to an "existent" stage without having to create/replace a new one. 

1. I have a central API Gateway which has multiple resources:
     - `/resource-A` -> points to an external resources defined by an external `Stack-A`
     - `/resource-B` -> points to an external resources defined by an external `Stack-B`.
     
    
I want to be able to re deploy `Stack-A` and `Stack-B` and also create a new deployment for an existent environment - the current options on CDK Constructs doesn't allow to easily do that and suggest creating a new stage by using `new Stage`
as define here https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.Deployment.html.

image

### Proposed Solution

A property in the deployment construct to use an existent schema (either by stageName or an `IStage` interface)

```typescript

 const deployment = new apigateway.Deployment(this, "Deployment", {
      api: restApi,
     stageName: stageName // this extra property to utilize an existent stage 
    });
    
 ```
 
 This is the current way of doing that and I can confirm it works:
 
 ```typescript
 
    (deployment as any).resource.stageName = "myStage";
    
```
    
 I also considered trying to assign a stage to the rest api construct, but the deployment Stage expects a `Stage` object and doesn't not accept `IStage`.
 
 ```typescript
 
   restApi.deploymentStage = apigateway.Stage.fromStageAttributes(
      this,
      "sggg",
      {
        restApi,
        stageName: "myStage",
      }
    );

```

### Other Information

_No response_

### Acknowledgements

### CDK version used

"aws-cdk-lib": "^2.42.1",

MacOS

### PR DESCRIPTION
**Pull request by gracelu0**
### Issue #

Closes #25582 .

### Reason for this change

If I have to manually create a new deployment for API gateway but want to deploy to an existing stage instead of creating a new stage, the current workaround is `(deployment as any).resource.stageName = "myStage";`

### Description of changes

Added a new property `stageName` to `Deployment`. Updated README with two new sections: `Deploying to an existing stage` and `Controlled triggering of deployments` to show how users can trigger new deployments with their latest API changes if they manually create a `Deployment` resource for RestApi. 

Since `RestApiBase` manages `Deployment` constructs, this change applies to both `SpecRestApi` and `RestApi`. 

Added unit tests and integration tests. Also successfully deployed a sample app to my account where I created a stage in the API gateway console and specified the stage name in the new deployment.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/custom-resource-handlers/lib/core/cross-region-ssm-writer-handler/index.ts
+++ b/packages/@aws-cdk/custom-resource-handlers/lib/core/cross-region-ssm-writer-handler/index.ts
@@ -102,8 +102,8 @@ async function throwIfAnyInUse(ssm: SSM, parameters: CrossRegionExports): Promis
   }));
 
   if (tagResults.size > 0) {
-    const message: string = Object.entries(tagResults)
-      .map((result: [string, string[]]) => `${result[0]} is in use by stack(s) ${result[1].join(' ')}`)
+    const message: string = Array.from(tagResults.entries())
+      .map((result) => `${result[0]} is in use by stack(s) ${Array.from(result[1]).join(' ')}`)
       .join('\n');
     throw new Error(`Exports cannot be updated: \n${message}`);
   }
--- a/packages/@aws-cdk/custom-resource-handlers/test/core/cross-region-ssm-writer-handler.test.ts
+++ b/packages/@aws-cdk/custom-resource-handlers/test/core/cross-region-ssm-writer-handler.test.ts
@@ -1,7 +1,7 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import { InvalidResourceId } from '@aws-sdk/client-ssm';
-import { SSM_EXPORT_PATH_PREFIX } from '../../lib/core/types';
 import { handler } from '../../lib/core/cross-region-ssm-writer-handler/index';
+import { SSM_EXPORT_PATH_PREFIX } from '../../lib/core/types';
 
 let mockPutParameter: jest.Mock = jest.fn() ;
 let mocklistTagsForResource: jest.Mock = jest.fn();
@@ -96,7 +96,7 @@ describe('cross-region-ssm-writer entrypoint', () => {
       });
 
       // THEN
-      await expect(handler(event)).rejects.toThrow(/Exports cannot be updated/);
+      await expect(handler(event)).rejects.toThrow('Exports cannot be updated: \n/cdk/exports/MyStack/MyExport is in use by stack(s) MyStack');
     });
 
     test('Create event does not throw for new parameters', async () => {

### CODE REVIEW

File: packages/aws-cdk-lib/aws-apigateway/README.md [hunk 1]
```diff
@@ -995,6 +995,21 @@ const api = new apigateway.RestApi(this, 'books', {
   cloudWatchRoleRemovalPolicy: cdk.RemovalPolicy.DESTROY,
 });
 ```
+### Deploying to an existing stage
+
+If you want to use an existing stage for a deployment, you can specify the stage name in the deployment as follows:
+```ts
+const restApi = new apigateway.RestApi(this, 'my-api', {
+  deploy: false,
+});
```
godwingrs22, line 1004: Seems like existing stage is not created while creating the api in the example. Is below required? If so can we update it in `README.md`

``` 
 deployOptions: {
    stageName: 'dev',
  }
```

packages/aws-cdk-lib/aws-apigateway/lib/deployment.ts
```diff
@@ -26,6 +26,13 @@ export interface DeploymentProps {
    * @default false
    */
   readonly retainDeployments?: boolean;
+
+  /**
+   * The stage of the API Gateway deployment.
+   *
+   * @default - If `stageName` is not specified, a "prod" stage will be automatically created for the API Gateway deployment.
```
godwingrs22, line 33: nit: It could be worth to mention this also in `README.md`

packages/aws-cdk-lib/aws-apigateway/test/deployment.test.ts
```diff
@@ -126,6 +126,21 @@ describe('deployment', () => {
...
+    const stack = new Stack();
+    const api = new apigateway.RestApi(stack, 'api', { deploy: false, cloudWatchRole: false });
+    api.root.addMethod('GET');
+
+    // WHEN
+    new apigateway.Deployment(stack, 'deployment', { api, stageName: 'dev' });
+
+    // THEN
+    Template.fromStack(stack).hasResourceProperties('AWS::ApiGateway::Deployment', {
+      StageName: 'dev',
```
godwingrs22, line 140: Can we also have a test if the stage name is not specified in the Deployment, are we expecting default stage name `prod` to be used for deployment?

packages/aws-cdk-lib/aws-apigateway/lib/deployment.ts
godwingrs22, line 33: qq: Does Deployment creates the `stage` or only refer the existing stage? 
Because default stage `prod` is created by `RestApi` I think. Worth to double check. If so we can change the description something like 
"...default existing `prod` stage will be used for API Gateway Deployment"

godwingrs22 (commented):
left few minor comments.

packages/aws-cdk-lib/aws-apigateway/lib/deployment.ts
gracelu0: from what I understand Deployment does not create a new stage and only refers to the existing stage by `stageName` - I'll update the description

packages/aws-cdk-lib/aws-apigateway/test/deployment.test.ts
gracelu0: If a new `Deployment` is created on its own, we do not expect `prod` to be used by default since the Deployment is not being created by the `RestApi`. 

It also seems that when `RestApi` automatically creates a new deployment, it doesn't set the `StageName` property (see [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-apigateway/lib/restapi.ts#L622-L625)). Instead it specifies the associated `deployment` when creating the stage (see [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-apigateway/lib/restapi.ts#L633). I think this makes sense because a stage can only be associated with one deployment, whereas a deployment could be deployed to multiple stages.

packages/aws-cdk-lib/aws-apigateway/README.md
gracelu0: I believe if `deploy: false` is set then `RestApi` does not automatically create a deployment and stage. As such it's also not possible to set `deployOptions` if `deploy` is set to false. Doing so and running `cdk synth` results in error: `Cannot set 'deployOptions' if 'deploy' is disabled`.

So this example is where the user wants to manually create a deployment and set the stage to one they've already created. I tried deploying it with a API gateway stage called 'dev' in my account and it didn't create a "prod" stage for me but let me know if you found otherwise.

godwingrs22: Thank you for the explanation. But I'm still not clear on where the existing stage `dev` is defined. I hope  `stageName: 'dev'` is only refering the existing stage which is already available and doesn't create if not available already. Please correct me if I'm wrong. Hence, It could be worth mentioning in the example with the stage (existing) defined that is used in the deployment.

```
const deployment = new apigateway.Deployment(this, 'my-deployment', {
  api: restApi,
  stageName: 'dev',
});
```

packages/aws-cdk-lib/aws-apigateway/lib/deployment.ts
```diff
@@ -137,6 +151,7 @@ class LatestDeploymentResource extends CfnDeployment {
     super(scope, id, {
       description: props.description,
       restApiId: props.restApi.restApiId,
+      stageName: props.stageName,
```
godwingrs22, line 154: Could you please elaborate why we need to add the `stageName` property to LatestDeployment? Based on my previous comment, I thought inorder to use Deployment, restApi will be created with `Deploy:false` which will not automatically create the deployment and stage. Hence in that case latestDeployment will not be used if `deploy:false` is set ([See here](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.RestApi.html#latestdeployment)).

packages/@aws-cdk-testing/framework-integ/test/aws-apigateway/test/integ.restapi.import-deploymentstage.ts
```diff
@@ -0,0 +1,23 @@
+import * as cdk from 'aws-cdk-lib/core';
+import * as apigateway from 'aws-cdk-lib/aws-apigateway';
+import * as integ from '@aws-cdk/integ-tests-alpha';
+
+const app = new cdk.App();
+const stack = new cdk.Stack(app, 'integtest-restapi-import-deployment-stage');
+
+const api = new apigateway.RestApi(stack, 'my-api', {
+  deploy: false,
```
godwingrs22, line 9: if `deploy:false` is used, seems we need to use [deploymentStage](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.RestApi.html#deploymentstage) to point the stage explicitly to the restapi.

godwingrs22 (commented):
left some comments

packages/aws-cdk-lib/aws-apigateway/lib/deployment.ts
gracelu0: When the user manually creates a `Deployment` the constructor creates a new `LatestDeploymentResource` (see [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-apigateway/lib/deployment.ts#L71)), so that's why it's getting added here

packages/@aws-cdk-testing/framework-integ/test/aws-apigateway/test/integ.restapi.import-deploymentstage.ts
```diff
@@ -0,0 +1,23 @@
...
+});
+api.root.addMethod('GET');
+
+// Create a new deployment that uses existing stage
+new apigateway.Deployment(stack, 'MyManualDeployment', {
+  api: api,
+  stageName: 'myStage',
+});
```
godwingrs22, line 17: I tried to run this integ test and found the below finding. Overall great work and I'm able to create deployment for existing stage with one minor addition to the integ test as below.

1. If we manually define the deployment with `stageName` included, the deployment creates the stage with the stageName provided and creates the deployment. If the stage already exists, then it deploys to the existing stage. Hence I would suggest to add to `README.md` and also in the `stageName` property docstring that something like stage will be created if it doesn't exist.

2. Also, When the restApi model has changes, the deployment will remain the same and doesn't reflect the new changes to the existing stage. In order for the deployment to reflect the new changes to the existing stage, the user also needs to invoke the `addToLogicalId(data)` method which generates a new logical Id. For more info see [here](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.Deployment.html). Without invoking this method, the deployment will not reflect new changes in the stage.

Hence, it would be worth to mention to use the method in our integ test and readme examples if deployment is manually defined using `Deployment` construct. But for `deploy:true`,  restApi latestDeployment takes care to generate new logical Id automatically and deploys to the existing stage.

```
const api = new apigateway.RestApi(stack, 'my-api', {
  deploy: false,
});
api.root.addMethod('GET');
api.root.addMethod('POST');

const deployment = new apigateway.Deployment(stack, 'MyManualDeployment', {
  api: api,
  stageName: 'myStage',
});

deployment.addToLogicalId(`Deployment-${Date.now()}`);

new integ.IntegTest(app, 'restapi-import-deployment-stage', {
  testCases: [stack],
});
```

Apart from that, it LGTM.

godwingrs22 (approved):
LGTM. Thanks @gracelu0 for your effort in implmenting this much awaited feature.

leoujjawal:
Hi @godwingrs22 : I am still unable to use this in CDK Python v 2.139.1

gracelu0:
> Hi @godwingrs22 @gracelu0 : I am still unable to use this in CDK Python v 2.139.1

Hi @leoujjawal it looks like this didn't make it into the latest release - it should be included the next release.

leoujjawal:
> > Hi @godwingrs22 @gracelu0 : I am still unable to use this in CDK Python v 2.139.1
> 
> Hi @leoujjawal it looks like this didn't make it into the latest release - it should be included the next release.

Thanks for the reply. Is there a way to use the workaround solution in Python? I tried `deployment.resource.stage_name = "myStage"`, but it complains deployment has no resource.

eripoll:
I have been trying to use this but to no avail. I see my resources in the resource view, I see a new deployment is active, but the stage hasn't changed.
What am I doing wrong here?

```
        deployment = apigw.Deployment(
            self,
            deployment_name,
            api=api,
            retain_deployments=True,
            description=description,
            stage_name=env_name,
        )
```

leoujjawal:
> I have been trying to use this but to no avail. I see my resources in the resource view, I see a new deployment is active, but the stage hasn't changed. What am I doing wrong here?
> 
> ```
>         deployment = apigw.Deployment(
>             self,
>             deployment_name,
>             api=api,
>             retain_deployments=True,
>             description=description,
>             stage_name= stage_name,
>         )
> ```
All you got to do is - 

# Always create a new deployment
deployment = Deployment(self, str(uuid.uuid4()), api=rest_api)

v1_stage = Stage(
            self,
            "StageV1",
            stage_name="v1",
            deployment=deployment,
            logging_level=MethodLoggingLevel.INFO,
            data_trace_enabled=False,
            tracing_enabled=True,
        )

 rest_api.deployment_stage = v1_stage

leoujjawal:
> I have been trying to use this but to no avail. I see my resources in the resource view, I see a new deployment is active, but the stage hasn't changed. What am I doing wrong here?
> 
> ```
>         deployment = apigw.Deployment(
>             self,
>             deployment_name,
>             api=api,
>             retain_deployments=True,
>             description=description,
>             stage_name= stage_name,
>         )
> ```

All you got to do is - 

```
# Always create a new deployment
deployment = Deployment(self, str(uuid.uuid4()), api=rest_api)

v1_stage = Stage(
            self,
            "StageV1",
            stage_name="v1",
            deployment=deployment,
            logging_level=MethodLoggingLevel.INFO,
            data_trace_enabled=False,
            tracing_enabled=True,
        )

 rest_api.deployment_stage = v1_stage
```

PR #29492, opened 2024-03-14
fix(CLI): `cdk diff` stack deletion causes a race condition

### LINKED ISSUE #29265: (cli): change-set diff not required for new stack diffs

comcalvi:
*Co-authored by*: @scanlonp

### Issue # (if applicable)

Closes #29265.

### Reason for this change

Creating a changeset for a stack that has not been deployed yet causes CFN to create a stack in state `REVIEW_IN_PROGRESS`. Previously we deleted this empty stack, but did not wait for the stack status to be `DELETE_COMPLETE`. This allowed `cdk diff` to exit while the stack status was still `DELETE_IN_PROGRESS`, which can cause subsequent CDK commands to fail, because a stack deletion operation is still in progress. 

### Description of changes

No longer create the changeset if the stack doesn't exist. Only perform the existence check if the changeset parameter is specified, to avoid a permission error when looking up a stack. 

### Checklist

----


TheRealAmazonKendra (changes requested):
Correct me if I'm wrong here, but since we reverted https://github.com/aws/aws-cdk/pull/29485, if this is a cdk migrate based stack, we will now always tell customers that we are creating new resources instead of importing them. The original change was to prevent this very thing.

TheRealAmazonKendra:
I think the correct fix here is _mostly_ what @scanlonp did in https://github.com/aws/aws-cdk/pull/29394. The one piece that would have prevented the role switch here was if in `packages/aws-cdk/lib/cdk-toolkit.ts`, instead of 
```
      const stackExistsOptions = {
        stack: stacks.firstStack,
        deployName: stacks.firstStack.stackName,
      };

      const stackExists = await this.props.deployments.stackExists(stackExistsOptions);

      const changeSet = (stackExists && options.changeSet) ? await createDiffChangeSet({
```
The change had been:
```
      const changeSet = options.changeSet && (await this.props.deployments.stackExists(stackExistsOptions) ? await createDiffChangeSet({
```
Because the if statement will exit if `options.changeSet` if false before looking up whether or not the stack exists.

scanlonp (commented):
Looks good to me. 

Only comments would be: 

- is there a functional difference in doing a try catch block in `cleanupOldChangeset()` as opposed to checking the exists parameter, or is this just for code clarity? 
- If we have a migrate stack, it will not exist beforehand, so the logic could be streamlined. But it is not blocking and I am good with how it is.

comcalvi:
>is there a functional difference in doing a try catch block in cleanupOldChangeset() as opposed to checking the exists parameter, or is this just for code clarity?

turns out that `deleteChangeset` will throw if the stack doesn't exist. This was preventing a changeset from being created for me at some point, but that was because I was creating a changeset when a stack didn't exist, which is the wrong behavior. I've removed this.

> If we have a migrate stack, it will not exist beforehand, so the logic could be streamlined. But it is not blocking and I am good with how it is.

I think I've altered the logic you're referring to, it should be more clear now.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -168,16 +180,29 @@ export class CdkToolkit {
...
+              uuid: uuid.v4(),
+              deployments: this.props.deployments,
+              willExecute: false,
+              sdkProvider: this.props.sdkProvider,
+              parameters: Object.assign({}, parameterMap['*'], parameterMap[stack.stackName]),
+              resourcesToImport,
+              stream,
+            });
+          } else {
+            debug(`the stack '${stack.stackName}' does not exist, skipping changeset creation.`);
```
scanlonp, line 203:
```suggestion
            debug(`the stack '${stack.stackName}' has not been deployed to CloudFormation, skipping changeset creation.`);
```
minor nit: thoughts on this debug message? Might be unclear what "does not exist" means.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -136,20 +136,32 @@ export class CdkToolkit {
...
+            stack: stacks.firstStack,
+            uuid: uuid.v4(),
+            deployments: this.props.deployments,
+            willExecute: false,
+            sdkProvider: this.props.sdkProvider,
+            parameters: Object.assign({}, parameterMap['*'], parameterMap[stacks.firstStack.stackName]),
+            stream,
+          });
+        } else {
+          debug(`the stack '${stacks.firstStack.stackName}' does not exist, skipping changeset creation.`);
```
scanlonp, line 157:
```suggestion
            debug(`the stack '${stack.stackName}' has not been deployed to CloudFormation, skipping changeset creation.`);
```
Same idea here.

scanlonp (commented):
Good to approve from me! I'll wait for Kendra to take a look. 

Minor nit on the debug message.

packages/aws-cdk/lib/cdk-toolkit.ts
comcalvi: this one is `firstStack`, since this is in the fixed template branch.

scanlonp: Copy & paste mistake 👍

comcalvi (commented):
test pipeline deployment is complaining there's no review...

TheRealAmazonKendra (approved):

PR #29507, opened 2024-03-15
feat(ec2): instance type properties

### LINKED ISSUE #12022: (aws-ec2): Expose instance properties on InstanceType
There are many supporting properties unique to EC2 [InstanceType](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.InstanceType.html). Some AWS supporting services are conditioned on these properties so they should be readily available to developers in the framework. 

### Use Case

The specific usecase I have in mind is [Cloudwatch monitoring of SageMaker endpoints](https://docs.aws.amazon.com/sagemaker/latest/dg/monitoring-cloudwatch.html), from there:

> CPUUtilization: The percentage of CPU units that are used by the containers on an instance. The value can range between 0 and 100, and is multiplied by the number of CPUs. For example, if there are four CPUs, CPUUtilization can range from 0% to 400%

This requires the manual process of looking up the relevant information for your current instance type on the web and hard-coding it in the CDK. An ideal CX would be one where the developer doesn't need to leave CDK to get the information required. 

### Proposed Solution

There could be some static mapping or interface of various instance properties which are keyed on [InstanceType](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.InstanceType.html). As a first step some of the properties from https://aws.amazon.com/ec2/instance-types/ could be incorporated with the intent to expand to all types + properties in the future. 

---

nmussy:
### Issue # (if applicable)

Closes #12022.

### Reason for this change

Provide CDK users with both a list of valid class/type combinations, as well as the properties associated to those types.
Additionally, these properties could be used internally to give user warnings/errors if they are associated to limitations, such as Nitro support, cluster networking, etc.

### Description of changes

* Instantiate each available `InstanceType.INSTANCE_CLASS_INSTANCE_SIZE`, both with and without its class alias
* Create a `instance-properties.json` file containing the complete output of the `ec2:DescribeInstanceTypes` API command
  * `aws ec2 describe-instance-types | jq '.InstanceTypes | { InstanceTypes: map({key: .InstanceType, value: del(.InstanceType)}) | sort_by(.key) | from_entries } | .InstanceTypes' > packages/aws-cdk-lib/aws-ec2/data/instance-properties.json`
* Map this data to a public `instanceType.instanceProperties`, allowing CDK users to access them

I'm going to add tests to make sure the data is exposed correctly, but given this data is only meant to be informative, it shouldn't have an impact on deployments

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ec2/README.md
+++ b/packages/aws-cdk-lib/aws-ec2/README.md
@@ -1264,7 +1264,6 @@ new ec2.Instance(this, 'Instance', {
     cpuType: ec2.AmazonLinuxCpuType.ARM_64,
   }),
 });
-
 ```
 
 ### Latest Amazon Linux Images
--- a/packages/aws-cdk-lib/rosetta/aws_ec2/default.ts-fixture
+++ b/packages/aws-cdk-lib/rosetta/aws_ec2/default.ts-fixture
@@ -1,4 +1,5 @@
 // Fixture with packages imported, but nothing else
+import assert = require('assert');
 import { Construct } from 'constructs';
 import { Aspects, Duration, Fn, Size, Stack, StackProps } from 'aws-cdk-lib';
 import ec2 = require('aws-cdk-lib/aws-ec2');


nmussy:
Looking for feedback here 👋 

This currently just a proof of concept to discuss the design. Once we settle on it, I'll add the missing properties (CPU, Memory, etc.), as well as generate the complete JSON data and static instance type values.

nmussy:
Exemption Request: An integration test won't be needed, this is not changing the `InstanceType` behavior

GavinZZ:
I don't quite understand the change and how the shell script was used. The script seems to be a manual invocation so whenever there's an update, does it require manually invoking this script?

Given the size of the changes in this PR, it makes me a bit uncomfortable to review or approve a PR of this size without any tests. Would you be able to introduce any integ tests?

GavinZZ:
Adding a `pr-linter/no-exemption` as I think we should add an integ test for this change.

pahud:
Hi @nmussy 

The team is now reviewing all community PRs and this one seems to be stale for a while. Are you still planning to continue the work? If yes, please resolve the conflicts, convert the state from Draft to "Ready for review" and make sure CI pass. Then tag me and I'll reach out to the maintainer for review. 

Thanks.

PR #29513, opened 2024-03-16
fix(elasticloadbalancerV2): logicalId supports switch from addTargetGroups (under feature flag)

### LINKED ISSUE #29496: ElasticLoadBalancerV2: migration to addAction() from addTargetGroups() fails because of logicalId change
### Describe the bug

Old code
```
      });
```

New code

```
      });
```

```
Resources
```

### Expected Behavior

### Current Behavior

### Reproduction Steps

Sigh. 

### Possible Solution

_No response_

### CDK CLI Version

2.100.0 (build e1b5c77)

### Framework Version

_No response_

### Node.js Version

20.5.1

### OS

MacOS

### Language

TypeScript

### Language Version

5.3.3

### Other information

_No response_

ahammond:
### Issue ELBv2 logicalId inconsistency of ApplicationListenerRule logicalIds

Mitigates #29496

### Reason for this change

People using ALBs who need to migrate from the `addTargetGroups()` convenience method to the lower level `addAction()` method should not be blocked due to inconsistent logicalId's. Further, the logicalIds should be consistent going forward.

### Description of changes

There are two feature flags, one which sets a migration compat mode and another which fixed the behaviour to be consistent.

Unit testing.

### Checklist

----


ahammond:
Per https://github.com/ahammond/aws-cdk/blob/62be9c7a06e3191ea8ada225eb69c626d22033e4/INTEGRATION_TESTS.md#L32 none of the conditions which require an integration test pertain.

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1034,6 +1036,41 @@ export const FLAGS: Record<string, FlagInfo> = {
...
+      enabling a migration from the \`addTargetGroups()\` method.
+
+      For new resources, see instead the ${ENABLE_ALBV2_EXTERNALAPPLICATIONLISTENER_ADDTARGETGROUP_CONSISTENT_LOGICALID} flag.
+    `,
+    introducedIn: { v2: 'V2NEXT' },
+    recommendedValue: false,
+  },
+
+  //////////////////////////////////////////////////////////////////////
+  [ENABLE_ALBV2_EXTERNALAPPLICATIONLISTENER_ADDTARGETGROUP_CONSISTENT_LOGICALID]: {
```
paulhcsun, line 1063: I think that just having the other feature flag is sufficient. I can only think of a couple of scenarios where we would want to have this second feature flag:
1. Users want to keep `Rule` in the logicalId
2. Eventually get everyone to use this feature flag and then we can remove the feature flags altogether and migrate to add `Rule` to the `addTargetGroups()` case which I can't see happening. 

Could you explain why you feel this second feature flag is necessary? My main worry is that it will cause confusion with the first feature flag.

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1034,6 +1036,41 @@ export const FLAGS: Record<string, FlagInfo> = {
     introducedIn: { v2: 'V2NEXT' },
     recommendedValue: true,
   },
+
+  //////////////////////////////////////////////////////////////////////
+  [ENABLE_ALBV2_EXTERNALAPPLICATIONLISTENER_ADDTARGETGROUP_TO_ADDACTION_MIGRATION]: {
```
paulhcsun, line 1041: I don't think we need to add `ENABLE` to the name here, also I feel like `MIGRATE` suggests future changes that might not necessarily be coming (if that makes sense). 

```suggestion
  [ALBV2_EXTERNALAPPLICATIONLISTENER_SWITCH_FROM_ADDTARGETGROUP_TO_ADDACTION]: {
```

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1034,6 +1036,41 @@ export const FLAGS: Record<string, FlagInfo> = {
...
+    type: FlagType.VisibleContext,
+    summary: 'When enabled, you can migrate from the \`addTargetGroups()\` method of declaring a \`ListenerRule\` to the \`addAction()\` method, without changing the logicalId and replacing your resource.',
```
paulhcsun, line 1043:
```suggestion
    summary: 'When enabled, you can switch from the \`addTargetGroups()\` method of declaring a \`ListenerRule\` to the \`addAction()\` method, without changing the logicalId and replacing your resource.',
```

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1034,6 +1036,41 @@ export const FLAGS: Record<string, FlagInfo> = {
...
+    detailsMd: `
+      When migrating from a less complex to a more complex use of ALB,
```
paulhcsun, line 1045:
```suggestion
      When switching from a less complex to a more complex use of ALB,
```

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1034,6 +1036,41 @@ export const FLAGS: Record<string, FlagInfo> = {
...
+      you will eventually need features not available in the \`addTargetGroups()\` convenience method.
+      In this case you will want to use the \`addAction()\` method.
+      Without this feature being activated, the change will also add a \`Rule\` suffix to the logicalId of your \`ListnerRule\`,
```
paulhcsun, line 1048:
```suggestion
      When this feature is not enabled, switching over to \`addAction()\` from using \`addTargetGroups()\` will add a \`Rule\` suffix to the logicalId of your \`ListnerRule\`,
```

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1034,6 +1036,41 @@ export const FLAGS: Record<string, FlagInfo> = {
...
+      causing CloudFormation to attempt to replace the resource.
+      Since ListenerRules have a unique priority,
+      CloudFormation will always fail when generating the new ListenerRule.
+
+      Setting this feature flag will cause the \`addAction()\` method to not add the \`Rule\` suffix on the logicalId,
+      enabling a migration from the \`addTargetGroups()\` method.
```
paulhcsun, line 1054:
```suggestion
      allowing you to switch from the \`addTargetGroups()\` method without having CloudFormation generate a new resource.
```

paulhcsun (changes requested):
Thank you for your contribution @ahammond! The changes look good to me overall but I've just let some comments regarding the`ENABLE_ALBV2_EXTERNALAPPLICATIONLISTENER_ADDTARGETGROUP_CONSISTENT_LOGICALID` feature flag as well as some wording changes. 

We apologize for the delay in reviewing your PR. We'll make sure this gets prioritized and hopefully merged as part of the 2.136.0 release this week. We appreciate your patience.

packages/aws-cdk-lib/cx-api/lib/features.ts
ahammond: Users don't care about the contents of logicalIds until it causes them problems. This flag allows anyone who actually does care to be future friendly, while allowing the vast majority of users who don't care to continue not caring. When the logicalId changes, Cfn will attempt to replace the resource and fail. There is no way I can see to make this change painless for existing users. That said, maybe we should invert the logic here: make things consistent by default (breaking existing users), but provide a feature flag which allows legacy behaviour so existing users can punt until vNext. That said, this isn't what's blocking us, and maybe we should split this into two separate PRs?

ahammond: I did that. The consistency stuff can be a separate PR. We just need compatibility to get us unblocked.

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1034,6 +1035,26 @@ export const FLAGS: Record<string, FlagInfo> = {
...
+
+  //////////////////////////////////////////////////////////////////////
+  [ALBV2_EXTERNALAPPLICATIONLISTENER_SWITCH_FROM_ADDTARGETGROUP_TO_ADDACTION]: {
+    type: FlagType.VisibleContext,
+    summary: 'When enabled, you can switch from the \`addTargetGroups()\` method of declaring a \`ListenerRule\` to the \`addAction()\` method, without changing the logicalId and replacing your resource.',
+    detailsMd: `
+      When switching from a less complex to a more complex use of ALB,
+      you will eventually need features not available in the \`addTargetGroups()\` convenience method.
+      In this case you will want to use the \`addAction()\` method.
+      Before this feature is enabled, switching over to \`addAction()\` from using \`addTargetGroups()\` will add a \`Rule\` suffix to the logicalId of your \`ListenerRule\`,
```
paulhcsun, line 1047:
```suggestion
      Before this feature is enabled, switching over to \`addAction()\` from using \`addTargetGroups()\` 
      will add a \`Rule\` suffix to the logicalId of your \`ListenerRule\`,
```

paulhcsun (commented):
Just adding a new line to keep the spacing consistent here.

paulhcsun (commented):
Just adding a new line to keep the spacing consistent here.

paulhcsun:
@Mergifyio update

scanlonp:
Hey @ahammond, saw that the recommended value for this feature flag is false. This makes me question if we need a feature flag or not. I see feature flags as a way to change default behavior while staying backwards compatible. If you only need this behavior some of the time, I would think setting a property is the way to go.

Is adding an optional prop to `AddApplicationActionProps` that controls how the suffix is generated possible?

ahammond:
@scanlonp I don't have a strong feeling either way here.

scanlonp:
Got it. I saw @TheRealAmazonKendra commented on your related PR. I think there has been a concerted effort to limit our feature flags, and properties looks like the best path here, so let's go with properties.

scanlonp:
Going to revert this specific PR, but I will be attentive to your two PRs so that we can get them in quickly for you.

ahammond:
@scanlonp to clarify, I don't care how it is solved, I care that it is solved. We are now 4 weeks since this fix was initially pushed. Regarding the props based PR I pushed on Fri Apr 5th, I addressed the comments on Monday Apr 8th in the morning. FYI, as a workaround we have bundled the entire `aws-cdk-lib` package into one of our internal libs and are using `patch-package` to apply a hack-around. This is incredibly fragile and I would like to get us back to stock ASAP.

Issue #29514, opened 2024-03-16
(aws-cloudwatch-actions): LambdaAction fails if added to multiple action types


morrijm4:
### Describe the bug

Adding the same lambda as the action for multiple status changes (alarm, ok, insufficient data) causes an error because of logical id conflicts.

### Expected Behavior

I expect to be able to add the same lambda for multiple action types on a CloudWatch alarm.

### Current Behavior

```
/Users/james.morrison/work/sales-domain/node_modules/.pnpm/constructs@10.3.0/node_modules/constructs/src/construct.ts:447
      throw new Error(`There is already a Construct with name '${childName}' in ${typeName}${name.length > 0 ? ' [' + name + ']' : ''}`);
            ^
Error: There is already a Construct with name 'ApiGatewayClientErrorAlarmAlarmPermission' in Import [SlackLambda]
    at Node.addChild (/Users/james.morrison/work/sales-domain/node_modules/.pnpm/constructs@10.3.0/node_modules/constructs/src/construct.ts:447:13)
    at new Node (/Users/james.morrison/work/sales-domain/node_modules/.pnpm/constructs@10.3.0/node_modules/constructs/src/construct.ts:71:17)
    at new Construct (/Users/james.morrison/work/sales-domain/node_modules/.pnpm/constructs@10.3.0/node_modules/constructs/src/construct.ts:499:17)
    at new CfnElement (/Users/james.morrison/work/sales-domain/node_modules/.pnpm/aws-cdk-lib@2.126.0_constructs@10.3.0/node_modules/aws-cdk-lib/core/lib/cfn-element.js:1:743)
    at new CfnRefElement (/Users/james.morrison/work/sales-domain/node_modules/.pnpm/aws-cdk-lib@2.126.0_constructs@10.3.0/node_modules/aws-cdk-lib/core/lib/cfn-element.js:2:823)
    at new CfnResource (/Users/james.morrison/work/sales-domain/node_modules/.pnpm/aws-cdk-lib@2.126.0_constructs@10.3.0/node_modules/aws-cdk-lib/core/lib/cfn-resource.js:1:1399)
    at new CfnPermission (/Users/james.morrison/work/sales-domain/node_modules/.pnpm/aws-cdk-lib@2.126.0_constructs@10.3.0/node_modules/aws-cdk-lib/aws-lambda/lib/lambda.generated.js:1:113187)
[...]

PR #29515, opened 2024-03-16
fix(cloudwatch-actions): `LambdaAction` fails if added to multiple action types

morrijm4:
Closes. #29514

### Reason for this change

Adding the same lambda as the action for multiple status changes (alarm, ok, insufficient data) causes an error because of logical id conflicts.

### Description of changes

Before adding the `lambda:InvokeFunction` permission to the lambda's resource policy, it checks to see if one already exists.

I considered not including this change under the `LAMBDA_PERMISSION_LOGICAL_ID_FOR_LAMBDA_ACTION` feature flag but, it breaks the `throws when multiple alarms are created for the same lambda if feature flag is set to false` test because it no longer throws. I understand that a major goal of the project is to keep behavior consistent however, it seems like it would be beneficial to fix an undesirable behavior without the need of configuring a feature flag.

This is my first contribution so I am new to this, could my change warrant its own feature flag?

Expanded upon existing unit tests. 

### Checklist

----


packages/aws-cdk-lib/aws-cloudwatch-actions/lib/lambda.ts
```diff
@@ -22,13 +22,17 @@ export class LambdaAction implements cloudwatch.IAlarmAction {
...
-    this.lambdaFunction.addPermission(`${idPrefix}AlarmPermission`, {
-      sourceAccount: Stack.of(scope).account,
-      action: 'lambda:InvokeFunction',
-      sourceArn: alarm.alarmArn,
-      principal: new iam.ServicePrincipal('lambda.alarms.cloudwatch.amazonaws.com'),
-    });
+    const flag = FeatureFlags.of(scope).isEnabled(LAMBDA_PERMISSION_LOGICAL_ID_FOR_LAMBDA_ACTION);
+    const idPrefix = flag ? alarm.node.id : '';
+    const permissionId = `${idPrefix}AlarmPermission`;
+    if (!this.lambdaFunction.permissionsNode.tryFindChild(permissionId) || !flag) {
```
lpizzinidev, line 28: Not sure I follow here 🤔

Why is the `|| !flag` necessary? 
From the feature flag definition:

> When enabled, the logical ID of a Lambda permission for a Lambda action includes an alarm ID.

Ie, it should just control how we define `permissionId`, not the flow of the function.

Also, how is this going to fix the mentioned bug?
With `if (!this.lambdaFunction.permissionsNode.tryFindChild(permissionId)` we're just ignoring the IDs that would cause an exception.
We wouldn't need a feature flag in that case, but we would simply skip adding lambda permissions without feedback to the user (with consequent potential issues).

lpizzinidev (changes requested):

packages/aws-cdk-lib/aws-cloudwatch-actions/lib/lambda.ts
morrijm4: Thanks for the review! I reverted the feature flag to only affect the logical ID and added a comparison with the permission source ARN and alarm ARN.

lpizzinidev: Thanks for the quick follow-up!
Now I get why you need the check on the feature flag:

> it breaks the throws when multiple alarms are created for the same lambda if feature flag is set to false test because it no longer throws.

I wonder if this fix is necessary. 
In what scenarios is useful to specify the same Lamdba for different actions on the same alarm? 🤔

morrijm4: The use case that stumbled me into this bug 😅 was I wanted to send a POST request from a lambda to a Slack webhook that would send a message in an alerts channel. I wanted to have updates on alert and ok to have proper observability.

lpizzinidev: I see. 
Just to understand the context fully:
Why use the same Lambda for different actions from the same alarm? 
Isn't it better to define a function for each action? (I guess you're probably defining the logic for handling the different actions in the function's code now)
My point is that we shouldn't fix the implementation to allow users to define non-optimal implementations.
But maybe I'm missing something, please let me know if that is the case.

morrijm4: I see your point. However, for my use case the only difference I have between the alarm and ok actions is the emoji that I send in the message lol. My thought is it really necessary to have two lambdas when it can be expressed as a ternary?

Additionally, should the CDK be a blocker for architectural decisions?

packages/aws-cdk-lib/aws-cloudwatch-actions/test/lambda.test.ts
```diff
@@ -173,6 +178,8 @@ def handler(event, context):
     handler: 'index.handler',
   });
   alarm1.addAlarmAction(new actions.LambdaAction(alarmLambda));
+  alarm1.addOkAction(new actions.LambdaAction(alarmLambda));
+  alarm1.addInsufficientDataAction(new actions.LambdaAction(alarmLambda));
```
lpizzinidev, line 182:
```suggestion
  expect(() => {
    alarm1.addOkAction(new actions.LambdaAction(alarmLambda));
  }).toThrow(/There is already a Construct with name 'AlarmPermission' in Function \[alarmLambda\]/);
```
We should define a dedicated test case for this scenario.
`throws when multiple actions are created for the same alarm and lambda if feature flag is set to false`

lpizzinidev (changes requested):
https://github.com/aws/aws-cdk/pull/29515#discussion_r1527586824

> should the CDK be a blocker for architectural decisions

Nope, but I think it should help enforce good development practices.
Anyway, it seems possible to define the same Lambda for different actions on the same alarm from the Console so I guess it should be allowed here as well.

I left a note to adjust the test cases and to add some comments on the new condition.
Also, can you please fix the title to not be truncated?

Thanks for your quick follow-ups and clarifications 👍🙂

morrijm4:
@lpizzinidev Just to clarify, you want me to change it back to my original implementation with it behind the feature flag? Would you mind elaborating why you prefer that over the existing one?

lpizzinidev:
@morrijm4 Will the current one throw with the feature flag disabled?

morrijm4:
@lpizzinidev With the feature flag disabled it will throw if the same lambda is added to multiple alarms because it will try to add multiple permissions with the same `'AlarmPermission'` logical ID. If the same lambda is added to different actions for one alarm it will not throw.

packages/aws-cdk-lib/aws-cloudwatch-actions/test/lambda.test.ts
lpizzinidev, line 1: Can you also add coverage for adding the same Lambda for the same action on the same alarm?
Eg:
```ts
alarm1.addOkAction(new actions.LambdaAction(alarmLambda));
alarm1.addOkAction(new actions.LambdaAction(alarmLambda));
```
(Should work under feature flag, throw without)

packages/aws-cdk-lib/aws-cloudwatch-actions/lib/lambda.ts
```diff
@@ -23,12 +23,17 @@ export class LambdaAction implements cloudwatch.IAlarmAction {
...
-    this.lambdaFunction.addPermission(`${idPrefix}AlarmPermission`, {
-      sourceAccount: Stack.of(scope).account,
-      action: 'lambda:InvokeFunction',
-      sourceArn: alarm.alarmArn,
-      principal: new iam.ServicePrincipal('lambda.alarms.cloudwatch.amazonaws.com'),
-    });
+    const permissionId = `${idPrefix}AlarmPermission`;
+    const permissionNode = this.lambdaFunction.permissionsNode.tryFindChild(permissionId) as lambda.CfnPermission | undefined;
+
+    if (permissionNode?.sourceArn !== alarm.alarmArn) {
```
lpizzinidev, line 29:
```suggestion
    // If the Lambda permission has already been added to this function
    // and the feature flag is disabled we skip adding it to avoid an exception being thrown
    // see https://github.com/aws/aws-cdk/issues/29514
    if (permissionNode?.sourceArn !== alarm.alarmArn) {
```

lpizzinidev (changes requested):
https://github.com/aws/aws-cdk/pull/29515#issuecomment-2002613289 Makes sense (unless I'm missing something).
Thanks for clarifying 👍
Left some comments for test coverage and documentation

packages/aws-cdk-lib/aws-cloudwatch-actions/test/lambda.test.ts
morrijm4: Yes! However, it will work with or without the feature flag.

packages/aws-cdk-lib/aws-cloudwatch-actions/lib/lambda.ts
morrijm4: Added the comment but modified it because we will avoid the exception with the feature flag enabled or disabled.

packages/aws-cdk-lib/aws-cloudwatch-actions/test/lambda.test.ts
morrijm4: I'm a little confused about this comment, would you mind helping me understand what you are looking for? The line `alarm1.addOkAction(new actions.LambdaAction(alarmLambda));` will not throw in this case with the current implementation.

packages/aws-cdk-lib/aws-cloudwatch-actions/lib/lambda.ts
lpizzinidev: Hey ✋ Sorry for the late reply.
I wanted to go through it a bit more to clarify some doubts.

I think that what we wanna do here is keep the behavior consistent when not under FF.
Ie, it should throw with these scenarios:
1. Same alarm, different action, same lambda
```ts
alarm1.addAlarmAction(new actions.LambdaAction(alarmLambda));
alarm1.addOkAction(new actions.LambdaAction(alarmLambda));
```
2. Same alarm, same action, same lambda
```ts
alarm.addAlarmAction(new actions.LambdaAction(alarmLambda));
alarm.addAlarmAction(new actions.LambdaAction(alarmLambda));
```
3. Different alarm, different action, same lambda
```ts
alarm1.addAlarmAction(new actions.LambdaAction(alarmLambda));
alarm2.addAlarmAction(new actions.LambdaAction(alarmLambda));
```
That's because we try to create the same `AlarmPermission` construct.

With FF enabled we should address cases 1 and 2 (since 3 is already handled), that's why I suggested we should use the condition from the first commit as the other approach would be adding a second FF to add another prefix to the construct name (seems cumbersome).

Let me know if you have any doubts or comments.

Thanks.

lpizzinidev: Hey ✋
Sorry for the late response, needed a bit of free time to go through it.
Anyway, thanks for the changes. 
I think you made valid points with your responses and the PR looks good to me 👍

lpizzinidev (approved):

packages/aws-cdk-lib/aws-cloudwatch-actions/lib/lambda.ts
morrijm4: No worries! Appreciate you taking the time to go through it all and thank you for the feedback.

GavinZZ (approved):
Perfect, thanks for fixing it!

morrijm4:
@mergifyio refresh

morrijm4:
Hey @GavinZZ, one of the mergify actions failed because it said it did not have update permissions to merge this PR. I verified I checked the "Allow edits by maintainers" box so I wanted to refresh the action. As you can see, I tried the refresh comment command and then updating the branch.

Sorry for ignoring the clear message by the mergify bot. I see why it is there now...

Do you have any other troubleshooting advice? Tagging @lpizzinidev too. Thanks for the help!

GavinZZ:
No worry, will update the branch and retry the mergify process. Will monitor it and retry if any of the actions failed.

GavinZZ (approved):

Flavioli:
Is this actually resolved? I'm still running into this issue. With the flag `@aws-cdk/aws-cloudwatch-actions:changeLambdaPermissionLogicalIdForLambdaAction` I can use the same `LambdaAction` on multiple alarms, but I still cannot use the same `LambdaAction` on both `addAlarmAction` and `addOkAction` of the same alarm.

A workaround is to duplicate the alarm just for the purpose of having a different alarm for each of `addAlarmAction` and  `addOkAction` with the same lambda, or two use two different lambdas, one for each alarm action type. Neither scenario is ideal. But on my end this particular issue still seems to be present.

jamescarter-le:
Agree, it is annoying we cannot disable this functionality.

I have permissions for any alarm to call my reporting Lambda, and want to create many Alarms via the CDK, without them trying to create their own permissions.

I grant all my CloudWatch alarms permission to invoke my Lambda, in the C# CDK.

```
function.AddPermission("InvokeByCloudWatch", new Permission
{
    Principal = new ServicePrincipal("lambda.alarms.cloudwatch.amazonaws.com"),
    Action = "lambda:InvokeFunction",
    SourceArn = $"arn:aws:cloudwatch:{this.Region}:{this.Account}:alarm:*"
});
```

Now when calling AddAlarmAction, it should not try and create the Permissions, or we should have a parameter to prevent this behavior.

`alarm.AddAlarmAction(new LambdaAction(AlarmReportingFunction));`

Because the Alarm tries to add permissions, when attaching all my Lambdas clashes with the Permission being added. I do not want to add the AlarmName to the Action Permission using the existing feature flag, because it means I cannot attach the same Function to both AlarmAction and OkAction.

` Error: There is already a Construct with name 'AlarmPermission' in Function [Lambda_CloudWatchAlarms]`

ssimmons806:
I am getting this error when trying to synthesize the stack in tests via `Template.fromStack(stack)`. I can `npx cdk synth` and deploy it fine.

```
    There is already a Construct with name 'AlarmPermission' in NodejsFunction [Fn]

      43 |     })
      44 |
    > 45 |     alarm.addAlarmAction(action)
```

Is there a way to enable this feature flag in tests?

Issue #29443, opened 2024-03-11
CDK CLI: execute deployment on CONFIGURATION_COMPLETE event


evgenyka:
### Describe the feature

When creating a CloudFormation stack, a resource can sometimes take longer to provision, making it appear as if it’s stuck in an IN_PROGRESS state. This can be because CloudFormation is waiting for the resource to complete consistency checks during its resource stabilization step. To improve visibility into resource provisioning status, CloudFormation has introduced a new “[CONFIGURATION_COMPLETE](https://aws.amazon.com/blogs/devops/how-we-sped-up-aws-cloudformation-deployments-with-optimistic-stabilization/)” event. This event is emitted at both the individual resource level and the overall stack level during the create workflow when resource(s) creation or configuration is complete, but stabilization is still in progress.
CDK can use this new event and speed up the deployment time. 

### Use Case

I'm annoyed by how long it takes to deploy or redeploy my CDK app while I'm developing and testing. I can speed it up by taking a risk with CloudFormation's stabilization phase, but it's not safe for production.

### Proposed Solution

Propose new CDK CLI command or additional switch to `cdk deploy` to execute a new optimistic stabilization strategy to shorten stack deployment times. In this mode, CDK will continue the deployment sequence using the new `CONFIGURATION_COMPLETE` event and not wait for `CREATE_COMPLETE`. 

### Other Information

There's a chance that resource creation might fail during the stabilization phase, which occurs between the `CONFIGURATION_COMPLETE` event and the `CREATE_COMPLETE` event. This shouldn't be the default deployment mode, and users should be warned about the risk. CDK deploy should be capable of rolling back in case of late-arriving resource creation failures.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.132.0

### Environment details (OS name and version, etc.)

MACOS

comcalvi:
The CDK should not take this feature. Here's a few reasons why:

[...]

PR #29536, opened 2024-03-19
feat(cli): support optimistic stabilization strategy(--exit-on-config-complete)

pahud:
This PR adds an `--exit-on-config-complete` flag for `cdk deploy`. When this flag is enabled, CDK will monitor the `DetailedStatus`of the stack being deployed, and return success when the stack enters the `CONFIGURATION_COMPLETE` state.

This [optimistic stabilization strategy](https://aws.amazon.com/blogs/devops/how-we-sped-up-aws-cloudformation-deployments-with-optimistic-stabilization/) can help reduce the total deployment time, especially when working with multiple stacks that have dependencies on each other. Normally, the CDK would wait for the entire deployment process to complete before returning, even if individual stacks had finished. With the `--exit-on-config-complete` flag, CDK will return as soon as each stack is in the `CONFIGURATION_COMPLETE` state, allowing subsequent dependent stacks to continue their deployment immediately. This can lead to significant time savings for complex CDK applications with many interdependent stacks.

### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29443

### Description of changes

When `--exit-on-config-complete` is provided with `cdk deploy`, `waitForStackDeploy()` monitors the DetailedStatus for `CONFIGURATION_COMPLETE` and returns success upon it. 

### Checklist

----

### DIFF
--- a/packages/aws-cdk/README.md
+++ b/packages/aws-cdk/README.md
@@ -482,7 +482,7 @@ For more details on this, you can refer to the blog post at
 
 The `--exit-on-config-complete` flag is supported by `cdk deploy` only.
 
-**⚠ Note **: This command is considered experimental,
+**Note**: This command is considered experimental,
 and might have breaking changes in the future.
 
 ### `cdk watch`


TheRealAmazonKendra:
@Mergifyio update

comcalvi:
This integration test is failing:

```
● deploy with optimistic stabilization
--


Test Suites: 1 failed, 2 passed, 3 total
```

Could you fix the failure? I can't find the actual error in the logs, it just says "exited with error code 1"

comcalvi (commented):
This integration test is failing:
```
● deploy with optimistic stabilization
--


Test Suites: 1 failed, 2 passed, 3 total
```
Could you fix the failure? I can't find the actual error in the logs, it just says "exited with error code 1"

pahud:
I will fix this next week. :-)

ericzbeard (changes requested):
The docs for the PR make it sound like we need this to get the benefits of optimistic stabilization for resources within a single stack. We need to clarify this, because optimistic stabilization is something that happens in the CloudFormation back end, and does not require any configuration or opt-in to get the benefits. If I understand this PR correctly, it is concerned with dependencies between different stacks within a CDK application. Let's make that very clear.

pahud:
[INTEG TEST::deploy with optimistic stabilization] Starting (pid 60649)...
[INTEG TEST::deploy with optimistic stabilization] Done (183282 ms).
 PASS  packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.js (185.406 s)
 
 The cli-integ-tests should be good now.

packages/aws-cdk/lib/cli.ts
```diff
@@ -255,7 +256,8 @@ async function parseCommandLineArguments(args: string[]) {
     .command('destroy [STACKS..]', 'Destroy the stack(s) named STACKS', (yargs: Argv) => yargs
```
bergjaak, line 256: would it make sense to also add an option for `cdk watch`??

packages/aws-cdk/lib/cli.ts
```diff
@@ -604,6 +606,7 @@ export async function exec(args: string[], synthesizer?: Synthesizer): Promise<n
           assetParallelism: configuration.settings.get(['assetParallelism']),
           assetBuildTime: configuration.settings.get(['assetPrebuild']) ? AssetBuildTime.ALL_BEFORE_DEPLOY : AssetBuildTime.JUST_IN_TIME,
           ignoreNoStacks: args.ignoreNoStacks,
+          exitOnConfigComplete: args.exitOnConfigComplete,
```
bergjaak, line 609: on line 658 of this file, for `case 'destroy'`. do we also need to pass this exitOnConfigComplete? What about `case: 'watch'`?

packages/aws-cdk/lib/cli.ts
```diff
@@ -255,7 +256,8 @@ async function parseCommandLineArguments(args: string[]) {
     .command('destroy [STACKS..]', 'Destroy the stack(s) named STACKS', (yargs: Argv) => yargs
       .option('all', { type: 'boolean', default: false, desc: 'Destroy all available stacks' })
       .option('exclusively', { type: 'boolean', alias: 'e', desc: 'Only destroy requested stacks, don\'t include dependees' })
-      .option('force', { type: 'boolean', alias: 'f', desc: 'Do not ask for confirmation before destroying the stacks' }))
+      .option('force', { type: 'boolean', alias: 'f', desc: 'Do not ask for confirmation before destroying the stacks' })
+      .option('exit-on-config-complete', { type: 'boolean', desc: 'Whether to exit on configuration complete', default: false }))
```
bergjaak, line 260: Is this being ignored?  I only see support for `deploy`? And if this isn't being ignored, can we add an integration test for `destroy` case?

packages/aws-cdk/lib/api/util/cloudformation/stack-status.ts
```diff
@@ -50,3 +50,28 @@ export class StackStatus {
     return this.name + (this.reason ? ` (${this.reason})` : '');
   }
 }
+
+/**
+ * A utility class to inspect CloudFormation stack detailed statuses.
+ *
+ * @see https://docs.aws.amazon.com/cli/latest/reference/cloudformation/describe-stacks.html
+ */
+export class StackDetailedStatus {
```
bergjaak, line 59: Rather than make a new class, would it make more sense to have this be part of the existing StackStatus class?

And it would be great to add unit tests for these new functions :)

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
```diff
@@ -417,6 +417,36 @@ class LambdaHotswapStack extends cdk.Stack {
...
+      maxAzs: 1,
+      subnetConfiguration: [
+        {
+          cidrMask: 24,
+          name: 'Public',
+          subnetType: ec2.SubnetType.PUBLIC,
+        },
+      ],
+    });
+
```
bergjaak, line 435: A high level question -- what if a resource fails deploying after CONFIGURATION_COMPLETE status is emitted? Do we handle that case? Is that a possible case? If it is a possible case, could we have an integration test for that?

packages/aws-cdk/lib/api/deployments.ts
```diff
@@ -427,6 +436,7 @@ export class Deployments {
       deployName: options.deployName,
       quiet: options.quiet,
       ci: options.ci,
+      exitOnConfigComplete: options.exitOnConfigComplete,
```
bergjaak, line 439: Where is this set? Is this unset for the delete case?

packages/aws-cdk/lib/cli.ts
pahud: Correct me if I am wrong but AFAIK `cdk watch` essentially monitors a stack and update it through SDK calls without redeploying it. I think `cdk watch` would not monitor the stack status so I can't see how this would fit in this scenario. I might be wrong though.

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
pahud: Will discuss this with the CFN team.

bergjaak: My thinking is,  if that happens, we should fail with exit status 1. One idea, without much thinking about how this would work, is to poll the statuses of all the stacks once the final stack has entered CONFIG_COMPLETE. So then it's just that the final stack of the deployment is not getting the speedup... which I think is fine since you then get the speed up on all the intermediate stacks.

I am considering if there is a case where a customer kicks off some sort of production thing as soon as the customer get a success signal from CDK deploy? If so, then we'd want to make sure that everything did successfully provision... unless CloudFormation handles that, somehow. I'm not too familiar with this new optimistic stabilization feature from CloudFormation.

packages/aws-cdk/lib/cli.ts
bergjaak: Ah, I thinkk we're both right. It's an options to turn off the --hot-swap for `watch` https://docs.aws.amazon.com/cdk/v2/guide/cli.html#cli-deploy.

This comment is just a suggestion, though. I would approve this PR with our without watch support, since that could be added in the future, if there was a demand for it

packages/aws-cdk/lib/cli.ts
pahud: Just confirmed the optimistic strategy from CFN does not support deleting stack so I will remove that from `destroy`. We'll only support `deploy`.

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
pahud: We probably should figure out, let's say if we have 2 stacks to deploy with Stack2 depends on Stack1:

Stack1 enters `CC` state -> CLI starts deploying Stack2 -> Stack2 enters `CC` state -> CLI exit 0 but Stack1 ended up with final status failed. Now we get boost from Stack1 and Stack2 but Stack1 is actually failed. What should we do?

bergjaak: My preconception (which could be wrong) is that stack 2 would fail if stack 1 eventually failed. But if the scenario you're describing is true, then that biases me in favor of including a warning about this not being for production, and that would be all we'd do for addressing the failure case. So, similar to how we advise users to not use hotswap in production, we'd advise users to not use this flag in production. What do you think of that? I personally feel satisfied with that.

pahud: I'll check with the CFN team:

-  Will stack2 fail when stack1 fails with `CC` status
a. if stack2 would fail then we just need to check the status of the last stack
b. if stack2 would NOT fail we will end up with a failed stack1 and a deployed stack2.

ridiss1: Yes! I agree with @bergjaak to warn customer to not use the flag in production. This is what CFN official user guide mentions. 

Stack 2 will not fail automatically. unless CDK implement a pulling mechanism that read the state an fail the stack based on the stack 1 state.

packages/aws-cdk/lib/api/deployments.ts
pahud: As this feature does not support destroyStack. I just removed this.

packages/aws-cdk/lib/cli.ts
pahud: This feature does not support destroy stacks so I removed all relevant code about the `destroy` case.

packages/aws-cdk/lib/api/util/cloudformation/stack-status.ts
pahud: DONE. Just refactored the existing class and add some unit tests.

comcalvi:
I've been asked to pick this up, so I'm closing this PR and opening a new one.

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
```diff
@@ -444,6 +444,36 @@ class LambdaHotswapStack extends cdk.Stack {
   }
 }
 
+class DetailedStatusStack extends cdk.Stack{
```
comcalvi, line 447: none of these resources currently cause the stack itself to emit the `CONFIGURATION_COMPLETE` event, so this test needs to be updated. The only resource I know of that does make it emit this event is ECS Services

pahud:
Update:

CDK can't support this feature at this moment because CFN would not emit the outputs in `CONFIGURATION_COMPLETE` phase and dependent stacks would fail if they reference the outputs from there.

We will evaluate this again when CFN has this support.

PR #29542, opened 2024-03-19
fix(sns): contentBasedDeduplication is always false for imported topic

### LINKED ISSUE #29532: sns.Topic.fromTopicArn: importinig topic forces contentBasedDeduplication = false
### Describe the bug

When importing a topic by Arn, the constructor sets contentBasedDeduplication = false. 

This means that when we call the resource method getContentBasedDeduplication(), it returns false always.

### Expected Behavior

I would expect that when the resource is imported, the ARN is checked to determine whether or not contentBasedDeduplication is enabled on the Topic, then return a result that contains this.

### Current Behavior

Currently .getContentBasedDeduplication() always returns false, regardless of the configuration of the actual Topic imported. This is only for imported Topics. Newly created topics return the actual value of the property. 

### Reproduction Steps

JAVA code:

        // Existing Topic with ContentBasedDeduplication enabled
        final ITopic assetTopic = Topic.fromTopicArn(this, "asset topic","arn:aws:sns:us-east-1:123456789012:myTopic.fifo");
        // returns false
        System.out.println(assetTopic.getContentBasedDeduplication());
        // New Topic created with ContentBasedDeduplication enabled
        Topic newTopic = Topic.Builder.create(this,"new topic")
                .contentBasedDeduplication(true)
                .displayName("my new topic")
                .fifo(true)
                .build();
        //Returns true, correctly
        System.out.println(newTopic.getContentBasedDeduplication());    

### Possible Solution

_No response_

See linked commit for when the property was set. Not sure if there is possibly an underlying reason why this value needs to be false for imports, if so please let me know.

### CDK CLI Version

2.131.0

### Framework Version

_No response_

### Node.js Version

NA

### OS

Win11

### Language

Java

### Language Version

openjdk version "21.0.2" 2024-01-16 LTS

### Other information

_No response_

msambol:
Closes #29532.

----

### DIFF
--- a/packages/aws-cdk/lib/api/deployments.ts
+++ b/packages/aws-cdk/lib/api/deployments.ts
@@ -265,6 +265,7 @@ export interface DestroyStackOptions {
 export interface StackExistsOptions {
   stack: cxapi.CloudFormationStackArtifact;
   deployName?: string;
+  tryLookupRole?: boolean;
 }
 
 export interface DeploymentsProps {
@@ -430,7 +431,12 @@ export class Deployments {
   }
 
   public async stackExists(options: StackExistsOptions): Promise<boolean> {
-    const { stackSdk } = await this.prepareSdkFor(options.stack, undefined, Mode.ForReading);
+    let stackSdk;
+    if (options.tryLookupRole) {
+      stackSdk = (await this.prepareSdkWithLookupOrDeployRole(options.stack)).stackSdk;
+    } else {
+      stackSdk = (await this.prepareSdkFor(options.stack, undefined, Mode.ForReading)).stackSdk;
+    }
     const stack = await CloudFormationStack.lookup(stackSdk.cloudFormation(), options.deployName ?? options.stack.stackName);
     return stack.exists;
   }
--- a/packages/aws-cdk/lib/cdk-toolkit.ts
+++ b/packages/aws-cdk/lib/cdk-toolkit.ts
@@ -139,10 +139,19 @@ export class CdkToolkit {
       let changeSet = undefined;
 
       if (options.changeSet) {
-        const stackExists = await this.props.deployments.stackExists({
-          stack: stacks.firstStack,
-          deployName: stacks.firstStack.stackName,
-        });
+        let stackExists = false;
+        try {
+          stackExists = await this.props.deployments.stackExists({
+            stack: stacks.firstStack,
+            deployName: stacks.firstStack.stackName,
+            tryLookupRole: true,
+          });
+        } catch (e: any) {
+          debug(e.message);
+          stream.write('Checking if the stack exists before creating the changeset has failed, will base the diff on template differences (run again with -v to see the reason)\n');
+          stackExists = false;
+        }
+
         if (stackExists) {
           changeSet = await createDiffChangeSet({
             stack: stacks.firstStack,
@@ -154,7 +163,7 @@ export class CdkToolkit {
             stream,
           });
         } else {
-          debug(`the stack '${stacks.firstStack.stackName}' has not been deployed to CloudFormation, skipping changeset creation.`);
+          debug(`the stack '${stacks.firstStack.stackName}' has not been deployed to CloudFormation or describeStacks call failed, skipping changeset creation.`);
         }
       }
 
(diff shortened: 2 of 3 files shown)


packages/aws-cdk-lib/aws-sns/lib/topic.ts
```diff
@@ -154,13 +154,14 @@ export class Topic extends TopicBase {
...
+   * @param contentBasedDeduplication If content-based deduplication is enabled
    */
-  public static fromTopicArn(scope: Construct, id: string, topicArn: string): ITopic {
+  public static fromTopicArn(scope: Construct, id: string, topicArn: string, contentBasedDeduplication?: boolean): ITopic {
     class Import extends TopicBase {
       public readonly topicArn = topicArn;
       public readonly topicName = Stack.of(scope).splitArn(topicArn, ArnFormat.NO_RESOURCE_NAME).resource;
       public readonly fifo = this.topicName.endsWith('.fifo');
-      public readonly contentBasedDeduplication = false;
+      public readonly contentBasedDeduplication = contentBasedDeduplication ?? false;
```
msambol, line 164: keep default `false` for backwards compatibility.

msambol:
Exemption request: I don't believe an integration test is needed here.

packages/aws-cdk-lib/aws-sns/lib/topic.ts
```diff
@@ -154,13 +154,14 @@ export class Topic extends TopicBase {
    * @param scope The parent creating construct
    * @param id The construct's name
    * @param topicArn topic ARN (i.e. arn:aws:sns:us-east-2:444455556666:MyTopic)
+   * @param contentBasedDeduplication If content-based deduplication is enabled
    */
-  public static fromTopicArn(scope: Construct, id: string, topicArn: string): ITopic {
+  public static fromTopicArn(scope: Construct, id: string, topicArn: string, contentBasedDeduplication?: boolean): ITopic {
```
lpizzinidev, line 159: 🤔 Not sure about this approach.
Do we have a way to detect if `contentBasedDeduplication` is enabled from the `topicArn`?
If not, I'd prefer to create a separate `fromTopicAttributes` method to address the use case as done in other constructs.

lpizzinidev (changes requested):

packages/aws-cdk-lib/aws-sns/lib/topic.ts
msambol: Yeah let me fix this....

msambol:
@lpizzinidev This should be better now..

lpizzinidev (changes requested):
Thanks for the changes 👍
Left some comments for further adjustments.
Also, can you please add integration test coverage for the `Topic.fromTopicArn` and `Topic.fromTopicAttributes` methods?

msambol:
@lpizzinidev This should be better now.

packages/@aws-cdk-testing/framework-integ/test/aws-sns/test/integ.sns.ts
```diff
@@ -40,6 +40,9 @@ class SNSInteg extends Stack {
       successFeedbackRole: feedbackRole,
       successFeedbackSampleRate: 50,
     });
+
+    // Can import topic from arn
+    Topic.fromTopicArn(this, 'ImportedTopic', topic.topicArn);
```
lpizzinidev, line 45:
```suggestion
    const topic2 = new Topic(this, 'MyTopic2', {
      topicName: 'fooTopic2',
      displayName: 'fooDisplayName2',
      masterKey: key,
    });
    const importedTopic = Topic.fromTopicArn(this, 'ImportedTopic', topic2.topicArn);
    
    importedTopic.addLoggingConfig({
      protocol: LoggingProtocol.HTTP,
      failureFeedbackRole: feedbackRole,
      successFeedbackRole: feedbackRole,
      successFeedbackSampleRate: 50,
    });
```
Can you please add an extra topic so that the snapshot will get updated and the pipeline will be green? 
Also, since `fromTopicArn` is calling `fromTopicAttributes`, I think you can remove the change in [integ.sns-fifo.ts](https://github.com/aws/aws-cdk/pull/29542/files#diff-bd9737da387a76c89a9eeb2d47667280f85b2dfb8435ccae19dd8653a44365af)

lpizzinidev (approved):
Thanks 👍
Looks good to me, just a final adjustment on the integration tests is needed in my opinion

shikha372:
Thanks @msambol for your contribution, as per my understanding content based deduplication is available only for the FIFO queue, so if we are allowing users to set this value for imported topics, they might mistakenly set it for standard queue and face issue during deployment. Could we add validation that this is being set from attributes only in case in fifo queue.

msambol:
> Thank you @msambol for your contribution, as per my understanding content based deduplication is available only for the FIFO queue, so if we are allowing users to set this value for imported topics, they might mistakenly set it for standard queue and face issue during deployment. Could we add validation that this is being set from attributes only in case in fifo queue.

Great catch! Updated.

shikha372 (approved):
LGTM .. Thank you @msambol for contributing and @lpizzinidev for review

PR #29563, opened 2024-03-20
fix(ec2): global vpc endpoint support

### LINKED ISSUE #29560: ec2: Missing and invalid global VPC endpoints
We also have an issue with global endpoints, e.g. [`S3_MULTI_REGION_ACCESS_POINTS`](https://github.com/aws/aws-cdk/blob/0fee99b473c17210a156d74263227cdf3274d253/packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts#L433). They are not supposed to have a region prefix ([docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/MultiRegionAccessPointsPrivateLink.html)), but currently do in the CDK:

```sh
$ aws ec2 describe-vpc-endpoint-services --region=us-east-1 --service-names=com.amazonaws.s3-global.accesspoint | jq '.ServiceDetails[] | .ServiceName'

"com.amazonaws.s3-global.accesspoint"
```

```ts
new CfnOutput(this, "endpoint", {
	value: ec2.InterfaceVpcEndpointAwsService.S3_MULTI_REGION_ACCESS_POINTS.name,
});

// TestDeployStack.endpoint = com.amazonaws.eu-west-1.s3-global.accesspoint
```

The region is currently always prefixed: https://github.com/aws/aws-cdk/blob/0fee99b473c17210a156d74263227cdf3274d253/packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts#L510

I haven't checked if there are other existing cases, but `aws.api.global.codecatalyst` is currently missing from the endpoint list, and will run into the same issue ([docs](https://docs.aws.amazon.com/codecatalyst/latest/adminguide/managing-vpcs.endpoint.html#managing-vpcs.endpoint-service-names))

_Originally posted by @nmussy in https://github.com/aws/aws-cdk/issues/29524#issuecomment-2009783493_

nmussy:
### Issue # (if applicable)

Closes #29560.

### Reason for this change

The service name generated by the CDK for global VPC endpoints was incorrect, as it contained the stack's region:

```sh
$ aws ec2 describe-vpc-endpoint-services --region=us-east-1 --service-names=com.amazonaws.s3-global.accesspoint | jq '.ServiceDetails[] | .ServiceName'

"com.amazonaws.s3-global.accesspoint"
```

```ts
new CfnOutput(this, "endpoint", {
	value: ec2.InterfaceVpcEndpointAwsService.S3_MULTI_REGION_ACCESS_POINTS.name,
});

// TestDeployStack.endpoint = com.amazonaws.eu-west-1.s3-global.accesspoint
```

In addition, another global endpoint was missing from `InterfaceVpcEndpointAwsService`.

### Description of changes

* The `InterfaceVpcEndpointAwsService` constructor was modified to
  * I would have preferred to switch to a single object for the optional constructor properties (`prefix`, `port`, and now `global`), but couldn't make a breaking change to a publicly accessible constructor
* `InterfaceVpcEndpointAwsService.S3_MULTI_REGION_ACCESS_POINTS` was changed to be a global VPC endpoint
* `InterfaceVpcEndpointAwsService.CODECATALYST` was added

I've added a unit test to check that the global endpoints' name were set correctly.

I also added an integration test for `InterfaceVpcEndpointAwsService.S3_MULTI_REGION_ACCESS_POINTS`.
To test it, I created a publicly accessible EC2 instance on the VPC, connected to it and ran `nslookup accesspoint.s3-global.amazonaws.com` to make sure it was resolvable (see [Configuring a Multi-Region Access Point for use with AWS PrivateLink](https://docs.aws.amazon.com/AmazonS3/latest/userguide/MultiRegionAccessPointsPrivateLink.html)):

Without the `InterfaceVpcEndpointAwsService.S3_MULTI_REGION_ACCESS_POINTS` interface endpoint:
```sh
$ nslookup accesspoint.s3-global.amazonaws.com
Server:         10.0.0.2
Address:        10.0.0.2#53

Non-authoritative answer:
*** Can't find accesspoint.s3-global.amazonaws.com: No answer
```

With the `InterfaceVpcEndpointAwsService.S3_MULTI_REGION_ACCESS_POINTS` interface endpoint:
```sh
$ nslookup accesspoint.s3-global.amazonaws.com
Server:         10.0.0.2
Address:        10.0.0.2#53

Non-authoritative answer:
Name:   accesspoint.s3-global.amazonaws.com
Address: 10.0.156.75
Name:   accesspoint.s3-global.amazonaws.com
Address: 10.0.246.83
```

### Checklist

----


msambol:
@nmussy want to rebase this and i'll take a look?

nmussy:
@msambol Should be good to go, thanks 👍

msambol (approved):
LGTM

packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts
```diff
@@ -553,10 +554,10 @@ export class InterfaceVpcEndpointAwsService implements IInterfaceVpcEndpointServ
    */
   public readonly privateDnsDefault?: boolean = true;
 
-  constructor(name: string, prefix?: string, port?: number) {
-    const region = Lazy.uncachedString({
+  constructor(name: string, prefix?: string, port?: number, global = false) {
```
msambol, line 557: actually, maybe we want this to be `props: InterfaceVpcEndpointProps` in case we need to add more in the future?

nmussy: Yeah, I mentioned it in the PR: 
> I would have preferred to switch to a single object for the optional constructor properties (prefix, port, and now global), but couldn't make a breaking change to a publicly accessible constructor

I can put `global` in a `props` object, but we'll still have `name, prefix?, port?, props?`. Might be better late than never?

msambol: Yeah, exactly: `name, prefix?, port?, props?`. Especially since global has a default value, will make it difficult to change in the future.

nmussy: Done 👍

packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts
```diff
@@ -553,10 +567,15 @@ export class InterfaceVpcEndpointAwsService implements IInterfaceVpcEndpointServ
    */
   public readonly privateDnsDefault?: boolean = true;
 
-  constructor(name: string, prefix?: string, port?: number) {
-    const region = Lazy.uncachedString({
+  constructor(
+    name: string,
+    prefix?: string,
+    port?: number,
+    { global = false }: InterfaceVpcEndpointAwsServiceProps = {},
```
msambol, line 574: possible to do `props?: InterfaceVpcEndpointAwsServiceProps` ? Easier to read ?

nmussy: Sure, done 👍

msambol: forget to stage?

nmussy: No, forgot to push this time 😅

msambol (approved):
NIce! Thanks.

packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts
```diff
@@ -570,7 +589,7 @@ export class InterfaceVpcEndpointAwsService implements IInterfaceVpcEndpointServ
       },
     });
 
-    this.name = `${prefix || defaultEndpointPrefix}.${region}.${name}${defaultEndpointSuffix}`;
+    this.name = `${prefix || defaultEndpointPrefix}.${regionPrefix}${name}${defaultEndpointSuffix}`;
```
aaythapa, line 592: Nit: adding `+ '.'` in line 578 makes this line look a little weird imo. There is a `.` in front of `${regionPrefix}` but the suffix `.` is added above. Either both `.`'s should be added when you make `regionPrefix` or both should be added here (i prefer here). Unless theres another reason you moved the suffix `.`?

aaythapa (commented):
Thanks for making this change! Just one nit and question

aaythapa:
> I would have preferred to switch to a single object for the optional constructor properties (prefix, port, and now global), but couldn't make a breaking change to a publicly accessible constructor

 Not sure I understand this part of the description, could you elaborate? Did you want to add prefix, port and global as one props like object?

packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts
nmussy: The issue is that if the endpoint is global, there will be no region. If I'd kept the region variable, we would have had two `.` between the prefix and the service name. That's why I renamed it to `regionPrefix` and gave it the `.` if it wasn't global. I can revert back to having a `region` variable and only append the `.` if it's not empty/not global, but I would be testing the same thing twice then. Another possibility would be to switch to a `[].filter.join('.')` pattern, it would be more readable imo but a bit excessive given the single modifying condition

nmussy:
> Did you want to add prefix, port and global as one props like object?

Ideally, yes. The Props interface we ended up with still leaves us with `undefined` args before it, but changing that would cause a breaking change. We could have a factory function I suppose, but I would require storing the `global` property to allow the constructor to access it.

aaythapa (approved):

msambol:
@Mergifyio refresh

nmussy:
@mergifyio refresh

nmussy:
@Mergifyio refresh

PR #29568, opened 2024-03-21
chore(integ-test-alpha): add log group retention days to waiter function

sugardon:
### Issue # (if applicable)

None
(related to #29277)

### Reason for this change

When call  `waitForAssertions()`  the assertion provider creates CloudWatch log groups, but set to never expire

### Description of changes

Added log retention days to assertion provider in `WaiterStateMachine`

Added unit test

### Checklist

----


packages/@aws-cdk/integ-tests-alpha/lib/assertions/waiter-state-machine.ts
```diff
@@ -33,6 +34,13 @@ export interface WaiterStateMachineOptions {
    * @default 1 (no backoff)
    */
   readonly backoffRate?: number;
+
+  /**
+   * How long, in days, the log contents will be retained.
+   *
+   * @default - no retention days specified
+   */
+  readonly logRetention?: RetentionDays;
```
msambol, line 43: for the maintainers to decide but do we actually want to set a reasonable default here? maybe 3 days or 1 week.

packages/@aws-cdk/integ-tests-alpha/lib/assertions/waiter-state-machine.ts
```diff
@@ -33,6 +34,13 @@ export interface WaiterStateMachineOptions {
    * @default 1 (no backoff)
    */
   readonly backoffRate?: number;
+
+  /**
+   * How long, in days, the log contents will be retained.
+   *
+   * @default - no retention days specified
```
msambol, line 41: nit: could you remove the hyphen to match the other defaults ?

msambol: 👍🏼👍🏼

TheRealAmazonKendra:
So, here's what I'm wondering overall: instead of a person having to specify this in each and every location we create these logs, could we just use where this is set on the assertions and then add the same here? Or, could we set this one the IntegTest construct and then use that value everywhere within the test infrastructure?

packages/@aws-cdk/integ-tests-alpha/lib/assertions/waiter-state-machine.ts
TheRealAmazonKendra: I think that no default is correct here. It's what was also done here https://github.com/aws/aws-cdk/pull/29277#discussion_r1525419143.

sugardon:
Right now, I think people should specify log retention every location, like below.

```ts
import * as logs from 'aws-cdk-lib/aws-logs';

declare const testCase: IntegTest;
declare const start: IApiCall;

const describe = testCase.assertions.awsApiCall('StepFunctions', 'describeExecution', {
  executionArn: start.getAttString('executionArn'),
  // already we can set log retention after #29277
  RetentionDays: logs.RetentionDays.ONE_WEEK,
}).expect(ExpectedResult.objectLike({
  status: 'SUCCEEDED',
})).waitForAssertions({
  totalTimeout: Duration.minutes(5),
  interval: Duration.seconds(15),
  backoffRate: 3,
  // this PR
});
```

> could we set this one the IntegTest construct and then use that value everywhere within the test infrastructure?

I think it's a good interface! But, I'm not sure how to implement it right now. If possible, I would like to implement it.

packages/@aws-cdk/integ-tests-alpha/README.md
```diff
@@ -555,6 +557,7 @@ const describe = testCase.assertions.awsApiCall('StepFunctions', 'describeExecut
   totalTimeout: Duration.minutes(5),
   interval: Duration.seconds(15),
   backoffRate: 3,
+  logRetention: logs.RetentionDays.ONE_WEEK,
```
comcalvi, line 560: could we add this property to an existing integration test?

comcalvi (changes requested):

packages/@aws-cdk/integ-tests-alpha/README.md
sugardon: Thank you for instructive comment! 
I think, it might take time, but I will try it!

Issue #27199, opened 2023-09-19
s3|ecr: auto-delete-[objects|images] breaks on cloudformation rollback (suspected)


kaizencc:
Based on the way the custom resource is implemented, it is likely that unexpected behavior happens on Cloudformation rollback, i.e. the custom resource will prematurely delete the objects.

Consider the following scenario:

    UPDATE target resource (replacement, creates a new resource)
    UPDATE custom resource (old -> new, objects in old bucket are deleted)
    (...stuff happens...)
    ERROR, triggers a rollback
    UPDATE custom resource (new -> old)
    DELETE target resource (deletes the new resource, remembers the existing one)

We will have deleted objects in the bucket that has been rolled back to in this scenario.

---

The correct way to handle this is what was done in synthetics: 

https://github.com/aws/aws-cdk/blob/3d6d042ec62135627a42ffeae94b13782d78d1cb/packages/%40aws-cdk/custom-resource-handlers/lib/aws-synthetics-alpha/auto-delete-underlying-resources-handler/index.ts#L26-L36

As opposed to

https://github.com/aws/aws-cdk/blob/3d6d042ec62135627a42ffeae94b13782d78d1cb/packages/%40aws-cdk/custom-resource-handlers/lib/aws-s3/auto-delete-objects-handler/index.ts#L23-L35

GavinZZ:
I can confirm that this is a reproducible issue. Here's the cdk code I used
```
const bucket = new s3.Bucket(this, 'Bucket', {
      removalPolicy: cdk.RemovalPolicy.DESTROY,
      bucketName: <a bucket name that's not used>,
      autoDeleteObjects: true,
    });

    // Intentionally failure since `mybucket-1` exists
    const bucket2 = new s3.Bucket(this, 'Bucket2', {
      removalPolicy: cdk.RemovalPolicy.DESTROY,
      bucketName: <a bucket name that's not used>,
    });

    bucket2.node.addDependency(bucket);
```

Once the deployment is successful, add some random content to the bucket, then update the code so that the first bucket's `bucketName` is updated to another valid name. Update the second bucket's `bucketName` to be an existing bucket name, which will trigger a deployment failure hence roll back. 

The stack is rolled back, but the content is the bucket is not deleted.

PR #29581, opened 2024-03-22
fix(custom-resource-handler): auto-delete-[objects|images] breaks on cloudformation rollback

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/27199

### Reason for this change

Based on the way the custom resource is implemented, it is likely that unexpected behavior happens on Cloudformation rollback, i.e. the custom resource will prematurely delete the objects.

Consider the following scenario:

```
UPDATE target resource (replacement, creates a new resource)
UPDATE custom resource (old -> new, objects in old bucket are deleted)
(...stuff happens...)
ERROR, triggers a rollback
UPDATE custom resource (new -> old)
DELETE target resource (deletes the new resource, remembers the existing one)
```

We will have deleted objects in the bucket that has been rolled back to in this scenario, but the content is now gone.

### Description of changes

Instead of deleting it right during update, we send back `PhysicalResourceId` in the event handler which if the id changes, it will let CFN to empty and delete the bucket at the end of the deployment.

New & updated tests. Also manually tested with deploying a template 
```
const bucket = new s3.Bucket(this, 'Bucket', {
      bucketName: <a bucket name that's not used>,
      autoDeleteObjects: true,
    });

    // Intentionally failure since `mybucket-1` exists
    const bucket2 = new s3.Bucket(this, 'Bucket2', {
      bucketName: <a bucket name that's not used>,
    });

    bucket2.node.addDependency(bucket);
```

Once the deployment is successful, add some random content to the bucket, then update the code so that the first bucket's bucketName is updated to another valid name. Update the second bucket's bucketName to be an existing bucket name, which will trigger a deployment failure hence roll back. 

After the change, the content will stay there if a deployment failure happens. The content & bucket will be deleted if deployment is successful.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/custom-resource-handlers/lib/aws-s3/auto-delete-objects-handler/index.ts
+++ b/packages/@aws-cdk/custom-resource-handlers/lib/aws-s3/auto-delete-objects-handler/index.ts
@@ -30,8 +30,9 @@ async function onUpdate(event: AWSLambda.CloudFormationCustomResourceEvent) {
   /* If the name of the bucket has changed, CloudFormation will try to delete the bucket
     and create a new one with the new name. Returning a PhysicalResourceId that differs
     from the event's PhysicalResourceId will trigger a `Delete` event for the custom
-    resource. The `Delete` event will trigger `onDelete` function which will
-    empty the content of the bucket and then proceed to delete the bucket. */
+    resource, note that this is default CFN behaviour. The `Delete` event will trigger 
+    `onDelete` function which will empty the content of the bucket and then proceed to 
+    delete the bucket. */
   return { PhysicalResourceId: newBucketName };
 }
 


packages/@aws-cdk/custom-resource-handlers/lib/aws-ecr/auto-delete-images-handler/index.ts
```diff
@@ -22,16 +23,14 @@ export async function autoDeleteHandler(event: AWSLambda.CloudFormationCustomRes
...
-    return onDelete(oldRepositoryName);
-  }
+  const newRepositoryName = updateEvent.ResourceProperties?.RepositoryName ?? oldRepositoryName;
+
+  /* If the name of the repository has changed, CloudFormation will try to delete the repo
+  and create a new one with the new name. Returning a PhysicalResourceId that differs
+  from the event's PhysicalResourceId will trigger a `Delete` event for the custom
+  resource. The `Delete` event will trigger `onDelete` function which will
+  empty the content of the repository and then proceed to delete the repository. */
+  return { PhysicalResourceId: newRepositoryName };
```
paulhcsun, line 33: What was the original return value from `onDelete()` and how was it used after this function call? Or does the return value not matter and it was just to perform the deletion as part of thie `onUpdate()`? 

Also where is the logic that checks if the repository name has changed or not and triggers the deletion of the bucket at the end of the deployment?

GavinZZ: `onDelete` doesn't return any value so it doesn't matter. We used to directly call the delete methods defined in `onDelete` not not using the default behaviour CFN provided (which is to do the deletion for us if the id changes).

That logic is default with in CFN custom resource. `The value returned for a PhysicalResourceId can change custom resource update operations. If the value returned is the same, it is considered a normal update. If the value returned is different, AWS CloudFormation recognizes the update as a replacement and sends a delete request to the old resource. `. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-requesttypes-delete.html

paulhcsun (approved):

paulhcsun:
@mergifyio update

PR #29582, opened 2024-03-22
docs(lambda-event-source): improve docstring in KinesisEventSourceProps

### LINKED ISSUE #27321: aws_lambda_event_sources: Missing support for maxRecordAge=-1 value in  KinesisEventSourceProps

roger-zhangg:
### Issue # (if applicable)

Closes #27321

### Reason for this change

Our customer found `maxRecordAge`  can't be set to `-1` Due to limitation of `cdk.Duration`. However, it's not obvious in the docstring that the [default value](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html#cfn-lambda-eventsourcemapping-maximumrecordageinseconds) of `maxRecordAge` is already -1. Our customer can simply leave this prop empty to set it to -1. 

So, This PR improves the docstring to make default value more apparent.

See more discussion in the ticket.

### Description of changes

Improve docstring of `maxRecordAge` and `retryAttempts` in `KinesisEventSourceProps` in `aws-lambda-event-source`

### Checklist

----


packages/aws-cdk-lib/aws-lambda-event-sources/lib/stream.ts
```diff
@@ -73,7 +75,8 @@ export interface StreamEventSourceProps extends BaseStreamEventSourceProps {
    * * Minimum value of 0
    * * Maximum value of 10000
    *
-   * @default - retry until the record expires
+   * @default -1 
+   * retry until the record expires
```
jonife, line 79: This seems like an oversight.

roger-zhangg: Thanks for the review @jonife , This one is -1 as well, [see doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html#cfn-lambda-eventsourcemapping-maximumrecordageinseconds), adding default here the same reason for maxRecordAge

packages/aws-cdk-lib/aws-lambda-event-sources/lib/stream.ts
```diff
@@ -63,7 +63,9 @@ export interface StreamEventSourceProps extends BaseStreamEventSourceProps {
    * * Minimum value of 60 seconds
    * * Maximum value of 7 days
    *
```
lpizzinidev, line 65:
```suggestion
   * * Maximum value of 7 days
   *
   * The default value is -1, which sets the maximum age to infinite.
   * When the value is set to infinite, Lambda never discards old records.
   *
```

packages/aws-cdk-lib/aws-lambda-event-sources/lib/stream.ts
```diff
@@ -63,7 +63,9 @@ export interface StreamEventSourceProps extends BaseStreamEventSourceProps {
...
-   * @default - the retention period configured on the stream
+   * @default -1
+   * sets the maximum age to infinite. When infinite, Lambda never discards old records.
+   * The retention period configured on the stream now decides max age for records.
```
lpizzinidev, line 68:
```suggestion
```
Let's add an explanation in the documentation above

packages/aws-cdk-lib/aws-lambda-event-sources/lib/stream.ts
```diff
@@ -73,7 +75,8 @@ export interface StreamEventSourceProps extends BaseStreamEventSourceProps {
    * * Minimum value of 0
    * * Maximum value of 10000
    *
```
lpizzinidev, line 77:
```suggestion
   *
   * The default value is -1, which sets the maximum number of retries to infinite.
   * When MaximumRetryAttempts is infinite, Lambda retries failed records until
   * the record expires in the event source.
   *
```

packages/aws-cdk-lib/aws-lambda-event-sources/lib/stream.ts
```diff
@@ -73,7 +75,8 @@ export interface StreamEventSourceProps extends BaseStreamEventSourceProps {
...
-   * @default - retry until the record expires
+   * @default -1
+   * retry until the record expires
```
lpizzinidev, line 79:
```suggestion
```
Same as above.

lpizzinidev (changes requested):
Thanks 👍
Left some suggestions, feel free to improve it further.

Note that I don't think this will solve https://github.com/aws/aws-cdk/issues/27321 completely as we still won't be able to update the `maxRecordAge` value for existing `KinesisEventSource` resources, eg:
1. User creates `KinesisEventSource` specifying `maxRecordAge` to some value
```ts
new KinesisEventSource(stream, {
  ...
  maxRecordAge: cdk.Duration.seconds(100),
});
```
2. User updates stack and removes `maxRecordAge` from declared `KinesisEventSource`:
```ts
new KinesisEventSource(stream, {
  ...
  maxRecordAge: undefined,
});
```

I think that the `maxRecordAge` will remain at 100 seconds (please correct me if I'm wrong).
Ie, we need a way to specify `maxRecordAge: -1` to support all use cases.

roger-zhangg:
Hi @lpizzinidev Thanks for the review and pointing out the update use case. However, I couldn't reproduce what you mentioned there. 

Steps:

- I created a stack with `maxRecordAge: cdk.Duration.seconds(100),`
- I updated stack by removing `maxRecordAge` prop.
- I can see `maxRecordAge` is set to -1 in Lambda console. 

If this is the case on your end, could you help to share your step to reproduce this issue? Thanks.

lpizzinidev:
Hi @roger-zhangg ✋
My assumption was wrong then, thanks for taking the time and running the scenario!
If you can please go ahead and update the documentation as suggested (feel free to improve it further of course) then this will be good to go for me 👍

roger-zhangg:
Hi @TheRealAmazonKendra @lpizzinidev Thanks for the suggestions and I'd like to get your review on this so we can move forward.

scanlonp (approved):
Thanks for this docs improvement; the wording is super clear.

PR #29594, opened 2024-03-24
feat(cloudwatch): construct for alarms based on anomaly detection (#10540)

### LINKED ISSUE #10540: [aws-cloudwatch] Missing ThresholdMetricId
When creating Cloudwatch Alarm that makes use of anomaly detection the following error is raised at deploy time:

```
ComparisonOperators for ranges require ThresholdMetricId to be set (Service: AmazonCloudWatch; Status Code: 400; Error Code: ValidationError; Request ID: xxx; Proxy: null)
```

### Reproduction Steps

```python
aws_cloudwatch.Alarm(
     self,
    "my_alarm",
    evaluation_periods=1,
    metric=sqs_queue.metric_number_of_messages_sent(),
    threshold=2,
    statistic="Average",
    comparison_operator=aws_cloudwatch.ComparisonOperator.LESS_THAN_LOWER_OR_GREATER_THAN_UPPER_THRESHOLD,
)
```

The Alarm construct works great for any of the other ComparisonOperators that do not rely on anomaly detection i.e. `GreaterThanThreshold`. 

### Environment

  - **CLI Version      :** 1.62.0
  - **Framework Version:** 1.64.0
  - **Node.js Version:** v12.14.1
  - **OS               :** MacOS Catalina
  - **Language (Version):** Python 3.7

--- 

This is :bug: Bug Report

vgolemis:
feat(cloudwatch): construct for alarms based on anomaly detection (#10540)

### Issue # (if applicable)

Closes #10540.

### Reason for this change
Introduce new construct `AnomalyDetectionAlarm` that covers the gap of creating an `Alarm` based on anomaly detection. 

### Description of changes
* Documentation for the API of `AnomalyDetectionAlarm`
* New class `AnomalyDetectionAlarm` that creates an alarm that utilizes [CloudWatch anomaly detection](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Anomaly_Detection.html
  * The underlying resource is a `AWS::CloudWatch::Alarm`
  * It can create the Alarm either directly if the `metric` property has an `ANOMALY_DETECTION_BAND` metric expression or can generate it. The user can provide the `generateAnomalyDetectionExpression` property to define this behavior.
  * Validates the the `ComparisonOperator` provided applies to the anomaly detection scenario.
  * The `thresholdMetricId` was not exposed to the user because the `IMetric` interface does not provide a property to define the id of the metric. This is done when defining the `MetricDataQueryProperty`; decided to not force the user to provide the array of `MetricDataQueryProperty`.
  * The `threshold` property is mandatory by the `CreateAlarmOptions`. It was repurposed to define the anomaly detection threshold if the function generates the `ANOMALY_DETECTION_BAND` metric expression.
  * The implementation of the `renderMetric()` verifies the expressions provided are correct and forces the appropriate reference ids to be in place.
* The static methods `fromAlarmName()` and `fromAlarmArn()` were the same for `Alarm` and `CompositeAlarm` so moved the to the `AlarmBase` so all of the classes can access them.

Integration test and Unit tests and  have been added for the `AnomalyDetectionAlarm`. Verified the existing tests for the `Alarm` and `CompositeAlarm` are not impacted.

### Checklist

----


packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -24,6 +24,20 @@ export interface AlarmProps extends CreateAlarmOptions {
   readonly metric: IMetric;
 }
 
+/**
+ * Properties for Anomaly Detection Alarms
+ */
+export interface AnomalyDetectionAlarmProps extends AlarmProps {
```
comcalvi, line 30: The `AnomalyDetectionAlarmProps` extend the `AlarmProps`, but the `AnomalyDetectionAlarm` does not extend `Alarm`. 

Why is the `AnomalyDetectionAlarm` a separate construct? It looks like it's a strict superset of the `Alarm` we have already, which suggests that instead of making a new construct we could add a prop to `Alarm`. 

To make this decision we have to understand what the differences between an `AnomalyDetectionAlarm` and an `Alarm` are. Are there any properties unique to the `AnomalyDetectionAlarm` which conflict with the props of a regular `Alarm`? (Props 'A' and 'B' are said to conflict if I can (or should) specify 'A' or 'B', but not both).

comcalvi (changes requested):
Why is there no `CfnAnomalyDetector` in this construct? Can you help me understand how this works? I thought the `AnomalyDetectorAlarm` required a reference to the metric of an [AnomalyDetector](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudwatch-anomalydetector.html)

PR #29607, opened 2024-03-25
chore(opensearchservice): t3 instance type does not support Multi-AZ with standby feature

tmokmss:
### Issue # (if applicable)
Related with #26026

### Reason for this change

#26082 enabled Multi-AZ with Standby by default, but deployment fails if we use t3 instance type, because it does not support the feature. To fail fast, this PR adds validation on synth time.

> Multi-AZ with Standby only works with the m5, c5, r5, r6g, c6g, m6g, r6gd and i3 instance types.
> https://docs.aws.amazon.com/opensearch-service/latest/developerguide/managedomains-multiaz.html

> You can use T3 instance types only if your domain is provisioned without standby.
> https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-instance-types.html#latest-gen

### Description of changes

If the instance type of data node or master node is t3, throws an error.

I also considered to automatically set `multiAzWithStandbyEnabled: false` if we detect any t3 instance type, but it would introduce unwanted behavior e.g. in the below case:

```ts
// Initial state
// multiAzWithStandbyEnabled: true as there's no t3 instance type
 new Domain(stack, 'Domain', {
    version: engineVersion,
    capacity: {
      dataNodeInstanceType: 'r5.large.search',
    },
})

// Update domain to add master nodes with t3 instance type
new Domain(stack, 'Domain', {
    version: engineVersion,
    capacity: {
      dataNodeInstanceType: 'r5.large.search',
      masterNodeInstanceType: 't3.medium.search',
      masterNodes: 3,
    },
})

// multiAzWithStandbyEnabled suddenly become false!
```

so we just throw an error.

Added some unit tests.

I also confirmed that it results in deployment error if we try to deploy with t3 instance type & `multiAzWithStandbyEnabled : true` for both data node and master node.

### Checklist

----


packages/aws-cdk-lib/aws-opensearchservice/lib/domain.ts
```diff
@@ -1827,6 +1827,10 @@ export class Domain extends DomainBase implements IDomain, ec2.IConnectable {
       }
     }
 
+    if (isSomeInstanceType('t3') && multiAzWithStandbyEnabled) {
+      throw new Error('T3 instance types does not support Multi-AZ with standby feature.');
```
go-to-k, line 1831: nit.
e.g.)
```suggestion
      throw new Error('T3 instance types do not support Multi-AZ with standby feature.');
```

go-to-k (changes requested):
From what I can see in the management console, OR1 and Im4gn instance types are apparently supported in the Multi-AZ with standby as well.

I selected "Domain with standby" and was able to select both OR1 and Im4gn. I also confirmed that t3 cannot be selected for "Domain with standby", and the t3 can be selected for "Domain without standby" (with the "Include previous generation instance types" toggle).

(Note that I did not actually create the domain.)

https://github.com/aws/aws-cdk/issues/26026#issuecomment-2018025031

open-search-console

tmokmss:
Thanks @go-to-k!

go-to-k (approved):
LGTM. Thanks @tmokmss

GavinZZ (approved):
LGTM, thanks for contributing.

PR #29580, opened 2024-03-22
chore(codebuild): aarch64 standard 2.0 and 3.0 build images are not deprecated

nikovirtala:
### Issue # (if applicable)

Closes #<issue number here>.

### Reason for this change

AWS CodeBuild Docker images `AMAZON_LINUX_2_ARM_3` aka `AMAZON_LINUX_2_STANDARD_3_0` aka `aws/codebuild/amazonlinux2-aarch64-standard:3.0` and `AMAZON_LINUX_2_ARM_2` aka `AMAZON_LINUX_2_STANDARD_2_0` aka `aws/codebuild/amazonlinux2-aarch64-standard:2.0`are not deprecated but the currently actively maintained ones according to https://github.com/aws/aws-codebuild-docker-images and https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html

### Description of changes

I removed the JSDoc Deprecated flag for the currently active images and added their image names as a comment so that one can find the images also using those references.

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #29616, opened 2024-03-26
chore(codebuild): update BuildImage classes

nmussy:
### Issue # (if applicable)

None as far as I could tell

Overlaps with #29580

### Reason for this change

Align the CodeBuild images listed by the CDK with those offered by the SDK

### Description of changes

* Added missing build images
  * Windows 2022 was added with a warning, see [PR comment](https://github.com/aws/aws-cdk/pull/29616#issuecomment-2019991742)
* Removed `@deprecated` messages, see #29580
* Updated JSDoc comments to indicate AL2023 based images, see [docs](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html)
  * It might be a good idea to deprecate and rename `AMAZON_LINUX_2_STANDARD_3_0` to `AMAZON_LINUX_2023_STANDARD_3_0`, despite how the images are named. I'll leave it up to the maintainers
* Added `{@link}` tags where missing

Using the [codebuild:ListCuratedEnvironmentImages](https://docs.aws.amazon.com/codebuild/latest/APIReference/API_ListCuratedEnvironmentImages.html) API command, and comparing it to the CDK.

### Checklist

----


nmussy:
Switching to a draft for now, I need to check whether or not `WINDOWS_SERVER_2022_CONTAINER` is available, the [docs aren't super clear about that](https://docs.aws.amazon.com/search/doc-search.html?searchPath=documentation-guide&searchQuery=WINDOWS_SERVER_2022_CONTAINER&this_doc_product=AWS%20CloudFormation&this_doc_guide=User%20Guide).

If it is, I'll probably update Windows in another PR to add an integ test

nmussy:
As the [Cfn docs assert](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codebuild-project-environment.html#cfn-codebuild-project-environment-type), `WINDOWS_SERVER_2022_CONTAINER` is not a valid `Environment::Type`, sort of.

When attempting to deploy `aws/codebuild/windows-base:2022-1.0` with `"Type": "WINDOWS_SERVER_2022_CONTAINER"`, Cfn returns the following error:
```
Windows Server 2022 not supported for on-demand compute
```

I've also tried deploying it with the incorrect `Type`, i.e. `WINDOWS_SERVER_2019_CONTAINER`:
```
AWS CodeBuild curated image aws/codebuild/windows-base:2022-1.0 is not supported for projects with environment type WINDOWS_SERVER_2019_CONTAINER and compute type BUILD_GENERAL1_MEDIUM
```

This matches [Working with reserved capacity in AWS CodeBuild](https://docs.aws.amazon.com/codebuild/latest/userguide/fleets.html):
> On-demand ﬂeets do not support Windows Server 2022.

The CDK should be able to be set this up with [`CfnProject:ProjectFleetProperty`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_codebuild.CfnProject.ProjectFleetProperty.html), but that's going to be outside the scope of this PR.

I'll open an issue to resolve this later, but in the meantime I'll mark both `WindowsBuildImage.WIN_SERVER_CORE_2022_BASE_3_0` and `WindowsImageType.SERVER_2022` as `@deprecated`, explaining the situation. This can easily be updated later on.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1964,6 +1977,18 @@ export enum WindowsImageType {
    * The WINDOWS_SERVER_2019_CONTAINER environment type
    */
   SERVER_2019 = 'WINDOWS_SERVER_2019_CONTAINER',
+
+  /**
+   * The WINDOWS_SERVER_2022_CONTAINER environment type
+   *
+   * @deprecated Windows Server 2022 images are not supported for on-demand compute.
```
TheRealAmazonKendra, line 1984: I don't think we should be adding version that we are immediately marking as deprecated.

TheRealAmazonKendra (changes requested):
Thanks for adding these. I'm not sure, however, that we should be adding types that we are also immediately marking as deprecated. Can you elaborate on your thinking here?

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
nmussy: Yeah, I just want to give the user a warning that they probably won't be able to use the image as is. I thought the `@deprecated` was a good compromise. I can delete them here and we can add them back with #29617's PR

nmussy: @TheRealAmazonKendra Nudge

nmussy:
Closing in favor of #29754, which will allow an integration test of Windows 2022

PR #29629, opened 2024-03-27
feat(ec2): add support for environment files and variables in systemd

razin99:
### Reason for this change

Allow users to define `Environment` and `EnvironmentFile` in a systemd service file.

### Description of changes

Added 2 new properties for `ec2.InitService.systemdConfigFile`:
1. `environmentVariables` which is a string key value pair
2. `environmentFiles` which is a list of file paths

I have added a unit test.
Documentations referred to:
- https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#EnvironmentFile=
- https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Environment= 

### Checklist

----


razin99:
Exemption Request

No README updates needed since this added optional features that goes beyond basic systemd service creation

packages/@aws-cdk/cx-api/FEATURE_FLAGS.md
lpizzinidev, line 1: Why is this file included? 🤔

packages/aws-cdk-lib/aws-ec2/lib/cfn-init-elements.ts
```diff
@@ -1108,4 +1123,18 @@ export interface SystemdConfigFileOptions {
...
+
+  /**
+   * Include environment variables when process is running
+   *
+   * @default - No environment variables set
+   */
+  readonly environmentVariables?: Record<string, string>;
+
+  /**
+   * Include file to load environment variables from when process is running
```
lpizzinidev, line 1135:
```suggestion
   * Loads environment variables from files when the process is running.
```

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.instance-init.ts
```diff
@@ -49,6 +49,15 @@ new ec2.Instance(stack, 'Instance2', {
         ec2.InitUser.fromName('sysuser2'),
         ec2.InitSource.fromAsset('/tmp/sourceDir', tmpDir),
       ]),
+      service: new ec2.InitConfig([
+        ec2.InitFile.fromString('/myvars.env', 'OTHER_VAR="im from the file :3"'),
+        ec2.InitService.systemdConfigFile('myapp', {
+          command: '/bin/bash -c "echo HELLO_WORLD=${MY_VAR} | FROM_FILE=${OTHER_VAR}"',
+          environmentVariables: {
```
lpizzinidev, line 56: Can you please add the `environmentFiles` variable to the test?

lpizzinidev (changes requested):
Thanks for contributing 👍
Left some comments for adjustments.
Also, you could add some documentation in [Configuring Instances using CloudFormation Init (cfn-init)](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/README.md#configuring-instances-using-cloudformation-init-cfn-init).

packages/@aws-cdk/cx-api/FEATURE_FLAGS.md
razin99: Yeah im not too sure :/
It popped up when I was doing some integ tests. Will remove :)

packages/aws-cdk-lib/aws-ec2/README.md
```diff
@@ -1426,6 +1426,22 @@ ec2.CloudFormationInit.fromElements(
 );
 ```
 
+If your services require environment variables, you can use the `environmentVariables` or `environmentFiles`
+parameter to do so:
+```ts
```
lpizzinidev, line 1431:
```suggestion
You can use the `environmentVariables` or `environmentFiles` parameters to specify environment variables
for your services:

```ts
```

lpizzinidev (approved):
Thanks 👍 Added a suggestion to change the wording for the documentation

razin99:
@lpizzinidev what do I do from here? Never got a PR this far in this repo. Do I just wait for it to be merged?

TheRealAmazonKendra (commented):
Thank you for your work on this. I'm thinking that it's likely that this change is good to go with one clarifying question: can a user provide both environmentVariables and environmentFiles or is it an either/or situation?

If they can provide both, this looks ready.

razin99:
> Thank you for your work on this. I'm thinking that it's likely that this change is good to go with one clarifying question: can a user provide both environmentVariables and environmentFiles or is it an either/or situation?
> 
> If they can provide both, this looks ready.

They can provide both, as this is a supported behaviour by systemd. From what I've read ([here](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#EnvironmentFile=:~:text=Settings%20from%20these%20files%20override%20settings%20made%20with%20Environment%3D.%20If%20the%20same%20variable%20is%20set%20twice%20from%20these%20files%2C%20the%20files%20will%20be%20read%20in%20the%20order%20they%20are%20specified%20and%20the%20later%20setting%20will%20override%20the%20earlier%20setting.)), anything `EnvironmentFile` can override `Environment`

Quote from documentation:
> Settings from these files override settings made with Environment=. If the same variable is set twice from these files, the files will be read in the order they are specified and the later setting will override the earlier setting.

GavinZZ (approved):
Thanks for answering the questions. Approving on behalf for Kendra since her question is answered and this PR looks good to me.

GavinZZ (approved):

Issue #29564, opened 2024-03-20
aws-cdk: S3 set publicReadAccess: true, fails deploy because of default deny public access policy


irobinsonDandH:
### Describe the bug

So, if you make a new s3 bucket 

const staticBucket = new aws_s3.Bucket(s3Stack, `static-Bucket`, {
    bucketName: `static-bucket`,
    publicReadAccess: true,
    }
  })
  
While this is fine code and you can deploy it will fail in the middle with a generic access denied error not telling you what stopped it even if you are full admin. This happens due to the default deny all public access rule.

### Expected Behavior

So if you make a new s3 bucket 

const staticBucket = new aws_s3.Bucket(s3Stack, `static-Bucket`, {
    bucketName: `static-bucket`,
    publicReadAccess: true,
    }
  })
  
it will create the s3 bucket with the policy and set the deny public access to false for all 4 options

### Current Behavior

Fails with access denied error while creating the bucket and doesn't say that it's because of the policy. 

### Reproduction Steps

Use the following code changing the bucket name to something unique.

const staticBucket = new aws_s3.Bucket(s3Stack, `static-Bucket`, {
    bucketName: `static-bucket`,
    publicReadAccess: true,
    }
  })
  
  npx cdk deploy app

### Possible Solution

A possible solution would be if you use publicReadAccess: true set all blockPublicAccess to false implicitly same if you use the grantPublicAccess() function.

or state in the documentation that you have to set blockPublicAceess to false and give a better error back.

blockPublicAccess: {
      blockPublicAcls: false,
      blockPublicPolicy: false,
      ignorePublicAcls: false,
      restrictPublicBuckets:false,
    }
    

### CDK CLI Version

2.124

### Node.js Version

18

### OS

Debian

### Language

TypeScript

### Language Version

5.3.3

### Other information

Current workaround is adding 

blockPublicAccess: {
      blockPublicAcls: false,
      blockPublicPolicy: false,
      ignorePublicAcls: false,
      restrictPublicBuckets:false,
    }
    
    but I feel that publicReadAccess: true should just handle the bucket level permissions fully.

pahud:
>A possible solution would be if you use publicReadAccess: true set all blockPublicAccess to false implicitly same if you use the grantPublicAccess() function.
>or state in the documentation that you have to set blockPublicAceess to false and give a better error back.

Yes we should do that with either of the suggested solutions.

At this moment the CFN is returning 403.

[...]

PR #29632, opened 2024-03-27
fix(s3): publicReadAccess causes deployment failure due to access denied 403

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29564

### Reason for this change

if you make a new s3 bucket
```
const staticBucket = new aws_s3.Bucket(s3Stack, `static-Bucket`, {
    bucketName: `static-bucket`,
    publicReadAccess: true,
})
```
While this is fine code and you can deploy it will fail in the middle with a generic access denied error not telling you what stopped it even if you are full admin. This happens due to the default deny all public access rule.

### Description of changes

When users only enable `publicReadAccess` without configuring `blockPublicAccess` to disable it, we will raise an exception and throw an more appropriate error message for easier diagnosis. 

We do not want to directly disable `blockPublicAccess` as it feels like a weird behaviour.

New unit tests and updated integ tests

### Checklist

----


packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1995,6 +1995,10 @@ export class Bucket extends BucketBase {
     (props.lifecycleRules || []).forEach(this.addLifecycleRule.bind(this));
 
     if (props.publicReadAccess) {
+      if (props.blockPublicAccess === undefined) {
```
TheRealAmazonKendra, line 1998: So, if possible, I'd like to see this enforced in the contract rather than validation after the fact. 

Looking deeper into this module, we actually have four fields that interplay in a complex and unideal way:  `accessControl`, `publicReadAccess`, `blockPublicAccess`,  and `objectOwnership`.

`accessControl` is a legacy property and generally shouldn't be used anymore BUT we can't just take it away altogether. The values that can be set here are the list of canned ACLs S3 provides: 
```
PRIVATE | Owner gets FULL_CONTROL.
PUBLIC_READ | Owner gets FULL_CONTROL.
PUBLIC_READ_WRITE | Owner gets FULL_CONTROL.
AUTHENTICATED_READ | Owner gets FULL_CONTROL.
LOG_DELIVERY_WRITE | The LogDelivery group gets WRITE and READ_ACP permissions on the bucket.
BUCKET_OWNER_READ | Object owner gets FULL_CONTROL.
BUCKET_OWNER_FULL_CONTROL | Both the object owner and the bucket owner get FULL_CONTROL over the object.
AWS_EXEC_READ | Owner gets FULL_CONTROL.
```
Newly created buckets have ACLs disabled by default so the default here is `PRIVATE`.

Unless the user sets `objectOwnership`, however, to enable ACLs, the only value that works is `PRIVATE`. If they do set `objectOwnership`, it can be one of the following options:
```
BUCKET_OWNER_ENFORCED | ACLs are disabled, and the bucket owner automatically owns and has full control over every object in the bucket.
BUCKET_OWNER_PREFERRED | The bucket owner will own the object if the object is uploaded with the bucket-owner-full-control canned ACL.
OBJECT_WRITER | The AWS account that uploads an object owns the object, has full control over it, and can grant other users access to it through ACLs.
```
Now, if a user sets `publicReadAccess` you can see that only a small subset of these values will work with that. Anything other than `accessControl: AccessControl.PUBLIC_READ` (or maybe also `accessControl: AccessControl.PUBLIC_READ_WRITE` will be invalid.

Lastly, `blockPublicAccess` again only works with specific settings and in specific configurations. (Too complicated, didn't write).

Alllllllll of this to say: What if we deprecate all of these fields and create a class that locks down the contract to the valid permutations via an enum like class?

TheRealAmazonKendra: It's worth also noting that the encryption fields are similarly messy. It's potentially worth investing some time into a BucketV2 construct to really tighten up the contract here and make it a better user experience overall.

TheRealAmazonKendra (dismissed):

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
GavinZZ: @TheRealAmazonKendra Thanks Kendra for the review. This seems like a new feature with BucketV2 than a bug fix. Would you classify the BucketV2 as a p1 feature request or a p2 feature request? Just want to know the priority.

GavinZZ: Thanks Kendra for the detailed description. While I appreciate your suggestion and I do think this would be great for customers, I don't think in place update for the current `Bucket` construct is a good idea. As you described, we already have 4 properties that are related to the bucket access control. Now if we deprecate them and come up with a new property to handle the access control, it is
1. Error prune as the logic is quite nasty and there's edge cases that we may miss.
2. More confusing for users for an additional properties.

I think what I'll do is to merge this PR while create a GitHub issue as a p2 feature request on a BucketV2 with better handling on the access control section.

packages/aws-cdk-lib/aws-s3/test/bucket.test.ts
```diff
@@ -879,6 +879,28 @@ describe('bucket', () => {
     });
   });
 
+  test('bucket with default block public access setting to throw error msg', () => {
+    const stack = new cdk.Stack();
+
+    expect(() => new s3.Bucket(stack, 'Bucket', {
+      publicReadAccess: true,
+    })).toThrow('Cannot use \'publicReadAccess\' property');
```
aaythapa, line 887: error msg needs updating?

```suggestion
    })).toThrow(''Cannot use \'publicReadAccess\' property on a bucket without allowing bucket-level public access through \'blockPublicAceess\' property.');
```

aaythapa (commented):
tiny comment

aaythapa (approved):

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -2000,6 +2000,10 @@ export class Bucket extends BucketBase {
...
+        throw new Error('Cannot use \'publicReadAccess\' property on a bucket without allowing bucket-level public access through \'blockPublicAceess\' property.');
```
richard-stafflink, line 2004: @GavinZZ 
blockPublicAceess -> blockPublicAccess

PR #29646, opened 2024-03-29
feat(apigatewayv2-integrations): sqs integrations

### LINKED ISSUE #24785: apigatewayv2-integrations: Add support for SQS-SendMessage integration
### Describe the feature

ApiGatewayv2 HTTP API integrates with SQS to redirect the request into a queue. It would be nice to have CDK implement this. See [documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-aws-services-reference.html#SQS-SendMessage) on `SQS-SendMessage` mapping.

### Use Case

I want an easy way to use an SQS Queue as a proxy for my HTTP API.

### Proposed Solution

Implement the construct similarly to the lambda integration.

It would look like the following:

```ts
    const stack = new Stack();
    const api = new HttpApi(stack, 'HttpApi');
    const queue = DefaultQueue(stack, 'Queue');
    new HttpRoute(stack, 'LambdaProxyRoute', {
      httpApi: api,
      integration: new HttpSqsQueueIntegration('Integration', queue, {
        parameterMapping: new ParameterMapping()
          .appendHeader('header2', MappingValue.requestHeader('header1'))
          .removeHeader('header1'),
      }),
      routeKey: HttpRouteKey.with('/pets'),
    });
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.70.0

Ubuntu 20

badmintoncryer:
### Issue # (if applicable)

Closes #24785.

### Reason for this change

HTTP API supports AWS service integrations with SQS but CDK does not support this.

### Description of changes

Added `HttpSqsIntegration` class

### Checklist

----


packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/http/sqs.ts
```diff
@@ -0,0 +1,117 @@
...
+  readonly parameterMapping?: apigwv2.ParameterMapping;
+
+  /**
+   * The subtype of the HTTP integration.
+   *
+   * Only subtypes starting with SQS_ can be specified.
+   *
+   * @default HttpIntegrationSubtype.SQS_SEND_MESSAGE
+   */
+  readonly subtype?: apigwv2.HttpIntegrationSubtype;
```
TheRealAmazonKendra, line 29: Since only a couple of these are relevant here, I think that this isn't quite the right type to use here. Of the 10 values, only 4 work.

Additionally, per the comments above where you mention that certain arns need to be added based off the subtype here, I think that these props aren't quite the right contract in general. For instance, in the case that this subtype is `START_EXECUTION`, the parameter mapping shouldn't be optional because it must include at least the `StateMachineArn`.

TheRealAmazonKendra: Wait, I actually went back and read through the SubTypes more and it looks like the subtypes listed in the `parameterMapping` docstring above don't start with `SQS_` so those comments don't mesh with the ones here in `subType`.

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/http/sqs.ts
```diff
@@ -0,0 +1,117 @@
...
+  /**
+   * Specifies how to transform HTTP requests before sending them to the backend.
+   *
+   * When the subtype is either `START_EXECUTION` or `START_SYNC_EXECUTION`,
+   * it is necessary to specify the `StateMachineArn`.
+   * Conversely, when the subtype is `STOP_EXECUTION`, the `ExecutionArn` must be specified.
+   *
+   * @see https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-parameter-mapping.html
+   *
+   * @default - specify only `StateMachineArn`
```
TheRealAmazonKendra, line 18: Below in the code you're using `this.props.queue.queueUrl` so this doesn't seem accurate.

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/http/sqs.ts
```diff
@@ -0,0 +1,117 @@
...
+   */
+  readonly parameterMapping?: apigwv2.ParameterMapping;
```
TheRealAmazonKendra, line 20: Are these parameter mappings typically defined or can they be whatever the user adds here?

TheRealAmazonKendra (changes requested):
Thank you for your contribution! I really like the thought put in here to default values and abstracting away some of the parameters you know users will need every time. I have a few questions/comments inline regarding the contract. I'm not sure it's quite right but this is a great start.

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/http/sqs.ts
badmintoncryer: I'm really sorry for describing crazy comments for `parameterMapping`. 
I have updated it and could you please re-confirm it?

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/http/sqs.ts
badmintoncryer: I have fixed it.

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/http/sqs.ts
badmintoncryer: Default parameter mappings are based on this [docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-aws-services-reference.html#SQS-SendMessage).
I defined only required parameter such as `QueueUrl` for `SQS_SEND_MESSAGE`.

If a user wants to set optional parameters or change the values of required parameters, it is assumed that they will provide a custom-defined `parameterMapping`.

Is this response addressing your question? Please feel free to correct me if my understanding is incorrect.

badmintoncryer:
@TheRealAmazonKendra Thank you for your review! I deeply apologize for the nonsensical comments regarding the `parameterMappings`. (It seems that it was likely mistakenly copied from somewhere...)
I have corrected the relevant sections and responded to your questions. Please review at your convenience.

packages/aws-cdk-lib/aws-apigatewayv2-integrations/README.md
```diff
@@ -117,6 +118,97 @@ httpApi.addRoutes({
 - The `executionArn` parameter is required for the `STOP_EXECUTION` subtype. It is necessary to specify the `executionArn` in the `parameterMapping` property of the `HttpStepFunctionsIntegration` object.
 - `START_SYNC_EXECUTION` subtype is only supported for EXPRESS type state machine.
 
+### SQS Integration
+
+SQS integrations enable integrating an HTTP API route with AWS SQS.
+This allows the HTTP API to send, receive and delete messages from an SQS queue.
+
+The following code configures a Step Functions integrations:
```
chriswilty, line 126: `configures an SQS integration`

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/http/sqs.ts
chriswilty: Why not narrow down the type to the sane choices?

```ts
readonly subtype: Pick<
  apigwv2.HttpIntegrationSubtype,
  'SQS_SEND_MESSAGE' | 'SQS_RECEIVE_MESSAGE' | 'SQS_DELETE_MESSAGE' | 'SQS_PURGE_QUEUE'
>
```

packages/aws-cdk-lib/aws-apigatewayv2-integrations/README.md
badmintoncryer: Thanks a lot!

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/http/sqs.ts
badmintoncryer: @chriswilty Thank you for your suggestion.
This `PickEnum` type will work correctly in TypeScript. However, it may not function properly when converted to other languages using jsii. 
For example, I believe even TypeScript's union types couldn't be used in CDK.

I would like to check with the maintainer if it is usable.

epoctic:
This looks awesome, can't wait to see it in a release!

badmintoncryer:
@TheRealAmazonKendra Could you please review this PR again?

packages/aws-cdk-lib/aws-apigatewayv2-integrations/lib/http/sqs.ts
GavinZZ: IMO, I think we can keep this since there's validation below that checks the subtype values.

GavinZZ (dismissed):

GavinZZ:
Going to approve on behalf of Kendra since this PR has been sitting here for almost half a year.

GavinZZ (dismissed):

GavinZZ:
@badmintoncryer can you fix the linting errors when available?

badmintoncryer:
@GavinZZ Thank you for your approval. I've fixed linting errors.

GavinZZ (approved):

PR #29596, opened 2024-03-24
feat(custom-resources): add optional logging flag to `AwsCustomResourceProps`

colifran:
### Reason for this change

SDK v2 and v3 handlers for `AwsCustomResource` log the event object passed to the handler, API responses, and caught /uncaught errors. This can potentially result in logging sensitive information that a user may wish to hide. This PR introduces a `disableLogging` flag that can be used to disable all logging configured in the SDK v2 and v3 handlers.

### Description of changes

Added a `disableLogging` flag to the `AwsCustomResourceProps` interface. The value of `disableLogging` is provided in the handler event object as a `ResourceProperty`. All logging is now logged conditionally based on the value of this property.

### Description of how you validated changes

* Existing integ tests have been updated
* A new integ test with `disableLogging` set to true was added
* A unit test was added to check that `disableLogging` is `false` by default
* A unit test was added to check that when `disableLogging` is `true` it is set correctly in the synthesized template

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #29648, opened 2024-03-29
feat(custom-resources): add logging property to `AwsSdkCall` and create `Logging` class

colifran:
### Reason for this change

SDK v2 and v3 handlers for `AwsCustomResource` log the event object passed to the handler, API responses, and caught /uncaught errors for each SDK call made. This can potentially result in logging sensitive information that a user may wish to hide. This PR introduces a new `logging` property on the `AwsSdkCall` interface that can be used to provide more control over logging in the SDK v2 and v3 handlers on a per SDK call basis. The `logging` flag is configurable via a new `Logging` class which exposes two static methods:
- all: all logging during lambda execution is turned on
- withDataHidden: hides all logged data associated with the API call response. This includes the raw response as well as the `Data` field on the response object

Additional logging configurations can be added in the future.

### Description of changes

Added a `logging` flag to the `AwsSdkCall` interface which is configurable via the new `Logging` class. The `Logging` class has an internal `render` method which renders the specified logging configuration which is passed as part of the `create`, `update`, and `delete` `ResourceProperties` to the lambda handler. These `logging` properties are then used throughout the handler to control what is logged based on their value

* A new integ test with `logging` as `withDataHidden` was added
* Unit tests to ensure calling `render` on a `Logging` instance produces the expected result
* Unit tests to ensure that using `logging` with `AwsSdkCall` while using `AwsCustomResource` produces the correct CloudFormation template

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ec2/README.md
+++ b/packages/aws-cdk-lib/aws-ec2/README.md
@@ -215,7 +215,7 @@ const provider = ec2.NatProvider.instanceV2({
 new ec2.Vpc(this, 'TheVPC', {
   natGatewayProvider: provider,
 });
-provider.connections.allowFrom(ec2.Peer.ipv4('1.2.3.4/8'), ec2.Port.tcp(80));
+provider.connections.allowFrom(ec2.Peer.ipv4('1.2.3.4/8'), ec2.Port.HTTP);
 ```
 
 You can also customize the characteristics of your NAT instances, including their security group,
@@ -266,7 +266,7 @@ const provider = ec2.NatProvider.instance({
 new ec2.Vpc(this, 'TheVPC', {
   natGatewayProvider: provider,
 });
-provider.connections.allowFrom(ec2.Peer.ipv4('1.2.3.4/8'), ec2.Port.tcp(80));
+provider.connections.allowFrom(ec2.Peer.ipv4('1.2.3.4/8'), ec2.Port.HTTP);
 ```
 
 ### Ip Address Management
@@ -724,13 +724,13 @@ declare const appFleet: autoscaling.AutoScalingGroup;
 declare const dbFleet: autoscaling.AutoScalingGroup;
 
 // Allow connections from anywhere
-loadBalancer.connections.allowFromAnyIpv4(ec2.Port.tcp(443), 'Allow inbound HTTPS');
+loadBalancer.connections.allowFromAnyIpv4(ec2.Port.HTTPS, 'Allow inbound HTTPS');
 
 // The same, but an explicit IP address
-loadBalancer.connections.allowFrom(ec2.Peer.ipv4('1.2.3.4/32'), ec2.Port.tcp(443), 'Allow inbound HTTPS');
+loadBalancer.connections.allowFrom(ec2.Peer.ipv4('1.2.3.4/32'), ec2.Port.HTTPS, 'Allow inbound HTTPS');
 
 // Allow connection between AutoScalingGroups
-appFleet.connections.allowTo(dbFleet, ec2.Port.tcp(443), 'App can call database');
+appFleet.connections.allowTo(dbFleet, ec2.Port.HTTPS, 'App can call database');
 ```
 
 ### Connection Peers
@@ -747,7 +747,7 @@ peer = ec2.Peer.anyIpv4();
 peer = ec2.Peer.ipv6('::0/0');
 peer = ec2.Peer.anyIpv6();
 peer = ec2.Peer.prefixList('pl-12345');
-appFleet.connections.allowTo(peer, ec2.Port.tcp(443), 'Allow outbound HTTPS');
+appFleet.connections.allowTo(peer, ec2.Port.HTTPS, 'Allow outbound HTTPS');
 ```
 
 Any object that has a security group can itself be used as a connection peer:
@@ -758,9 +758,9 @@ declare const fleet2: autoscaling.AutoScalingGroup;
 declare const appFleet: autoscaling.AutoScalingGroup;
 
 // These automatically create appropriate ingress and egress rules in both security groups
-fleet1.connections.allowTo(fleet2, ec2.Port.tcp(80), 'Allow between fleets');
+fleet1.connections.allowTo(fleet2, ec2.Port.HTTP, 'Allow between fleets');
 
-appFleet.connections.allowFromAnyIpv4(ec2.Port.tcp(80), 'Allow from load balancer');
+appFleet.connections.allowFromAnyIpv4(ec2.Port.HTTP, 'Allow from load balancer');
 ```
 
 ### Port Ranges
@@ -770,6 +770,7 @@ the connection specifier:
 
 ```ts
 ec2.Port.tcp(80)
+ec2.Port.HTTPS
 ec2.Port.tcpRange(60000, 65535)
 ec2.Port.allTcp()
 ec2.Port.allIcmp()
@@ -823,7 +824,7 @@ const mySecurityGroupWithoutInlineRules = new ec2.SecurityGroup(this, 'SecurityG
   disableInlineRules: true
 });
 //This will add the rule as an external cloud formation construct
-mySecurityGroupWithoutInlineRules.addIngressRule(ec2.Peer.anyIpv4(), ec2.Port.tcp(22), 'allow ssh access from the world');
+mySecurityGroupWithoutInlineRules.addIngressRule(ec2.Peer.anyIpv4(), ec2.Port.SSH, 'allow ssh access from the world');
 ```
 
 ### Importing an existing security group
--- a/packages/aws-cdk-lib/aws-ec2/lib/port.ts
+++ b/packages/aws-cdk-lib/aws-ec2/lib/port.ts
@@ -188,6 +188,41 @@ export interface PortProps {
  * Interface for classes that provide the connection-specification parts of a security group rule
  */
 export class Port {
+  /** Well-known SSH port (TCP 22) */
+  public static readonly SSH = Port.tcp(22);
+  /** Well-known SMTP port (TCP 25) */
+  public static readonly SMTP = Port.tcp(25);
+  /** Well-known DNS port (UDP 53) */
+  public static readonly DNS_UDP = Port.udp(53);
+  /** Well-known DNS port (TCP 53) */
+  public static readonly DNS_TCP = Port.tcp(53);
+  /** Well-known HTTP port (TCP 80) */
+  public static readonly HTTP = Port.tcp(80);
+  /** Well-known POP3 port (TCP 110) */
+  public static readonly POP3 = Port.tcp(110);
+  /** Well-known IMAP port (TCP 143) */
+  public static readonly IMAP = Port.tcp(143);
+  /** Well-known LDAP port (TCP 389) */
+  public static readonly LDAP = Port.tcp(389);
+  /** Well-known HTTPS port (TCP 443) */
+  public static readonly HTTPS = Port.tcp(443);
+  /** Well-known SMB port (TCP 445) */
+  public static readonly SMB = Port.tcp(445);
+  /** Well-known IMAPS port (TCP 993) */
+  public static readonly IMAPS = Port.tcp(993);
+  /** Well-known POP3S port (TCP 995) */
+  public static readonly POP3S = Port.tcp(995);
+  /** Well-known Microsoft SQL Server port (TCP 1433) */
+  public static readonly MSSQL = Port.tcp(1433);
+  /** Well-known NFS port (TCP 2049) */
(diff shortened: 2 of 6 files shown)


packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/aws-custom-resource.ts
```diff
@@ -394,6 +394,16 @@ export interface AwsCustomResourceProps {
    * @default - the Vpc default strategy if not specified
    */
   readonly vpcSubnets?: ec2.SubnetSelection;
+
+  /**
+   * Whether or not to include API response data in the logs for the underlying lambda.
+   *
+   * If this value is false, the raw API response and the `Data` field in the response
```
vinayak-kukreja, line 401: I think we should add an example of `Response Object` here so that we explicitly mention what we will still be logging with this. And also add this to our documentation

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -607,9 +607,16 @@ new cr.AwsCustomResource(this, 'ListObjects', {
 Note that even if you restrict the output of your custom resource you can still use any
 path in `PhysicalResourceId.fromResponse()`.
 
-### Custom Resource Examples
+### Custom Resource Logging
+
+The underlying lamdda function used by the `AwsCustomResource` construct logs the following information:
```
vinayak-kukreja, line 612: NIT
```suggestion
The underlying lambda function used by the `AwsCustomResource` construct logs the following information:
```

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -607,9 +607,16 @@ new cr.AwsCustomResource(this, 'ListObjects', {
 Note that even if you restrict the output of your custom resource you can still use any
 path in `PhysicalResourceId.fromResponse()`.
 
-### Custom Resource Examples
+### Custom Resource Logging
```
vinayak-kukreja, line 610: Lets add an example Response Object in these docs

vinayak-kukreja (changes requested):
I agree with the changes. But, we should get consensus among the team too about this. Especially stakeholders you have already synced with about this.

packages/aws-cdk-lib/custom-resources/README.md
colifran: Good call out. This has been updated.

packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/aws-custom-resource.ts
colifran: Updated.

packages/aws-cdk-lib/custom-resources/README.md
colifran: Updated.

TheRealAmazonKendra:
Did you close the original PR altogether? I was looking for my comments on it but couldn't find them. This still has the issue of this being a boolean without an option for additional customization.

colifran:
@TheRealAmazonKendra Can you provide more detail on why you think that? We can provide this flag and still keep the door open for additional customization. This flag would only control the logging of the `Data` field in the response object. Any additional customization could come with a future PR and this flag could be set or unset without blocking a future property that focuses on customization.

Say, for example, we wanted to add something to control log levels. If `logApiResponseData` is set to false and the log level is `DEBUG` then CloudWatch Logs still has all errors (caught and uncaught), the event received by the handler, and the response object (just without the `Data` field). That would open the door for a user to have all the benefits of a `DEBUG` log level while making an API call that returns sensitive data. Except now they don't have to worry about exposing data in CloudWatch Logs that they want hidden. If that isn't an issue for them then they can ignore `logApiResponseData` or explicitly set it to `true`.

The previous PR is linked [here](https://github.com/aws/aws-cdk/pull/29596). The flag proposed in that PR was too overbearing and would have blocked us from implementing additional customization in the future (as you correctly stated). I don't think this has that problem.

colifran:
@TheRealAmazonKendra pushed initial changes we discussed offline. To summarize we should provide a `Logging` class that consolidates all of logging levels into an enum-like class. Additionally, this doesn't need to follow standard logging convention (info, trace, debug, etc.). We can provide something like `on`, `off`, and `selective` where `selective` would give granular control over logged data to the user.

piradeepk (commented):
Looks great, left a couple minor comments.

packages/aws-cdk-lib/custom-resources/test/aws-custom-resource/aws-custom-resource.test.ts
```diff
@@ -46,13 +46,23 @@ test('aws sdk js custom resource with onCreate and onDelete', () => {
       'physicalResourceId': {
         'id': 'loggroup',
       },
+      'logHandlerEvent': true,
+      'logApiResponse': true,
+      'logResponseObject': true,
+      'logSdkVersion': true,
+      'logErrors': true,
```
colifran, line 53: Calling out here that while this is an existing unit test being updated, it is not a breaking change. `AwsSdkCall` now has logging configured as `on` by default. Thus, the event passed to the Lambda has additional properties. This maintains the current default behavior in the handler code by logging all data by default.

packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/logging.ts
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * Whether or not to log errors that were encountered during lambda execution.
+   *
+   * @default false
+   */
+  readonly logErrors?: boolean;
+}
+
+/**
+ * A class used to configure Logging during for AwsCustomResource SDK calls.
```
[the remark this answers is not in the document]
colifran: Yes, agreed and updated.

packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/logging.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+export abstract class Logging {
+  /**
+   * Enables logging of all logged data in the lambda handler.
+   *
+   * This includes the event object, the API call response, all fields in the response object
+   * returned by the lambda, and any errors encountered.
+   */
+  public static on(): Logging {
```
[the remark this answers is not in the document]
colifran: Yeah I agree. Aligns better with the verbiage too. Updated.

packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/aws-custom-resource.ts
```diff
@@ -188,6 +189,13 @@ export interface AwsSdkCall {
    * @default - run without assuming role
    */
   readonly assumedRoleArn?: string;
+
+  /**
+   * A property used to configure logging during lambda function execution.
+   *
+   * @default Logging.on()
```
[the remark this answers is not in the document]
colifran: I think the more detail the better. I've updated this. Let me know what you think.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources/aws-custom-resource-handler/aws-sdk-v2-handler.ts
```diff
@@ -62,6 +62,16 @@ function patchSdk(awsSdk: any): any {
...
 export async function handler(event: AWSLambda.CloudFormationCustomResourceEvent, context: AWSLambda.Context) {
+  event.ResourceProperties.Create = decodeCall(event.ResourceProperties.Create);
+  event.ResourceProperties.Update = decodeCall(event.ResourceProperties.Update);
+  event.ResourceProperties.Delete = decodeCall(event.ResourceProperties.Delete);
+
+  const call: AwsSdkCall | undefined = event.ResourceProperties[event.RequestType];
+
+  // if there is a call there will always be logging configured -- otherwise, in the event of no call, logging
+  // wasn't configure so just default to existing behavior
+  const logApiResponseData = call?.logApiResponseData ?? true;
```
piradeepk, line 73: Do we need this in the v2 handler (I don't see where we use it)?

colifran: A bit further down (lines 172 and 177). This is passed to `respond` which is in `shared.ts`. The v2 handler and the v3 handler share logic contained in `shared.ts`

piradeepk: Got it! Lgtm!

packages/@aws-cdk-testing/framework-integ/test/aws-synthetics/test/integ.canary-auto-delete-lambda.ts
```diff
@@ -10,7 +10,7 @@ class TestStack extends Stack {
 
     new synthetics.Canary(this, 'Canary', {
       canaryName: 'next',
-      runtime: synthetics.Runtime.SYNTHETICS_NODEJS_PUPPETEER_4_0,
+      runtime: synthetics.Runtime.SYNTHETICS_NODEJS_PUPPETEER_7_0,
```
colifran, line 13: `Runtime.SYNTHETICS_NODEJS_PUPPETEER_4_0` was causing the integ test to fail due to this being a deprecated runtime. The error message was as follows:

```
Invalid request provided: Deprecated runtime version specified.
```

I've updated the runtime to the latest.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/emrcontainers/integ.start-job-run.ts
```diff
@@ -22,7 +22,7 @@ const stack = new cdk.Stack(app, 'aws-stepfunctions-tasks-emr-containers-start-j
 stack.node.setContext(EC2_RESTRICT_DEFAULT_SECURITY_GROUP, false);
 
 const eksCluster = new eks.Cluster(stack, 'integration-test-eks-cluster', {
-  version: eks.KubernetesVersion.V1_22,
+  version: eks.KubernetesVersion.V1_29,
```
colifran, line 25: `KubernetesVersion.V1_22` was causing the integ test to fail since version 1.22 is no longer supported. The integ test was failing with the following error message:

```
Unsupported Kubernetes version 1.22
```

I've update to the latest version.

colifran:
A number of unit tests needed to be updated as a result of the new `logging` property added to `AwsSdkCall`. This does not represent a breaking change. `logging` is optional, but will default to `Logging.all()` which represents the current logging behavior in the v2 and v3 handlers. `Logging.all()` adds the `logApiResponseData: true` to the event object passed to the `AwsCustomResource` provider Lambda's handler by default. This is a new field that is used for configurable `logging`.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources/aws-custom-resource-handler/construct-types.ts
```diff
@@ -19,4 +21,5 @@ export interface AwsSdkCall {
   readonly outputPath?: string;
   readonly outputPaths?: string[];
   readonly assumedRoleArn?: string;
-}
\ No newline at end of file
+  readonly logApiResponseData: boolean;
```
scanlonp, line 24: Question on if this property should be optional or not. By the way we have set up the logic with `_render()`, we will always have a value for this field, so it can be "required". 

@colifran and I talked offline, I do not think that this is a problem. The main discussion is at what level we decide to set the default, and having it passed explicitly to the call is fine.

scanlonp (approved):
One comment as a point of discussion. But overall good from me.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources/aws-custom-resource-handler/construct-types.ts
colifran: For historical purposes, I want to make sure it is clear that the `AwsSdkCall` interface referenced above is not a type used in the `aws-cdk-lib`. Instead, this is a mirror of `AwsSdkCall` which is defined as part of the `AwsCustomResource` construct. However, the format of this represents the `AwsSdkCall` type as it would be received by the v2 and v3 handler, e.g., after `render` has been called on `Logging` and all `logging` properties are rendered. This interface just exists to provide a strongly typed `call` in the v2 and v3 handlers.

My opinion is that we should set the default prior to passing the event so that the event properties are visible in the synthesized CloudFormation template.

PR #29652, opened 2024-03-29
docs(cli): typos in flagname, make test description accurate

Smeb:
Tiny fixup for code documentation - I was using the wrong flag because I took it from the comments on the diff command.

I also changed the description for a test, since it doesn't actually check the flag name.


comcalvi (dismissed):

Smeb:
Ah shoot - I hit `update` before reading the comment. My bad

packages/aws-cdk/test/diff.test.ts
```diff
@@ -336,7 +336,7 @@ describe('non-nested stacks', () => {
     expect(exitCode).toBe(0);
   });
 
-  test('diff does not check for stack existence when --no-changeset is passed', async () => {
+  test('diff does not check for stack existence when changeSet is false', async () => {
```
scanlonp, line 339:
```suggestion
  test('diff does not check for stack existence when --no-change-set is passed, async () => {
```

Smeb: Just a comment on this - I removed that line because we don't actually check the flags - we just call diff with preconfigured arguments which don't map to the actual flag names.

comcalvi (dismissed):

TheRealAmazonKendra:
@Mergifyio update

TheRealAmazonKendra:
I'm not sure what the deal is with all the issues deploying our tests but it's certainly not the content of this PR. Let's see if I can get this merged.

TheRealAmazonKendra:
Actually, it looks like we've made changes since this PR was originally made and the comment added is no longer relevant. I'm going to go ahead and close this one.

PR #29403, opened 2024-03-07
fix(stepfunctions): the retry field in CustomState is not iterable

abdelnn:
### Issue # (if applicable)

Closes #29274 

### Reason for this change

CDK users were unable to specify their retry strategy if it was specified inline in their ASL state machine definition

### Description of changes

Checks if the state definition has an inline retry policy defined. If it does, add it to the existing strategy defined using `addRetry` (if there is one defined, this is where it was failing before)

### Description of how you validated changes

Added unit and integration tests

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #29654, opened 2024-03-29
fix(stepfunctions): the catch field in CustomState is not rendered

jormello:
### Issue # (if applicable)

N/A

### Reason for this change

Customers that specify `Catch` fields in their CustomState's `stateJson` do not have Catchers defined in the rendered state definition. The reason for this is that the `Catch` fields from the `stateJson` is overridden by Catchers added through `addCatch()`.

### Description of changes

This change updates the way the state's `Catch` field is rendered to merge Catchers provided inline with those provided through `addCatch()`. Catchers from `addCatch()` will be rendered first, followed by those provided inline. This is consistent with the merge behaviour for Retriers. 

Unit test coverage for Catchers provided just inline, just through `addCatch()`, and a combination of both. 

### Checklist

----


packages/aws-cdk-lib/aws-stepfunctions/test/custom-state.test.ts
```diff
@@ -1,8 +1,9 @@
 import { render } from './private/render-util';
 import * as cdk from '../../core';
 import * as sfn from '../lib';
+import { Errors } from '../lib/types';
 
-describe('Custom State', () => {
+describe.only('Custom State', () => {
```
wong-a, line 6: Should not skip other tests

wong-a (changes requested):

packages/aws-cdk-lib/aws-stepfunctions/test/custom-state.test.ts
```diff
@@ -309,4 +310,145 @@ describe('Custom State', () => {
       },
     );
   });
+
+  test('expect retry to merge when specifying strategy inline and through construct', () => {
```
abdelnn, line 314: (non-blocking) this is already covered as part of this test:

 https://github.com/aws/aws-cdk/blob/8ede40976392893a4a7c0e90f3d6638209670fb6/packages/aws-cdk-lib/aws-stepfunctions/test/custom-state.test.ts#L168

abdelnn (approved):

wong-a (approved):

packages/aws-cdk-lib/aws-stepfunctions/lib/states/custom-state.ts
```diff
@@ -74,11 +74,16 @@ export class CustomState extends State implements IChainable, INextable {
...
+    // Retriers and Catchers can be specified directly in the stateJson or indirectly to the construct with addRetry() and addCatch().
+    // renderRetryCatch() only renders the indirectly supplied Retriers and Catchers, so we need to manually merge in those directly in the stateJson
     if (Array.isArray(this.stateJson.Retry)) {
       state.Retry = Array.isArray(state.Retry) ? [...state.Retry, ...this.stateJson.Retry] : [...this.stateJson.Retry];
     }
 
+    if (Array.isArray(this.stateJson.Catch)) {
+      state.Catch = Array.isArray(state.Catch) ? [...state.Catch, ...this.stateJson.Catch] : [...this.stateJson.Catch];
+    }
+
```
aaythapa, line 86: Question I probably should've asked when reviewing https://github.com/aws/aws-cdk/pull/29403 but is there a reason we allow customers to define both and not just throw an error here saying they can only define one?

aaythapa (commented):
Thanks for making this change!

abdelnn (approved):

packages/aws-cdk-lib/aws-stepfunctions/lib/states/custom-state.ts
```diff
@@ -74,11 +75,64 @@ export class CustomState extends State implements IChainable, INextable {
...
     };
 
-    // merge the Retry field defined in the stateJson into the state
+    if (this.hasMultipleRetrySources(state)) {
+      this.addMultipleRetrySourcesWarning();
+    }
+
+    if (this.hasMultipleCatchSources(state)) {
+      this.addMultipleCatchSourcesWarning();
+    }
```
aaythapa, line 84: We check to see if we have multiple retry/catch sources to add warnings then check again to generate the state. Can we do 1 check, send a warning then generate the stateJson

packages/aws-cdk-lib/aws-stepfunctions/lib/states/custom-state.ts
aaythapa: Discussed offline, we decided the best way would be for addRetry and addCatch to have the same behavior but to add warnings so the customer understands how the merge works.

packages/aws-cdk-lib/aws-stepfunctions/lib/states/custom-state.ts
aaythapa: I meant can't we do something like this in line 92
```suggestion
   if (Array.isArray(this.stateJson.Catch)) {
        if(Array.isArray(state.Catch)) {
               this.addMultipleCatchSourcesWarning();
               state.Catch = [...state.Catch, ...this.stateJson.Catch];
        }
        else
        {
               state.Catch = [...this.stateJson.Catch];
        }
    }
```

jormello: As mentioned, we initialize the state's `catch` field as an empty array, so even if customers do not call `addCatch()` on their custom state they will incorrectly be flagged for providing both sources. The same is true for retries.

aaythapa: Talked offline, I was confused by the `if(Array.isArray(state.Catch))` since its always an array but its only there to act as a type check which i believe is fine

aaythapa (approved):

aaythapa (approved):

PR #29669, opened 2024-04-01
fix(triggers): store provider as class property

### LINKED ISSUE #29657: triggers: Trigger provider does not show up in construct tree
### Describe the bug

The Trigger provider construct and its children don't seem to be showing up in the Construct tree. This means that it can't be used with Aspects to apply tags or execute any kind of overrides. 

### Expected Behavior

1. 
```typescript
const triggerCustomResourceProvider = stack.node.tryFindChild(
      "AWSCDK.TriggerCustomResourceProviderCustomResourceProvider",
    );
```
This should return the trigger resource provider.

2. Similarly, this should allow me to make modifications to the Trigger resources
```typescript
Aspects.of(stack).add(new SomeModificationToTrigger());
```

### Current Behavior

1. `tryFindChild` returns undefined
2. Aspects does not pass the class any of the children resources of the Trigger Provider
3. Listing the entire construct tree via `console.log(this.node.findAll().map((n) => n.node.path));` does not show the Trigger provider constructs

### Reproduction Steps

This was tested in two places: Fine-grained assertion tests and Python code using the published code from a custom CDK library.

### Possible Solution

Possibly the use of a `.` in the [construct ID](https://github.com/aws/aws-cdk/blob/df48fd781a476ada79b6b49f006a60f33f73d409/packages/aws-cdk-lib/triggers/lib/trigger.ts#L117) is causing this? 

_No response_

### CDK CLI Version

2.131.0

### Framework Version

_No response_

### Node.js Version

20

### OS

macOS Sonoma

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

automartin5000:
### Issue # (if applicable)

Closes #29657

### Reason for this change

Unable to modify Triggers provider using conventional escape hatches

### Description of changes

Stored Trigger provider as class property to be user-accessible

Was unable to test changes due to issues getting project installed locally

### Checklist

----


automartin5000:
Clarification Request
I can't tell why builds are failing. There seem to be a lot of unrelated failures? The only thing I can tell is a linting error in the Trigger file, but that looks to have already been there?

Also does this need an integration test beyond the tests I already added? If so where would it go/what would it do?

packages/aws-cdk-lib/triggers/lib/trigger.ts
```diff
@@ -111,20 +111,25 @@ export interface TriggerProps extends TriggerOptions {
  * Triggers an AWS Lambda function during deployment.
  */
 export class Trigger extends Construct implements ITrigger {
+  /**
+   * Singleton Trigger provider
+   * @type {CustomResourceProvider} 
+   */
+  crProvider: CustomResourceProvider;
```
TheRealAmazonKendra, line 118: Can we either call this `provider` or use the whole `customResourceProvider`?

TheRealAmazonKendra (changes requested):
Thank you for your contribution! I have just one comment inline but I think we also need to see an integ test added for this to show that it works for your intended use case.

TheRealAmazonKendra:
Also, please note that the title of a fix should describe the bug, not the solution. Could you please adjust that?

TheRealAmazonKendra:
@Mergifyio update

TheRealAmazonKendra:
Also, I hadn't noticed that your build failed and that you asked a question about it. The failure lines here are pretty hard to find but they're:
```
aws-cdk-lib:   116:36  error  Trailing spaces not allowed  no-trailing-spaces
aws-cdk-lib: âœ– 1 problem (1 error, 0 warnings)
aws-cdk-lib:   1 error and 0 warnings potentially fixable with the `--fix` option.
```
The docstring on line 116 of `trigger.ts` is running afoul of our linter. The other lines that look like failures are just jsii warnings. They are noisy but they don't cause failures.

PR #29680, opened 2024-04-01
feat(lambda): adding support for Ruby3.3 lambda runtime

jonife:
Adds `ruby 3.3` to the available [Lambda Runtimes constants](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.Runtime.html#initializer)


xazhao (approved):
Looks good to me!

PR #29681, opened 2024-04-01
chore(lambda): upgrade default runtime for custom resource provider and function to nodejs_20

shikha372:
### Issue # (if applicable)

Closes: https://github.com/aws/aws-cdk/issues/28125
Tracking Ticket: https://github.com/aws/aws-cdk/issues/29786

### Reason for this change

Keeping dependencies up to date with latest release in custom resource provider function and lambda default runtime value for Nodejs

### Description of changes

Update default runtime defined in HandlerFrameworkModule to update custom resource provider
Update runtime for custom resource provider still using CfnResource.
Update for providers in alpha modules.
Unit Test updated with expected runtime value
Update to expected runtime value in integ test along with snapshots.

Ran Integration Tests(framework-integ) to validate custom resources behaviour after the upgrade.

### Checklist

----


packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
```diff
@@ -101,6 +101,7 @@ function createEvalFn(runtime: lambda.Runtime | undefined, scope: Construct) {
   const lambdaPurpose = 'Eval';
 
   const nodeJsGuids = {
+    [lambda.Runtime.NODEJS_20_X.name]: 'f66698ab-28ac-4879-83d1-615be69028bc',
```
GavinZZ, line 104: Curious where is this id from?

GavinZZ (commented):
Curious how do you confirm that we didn't miss anywhere?

aaythapa (approved):

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
shikha372: confirmed with @colifran , these are random GUIDs, this one is generated using https://www.uuidgenerator.net/guid

shikha372:
> how do you confirm that we didn't miss anywhere?

ran `yarn build` and `yarn test` on local under aws-cdk-lib and @aws-cdk/custom-resource-handlers to confirm that unit tests are passing, changes to the unit tests are confirmed on the basis of that. 
Also, we verify the generated providers created under aws-cdk-lib and aws-cdk `dist` folder on running build, these generated handlers have nodejs20 as their runtime.
example path: ~/aws-cdk/packages/@aws-cdk/custom-resource-handlers/dist/aws-cloudfront/cross-region-string-param-reader-provider.generated.ts
Second verification will be with the below command  under each of this package to confirm nodejs18 is mentioned only in expected places 
`grep -irl 'nodejs18' ` or `grep -irl 'NODEJS_20_X'`

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/evaluate-expression.ts
colifran: These are just uuids. The `SingletonFunction` construct uses this to help determine if a`Function` has already been created within the same scope. As a result, I think the only requirement is that the uuid assigned needs to be unique.

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/evaluate-expression.test.ts
```diff
@@ -89,17 +89,17 @@ test('with dash and underscore in path', () => {
   });
 });
 
-test('With Node.js 18.x', () => {
+test('With Node.js 20.x', () => {
```
colifran, line 92: Generally we don't want to change any existing tests. If they were accepted at one point they should continue working event with future changes otherwise it signal that we're making a breaking change. I would keep the existing node18 test and add a new test for node20. In fact, maybe we just want to keep an array of runtime versions and use `forEach` to test each case. As an example - https://github.com/aws/aws-cdk/blob/f4b825e34f08d29bea19fe29be422fe6143acabd/packages/%40aws-cdk/aws-glue-alpha/test/job-executable.test.ts#L132

colifran (changes requested):
Looks good so far! This needs to be included too - You may be working on it but this should be included too - https://github.com/aws/aws-cdk/blob/00e8a7b877a732b4236227f6462bf3914ce7a214/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/providers/provider.ts#L81

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/evaluate-expression.test.ts
shikha372: Thanks @colifran ,I updated in lambda-node-js too, let me fix in both cases

shikha372: EvalNodejsSingletonFunction has now runtime as nodejs20 updated in generated provider through framework:
aws-cdk/packages/aws-cdk-lib/custom-resource-handlers/dist/aws-stepfunctions-tasks/eval-nodejs-provider.generated.ts

` "runtime": lambda.Runtime.NODEJS_20_X`

 even though below function is taking the runtime as input props, only the uuid is being passed to provider, runtime remains `nodejs20`, so test case for Node.js 18x seems to be no longer valid , let me know if my understanding is not correct
```
function createEvalFn(runtime: lambda.Runtime | undefined, scope: Construct) 

.......
return new EvalNodejsSingletonFunction(scope, 'EvalFunction', {
    uuid,
    lambdaPurpose,
  });
```

```
export interface EvalNodejsSingletonFunctionProps extends lambda.FunctionOptions {
  readonly uuid: string;
  readonly lambdaPurpose?: string;
}
```

packages/aws-cdk-lib/aws-lambda-nodejs/test/bundling.test.ts
```diff
@@ -11,8 +11,8 @@ import { PackageInstallation } from '../lib/package-installation';
 import { Charset, LogLevel, OutputFormat, SourceMapMode } from '../lib/types';
 import * as util from '../lib/util';
 
-const STANDARD_RUNTIME = Runtime.NODEJS_18_X;
-const STANDARD_TARGET = 'node18';
+const STANDARD_RUNTIME = Runtime.NODEJS_20_X;
+const STANDARD_TARGET = 'node20';
```
shikha372: These test changes are fixed in latest commit to update only expected target value for nodejs latest runtime.

shikha372:
> Looks good so far! You may be working on it but this should be included too -
> 
> https://github.com/aws/aws-cdk/blob/00e8a7b877a732b4236227f6462bf3914ce7a214/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/providers/provider.ts#L81

yes, no changes done in 

> Looks good so far! You may be working on it but this should be included too -
> 
> https://github.com/aws/aws-cdk/blob/00e8a7b877a732b4236227f6462bf3914ce7a214/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/providers/provider.ts#L81

Added in latest commit

packages/aws-cdk-lib/aws-lambda/lib/runtime.ts
```diff
@@ -113,7 +113,7 @@ export class Runtime {
    * The latest NodeJS version currently available in ALL regions (not necessarily the latest NodeJS version
    * available in YOUR region).
    */
-  public static readonly NODEJS_LATEST = new Runtime('nodejs18.x', RuntimeFamily.NODEJS, { supportsInlineCode: true, isVariable: true });
+  public static readonly NODEJS_LATEST = new Runtime('nodejs20.x', RuntimeFamily.NODEJS, { supportsInlineCode: true, isVariable: true });
```
TheRealAmazonKendra, line 116: Instead of hardcoding node20 in the many places we have 18 currently hardcoded, can we just use this?

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/utils/framework-utils.ts
```diff
@@ -13,6 +13,9 @@ export function buildComponentName(fqn: string, type: ComponentType, entrypoint:
 
 export function toLambdaRuntime(runtime: Runtime) {
   switch (runtime) {
+    case Runtime.NODEJS_20_X: {
```
TheRealAmazonKendra, line 16: What do you think about adding a `LATEST` in here and pointing that to node20 as well?

TheRealAmazonKendra (changes requested):
Just a couple of comments but overall I think this is looking good so far. One quick question I have is what the scope of this task is intended to be. We have some testing infra that is likely hardcoded to node18 (not meaning the framework integ tests that I had asked to be split out, other places). So, fi this change is meant to be exhaustive for the repo, some things may have been missed.

shikha372:
> intended to be. We have some testing infra that is likely hardcoded to node18 (not meaning t

Hey @TheRealAmazonKendra ,
Thanks for the review, as I understand scope of this work is to update the runtime to nodejs20 for custom resources and  lambda.
do you mean there are some places in other unit tests as well which might be still using the node18 ? shouldn't dependent tests fail if we ran `yarn test` in `aws-cdk-lib` directory, unless it is hardcoded to use something other than `NODEJS_LATEST` in the service itself.

packages/aws-cdk-lib/aws-lambda/lib/runtime.ts
shikha372: Along with introducing NODEJS_LATEST in resource framework, modified this in alpha modules and other places in this PR to point to this except in Readme above(assuming documentation should reflect the latest version being supported).

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/utils/framework-utils.ts
shikha372: makes sense after pointing the DEFAULT_RUNTIME  in framework to Runtime.NODEJS_LATEST, next time when we upgrade we'll only need to cover another type in enum and add it here, framework should pick the latest version from this value. 
Updated in latest commit for review.

packages/aws-cdk-lib/aws-lambda-nodejs/test/bundling.test.ts
```diff
@@ -13,6 +13,7 @@ import * as util from '../lib/util';
 
 const STANDARD_RUNTIME = Runtime.NODEJS_18_X;
 const STANDARD_TARGET = 'node18';
+const LATEST_TARGET = 'node20';
```
TheRealAmazonKendra, line 16: Can we just change `STANDARD_TARGET` and `STANDARD_RUNTIME` instead of adding `LATEST_TARGET`?

TheRealAmazonKendra (changes requested):
Just one quick comment

TheRealAmazonKendra (approved):
As long as all the integ tests pass, we should be good.

TheRealAmazonKendra:
Well, I've got some unfortunate news, I was searching for something unrelated to this change and discovered that node20 isn't available in GovCloud, ADC, or China regions so in places where `DEFAULT` was specifying node18 and anywhere that the version is not configurable to the user still needs to be that.

colifran:
@TheRealAmazonKendra Given that news it seems like moving forward here doesn't make sense until node20 is available in GovCloud, ADC, or China regions.

PR #29683, opened 2024-04-01
chore(CLI): add `--changeset` as an alias for `--change-set`

comcalvi:
Some customers have typo'd `--no-changeset` where `--no-change-set` was expected. Make these two aliases of each other since they're easy to mix up, and this flag being valid changes the permissions used by diff, which will cause errors in certain environments. 

tested manually.

----


scanlonp (approved):

scanlonp:
This alias should not need a test pipeline run.

Issue #13526, opened 2021-03-10
(aws-apigateway): RestApi imported with method fromRestApiAttributes not updating current deployment in stage


evilsephiroth:
I've imported from a serverless project this restApi.
```
const restApiId = cdk.Fn.importValue(envRelated(`ApiGateway-restApiId`));
const rootResourceId = cdk.Fn.importValue(envRelated(`ApiGateway-rootResourceId`));
this.apiGw = apigateway.RestApi.fromRestApiAttributes(this, envRelated(`apigw`), {
	restApiId: restApiId,
	rootResourceId: rootResourceId,
});
```
When deploying with cdk deploy, all resources are updated but not the stage.
No deployment is created.
I've tried also to force a deployment with this snippet.
```
new Deployment(this,envRelated('deployment-stage') , {
      api: apiGw
    });
```
Salting the deployment id makes no difference.
After the deployment is done, the api gateway is still pointing to the previous deployment instead of the new deployment done.

![chrome_sGUK4TBoqT](https://user-images.githubusercontent.com/199461/110676506-52f88280-81d4-11eb-89b7-e12889627b5a.png)

Specifying the stage with the forced deployment fails because stage already exists.

### Environment

  - **CDK CLI Version:** 1.9.2
  - **Module Version:** 1.9.2
  - **Node.js Version:** 12.14.1
  - **OS:** Ubuntu Linux 20.04
  - **Language: ts

baumannalexj:
```
const apiDeployment = new cdkApiGateway.Deployment(this, `api-deployment`, {
            api: <importedApi>,
            description: `maybe put datetime or artifact version here`,
            retainDeployments: true,
        });
        apiDeployment.addToLogicalId(cdkConfig.DATETIME_ISO); //need force change deployment hash to force new deployment

```

and for a specific stage
```
        apiDeployment.resource.stageName = cdkConfig.ENV_UPPER;          // https://stackoverflow.com/a/64369331/6791342 // if the 'stageName' already exists (from the core apigateway deployment) then the existing stage will be used !


```

evilsephiroth:
> @evilsephiroth try this
> 
> you need to bust the deployment sha with `addToLogicalId`
> 
> ```
> const apiDeployment = new cdkApiGateway.Deployment(this, `api-deployment`, {
>             api: <importedApi>,
>             description: `maybe put datetime or artifact version here`,
>             retainDeployments: true,
>         });
[...]

Issue #12417, opened 2021-01-08
(apigateway): support deployment salting for imported RestApi


taoatmars:
If the RestApi is define by RestApi.fromRestApiAttributes(), any deployment will not add salt to the deployment resource.

### Reproduction Steps
This is my code.
```
const api = RestApi.fromRestApiAttributes(this, 'ImportedApi', {
      restApiId: XXX,
      rootResourceId: XXX,
    });

    const allResources = api.root.addResource('cdk');

    allResources.addCorsPreflight({ allowOrigins: Cors.ALL_ORIGINS });

    const method = allResources.addMethod('PUT', ifn, {
      methodResponses: [
        {
          statusCode: '200',
          responseParameters: {
            'method.response.header.Access-Control-Allow-Origin': true,
          },
        },
      ],
      authorizationType: AuthorizationType.IAM,
    });

    // Handle deployment
    const deployment = new Deployment(this, '123_deployment', {
      api: RestApi.fromRestApiId(this, 'abc', apiCDKApiApiId),
    });
    new Stage(this, 'CDKApiStageDev', { deployment, stageName: `${ENV}` });
```

"taotest123deploymentC642C343" is the resource name in the Cloudformation template.

### What did you expect to happen?

add salt to the resource name

### What actually happened?

no salt was added to the resource name

### Environment

  - **CDK CLI Version  :*1.83*
  - **Framework Version:**
  - **Node.js Version:** 
  - **OS               :*MAC*
  - **Language (Version):*TypeScript* 
  
### Other

**WORKAROUND**: Just in case anyone ran into the same problem, you can force a deployment by manually adding salt to deployment id. in above example, 123_deployment is my deployment id, you can change the id 1234_deployment to force new deployment to deploy. 

--- 

This is :bug: Bug Report

nija-at:
Unfortunately, this isn't easy to implement since import APIs are intended to create API Gateway resources that are unaware of others. This is how we are able to split off these resources into different stacks and not cause cyclic dependencies.

Manually salting the deployment (as described in your workaround) is the best option available.

Moving this to a feature request.

taoatmars:
@nija-at Thanks for the reply, I know this is difficult to implement, but

tmclaugh:
Per https://github.com/aws/aws-cdk/issues/13383, if _latestDeployment_ were to be set when calling _RestApi.fromRestApiAttributes()_ this issue should be resolved without even needing to create a Deployment resource manually within the stack.

[...]

PR #29691, opened 2024-04-02
docs(apigateway): add warning about split stack technique

### LINKED ISSUE #29690: aws-apigateway: Documentation lacks information about stack separation for RestApi and Resources
### Overview

In this documentation, I would like to note that when separating the stacks for `RestApi` and `Resource`, `Deployment` is not automatically created, and to avoid this issue, it is necessary to use `Deployment.addToLogicalId()`.

### Details

When a `Resource` is added to an `IRestApi` imported via `RestApi.fromRestApiAttribute` or `RestApi.fromRestApiId`, `Deployment` is not created, so the latest resource state is not associated with the API Gateway stage, which is a known issue.
This has been discussed in several issues, including the following:

- https://github.com/aws/aws-cdk/issues/13526
- https://github.com/aws/aws-cdk/issues/29079

We understand that it is not easy to automatically create Deployment, and there may be differing opinions on whether it should be fixed.
However, since there is no mention of this issue in the documentation, we feel that many developers who are separating stacks related to API Gateway may unintentionally create an environment where `Deployment` is not created.
Due to this issue, developers need to manually deploy to API Gateway after running `cdk deploy`.

While it is possible to share a workaround implementation, it is difficult to implement an automatic `Deployment` creation that detects changes in `Resource` without increasing complexity in the current configuration.

Therefore, we would like to propose adding a note to the documentation first.

### Links

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway-readme.html#breaking-up-methods-and-resources-across-stacks

engineer-taro:
### Issue #29690

Closes #29690

### Reason for this change

Regarding the stack separation of RestApi and Resource, there is no documentation about the fact that Deployment is not automatically created. When I actually add resources to the code documented and try cdk deploy for the second time and beyond, a new deployment is not created, and the latest resources are not reflected.

### Description of changes

I added a note and related links to the documentation.

Nothing. It is just to change the description.

### Checklist

----


packages/aws-cdk-lib/aws-apigateway/README.md
```diff
@@ -267,6 +267,10 @@ The following example uses sets up two Resources '/pets' and '/books' in separat
 
 [Resources grouped into nested stacks](test/integ.restapi-import.lit.ts)
 
+> **Warning:** With the above code, no Deployment is created even if there are changes to the resources, and the latest resources are not reflected.
+To create a Deployment, you need to use [`Deployment.addToLogicalId()`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.Deployment.html#addwbrtowbrlogicalwbriddata).
+Currently, it is possible to work around this issue with a workaround implementation. Please refer to the [related issue](https://github.com/aws/aws-cdk/issues/13526).
```
go-to-k, line 272: I told you to write specific workarounds in the threads above, because I thought it's not clear to users what to do just to post the link to the issue.

If we write the workarounds, do we need to post the issue link here? Is there more information packed into this issue than that? (It's already closed, so just posting the link may cause users to close the page. If you want to include it, it might be a good idea to tell them what they should look for.)

go-to-k: (Is this a bug in the first place...? If it is a bug that needs to be resolved, perhaps the best course of action is to open a new issue and work on it. If it's not a bug, then I'm not sure, but maybe there's no need to post a link to the issue here...)

engineer-taro: @go-to-k 

Thank you very much for the detailed review. It is challenging to determine whether this is a bug, as similar issues have been raised multiple times in the past. In those cases, workarounds have been shared, and the issues have either been closed or remain unresolved due to the time required to address them.

The reason why the deployment is not created is that the object obtained by `RestApi.fromRestApiAttributes()` does not have dynamic attributes such as `latestDaployment`.

Reference Issues:
- https://github.com/aws/aws-cdk/issues/13526#issuecomment-1011216177
- https://github.com/aws/aws-cdk/issues/12417#issuecomment-1551338526

The concern is that developers considering stack splitting may find it difficult to discover this problem.

The current issue with the documentation is that it shares the implementation of stack splitting where automatic deployment is not performed. Would it be appropriate to add the implementation of `Deployment.addToLogicalId` to the sample code? 
However, even with this implementation, the issue of creating the deployment at the appropriate timing remains.

```
class DeployStack extends NestedStack {
  constructor(scope: Construct, props: DeployStackProps) {
    super(scope, 'integ-restapi-import-DeployStack', props);

    const deployment = new Deployment(this, 'Deployment', {
      api: RestApi.fromRestApiId(this, 'RestApi', props.restApiId),
    });
    if (props.methods) {
      for (const method of props.methods) {
        deployment.node.addDependency(method);
      }
    }
    const logicalId = // salted id
    deployment.addToLogicalId(logicalId)
    new Stage(this, 'Stage', { deployment });
  }
}
```

Alternatively, if this is a long-standing issue, should we wait for a resolution, even though the timeline is uncertain?

go-to-k: Thanks for clearing.

> Would it be appropriate to add the implementation of Deployment.addToLogicalId to the sample code?
> However, even with this implementation, the issue of creating the deployment at the appropriate timing remains.

I'm not sure, if the implementation is incomplete and remains the another issue, it may not be appropriate to recommend it... It would also be helpful to just state that deployment must be done manually when resources are changed. I think it's important to write only what we know completely.

> Alternatively, if this is a long-standing issue, should we wait for a resolution, even though the timeline is uncertain?

I don't see the need to wait all, as there is no guarantee that the solution will be found, and in the meantime, there will be new users who will be in trouble.

packages/aws-cdk-lib/aws-apigateway/README.md
```diff
@@ -267,6 +267,10 @@ The following example uses sets up two Resources '/pets' and '/books' in separat
 
 [Resources grouped into nested stacks](test/integ.restapi-import.lit.ts)
 
+> **Warning:** With the above code, no Deployment is created even if there are changes to the resources, and the latest resources are not reflected.
+To create a Deployment, you need to use [`Deployment.addToLogicalId()`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.Deployment.html#addwbrtowbrlogicalwbriddata).
```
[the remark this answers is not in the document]
go-to-k: This may not be necessary.

https://github.com/aws/aws-cdk/pull/29691#discussion_r1558839033

engineer-taro:
@go-to-k 

Thank you for your review.
I have made the following changes. I would appreciate it if you could check them when you have time.

- I have noted that an API Gateway deployment is created during the initial deployment.
- I have omitted mentioning the incomplete implementation of `addToLogicalId`.
- I have noted that manual deployment is necessary after the second and subsequent CDK deployments.

packages/aws-cdk-lib/aws-apigateway/README.md
```diff
@@ -267,6 +267,10 @@ The following example uses sets up two Resources '/pets' and '/books' in separat
 
 [Resources grouped into nested stacks](test/integ.restapi-import.lit.ts)
 
+> **Warning:** In the code above, an API Gateway deployment is created during the initial CDK deployment.
+However, if there are changes to the resources in subsequent CDK deployments, a new API Gateway deployment is not automatically created. As a result, the latest state of the resources is not reflected.
+To ensure the latest state of the resources is reflected, a manual deployment of the API Gateway is required after the CDK deployment.
```
go-to-k, line 272: It is good to insert line breaks at visually appropriate places in the document. (We can feel relief that these line breaks are not applied when they are actually displayed in the browser.)

```suggestion
> **Warning:** In the code above, an API Gateway deployment is created during the initial CDK deployment.
However, if there are changes to the resources in subsequent CDK deployments, a new API Gateway deployment is not
automatically created. As a result, the latest state of the resources is not reflected. To ensure the latest state
of the resources is reflected, a manual deployment of the API Gateway is required after the CDK deployment.
```

go-to-k (changes requested):
Thanks for your changes. I have one last comment. I will approve this as soon as it is done.

packages/aws-cdk-lib/aws-apigateway/README.md
engineer-taro: @go-to-k 
Thank you! Made corrections!

go-to-k (approved):
Thank you! Looks good to me.

aaythapa (approved):

aaythapa:
@mergify update

aaythapa (approved):

PR #29698, opened 2024-04-02
feat(s3): autoDeleteObjects log group allows retention period and removal policy definitions

### LINKED ISSUE #24815: (aws-s3): retention on S3 autoDeleteObjects lambda CloudWatch log group is Never expire
### Describe the bug

### Expected Behavior

### Current Behavior

### Reproduction Steps

```typescript
  autoDeleteObjects: true,
});
```

### Possible Solution

```typescript
  autoDeleteObjects: true,
});
```

_No response_

### CDK CLI Version

2.69.0 (build 60a5b2a)

### Framework Version

_No response_

### Node.js Version

v18.12.0

### OS

macOS 12.6

### Language

Typescript

### Language Version

4.9.5

### Other information

_No response_

kishiel:
### Issue #24815

Closes #24815

### Reason for this change

S3 bucket autoDeleteObjects leaves behind a log group for each bucket that uses the feature.  This results in a lot of cruft, especially in test accounts, which should be configurable by the bucket owner.  The account limit for log groups is 10,000 and I've got test accounts that have hit this limit several times.

### Description of changes

* Creates a log group rather than relying on the underlying custom-resource to create it automatically (a side effect of using CfnResource for AWS::Lambda::Function)
* Sets a default retention period of 90 days on the log group (I picked a number)
* Sets a default removal policy of delete on the log group (I don't think anyone wants these after they delete a bucket)
* Denies the custom-resource Lambda role permission to create a log group (prevents log group recreation on delete)
* Adds log group name as an optional to the interface of the custom-resource.  This is plumbed into the loggingConfig and results in an undefined entry if not provided.

Unit tests in addition to some simple functional tests.  

When making a bucket with autoDeleteObjects enabled I wanted to confirm that the log group for the lambda was, in fact, gone after I deleted the stack.  This is how I found that I needed to modify the permission of the Lambda role to deny log group creation.

I also confirmed that the custom-resources which do not provide a log group name still produce a log group and logs within.

Also, over 100 snapshot tests (RIP me).

### Checklist

----


kishiel:
There are about 10 snapshot tests which are failing that I'm unable to resolve on my own and I could use some help in running them. I believe that a few of them are because I'm using an internal AWS account so hopefully they just need to be run from someone's not-quite-so-special account.

kishiel:
Something appears to be misconfigured with the @aws-cdk/aws-route53resolver-alpha directory.  I can reproduce the problem with a single test in @aws-cdk/aws-route53resolver-alpha/test:

```
  const app = new App({
    context: {
      [cxapi.NEW_STYLE_STACK_SYNTHESIS_CONTEXT]: false,
    },
  });
  const stack = new Stack(app);
  const filePath = path.join(__dirname, 'domains.txt');
  new Asset(stack, 'MyAsset', { path: filePath });
```

Copying this same test over to @aws-cdk/aws-redshift-alpha/test does not cause the error to manifest.

mwebber:
@kishiel are you able to fix the merge conflicts, so your PR can proceed?

PR #29706, opened 2024-04-03
chore(rds): improve error messages for providing a vpc to cluster

prasanthlouis:
Just an idea/I was curious.

Personally, when I stumbled upon this error I got a bit confused because I wasn't sure what instanceProps was. I think the message  'Provide either vpc or instanceProps.vpc, but not both" provides the same explanation for both cases?

Or we rephrase the error message to not use "If instanceProps was provided", because the code can figure whether instanceProps was provided with another if condition Then the message would just be "VPC must be provided". Hence, someone who is unaware of instanceProps, doesn't need to be notified of it. Just an idea though and it's nitpicking! Thoughts?

### Reason for this change

Better error message when running into this error

### Description of changes

Consolidated logs statements

I have not validated anything because this was just an idea/I was curious

### Checklist
- [X ] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----


scanlonp:
Hey @prasanthlouis, I think you should write the error messages in a way that is the most clear to you. If the current message did not make sense, of course feel free to improve it!

I do not have too much experience using rds clusters, but I could imagine some helpful errors to be:

`A VPC must be provided, either through vpc or instanceProps.vpc`
`Only one of vpc and instanceProps.vpc can be provided`

If you think consolidating is best, let's do that.

However, you will need to edit the unit tests in `aws-rds/test/cluster.test.ts` to use your new error messages. That is why your build is failing.

prasanthlouis:
> Hey @prasanthlouis, I think you should write the error messages in a way that is the most clear to you. If the current message did not make sense, of course feel free to improve it!
> 
> I do not have too much experience using rds clusters, but I could imagine some helpful errors to be:
> 
> `A VPC must be provided, either through vpc or instanceProps.vpc` `Only one of vpc and instanceProps.vpc can be provided`
> 
> If you think consolidating is best, let's do that.
> 
> However, you will need to edit the unit tests in `aws-rds/test/cluster.test.ts` to use your new error messages. That is why your build is failing.

Thanks for the reply @scanlonp! I like the consolidation message because there's another similar message regarding VPC subnets right underneath this line as well! 

Thanks for informing me of the failing test. I've updated that as well! Let me know if there's anything else I can do to assist. Cheers!

TheRealAmazonKendra (approved):

PR #29712, opened 2024-04-03
fix(elasticloadbalancerV2): consistent logicalId for addTargetGroups (under feature flag)

### LINKED ISSUE #29496: ElasticLoadBalancerV2: migration to addAction() from addTargetGroups() fails because of logicalId change

ahammond:
Convention is that `ListenerRule`s should have a logicalId with a `Rule` suffix. Follow convention while providing a feature flag to support people who already have resources deployed and need to stick with the legacy naming convention.

### Issue

Closes #29496
Related to PR #29513

### Reason for this change

Going forward, the `logicalId` produced should be consistent.

### Description of changes

Made logicalId consistent by default while providing a feature flag to enable backwards compat for apps which have pre-deployed infra.

UTs applied.
`Exemption Request`: ITs are not required per docs

### Checklist

----


ahammond:
`Exemption Request`: ITs are not required for this change, per docs.

ahammond:
I'm not sure if this is considered a Breaking Change or not.

TheRealAmazonKendra:
I think that we need to revert the previous change and combine these two using optional props, not feature flags. If you are defaulting to false, it's generally an indication that the change isn't quite right.

TheRealAmazonKendra:
Alright, we reverted the other one so here are my thoughts on the path forward here. I totally get why you originally went the route that you did but we try to minimize the amount of feature flags we add and adding two here didn't feel right.

Upon looking through your code and this package, I noticed additional inconsistencies. In the `ApplicationListener` class we add `Rule` as a suffix in both `addAction` on line [294](https://github.com/aws/aws-cdk/blame/main/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts#L294) and in `addTargetGroups` on line  [322](https://github.com/aws/aws-cdk/blame/main/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts#L322). On the other hand, in the `ExternalApplicationListener` class we do not add `Rule` as a suffix on `addTargetGroups` on line [629](https://github.com/aws/aws-cdk/blame/main/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts#L629) but we do add it in `addAction` on line [675](https://github.com/aws/aws-cdk/blame/main/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts#L675).

What I'm thinking is this: what if we add an optional prop to these functions that is something to the extent of `overrideDefaultId`, or maybe something to indicate that it's a migration from on of the other functions, and use that to control whether or not `Rule` is added as a suffix. That way it's not a breaking change, and if nothing is added for that prop then we go with the current default. 

That prop name probably isn't very good, though, so I'm totally open to better names.

What do you think?

PR #29718, opened 2024-04-04
fix(cli): diff with changeset fails if deploy role cannot be assumed

### LINKED ISSUE #29650: cdk diff: in 2.134.0 diff fails if `deploy` cannot be assumed
### Describe the bug

In our CI we have a readonly role with access to the `lookup` role, but not the `deploy` role for non protected branches.

In version 2.133.0 running diff works - in 2.134.0 we get a crash.

For now we fixed it by pinning the previous version.

### Expected Behavior

Diffing with a role which can assume the `lookup` role works

### Current Behavior

Diffing with a role which cannot assume `deploy` fails.

```
current credentials could not be used to assume '<deploy role>' but are for the right account. Proceeding anyway.
User: <ci_role> is not authorized to perform: cloudformation:DescribeStacks on <stack> because no identity-based policy allows the cloudformation:DescribeStacks action
```

Not clear what's happening here - I guess after failing to assume the `deploy` role the current role is the ci role which attempted to assume `deploy`, and then after `DescribeStacks` is called.

### Reproduction Steps

I don't think I can easily provide this, since you would need to set up the same roles, stacks, etc. This does work with a higher privilege role, which indicates it's to do with the role assumption.

- Bootstrapped CDK
- CDK setup (we use Go, but it seems very clearly to be a permission issue).
- Run `cdk diff --all --ci`

### Possible Solution

Ensure that the `lookup` role is still used after trying to assume the `deploy` role and failing.

_No response_

### CDK CLI Version

v2.134.0

### Framework Version

_No response_

### Node.js Version

20

### OS

Darwin

### Language

Go

### Language Version

1.22

### Other information

_No response_

scanlonp:
Closes #29650

### Description of changes

This addresses the issue in two ways:

1. If the describeStacks call errors out, we now catch it and default to classic diff behavior.
2. The describeStacks call now tries to use the lookup role rather than the deploy role.

Manual testing with a user that could only assume lookup roles.

### Checklist

----


scanlonp:
@Mergifyio update

TheRealAmazonKendra (dismissed):
We've already messed this up a few times so I'd like to see significantly more tests here.

scanlonp:
@TheRealAmazonKendra, these tests should cover the cases for the stack exists call. I did not find a good way to test which permissions we grab for the stack exists call, but I did not find too much prior art around which role is assumed. Additionally, the new functionality in `stackExists()` is exclusive to the call made in our diff logic.

scanlonp:
@Mergifyio update

packages/aws-cdk/lib/api/deployments.ts
```diff
@@ -430,7 +431,12 @@ export class Deployments {
   }
 
   public async stackExists(options: StackExistsOptions): Promise<boolean> {
-    const { stackSdk } = await this.prepareSdkFor(options.stack, undefined, Mode.ForReading);
+    let stackSdk;
+    if (options.tryLookupRole) {
+      stackSdk = (await this.prepareSdkWithLookupOrDeployRole(options.stack)).stackSdk;
+    } else {
+      stackSdk = (await this.prepareSdkFor(options.stack, undefined, Mode.ForReading)).stackSdk;
+    }
```
comcalvi, line 439: `prepareSdkWithLookupOrDeployRole` defaults to `ForReading`. If we can't assume the lookup role, it will fall back to the deploy role. 

Don't use `prepareSdkFor` directly, use the wrappers like `prepareSdkWithLookupRole` or whichever one you want to use; those methods have names that make it clear which role is being assumed.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -183,11 +192,22 @@ export class CdkToolkit {
...
-          // only perform this check if we're going to make a changeset. This check requires permissions that --no-changeset users might not have.
-          const stackExists = await this.props.deployments.stackExists({
-            stack: stack,
-            deployName: stack.stackName,
-          });
+
+          let stackExists = false;
+          try {
+            // only perform this check if we're going to make a changeset. This check requires permissions that --no-changeset users might not have.
+            // the call should now use the lookup role, but keep behind the flag since we only need it if the changeset is being made
```
comcalvi, line 199: 1. We know this is kept behind the flag from reading it, you don't need this comment here.
2. "the call should now use the lookup role" - sure, but it should have only ever used that in the first place. This isn't particularly valuable here.

packages/aws-cdk/lib/api/deployments.ts
```diff
@@ -430,7 +431,12 @@ export class Deployments {
   }
 
   public async stackExists(options: StackExistsOptions): Promise<boolean> {
-    const { stackSdk } = await this.prepareSdkFor(options.stack, undefined, Mode.ForReading);
+    let stackSdk;
+    if (options.tryLookupRole) {
```
comcalvi, line 435: This option is confusing...if it's on, then try to assume the lookup role or the deploy role, but if it's off, we assume the lookup role only?

comcalvi (changes requested):

packages/aws-cdk/lib/api/deployments.ts
comcalvi: We should still try to fallback to the deploy role, in case somebody doesn't have permissions to the lookup role. Why don't we catch any errors we get when trying to assume the deploy role? If we can't assume it then we stop changeset creation. 

Inside of `prepareSdkFor`, `cachedSdkForEnvironment` returns an object with `didAssumeRole` on it. `didAssumeRole` is not returned to the caller, but you could add it as a property to the returned object. Then you have a way of checking if the fallback was successful.

comcalvi (approved):

comcalvi:
LGTM; let's test out removing the option and only using `lookupOrDeployRole` instead

packages/aws-cdk/lib/cdk-toolkit.ts
scanlonp, line 199:
```suggestion

```

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -183,11 +192,21 @@ export class CdkToolkit {
...
         if (options.changeSet) {
-          // only perform this check if we're going to make a changeset. This check requires permissions that --no-changeset users might not have.
-          const stackExists = await this.props.deployments.stackExists({
-            stack: stack,
-            deployName: stack.stackName,
-          });
+
+          let stackExists = false;
+          try {
+            // only perform this check if we're going to make a changeset. This check requires permissions that --no-changeset users might not have.
```
scanlonp, line 198:
```suggestion

```

comcalvi (approved):

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -139,10 +139,19 @@ export class CdkToolkit {
       let changeSet = undefined;
 
       if (options.changeSet) {
-        const stackExists = await this.props.deployments.stackExists({
-          stack: stacks.firstStack,
-          deployName: stacks.firstStack.stackName,
-        });
+        let stackExists = false;
```
comcalvi, line 142:
```suggestion
        let stackExists = false;
        // foo
```

packages/aws-cdk/lib/api/deployments.ts
comcalvi: discussed off(on?)line, we can investigate this in a follow up next week

PR #29723, opened 2024-04-04
feat(eks): pass helm chart values to aws-load-balancer-controller

### LINKED ISSUE #29707: [aws-eks]: add support for passing helm chart values to AlbController
### Describe the feature

Allow users to pass optional helm chart values to the aws-load-balancer-controller helm chart to configure its behavior.

### Use Case

We are using AWS Firewall Manager as well as the wafv2.CfnWebACLAssociation construct to associate our load balancers with WAFs so that we don't have to hard-code the waf arn in our manifests. aws-load-balancer-controller is constantly removing these associations. We can disable this behavior by setting `enableWafv2: false` in the helm chart values. However, this is not currently possible without disabling the built-in albController and using the HelmChart construct directly.

### Proposed Solution

expose an additional property `values` on the [AlbControllerProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_eks.AlbControllerProps.html) interface. Then apply the values to the defaults already being used when AlbController instantiates the HelmChart construct under the hood.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.133.0

MacOS Ventura 13.6.4

mtrspringer:
### Issue # (if applicable)

Closes #29707 

### Reason for this change

Need to be able to disable aws-load-balancer-controller wafv2 behavior so it doesn't remove waf associations created by AWS FMS.

### Description of changes

I added a `values` property to the `AlbControllerOptions` interface for passing optional values to the underlying helm chart.

I added two tests to validate my changes:
- a test to check that values passed appear in the template output
- a test to check that values passed do not override values currently set by the AlbController construct

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-eks/README.md
+++ b/packages/aws-cdk-lib/aws-eks/README.md
@@ -692,7 +692,7 @@ if (cluster.albController) {
 ```
 
 If you need to configure the underlying ALB Controller, you can pass optional values that will be forwarded to the underling HelmChart construct.
-For example, if your organization uses AWS Firewall Manager you will need to disable aws-load-balancer-controller's waf modification behavior or it will disassociate WAFs that are not specified in the Ingress' annotations.
+For example, if your organization uses AWS Firewall Manager you will need to disable aws-load-balancer-controller's waf modification behavior or it will periodically disassociate WAFs that are not specified in the Ingress' annotations.
 
 Note: supported value options may vary depending on the version of the aws-load-balancer-controller helm chart you are using. You should consult the [ArtifactHub documentation](https://artifacthub.io/packages/helm/aws/aws-load-balancer-controller) for your version to ensure you are configuring the chart correctly.
 
@@ -701,7 +701,7 @@ new eks.Cluster(this, 'HelloEKS', {
   version: eks.KubernetesVersion.V1_29,
   albController: {
     version: eks.AlbControllerVersion.V2_6_2,
-    values: {
+    helmChartValues: {
       enableWaf: false,
       enableWafv2: false,
     },
--- a/packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
+++ b/packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
@@ -275,7 +275,7 @@ export interface AlbControllerOptions {
   /**
    * Override values to be used by the chart.
    * For nested values use a nested dictionary. For example:
-   * values: {
+   * helmChartValues: {
    *  autoscaling: false,
    *  ingressClassParams: { create: true }
    * }
@@ -291,7 +291,7 @@ export interface AlbControllerOptions {
    *
    * @default - No values are provided to the chart.
    */
-  readonly values?: {[key: string]: any};
+  readonly helmChartValues?: {[key: string]: any};
 }
 
 /**
@@ -352,6 +352,8 @@ export class AlbController extends Construct {
     }
 
     // https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/deploy/installation/#add-controller-to-cluster
+    const { helmChartValues = {} } = props;
+    this.validateHelmChartValues(helmChartValues);
     const chart = new HelmChart(this, 'Resource', {
       cluster: props.cluster,
       chart: 'aws-load-balancer-controller',
@@ -363,7 +365,8 @@ export class AlbController extends Construct {
       wait: true,
       timeout: Duration.minutes(15),
       values: {
-        ...(props.values ?? {}),
+        ...helmChartValues,
+        // if you modify these values, you must also update this.restrictedHelmChartValueKeys
         clusterName: props.cluster.clusterName,
         serviceAccount: {
           create: false,
@@ -401,5 +404,23 @@ export class AlbController extends Construct {
       return rewriteResource(resources);
     }
     return resources.map(rewriteResource);
+  }
+
+  private validateHelmChartValues(values: {[key: string]: any}) {
+    const valuesKeySet = new Set(Object.keys(values));
+    const invalidKeys = new Set([...valuesKeySet].filter((key) => this.restrictedHelmChartValueKeys.has(key)));
+    if (invalidKeys.size > 0) {
+      throw new Error(`The following aws-load-balancer-controller HelmChart value keys are restricted and cannot be overridden: ${Array.from(this.restrictedHelmChartValueKeys).join(', ')}. (Invalid keys: ${Array.from(invalidKeys).join(', ')})`);
+    }
+  }
+
+  private get restrictedHelmChartValueKeys(): Set<string> {
+    return new Set([
+      'clusterName',
+      'serviceAccount',
+      'region',
+      'vpcId',
+      'image',
+    ]);
   }
 }
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-eks/test/integ.alb-controller.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-eks/test/integ.alb-controller.ts
@@ -12,7 +12,7 @@ import { IAM_OIDC_REJECT_UNAUTHORIZED_CONNECTIONS } from 'aws-cdk-lib/cx-api';
 const LATEST_VERSION: eks.AlbControllerVersion = eks.AlbControllerVersion.V2_8_2;
 
 interface EksClusterAlbControllerStackProps extends StackProps {
-  albControllerValues?: {[key:string]: any};
+  albControllerHelmChartValues?: {[key:string]: any};
 }
 
 class EksClusterAlbControllerStack extends Stack {
@@ -28,7 +28,7 @@ class EksClusterAlbControllerStack extends Stack {
       ...getClusterVersionConfig(this, eks.KubernetesVersion.V1_30),
       albController: {
         version: LATEST_VERSION,
-        values: props.albControllerValues,
+        helmChartValues: props.albControllerHelmChartValues,
       },
     });
 
@@ -82,7 +82,7 @@ const app = new App({
   },
 });
 const stack = new EksClusterAlbControllerStack(app, 'aws-cdk-eks-cluster-alb-controller');
-const stackWithAlbControllerValues = new EksClusterAlbControllerStack(app, 'aws-cdk-eks-cluster-alb-controller-values', { albControllerValues: { enableWafv2: false } });
+const stackWithAlbControllerValues = new EksClusterAlbControllerStack(app, 'aws-cdk-eks-cluster-alb-controller-values', { albControllerHelmChartValues: { enableWafv2: false } });
 new integ.IntegTest(app, 'aws-cdk-cluster-alb-controller-integ', {
   testCases: [stack, stackWithAlbControllerValues],
   // Test includes assets that are updated weekly. If not disabled, the upgrade PR will fail.
--- a/packages/aws-cdk-lib/aws-eks/test/alb-controller.test.ts
+++ b/packages/aws-cdk-lib/aws-eks/test/alb-controller.test.ts
@@ -164,7 +164,7 @@ test('can pass values to the aws-load-balancer-controller helm chart', () => {
     cluster,
     version: AlbControllerVersion.V2_6_2,
     repository: 'custom',
-    values: {
+    helmChartValues: {
       enableWafv2: false,
     },
   });
@@ -190,18 +190,18 @@ test('can pass values to the aws-load-balancer-controller helm chart', () => {
   });
 });
 
-test('values passed to the aws-load-balancer-controller do not override values set by construct', () => {
+test('values passed to the aws-load-balancer-controller result in an error if they conflict with restricted values set by the construct', () => {
   const { stack } = testFixture();
 
   const cluster = new Cluster(stack, 'Cluster', {
     version: KubernetesVersion.V1_27,
   });
 
-  AlbController.create(stack, {
+  expect(() => AlbController.create(stack, {
     cluster,
     version: AlbControllerVersion.V2_6_2,
     repository: 'custom',
-    values: {
+    helmChartValues: {
       clusterName: 'test-cluster',
       serviceAccount: {
         create: true,
@@ -214,27 +214,7 @@ test('values passed to the aws-load-balancer-controller do not override values s
         tag: 'test-tag',
       },
     },
-  });
-
-  Template.fromStack(stack).hasResourceProperties(HelmChart.RESOURCE_TYPE, {
-    Version: '1.6.2', // The helm chart version associated with AlbControllerVersion.V2_6_2
-    Values: {
-      'Fn::Join': [
-        '',
-        [
-          '{"clusterName":"',
-          {
-            Ref: 'Cluster9EE0221C',
-          },
-          '","serviceAccount":{"create":false,"name":"aws-load-balancer-controller"},"region":"us-east-1","vpcId":"',
-          {
-            Ref: 'ClusterDefaultVpcFA9F2722',
-          },
-          '","image":{"repository":"custom","tag":"v2.6.2"}}',
-        ],
-      ],
-    },
-  });
+  })).toThrow(/The following aws-load-balancer-controller HelmChart value keys are restricted and cannot be overridden: .*/);
 });
 
 describe('AlbController AwsAuth creation', () => {


packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
```diff
@@ -241,6 +241,27 @@ export interface AlbControllerOptions {
...
+   * }
+   *
+   * Note that the following values are set by the controller and cannot be overridden:
+   * - clusterName
+   * - serviceAccount.create
+   * - serviceAccount.name
+   * - region
+   * - vpcId
+   * - image.repository
+   * - image.tag
```
lpizzinidev, line 260:
```suggestion
   * The values to be used by the chart.
   * For nested values use a nested dictionary. For example:
   * values: {
   *  autoscaling: false,
   *  ingressClassParams: { create: true }
   * }
   *
   * Note that the following values are set by the controller and cannot be overridden:
   *  - clusterName
   *  - serviceAccount.create
   *  - serviceAccount.name
   *  - region
   *  - vpcId
   *  - image.repository
   *  - image.tag
```
Technically, we're not overriding the values.

packages/aws-cdk-lib/aws-eks/README.md
```diff
@@ -684,6 +684,25 @@ if (cluster.albController) {
 }
 ```
 
+If you need to configure the underlying ALB Controller, you can pass optional values that will be forwarded to the underling HelmChart construct.
+For example, if your organization uses AWS Firewall Manager you will need to disable aws-load-balancer-controller's waf modification behavior or it will disassociate WAFs that are not specified in the Ingress' annotations.
+
+Note: supported value options may vary depending on the version of the aws-load-balancer-controller helm chart you are using. You should consult the [ArtifactHub documentation](https://artifacthub.io/packages/helm/aws/aws-load-balancer-controller) for your version to ensure you are configuring the chart correctly.
```
lpizzinidev, line 690:
```suggestion
For example, if your organization uses AWS Firewall Manager you will need to disable `aws-load-balancer-controller`'s waf modification behavior or it will disassociate WAFs that are not specified in the Ingress' annotations.

**Note:** Supported value options may vary depending on the version of the `aws-load-balancer-controller` Helm chart you use. 
Visit the [ArtifactHub documentation](https://artifacthub.io/packages/helm/aws/aws-load-balancer-controller) to ensure you are configuring the chart correctly for your version.
```

lpizzinidev (approved):
Looks good, thanks! 👍
Left some minor suggestions for updating the documentation.

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
```diff
@@ -241,6 +241,27 @@ export interface AlbControllerOptions {
...
+   *  - serviceAccount.create
+   *  - serviceAccount.name
+   *  - region
+   *  - vpcId
+   *  - image.repository
+   *  - image.tag
+   *
+   * @default - No values are provided to the chart.
+   */
+  readonly values?: {[key: string]: any};
```
paulhcsun, line 264:
```suggestion
  readonly helmChartValues?: {[key: string]: any};
```

Can you rename this to `helmChartValues`? Just having this be `values` in this context is not clear what these values are being used for. Ideally I'd have like the property within Helm Chart to be named something other than `values` as well but it's too late for that now.

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
```diff
@@ -241,6 +241,27 @@ export interface AlbControllerOptions {
...
+   * }
+   *
+   * Note that the following values are set by the controller and cannot be overridden:
+   *  - clusterName
+   *  - serviceAccount.create
+   *  - serviceAccount.name
+   *  - region
+   *  - vpcId
+   *  - image.repository
+   *  - image.tag
```
paulhcsun, line 260: Do we know the full set of values that can be passed in to the `helmChart.values` property? If so could we add some stronger typing checks to them? If not then could we add some validation for the values that cannot be overridden? 

From your unit test case it seems like Helm Chart will just ignore these values but I would rather prevent these values from being set at all instead of allowing it to be passed in and have it silently fail.

paulhcsun (changes requested):

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
mtrspringer: i think the set of values varies (or can potentially vary) depending on the helm chart version. I think defining a type (e.g. AlbControllerHelmChartValues) will create a loose link to the helm chart that has the potential for breaking. Validating the passed values against the keys we know we explicitly set seems like a better approach to me.

paulhcsun: Gotcha, thanks for clarifying. Then let's go with the second option to validate the passed values that cannot be overridden.

mtrspringer: @paulhcsun i've updated with the requested changes. Assuming the suite passes, would you (or someone else) be able to run/update the integration tests? I ran into some issues with the teardown last time.

paulhcsun: Hey @mtrspringer, thanks for making the changes! and ya for sure, I'll give that a run later today.

paulhcsun: Hey @mtrspringer, apologies some other items came up. I will run the tests now.

paulhcsun: It took a while to run on Friday but looks like it failed due to a credentials related error. Possibly because it hung and timed out? I will try running it again but if it still fails then I'll ask another team member to help look into this as I'll be oncall this week.

mtrspringer: @paulhcsun thanks very much! yes it takes ages, looks like the build is passing though

paulhcsun:
Hi @mtrspringer, apologies for another delay to getting around to reviewing this. I was oncall last week. Thank you for making the requested changes for validating the values that can be passed to `helmChart.values`. 

Would you mind explaining what the changes within `@aws-cdk/sdk-v2-to-v3-adapter` are for? I don't remember seeing them during the initial review and I'm not quite sure how they're related to this change.

mtrspringer:
@paulhcsun hi sorry i've been on vacation the last 2 weeks. I dont remember making any changes to those files, maybe they got included when i updated my branch via the github button? im happy to leave them out

mtrspringer:
@paulhcsun looks like they were .d.ts and .js files that somehow got included. i've just removed them so this PR should be ready for review now

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
```diff
@@ -241,6 +241,27 @@ export interface AlbControllerOptions {
...
+   * - serviceAccount.create
+   * - serviceAccount.name
+   * - region
+   * - vpcId
+   * - image.repository
+   * - image.tag
+   *
+   * @default - No values are provided to the chart.
+   */
+  readonly helmChartValues?: {[key: string]: any};
```
TheRealAmazonKendra, line 264: I don't think that this is quite the right direction here. From what I can tell, only one or two values need to be overridden for your use case and this is basically giving the user carte blanche to override far more than that. I chatted with @paulhcsun about this PR and I'm thinking that it would likely be better to have a function that can be called after instantiation to disable the load balancer wafv2 behavior instead of adding a prop.

In general, we don't want props to contradict one another or allow values that we know are not allowed (i.e. the list of values you added in the docstring).

What I'm really having trouble with here is understanding how those values are being set in the first place, in your use case. Can you explain what's going on in your specific case a bit more so that our suggestions for the fix don't send you in circles?

TheRealAmazonKendra: I would also probably suggest that the function to update this behavior should be scoped to the ALB controller but, again, I'd like to understand better how these fields are being set before saying you should make this code change.

TheRealAmazonKendra (dismissed):

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
mtrspringer: hi @TheRealAmazonKendra thanks for your review.

you are correct that for my specific use case I only need to update one or two values, but my intention with this PR is to address the issue of aws-cdk-lib/aws-eks package not exposing the ability to set whatever values the user wants (apart from the ones it is already setting).

the AlbController is just a thin wrapper around the HelmChart construct, yet its constructor's props do not expose any ability to pass values to the underlying helm chart. IMO this was a poor design choice, as I don't think aws-cdk should be the arbiter of which values can be passed to the aws-load-balancer-controller helm chart.

the initial implementation of the Cluster construct did not include any props related to the albController. They were later added as a convenience due to the fact that it is commonly utilized to integrate with aws elbv2. my assumption is that the AlbController construct was defined at the same time, due to the fact that `AlbControllerOptions = Omit<AlbControllerProps, 'cluster'>`. However, the lack of an interface exposing configuration of the helm chart's values means that any user leveraging aws firewall manager either cannot use the "official" construct for the helm chart, or they have to accept that aws-load-balancer-controller will periodically disassociate an fms-applied web acl and their albs willbe unprotected until fms can detect & remediate (usually this is a couple minutes, but we have had occurrences of the remediation failing and albs being unprotected for hours).

Due to the fact that a helm chart's values.yml is the way in which every helm chart's default behavior is overridden, i feel that adding a prop for passing values to the AlbController is a valid design choice. There may be other default behaviors of the helm chart beyond waf management that users may want to disable. Due to the fact that the supported values for aws-load-balancer-controller may vary from one version to the next, I didn't feel it was wise to define a type/interface for a subset of configurable values that will then need to be kept up-to-date. I think if users decide to pass values to the AlbController, they are at the point where they know they need to and we should trust that they are able to decide which values to set.

as far as your suggestion for a function to update values after instantiation, this would need to be exposed on both the AlbController construct and the HelmChart construct. i feel like this is not a great approach and it expands the scope of the solution beyond the AlbController construct.

in short, i feel that this is a fix for an overly-restrictive construct props definition.

Necrokefalos: Hello, I would also want to be able to edit values through this way, in my case for values `enableCertManager`, `cluster.dnsDomain` and some more.
I wonder why this functionality wasn't introduced from the beginning...

Necrokefalos:
Hi, any updates or estimation on this? Can i help somehow?

mtrspringer:
hi @Necrokefalos im still waiting on a response as to what changes, if any, need to be made. good to know that the values also need to be modified for cert-manager as we are looking to adopt that sometime soon. in the meantime i think the integration tests need to be run again before the checks pass. @paulhcsun would you be able to do that again at your earliest convenience? thanks.

mtrspringer:
@Necrokefalos not sure if this will work for your use-case (and it's hacky) but in the interim while this is being deliberated I realized we can "disable" alb controller behavior via the serviceAccount policy, which we can access via the construct node.findChild api:

```ts
    // TODO: remove once cdk supports passing values to albController Helm Chart:
    const {enableWafv2 = false} = albControllerValues;
    if (!enableWafv2) {
      const albControllerServiceAccount =
        this.eksCluster.albController!.node.findChild(
          'alb-sa'
        ) as eks.ServiceAccount;
      albControllerServiceAccount.addToPrincipalPolicy(
        new iam.PolicyStatement({
          effect: iam.Effect.DENY,
          actions: [
            'waf-regional:GetWebACL',
            'waf-regional:GetWebACLForResource',
            'waf-regional:AssociateWebACL',
            'waf-regional:DisassociateWebACL',
            'wafv2:GetWebACL',
            'wafv2:GetWebACLForResource',
            'wafv2:AssociateWebACL',
            'wafv2:DisassociateWebACL',
          ],
          resources: ['*'],
        })
      );
    }
```

i still have to test how this affects the controller's operation (i.e. can it gracefully handle the permission errors) but it might be easier than replacing the cdk.eks.Cluster.albController with a custom HelmChart implementation.

paulhcsun:
Hi @mtrspringer, apologies for the late response I was away recently and just got back to this. I've started running the integ tests so hopefully i'll get those pushed today or tmr. I think when I last discussed with Kendra there was one design change for how we allow the users to update the helm chart values. I will confirm with her and get back asap. Apologies for the churn.

paulhcsun:
I've pushed the snapshot changes for the `integ.alb-controller.js` test however the `test/aws-elasticloadbalancingv2/test/integ.nlb-attributes.js` test currently fails with this error:

```
Tests:    1 failed, 921 total
Failed: /Volumes/workplace/forks-cdk/mtrspringer/packages/@aws-cdk-testing/framework-integ/test/aws-elasticloadbalancingv2/test/integ.nlb-attributes.js

Running integration tests for failed tests...

Running test /Volumes/workplace/forks-cdk/mtrspringer/packages/@aws-cdk-testing/framework-integ/test/aws-elasticloadbalancingv2/test/integ.nlb-attributes.js in us-east-1
/Volumes/workplace/forks-cdk/mtrspringer/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.js:93
                throw new Error(`'denyAllIgwTraffic' may only be set on load balancers with ${enums_1().IpAddressType.DUAL_STACK} addressing.`);
                ^

Error: 'denyAllIgwTraffic' may only be set on load balancers with dualstack addressing.
    at new BaseLoadBalancer (/Volumes/workplace/forks-cdk/mtrspringer/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.js:93:23)
    at new NetworkLoadBalancer (/Volumes/workplace/forks-cdk/mtrspringer/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/nlb/network-load-balancer.js:165:9)
    at Object.<anonymous> (/Volumes/workplace/forks-cdk/mtrspringer/packages/@aws-cdk-testing/framework-integ/test/aws-elasticloadbalancingv2/test/integ.nlb-attributes.js:13:1)
    at Module._compile (node:internal/modules/cjs/loader:1368:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1426:10)
    at Module.load (node:internal/modules/cjs/loader:1205:32)
    at Module._load (node:internal/modules/cjs/loader:1021:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:142:12)
    at node:internal/main/run_main_module:28:49

Node.js v21.7.1
  ERROR      /Volumes/workplace/forks-cdk/mtrspringer/packages/@aws-cdk-testing/framework-integ/test/aws-elasticloadbalancingv2/test/integ.nlb-attributes.js (undefined/us-east-1) 0.17s
      Error during integration test: Error: Command exited with status 1

Test Results:

Tests:    1 failed, 1 total
Error: Some integration tests failed!
    at main (/Volumes/workplace/forks-cdk/mtrspringer/packages/@aws-cdk/integ-runner/lib/cli.js:166:23)
error Command failed with exit code 1.
```

I'm not quite sure how this test is related or why it's failing now. Don't se anything related to `denyAllIgwTraffic` within that test or in any of your changes...

mtrspringer:
@paulhcsun thanks yeah thats bizarre, i didnt make any changes to the nlb tests or source code. maybe something strange happened with the test assets during rebasing?

should i revert all asset/generated changes so that its just my desired code and let you run the suite again?

shikha372 (dismissed):
Thank you @mtrspringer for submitting the PR, looking at the[ initial suggestion here](https://github.com/aws/aws-cdk/pull/29723/files#r1644794503) I would also prefer a function based implementation that can give users the option to override only set of allowed values possible for alb-controller.

mtrspringer:
@paulhcsun hi sorry i've been busy with my end-of-year work. i've just rebased the PR is there any chance you could run the integration tests again please? the iam-based disabling hack i commented about [here](https://github.com/aws/aws-cdk/pull/29723#issuecomment-2233445348) has not worked as the aws-load-balancer-controller helm chart fails to complete the ingress update if the waf action it wants to perform fails. in this light i'd really like to try and get this feature over the line.

@shikha372 could you please elaborate on what you are looking for that's different from the current implementation? i have already updated to throw an error if a user attempts to set a helm chart value other that the AwsLoadBalancerController construct wants to set in response to previous comments. is this sufficient?

thanks

paulhcsun:
> @shikha372 could you please elaborate on what you are looking for that's different from the current implementation? i have already updated to throw an error if a user attempts to set a helm chart value other that the AwsLoadBalancerController construct wants to set in response to previous comments. is this sufficient?
> 
> thanks

What @shikha372 is referring to is the solution that @TheRealAmazonKendra recommended to use a function that can be called after instantiation to override the desired values. I believe this is a good halfway compromise because we can still add in more methods for the properties that we want to support overrides for. I understand that for your use case and probably @Necrokefalos too, it would be fine to let you have control over all of the properties you want to override but this change would be applied to all other users too and we want to have some sort of safety net. In general in the CDK we try to make it impossible to set conflicting property values or properties that cannot be changed rather than to allow it to be set but prevent the deployment with error handling.

packages/aws-cdk-lib/aws-eks/lib/alb-controller.ts
```diff
@@ -380,4 +405,22 @@ export class AlbController extends Construct {
     }
     return resources.map(rewriteResource);
   }
+
+  private validateHelmChartValues(values: {[key: string]: any}) {
```
mtrspringer, line 409: @paulhcsun i have already implemented the validation as a function that restricts the keys the user is allowed to modify. do you mean that overriding the values should be an instance method? this does not make sense to me and does not align with how other properties of the AlbController HelmChart wrapper are set. for example, there is no albController.updateVersion() method, the helm chart version is passed in the constructor.

an instance method would also require updating the HelmChart construct to support modifying values after initialization (the AlbController constructor initializes the underlying HelmChart construct), which again makes the solution more complicated and expands beyond the scope of the AlbController - the HelmChart construct does not need to support value modification as it allows for passing arbitrary values already.

let me know if any modifications to this implementation are required and i'll be happy to make them.

mtrspringer:
@paulhcsun thanks for the comments. im unsure what could be causing those errors - none of my test cases specify the `denyAllIgwTraffic` field. maybe it was caused by an underlying change? i can try to dig in more

i've responded to your comment about filtering the allowed values above, happy to hear your thoughts there.

GavinZZ (changes requested):
I see that the integ test `integ.alb-controller` fails. This is likely due to you changes to this file. Please follow this [link](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md#L308-L309) to update the test snapshots.

FYI, you can probably just run `yarn integ aws-eks/test/integ.alb-controller --update-on-failed` from `packages/@aws-cdk-testing/framework-integ/test`.

shikha372:
Thank you @mtrspinger for contributing to CDK, this PR seems to be inactive for a while so we've opened PR https://github.com/aws/aws-cdk/pull/34077 to handle requested feature for `enableWaf` property in ALB controller props.

PR #29726, opened 2024-04-04
docs(elasticloadbalancingv2): set proper default for stickiness

### LINKED ISSUE #29725: aws-cdk-lib: Possibly incorrect default value spec in type-doc for stickinessCookieDuration for ApplicationTargetGroup
### Describe the bug

I believe [this](https://github.com/aws/aws-cdk/blob/b82320b08ebcda98b85be8ceb56a5a4b39511d4a/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-target-group.ts#L63) type-doc that claims the default value to be "1 days" is not accurate.

Following the code path, I think, the behavior rather is that, if `stickinessCookieDuration` wasn't set, then the stickiness is not enabled; and if it is defined it must be between 1 second and  7 days (604800 seconds) —this last bit is correctly specified in the type-doc.

### Expected Behavior

See issue description

### Current Behavior

See issue description

### Reproduction Steps

See issue description

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.125.0 (build 5e3c3c6)

### Framework Version

_No response_

### Node.js Version

node v18.19.1

### OS

macOS 14.4.1 (23E224)

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

msambol:
By default, stickiness is disabled. The duration _is_ one day if stickiness is enabled,
but we don't surface that option to users. If `stickinessCookieDuration` is not 
supplied, stickiness will not be enabled so the 1 day default doesn't make sense here.

Closes #29725.

----

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -13957,16 +13957,7 @@ string-length@^4.0.1:
     char-regex "^1.0.2"
     strip-ansi "^6.0.0"
 
-"string-width-cjs@npm:string-width@^4.2.0":
-  version "4.2.3"
-  resolved "https://registry.npmjs.org/string-width/-/string-width-4.2.3.tgz#269c7117d27b05ad2e536830a8ec895ef9c6d010"
-  integrity sha512-wKyQRQpjJ0sIp62ErSZdGsjMJWsap5oRNihHhu6G7JVO/9jIB6UyevL+tXuOqrng8j/cxKTWyWUwvSTriiZz/g==
-  dependencies:
-    emoji-regex "^8.0.0"
-    is-fullwidth-code-point "^3.0.0"
-    strip-ansi "^6.0.1"
-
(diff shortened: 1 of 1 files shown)


nmussy (approved):
LGTM, matches the behavior:
https://github.com/aws/aws-cdk/blob/e52510c9d51abc5703c1dc61653aa3c4d04f7b37/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-target-group.ts#L340-L344

TheRealAmazonKendra:
I die on the inside a little every time I see this https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-target-group.ts#L73
```
   * Note: `stickinessCookieName` parameter depends on the presence of `stickinessCookieDuration` parameter.
   * If `stickinessCookieDuration` is not set, `stickinessCookieName` will be omitted.
```

But, that's not your change. Yours is good. Thanks for fixing that. I'm going to rename this a docs change for better clarity on what was updated.

TheRealAmazonKendra (approved):

PR #29730, opened 2024-04-04
chore: remove sdkv2 from cloudformation-diff

TheRealAmazonKendra:
### Issue # (if applicable)

Closes #<issue number here>.

### Checklist

----

### DIFF
--- a/packages/aws-cdk/test/diff.test.ts
+++ b/packages/aws-cdk/test/diff.test.ts
@@ -405,7 +405,7 @@ describe('stack exists checks', () => {
         Changes: [
           {
             ResourceChange: {
-              Action: 'Dummy',
+              Action: 'Add',
               LogicalResourceId: 'Object',
             },
           },


packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
```diff
@@ -1104,7 +1104,7 @@ describe('changeset', () => {
             ResourceType: 'AWS::S3::Bucket',
             Replacement: 'True',
             Details: [{
-              Evaluation: 'Direct',
+              Evaluation: 'Static',
```
piradeepk, line 1107: Curious, do you mind explaining why we needed to change the Evaluation type from Direct to Static?

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
```diff
@@ -1,12 +1,14 @@
...
 // The SDK should not make network calls here
-import type { CloudFormation } from 'aws-sdk';
+import type { DescribeChangeSetOutput as DescribeChangeSet } from '@aws-sdk/client-cloudformation';
 import * as impl from './diff';
 import * as types from './diff/types';
 import { deepEqual, diffKeyedEntities, unionOf } from './diff/util';
 
 export * from './diff/types';
 
+export type DescribeChangeSetOutput = DescribeChangeSet;
```
comcalvi, line 10: why both this and the import? The change on line 3 lets us reference this type as `DescribeChangeSet`, but without the `as`, you could achieve the same result we get from both lines 3 and 10.

yarn.lock
```diff
@@ -1848,6 +2113,19 @@
     "@smithy/types" "^2.5.0"
     tslib "^2.5.0"
 
+"@aws-sdk/credential-provider-sso@3.540.0":
+  version "3.540.0"
```
comcalvi, line 2117: I guess these are fine? ideally the changes to `yarn.lock` would reflect only the diff in `package.json`, but node dependency tooling is funny

comcalvi (changes requested):

packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
TheRealAmazonKendra: This was not a valid value before but in v2 it wasn't strongly typed enough for it to matter in our test case.

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
TheRealAmazonKendra: This type is used extensively throughout aws-cdk (the cli, not the entire package). This re-export is a workaround In order to not cause a conflict of types in that package. You could argue that we should just migrate that whole package, but that's a much more complex task that we have not yet prioritized.

comcalvi (approved):

piradeepk (approved):
Lgtm! 🚀

TheRealAmazonKendra:
@Mergifyio update

PR #29746, opened 2024-04-05
feat(elasticloadbalancingv2): add removeSuffix param for ExternalApplicationListener.addAction()

### LINKED ISSUE #29496: ElasticLoadBalancerV2: migration to addAction() from addTargetGroups() fails because of logicalId change

ahammond:
### Issue # (if applicable)

Closes #29496

### Reason for this change

See https://github.com/aws/aws-cdk/pull/29513 (props based solution instead of feature-flag)

### Description of changes

Adds a `removeSuffix` property to addAction method to address problems due to logicalId inconsistency.

UTs. Per IT document, integration tests are not necessary.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/custom-resource-handlers/lib/aws-eks/kubectl-handler/helm/__init__.py
+++ b/packages/@aws-cdk/custom-resource-handlers/lib/aws-eks/kubectl-handler/helm/__init__.py
@@ -100,7 +100,7 @@ def helm_handler(event, context):
 
 def get_oci_cmd(repository, version):
     # Generates OCI command based on pattern. Public ECR vs Private ECR are treated differently.
-    private_ecr_pattern = 'oci://(?P<registry>\d+\.dkr\.ecr\.(?P<region>[a-z0-9\-]+)\.amazonaws\.com)*'
+    private_ecr_pattern = 'oci://(?P<registry>\d+\.dkr\.ecr\.(?P<region>[a-z0-9\-]+)\.amazonaws\.com(\.cn)?)*'
     public_ecr_pattern = 'oci://(?P<registry>public\.ecr\.aws)*'
 
     private_registry = re.match(private_ecr_pattern, repository).groupdict()


ahammond:
`Exemption Request`: this change doesn't touch any of the aspects which would require IT.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/README.md
```diff
@@ -750,3 +750,18 @@ const targetGroup = elbv2.ApplicationTargetGroup.fromTargetGroupAttributes(this,
...
 const targetGroupMetrics: elbv2.IApplicationTargetGroupMetrics = targetGroup.metrics; // throws an Error()
 ```
+
+## logicalIds on ExternalApplicationListener.addTargetGroups() and .addAction()
+
+Legacy behavior for the `addTargetGroups()` method did not follow the standard of adding
+a `Rule` suffix to the logicalId.
+If you have `ListenerRule`s deployed using the legacy behavior and are upgrading,
+you will need to enable the `removeRuleSuffixFromLogicalId: true` property
+so that CloudFormation will not attempt to replace your existing `ListenerRule`s.
```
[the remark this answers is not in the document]
ahammond: So... invert the default in this case? And then the next major release fix the default to be consistent?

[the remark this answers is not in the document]
ahammond: I have change the PR to preserve legacy behavior.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts
```diff
@@ -620,13 +620,25 @@ abstract class ExternalApplicationListener extends Resource implements IApplicat
...
+   * If you have already deployed a `ListenerRule` using this method
+   * and need to migrate to using the more feature rich `addAction()`
+   * method, then you will need to set the `removeRuleSuffixFromLogicalId: true`
+   * property there to avoid having CloudFormation attempt to replace your resource.
    */
   public addTargetGroups(id: string, props: AddApplicationTargetGroupsProps): void {
     checkAddRuleProps(props);
 
     if (props.priority !== undefined) {
+      const removeRuleSuffixFromLogicalId = props.removeRuleSuffixFromLogicalId ?? true;
```
TheRealAmazonKendra, line 638: I'm not sure I see a reason to have the option to add `Rule` at all to this function, unless you think that users might want to switch back from `addAction()` to `addTargetGroups()`, but from your descriptions it sounds like `addAction()` is just better. 

I realize that making them all consistent was totally my suggestion, but I now that I'm seeing it implemented, I think I was wrong. If a user wants to migrate back this way, they can just add `Rule` to the end of the `id`.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts
```diff
@@ -620,13 +620,25 @@ abstract class ExternalApplicationListener extends Resource implements IApplicat
    *
    * It's possible to add conditions to the TargetGroups added in this way.
    * At least one TargetGroup must be added without conditions.
+   *
+   * NOTE: if you are deploying new infrastructure using this method,
+   * it is recommended to set the `removeRuleSuffixFromLogicalId: false`
+   * property so that the logicalId for the generated `ListenerRule`
+   * is consistent with other methods for managing `ListenerRule`s.
```
TheRealAmazonKendra, line 627: Sort of, in this case we don't want users to have to think about the existence of this prop at all. Any description of it should focus just on the migration use case. We don't have any reason to assume that all users will want to migrate in the future so this feels like cognitive overload.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts
```diff
@@ -620,13 +620,25 @@ abstract class ExternalApplicationListener extends Resource implements IApplicat
...
+   *
+   * If you have already deployed a `ListenerRule` using this method
+   * and need to migrate to using the more feature rich `addAction()`
+   * method, then you will need to set the `removeRuleSuffixFromLogicalId: true`
+   * property there to avoid having CloudFormation attempt to replace your resource.
```
TheRealAmazonKendra, line 632: This comment feels like it only belongs on `addAction()`

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts
```diff
@@ -664,15 +676,21 @@ abstract class ExternalApplicationListener extends Resource implements IApplicat
...
+   * If you previously deployed a `ListenerRule` using the `addTargetGroups()` method
+   * and need to migrate to using the more feature rich `addAction()`
+   * method, then you will need to set the `removeRuleSuffixFromLogicalId: true`
+   * property here to avoid having CloudFormation attempt to replace your resource.
    */
   public addAction(id: string, props: AddApplicationActionProps): void {
     checkAddRuleProps(props);
 
     if (props.priority !== undefined) {
+      const ruleId = props.removeRuleSuffixFromLogicalId ? id : id + 'Rule';
```
TheRealAmazonKendra, line 689: The name here feels really long. Can we go with something more brief like `removeSuffix` and add in the description that this is only for the sake of migration from the older function?

TheRealAmazonKendra (changes requested):

vinayak-kukreja (commented):
I have a general question and please let me know if I am missing historical context here :)

* Is `addAction` covering all that is done by `addTargetGroups` and adds more to it? If yes, do we want two functions doing somewhat similar things to exist? I believe it adds more confusion
* If prior scenario is true, where `addAction` just supercedes `addTargetGroups`, then why can we just not deprecate `addTargetGroups` and remove `Rule` suffix from `addAction` and add all of this behind a feature flag?

vinayak-kukreja:
> I have a general question and please let me know if I am missing historical context here :)
> 
>     * Is `addAction` covering all that is done by `addTargetGroups` and adds more to it? If yes, do we want two functions doing somewhat similar things to exist? I believe it adds more confusion
> 
>     * If prior scenario is true, where `addAction` just supercedes `addTargetGroups`, then why can we just not deprecate `addTargetGroups` and remove `Rule` suffix from `addAction` and add all of this behind a feature flag?

Ok, so I see `addAction` provides more control but `addTargetGroups` provides an abstracted version of it's functionality with some defaults. 
In that case, IMO this configuration via props is becoming a bit confusing. I understand that we are trying to minimize using feature flags but adding it here probably will make a cleaner contract for the user. Thoughts?

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts
ahammond: This addresses the problem of a user wanting to use `addTargetGroup()` while having a consistent logicalId. I... don't think most people would ever discover this. The interesting question is how to move forward with consistent logiacalIds. I'm pretty sure that the Right Way to do this is to change the default and provide a FeatureFlag to revert to legacy behaviour.

TBH, I'm not particular so long as we unblock the ability for `addAction()` to deal with the inconsistency.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts
ahammond: I kind of agree. Maybe addressing changes to the `addTargetGroups()` method should be a separate PR?

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-listener.ts
ahammond: Long, but only used in a very specific edge case, which is usually the justification for a verbose name.

ahammond:
I removed `addTargetGroup()` related changes from this PR. I really need the `addAction()` stuff resolved so we can move forward.

ahammond:
> Ok, so I see `addAction` provides more control but `addTargetGroups` provides an abstracted version of it's functionality with some defaults. In that case, IMO this configuration via props is becoming a bit confusing. I understand that we are trying to minimize using feature flags but adding it here probably will make a cleaner contract for the user. Thoughts?

I've pulled out the stuff addressing the `addTargetGroups` from this PR to only address the problem which is blocking us. I think that the Right Way to address this going forward for the `addTargetGroups()` is a feature flag along the lines of what I wrote with https://github.com/aws/aws-cdk/pull/29513

But let's go one step at a time.

TheRealAmazonKendra:
@Mergifyio update

TheRealAmazonKendra:
Oh mergify...

TheRealAmazonKendra (approved):

PR #29729, opened 2024-04-04
feat(ec2): `NatInstanceProviderV2` improvements

nmussy:
### Issue # (if applicable)

Closes #29720

Might fix other issues, will need check the backlog

### Reason for this change

The user data script used to initialize NAT instances was at present incorrect, assuming an incorrect default interface name. I wanted to both update this script, and allow CDK users to update this user data if needed in the future, both in case of an issue with the script or just to extend/override the initialization procedure.

### Description of changes

Apologies for the non-atomic PR, as I was working on the unit and integration tests, I noticed additional issues and missing features. These are the initially targeted changes:

* Add `userData` field to `NatInstanceProps`, allowing the user to overwrite the NAT instances initialization script
  * Add static `NatInstanceProps.DEFAULT_USER_DATA_COMMANDS` field, allowing the user to more easily extend the default script
* Fix default `userData` to query the instance `default` network interface, instead of always using `eth0`

These were found along the way:

* Add getter `gatewayInstances` method to `NatInstanceProps`, allowing the user to interact with the created NAT instances once they're attached to the VPC
  * This allowed me to give the instance role additional permissions, but has more potential uses
* Fix `creditSpecification` not being implemented correctly
* Updated existing unit and integ tests to increase coverage
* Add warning in `README` about potential issues with NAT instances as is implemented

### Description of how you validated changes

I've updated and added unit and integration tests

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #29769, opened 2024-04-09
chore(ec2): `NatInstanceProviderV2` security group example

### LINKED ISSUE #27527: aws-cdk/aws-ec2: Better Support for NAT Gateway Construction in L2 Vpc
### Describe the feature

There is currently a cyclic dependency between Security Groups and VPCs within NatInstanceProvider.

The dependency chain looks something like this:

Vpc
  constructor props: VpcProps
    natGatewayProvider: [NatProvider](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.NatProvider.html)
      NatInstanceProvider -> NatProvider
        constructor props: NatInstanceProps
          securityGroup: ISecurityGroup
            SecurityGroup -> ISecurityGroup
              constructor props: [SecurityGroupProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.SecurityGroupProps.html)
                **vpc: IVpc**

As discussed during the office hours session, I believe you could replicate the design used for subnet selection to create named security groups.  You could add a property to VpcProps that allows the definition of Security groups to be created, which would include some identifying name property in addition to the other parameters.  And then you could create a NatInstanceProps parameter that accepts the security group by the identifying name instead of as an ISecurityGroup, or perhaps some other NatProvider class that would integrate such options to construct NAT Gateway instances.

### Use Case

The use case is creating NatGateways on VPC construction as appears to be possible given the existing API, but doesn't seem to be fully fleshed out.  Currently that feature seems to be broken around the security group assignment for the NAT gateways creating a cyclic dependency on the need for a security group to be assigned a VPC.

### Proposed Solution

As discussed during the office hours session, I believe you could replicate the design used for subnet selection to create named security groups.  You could add a property to VpcProps that allows the definition of Security groups to be created, which would include some identifying name property in addition to the other parameters.  And then you could create a NatInstanceProps parameter that accepts the security group by identifying name instead of as an ISecurityGroup, or perhaps some other NatProvider class that would integrate such options to construct NAT Gateway instances.

### Other Information

I think we're constrained by the limitations of the corresponding CloudFormation constructs here, so creating security groups without needing a VPC, such that they may be passed as arguments to the NAT Gateway instance providers and subsequently assigned the VPC doesn't seem to be an option.
Currently the only workarounds seem to include:
- using entirely L1 constructs
- dropping the NAT Gateway feature from VPC construction entirely
- ignoring the NAT Gateway Vpc-construction-time feature and manually adding security groups and NAT gateways after the fact (after the VPC is constructed)

### Acknowledgements

### CDK version used

2.99.1

Any, MacOS/AL2

nmussy:
### Issue # (if applicable)

Closes #27527

### Reason for this change

The current `NatInstanceProviderV2.securityGroup` property is unusable, given the dependency loop between the construct props (`NatInstanceProviderV2` > `VPC` > `SecurityGroup` > `NatInstanceProviderV2`).
When creating the integration for #29729, adding a getter for the instances generated by the provider to update the instance role was required to test the `userData` overload. This solution also allows to bypass the circular dependency describe above, given that both the VPC and the instances are generated once the VPC is created with the `natGatewayProvider`.

### Description of changes

* Deprecate `NatInstanceProviderV2.securityGroup`
  * Add `@example` tag to demo `NatInstanceProviderV2.gatewayInstances`
* Update `README` to demo setting the security group
* Update integ to test the demo

Updated integration test

### Checklist

----


nmussy:
@arachnilith Can you confirm that this covers the feature request you raised in #27527?

GavinZZ (dismissed):
LGTM, thanks for making the change.

GavinZZ (approved):

PR #29784, opened 2024-04-10
fix(aws-cdk-lib): allow create explicitly empty trees on aws-cdk-lib core

Rondineli:
### Issue # (if applicable)

Closes #[29165](https://github.com/aws/aws-cdk/issues/29165).

### Reason for this change

Using `addOverride` calls `deepMerge`, and few edge cases like wafv2 rules needs Actions and other keys with objects and empty objects. Fixing reviews on (29217)[https://github.com/aws/aws-cdk/pull/29217]

### Description of changes

Changing the code to be more realistic with the comment, where only null data is deleted

No tests, just making sure the current change would be compatible with current tests, should not break anything

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-s3-deployment/README.md
+++ b/packages/aws-cdk-lib/aws-s3-deployment/README.md
@@ -24,13 +24,13 @@ This is what happens under the hood:
 
 1. When this stack is deployed (either via `cdk deploy` or via CI/CD), the
    contents of the local `website-dist` directory will be archived and uploaded
-   to an intermediary assets bucket. If there is more than one source, they will
-   be individually uploaded.
-2. The `BucketDeployment` construct synthesizes a custom CloudFormation resource
+   to an intermediary assets bucket (the `StagingBucket` of the CDK bootstrap stack).
+   If there is more than one source, they will be individually uploaded.
+2. The `BucketDeployment` construct synthesizes a Lambda-backed custom CloudFormation resource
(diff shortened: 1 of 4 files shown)


Rondineli:
@paulhcsun I fixed the PR based on your requests and also, changed it to have a safeguard where we are not deleting **only** the waf objects list and everything else should be exactly the same. If you check the original issue (29165)[https://github.com/aws/aws-cdk/issues/29165] The suggested scenarios is currently not working and this should fix part of the problem. 

LEt me know if there is anything else I can do to make it more to what aws expects.

vinayak-kukreja:
Hey @Rondineli, thank you for opening a PR with us and apologies for the delay. 

I do not agree with the approach in this PR that we are taking to resolve the issue. The keys being mentioned to be excluded can be used in some other scenarios and could lead to potential issues. Even if it works for wafv2, we will want this to be more generic for other resources too.

IMO instead of adding exemption/exception keys for this, we need to fix the underlying issue and that is related to how `deepMerge` is working. 

I have added a [comment](https://github.com/aws/aws-cdk/issues/29165#issuecomment-2048316953) to the issue you opened. Please go through it and let us know if you feel comfortable working on the suggested approach, or if you would like us to work on this? :)

Rondineli:
@vinayak-kukreja Will adjust this PR to the requested approach :) 

Thanks for the explanation and review!

PR #29793, opened 2024-04-11
feat(ec2): well-known port aliases

nmussy:
### Issue # (if applicable)

None as far as I can tell

### Reason for this change

The web console lists commonly used ports when adding a rule to a security group, this aims to reproduce this simple quality of life shortcut. It can also help with code readability, and might save people from a typo

### Description of changes

* Add well-known static `Port` instances
  * Intersection of the AWS web console listed ports and the [IANA Service Name and Transport Protocol Port Number Registry](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt)

Compared the AWS web console values to the IANA list, and added a unit test to make sure the alias behaved properly 

### Checklist

----


packages/aws-cdk-lib/aws-ec2/lib/port.ts
```diff
@@ -188,6 +188,41 @@ export interface PortProps {
...
+  /** Well-known Microsoft SQL Server port (TCP 1433) */
+  public static readonly MSSQL = Port.tcp(1433);
+  /** Well-known NFS port (TCP 2049) */
+  public static readonly NFS = Port.tcp(2049);
+  /** Well-known MySQL and Aurora port (TCP 3306) */
+  public static readonly MYSQL_AURORA = Port.tcp(3306);
+  /** Well-known Microsoft Remote Desktop Protocol port (TCP 3389) */
+  public static readonly RDP = Port.tcp(3389);
+  /** Well-known PostgreSQL port (TCP 5432) */
+  public static readonly POSTGRES = Port.tcp(5432);
```
nmussy, line 224: Let me know if you would like to see more, but I think we should stick to the officially IANA assigned port, even if the AWS web console lists additional ones

nmussy:
Exemption Request: These are only aliases of existing methods, a new integration is not necessary

packages/aws-cdk-lib/aws-ec2/lib/port.ts
```diff
@@ -188,6 +188,41 @@ export interface PortProps {
  * Interface for classes that provide the connection-specification parts of a security group rule
  */
 export class Port {
+  /** Well-known SSH port (TCP 22) */
```
msambol, line 191: only a nit: do we need "Well-known" in all these?  Cleaner to just have `SSH port (TCP 22)` ?

msambol (approved):
Super helpful. Nice job 👍🏼

packages/aws-cdk-lib/aws-ec2/lib/port.ts
nmussy: I would have added these at the top of the class/enum JSDoc if they weren't declared alongside the rest of the `Port` class, but I don't think separating them is worth avoiding the repetition

godwingrs22 (approved):
This is great and helpful. Thanks @nmussy for your contribution and @msambol for the review.

Issue #29800, opened 2024-04-11
aws_ecs: allow to specify kernel when creating the ecsOptimizedImage


segundovolante:
### Describe the feature

Current CDK api does not allow to specify the kernel version, and you end up with 4.14 which is marked as EOL Jan 2024.

This page says 4.14 has reached EOL: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html 

API link: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.EcsOptimizedImage.html 

It would help if:
*  default kernel is the recommended kernel
* it could be passed like in this construct: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.AmazonLinux2Kernel.html 
* create the AMI id using https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.AmazonLinux2ImageSsmParameter.html 

### Use Case

I want to create an ECS autoscaling group with the recommended kernel or using another supported ami from the ECS optimized image.

### Proposed Solution

Option 1: update the api to use the recommended kernel version (5.10)
Option 2: allow users to pass the kernel version to use, as the AMI exists. 

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.137.0

### Environment details (OS name and version, etc.)

AL2

khushail:
Thanks @segundovolante for requesting this feature. This totally makes sense!

Konosh93:
> it could be passed like in this construct: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.AmazonLinux2Kernel.html

I concur with the proposal to allow manual specification of the kernel version in ecsOptimizedImage. While updating the default kernel version to 5.10 might seem beneficial, it does carry the risk of introducing breaking changes for current users.

Additionally, changing the default kernel would necessitate updates to a large number of integration tests

segundovolante:
Yeah, I agree that the better option long term is to allow to specify the kernel as an optional parameter. This would behave similar to the ec2.

PR #29809, opened 2024-04-12
feat(ecs): enable Kernel 5.10 support on Amazon Linux 2

Konosh93:
### Issue 

#29800

### Reason for this change

Allow using [recommended kernel version](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html) instead of 4.14 which is marked as EOL Jan 2024, by opting in for the new kernel.

> We recommend that you use an AMI with Linux kernel 5.10 because Linux kernel 4.14 reached end-of-life on January 10, 2024.

### Description of changes
- Deprecate the existing `amazonLinux2` in favor of a new `amazonLinux2WithKernel510`.

Unit and integration tests.

### Checklist

----


packages/aws-cdk-lib/aws-ecs/lib/amis.ts
```diff
@@ -205,6 +218,9 @@ export class EcsOptimizedImage implements ec2.IMachineImage {
    * @param hardwareType ECS-optimized AMI variant to use
    */
   public static amazonLinux2023(hardwareType = AmiHardwareType.STANDARD, options: EcsOptimizedImageOptions = {}): EcsOptimizedImage {
+    if (options.kernel) {
+      throw new Error('Kernel version can only be specified for Amazon Linux 2');
+    }
```
lpizzinidev, line 223: I think we should provide the ability to specify the `kernel` property for all Linux generations.
What about declaring dedicated interfaces?
```ts
export interface EcsOptimizedAmazonLinux2023ImageOptions extends EcsOptimizedImageOptions {
  /**
   * The kernel version to use with Amazon Linux 2023
   *
   * @default - The recommended kernel version
   */
  readonly kernel?: ec2.AmazonLinux2022Kernel;
}
...
```

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
```diff
@@ -275,6 +303,7 @@ export class EcsOptimizedImage implements ec2.IMachineImage {
       + (this.generation === ec2.AmazonLinuxGeneration.AMAZON_LINUX_2 ? 'amazon-linux-2/' : '')
       + (this.generation === ec2.AmazonLinuxGeneration.AMAZON_LINUX_2023 ? 'amazon-linux-2023/' : '')
       + (this.windowsVersion ? `Windows_Server-${this.windowsVersion}-English-Full-ECS_Optimized/` : '')
+      + kernelPath
```
lpizzinidev, line 306:
```suggestion
      + (this.kernel ? `${props.kernel}/` : '')
```
Could be simplified after implementing the property for all Linux generations.

lpizzinidev (changes requested):
Thanks 👍
I left some suggestions for adjustments.
Also, does the `recommended` [here](https://github.com/aws/aws-cdk/blob/b0490640a25eb1946aa9db748cf85867e841f243/packages/aws-cdk-lib/aws-ecs/lib/amis.ts#L281) refer to the kernel version? 🤔
If so, we should remove it when `kernel` is specified.

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
Konosh93: I considered that but decided against it for 2 reasons:
1. The [AMI metadata SSM parameters](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/retrieve-ecs-optimized_AMI.html) do not seem to contain explicit kernel path except for Amazon Linux 2 AMIs

2. introducing a new interface like `EcsOptimizedAmazonLinux2023ImageOptions` seems to cause lint errors like:
```
Some packages seem to have undergone breaking API changes. Please avoid.
```

Please let me know if I am misunderstanding something. 🙏

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
Konosh93: Thanks for the suggestion 🙏 
I added a comment about why kernel is specified for Amazon Linux 2 only:
- https://github.com/aws/aws-cdk/pull/29809#discussion_r1564718696

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
lpizzinidev: Thanks for the follow-up, you're right about point 1 👍

However, I think that the idea of adding an interface for Amazon Linux 2 `kernel` property is still valuable, given that we would be able to simplify the implementation and remove the manual validations:
```ts
export interface EcsOptimizedAmazonLinux2ImageOptions extends EcsOptimizedImageOptions {
  /**
   * The kernel version to use with Amazon Linux 2
   *
   * @default - The recommended kernel version
   */
  readonly kernel?: ec2.AmazonLinux2Kernel;
}
```
We would then need to change options type for the `amazonLinux2` method:
```ts
public static amazonLinux2(hardwareType = AmiHardwareType.STANDARD, options: EcsOptimizedAmazonLinux2ImageOptions = {}): EcsOptimizedImage { ... }
```
What do you think?

On point 2, you can allow the change by adding an entry to `allowed-breaking-changes.txt` ([example](https://github.com/aws/aws-cdk/blob/b0490640a25eb1946aa9db748cf85867e841f243/allowed-breaking-changes.txt#L92)).

lpizzinidev: @Konosh93 

Thanks for the follow-up, you're right about point 1 👍

However, I think that the idea of adding an interface for Amazon Linux 2 `kernel` property is still valuable, given that we would be able to simplify the implementation and remove the manual validations:
```ts
export interface EcsOptimizedAmazonLinux2ImageOptions extends EcsOptimizedImageOptions {
  /**
   * The kernel version to use with Amazon Linux 2
   *
   * @default - The recommended kernel version
   */
  readonly kernel?: ec2.AmazonLinux2Kernel;
}
```
We would then need to change options type for the `amazonLinux2` method:
```ts
public static amazonLinux2(hardwareType = AmiHardwareType.STANDARD, options: EcsOptimizedAmazonLinux2ImageOptions = {}): EcsOptimizedImage { ... }
```
What do you think?

On point 2, you can allow the change by adding an entry to `allowed-breaking-changes.txt` ([example](https://github.com/aws/aws-cdk/blob/b0490640a25eb1946aa9db748cf85867e841f243/allowed-breaking-changes.txt#L92)).

Konosh93: @lpizzinidev 

Thanks for the tips and detailed code examples 🙏 

> adding an interface for Amazon Linux 2 kernel property is still valuable

Indeed, I added proposed interface and and updated the breaking change file accordingly.

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
Konosh93: @lpizzinidev 
Just a follow-up about the style of the kernel part in the parameter path, I think it is better to still do the check and throw an error, since the constructor accepts a kernel argument, but only handles the Amazon Linux 2 kernel. 

It might be more confusing If we add a kernel property to the class, given that it only concerns Amazon Linux 2 AMI.
```ts
export class EcsOptimizedImage implements ec2.IMachineImage {
   ...
   private readonly kernel?: ec2.AmazonLinux2Kernel;
   ...
}
```

Konosh93:
> Also, does the recommended [here](https://github.com/aws/aws-cdk/blob/b0490640a25eb1946aa9db748cf85867e841f243/packages/aws-cdk-lib/aws-ecs/lib/amis.ts#L281) refer to the kernel version? 🤔

Thanks for spotting this, it might be I will test and let you know 🙏

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
```diff
@@ -267,6 +287,14 @@ export class EcsOptimizedImage implements ec2.IMachineImage {
...
       throw new Error('This error should never be thrown');
     }
+    const kernelPath = (() => {
+      if (!props.kernel) {
+        return '';
+      } else if (props.generation !== ec2.AmazonLinuxGeneration.AMAZON_LINUX_2) {
+        throw new Error('Kernel version can only be specified for Amazon Linux 2');
+      }
+      return props.kernel.toString() + '/';
+    })();
```
lpizzinidev, line 297:
```suggestion
    if (props.kernel && props.generation !== ec2.AmazonLinuxGeneration.AMAZON_LINUX_2) {
      throw new Error('Kernel version is supported only for Amazon Linux 2');
    }
```
I'd prefer to keep it simple instead of using a function.
We can the concatenate the string with `(props.kernel ? `${props.kernel}/` : '')`.

packages/aws-cdk-lib/aws-ecs/test/amis.test.ts
```diff
@@ -24,4 +25,46 @@ describe('amis', () => {
...
+
+      // THEN
+      const assembly = app.synth();
+      const parameters = assembly.getStackByName(stack.stackName).template.Parameters;
+      expect(Object.entries(parameters).some(
+        ([k, v]) => k.startsWith('SsmParameterValueawsserviceecsoptimizedamiamazonlinux2') && (v as any).Default.includes('/aws/service/ecs/optimized-ami/amazon-linux-2/recommended/image_id'),
+      )).toEqual(true);
+    });
+  });
+
```
lpizzinidev, line 69: Can you please add coverage for the `Amazon Linux 2` validation?

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
```diff
@@ -97,6 +97,13 @@ export interface EcsOptimizedAmiProps {
    * @default false
    */
   readonly cachedInContext?: boolean;
+
+  /**
+   * What kernel version of Amazon Linux 2 to use
+   *
+   * @default none, uses the recommended kernel version
+   */
+  readonly kernel?: ec2.AmazonLinux2Kernel;
```
lpizzinidev, line 106:
```suggestion
  /**
   * Kernel version used.
   * Only supported for Amazon Linux 2.
   *
   * @default none, uses the recommended kernel version
   */
  readonly kernel?: string;
```
What if we need to support the property for other Linux images in the future? 🤔
```

**Review by lpizzinidev — CHANGES_REQUESTED** [review]
Thanks 👍
Left some comments for further adjustments, feel free to comment on those.

File: packages/aws-cdk-lib/aws-ecs/lib/amis.ts (hunk 3 above)
Comment (Konosh93) [reply packages/aws-cdk-lib/aws-ecs/lib/amis.ts:297]: Thanks for the suggestion. I will fix it accordingly. It will look like:
```ts
+ (props.kernel ? props.kernel.toString() + '/' : '')
```

packages/aws-cdk-lib/aws-ecs/test/amis.test.ts
Konosh93: @lpizzinidev Can you explain more?

Do you mean adding tests that should throw errors when specifying the kernel but the AMI is not Amazon Linux 2?
If so, then I think that won't be necessary because TypeScript will throw a compile error as the `kernel` property is defined only in `EcsOptimizedImage.amazonLinux2`'s arguments.

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
Konosh93: That's better. Thanks a lot 🙏

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
Konosh93: Looks good, thannks.

TheRealAmazonKendra (changes requested):
Linux 2 seems to be an exception to the usual pattern because AWS extended it's lifecycle by two years, necessitating a new kernel. It doesn't look like that will be the pattern going forward and that, for the remainder of it's supported life, only 5.10 will be valid. Given all that, I'm not a fan of adding `kernel` as a prop/option.

I think the best path forward on this one is to add a feature flag and to change the kernel based off that. Since the kernel change requires a replacement, we definitely want users to have to make the active choice to switch it on.

Konosh93:
@TheRealAmazonKendra, thank you for your insightful review. I find myself pondering a particular aspect regarding the use of FeatureFlags in our current setup.

Given that the class `EcsOptimizedImage` does not inherit from `Construct`, it necessitates passing the scope directly to the constructor to facilitate FeatureFlags instantiation. I'm considering modifying the static method to allow scope passing by the user, like so:
```ts
public static amazonLinux2(hardwareType = AmiHardwareType.STANDARD, options: EcsOptimizedImageOptions = {}, scope?: Construct): EcsOptimizedImage {
```
To preserve compatibility, making `scope` optional seems imperative. Yet, this approach raises a concern: if the scope is omitted, how do we ascertain whether the feature is activated? Without this information, enabling the feature might have no effect. For example:
```ts
  private constructor(props: EcsOptimizedAmiProps, scope?: Construct) {
    this.hwType = props && props.hardwareType;

    if (props.windowsVersion) {
      this.windowsVersion = props.windowsVersion;
    } else if (props.generation) {
      this.generation = props.generation;
    } else {
      throw new Error('This error should never be thrown');
    }

    const kernelSpecified = scope && FeatureFlags.of(scope).isEnabled(cx_api.ECS_OPTIMIZED_IMAGE_KERNEL_5_10);
```
`kernelSpecified` is true only if scope is provided and the feature is turned on.

Could you share your thoughts on this approach, especially regarding the management of scope and feature flags? How might others be handling similar scenarios?

TheRealAmazonKendra:
> > I think the best path forward on this one is to add a feature flag and to change the kernel based off that.
> 
> @TheRealAmazonKendra, thank you for your insightful review. I find myself pondering a particular aspect regarding the use of FeatureFlags in our current setup.
> 
> Given that the class `EcsOptimizedImage` does not inherit from `Construct`, it necessitates passing the scope directly to the constructor to facilitate FeatureFlags instantiation. I'm considering modifying the static method to allow scope passing by the user, like so:
> 
> ```ts
> public static amazonLinux2(hardwareType = AmiHardwareType.STANDARD, options: EcsOptimizedImageOptions = {}, scope?: Construct): EcsOptimizedImage {
> ```
> 
> To preserve compatibility, making `scope` optional seems imperative. Yet, this approach raises a concern: if the scope is omitted, how do we ascertain whether the feature is activated? Without this information, enabling the feature might have no effect. For example:
> 
> ```ts
>   private constructor(props: EcsOptimizedAmiProps, scope?: Construct) {
>     this.hwType = props && props.hardwareType;
> 
>     if (props.windowsVersion) {
>       this.windowsVersion = props.windowsVersion;
>     } else if (props.generation) {
>       this.generation = props.generation;
>     } else {
>       throw new Error('This error should never be thrown');
>     }
> 
>     const kernelSpecified = scope && FeatureFlags.of(scope).isEnabled(cx_api.ECS_OPTIMIZED_IMAGE_KERNEL_5_10);
> ```
> 
> `kernelSpecified` is true only if scope is provided and the feature is turned on.
> 
> Could you share your thoughts on this approach, especially regarding the management of scope and feature flags? How might others be handling similar scenarios?

Ah, this is a really good point. So, we usually use a `_bind()` internal function for this but I'm kind of thinking that doing so might be more of a pain than it's worth, and without doing much digging, it might not really be feasible to know everywhere these get used. 

In light of that, I'll revert back to what was going to be my original suggestion that I didn't like very much but that is probably still our best path forward: let's just deprecate the original function and make one that will include the new kernel version. We should probably also deprecate the `amazonLinux` function as well, since AL1 went EOL in January of this year. As for naming the new function, this is the part I didn't like very much because the best I could think of was `amazonLinux2Kernel5` or something like that. Any thoughts on a better name for it?

Konosh93:
@TheRealAmazonKendra 

> In light of that, I'll revert back to what was going to be my original suggestion that I didn't like very much but that is probably still our best path forward: let's just deprecate the original function and make one that will include the new kernel version. We should probably also deprecate the `amazonLinux` function as well, since AL1 went EOL in January of this year. As for naming the new function, this is the part I didn't like very much because the best I could think of was `amazonLinux2Kernel5` or something like that. Any thoughts on a better name for it?

Thank you very much for the suggestion. I added the deprecated decorators to both `amazonLinux` and `amazonLinux2` and added a new static method `amazonLinux2Kernel510` which sets the new kernel in the SSM path.

Let me know if this resembles what you had in mind.

TheRealAmazonKendra:
@Mergifyio update

TheRealAmazonKendra:
@Mergifyio update

Konosh93:
@Mergifyio update

Konosh93:
@TheRealAmazonKendra Can we go ahead and merge this change, or is it missing something?

Konosh93:
@TheRealAmazonKendra Do you any other suggestions to add to this PR?

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
```diff
@@ -97,6 +97,14 @@ export interface EcsOptimizedAmiProps {
...
    */
   readonly cachedInContext?: boolean;
+
+  /**
+   * Kernel version used.
+   * Only supported for Amazon Linux 2.
+   *
+   * @default none, uses the recommended kernel version
+   */
+  readonly kernel?: string;
```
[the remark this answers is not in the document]
comcalvi: As @TheRealAmazonKendra said, this is a deprecated interface, so we shouldn't be adding a new property to this...except, there's a note that says `remove @deprecated` above in V2. @rix0rrr I'm thinking we should un-deprecate this property, WDYT? 

This should not be a string value, it should be an enum. We should not force users to call the `.toString()` on the enum, they should just pass it directly. 

The enum provides type discoverability; if the `kernel` property is only a string type, there's nothing the IDE can do to tell you that you should really be passing the result of a `toString()` call. Instead, with the enum, it tells you what type is expected, so the user discovers what they need to pass much more easily.

comcalvi (changes requested):

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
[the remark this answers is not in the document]
Konosh93: @comcalvi Thanks for the review.

1) Regarding the deprecated interface:
> This is a deprecated interface, so we shouldn't be adding a new property to this.

There was some back and forth on this point. I opted for this workaround to avoid introducing a breaking change, which is deemed unacceptable as per [this comment](https://github.com/aws/aws-cdk/pull/29809#discussion_r1568061197).

> So, it's true that you can get around our linter by adding that, but this is a stable module, so we can't actually accept a breaking change here.

2) On using an enum instead of a string:
> This should not be a string value, it should be an enum. We should not force users to call the .toString() on the enum.

I have updated the implementation to pass the enum directly. However, I didn't foresee this option being passed by a user. The constructor is private, and only the static methods are instantiating the class. Therefore, there was no way for any user (except through the static methods) to call the constructor directly.

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
```diff
@@ -97,6 +97,13 @@ export interface EcsOptimizedAmiProps {
    * @default false
    */
   readonly cachedInContext?: boolean;
+  /**
+   * Kernel version used.
+   * Only supported for Amazon Linux 2.
+   *
+   * @default none, uses the recommended kernel version
+   */
+  readonly kernel?: ec2.AmazonLinux2Kernel;
```
comcalvi, line 106: Discussed offline, this should not be undeprecated...sorry about that, I was wrong. This entire enum should be deprecated and replaced with the linked `EcsOptimizedImage` instead.

comcalvi (dismissed):

packages/aws-cdk-lib/aws-ecs/lib/amis.ts
Konosh93: @comcalvi I replied to your comment [here](https://github.com/aws/aws-cdk/pull/29809#discussion_r1644357838). Can you have a look?

>  This entire props should be deprecated and replaced with the linked EcsOptimizedImage instead.

`EcsOptimizedImage` is a class and its private constructor is still using the deprecated `EcsOptimizedAmiProps`. There are no other Props interfaces for `EcsOptimizedImage` constructor. 

Is it time to remove `@deprecated` already?

comcalvi: It looks like we deprecated the props and decided to create a bunch of factory methods that take an `Options` instead of `props`. I think your original solution was actually correct; leave these as deprecated (since they should be marked internal, and if we remove the deprecation we'll never remember to fix this) and just add the new property.

@TheRealAmazonKendra is correct about this, in principle, since a deprecated marker implies that these props have been directly replaced by something else. That is not what is going on here. Instead, these props are supposed to be hidden away behind a set of factory methods. So, even if we had not missed these when we did v2, the optimal would still be what you've done here; add `kernel` to these props. 

Sorry for the churn on this one, please leave the `@deprecated` and add the new property to those deprecated props.

Konosh93: @comcalvi Thanks for the update 🙏 
I am not entirely confident I am following what has been the final decision regarding how this change should be incorporated but I tried to revert to a previous state in this branch history; namely (c107fe0f).

Please let me know if I misunderstood your last comment. Thanks again

Issue #29813, opened 2024-04-12
aws_lambda: Version.fromVersionArn creates invalid Version object


rehos:
### Describe the bug

The version object created by Version.fromVersionArn is created with a lambda property that still contains the version qualifier in its function url. Which can cause CloudFormation deployment errors. For example when you create a lambda alias from it and then try to use that alias fro a lambda function url.

### Expected Behavior

I expect Version.fromVersionArn to create the equivalent object as Version.fromVersionAttributes when referencing the same version.

The following test when using Version.fromVersionAttributes succeeds:

```
import {
    Stack,
    aws_lambda as lambda,
} from 'aws-cdk-lib';

test('lambda.Version.fromVersionAttributes generates Version with correct lambda property', () => {
    const stack = new Stack();
    const version = lambda.Version.fromVersionAttributes(stack, 'ExampleVersion', {
        lambda: lambda.Function.fromFunctionArn(stack, 'Examplelambda', 'arn:aws:lambda:eu-central-1:123456789012:function:ExampleLambda'),
        version: '13',
    });

    expect(version.version).toStrictEqual('13') // Ok
    expect(version.lambda.functionArn).toStrictEqual('arn:aws:lambda:eu-central-1:123456789012:function:ExampleLambda') // Ok
})

```

### Current Behavior

However the following test when using Version.fromVersionArn fails because the version qualifier is not stripped from the arn:

```
import {
    Stack,
    aws_lambda as lambda,
} from 'aws-cdk-lib';

test('lambda.Version.fromVersionArn generates Version with correct lambda property', () => {
    const stack = new Stack();
    const version = lambda.Version.fromVersionArn(stack, 'ExampleVersion', 'arn:aws:lambda:eu-central-1:123456789012:function:ExampleLambda:13');

    expect(version.version).toStrictEqual('13') // Ok
    expect(version.lambda.functionArn).toStrictEqual('arn:aws:lambda:eu-central-1:123456789012:function:ExampleLambda') // Not Ok
})
```

### Reproduction Steps

Run the above tests

### CDK CLI Version

2.137.0

### Node.js Version

v20.11.0

### OS

macOS Sonoma 14.4.1

### Language

TypeScript

### Language Version

Typescript (5.4.5)

pahud:
Yes, it should be a tiny fix here

[...]

PR #29820, opened 2024-04-12
fix(lambda): version.fromVersionArn creates invalid Version object

pahud:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29813

### Reason for this change

improve the fromFunctionArn()  to better handle the input ARN

### Description of changes

fromFunctionArn() does not handle the ARN correctly if the input ARN has trailing version or alias.

See unit tests

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-lambda/lib/function.ts
+++ b/packages/aws-cdk-lib/aws-lambda/lib/function.ts
@@ -672,8 +672,8 @@ export class Function extends FunctionBase {
    */
   public static fromFunctionArn(scope: Construct, id: string, functionArn: string): IFunction {
     /**
-     * First, validate the functionArn string. If it's ARN with trailing version or alias,
-     * we trim off the trailing to retrieve the real functionArn.
+     * If the functionArn has a trailing version or alias (more than 7 parts when split by ":",
+     * we trim off the trailing version/alias to retrieve the real functionArn.
      * See lambda resource ARN format here: https://docs.aws.amazon.com/lambda/latest/dg/lambda-api-permissions-ref.html
      */
     const parts = functionArn.split(':');


pahud:
Exemption Request - unit tests added

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -671,7 +671,18 @@ export class Function extends FunctionBase {
    * in the same account and region as the stack you are importing it into.
    */
   public static fromFunctionArn(scope: Construct, id: string, functionArn: string): IFunction {
-    return Function.fromFunctionAttributes(scope, id, { functionArn });
+    /**
+     * First, validate the functionArn string. If it's ARN with trailing version or alias,
+     * we trim off the trailing to retrieve the real functionArn.
```
paulhcsun, line 676:
```suggestion
     * If the functionArn has a trailing version or alias (more than 7 parts when split by ":",
     * we trim off the trailing version/alias to retrieve the real functionArn.
```
Let me know if you think this makes more sense. I don't think we need to say that we're validating the functionArn string.

GavinZZ (approved):
LGTM

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -671,7 +671,18 @@ export class Function extends FunctionBase {
...
+     * See lambda resource ARN format here: https://docs.aws.amazon.com/lambda/latest/dg/lambda-api-permissions-ref.html
+     */
+    const parts = functionArn.split(':');
```
paulhcsun, line 679:
```suggestion
    const arnParts = functionArn.split(':');
```

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -671,7 +671,18 @@ export class Function extends FunctionBase {
...
+    if (parts.length > 7) {
```
paulhcsun, line 680: I'm not really a fan hardcoding this check with `> 7` but I also don't really foresee this ever changing in terms of sections of an ARN so I guess it's ok.

paulhcsun (changes requested):
Left a couple comments

paulhcsun (approved):
Thanks @pahud!

rehos:
I reviewed the code and I don't think it works when the function arn is an unresolved token (i.e. passed as a parameter to the CloudFormation stack). See my comment [here](https://github.com/aws/aws-cdk/commit/819888470248fed2ddbdac689f14254688eef88f#r140934472)

packages/aws-cdk-lib/aws-lambda/lib/function.ts
rehos: @paulhcsun see my comment [here](https://github.com/aws/aws-cdk/commit/819888470248fed2ddbdac689f14254688eef88f#r140934472) too.  It is about the fact that the functionArn can be a token instead of a literal string too (for example when it is passed as a parameter to the CloudFormation stack). I also added a code snippet on how to handle that case too.

pahud: fromFunctionArn only accepts functionArn as `string`.  Should we consider that condition?

rehos: @pahud and @paulhcsun in the following code snippet is a stack that accepts the lambda function version arn as a parameter (for example because the actual function was created in another stack earlier in the ci/cd pipeline). 

```
const myFunctionVersionArn = new cdk.CfnParameter(stack, 'MyFunctionVersionArn');

const version = lambda.Version.fromVersionArn(stack, 'MyVersionFunction', myFunctionVersionArn.valueAsString);
```

The arn that is passed to fromVersionArn is a tokenized string (value not known at compile time) and is eventually passed to fromFunctionArn because it is called by fromVersionArn under the hood. And fromVersionArn passes it down without stripping the version part in the arn.

At the moment the fix has been applied to Function.fromFunctionArn, but I think you may assume people pass the correct arn to that function. The real problem is in fromVersionArn. This function "knows" it will receive a version arn so it also should know that it should strip the version to get a valid functionArn. So I think [this line](https://github.com/aws/aws-cdk/blob/5a918d12ec0455626b5630b35aa47e6d9da6df7c/packages/aws-cdk-lib/aws-lambda/lib/lambda-version.ts#L125) should be fixed. If you do that then the `Fn.join(Fn.select(.., Fn.split...)` will only be generated when it is really needed.

```
public static fromVersionArn(scope: Construct, id: string, versionArn: string): IVersion {
      const version = extractQualifierFromArn(versionArn);

      let functionArn: string;

      if (Token.isUnresolved(versionArn)) {
            const parts =  Fn.split(':', versionArn);
            functionArn = cdk.Fn.join(':', [
                Fn.select(0, parts),
                Fn.select(1, parts),
                Fn.select(2, parts),
                Fn.select(3, parts),
                Fn.select(4, parts),
                Fn.select(5, parts),
                Fn.select(6, parts),
            ]);
      } else {
            const parts = versionArn.split(':');
            functionArn = parts.slice(0, 7).join(':');
      }

      const lambda = Function.fromFunctionArn(scope, `${id}Function`, functionArn);
  
      class Import extends QualifiedFunctionBase implements IVersion {
        public readonly version = version;
        public readonly lambda = lambda;
        public readonly functionName = `${lambda.functionName}:${version}`;
        public readonly functionArn = versionArn;
        public readonly grantPrincipal = lambda.grantPrincipal;
        public readonly role = lambda.role;
        public readonly architecture = lambda.architecture;
  
        protected readonly qualifier = version;
        protected readonly canCreatePermissions = this._isStackAccount();
  
        public addAlias(name: string, opts: AliasOptions = {}): Alias {
          return addAlias(this, this, name, opts);
        }
  
        public get edgeArn(): string {
          if (version === '$LATEST') {
            throw new Error('$LATEST function version cannot be used for Lambda@Edge');
          }
          return this.functionArn;
        }
      }
      return new Import(scope, id);
    }
```

mindstorms6:
I think this change is backwards incompatible - it broke our stack.

Previously, if you passed a function arn with a version, the `getFunctionArn` would retain that version. It no longer does - and this caused an outage for us (by causing an APIG authorizer to point to the "root" lambda function instead of a specific alias - where it didn't haver permissions)

rehos:
Sorry to hear the change caused an outage. My issue still isn't fixed with the change (see also my [comment](https://github.com/aws/aws-cdk/pull/29820#discussion_r1570054030) above, in which I also suggested not to change Function.fromFunctionArn, but instead fix Version.fromVersionArn).

This still generates an invalid stack:

```
const DemoClientVersionArn = new cdk.CfnParameter(this, 'DemoClientVersionArn');

const demoClientVersion = lambda.Version.fromVersionArn(this, 'DemoClientVersion', DemoClientVersionArn.valueAsString);

const demoClientCurrentAlias = new lambda.Alias(this, 'DemoClientCurrentAlias', {
    aliasName: 'current',
    version: demoClientVersion,
});

// The function url resource created by the CDK causes a CloudFormation deployment error
// because the version qualifier is still part of the functionArn of the lambda.Function
// created by lambda.Version.fromVersionArn
const demoClientFunctionUrl = new lambda.FunctionUrl(this, 'DemoClientFunctionUrl', {
    authType: lambda.FunctionUrlAuthType.AWS_IAM,
    invokeMode: lambda.InvokeMode.RESPONSE_STREAM,
    function: demoClientCurrentAlias,
});
```

PR #29822, opened 2024-04-12
feat(settings): allow project local settings with cdk.local.json

### LINKED ISSUE #29821: (settings): ability to specific developer's local settings via cdk.local.json
### Describe the feature

I'd want my developers to be able to specify certain context settings via untracked `cdk.local.json` file.
Something similar to `.env.local` that is untracked and presents an opportunity for developers to configure their stacks without stepping on each other toes. 

I am aware that there is `~/cdk.json` being a global configuration developer can apply but this particular's setting issue is its global impact on anything that's developed/deployed whereas I am looking for more isolated approach that allows to define settings per project.

### Use Case

Point here is not to bloat either `cdk.json` or `cdk.context.json` files that are tracked, or in my case can be tracked when it comes to `cdk.context.json`. Reason for not tracking `cdk.context.json` lies in a fact that my organization develops various services as L3 components (packaged as NPM packages) that are combined with `preview`. Preview actually uses the L3 component to deploy isolated micro service. Given that all developers have their accounts it would be quite problematic to manage `cdk.context.json` if it were comitted.

### Proposed Solution

Modify [settings.ts](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/settings.ts) by adding `export const PROJECT_LOCAL_CONFIG = 'cdk.local.json';` and adjusting code to load it up similarly to `USER_DEFAULTS`. 

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.137.0

MacOS Sierra

kornicameister:
### Issue # (if applicable)

Closes #29821

### Reason for this change

Ability to define project's local settings via `cdk.local.json`

### Description of changes

Add 4th source of settings that are localized to given application.
Point is that `cdk.local.json` should not be tracked in repository.

I haven't added tests yet, will do once change is accepted in proposed shape.

### Checklist

----

### DIFF
--- a/packages/aws-cdk/lib/settings.ts
+++ b/packages/aws-cdk/lib/settings.ts
@@ -8,6 +8,7 @@ import * as util from './util';
 export type SettingsMap = {[key: string]: any};
 
 export const PROJECT_CONFIG = 'cdk.json';
+export const PROJECT_LOCAL_CONFIG = 'cdk.local.json';
 export const PROJECT_CONTEXT = 'cdk.context.json';
 export const USER_DEFAULTS = '~/.cdk.json';
 
@@ -111,6 +112,7 @@ export class Configuration {
    */
   public async load(): Promise<this> {
     const userConfig = await loadAndLog(USER_DEFAULTS);
+    const localConfig = await loadAndLog(PROJECT_LOCAL_CONFIG);
     this._projectConfig = await loadAndLog(PROJECT_CONFIG);
     this._projectContext = await loadAndLog(PROJECT_CONTEXT);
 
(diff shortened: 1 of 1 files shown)


packages/aws-cdk/lib/settings.ts
```diff
@@ -135,6 +140,7 @@ export class Configuration {
     this.settings = this.defaultConfig
       .merge(userConfig)
       .merge(this.projectConfig)
+      .merge(localConfig)
```
kornicameister, line 143: I think it makes sense to override `cdk.json` but still let CLI take precedence.

PR #29849, opened 2024-04-15
fix(batch): allow `endNode` to be optional for multinode containers

### LINKED ISSUE #29415: (aws-batch): Cannot omit the start/end of target node on batch.MultiNodeContainer
### Describe the feature

When using CDK to deploy aws batch multi-node jobs, we have to specify both the start_node and end_node, FYI: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_batch/MultiNodeContainer.html#aws_cdk.aws_batch.MultiNodeContainer. But refer to the https://docs.aws.amazon.com/batch/latest/APIReference/API_NodeRangeProperty.html#API_NodeRangeProperty_Contents, it is possible to omit start_node/end_node here. 

### Use Case

Our use case is as following: after I deploy the aws batch cloud infrastructure, I can use boto3 with python to submit a job like:

```
import boto3 

response = batch_client.submit_job(
    jobName=job_name,
    jobQueue=job_queue,
    jobDefinition=multi_job_definition,
    parameters=job_parameters,
    nodeOverrides={
        "numNodes": overridden_num_nodes,
    },
)
```

Currently It will throw error like: 
```
botocore.errorfactory.ClientException: An error occurred (ClientException) when calling the SubmitJob operation: NumNodes override can only be applied if the job definition has at least 1 target node without a range_end i.e (:) or (range_start:).
```

This is because in CDK we have to specify the start_node and end_node. But if we support to omit the end_node in CDK, we can avoid this problem. And this is valid according to the aws batch multi-node job definition. Currently the only work around is to create another job_definition based on what we deploy and modify the target node in the batch console.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.131.0

Amazon Linux 2

comcalvi:
### Issue #

Closes #29415

### Reason for this change

The Batch L2 forces users to specify endNode, even though this is technically optional, on batch multinode jobs. This was initially done intentionally, since with endNode we can compute the always-required `numNodes` for the user. `numNodes` can be overwritten when actually submitting a job, but only if the job has at least one container with `endNode` omitted. 

If any container omits `endNode`, we can no longer compute `numNodes`, which is a required property; in this case, the user must specify it.

### Description of changes

Adds a new optional property, `numNodes`, and makes `endNode` optional on multinode containers. 

unit and integration tests

### Checklist

----


comcalvi:
ah...naturally weakening a property is a breaking change

PR #29856, opened 2024-04-16
feat(lambda): add grantInvokeLatestVersion to grant invoke only to latest function version

### LINKED ISSUE #20177: (lambda): Function.grant_invoke should not grant for all versions
### Describe the bug

While upgrading from 1.140.0 to 2.20.0, we noticed that the behavior of `aws_lambda.Function.grant_invoke` has changed. Previously it granted `lambda:InvokeFunction` just on the function ARN itself. Now it also grants `lambda:InvokeFunction` on `${arn}:*`, meaning all function versions.

This change does not make sense. At best, granting this wildcard permission is pointless, because CDK doesn't publish function versions. At worst, this is actually a potential security issue, because when I grant invoke rights to a principal, I mean for them to invoke it without any version number, which means `$LATEST`. If they have permission to invoke any version, then it is possible for them to invoke an older version that is not intended to be used anymore.

### Expected Behavior

`grant_invoke` should retain its original behavior of only granting the ability to invoke the function itself, without an explicit version. If for some reason someone wants to grant wildcard access to all versions, there can be an optional boolean parameter to `grant_invoke`.

### Current Behavior

See above.

### Reproduction Steps

Create an `aws_lambda.Function` and use `grant_invoke` to grant invoke rights to some principal. Then look at the IAM policy that is generated.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.20.0

### Framework Version

_No response_

### Node.js Version

16.14.2

### OS

Alpine 3.15

### Language

Python

### Language Version

3.9.12

### Other information

_No response_

roger-zhangg:
### Issue # (if applicable)

Closes #20177

### Reason for this change

`fn.grantInvoke()` will grant invoke permission to invoke both the latest version and all pervious version of the lambda function. We can see this behavior could bring some security concern for some of our customers.

### Description of changes

We provides a new function `fn.grantInvokeLatestVersion()`  to grant invoke only to the Latest version of function and the unqualified lambda arn

Example:
```ts
// Grant permissions to a service
declare const fn: lambda.Function;
const principal = new iam.ServicePrincipal('my-service');

fn.grantInvokeLatestVersion(principal);

```

Added unit tests and integration tests.
When using `fn.grantInvokeLatestVersion()` granted principle to invoke a function's past version, it will get the following error:
```
An error occurred (AccessDeniedException) when calling the Invoke operation: User: {$principle} is not authorized to perform: lambda:InvokeFunction on resource: {$LambdaArn:$version} because no identity-based policy allows the lambda:InvokeFunction action
```

### Alternative design (to discuss)
setup a `grantInvokeProp` including `grantVersionAccess` flag to pass in the `grantInvokeLatestVersion` instead using `grantVersionAccess` flag directly on `grantInvokeLatestVersion`
-> This is discussed in the comments, I agree having props will have future extensibility but usually for grant methods specifically we haven't seen before. So we will not add prop to the new function `grantInvokeLatestVersion`

### Checklist

----


nmussy (changes requested):
A cleaner way to implement this feature would be to extend the current implementation of `grantInvoke` and to add a `onlyGrantLatestVersion` parameter with a default value. Using a `Props` object is also preferable for future extensibility:

```ts
interface GrantInvokeProps {
  onlyGrantLatestVersion?: boolean;
}
grantInvoke(grantee: iam.IGrantable, { onlyGrantLatestVersion = false }: GrantInvokeProps = {}): iam.Grant;
```

This would not cause a breaking change, given existing calls to `grantInvoke` would remain correct and unchanged.

packages/aws-cdk-lib/aws-lambda/lib/function-base.ts
```diff
@@ -418,20 +431,29 @@ export abstract class FunctionBase extends Resource implements IFunction, ec2.IC
...
+   *  - When set to `false`, both the function and functions with specific versions can be invoked.
+   *  - When set to `true`, only the function without a specific version (`$Latest`) can be invoked.
    */
-  public grantInvoke(grantee: iam.IGrantable): iam.Grant {
+  public grantInvoke(grantee: iam.IGrantable, props: GrantInvokeProps = {}): iam.Grant {
     const hash = createHash('sha256')
       .update(JSON.stringify({
         principal: grantee.grantPrincipal.toString(),
         conditions: grantee.grantPrincipal.policyFragment.conditions,
+        onlyGrantLatestVersion: props.onlyGrantLatestVersion,
```
nmussy, line 444: Switching between `grantInvoke(grantee)` and `grantInvoke(grantee, { onlyGrantLatestVersion: false })` would cause an unnecessary hash change

```suggestion
        onlyGrantLatestVersion: props.onlyGrantLatestVersion ?? false,
```

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -201,6 +201,15 @@ You can also restrict permissions given to AWS services by providing
 a source account or ARN (representing the account and identifier of the resource
 that accesses the function or layer).
 
+**Important**: By default `fn.grantInvoke()` grants permission to the principal to invoke any version of the function, including all past ones. If you only want the principal to invoke the latest version, use `grantInvoke(grantee, { onlyGrantLatestVersion:true })`.
```
nmussy, line 204:
```suggestion
**Important**: By default `fn.grantInvoke()` grants permission to the principal to invoke any version of the function, including all past ones. If you only want the principal to be granted permission to invoke the latest version, use `grantInvoke(grantee, { onlyGrantLatestVersion: true })`.
```

You can also use a Markdown quote to increase the visibility, [e.g. this note](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2-readme.html#port-ranges)

```
> By default...
```

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
nmussy, line 1: Could you add unit tests with imported Lambda functions as well? I am especially anticipating issues with the following:

```ts
lambda.Function.fromFunctionArn(
  stack,
  'unqualified',
  'arn:aws:lambda:us-east-1:123456789012:function:my-function'
);

lambda.Function.fromFunctionArn(
  stack,
  'v1',
  'arn:aws:lambda:us-east-1:123456789012:function:my-function:1'
);

lambda.Function.fromFunctionArn(
  stack,
  'latest',
  'arn:aws:lambda:us-east-1:123456789012:function:my-function:$LATEST'
);
```

If you `onlyGrantLatestVersion` for the first example, I would assume it would actually only grant the ARN version, which might not actually be the latest version

packages/aws-cdk-lib/aws-lambda/lib/function-base.ts
```diff
@@ -418,20 +431,29 @@ export abstract class FunctionBase extends Resource implements IFunction, ec2.IC
...
       .digest('base64');
     const identifier = `Invoke${hash}`;
 
     // Memoize the result so subsequent grantInvoke() calls are idempotent
     let grant = this._invocationGrants[identifier];
     if (!grant) {
-      grant = this.grant(grantee, identifier, 'lambda:InvokeFunction', this.resourceArnsForGrantInvoke);
+      let resouceArns = this.resourceArnsForGrantInvoke;
+      if (props.onlyGrantLatestVersion) {
+        resouceArns = [this.functionArn];
```
nmussy, line 454: Is `functionArn` always fully qualified? If it is, what happens if `functionArn` points to a specific version, e.g. `my-function:1`, but other versions were published since? You are no longer granting the latest version, just "a" version

See the comment on the unit tests file below

packages/@aws-cdk-testing/framework-integ/test/aws-lambda/test/integ.permissions.ts
nmussy, line 1: The integration feels a bit lackluster here, we're only testing that the policy is synthetically correct, not that the grants work as expected

Adding integration assertions to check that the functions are invokable with the provided examples would be ideal, either with `AwsApiCall`s, or by adding an API Gateway to integrate the Lambdas and running `HttpApiCall`s

packages/aws-cdk-lib/aws-lambda/lib/function-base.ts
```diff
@@ -227,6 +227,19 @@ export interface FunctionAttributes {
...
 }
 
+/**
+ * Parameters to pass into grantInvoke method
+ */
+export interface GrantInvokeProps {
+  /**
+   * Controls whether to grant invoke access to all function versions. Defaults to `false`.
+   *  - When set to `false`, both the function and functions with specific versions can be invoked.
+   *  - When set to `true`, only the function without a specific version (`$Latest`) can be invoked.
```
nmussy, line 237: Is this accurate? I would imagine that at least `functionname:$LATEST` would be invokable, and probably even `functionname:3`, if 3 was the latest published version

nmussy (changes requested):

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
roger-zhangg: Thanks for the hint, and this is interesting. For the current design `onlyGrantLatestVersion` will grant permission to `this.functionArn`. While the previous design without `onlyGrantLatestVersion` grants to `[this.functionArn, ${this.functionArn}:*]`. I would assume if `this.functionArn` is something like `arn:aws:lambda:us-east-1:123456789012:function:my-function:1` the previous method won't work either. I'll update once I checked this out.

packages/aws-cdk-lib/aws-lambda/lib/function-base.ts
roger-zhangg: Thanks for catching this, I think you are right👍

packages/@aws-cdk-testing/framework-integ/test/aws-lambda/test/integ.permissions.ts
roger-zhangg: I do agree only testing synthetically is kind of wired, but doing `AwsApiCalls` seems to be testing against Lambda/IAM rather than testing CDK. So I just followed the previous test behavior

packages/aws-cdk-lib/aws-lambda/lib/function-base.ts
```diff
@@ -95,12 +95,12 @@ export interface IFunction extends IResource, ec2.IConnectable, iam.IGrantable {
   /**
    * Grant the given identity permissions to invoke this Lambda
    */
-  grantInvoke(identity: iam.IGrantable): iam.Grant;
+  grantInvoke(grantee: iam.IGrantable, props?: GrantInvokeProps): iam.Grant;
```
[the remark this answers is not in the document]
godwingrs22: updating the existing parameter name `identity` would it cause a breaking change?

godwingrs22:
1. Instead of adding props to the existing grantInvoke method can we create new method instead as stated in the description? Because I saw most of our grant methods in other modules accepts only one parameter grantee and doesn't take props as a param. Hence we propose to follow the similar approach.

Examples:
[grantReadData(grantee)](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_dynamodb.TableV2.html#grantwbrreadwbrdatagrantee) in `TableV2` Construct of `DynamoDB`
[grantManageConnections(identity)](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigatewayv2.WebSocketApi.html#grantwbrmanagewbrconnectionsidentity) in `WebSocketApi` construct of `apigateway`

2. Also for naming the new method, we suggest not to have `V2` in the method name and have something like `grantInvokeLatestVersion()` or  `grantInvokeOnlyLatestVersion()`

roger-zhangg:
Hi @godwingrs22 thanks for the review, I can see your point. But for the naming, If we name the new function like `grantInvokeLatestVersion()` Then maybe we don't need to provide props to the new function. But this also eliminates possible extensibility. What do you think?

godwingrs22:
Hi @roger-zhangg, Yes with new method approach we don't need to provide props. I agree having props will have future extensibility but usually for grant methods specifically we haven't seen before.

`grantInvokeLatestVersion(grantee: iam.IGrantable)`

roger-zhangg:
Hi @godwingrs22 should we consider supporting granting invoke to a certain version? Could be something like `grantInvokeVersion(grantee: iam.IGrantable, version?: IVersion)` and defaults to `$LATEST`

nmussy:
A good compromise might be to have both `grantInvokeLatestVersion(grantee: iam.IGrantable)` and `grantInvokeVersion(grantee: iam.IGrantable, version: IVersion)`, with the former calling the latter. We'd get both the more commonly used `$LATEST` grant ,and the flexibility of pinpointing a specific version

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
roger-zhangg: Update: I tried out `fromFunctionArn` and seems CDK doesn't allow calling `grantInvoke` on a imported Lambda function 

### sample code
```typescript
const func2 = lambda.Function.fromFunctionArn(
  this,
  'v1',
  'arn:aws:lambda:us-east-1:123456789012:function:my-function:1'
);

const lambdaRole = new iam.Role(this, 'LambdaRole', {
  assumedBy: new iam.ServicePrincipal('lambda.amazonaws.com'),
});

func2.grantInvoke(lambdaRole);
```

### error
```
/local/home/ruojiazh/proj/aws-cdk/packages/aws-cdk-lib/aws-lambda/lib/function-base.ts:563
            throw new Error('Cannot modify permission to lambda function. Function is either imported or $LATEST version.\n'
                  ^
Error: Cannot modify permission to lambda function. Function is either imported or $LATEST version.
If the function is imported from the same account use `fromFunctionAttributes()` API with the `sameEnvironment` flag.
If the function is imported from a different account and already has the correct permissions use `fromFunctionAttributes()` API with the `skipPermissions` flag.
```

### Using the suggested method `fromFunctionAttributes`
```typescript
    const func2 = lambda.Function.fromFunctionAttributes(this, 'Function', {
      functionArn: 'arn:aws:lambda:us-east-1:123456789012:function:my-function:1',
      sameEnvironment: true,
    });


    const lambdaRole = new iam.Role(this, 'LambdaRole', {
      assumedBy: new iam.ServicePrincipal('lambda.amazonaws.com'),
    });

    func2.grantInvoke(lambdaRole);
```

output
```
PolicyDocument:
        Statement:
          - Action: lambda:InvokeFunction
            Effect: Allow
            Resource:
              - arn:aws:lambda:us-east-1:123456789012:function:my-function:1
              - arn:aws:lambda:us-east-1:123456789012:function:my-function:1:*
        Version: "2012-10-17"
```

You can see the output here is actually wrong. I assume CDK doesn't expect you to passin the function arn with a version in `fromFunctionAttributes`. If we follow this design, we can also assume the Arn passed in shouldn't have a version.

packages/aws-cdk-lib/aws-lambda/lib/function-base.ts
roger-zhangg: Refer to the comment above, the previous design doesn't expect a `functionArn` with version

packages/@aws-cdk-testing/framework-integ/test/aws-lambda/test/integ.permissions.ts
```diff
@@ -17,6 +17,10 @@ fn.grantInvoke(new iam.AnyPrincipal().inOrganization('o-yyyyyyyyyy'));
 
 fn.grantInvoke(new iam.OrganizationPrincipal('o-xxxxxxxxxx'));
 
+fn.grantInvoke(new iam.AnyPrincipal().inOrganization('o-yyyyyyyyyy2'), { onlyGrantLatestVersion: true });
```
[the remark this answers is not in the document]
roger-zhangg: Hi @godwingrs22 I'm looking into this but could you give me a headup on how to write integ test on this? I think I need to assume to the role I created here. But I didn't find how can I assume role in integration test. I checked `invokeFucntion` [here](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/private/deploy-assert.ts#L97) and `invokefunctionprop` [here](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/integ-tests-alpha/lib/assertions/sdk.ts#L196) but nothing related to assume role is found.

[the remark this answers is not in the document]
roger-zhangg: I had a inspiration on this from my teammate: I can setup 2 lambda function f1,f2 and a role in the integ-test, f1 grants invoke to role, then f2 uses this role to try to invoke f1.  What do you think?

roger-zhangg:
Hi @nmussy , @godwingrs22 I'm really grateful for your great and helpful ideas, I have updated and replied to some of the conversation above. Please help to review so I can process and merge it sooner. Thanks!

packages/aws-cdk-lib/aws-lambda/lib/function-base.ts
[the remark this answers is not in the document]
roger-zhangg: I'm also confused on this part, this interface uses `identity` as param, however the implementation uses `grantee`. I'm not sure if we should just keep it as is or should we change one to match the other?

roger-zhangg:
Hi @godwingrs22 , could you please help to take a look when available, Thanks!

roger-zhangg:
😴

godwingrs22:
Hi @roger-zhangg , Apologies for the delay. I'll review within this week.

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -1309,6 +1309,72 @@ describe('function', () => {
...
+      Template.fromStack(stack).hasResourceProperties('AWS::IAM::Policy', {
+        PolicyDocument: {
+          Version: '2012-10-17',
+          Statement: [
+            {
+              Action: 'lambda:InvokeFunction',
+              Effect: 'Allow',
+              Resource: [
+                { 'Fn::GetAtt': ['Function76856677', 'Arn'] },
+                { 'Fn::Join': ['', [{ 'Fn::GetAtt': ['Function76856677', 'Arn'] }, ':$LATEST']] },
```
godwingrs22, line 1337: We would be able to call the latest lambda by also specfying the version right? In that case, can we assert for the role contains specific latest version also along with $LATEST?

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -201,6 +201,27 @@ You can also restrict permissions given to AWS services by providing
...
+
+```ts
+declare const fn: lambda.Function;
+const principal = new iam.ServicePrincipal('my-service');
+// Grant invoke only to latest version and unqualified lambda arn
+fn.grantInvokeLatestVersion(principal);
+
+```
+
+If you want to grant access for invoking a specific version of Lambda function, you can use `fn.grantInovkeVersion(grantee, version)`
```
godwingrs22, line 215: nit: typo in `grantInvokeVersion`

godwingrs22 (changes requested):
@roger-zhangg Apologies for the delay in response. Thank you for making the changes. I've few followup comments left.

packages/@aws-cdk-testing/framework-integ/test/aws-lambda/test/integ.permissions.ts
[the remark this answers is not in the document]
godwingrs22: Yes this should work. However let me get a second review within my team and get back to you on this integ test handling.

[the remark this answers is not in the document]
godwingrs22: @roger-zhangg , Seems the current integ test framework is designed for handling success cases. Hence we can write a success case and perform assertion if the right permission is set. This should be good enough to cover the integ test.

roger-zhangg:
Will resolve comments this week

roger-zhangg (commented):
Addressed comments

roger-zhangg:
Hi @godwingrs22 , could you help with a final review? Thanks

godwingrs22 (approved):
LGTM. Thanks @roger-zhangg for your effort in implementing this feature.

PR #29860, opened 2024-04-17
chore: add `aws-sdk` as dev dependency for unit tests and satisfy `AwsSdkCall` in unit tests

colifran:
### Reason for this change

https://github.com/aws/aws-cdk/pull/29648 introduced a change to the `AwsSdkCall` representation used in the v2 and v3 handler code. Our handler unit tests use `satisfies` to validate that the event object satisfies `AwsSdkCall`. All unit tests and the build still pass, but the linter calls out that the event object doesn't actually satisfy `AwsSdkCall`.

https://github.com/aws/aws-cdk/pull/29845 removed the dependency `@aws-cdk/custom-resource-handlers` had on `aws-sdk`. We should add this as devDependency since we're using `aws-sdk` in v2 handler mocks.

### Description of changes

I added `logApiResponseData` property to the event objects being tested to make the event satisfy `AwsSdkCall`. I added `aws-sdk` as a dev dependency. We will remove this as part of the v2 handler removal.

### Checklist

----


piradeepk (approved):
Lgtm! 🚀

PR #28901, opened 2024-01-28
feat(ec2): support network interface L2

WinterYukky:
# What change?
Add `ec2.NetworkInterface` L2 constructs and `ec2.Instance.addNetworkInterface()` method. 

# Why need this change? 
There are three reasons. 

1. Currently (v2.124.0), `ec2.Instance` doesn't support additional IPv4 address. It is hard in the cace of advanced network architects (e.g. [Multi-VPC ENI Attacements](https://aws.amazon.com/about-aws/whats-new/2023/10/multi-vpc-eni-attachments/?nc1=h_ls)). To support network interface L2 is very usefull. 
1. L2 interface (INetworkInterface) is neccesarry to support other L2 resources that using network interface such as VerifiedAccess Endpoint. 
1. A created network interface by `ec2.Instance` can not get ID of network interface. Customers who need a network interface ID wants configure network interface at initialize `ec2.Instance`. 

# How to use this?

 Following document is README.md. 

## Network Interface

You can attach additional network interfaces to an EC2 instance.  Attaching multiple network interfaces to an instance is useful when you want to:
- Create a management network.
- Use network and security appliances in your Virtual Private Cloud (VPC).
- Create dual-homed instances with workloads/roles on distinct subnets.
- Create a low-budget, high-availability solution.

The following code how to add additional network interfaces for an EC2 instance.

```ts
declare const vpc: ec2.IVpc;
declare const instance: ec2.Instance;

const eni1 = new ec2.NetworkInterface(this, 'NetworkInterface', {
  vpc,
});
// device index is ordered by default
instance.addNetworkInterface(eni1);

// or, you can also specify the device index
declare const eni2: ec2.INetworkInterface;
instance.addNetworkInterface(eni2, {
  deviceIndex: 3,
});
```

You can also assign private IPv4 address from prefixes or specific IPv4 addresses.
The following code assigns IPv4 address from prefixes.

```ts
declare const vpc: ec2.IVpc;

new ec2.NetworkInterface(this, 'NetworkInterface', {
  vpc,
  ipv4: ec2.Ipv4Assign.fromPrefixes('10.0.0.0/28', '10.1.0.0/28')
    .addPrimaryAddress('10.0.0.10'),
});
```

For more information see [Scenarios for network interfaces](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/scenarios-enis.html).

# Design decision-making
[...]

PR #29875, opened 2024-04-17
feat(ec2): network interface definitions for launch templates

### LINKED ISSUE #14494: EC2 Launch Template: Vpc Subnets support
Would be great to be able to specify network interfaces for a [LaunchTemplate](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.LaunchTemplate.html). 

### Use Case

We are building an application where EC2 instances need to be launched/autoscaled in the application logic and the best abstraction for this would be to have a launch template that contains all the necessary details for launching an instance, without having to worry about things like subnets.

### Proposed Solution

Include the [vpcSubnets](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.Instance.html#vpcsubnets) fields in LaunchTemplate, similarly to what you can do with an Instance.

### Other

I'm not super deep into CDK/CloudFormation/EC2 details, so my proposed change may have the wrong abstractions. Still, being able to specify the networking configuration in a LaunchTemplate would be awesome. Just for reference, I've done the same thing earlierly with Terraform, which does support Network Interfaces in launch tempalates, as described in [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template).

Thanks for your great work!

---

pasiorovuo:
This PR implements basic functionality for defining network interfaces as part of Launch Template.

### Issue # (if applicable)

Fixes #14494

### Reason for this change

Current Launch Template L2 is missing the ability to define multiple interfaces, their subnets etc. This functionality is required in advanced scenarios where instances need multiple public IP addresses or need to be multi-homed etc.

### Description of changes

Multiple new types have been introduced in `packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts`. Many of them are just renamed `extends` of the `ec2-generated` types. This made sense as the Cloudformation type definition is suitable, but I'm a bit uncertain if this violates the design requirement _Do not “leak” the details or types of the CFN layer when defining your construct API_.

Main type is `NetworkInterface` which defines the attributes for the interfaces. Additional changes are in `LaunchTemplate` constructor to accommodate the new functionality.

README has been updated to include a simple usage example.

A single test has been added. I have also verified the implementation by creating stacks in different configurations and deployed them. There too many combinations to check them all, so only a subset has been tested, unfortunately.

### Checklist

----


pasiorovuo:
I do not know what the failing check is about.

packages/aws-cdk-lib/aws-ec2/README.md
```diff
@@ -2212,6 +2212,35 @@ const launchTemplate = new ec2.LaunchTemplate(this, 'LaunchTemplate', {
...
+declare const vpc: ec2.Vpc;
+
+const sg1 = new ec2.SecurityGroup(this, 'sg1', {
+  vpc: vpc,
+});
+const sg2 = new ec2.SecurityGroup(this, 'sg2', {
+  vpc: vpc,
+});
+
+new LaunchTemplate(stack, 'LaunchTemplate', {
```
nmussy, line 2227: To fix the failing build:
```suggestion
new ec2.LaunchTemplate(stack, 'LaunchTemplate', {
```

packages/aws-cdk-lib/aws-ec2/README.md
```diff
@@ -2212,6 +2212,35 @@ const launchTemplate = new ec2.LaunchTemplate(this, 'LaunchTemplate', {
...
+declare const vpc: ec2.Vpc;
+
+const sg1 = new ec2.SecurityGroup(this, 'sg1', {
+  vpc: vpc,
+});
+const sg2 = new ec2.SecurityGroup(this, 'sg2', {
+  vpc: vpc,
+});
+
+new ec2.LaunchTemplate(stack, 'LaunchTemplate', {
```
nmussy, line 2227: Should be good this time 😅 
```suggestion
new ec2.LaunchTemplate(this, 'LaunchTemplate', {
```

pasiorovuo:
This seems to be overlapping in parts with https://github.com/aws/aws-cdk/pull/28901. If it would be accepted, I could update this PR to account for the new types etc. introduced by it.

packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts
```diff
@@ -763,7 +1019,7 @@ export class LaunchTemplate extends Resource implements ILaunchTemplate, iam.IGr
         tagSpecifications: tagsToken,
         userData: userDataToken,
         metadataOptions: this.renderMetadataOptions(props),
-        networkInterfaces,
+        networkInterfaces: networkInterfaces ? synthesizeNetworkInterfaces(networkInterfaces) : undefined,
```
chenhe95, line 1022: I think that if `networkInterfaces` is defined, we should not be providing anything in `securityGroupIds`. Since the options provided in `securityGroupsToken` are not used at all unless `props.networkInterfaces` is undefined and `props.associatePublicIpAddress` is defined   
  
On line `securityGroupIds: networkInterfaces ? undefined : securityGroupsToken,` right above

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.launch-template.js.snapshot/aws-cdk-ec2-lt-metadata-1.template.json
chenhe95, line 1: Would it be possible to also have unit tests in https://github.com/aws/aws-cdk/blob/7b2a65384a1592410cd820b745c680fea00d8c95/packages/aws-cdk-lib/aws-ec2/test/launch-template.test.ts#L4 that have cases for when different network interface options are provided and also a verification that when `props.associatePublicIpAddress` is incompatible with `props.networkInterfaces`, the correct error is thrown?

packages/aws-cdk-lib/aws-ec2/lib/launch-template.ts
```diff
@@ -68,6 +69,234 @@ export enum InstanceInitiatedShutdownBehavior {
...
+   * The primary private IPv4 address of the network interface.
+   *
+   * @default A random IP address is assigned from one of the subnets.
+   */
+  readonly privateIpAddress?: string;
+
+  /**
+   * One or more private IPv4 addresses.
+   *
+   * @default A random IP address is assigned from one of the subnets.
```
chenhe95, line 263: I was never able to figure out how these defaults are resolved since the AWS documentation doesn't have any info  
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-privateipadd.html  
```
PrivateIpAddresses

    One or more private IPv4 addresses.

    Required: No

    Type: Array of [PrivateIpAdd](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-privateipadd.html)

    Update requires: [No interruption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-no-interrupt)
```

chenhe95:
Hi Pasi. Thanks for looking into this! I actually had the exact same issue as the one you're trying to solve. Your PR looks very similar to my local prototype with the differences being in how security group IDs are handled and how the unit tests are done. I was also not sure how AWS determines default behaviors when input fields are left undefined. Left comments on those parts. 
  
Nice work!

chenhe95:
Hi Pasi! I am wondering if this ticket is still being worked on, since I am very invested in getting launch templates to be more configurable  
Thank you for all the work so far!

GavinZZ (changes requested):
I want to start off by apologizing that we did not review this PR sooner.

Thank you for introducing this new L2. I appreciate the effort but I see a lot of the properties were simply copy/paste of the L1 construct. I would love to use this chance to see if there's any improvements we can make to make L2 more usable but flattening the properties/providing better default/etc. 

Feel free to find me in CDK slack workspace and we can have a deep sync on this.

GavinZZ:
Apologize in advance if this is something you're actively working on. However, this PR looks stale to me and I'm going to close this PR for staleness as I see that there's a new PR for this same feature request.

PR #29893, opened 2024-04-18
docs: elaborate that crossZoneEnabled property cannot be changed for ALB in ELBv2.

### LINKED ISSUE #29866: aws-elb: Disabling cross zone load balancing is not respected in Java CDK
### Describe the bug

```typescript
        ApplicationLoadBalancer loadBalancer = ApplicationLoadBalancer.Builder.create(this, "ApplicationLoadBalancer")
                .vpc(vpc)
                .crossZoneEnabled(false) // BUG: This is not respected and cross zone is enabled
                .desyncMitigationMode(DesyncMitigationMode.DEFENSIVE)
                .internetFacing(true)
                .build();
```

### Expected Behavior

I expect cross zone to be disabled or a warning why it was enabled

### Current Behavior

Cross zone load balancer is not respected

### Reproduction Steps

See source

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.134.0

### Framework Version

2

### Node.js Version

v21.7.1

### OS

MacOS

### Language

Java

### Language Version

Java (17)

### Other information

_No response_

ashishdhingra:
### Issue #29866

Closes #29866.

### Reason for this change
Per below documentation:
- [Cross-zone load balancing for target groups](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/disable-cross-zone.html), it mentions that `With Application Load Balancers, cross-zone load balancing is always turned on at the load balancer level, and cannot be turned off.`. It also specifies, `For target groups, the default is to use the load balancer setting, but you can override the default by explicitly turning cross-zone load balancing off at the target group level.`
- [AWS::ElasticLoadBalancingV2::LoadBalancer LoadBalancerAttribute](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattribute.html) also specifies for `load_balancing.cross_zone.enabled` that `Indicates whether cross-zone load balancing is enabled. The possible values are true and false. The default for Network Load Balancers and Gateway Load Balancers is false. The default for Application Load Balancers is true, and cannot be changed.`.

So for Application Load Balancer, cross-zone load balancing is always turned on at the load balancer level, and cannot be turned off.

### Description of changes
Elaborate API documentation for ELBv2 that `crossZoneEnabled` property cannot be changed for `ALB` in ELBv2.

N/A

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts
+++ b/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts
@@ -51,7 +51,7 @@ export interface BaseLoadBalancerProps {
   /**
    * Indicates whether cross-zone load balancing is enabled.
    *
-   * @default - false for Network Load Balancers and true for Application Load Balancers.
+   * @default - false for Network Load Balancers and true for Application Load Balancers (this cannot be changed).
    */
   readonly crossZoneEnabled?: boolean;
 


pahud:
I would suggest we fix here as well:

https://github.com/aws/aws-cdk/blob/c2766ecfbd0f871cf7d9c84ca78d3a7fdec63194/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts#L246

Because the L2 `crossZoneEnabled` prop literally allows you to specify as `false` but not well handled here. For example, if user set `crossZoneEnabled: false` in the L2 props, this attribute will always be undefined, which is exactly the same behavior leaving the L2 prop undefined. This is weird to me. What do you think?

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts
```diff
@@ -51,7 +51,7 @@ export interface BaseLoadBalancerProps {
   /**
    * Indicates whether cross-zone load balancing is enabled.
    *
-   * @default - false for Network Load Balancers and true for Application Load Balancers.
+   * @default - false for Network Load Balancers and true for Application Load Balancers (this cannot be changed).
```
aaythapa, line 54: Need some clarification, I see [here](https://github.com/aws/aws-cdk/issues/29866#issuecomment-2064577629) and [here](https://github.com/aws/aws-cdk/issues/29866#issuecomment-2061508349) that there are way to disable crossZone for ALB. Why are we saying this cannot be changed here? If we can change it then this should be implement the change suggested [here](https://github.com/aws/aws-cdk/issues/29866#issuecomment-2065180727) imo. Let me know if I'm understanding this wrong.

ashishdhingra: @aaythapa In https://github.com/aws/aws-cdk/issues/29866#issuecomment-2064577629 and   https://github.com/aws/aws-cdk/issues/29866#issuecomment-2061508349, it is mentioned that the `crossZone` could be disabled at `Target Group` level per docs and it already works per example. However, it cannot be changed at **ALB level** per docs. This PR is only to make it clear in LoadBalancer docs that it cannot be changed at ALB level.

**NOTE:** I'm unsure what will happen in CloudFormation deployment if we enforce `crossZone` to `false`, when service itself doesn't support disabling it at ALB level.

ashishdhingra:
Closing this PR in favor of https://github.com/aws/aws-cdk/pull/29907.

PR #29907, opened 2024-04-19
fix(elasticloadbalancingv2): crossZoneEnabled does not support false for ALB

pahud:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29866

### Reason for this change

`crossZoneEnabled` is not well handled when it's `false` with ALB.  Because:

1. When the L2 prop is set `false`, it will not pass down to the L1 and won't throw any error as ALB does not support being disabled. It just silently ignore it.
2. When the prop is `false` for NLB, the L1 attribute will be `undefined`, which is having the same result but it should be explicitly set as `false` in L1.

This PR covers the following cases:

1. When `crossZoneEnabled` is `true`, `load_balancing.cross_zone.enabled` should be `true`.
2. When `crossZoneEnabled` is `false`, `load_balancing.cross_zone.enabled` should be `false`, rather than `undefined`.
3. When `crossZoneEnabled` is `false` with ALB, cdk throws an error because ALB does not support disabling it per [doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattribute.html) description.
4. NLB supports either `true` or `false`.
5. This prop can be `undefined` for ALB or NLB.
6. Improve the doc string for the `crossZoneEnabled` prop.

### Checklist

----


GavinZZ (approved):
LGTM

PR #29913, opened 2024-04-20
fix(stepfunctions): distributed maps under branches do not have necessary permissions

### LINKED ISSUE #29266: stepfunctions: DistributedMap in Parallel doesn't create DistributedMapPolicy
### Describe the bug

When a DistributedMap is a top level item in a state machine it, 'synth ' will generate a DistributedMap Policy 
image

When you run the DistributedMap in a Parallel State it will not generate that policy and the step function will not be able to start the distributed step.

### Expected Behavior

Whether in Parallel or not the DistributedMap should be generated. 

### Current Behavior

https://github.com/sannies/legendary-octo-computing-machine

### Reproduction Steps

See simple python project to reproduce

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.130.0

### Framework Version

_No response_

### Node.js Version

21.1.0

### OS

Mac

### Language

Python

### Language Version

_No response_

### Other information

_No response_

mschelstastic:
### Issue #
Closes #29266

### Reason for this change

distributed maps under branch states (i.e., Parallel) do not apply the necessary permissions to run the state.

### Description of changes

this moves the bind functionality into state and calls it on both state and all child (branch) states. Previously it did not run on branch states and as such did not work for example a distributed map under a parallel state without adding the permissions out of band.

An alternative would be to have a bind method that the user calls out of band on the distributed map (or in my case, in a wrapper class, but it's clunky)

Unit tests

### Checklist

----


curquhart:
ahh re pull request validator failing for no integration test update, i will look at that later this weekend of sometime next week.

curquhart:
added a couple integration tests and fixed the linting problems

curquhart:
i was wondering about my approach... because branches aren't exposed, i moved the policy stuff into State. however, State cannot depend on DistributedMap because of a cyclic dependency. To deal with this, I just passed isDistributedMap into the method call. This is based on the assumption that branches is hidden for a reason.

This works but could be cleaner. I was wondering about thoughts on a few alternative solutions:
- could i expose branches in State? then this stuff could be entirely (other than that visibility change in State) be in StateGraph
- add isDistributedMap to State (as in, StateGraph would call state.isDistributedState())... in the base class, this would return true, in DistributedMap this would return false
- processorMode is already in State... it's protected but if we made it public we might be also use that to determine if distributed map (by swapping with a getter, throwing the real value in _processorMode, and then having the getter traverse any children)

from my point of view, anything is fine, i just want to remove the manual perms from my cdk since it's weird that cdk sometimes but not always adds them (depending on how the graph is shaped) heh

anyway will update if needed!

curquhart:
👋 just wanted to touch base on this... I'm happy to make changes if yall don't like the approach!

thanks :)

curquhart:
just saw there is also an issue reported for this - linked it in desc... just following up again though, is someone able to look at this? Thanks :)

comcalvi (dismissed):
`branches` is protected because many of the stepfunctions features live on the base class but are defined and exposed in subclasses, as explained here: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-stepfunctions/lib/states/state.ts#L163, and since `branches` is only usable with `Parallel` states, I don't think it makes sense to expose it. 

I don't think the processor mode approach is the right one either, for the the reason you outlined. 

The design suggests we should include `IsDistributedMap` on the base class, as you describe in option 2. This is an odd placement, because the base class should not know anything about its Distributed subclass variant - but we've already broken this assumption by declaring all of these subclass-unique properties on the base class, so we may as well include `isDistributedMap` on the base class.

curquhart:
sorry finally got back to this... I realized adding isDistributedMap was impractical because of `DistributedMap. isDistributedMap` being public which would make making it an instance method a breaking change to the public interface and dug a bit more through the innards

i realized that child graphs are contained within allContainedStates which means we can just recursively call bind on each graph in that collection, which in turn avoids making changes outside of the original StateGraph logic (so less code touched, woohoo).

anyway! updated to do that, let me know if this makes sense

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions/test/integ.map-distributed-iam.ts
```diff
@@ -0,0 +1,59 @@
+import { IntegTest } from '@aws-cdk/integ-tests-alpha';
+import * as cdk from 'aws-cdk-lib';
+import * as sfn from 'aws-cdk-lib/aws-stepfunctions';
+
+/**
+ * Stack verification steps:
+ *
+ * -- aws stepfunctions describe-state-machine --state-machine-arn <stack-output> has a status of `ACTIVE`
```
TheRealAmazonKendra, line 8: Could you please add an assertion that does this instead of making it a manual verification step?

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions/test/integ.map-distributed-iam.ts
```diff
@@ -0,0 +1,59 @@
...
+    executionType: sfn.ProcessorType.STANDARD,
+  });
+}
+
+const sm1 = new sfn.StateMachine(stack, 'StateMachine1', {
+  definitionBody: sfn.DefinitionBody.fromChainable(createMap()),
+  timeout: cdk.Duration.seconds(30),
+});
+
+new cdk.CfnOutput(stack, 'StateMachine1ARN', {
```
TheRealAmazonKendra, line 36: These outputs are great but we should do something with them. Perhaps add assertions that get the machine state of each of these?

packages/aws-cdk-lib/aws-stepfunctions/test/state-graph.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    });
+    const stateMachineLogicalId = stack.getLogicalId(stateMachine.node.defaultChild as stepfunctions.CfnStateMachine);
+    const template = assertions.Template.fromStack(stack);
+
+    // THEN
+    // might as well check policy since this is around distributed map perms; however, the actual thing under test here
+    // is that our permissions adding doesn't create infinite recursion as we iterate through all nested graphs.
+    // a failure in this test case (for the thing it is meant to test) would present as a stack overflow ("maximum call
+    // stack size exceeded" error.)
+    template.hasResource('AWS::IAM::Policy', createPolicyProps(stateMachineLogicalId));
```
TheRealAmazonKendra, line 64: Can we check for the expected amount of policies based off the amount of child states?

TheRealAmazonKendra (changes requested):
Thank you for your contribution!  I can see that a lot of hard work went into this and it looks right. I just have a couple requests inline for improving the testing a bit.

curquhart:
oh i just saw your comments! thanks for looking :D I'll try to address them this weekend - i looked at a couple integ tests and the (manual) way i did it was based on an existing test, i didn't realize there's a framework in other tests for validating via api, that'll be much nicer :)

PR #29928, opened 2024-04-22
feat(kms): key rotation period

### LINKED ISSUE #29927: kms: support for key rotation period
### Describe the feature

Configuring automatic key rotation period.

```ts
new kms.Key(this, 'MyKey', {
  enableKeyRotation: true,
  rotationPeriod: Duration.days(180), // Add
});
```

### Use Case

Cloudformation [supports for configuring period of automatic key rotation](https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-resource-kms-key.html#cfn-kms-key-rotationperiodindays) but CDK does not.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.127.0

irrelevant

badmintoncryer:
### Issue # (if applicable)

Closes #29927.

### Reason for this change

Cloudformation [supports for configuring period of automatic key rotation](https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-resource-kms-key.html#cfn-kms-key-rotationperiodindays) but CDK does not.

### Description of changes

Added `rotationPeriod` to `KeyProps`.

### Checklist

----


packages/aws-cdk-lib/aws-kms/lib/key.ts
```diff
@@ -434,6 +434,13 @@ export interface KeyProps {
    */
   readonly enableKeyRotation?: boolean;
 
+  /**
+   * The period between each automatic rotation.
+   *
+   * @default - 365 days.
+   */
+  readonly rotationPeriod?: Duration;
```
badmintoncryer, line 442: I've simply added `rotationPeriod` properties to `KeyProps`.

However, there are other possible implementation approaches, and I'm struggling to decide which one is best. Could I please get your opinions, reviewers?


1. add rotationPeriod property (current implementation)

```ts
new kms.Key(this, 'MyKey', {
  enableKeyRotation: true,
  rotationPeriod: Duration.days(180), // Add
});
```
3. deprecate `enableKeyRotation` and define `KeyRotation` property
```ts
export interface KeyRotation {
  enableKeyRotation: boolean,
  rotationPeriod?: Duration,
}

new kms.Key(this, 'MyKey', {
  keyRotation: {
    enableKeyRotation: true,
    rotationPeriod: Duration.days(180), // optional
  },
});
```

5. deprecate `enableKeyRotation` and add only `rotationPeriod`

```ts
new kms.Key(this, 'MyKey', {
  rotationPeriod: Duration.days(180), // Implicitly enable key rotation by defining this property
});
```

nmussy: I don't think the current implementation is that terrible. It probably wouldn't have been my first choice if `enableKeyRotation` wasn't already there, but given the circumstances, you're not introducing a deprecation or breaking change, you're allowing the existing construct to customize its default value.

badmintoncryer: @nmussy Thank you for your opinion. I'm relieved to hear that.

comcalvi: I like the implementation you chose @badmintoncryer, just with a small modification; if `rotationPeriod` is defined and `enableKeyRotation` is `undefined`, let's set `enableKeyRotation` to `true`.

packages/aws-cdk-lib/aws-kms/lib/key.ts
```diff
@@ -722,6 +729,15 @@ export class Key extends KeyBase {
       throw new Error('key rotation cannot be enabled on asymmetric keys');
     }
 
+    if (props.rotationPeriod) {
+      if (!props.enableKeyRotation) {
```
comcalvi, line 733:
```suggestion
      if (props.enableKeyRotation === false) {
```

packages/aws-cdk-lib/aws-kms/lib/key.ts
```diff
@@ -434,6 +434,13 @@ export interface KeyProps {
    */
   readonly enableKeyRotation?: boolean;
 
+  /**
+   * The period between each automatic rotation.
+   *
+   * @default - 365 days.
```
comcalvi, line 440:
```suggestion
   * @default - set by CFN to 365 days.
```

This clarifies to future contributors that the value is intentionally not set to 365 by the CDK, and instead defers to CFN for that value.

comcalvi (changes requested):

badmintoncryer:
@comcalvi 
Thank you for your review! I've updated my implementation. Could you please confirm it again?

packages/aws-cdk-lib/aws-kms/test/key.test.ts
```diff
@@ -518,12 +520,30 @@ test('key with some options', () => {
   });
 });
 
+test('set rotationPeriod without enabling enableKeyRotation', () => {
```
comcalvi, line 523: Can we also add a test with `enableKeyRotation` explicitly set to true? Other than that this looks ready to ship.

comcalvi (changes requested):

packages/aws-cdk-lib/aws-kms/test/key.test.ts
comcalvi: you are correct, it does.

comcalvi (approved):

Issue #29931, opened 2024-04-23
ec2: change default  Volume.volumeType and/or change generic EbsDeviceVolumeType


azatoth:
### Describe the feature

Change `Volume.volumeType` default to be `EbsDeviceVolumeType.GENERAL_PURPOSE_SSD_GP3`, and rename `EbsDeviceVolumeType.GENERAL_PURPOSE_SSD` to `EbsDeviceVolumeType.GENERAL_PURPOSE_SSD_GP2`

### Use Case

People incur extra cost because they might not know that they should change the volume type from the default GP2 which is discouraged to use.

### Other Information

Currently `Volume.volumeType` defaults to `EbsDeviceVolumeType.GENERAL_PURPOSE_SSD` which refers to GP2; I think it might be time to switch the default to GP3; The naming of `EbsDeviceVolumeType.GENERAL_PURPOSE_SSD` for GP2 is unfortunate and should probably be deprecated for a `EbsDeviceVolumeType.GENERAL_PURPOSE_SSD_GP2` instead.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [X] This feature might incur a breaking change

### CDK version used

v2.138

### Environment details (OS name and version, etc.)

Ubuntu 22.04.4

pahud:
Love this idea but I guess we can't switch the [EbsDeviceVolumeType](https://github.com/aws/aws-cdk/blob/31492c621c835e3fa08de2bbac24a0adc6fc2d69/packages/aws-cdk-lib/aws-ec2/lib/volume.ts#L174) enum using feature flag if we plan to make `EbsDeviceVolumeType.GENERAL_PURPOSE_SSD` point to `EbsDeviceVolumeType.GENERAL_PURPOSE_SSD_GP3`.  But we can change the default volume type for EBS [here](https://github.com/aws/aws-cdk/blob/31492c621c835e3fa08de2bbac24a0adc6fc2d69/packages/aws-cdk-lib/aws-ec2/lib/volume.ts#L624) if volumeType is undefined.

i.e.

```ts
volumeType: props.volumeType ?? EbsDeviceVolumeType.GENERAL_PURPOSE_SSD_GP3,
```

Looking at the EBS console, the default volume type for a new ebs is now gp3 so I think it makes sense here.

<img width="918" alt="image" src="https://github.com/aws/aws-cdk/assets/278432/137a113a-5d57-436f-aca0-017acfda7776">

PR #29934, opened 2024-04-23
chore(ebs): set default volumeType to gp3(under feature flag)

pahud:
### Issue # (if applicable)

As the EBS console is now having `gp3` as the default volumeType, this PR set the default volumeType to gp3 if undefined under feature flag.

Closes https://github.com/aws/aws-cdk/issues/29931

I have deployed the sample below and verified the volume type is `gp3` from console.

```ts
import { Stack, App, Size, aws_ec2 as ec2 } from 'aws-cdk-lib';
import * as cxapi from 'aws-cdk-lib/cx-api';

const app = new App();

const stack = new Stack(app, 'demo-stack');

stack.node.setContext(cxapi.EBS_DEFAULT_GP3, true);

// should create a gp3 volume
new ec2.Volume(stack, 'Volume', {
  availabilityZone: 'us-east-1a',
  size: Size.gibibytes(500),
});


```

### Checklist

----


packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
```diff
@@ -128,7 +127,8 @@ The following json shows the current recommended set of flags, as `cdk init` wou
     "@aws-cdk/aws-codepipeline:crossAccountKeysDefaultValueToFalse": true,
     "@aws-cdk/aws-codepipeline:defaultPipelineTypeToV2": true,
     "@aws-cdk/aws-kms:reduceCrossAccountRegionPolicyScope": true,
-    "@aws-cdk/aws-eks:nodegroupNameAttribute": true
+    "@aws-cdk/aws-eks:nodegroupNameAttribute": true,
+    "@aws-cdk/aws-ec2:ebsDefaultGp3Volume": true
```
nmussy, line 131: Any reason not to have something like `"@aws-cdk/aws-ec2:ebsDefaultVolume": "gp3"`? The current setup makes future changes in the default volume type a little more awkward, we'll have to have multiple `aws-cdk/aws-ec2:ebsDefaultXXXVolume` flags, make sure there are no conflicts, etc.

pahud: Let's say if we introduce `"@aws-cdk/aws-ec2:ebsDefaultVolume": "gp3"` in this PR and after a while AWS announces `gp4` as a new default and we modify this feature flag to `"@aws-cdk/aws-ec2:ebsDefaultVolume": "gp4"`, this would cause breaking changes for those opting in `gp3` and we still need a check to see if they have explicitly enable this flag as `gp3`. I am not sure if that would be easier to maintain. Another consideration is conventionally our feature flags are boolean except for [target partitions](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md#aws-cdkcoretarget-partitions). Will check this with the maintainers.

nmussy: > we modify this feature flag to "@aws-cdk/aws-ec2:ebsDefaultVolume": "gp4", this would cause breaking changes for those opting in gp3 and we still need a check to see if they have explicitly enable this flag as gp3

Wouldn't we only change the recommended value to `gp4`? Users that opted in to `gp3` would be in the same situation as those who enabled `ebsDefaultGp3Volume` if we introduce `ebsDefaultGp4Volume`. Arguably even worse off, since they would both need to disable `ebsDefaultGp3Volume` and enable `ebsDefaultGp4Volume`, vs just updating `ebsDefaultVolume` to `gp4`.

I'm leaving it up to you guys, I might not be seeing the whole picture, just throwing it out there 👍

TheRealAmazonKendra:
If we make this switch without a feature flag, what is the behavior of this change? Obviously there will be a replacement of instances but will this cause downtime? If we can just update it, no reason not to.

pahud:
@TheRealAmazonKendra  Let me check and update here.

packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
TheRealAmazonKendra: One thing to note here is that feature flags are handled as a boolean so we can't use `gp3`.

TheRealAmazonKendra: Oh, Pahud already commented that. I missed that on the initial read.

TheRealAmazonKendra:
I was looking around to see what the EOL date is for these instance types (there is none) and found this https://aws.amazon.com/blogs/storage/migrate-your-amazon-ebs-volumes-from-gp2-to-gp3-and-save-up-to-20-on-costs/. It seems that there isn't actually downtime when a replacement happens but that there may be other considerations we need to take into account if someone is migrating.

TheRealAmazonKendra:
Also relevant https://docs.aws.amazon.com/ebs/latest/userguide/modify-volume-requirements.html

TheRealAmazonKendra:
All that to say, I totally agree that we need to add a feature flag here but we also probably need to add a bit more than just that. Our documentation should make these limitations clear. I'm not sure if there's much else we can do to assist in a smooth upgrade.

From the fact that gp3 was released in 2020 and we haven't had any new ones since then, I'm a bit torn about how much we need to future proof this. 

If we want to really future proof it, we could update the default here to be `LATEST` and  say in the documentation that we will use the latest volume type available at the time of implementation. To ensure we're not updating that version without the user intending to, we could save that version in context and require the user to clear that out in order to upgrade. I'm modeling this idea off of what we do with Amazon Linux AMIs.

This approach would still require a feature flag but it would ensure that we aren't having to update that feature flag and/or overwrite it in the future.

Thoughts on this approach? If it feels like overkill, I'm not opposed to how you've added the feature flag as is. Perhaps we should check with the EBS team to see if there's anything on the horizon for gp4 or similar and to see if there's anything thorny we need to know about updating via CloudFormation so we can warn users.

pahud:
>To ensure we're not updating that version without the user intending to, we could save that version in context and require the user to clear that out in order to upgrade. I'm modeling this idea off of what we do with Amazon Linux AMIs.

I'm interested to know more about this. Are you referring [latestamazonlinux2()](https://github.com/aws/aws-cdk/blob/466f170af409d0c9c44f0f03a6eb5a72553db29b/packages/aws-cdk-lib/aws-ec2/lib/machine-image/machine-image.ts#L38) which essentially lookup the value from ssm parameter through a context provider and cache in the context? Is there any existing context provider we can reuse for this use case? Adding a new context provider just for this seems a little bit overkill to me.

I am not a big fan of `LATEST` and I prefer explicit, clear and self-explanatory configuration without users dealing with the cache invalidation. For example, if I use `LATEST` for the default ebs volume type, I won't know what exactly volume type would be picked by CDK from my code and I'll need to check if there's something cached in my local context and double check my synthesized template. But if I explicitly set GP3  as the default with feature flag I have the confidence it would always be GP3 unless I explicitly set something else.

RE `gp4`, by the time `gp4` is announced as the default volume type and we are still at CDK v2, we could introduce `ebsDefaultGp4Volume` which has high priority over `ebsDefaultGp3Volume` if both exists and we could have a helper function to determine the default type if multiple feature flags are enabled. The advantage is that the user code would be very clear and self-explanatory without ambiguity. I am also fine with `ebsDefaultVolume: gp4` as mentioned above by @nmussy but that seems more like a context variable to switch the defaults rather than a feature flag to me.

GavinZZ (commented):
Let's also add the flag to https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/cx-api/README.md

pahud:
@GavinZZ  DONE

GavinZZ:
Read through all the comments and I actually like the idea proposed by @nmussy. Haven't thought about the implementation difficulty but my first instinct is that we don't need anything else to support string type value in feature flag. When `gp4` is out, we can switch the recommended value to `gp4` and stacks created before shouldn't get impacted while new stacks get default support using `gp4`.

GavinZZ (approved):
LGTM, discussed with the team and decided to continue using boolean value for feature flags.

Issue #29004, opened 2024-02-06
aws_s3_notifications : CDK destroy deletes existing S3 event notifications


nleut:
### Describe the bug

`cdk destroy` removes all event notifications configured on an existing S3 bucket instead of only CDK managed event notifications. This occurs whenever a stack that creates an event notification for an existing bucket is deleted or rolled back. 

### Expected Behavior

`cdk destroy` or a rollback from a failed stack should only delete S3 event notifications created by the stack

### Current Behavior

`cdk destroy` or a rollback from a failed stack removes all S3 event notifications on the bucket 

### Reproduction Steps

1. Create S3 bucket manually through the console
2. Create SNS topic manually through the console 
3. Add event notification on the S3 bucket to the SNS topic through the console
4. Create CDK stack that references an existing bucket and creates a new event notification:
```
bucket = s3.Bucket.from_bucket_arn(self, id="mybucket", bucket_arn="arn:aws:s3:::mybucket")
topic = sns.Topic(self, id="mytopic")
bucket.add_event_notification(s3.EventType.OBJECT_CREATED, s3_notifications.SnsDestination(topic), s3.NotificationKeyFilter(suffix=".1"))
```
5. `cdk deploy` the stack
6. `cdk destroy` the stack
7. Observe that all event notifications are now removed from the bucket, including the manually created event notification from step 3. 

### Possible Solution

The BucketNotificationHandler Lambda function described in #2004 appears to include handling for unmanaged event notifications. Modifying this function to support this scenario should resolve the issue. 

### CDK CLI Version

2.126.0 (build fb74c41)

### Node.js Version

v20.11.0

### OS

macOS Ventura 13.6.4

### Language

Python

### Language Version

Python (3.9.7)

spKavvadias:
We also had the same issue. We have multiple CDK stacks. When I tried to add or remove bucket notifications using the method you described (to a specific stack) it also removed all existing ones (not part of the stack).

pahud:
probably related to https://github.com/aws/aws-cdk/issues/28915

wchaws:
This is caused by https://github.com/aws/aws-cdk/commit/37be7b9b5d2abe04aa5f632a7b78aed3e9c233d8#r137776660

alpcaniscan:
> We also had the same issue. We have multiple CDK stacks. When I tried to add or remove bucket notifications using the method you described (to a specific stack) it also removed all existing ones (not part of the stack).

having same issue here.

PR #29939, opened 2024-04-23
fix(s3-notifications): cdk destroy deletes external/existing s3 notification events

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29004

### Reason for this change

`cdk destroy` removes all event notifications configured on an existing S3 bucket instead of only CDK managed event notifications. This occurs whenever a stack that creates an event notification for an existing bucket is deleted or rolled back.

### Description of changes

Add a `Delete` statement which will only remove the ones created from within the stack

Manually tested this.

### Checklist

----


packages/@aws-cdk/custom-resource-handlers/test/aws-s3/notifications-resource-handler/test_index.py
```diff
@@ -95,8 +95,8 @@ def test_create(self, _, put: MagicMock):
         index.handler(event, {})
 
         put.assert_called_once_with(
-            event["ResourceProperties"]["BucketName"],
-            event["ResourceProperties"]["NotificationConfiguration"],
+            Bucket=event["ResourceProperties"]["BucketName"],
```
GavinZZ, line 98: Need to update this test file because this [PR](https://github.com/aws/aws-cdk/commit/37be7b9b5d2abe04aa5f632a7b78aed3e9c233d8) only updated the handler file but didn't change or run the test files. This results in all these tests failing.

packages/@aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py
```diff
@@ -48,6 +48,10 @@ def with_id(n):
         old_incoming_ids = [n['Id'] for n in ids]
         # if the notification was created by us, we know what id to expect so we can filter by it.
         external_notifications[t] = [n for n in existing_notifications.get(t, []) if not n['Id'] in old_incoming_ids]
+    elif request_type == 'Delete':
+        # For 'Delete' request, old parameter is an empty dict so we cannot use this to determine which are external
+        # notifications. Fall back to rely on the stack naming logic.
+        external_notifications[t] = [n for n in existing_notifications.get(t, []) if not n['Id'].startswith(f"{stack_id}-")]
```
GavinZZ, line 54: Falling back to the previous correct solution https://github.com/aws/aws-cdk/commit/37be7b9b5d2abe04aa5f632a7b78aed3e9c233d8#diff-00d1f29e73955f487795d4a9873bcff824e70d6268e7926e44b7defb31b4ecb1L74

aaythapa:
Does this suggested fix not work? https://github.com/aws/aws-cdk/commit/37be7b9b5d2abe04aa5f632a7b78aed3e9c233d8#r137776660

GavinZZ:
> Does this suggested fix not work? [37be7b9#r137776660](https://github.com/aws/aws-cdk/commit/37be7b9b5d2abe04aa5f632a7b78aed3e9c233d8#r137776660)

No, I added comment in code to explain why I had to do it this way. Basically the `old` parameter for delete is empty dict unlike update.

aaythapa:
> > Does this suggested fix not work? [37be7b9#r137776660](https://github.com/aws/aws-cdk/commit/37be7b9b5d2abe04aa5f632a7b78aed3e9c233d8#r137776660)
> 
> No, I added comment in code to explain why I had to do it this way. Basically the `old` parameter for delete is empty dict unlike update.

ah, didn't notice. LGTM

PR #29958, opened 2024-04-25
fix(custom-resources): fails to use latest SDK version

### LINKED ISSUE #29891: (AwsCustomResource): fails to use latest SDK version
### Describe the bug

When setting `AwsCustomResource` with `installLatestAwsSdk: true` it fails to upgrade aws-sdk to latest version (at this moment `v3.556.0`) and fails back to the default (currently `v3.515.0` in Lambda).

Examples with DynamoDB:

```
 INFO	Installing latest AWS SDK v3: @aws-sdk/client-dynamodb
 Task timed out after 120.10 seconds
        .
        .
        .
 INFO	Installing latest AWS SDK v3: @aws-sdk/client-dynamodb
 INFO	Failed to install latest AWS SDK v3. Falling back to pre-installed version. Error: Error: Cannot find module 'mnemonist/lru-cache'
 Require stack:
 - /tmp/node_modules/@aws-sdk/endpoint-cache/dist-cjs/index.js
 - /tmp/node_modules/@aws-sdk/middleware-endpoint-discovery/dist-cjs/index.js
 - /tmp/node_modules/@aws-sdk/client-dynamodb/dist-cjs/index.js
 - /var/task/index.js
 - /var/runtime/index.mjs
          .
          .

```

 and SSM modules:

```
INFO	Installing latest AWS SDK v3: @aws-sdk/client-ssm
Task timed out after 120.16 seconds 
          .
          .
 INFO	Installing latest AWS SDK v3: @aws-sdk/client-ssm
 INFO	Failed to install latest AWS SDK v3. Falling back to pre-installed version. Error: Error: Cannot find module '@smithy/shared-ini-file-loader'
Require stack:
- /tmp/node_modules/@smithy/node-config-provider/dist-cjs/index.js
- /tmp/node_modules/@smithy/middleware-endpoint/dist-cjs/adaptors/getEndpointFromConfig.js
- /tmp/node_modules/@smithy/middleware-endpoint/dist-cjs/index.js
- /tmp/node_modules/@smithy/core/dist-cjs/index.js
- /tmp/node_modules/@aws-sdk/client-ssm/dist-cjs/index.js
- /var/task/index.js
- /var/runtime/index.mjs
          .
          . 

```
The custom resource eventually succeeds, after failing back to the default aws-sdk.

### Expected Behavior

Be able to update the Lambda Nodejs18 runtime with the latest SDKv3 version.

### Current Behavior

Unable to retrieve and upgrade Lambda Nodejs18 for SDKv3 latest version

### Reproduction Steps

Deploy a CDK App with a Custom Resource that has `installLatestAwsSdk: true`:

```
const myTable = new dynamodb.Table(this, 'myTable', {
      partitionKey: {
        name: 'id',
        type: dynamodb.AttributeType.STRING,
      },
    });

const myCustomResource = new cr.AwsCustomResource(this, 'myCR', {​​
      policy: cr.AwsCustomResourcePolicy.fromSdkCalls({​​
          resources: cr.AwsCustomResourcePolicy.ANY_RESOURCE,
      }​​),
      installLatestAwsSdk: true,
      onCreate: {​​
          service: 'DynamoDB',
          action: 'PutItem',
          parameters: {​​
              Item: {​​
                id: {"S": 'test-value'}
              }​​,
              TableName: myTable.tableName,
          }​​,
          physicalResourceId: cr.PhysicalResourceId.of('myCRphysicalResourceID'),
      }​​,
      onUpdate: {​​
          service: 'DynamoDB',
          action: 'PutItem',
          parameters: {​​
            service: 'DynamoDB',
            action: 'PutItem',
            parameters: {​​
                Item: {​​
                  id: {"S": 'test-value'}
                }​​,
                TableName: myTable.tableName,
            }​​,
          physicalResourceId: cr.PhysicalResourceId.of('myCRphysicalResourceID'),
      }​​,
      },
  }​​);

```

Check the underlying Lambda Function's logs to see the messages above. 

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.137.0

### Framework Version

_No response_

### Node.js Version

18

### OS

Mac

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

moelasmar:
### Issue # (if applicable)

Closes #29891

### Reason for this change
When setting AwsCustomResource with installLatestAwsSdk: true it fails to upgrade aws-sdk to latest version. The Lambda function created to support the custom resource usually time out after 120 seconds, and for some cases it get time out even after 900 seconds. 

### Description of changes
Update the Lambda function created for the custom resource to set its MemorySize to be 512 in case if installLatestAwsSdk flag is true instead of the default value 128.

Also, this change will expose the MemorySize to the AwsCustomResource construct, so customers can customize the MemorySize of the Lambda function to fulfill use cases that requires higher memory, CPU, or I/O performance.

I added the unit test cases, and updated the integration test cases.

### Checklist

----


TheRealAmazonKendra:
I'm not a fan of this change unless we really understand what's going on and why this suddenly started. It seems more like a bandaid than an actual fix.

We need to understand why this issue just started. We didn't see it prior to last week and, from what I know, there were no changes on our end. 

Did Lambda do something that is impacting this on their end? Or, is there actually a change we made that is related. Is npm experiencing outages and latency issues? Regardless, we shouldn't just make an update without understanding root cause as we are seeing this issue independent of the cdk version a customer is using.

@colifran was also doing some experiments around this issue on Friday and can confirm that was independent of cdk version.

TheRealAmazonKendra:
I also commented this is the issue but just to make sure the comment doesn't get lost:

I'm going to suspect there's something going on with the SDK and/or npm here. On https://www.npmjs.com/package/@aws-sdk/client-s3 it says the most recent version is from 9 days ago and lists 3.556.0 as the most up-to-date version. On https://github.com/aws/aws-sdk-js-v3 latest is 3.562.0 published 12 hours ago.

mrgrain:
This is interesting. I tend to agree with you @moelasmar that `128MB` memory is not a good choice for this function. Lambda Functions use memory as a proxy configuration for other resources. In particular, with higher memory the Function also receives increased CPU and I/O performance.

CPU performance is extremely relevant to cryptographic code, e.g. TLS connections. In the context of making AWS API requests, this is highly relevant. Not only is it using TLS to connect to APIs, it will also [sign requests for authentication](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html). You might find existing literature on this; from personal experience crypto performance starts to plateau around `768MB`. 

I/O performance will be relevant mostly for `npm install` because that is going to write a LOT of files to disk.

comcalvi (approved):

PR #29965, opened 2024-04-25
feat(rds): implement setting parameter group name

LaurenceWarne:
### Reason for this change

I'd like to be able to set RDS parameter group names using CDK.

### Description of changes

I've added a new field `name` to `ParameterGroupProps` used to populate `dbClusterParameterGroupName` and `dbParameterGroupName` where appropriate - AFAICS this couldn't be done at this level previously?

I've altered a couple of unit tests and an integration test.  However I'm unable to run the integration test as the postgres version in the existing snapshot is deprecated (I believe this needs to be deployed before the new snapshot?) which blocks me from deploying.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ses/lib/vdm-attributes.ts
+++ b/packages/aws-cdk-lib/aws-ses/lib/vdm-attributes.ts
@@ -3,11 +3,11 @@ import { CfnVdmAttributes } from './ses.generated';
 import { IResource, Resource } from '../../core';
 
 /**
- * Virtual Deliverablity Manager (VDM) attributes
+ * Virtual Deliverability Manager (VDM) attributes
  */
 export interface IVdmAttributes extends IResource {
   /**
-   * The name of the resource behind the Virtual Deliverablity Manager attributes.
+   * The name of the resource behind the Virtual Deliverability Manager attributes.
    *
    * @attribute
    */
@@ -15,7 +15,7 @@ export interface IVdmAttributes extends IResource {
 }
 
 /**
- * Properties for the Virtual Deliverablity Manager (VDM) attributes
+ * Properties for the Virtual Deliverability Manager (VDM) attributes
  */
 export interface VdmAttributesProps {
   /**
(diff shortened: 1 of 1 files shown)


LaurenceWarne:
Clarification Request: please let me know what README changes you think would be appropriate :slightly_smiling_face:

paulhcsun (changes requested):
Hi @LaurenceWarne, thank you for the contribution!

Regarding your clarification request, you could update the [README here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-rds/README.md?plain=1#L1065-L1084) by adding the new `name` property to the sample code. It's a fairly simple change so that should be sufficient. 

With regards to the integ test, I've pulled your code down and tested it an ran into the same issue. I was able to resolve it by doing 2 things:
1. Delete the existing [snapshot folder for that integ test](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/framework-integ/test/aws-rds/test/integ.instance-with-parameter-group.js.snapshot)
2. Change the `rds.PostgresEngineVersion` to a supported version: ``rds.PostgresEngineVersion/VER_16_2` and rerun the test with the `--update-on-failed` flag to regenerate the new snapshots.

LaurenceWarne:
> With regards to the integ test, I've pulled your code down and tested it an ran into the same issue. I was able to resolve it by doing 2 things:
>
>    Delete the existing [snapshot folder for that integ test](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/framework-integ/test/aws-rds/test/integ.instance-with-parameter-group.js.snapshot)
>    Change the rds.PostgresEngineVersion to a supported version: ``rds.PostgresEngineVersion/VER_16_2and rerun the test with the--update-on-failed` flag to regenerate the new snapshots.

Thanks for the guidance, should be done in the most recent commit!

packages/aws-cdk-lib/aws-rds/lib/parameter-group.ts
```diff
@@ -70,6 +70,13 @@ export interface ParameterGroupProps {
    */
   readonly engine: IEngine;
 
+  /**
+   * The name of this parameter group.
+   *
+   * @default CloudFormation-generated name
```
paulhcsun, line 76:
```suggestion
   * @default - CloudFormation-generated name
```
Something I learned recently, if the value immediately after a `@default` tag is not the actual value but rather a sentence describing the value, you need to separate it with a dash `-` or else it would think that `Cloudformation-generated` is the default value.

paulhcsun (approved):
This looks good, nice work @LaurenceWarne! 

And no worries, I can create an issue to update the existing integ tests using the deprecated postgres versions. Thanks again!

paulhcsun:
@mergifyio update

LaurenceWarne:
> And no worries, I can create an issue to update the existing integ tests using the deprecated postgres versions. Thanks again!

Cool :+1:. Do you need me to merge master - looks to be an issue with mergify?

paulhcsun:
@mergifyio update

paulhcsun (approved):

paulhcsun:
Reran, should hopefully work this time!

LaurenceWarne:
:tad

PR #29976, opened 2024-04-26
feat(redshift): multi AZ cluster

badmintoncryer:
### Issue # (if applicable)

Closes #<issue number here>.

### Reason for this change

AWS CDK cannot configure Redshift multi-AZ cluster.

### Description of changes

Add `multiAz` to `clusterProps`.

### Checklist

----


packages/@aws-cdk/aws-redshift-alpha/README.md
```diff
@@ -556,6 +556,23 @@ const cluster = new Cluster(this, 'Redshift', {
 cluster.addIamRole(role);
 ```
 
+## [Multi-AZ](https://docs.aws.amazon.com/redshift/latest/mgmt/managing-cluster-multi-az.html)
```
go-to-k, line 559: Instead of putting the link in the title, how about putting the link in the body of the text?

packages/@aws-cdk/aws-redshift-alpha/README.md
```diff
@@ -556,6 +556,23 @@ const cluster = new Cluster(this, 'Redshift', {
...
+
+Amazon Redshift supports multiple Availability Zones (Multi-AZ) deployments for provisioned RA3 clusters. By using Multi-AZ deployments, your Amazon Redshift data warehouse can continue operating in failure scenarios when an unexpected event happens in an Availability Zone.
```
go-to-k, line 561: Please insert appropriate line breaks. And it would be good to include an explanation of using the `multiAz` property.

packages/@aws-cdk/aws-redshift-alpha/README.md
```diff
@@ -556,6 +556,23 @@ const cluster = new Cluster(this, 'Redshift', {
...
+  masterUser: {
+    masterUsername: 'admin',
+  },
+  vpc, // 3 AZs are required for Multi-AZ
+  nodeType: redshift.NodeType.RA3_XLPLUS, // must be RA3 node type
+  clusterType: redshift.ClusterType.MULTI_NODE, // must be MULTI_NODE
+  numberOfNodes: 2, // must be 2 or more
+  multiAz: true,
+});
+```
```
go-to-k, line 574: How about actually creating an instance or defining `declare` for the `vpc`?(It does not occur a CI error, but..,)

packages/@aws-cdk/aws-redshift-alpha/lib/cluster.ts
```diff
@@ -385,6 +385,13 @@ export interface ClusterProps {
    * @default - false
    */
   readonly enhancedVpcRouting?: boolean;
+
+  /**
+   * Indicating whether Amazon Redshift should deploy the cluster in two Availability Zones.
+   *
+   * @default false
```
go-to-k, line 392: I thought this would be more appropriate since we are not actually setting false in CDK.

```suggestion
   * @default - false
```

packages/@aws-cdk/aws-redshift-alpha/lib/cluster.ts
```diff
@@ -558,6 +565,17 @@ export class Cluster extends ClusterBase {
...
+    const nodeType = props.nodeType || NodeType.DC2_LARGE;
+
+    if (props.multiAz) {
+      if (!nodeType.startsWith('ra3')) {
+        throw new Error('Multi-AZ cluster is only supported for RA3 node types.');
+      }
+      if (clusterType === ClusterType.SINGLE_NODE) {
+        throw new Error('Multi-AZ cluster is not supported for single-node clusters.');
+      }
+    }
```
go-to-k, line 577: It would be nice to output the value set in props by users: `got: xxx`

go-to-k (changes requested):
Thanks for this PR. I made just some minor comments.

packages/@aws-cdk/aws-redshift-alpha/lib/cluster.ts
```diff
@@ -558,6 +565,17 @@ export class Cluster extends ClusterBase {
       );
     }
 
+    const nodeType = props.nodeType || NodeType.DC2_LARGE;
+
+    if (props.multiAz) {
+      if (!nodeType.startsWith('ra3')) {
+        throw new Error(`Multi-AZ cluster is only supported for RA3 node types. got: ${nodeType}`);
```
go-to-k, line 572: for the cases that users do not specify the props.

```suggestion
        throw new Error(`Multi-AZ cluster is only supported for RA3 node types, got: ${props.nodeType}`);
```

packages/@aws-cdk/aws-redshift-alpha/lib/cluster.ts
```diff
@@ -558,6 +565,17 @@ export class Cluster extends ClusterBase {
...
+      }
+      if (clusterType === ClusterType.SINGLE_NODE) {
+        throw new Error('Multi-AZ cluster is not supported for single-node clusters. got: `ClusterType.SINGLE_NODE`');
```
go-to-k, line 575: Might want to change the message a bit, according to [the message](https://github.com/go-to-k/aws-cdk/blob/247aa35676b3b1ead91a56fd00e819c67a9a7285/packages/@aws-cdk/aws-redshift-alpha/lib/cluster.ts#L669).

```suggestion
        throw new Error('Multi-AZ cluster is not supported for cluster type single-node');
```

packages/@aws-cdk/aws-redshift-alpha/test/cluster.test.ts
```diff
@@ -614,6 +614,67 @@ test('elastic ip address', () => {
...
+    expect(() => {
+      new Cluster(stack, 'Redshift', {
+        masterUser: {
+          masterUsername: 'admin',
+        },
+        vpc,
+        nodeType: NodeType.DS2_XLARGE,
+        multiAz: true,
+      });
+    }).toThrow('Multi-AZ cluster is only supported for RA3 node types.');
```
go-to-k, line 660: Please match the changed message.

packages/@aws-cdk/aws-redshift-alpha/test/cluster.test.ts
```diff
@@ -614,6 +614,67 @@ test('elastic ip address', () => {
...
+      new Cluster(stack, 'Redshift', {
+        masterUser: {
+          masterUsername: 'admin',
+        },
+        vpc,
+        nodeType: NodeType.RA3_XLPLUS,
+        multiAz: true,
+        clusterType: ClusterType.SINGLE_NODE,
+      });
+    }).toThrow('Multi-AZ cluster is not supported for single-node clusters.');
```
go-to-k, line 674: Please match the changed message.

go-to-k (changes requested):
[This comment](https://github.com/aws/aws-cdk/pull/29976#discussion_r1586253457) (`And it would be good to include an explanation for the need of the multiAz property.`) has not been reflected yet, so it may be in the process of being corrected, but I wanted to comment first on what I noticed.

packages/@aws-cdk/aws-redshift-alpha/lib/cluster.ts
go-to-k: Or the following:

```suggestion
        throw new Error('Multi-AZ cluster is not supported for `clusterType` single-node');
```

Or

```suggestion
        throw new Error('Multi-AZ cluster is not supported for single-node `clusterType`');
```

If you have something you think is good...

packages/@aws-cdk/aws-redshift-alpha/README.md
badmintoncryer: > And it would be good to include an explanation for the need of the multiAz property.
Does this mean to add a sentence about the benefits of enabling the multiAz property? (Is 'By using Multi-AZ deployments...' not enough for that??)

badmintoncryer:
@go-to-k Thank you for your review!! I'm sorry that I've pushed temporary commits. I've addressed all your comments.

packages/@aws-cdk/aws-redshift-alpha/README.md
badmintoncryer: I see! Thanks a lot!

go-to-k:
Oh... failed...

go-to-k:
> Oh... failed...

Succeeded! I'll check them.

go-to-k (approved):
LGTM. Thanks for the changes!

badmintoncryer:
Thank you always!

packages/@aws-cdk/aws-redshift-alpha/lib/cluster.ts
```diff
@@ -558,6 +565,17 @@ export class Cluster extends ClusterBase {
       );
     }
 
+    const nodeType = props.nodeType || NodeType.DC2_LARGE;
```
shikha372, line 568: do we need to specify it here its already set as the default for this prop?

badmintoncryer: This specification is not new.

https://github.com/aws/aws-cdk/pull/29976/files#diff-8666885f385d3c87fb98f835d762b86e7a98bfd217bd69c9a210041ab07585d6L577

If we change this implementation, it will cause a breaking change. Therefore, I think it is better to keep it as is.

shikha372: understand, I should've been more specific wanted to check if we could keep it as `const nodeType = props.nodeType` , since NodeType.DC2_LARGE is already set as default but I see that we've kept it as before, thank you for explaning.

shikha372 (approved):
Thank you for your contribution @badmintoncryer

badmintoncryer:
@mergify update

PR #29978, opened 2024-04-26
chore: stop emmitting sourcemaps in integ-tests

mrgrain:
### Reason for this change

While working on a related issue, I noticed that some integ-test snapshots are very brittle because assets include inline sourcemaps and ts files.

### Description of changes

With this change, we are disabling all sourcemaps for integ-tests and explicitly exclude ts source files from assets.
This is okay because the handlers are just written to be used inside integ tests.

Another change was made to disable the inline sourcemap for the custom resource handler entrypoint. The actual handlers do not have sourcemaps, so this was silly and probably just an oversight.

Run integ tests to update snapshots.

### Checklist

----


colifran (approved):
Thanks for getting this cleaned up!

PR #29987, opened 2024-04-27
feat(pipes-targets): add step functions state machine target

### LINKED ISSUE #29665: (pipes-targets): Add Step functions target for EventBridge pipes
### Describe the feature

Add a implementation for the Step function target for EventBridge pipes in the `pipes-targets` module.

### Use Case

For feature completeness of the L2 pipes module this source is required

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

-

-

WtfJoke:
### Issue #29665

Closes #29665

### Reason for this change
Step Function target is not supported yet by pipes-targets.

### Description of changes
- Added step function as a pipes target.
- I've decided to make the `invocationType` a required parameter, since this made the code clearer and make users aware of how they want the step function to be invoked. 
The [AWS::Pipes::Pipe PipeTargetStateMachineParameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetstatemachineparameters.html) has this as an optional parameter (defaulting to Request-Response), which can lead the user unknowingly in a broken pipe, because cdk's StateMachines [default](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions.StateMachine.html#statemachinetype) to Standard Workflow, which is not compatible with Request-Response Invocation Type.
- Currently there seems no way to prevent users from creating a pipe with an imported Standard StateMachine and InvocationType Request-Response as the stateMachineType cant be read (or I dont know how :D)

- Added unit tests
- Added integration test

### Checklist

----
I've talked with @RaphaelManke and he was fine for me opening up a PR (put him as a co-author nevertheless) :)


packages/@aws-cdk/aws-pipes-targets-alpha/lib/index.ts
```diff
@@ -1 +1,2 @@
 export * from './sqs';
+export * from './step-function';
```
mrgrain, line 2: Please make sure these are named consistently 
```suggestion
export * from './stepfunctions';
```

WtfJoke: Thanks for taking a look already :)

Currently its consistent using `step-function` everywhere. Should I change it to the plural version of it?

mrgrain: How do you mean "everywhere"? I was referring to the module name which is `aws-stepfunctions`.

(You might totally be right about this by the way)

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
```diff
@@ -63,3 +68,42 @@ const pipe = new pipes.Pipe(this, 'Pipe', {
     target: pipeTarget
 });
 ```
+
+### AWS Step Function
```
mrgrain, line 72:
```suggestion
### AWS Step Functions
```

packages/@aws-cdk/aws-pipes-targets-alpha/lib/index.ts
WtfJoke: >How do you mean "everywhere"?

I meant the filenames. 

> I was referring to the module name which is aws-stepfunctions.

Ah I think I know now what you meant. The package name is `aws-stepfunctions` and I should keep that name consistent also in this pipes-targets module, changed it to stepfunctions :)

I kept the name `StepFunctionTarget` if you think I should change that as well let me know.

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
WtfJoke: Good catch! Changed it

WtfJoke:
@mrgrain just wanted to ask if there is something else I should change or does an additional person needs to review my PR (or I just need to wait a bit more)?

mrgrain:
> @mrgrain just wanted to ask if there is something else I should change or does an additional person needs to review my PR (or I just need to wait a bit more)?

Someone else should get around to this eventually. I'm not actually working on the CDK anymore 🙈 
But ping me if you don't get a review next week!

WtfJoke:
Thanks, appreciate it! Lets hope its not necessary :D

WtfJoke:
@mrgrain unfortunately I need to come back to your offer as the PR hasnt been merged and a week passed. So I hope pinging you helps :D Thanks!

packages/@aws-cdk/aws-pipes-targets-alpha/lib/index.ts
```diff
@@ -1 +1,2 @@
 export * from './sqs';
+export * from './stepfunctions';
```
mrgrain, line 2:
```suggestion
export * from './stepfunctions';

```

packages/@aws-cdk/aws-pipes-targets-alpha/lib/stepfunctions.ts
```diff
@@ -0,0 +1,83 @@
...
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetparameters.html#cfn-pipes-pipe-pipetargetparameters-inputtemplate
+   * @default none
+   */
+  readonly inputTransformation?: IInputTransformation;
+
+  /**
+   * Specify whether to invoke the Step Functions state machine synchronously (`REQUEST_RESPONSE`) or asynchronously (`FIRE_AND_FORGET`).
+   *
+   * @see http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetsqsqueueparameters.html#cfn-pipes-pipe-pipetargetsqsqueueparameters-messagededuplicationid
```
mrgrain, line 21: wrong link

packages/@aws-cdk/aws-pipes-targets-alpha/lib/stepfunctions.ts
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Invoke StepFunction synchronously (`StartSyncExecution`) and wait for the execution to complete. See https://docs.aws.amazon.com/step-functions/latest/apireference/API_StartSyncExecution.html for more details.
+   */
+  REQUEST_RESPONSE = 'REQUEST_RESPONSE',
+}
+
+/**
+ * An EventBridge Pipes target that sends messages to a StepFunction.
+ */
+export class StepFunctionTarget implements ITarget {
```
mrgrain, line 45:
```suggestion
export class StateMachineTarget implements ITarget {
```

packages/@aws-cdk/aws-pipes-targets-alpha/lib/stepfunctions.ts
```diff
@@ -0,0 +1,83 @@
...
+   * @see http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetsqsqueueparameters.html#cfn-pipes-pipe-pipetargetsqsqueueparameters-messagededuplicationid
+   */
+  readonly invocationType: StepFunctionInvocationType;
+}
+
+/**
+ * InvocationType for invoking StepFunction.
+ * @see https://docs.aws.amazon.com/eventbridge/latest/pipes-reference/API_PipeTargetStateMachineParameters.html
+ */
+export enum StepFunctionInvocationType {
```
mrgrain, line 30:
```suggestion
export enum StateMachineInvocationType {
```

packages/@aws-cdk/aws-pipes-targets-alpha/lib/stepfunctions.ts
```diff
@@ -0,0 +1,83 @@
...
+  constructor(stateMachine: sfn.IStateMachine, parameters: StateMachineTargetParameters) {
+    this.stateMachine = stateMachine;
+    this.targetArn = stateMachine.stateMachineArn;
+    this.invocationType = parameters.invocationType;
+    this.inputTemplate = parameters.inputTransformation;
+
+    if (this.stateMachine instanceof StateMachine
+      && this.stateMachine.stateMachineType === StateMachineType.STANDARD
+      && this.invocationType === StepFunctionInvocationType.REQUEST_RESPONSE) {
+      throw new Error('Standard Workflows do not support REQUEST_RESPONSE invocation type');
```
mrgrain, line 61: Can you phrase this to include a solution for the user.
```suggestion
      throw new Error('STANDARD state machine workflows do not support the REQUEST_RESPONSE invocation type. Use FIRE_AND_FORGET instead.');
```

Also let's default to FIRE_AND_FORGET

packages/@aws-cdk/aws-pipes-targets-alpha/lib/stepfunctions.ts
```diff
@@ -0,0 +1,83 @@
...
+   * @default none
+   */
+  readonly inputTransformation?: IInputTransformation;
+
+  /**
+   * Specify whether to invoke the Step Functions state machine synchronously (`REQUEST_RESPONSE`) or asynchronously (`FIRE_AND_FORGET`).
+   *
+   * @see http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetsqsqueueparameters.html#cfn-pipes-pipe-pipetargetsqsqueueparameters-messagededuplicationid
+   */
+  readonly invocationType: StepFunctionInvocationType;
```
mrgrain, line 23: I think we can default to FIRE_AND_FORGET here.

mrgrain (dismissed):
Looks great! Got a couple more minor renames and let's set a default for InvocationType so that most users won't have to specify parameters at all.

Also trying to get @RaphaelManke to review as well

RaphaelManke (commented):
Overall great job @WtfJoke. 
I have only one thing that I am not 100% happy about.
Can we define a common name for Step function that is used all over the place? 
Currently we have : 
- AWS Step Functions
- StepFunction
- StateMachine
- stepFunctionStateMachine
- sfn

The EventBridge targets constructs calls the target 
[SfnStateMachine](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_events_targets.SfnStateMachine.html)

Maybe we should align to the naming of that because its a name from an already stable package. 
Whats your thought in that @mrgrain ?

My suggestion is, name the target `SfnStateMachine`

WtfJoke:
> Can we define a common name for Step function that is used all over the place?


This is something I am not happy about it as well. 
I tried to Stick mostly to the term "Step Function", but the terms Step Function and State Machine are mostly used interchangeable. I'm open for either name.

Please keep in mind its not only the Targets name (StepFunctionTarket vs StateMachineTarget vs SfnStateMachineTarget (@RaphaelManke you meant to include `Target` right?)) but also the in the corresponding docs.

Currently in the docs:
> StateMachine StepFunction target properties.

or  

> InvocationType for invoking StepFunction.
> Invoke StepFunction asynchronously (`StartExecution`). See https://docs.aws.amazon.com/step-functions/latest/apireference/API_StartExecution.html for more details.

or in the README  
>### AWS Step Functions

I am not so eager on writing SfnStateMachine everywhere in the docs. I would favor choosing StepFunction or StateMachine there.
What do you think @RaphaelManke @mrgrain?

P.S: I've rebased my changes on latest main and incorporated the review feedback from @mrgrain.

packages/@aws-cdk/aws-pipes-targets-alpha/lib/stepfunctions.ts
WtfJoke: good catch, fixed it

packages/@aws-cdk/aws-pipes-targets-alpha/lib/stepfunctions.ts
WtfJoke: adjusted to your suggestion and defaulted to `FIRE_AND_FORGET`

mrgrain:
> * AWS Step Functions

This should be used in docs, when referring to the service. 
"AWS Step Functions State Machine" can be used to refer to the resource in the docs, but when used multiple times "State Machine" is enough.

> * StepFunction

This should never be used, given that it is missing an `s` ;)

> * StateMachine

This the correct way to refer to just the resource.

> * stepFunctionStateMachine

Ignoring the case, this Would be correct way to name the combination of service & resource

> * sfn
> 
> The EventBridge targets constructs calls the target [SfnStateMachine](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_events_targets.SfnStateMachine.html)

I'm inclined to agree. `sfn` is the short version [used by the SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/sfn/) as well.

> Maybe we should align to the naming of that because its a name from an already stable package. Whats your thought in that @mrgrain ?

Probably. I think we named the queue target wrong: `SqsTarget` when it really should be `SqsQueueTarget`.
But then all classes in the modules are targets, so maybe `SqsQueue` would be enough.

> My suggestion is, name the target `SfnStateMachine` 

I come to the same conclusion, in particular about dropping the `Target` suffix.
There could be an argument about calling it `StepFunctionsStateMachine` (I kind of like it better) but given the precedence of  EventBridge targets and that `sfn` is a semi official abbreviation, I'm okay with it.

mrgrain:
> This is something I am not happy about it as well. I tried to Stick mostly to the term "Step Function", but the terms Step Function and State Machine are mostly used interchangeable. I'm open for either name.

They should not be used interchangeably. We should make sure to explicitly call out the resource we are integrating with, which is the AWS Step Functions State Machine. Or shorter State Machine.

When grouping things, we tend to use the service. So that's why the file is named `stepfunctions.ts` or in the docs we might have a heading "AWS Step Functions". The idea being that if there were a second resource in the same service/module they would go into the same group. Of course in practice this rarely happens.

> Please keep in mind its not only the Targets name (StepFunctionTarket vs StateMachineTarget vs SfnStateMachineTarget (@RaphaelManke you meant to include `Target` right?)) but also the in the corresponding docs.

Yeah I think we should drop `Target` from the name. We can still refer to it in the docs as "`SfnStateMachine` target" when we want to make it explicit we are talking about a target.

> > InvocationType for invoking StepFunction.
> > Invoke StepFunction asynchronously (`StartExecution`). See https://docs.aws.amazon.com/step-functions/latest/apireference/API_StartExecution.html for more details.

Just because you called it out here, I think this is a nice example why it's important to differentiate between the service and the resource. StepFunction (without `s) doesn't exist. And StepFunctions (the service) doesn't have an InvocationType. Only the State Machine can be invoked with an InvocationType.

> I am not so eager on writing SfnStateMachine everywhere in the docs. I would favor choosing StepFunction or StateMachine there. What do you think @RaphaelManke @mrgrain?

I can go over the docs again to give some concrete notes. In example code it would be `SfnStateMachine`. In copy text, it would likely be "State Machine" or "the State Machine target" or "parameters of the State Machine target", etc.

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
```diff
@@ -27,6 +27,11 @@ For more details see the service documentation:
 
 Pipe targets are the end point of a EventBridge Pipe.
 
+The following targets are supported:
+
+1. `targets.SqsTarget`: [Send event source to a Queue](#amazon-sqs)
+2. `targets.StateMachineTarget`: [Start a State Machine from an event source](#aws-step-functions)
```
mrgrain, line 33: Pending the code rename
```suggestion
2. `targets.SfnStateMachine`: [Invoke a State Machine from an event source](#aws-step-functions)
```

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
```diff
@@ -63,3 +68,57 @@ const pipe = new pipes.Pipe(this, 'Pipe', {
     target: pipeTarget
 });
 ```
+
+### AWS Step Functions
```
mrgrain, line 72: This makes sense as a heading. It could also be
```suggestion
### AWS Step Functions State Machine
```

WtfJoke:
I kinda prefer `StepFunctionsStateMachine` over `SfnStateMachine`, but in the end its up to you to decide.

> I can go over the docs again to give some concrete notes. In example code it would be SfnStateMachine. In copy text, it would likely be "State Machine" or "the State Machine target" or "parameters of the State Machine target", etc.

Once we set on a name, I can go through the code+docs again and afterwards it would be great to have a review again. IMHO no need for docs review in advance (but as you like).

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
```diff
@@ -63,3 +68,57 @@ const pipe = new pipes.Pipe(this, 'Pipe', {
...
+
+const pipeTarget = new targets.StateMachineTarget(targetStateMachine,{});
+
+const pipe = new pipes.Pipe(this, 'Pipe', {
+    source: new SomeSource(sourceQueue),
+    target: pipeTarget
+});
+```
+
+Specifying the Invocation Type of the State Machine.
```
mrgrain, line 88: This was already good. Slightly more explicit:
```suggestion
Specifying the Invocation Type when the target State Machine is invoked:
```

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
```diff
@@ -63,3 +68,57 @@ const pipe = new pipes.Pipe(this, 'Pipe', {
...
+);
+
+
+const pipe = new pipes.Pipe(this, 'Pipe', {
+    source: new SomeSource(sourceQueue),
+    target: pipeTarget
+});
+```
+
+The target input can be transformed:
```
mrgrain, line 107:
```suggestion
The input to the target State Machine can be transformed:
```

packages/@aws-cdk/aws-pipes-targets-alpha/lib/stepfunctions.ts
```diff
@@ -0,0 +1,84 @@
+import { IInputTransformation, IPipe, ITarget, TargetConfig } from '@aws-cdk/aws-pipes-alpha';
+import { IRole } from 'aws-cdk-lib/aws-iam';
+import * as sfn from 'aws-cdk-lib/aws-stepfunctions';
+import { StateMachine, StateMachineType } from 'aws-cdk-lib/aws-stepfunctions';
+
+/**
+ * StateMachine StepFunction target properties.
```
mrgrain, line 7:
```suggestion
 * Parameters for the SfnStateMachine target
```

packages/@aws-cdk/aws-pipes-targets-alpha/lib/stepfunctions.ts
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * The input transformation to apply to the message before sending it to the target.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetparameters.html#cfn-pipes-pipe-pipetargetparameters-inputtemplate
+   * @default none
+   */
+  readonly inputTransformation?: IInputTransformation;
+
+  /**
+   * Specify whether to invoke the Step Functions state machine synchronously (`REQUEST_RESPONSE`) or asynchronously (`FIRE_AND_FORGET`).
```
mrgrain, line 19:
```suggestion
   * Specify whether to invoke the State Machine synchronously (`REQUEST_RESPONSE`) or asynchronously (`FIRE_AND_FORGET`).
```

packages/@aws-cdk/aws-pipes-targets-alpha/lib/stepfunctions.ts
```diff
@@ -0,0 +1,84 @@
...
+   *
+   * @see http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetsqsqueueparameters.html#cfn-pipes-pipe-pipetargetsqsqueueparameters-messagededuplicationid
+   * @default StateMachineInvocationType.FIRE_AND_FORGET
+   */
+  readonly invocationType?: StateMachineInvocationType;
+}
+
+/**
+ * InvocationType for invoking StepFunction.
```
mrgrain, line 28:
```suggestion
 * InvocationType for invoking the State Machine.
```

mrgrain (approved):

WtfJoke:
Thanks a bunch!

PR #29998, opened 2024-04-29
(feat): introduce custom runtime aspect class for nodejs

shikha372:
### Issue # (if applicable)

Closes #<issue number here>.

### Description of changes

Introduce custom aspect class

### Checklist

----


packages/aws-cdk-lib/core/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,138 @@
...
+      const targetnode = node.node.children.find((child) => child instanceof cdk.CfnResource &&
+      child.cfnResourceType === 'AWS::Lambda::Function') as cdk.CfnResource;
+      targetnode.addPropertyOverride('Runtime', 'nodejs20.x');
+    }
+  }
+
+  private handleProvider(provider: Provider): void {
+    const functionNodes = provider.node.children.filter((child) => child instanceof lambda.Function);
+    for ( var node of functionNodes) {
+      const targetNode = node.node.children.find((child) => child instanceof cdk.CfnResource && child.cfnResourceType === 'AWS::Lambda::Function') as cdk.CfnResource;
```
colifran, line 54: A few things about this logic:
1. You have this same thing happening in three different spots. This is something that can be factored out.
2. You're visiting child nodes that will be visited anyways. Let me explain. All custom resource providers are just Lambda functions. This means they all extend `CfnResource` and have a `cfnResourceType` of `AWS::Lambda::Function`. You should only need to just have logic that checks for an instance of `CfnResource` and a `cfnResourceType` of `AWS::Lambda::Function`. You don't need all these branches.

packages/aws-cdk-lib/core/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  constructor(key: string) {
+    super(key);
+  }
+}
+
+/**
+ * Runtime
+ */
+export class Runtime {
```
colifran, line 119: This class isn't really a `Runtime`. You're building an abstraction on top of an `Aspect`, but I think this is going to be confusing as it is. People should be able to use the aspect with the `Aspects` abstraction. Also a `Runtime` in core doesn't really make sense. `Runtimes` should be attached to resources, i.e., Lambda functions or synthetics canaries. Let's remove this and just let people use `Aspects.of(...).add(...)`

packages/aws-cdk-lib/core/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,138 @@
...
+    if (ses.runtime && this.isValidRuntime(ses.runtime)) {
+      ses.addPropertyOverride('Runtime', 'nodejs20.x');
+    }
+  }
+
+  /**
+ *Runtime Validation
+ *
+ */
+  private isValidRuntime(runtime: string) : boolean {
```
colifran, line 88: This is just looking for an equivalent runtime. I think you want this to look for a runtime family. The way it is set up will mean we only replace `nodejs20.x` runtimes with other `nodejs20.x` runtimes. I think you want this to look if the runtime for the visited node is in the runtime family you're trying to replace.

packages/aws-cdk-lib/core/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,138 @@
...
+import { IAspect, Aspects } from './aspect';
+import * as cdk from '../../../aws-cdk-lib';
+import * as lambda from '../../aws-lambda';
+import { ReceiptRuleSet } from '../../aws-ses/lib/receipt-rule-set';
+import { Provider } from '../../custom-resources/lib';
+
+/**
+ * Runtime aspec
+ */
+abstract class RuntimeAspectsBase implements IAspect {
```
colifran, line 12: This shouldn't live in `core`. How about putting all of this in `aws-lambda` since this aspect will act on Lambda functions?

packages/aws-cdk-lib/core/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  private getRuntimeProperty(node: cdk.CfnResource) {
+    return (node.getResourceProperty('runtime') || node.getResourceProperty('Runtime')) as string;
+  }
+}
+//}
+/**
+ * RuntimeAspect
+ */
+export class RuntimeAspect extends RuntimeAspectsBase {
```
colifran, line 100: Let's opt for an enum-like class. This class is another layer of abstraction on top of `Runtime` which is an abstraction on top of `Aspect`. Let's enable users to use `Aspects.of(...).add(...)` with the aspect being created. I would see this looking like: `Aspects.of(scope).add(RuntimeAspect.nodeJs20())` or something similar.

packages/aws-cdk-lib/core/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,138 @@
...
+  public readonly targetRuntime: string;
+
+  constructor(runtime: string) {
+    this.targetRuntime = runtime;
+  }
+
+  public visit(construct: IConstruct): void {
+
+    //To handle custom providers
+    if (construct instanceof cdk.CustomResourceProviderBase) {
```
colifran, line 26: I don't think we need to check for instances of `CustomResourceProviderBase`, `Provider`, or `ReceptRule`. At their core these are all the same construct: `CfnResource` of type `AWS::Lambda::Function`. This should be all we need to look for.

packages/aws-cdk-lib/core/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,138 @@
...
+    const functionNodes = provider.node.children.filter((child) => child instanceof lambda.Function);
+    for ( var node of functionNodes) {
+      const targetNode = node.node.children.find((child) => child instanceof cdk.CfnResource && child.cfnResourceType === 'AWS::Lambda::Function') as cdk.CfnResource;
+      if (this.isValidRuntime(this.getRuntimeProperty(targetNode))) {
+        targetNode.addPropertyOverride('Runtime', 'nodejs20.x');
+      }
+
+      // onEvent Handlers
+      const onEventHandler = provider.onEventHandler as lambda.Function;
+      const onEventHandlerRuntime = onEventHandler.node.children.find((child) => child instanceof cdk.CfnResource && child.cfnResourceType === 'AWS::Lambda::Function') as cdk.CfnResource;
```
colifran, line 61: Rather than checking for `instanceof CfnResource` please use the `CfnResource.isResource()` static method.

packages/aws-cdk-lib/core/lib/nodejs-aspect.ts
shikha372: This is more to the point that we want to update it only 'CustomResources' and not for all lambda functions which i think we can cover with `CfnResource` of type `AWS::Lambda::Function` .

packages/@aws-cdk/aws-nodejs-aspect/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,137 @@
...
+import { AwsCustomResource, Provider } from 'aws-cdk-lib/custom-resources';
+import { ReceiptRuleSet } from 'aws-cdk-lib/aws-ses';
+
+/**
+ * Runtime aspect
+ */
+abstract class RuntimeAspectsBase implements IAspect {
+
+  /**
+   * The string key for the runtime
```
colifran, line 15: This isn't descriptive of what `targetRuntime` is. How about something like, "The runtime that the aspect will target for updates while walking the construct tree."

colifran: Also why is this public? This seems like an implementation detail.

packages/@aws-cdk/aws-nodejs-aspect/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,137 @@
...
+        return 'Unsupported';
+    }
+  }
+
+}
+
+/**
+ * RuntimeAspect
+ */
+export class RuntimeAspect extends RuntimeAspectsBase {
```
colifran, line 126: Do we really want to expose this? Additionally, if I'm using this I'm not clear what `key` is. Also, `string` is extremely loose. I could pass anything as a string. It seems like we should be building an enum-like class on `RuntimeAspect` which is extensible. Introducing `NodeJsAspect` means we would need a new class for every single runtime. I realize we may never introduce new runtimes, but we should make this extensible. How about something like this:
```ts
export class RuntimeAspect extends RuntimeAspectsBase {
	public static nodeJs20() {
		// return class implementation
	}
}
```

This may be better to put everything on `RuntimeAspectsBase` though since we would just be returning the implementation and another class is probably extra unnecessary code.

packages/@aws-cdk/aws-nodejs-aspect/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,137 @@
...
+  constructor(key: string) {
+    super(key);
+  }
+}
+
+export class NodeJsAspect {
+  public static modifyRuntimeTo(key: lambda.Runtime) {
```
colifran, line 133: See my above comment. I think a better user experience here is to do all the construction for them. We can expose options like `nodeJs20` and then we can to all the construction like passing the associated `Runtime` for the user. This also prevents the user from breaking themselves, like passing node 16 which would be breaking since it only comes with SDK v2.

packages/@aws-cdk/aws-nodejs-aspect/lib/nodejs-aspect.ts
```diff
@@ -0,0 +1,137 @@
...
+import * as cdk from 'aws-cdk-lib';
+import * as lambda from 'aws-cdk-lib/aws-lambda';
+import { IAspect } from 'aws-cdk-lib/core';
+import { AwsCustomResource, Provider } from 'aws-cdk-lib/custom-resources';
+import { ReceiptRuleSet } from 'aws-cdk-lib/aws-ses';
+
+/**
+ * Runtime aspect
+ */
+abstract class RuntimeAspectsBase implements IAspect {
```
colifran, line 12: Nit: Let's change this to `RuntimeAspectBase`

packages/@aws-cdk/aws-nodejs-aspect/lib/nodejs-aspect.ts
shikha372: yes, was trying with implementation earlier too but it was not reflecting the updated nodejs version in template this way, gave it a second attempt and it seems to be working now updated in latest commit.

packages/@aws-cdk/aws-nodejs-aspect/lib/nodejs-aspect.ts
shikha372: updated in latest commit for your review

packages/@aws-cdk/aws-nodejs-aspect/lib/nodejs-aspect.ts
shikha372: yeah, these strings was to pass initial lints, have more formatted version and correct typing in latest commit.

PR #30009, opened 2024-04-30
feat: support for IAM Identity Center in security diff

### LINKED ISSUE #29835: Big: CDK Diff does not include AWS::SSO::Assignment in security related changes
### Context
We received a security advisory regarding diff not handling AWS::SSO::Assignment resource changes as we security related change. See: https://github.com/aws/aws-cdk/security/advisories/GHSA-qpw6-2c95-6gcq We currently handle two packages as security related in [cloudformation-diff](https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk/cloudformation-diff): `iam` and `network`. We should add similar logic for SSO.

### Task
Update diff to include these resource changes. As part of this task we should also check to see if any other security related libraries have been added since this was last updated (Organizations comes to mind) and check to see if any updates have been added within the known libraries that we may not be checking for already. If those are found, new tasks should be added for each of those.

### Acceptance Criteria
- AWS::SSO::Assignment resources are added to the list of security changes when cdk diff is run
- A integ test is added/updated to include checking for these resources
- New tasks are created for each other resource/library type that we find are not currently covered

bergjaak:
### Issue # (if applicable)

Closes #29835 

### Reason for this change

IAM Identity Center resources were ignored in the security diff

### Description of changes

* Adds the  IAM Identity Center resources to CDK diff
* fixes not presenting property changes when a resource is removed from the template

* Added unit tests and integration tests.
* Ran the integration tests that mention cdk diff (`bin/run-suite -a cli-integ-tests -t 'cdk diff'`):
```
Test Suites: 2 skipped, 1 passed, 1 of 3 total
Tests:       90 skipped, 13 passed, 103 total
Snapshots:   0 total
Time:        312.397 s
Ran all test suites with tests matching "cdk diff": 
```

### Dependent PRs

* Before this change can be merged, this change https://github.com/cdklabs/awscdk-service-spec/pull/1052 must be merged.

### Checklist
- [Y] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----


bergjaak:
Exemption Request

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -775,6 +775,158 @@ integTest('cdk diff --fail with multiple stack exits with if any of the stacks c
...
+}));
+
+integTest('cdk diff --security-only successfully outputs sso-access-control information', withDefaultFixture(async (fixture) => {
+  const diff = await fixture.cdk(
+    ['diff', '--security-only', fixture.fullStackName('sso-access-control')],
+  );
+  `┌───┬────────────────────────────────┬────────────────────────┬─────────────────────────────────┐
+   │   │ Resource                       │ InstanceArn            │ AccessControlAttributes         │
+   ├───┼────────────────────────────────┼────────────────────────┼─────────────────────────────────┤
+   │ + │ \${instanceAccessControlConfig} │ arn:aws:test:testvalue │ Key: first, Values: [a]         │
```
comcalvi, line 871: the pipe characters are off by a space on this line, could we fix that alignment?

packages/@aws-cdk/cloudformation-diff/lib/diff/types.ts
```diff
@@ -128,11 +128,11 @@ export class TemplateDiff implements ITemplateDiff {
         continue;
       }
 
-      if (!resourceChange.newResourceType) {
+      if (!resourceChange.resourceType) {
```
comcalvi, line 131: why change from `newResourceType` to `resourceType`?

packages/@aws-cdk/cloudformation-diff/lib/iam/iam-identity-center.ts
```diff
@@ -0,0 +1,116 @@
...
+    return deepEqual(this, other);
+  }
+}
+
+/**
+ * AWS::SSO::InstanceAccessControlAttributeConfiguration
+ */
+export interface ISsoInstanceACAConfigProps {
+  ssoInstanceArn: string;
+}
```
comcalvi, line 88:
```suggestion
}

```

comcalvi (changes requested):
Nothing but minor comments, this looks really good! Nice work.

packages/@aws-cdk/cloudformation-diff/lib/diff/types.ts
bergjaak: Great question! I will leave a comment in the commit I push explaining why this is resourceType, because I think it's worth documenting in some form.

The reason is, when a resource  (let's say resource X, an AWS::IAM::Policy) is removed from a template, then X.oldResourceType == AWS::IAM::Policy, but X.newResourceType == undefined -- and we need the resourceType to be able to detect if there is a security relevant diff! (Note that X.resourceType == X.oldResourceType || X.newResourceType). However, suppose X had a policy on it stating that user 012345678901 wasn't allowed to get items from a DDB table or wasn't allowed to invoke a lambda function -- now user 012345678901 is allowed to do those things, and so the diff is security relevant.

Therefore, we should check the resourceType instead of newResourceType, which might be undefined.

bergjaak: For a concrete example, here's a test case I added for this issue: https://github.com/aws/aws-cdk/blob/38f1e5ca3860ab365198d670eb5f75dab5c3e452/packages/%40aws-cdk/cloudformation-diff/test/iam/detect-changes.test.ts#L461-L482

So before this change, that test case would have failed

comcalvi (approved):

bergjaak:
https://tiny.amazon.com/dj097tok/IsenLink test run

Issue #28588, opened 2024-01-05
aws_eks: Support AccessConfig on aws_eks.Cluster


TirTech:
### Describe the feature

Support setting AccessConfig (or at least AuthenticationMode) on `aws_eks.Cluster`.

### Use Case

We currently use `aws_eks.Cluster` to create our clusters, and would like to start using EKS access modes (specifically `API_AND_CONFIG_MAP`)

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.118.0

### Environment details (OS name and version, etc.)

Ubuntu 22.04.3 LTS on Windows 11 x86_64, python 3.10

pahud:
Yes this is an awesome feature we just announced and we definitely should support that. Would you tell me more about your expected CDK experience with that?

TirTech:
> Yes this is an awesome feature we just announced and we definitely should support that. Would you tell me more about your expected CDK experience with that?

Sure. When creating an EKS cluster, we would like to opt in to access entries by passing the authentication mode to the constructor:
```python
cluster = aws_eks.Cluster(
            self,
            id="MyCluster",
            # ...snipped
            authentication_mode=aws_eks.AuthenticationMode.API_AND_CONFIG_MAP
        )
```
with `aws_eks.AuthenticationMode` being one of the available [cluster authentication modes](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-eks-cluster-accessconfig.html#cfn-eks-cluster-accessconfig-authenticationmode) (default `CONFIG_MAP`).

For now we would go on to create `CfnAccessEntry` resource for each of the IAM roles we want to grant cluster access to, but in the future we would be looking to get the higher level `AccessEntry` resource added so we could create these entries for some `aws_iam.Role` "role" either on their own:
```python
aws_eks.AccessEntry(
            self,
            id="AdministatorAccessEntry",
            cluster_name=cluster.cluster_name,
            principal_arn=role.role_arn,
            # other options
        )
```
or with a helper method on the cluster:
```python
cluster.add_access_entry(self, id="AdministatorAccessEntry", principal_arn=role.role_arn, ...)
```

codeJack:
Thanks @TirTech for raising this one. 

@pahud to complement what has been shared above around the expected CDK experience with `AccessConfig`.

[...]

Issue #8276, opened 2020-05-29
attribute error when adding SA to existing EKS cluster


andrea-spoldi:
I'm trying to add a SA following the guide here: 

https://docs.aws.amazon.com/cdk/api/latest/docs/aws-eks-readme.html#service-accounts

to an existing EKS cluster deployed with CDK (version 1.15 and 1.16, both tested)

Issuing "cdk deploy" an "attribute error" is raised on Custom::AWSCDKOpenIdConnectProvider action

### Reproduction Steps

simply adding: 
const sa = cluster.addServiceAccount('MyServiceAccount');

to an existing EKS cdk stack 

### Error Log

`10/33 | 1:30:10 PM | CREATE_FAILED        | Custom::AWSCDKOpenIdConnectProvider   | EksStack/eks/OpenIdConnectProvider/Resource/Default (eksOpenIdConnectProviderA19AE8AF) CustomResource attribute error: Vendor response doesn't contain OpenIdConnectIssuerUrl key in object arn:aws:cloudformation:us-east-1:<account-id>:stack/*****-eks/74463170-7f06-11ea-8a97-12217e59f19f|eksB49B8EA3|d4bd0a87-4326-4ed2-b473-99471c6b71e5 in S3 bucket cloudformation-custom-resource-storage-useast1
	new CustomResource (/usr/src/packages/infrastructure/node_modules/@aws-cdk/core/lib/custom-resource.ts:115:21)
	\_ new OpenIdConnectProvider (/usr/src/packages/infrastructure/node_modules/@aws-cdk/aws-iam/lib/oidc-provider.ts:120:22)
	\_ Cluster.get openIdConnectProvider [as openIdConnectProvider] (/usr/src/packages/infrastructure/node_modules/@aws-cdk/aws-eks/lib/cluster.ts:663:37)
	\_ new ServiceAccount (/usr/src/packages/infrastructure/node_modules/@aws-cdk/aws-eks/lib/service-account.ts:72:58)
	\_ Cluster.addServiceAccount (/usr/src/packages/infrastructure/node_modules/@aws-cdk/aws-eks/lib/cluster.ts:725:12)
	\_ new EksResourcesStack (/usr/src/packages/infrastructure/lib/eks-resources-stack.ts:60:24)
	\_ Object.<anonymous> (/usr/src/packages/infrastructure/bin/infrastructure.ts:45:19)
	\_ Module._compile (internal/modules/cjs/loader.js:1133:30)
[...]

PR #8317, opened 2020-06-02
feat(eks): expose cluster security group and encryption configuration

eduardomourar:
This PR will have the EKS Cluster construct expose [**ClusterSecurityGroupId**](https://docs.aws.amazon.com/eks/latest/APIReference/API_VpcConfigResponse.html#AmazonEKS-Type-VpcConfigResponse-clusterSecurityGroupId) (ID of Security group that was created by Amazon EKS for the cluster) and [**EncryptionConfigKeyArn**](https://docs.aws.amazon.com/eks/latest/APIReference/API_Provider.html#AmazonEKS-Type-Provider-keyArn) (ARN of the customer master key used in the encryption configuration for the cluster) attributes for both custom resource and native CloudFormation option.

This also fixes #8276 in the following way: if a custom resource returns an attribute with an "undefined" value, CFN will fail with a "vendor response doesn't contain key" error. To avoid this, we return empty strings in case an attribute is undefined. This is also true for when adding new attributes, in which case updating to the new version will fail on previously deployed clusters with the same error. To mitigate this (and fix #8276 along the way), we add a fake property called "AttributesRevision" with a number that needs to be manually incremented every time new attributes are introduced. This will cause old clusters to be updated and the new attributes returned.

Closes #8236 

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30016, opened 2024-04-30
feat(eks): support AccessConfig for the Cluster construct

pahud:
###  Background

Amazon EKS originally uses `ConfigMap` as its access management and in aws-eks we use AwsAuth to leverage the kubectl from kubectl-lambda-layer to create the AwsAuth configmap for that. The ConfigMap has been very difficult to maintain due to its lack support of EKS API but thanks to the AwsAuth class, it's been very smooth in CDK.

In AWS reInvent 2023 we [announced](https://aws.amazon.com/blogs/containers/a-deep-dive-into-simplified-amazon-eks-access-management-controls/) the access API support that simplifies the management as a replacement of the traditional ConfigMap. In CloudFormation we have the [AccessConfig](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-eks-cluster.html#cfn-eks-cluster-accessconfig) with [AuthenticationMode](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-eks-cluster-accessconfig.html#cfn-eks-cluster-accessconfig-authenticationmode) and `BootstrapClusterCreatorAdminPermissions` now.

The `AuthenticationMode` supports `CONFIG_MAP`, which is the default, `API_AND_CONFIG_MAP` and `CONFIG_MAP`. It allows users to switch the mode on cluster creation or update. When the mode has API support, users have to define the `AccessEntry` to map the access policies and the IAM principals. This PR introduces the `AccessEntry` and `AccessPolicy` classes for that to simplify it with similar experience just as the [iam.ManagedPolicy ](https://github.com/aws/aws-cdk/blob/3928eae1ee92a03ba9959288f05f59d6bd5edcba/packages/aws-cdk-lib/aws-iam/lib/managed-policy.ts#L104)class. This PR also introduces the `grantAccess()` method that allows a cluster to `grant` its access to a specific principal and abstracts away the complexity.

Overview of the API experience from this PR:

```ts
const cluster = new eks.Cluster(this, 'Cluster', {
  vpc,
  mastersRole: clusterAdminRole,
  version: eks.KubernetesVersion.V1_30,
  kubectlLayer: new KubectlV29Layer(this, 'KubectlLayer'),
  authenticationMode: eks.AuthenticationMode.API_AND_CONFIG_MAP,
});

// Cluster Admin role for this cluster
cluster.grantAccess('clusterAdminAccess', clusterAdminRole.roleArn, [
  eks.AccessPolicy.fromAccessPolicyName('AmazonEKSClusterAdminPolicy', {
    accessScopeType: eks.AccessScopeType.CLUSTER,
  }),
]);

// EKS Admin role for specified namespaces of thie cluster
cluster.grantAccess('eksAdminRoleAccess', eksAdminRole.roleArn, [
  eks.AccessPolicy.fromAccessPolicyName('AmazonEKSAdminPolicy', {
    accessScopeType: eks.AccessScopeType.NAMESPACE,
    namespaces: ['foo', 'bar'],
  }),
]);

// EKS Admin Viewer role for specified namespaces of thie cluster
cluster.grantAccess('eksAdminViewRoleAccess', eksAdminViewRole.roleArn, [
  eks.AccessPolicy.fromAccessPolicyName('AmazonEKSAdminViewPolicy', {
    accessScopeType: eks.AccessScopeType.NAMESPACE,
    namespaces: ['foo', 'bar'],
  }),
]);
```

### Issue # (if applicable)

Closes  https://github.com/aws/aws-cdk/issues/28588

This PR introduces the `authenticationMode`, `AccessEntry` and `AccessPolicy` for both `Cluster` and `FargateCluster` construct.

- [x] bump `@aws-sdk/client-eks` to [v3.476.0](https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.476.0)(the minimal version with EKS Cluster Access Management support)
- [x] make sure it deploys with the new AccessConfig support for a new cluster
- [x] make sure an existing cluster can update by adding this new prop
- [x] make sure it deploys with a new FargateCluster
- [x] make sure an existing FargateCluster can update by adding this new prop
- [x] make sure it works with CfnAccessEntry L1 resources
- [x] AccessEntry L2 construct support
- [x] AccessPolicy class
- [x] bootstrapClusterCreatorAdminPermissions
- [x] unit tests
- [x] integ tests
- [x] update README
- [x] add PR notes

### Notes

1. Switching authentication modes on an existing cluster is a one-way operation like:

undefined(CONFIG_MAP) -> API_AND_CONFIG_MAP -> API

You can switch from undefined or CONFIG_MAP to API_AND_CONFIG_MAP. You can then switch from API_AND_CONFIG_MAP to API. You cannot revert these operations in the opposite direction. Meaning you cannot switch back to CONFIG_MAP or API_AND_CONFIG_MAP from API. And you cannot switch back to CONFIG_MAP from API_AND_CONFIG_MAP. (see [here](https://aws.amazon.com/blogs/containers/a-deep-dive-into-simplified-amazon-eks-access-management-controls/)) This PR adds relevant checks in the custom resource and add docstring in the `authenticationMode` prop.

2. Switching `bootstrapClusterCreatorAdminPermissions` would cause cluster replacement, we callout in the README and construct prop docstring as a headsup. This option is [available](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-eks-cluster-accessconfig.html#cfn-eks-cluster-accessconfig-bootstrapclustercreatoradminpermissions) in CFN which triggers replacement on resource update as well. I have created #30107 for further improvement.

3. This feature does not support AWS China regions at this moment as the JS SDK version of lambda node18 runtime in China regions is `3.462.0` while this feature requires SDK [3.476.0](https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.476.0) or above. It's `3.552.0` in us-east-1. Use [this example](https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html#nodejs-sdk-included) to check the version.

### Checklist

----


packages/aws-cdk-lib/aws-eks/README.md
```diff
@@ -1100,6 +1103,104 @@ consoleReadOnlyRole.addToPolicy(new iam.PolicyStatement({
...
+  vpc,
+  version: eks.KubernetesVersion.V1_29,
+  kubectlLayer: new KubectlV29Layer(this, 'KubectlLayer'),
+  authenticationMode: eks.AuthenticationMode.API_AND_CONFIG_MAP,
+});
+```
+
+> **Note** - Switching authentication modes on an existing cluster is a one-way operation. You can switch from
+> `CONFIG_MAP` to `API_AND_CONFIG_MAP`. You can then switch from `API_AND_CONFIG_MAP` to `API`.
+> You cannot revert these operations in the opposite direction. Meaning you cannot switch back to
```
GavinZZ, line 1129: For my learning purposes, what if I switch it in opposite direction? Is it going to cause a deployment failure or a resoruce replacement?

packages/aws-cdk-lib/aws-eks/lib/access-entry.ts
```diff
@@ -0,0 +1,340 @@
...
+  constructor(scope: Construct, id: string, props: AccessEntryProps ) {
+    super(scope, id);
+
+    this.cluster = props.cluster;
+    this.principal = props.principal;
+    this.accessPolicies = props.accessPolicies;
+
+    const resource = new CfnAccessEntry(this, 'Resource', {
+      clusterName: this.cluster.clusterName,
+      principalArn: this.principal,
```
GavinZZ, line 313: Are we going to allow users to customize `type` on access entry? Or we don't plan to support it on first launch but can be an added feature in the future?

packages/aws-cdk-lib/aws-eks/lib/access-entry.ts
```diff
@@ -0,0 +1,340 @@
...
+      accessPolicies: Lazy.any({
```
GavinZZ, line 314: For my learning purposes, the reason that we use `Lazy` here is because we have function below `addAccessPolicies` where users can add additional access policy outside of the constructor. So we make it lazy to make sure the conversion happens lazily?

packages/aws-cdk-lib/aws-eks/lib/aws-auth.ts
```diff
@@ -31,6 +31,13 @@ export class AwsAuth extends Construct {
   constructor(scope: Construct, id: string, props: AwsAuthProps) {
     super(scope, id);
 
+    // Throw if the cluster does not support ConfigMap.
+    const supportConfigMap = props.cluster.authenticationMode !== AuthenticationMode.API ? true : false;
+
+    if (!supportConfigMap) {
+      throw new Error('ConfigMap not supported in the AuthenticationMode');
```
GavinZZ, line 38: nit: can we give better messaging on why `ConfigMap` is not supported?

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
```diff
@@ -1677,12 +1737,23 @@ export class Cluster extends ClusterBase {
...
     // do not create a masters role if one is not provided. Trusting the accountRootPrincipal() is too permissive.
     if (props.mastersRole) {
       const mastersRole = props.mastersRole;
 
-      // map the IAM role to the `system:masters` group.
-      this.awsAuth.addMastersRole(mastersRole);
+      // if we support authentication API we create access entry
+      if (supportAuthenticationApi) {
+        this.grantAccess('mastersRoleAccess', props.mastersRole.roleArn, [
+          AccessPolicy.fromAccessPolicyName('AmazonEKSClusterAdminPolicy'),
```
GavinZZ, line 1750: Is it a good idea to add Admin Policy by default?

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
```diff
@@ -1716,6 +1787,47 @@ export class Cluster extends ClusterBase {
...
+   *
+   * This method creates an `AccessEntry` construct that grants the specified IAM principal the access permissions
+   * defined by the provided `IAccessPolicy` array. This allows the IAM principal to perform the actions permitted
+   * by the access policies within the EKS cluster.
+   *
+   * @param id - The ID of the `AccessEntry` construct to be created.
+   * @param principal - The IAM principal (role or user) to be granted access to the EKS cluster.
+   * @param accessPolicies - An array of `IAccessPolicy` objects that define the access permissions to be granted to the IAM principal.
+   */
+  public grantAccess(id: string, principal: string, accessPolicies: IAccessPolicy[]) {
```
GavinZZ, line 1828: Since we have `grantAccess` which also calls `addToAccessEntry`, do you think it makes sense to make `addToAccessEntry` private and let users use `grantAccess` directly? (Instead of providing two fucntions that do the same thing).

GavinZZ (commented):
Great work Pahud! Thanks for doing this. FYI I only reviewed `aws-cdk-lib/aws-eks/lib/*` file changes. Will review the rest in a next review. Some of my questions are for my own learning purposes.

packages/aws-cdk-lib/aws-eks/README.md
pahud: @GavinZZ Thank you for bringing it up.

Amazon EKS does not allow you to turn off the `API` support after you opt in or re-enable the ConfigMap after you disabled it. That is,

❌  `API_AND_CONFIG_MAP` to `CONFIG_MAP` -> not allowed as you are turning off the API support
❌  `API` to `CONFIG_MAP` -> not allowed as you are turning off the API support and also you are re-enabling the config map from disabled.
❌  `API` to `API_AND_CONFIG_MAP` -> not allowed as you are enabling the ConfigMap from disabled.
❌  `API_AND_CONFIG_MAP` to `CONFIG_MAP` or `undefined` -> not allowed as you are disabling API from enabled.
❌  Any defined value to undefined -> not allowed, EKS API does not allow you to switch back once you opt in.

This PR essentially blocks all those cases in the CDK code but we can only detect that in the custom resource so users still have to deploy it through the cloudformation and the custom resource would throw. 

When this happen:

1. no resource would be replaced
2. no SDK calls would happen because CDK custom resource would throw first

packages/aws-cdk-lib/aws-eks/lib/access-entry.ts
pahud: Yes Lazy is required if we need to evaluate and render an property on synthesize stage.

https://github.com/pahud/aws-cdk/blob/ceac30ac7355f712a1c96102464102afd4b358be/packages/aws-cdk-lib/aws-eks/lib/access-entry.ts#L314-L322

If we just set

```
accessPolicies: this.accessPolicies
```

then we won't be allowed to add more policies with the [addAccessPolicies](https://github.com/pahud/aws-cdk/blob/ceac30ac7355f712a1c96102464102afd4b358be/packages/aws-cdk-lib/aws-eks/lib/access-entry.ts#L336) method because that property would be determined on construct instantiation.

With Lazy, we could build the experience like

```ts
const entry = new AccessEntry(scope, id, {
  accessPolicies: myInitialAccessPolicy
})
entry.addAccessPolicies(...); // add additional policies on it
entry.addAccessPolicies(...); // add even more policies 
```

The Lazy function will render the final value for the L1 construct.

We have similar implementation in ECS [TaskDefinition](https://github.com/aws/aws-cdk/blob/205163fc0d2cac84d3d746a98c393e137f0e2388/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts#L485-L517) that renders the final container definition so our task definition allows us to keep adding containers like this and eventually render the final one for the L1.

```ts
task.addContainer(); // my first container def
task.addContainer(); // my second container def
task.addContainer(); // my 3rd container def
```

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
```diff
@@ -1677,12 +1737,23 @@ export class Cluster extends ClusterBase {
...
+
     // do not create a masters role if one is not provided. Trusting the accountRootPrincipal() is too permissive.
     if (props.mastersRole) {
       const mastersRole = props.mastersRole;
 
-      // map the IAM role to the `system:masters` group.
-      this.awsAuth.addMastersRole(mastersRole);
+      // if we support authentication API we create access entry
+      if (supportAuthenticationApi) {
+        this.grantAccess('mastersRoleAccess', props.mastersRole.roleArn, [
```
shikha372, line 1749: what would happen if this mastersRole is not provided in the props?

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
pahud: In current aws-eks implementation, if `props.mastersRole` is defined, we add that IAM principal to AwsAuth as admin role using [addMastesrRole method](https://github.com/aws/aws-cdk/blob/205163fc0d2cac84d3d746a98c393e137f0e2388/packages/aws-cdk-lib/aws-eks/lib/aws-auth.ts#L64), which essentially create the classic ConfigMap with that `props.mastersRole` as `system:masters`, and that is essentially what we'd like it to happen using the new AccessEntry approach - to grant that principal the admin policy

https://github.com/aws/aws-cdk/blob/205163fc0d2cac84d3d746a98c393e137f0e2388/packages/aws-cdk-lib/aws-eks/lib/cluster.ts#L1680-L1685

But you are right, it's not clear if we should grant it `AmazonEKSAdminPolicy` or `AmazonEKSClusterAdminPolicy` and I can't find it in our doc.

This is what Amazon Q told me:

image

I'll cut an internal ticket to the team to get this clarified.

shikha372: I guess masters role on the K8s cluster is meant for granting the cluster full access, but i'm not confident about the difference between `AmazonEKSClusterAdminPolicy` and `AmazonEKSAdminPolicy`, and which one to use here.

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
pahud: Yes this makes perfect sense. Will fix that.

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
```diff
@@ -1716,6 +1787,47 @@ export class Cluster extends ClusterBase {
...
+   * If no entry exists for the given principal, it creates a new access entry with the provided access policies.
+   *
+   * @param principal - The principal (e.g., IAM user or role) for which the access entry is being added.
+   * @param policies - An array of access policies to be associated with the principal.
+   *
+   * @throws {Error} If the uniqueName generated for the new access entry is not unique.
+   *
+   * @returns {void}
+   */
+  public addToAccessEntry(id: string, principal: string, policies: IAccessPolicy[]) {
```
shikha372, line 1803: can this be a `private` function if we customers are using grantaccess()?

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
pahud: At current implementation, if `props.mastersRole` is not provided, aws-eks would NOT add additional iam principal as `system:masters` to the ConfigMap and only the cluster creator role would be `system:masters`. Now if you need to administer the cluster using a common IAM role with `kubectl`, you will need to assume to that cluster creator role and that could be a permission concern. The reason `prop.mastersRole` exists is that it allow users to pass any existing role to that property and aws-eks would add that into its ConfigMap with `system:masters` permission. If undefined, no additional ConfigMap entry would be added.

Now, with the introduction of access entry with access api, we essentially keep this support but we have to determine the AuthenticationMode in the following conditions:

1. AuthenticationMode is `API` - this means ConfigMap not supported anymore and we should not do this.cluster.awsAuth.addMasters() to that, instead, we should create AccessEntry for the passed mastersRole.
2. AuthenticationMode is `API_AND_CONFIG_MAP` - this means both are supported. This PR would leave everything as before and continue update the ConfigMap using awsAuth.addMastersRole().
3. 2. AuthenticationMode is `CONFIG_MAP`  - same as option 2.

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
pahud: yes will fix it!

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
pahud: `AmazonEKSClusterAdminPolicy` should have higher privileges than `AmazonEKSAdminPolicy` as it's cluster-wise while `AmazonEKSAdminPolicy` is per namespace(s). I'll cut a ticket to get it sorted.

packages/aws-cdk-lib/aws-eks/lib/aws-auth.ts
```diff
@@ -31,6 +31,13 @@ export class AwsAuth extends Construct {
   constructor(scope: Construct, id: string, props: AwsAuthProps) {
     super(scope, id);
 
+    // Throw if the cluster does not support ConfigMap.
+    const supportConfigMap = props.cluster.authenticationMode !== AuthenticationMode.API ? true : false;
```
shikha372, line 35: if authentication mode is AuthenticationMode.API_AND_CONFIG_MAP would this be true or false ?

packages/aws-cdk-lib/aws-eks/lib/access-entry.ts
```diff
@@ -0,0 +1,340 @@
...
+ */
+export class AccessPolicy implements IAccessPolicy {
+  /**
+   * Import AccessPolicy by name.
+   */
+  public static fromAccessPolicyName(policyName: string, options?: AccessPolicyNameOptions): IAccessPolicy {
+    class Import implements IAccessPolicy {
+      public readonly policy = `arn:${Aws.PARTITION}:eks::aws:cluster-access-policy/${policyName}`
+      public readonly accessScope: AccessScope = options?.accessScope ?? {
+        type: policyName === 'AmazonEKSClusterAdminPolicy' ? AccessScopeType.CLUSTER : AccessScopeType.NAMESPACE,
```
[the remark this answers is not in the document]
pahud: Thank you. I am checking this with the EKS team.

Internal tracking: V1385204928

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
pahud: Internal Tracking: V1385204928

packages/aws-cdk-lib/aws-eks/lib/access-entry.ts
pahud: The `type`(eks.AccessScopeType) is per AccessPolicy which can be defined this way and attach to an AccessEntry.

```ts
const clusterScopedClusterAdminPolicy = AccessPolicy.fromAccessPolicyName('AmazonEKSClusterAdminPolicy', {
  accessScope: { type: eks.AccessScopeType.CLUSTER },
});
const namespaceScopedAdminPolicy = AccessPolicy.fromAccessPolicyName('AmazonEKSAdminPolicy', {
  accessScope: { type: eks.AccessScopeType.NAMESPACE, namespaces: ['foo'] },
});

new AccessEntry(stack, 'Entry', {
  cluster,
  accessPolicies: [
    clusterScopedClusterAdminPolicy,
    namespaceScopedAdminPolicy,
  ],
  principal: clusterAdminRole.roleArn,
});
```

We have an internal ticket(V1385204928) to check if all policies support both `Cluster` and `Namespaces` AccessScopeType.

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
pahud: internal tracking: V1385204928

packages/aws-cdk-lib/aws-eks/lib/aws-auth.ts
pahud: DONE. Just improved the doc string.

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
pahud: DONE

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
pahud: DONE

packages/aws-cdk-lib/aws-eks/lib/access-entry.ts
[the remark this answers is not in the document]
pahud: internal checking: V1385204928

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
shikha372: Understood, Thank you @pahud for explaining it well

richardgroves:
@pahud is there anything that can be done to encourage getting this merged in? We really need it in our project to be able to setup eks clusters properly in one phase.

packages/aws-cdk-lib/aws-eks/README.md
```diff
@@ -1100,6 +1103,115 @@ consoleReadOnlyRole.addToPolicy(new iam.PolicyStatement({
...
+});
+
+// Cluster Admin role for this cluster
+cluster.grantAccess('clusterAdminAccess', clusterAdminRole.roleArn, [
+  eks.AccessPolicy.fromAccessPolicyName('AmazonEKSClusterAdminPolicy', {
+    accessScopeType: eks.AccessScopeType.CLUSTER,
+  }),
+]);
+
+// EKS Admin role for specified namespaces of thie cluster
```
nnmin-aws, line 1197: this

packages/aws-cdk-lib/aws-eks/README.md
```diff
@@ -1100,6 +1103,115 @@ consoleReadOnlyRole.addToPolicy(new iam.PolicyStatement({
...
+cluster.grantAccess('eksAdminRoleAccess', eksAdminRole.roleArn, [
+  eks.AccessPolicy.fromAccessPolicyName('AmazonEKSAdminPolicy', { 
+    accessScopeType: eks.AccessScopeType.NAMESPACE,
+    namespaces: ['foo', 'bar'],
+  }),
+]);
+
+// EKS Admin Viewer role for specified namespaces of thie cluster
```
nnmin-aws, line 1205: nit: this cluster

packages/aws-cdk-lib/aws-eks/lib/cluster-resource.ts
```diff
@@ -96,7 +98,7 @@ export class ClusterResource extends Construct {
         // doesn't contain XXX key in object" (see #8276) by incrementing this
         // number, you will effectively cause a "no-op update" to the cluster
         // which will return the new set of attribute.
-        AttributesRevision: 2,
+        AttributesRevision: 3,
```
nnmin-aws, line 101: for my learning, do we bump Rv each time when there is new attribute added?

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
```diff
@@ -809,6 +822,16 @@ export interface ClusterProps extends ClusterOptions {
...
 
+  /**
+   * Whether or not IAM principal of the cluster creator was set as a cluster admin access entry
+   * during cluster creation time.
+   *
+   * Changing this value after the cluster has been created will result in the cluster being replaced.
+   *
+   * @default true
+   */
+  readonly bootstrapClusterCreatorAdminPermissions?: boolean;
```
nnmin-aws, line 833: can this value be change by user?

packages/aws-cdk-lib/aws-eks/lib/cluster.ts
```diff
@@ -1677,12 +1737,26 @@ export class Cluster extends ClusterBase {
...
+
     // do not create a masters role if one is not provided. Trusting the accountRootPrincipal() is too permissive.
     if (props.mastersRole) {
       const mastersRole = props.mastersRole;
 
-      // map the IAM role to the `system:masters` group.
-      this.awsAuth.addMastersRole(mastersRole);
+      // if we support authentication API we create an access entry for this mastersRole
+      // with cluster scope.
+      if (supportAuthenticationApi) {
```
nnmin-aws, line 1749: if bootstrapClusterCreatorAdminPermissions is false, will it still create the CAM entry for cluster creator role?

packages/aws-cdk-lib/aws-eks/README.md
```diff
@@ -1100,6 +1103,115 @@ consoleReadOnlyRole.addToPolicy(new iam.PolicyStatement({
 cluster.awsAuth.addMastersRole(consoleReadOnlyRole)
 ```
 
+### Access Config
```
nnmin-aws, line 1106: can we introduce the usage for bootstrapClusterCreatorAdminPermissions in Access Config? and its impact on whether the access entry for master role will be created or not.

pahud:
>@pahud is there anything that can be done to encourage getting this merged in? We really need it in our project to be able to setup eks clusters properly in one phase.

Thank you @richardgroves  I appreciate your concern. I will make another iteration this week and get the CDK team and EKS team review again. Hopefully to clarify some details and move it forward.

packages/aws-cdk-lib/aws-eks/lib/cluster-resource.ts
pahud: I believe so. It was introduced in this [PR](https://github.com/aws/aws-cdk/pull/8317).


>This also fixes https://github.com/aws/aws-cdk/issues/8276 in the following way: if a custom resource returns an attribute with an "undefined" value, CFN will fail with a "vendor response doesn't contain key" error. To avoid this, we return empty strings in case an attribute is undefined. This is also true for when adding new attributes, in which case updating to the new version will fail on previously deployed clusters with the same error. To mitigate this (and fix https://github.com/aws/aws-cdk/issues/8276 along the way), we add a fake property called "AttributesRevision" with a number that needs to be manually incremented every time new attributes are introduced. This will cause old clusters to be updated and the new attributes returned.

packages/aws-cdk-lib/aws-eks/README.md
```diff
@@ -1103,6 +1106,134 @@ consoleReadOnlyRole.addToPolicy(new iam.PolicyStatement({
...
+1. `undefined` ->  `API_AND_CONFIG_MAP` -> `API`
+2. `undefined` -> `API` (as `undefined` is equivalent `API_AND_CONFIG_MAP`)
+
+If you have explicitly declared `AwsAuth` resources and then try to switch to the `API` mode, which no longer supports the
+`ConfigMap`, AWS CDK will throw an error as a protective measure to prevent you from losing all the access entries in the `ConfigMap`. In this case, you will need to remove all the declared `AwsAuth` resources explicitly before you are allowed to transition to the
+`API` mode.
+
+> **Note** - This is a one-way transition. Once you switch to the `API` mode,
+you will not be able to switch back. Therefore, it is crucial to ensure that you have defined all the necessary access entries before making the switch to the `API` mode.
+
```
pahud, line 1236: Just added this section as a heads-up for migration. cc @nnmin-aws

GavinZZ (approved):
LGTM, thanks for the hardwork, discussion with EKS team and getting this merged!

GavinZZ:
Temporarily adding a `do-not-merge` label to confirm with team that there's no outstanding feedback. Will remove it and merge the PR if none.

shikha372 (approved):

pahud:
Obviously I did update the snaps but the PR Linter didn't think so.

Exemption Request just in case.

PR #29670, opened 2024-04-01
feat(ecs): support `pidMode` for `FargateTaskDefinition`

lpizzinidev:
### Issue # (if applicable)

Closes #29619.

### Reason for this change

Support [`pidMode`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-pidmode) for `FargateTaskDefinition`.

### Description of changes

Added support for the `pidMode` property along with the necessary validation, documentation, and test coverage.

### Description of how you validated changes

- [x] Unit tests
- [x] Integration tests

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30020, opened 2024-04-30
fix(ecs): require task pidMode for Linux-based Fargate tasks, not host

### LINKED ISSUE #29995: aws_ecs.FargateTaskDefinition: pidmode Task unsupported
### Describe the bug

Reading the pidmode announcement [here](https://aws.amazon.com/about-aws/whats-new/2023/08/aws-fargate-pid-namespace-sharing-kernel-configuration/), the cloudformation description for [pidMode here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-pidmode). It explains `task` is the only option. But currently the [FargateTaskDefinition](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ecs/lib/fargate/fargate-task-definition.ts#L174) only allows `host`.

Changing this to host will result in a failed deployment, because host isn't allowed on Fargate (linux) tasks.

### Expected Behavior

Consistency between documentation and usable options reflected by synth/deployments in CDK.
Task is the one that is the only one actually supported.

### Current Behavior

> Error: 'pidMode' can only be set to 'host' for Fargate containers, got: 'task'.
    at new FargateTaskDefinition (C:\projects\push-gateway-cdk\node_modules\aws-cdk-lib\aws-ecs\lib\fargate\fargate-task-definition.js:1:2562)

But using the host variant will results in:
> Deployment failed: Error: The stack named XXXX failed to deploy: UPDATE_ROLLBACK_COMPLETE: Resource handler returned message: "Invalid request provided: Create TaskDefinition: Tasks using the Fargate launch type do not support pidMode 'host'. The supported value for pidMode is 'task'.

### Reproduction Steps

```typescript
const ftdTask = new ecs.FargateTaskDefinition(stack, `task-that-will-fail-to-synth`, {
    cpu: 1024,
    memoryLimitMiB: 2048,
    pidMode: PidMode.TASK,
});
```
  
```typescript
const ftdHost = new ecs.FargateTaskDefinition(stack, `task-that-will-fail-to-deploy`, {
    cpu: 1024,
    memoryLimitMiB: 2048,
    pidMode: PidMode.HOST,
});
```

### Possible Solution

Switch [the checks](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ecs/lib/fargate/fargate-task-definition.ts#L174) from pidMode.HOST to pidMode.TASK.

_No response_

### CDK CLI Version

2.139.0

### Framework Version

_No response_

### Node.js Version

v20.12.0

### OS

windows

### Language

TypeScript

### Language Version

_No response_

### Other information

(OS is windows for where I ran CDK, the containers are linux as those are the only one allowing pidMode anyway)

CharlesHolmes:
### Issue # (if applicable)

Closes #29995.

### Reason for this change

Only the `task` option is allowed for [`pidMode`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-pidmode) on Linux-based Fargate tasks.

### Description of changes

This PR builds on the changes introduced in #29670 but fixes the handling of `pidMode` so that it matches the behavior allowed by CloudFormation and described in the [AWS User Guide](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-pidmode).

Updated the existing tests so that `task` is the only allowable `pidMode` setting if a Fargate task's OS is Linux-based.

### Checklist

----


vduits:
Appreciate the fix. I see you also [added a check for it being a linux os family](https://github.com/aws/aws-cdk/pull/30020/files#diff-56d2a0111b6433f90eea5f9e8d984de785ddbb0731b5e761f4e7dc0caa3f87d3R176), which makes sense being the alternative already being present for denying pidmode usage. Though that does change the behavior, where it will now only warn when you set an OS family, I think that is a good thing as it's more consistent to only message on knowns.

It does however mean that without an os family you have to will wait until deployment to notice your mistake. If we want a synth level protection, then it would be better to also lock pidmode behind having set an os family. This new flow means that if you set `PidMode.HOST` on a non os family tagged resource, it will synth but fail on deployment.

CharlesHolmes:
> If we want a synth level protection, then it would be better to also lock pidmode behind having set an os family.

I like this because it only affects people who are already using the L2 pidMode (which was introduced only a few days ago) and only if they're not specifying an OS family.  I'll put together the change.

packages/aws-cdk-lib/aws-ecs/lib/fargate/fargate-task-definition.ts
```diff
@@ -168,11 +170,16 @@ export class FargateTaskDefinition extends TaskDefinition implements IFargateTas
     }
 
     if (props.pidMode) {
+      if (!props.runtimePlatform?.operatingSystemFamily) {
+        throw new Error('Specifying \'pidMode\' requires that operating system family also be provided.');
+      }
       if (props.runtimePlatform?.operatingSystemFamily?.isWindows()) {
         throw new Error('\'pidMode\' is not supported for Windows containers.');
       }
```
xazhao, line 178: nit: I think what we care here is it should be linux. So we can move the `isLinux()` check up here and update the error message.
```
if (!props.runtimePlatform?.operatingSystemFamily?.isLinux())
```

packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
```diff
@@ -194,8 +194,10 @@ export interface TaskDefinitionProps extends CommonTaskDefinitionProps {
    * The process namespace to use for the containers in the task.
    *
    * Only supported for tasks that are hosted on AWS Fargate if the tasks
-   * are using platform version 1.4.0 or later (Linux).
-   * Not supported in Windows containers.
+   * are using platform version 1.4.0 or later (Linux). Only the TASK option
+   * is supported for Linux-based Fargate containers. Not supported in Windows
+   * containers. If pidMode is specified for a Fargate task, then
+   * runtimePlatform.operatingSystemFamily must also be specified.
    *
```
xazhao, line 201: nit: we can put the doc link here for reference.

xazhao (commented):
Agree that it's not a breaking change because the deployment would fail with wrong pidMode value before. Left some nit comments

packages/aws-cdk-lib/aws-ecs/lib/fargate/fargate-task-definition.ts
CharlesHolmes: From what I understand, that check was put there to verify that the user isn't trying to set pidMode on a Windows container.  The documentation says it's not supported.  Given that, do you think it would be worth keeping?

image

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_definition_pidmode

packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
CharlesHolmes: Sure!  I added a link to it.

packages/aws-cdk-lib/aws-ecs/lib/fargate/fargate-task-definition.ts
xazhao: Got it. I'm fine keeping the check here.

xazhao (approved):

PR #30022, opened 2024-04-30
fix(core): stack tags should not allow tokens

### LINKED ISSUE #29424: core: Stack Construct: when set stack-level tags with pseudo parameters such as Class Aws.STACK_NAME should emit error
### Describe the bug

In previous closed issue "core: improve experience around stack-level tagging"
https://github.com/aws/aws-cdk/issues/4134

Should have feature "We should emit an error if you try to apply a tag that uses tokens to the Stack itself."

However, when use Stack Class to work on tags, example Python code
```
        bucket = s3.Bucket(self, "testCDKBucketwithImportedValuejiemying",
                           removal_policy= RemovalPolicy.DESTROY,
                           auto_delete_objects= True
                           )
        
        Stack.of(self).tags.set_tag("StackNameViaAws",Aws.STACK_NAME)
``` 
CDK ``` ls and synth ```do not emit the error, only when ```cdk deploy``` the expected invalid token value error will be incurred during stack operation when trying to tag resources in the scope:
```
Deployment failed: Error: The stack named CdkCaseworkingPythonStack failed to deploy: UPDATE_ROLLBACK_COMPLETE: Resource handler returned message: "1 validation error detected: Value '${Token[AWS.StackName.13]}' at 'tags.2.member.value' failed to satisfy constraint: Member must satisfy regular expression pattern: [\p{L}\p{Z}\p{N}_.:/=+\-@]* (Service: Iam, Status Code: 400, Request ID: df4a5550-9d06-4cc0-bd1d-46331da9858e)" (RequestToken: ce3f7dc0-4937-80e1-4b98-c639edead1b0, HandlerErrorCode: GeneralServiceException)
```

### Expected Behavior

When user pass pseudo parameters such as Aws.STACK_NAME, and Aws.REGION to Stack Construct tags. An error should be given due to https://github.com/aws/aws-cdk/issues/4106#issuecomment-532761271

### Current Behavior

CDK ls and synth does not give error. 

The Stack construct will pass {Token[AWS.StackName.13]} as stack-level tag value when CDK create-changeset, subsequently causing error when the stack level tags propagate to resources inside of the stack. 

### Reproduction Steps

1. Create a tag-able resource in a CDK stack and set tags value with pseudo parameter Class Aws

```
        bucket = s3.Bucket(self, "testCDKBucketwithImportedValuejiemying",
                           removal_policy= RemovalPolicy.DESTROY,
                           auto_delete_objects= True
                           )
        
        Stack.of(self).tags.set_tag("StackNameViaAws",Aws.STACK_NAME)
``` 

2. cdk deploy

### Possible Solution

We should emit an error if you try to apply a tag that uses tokens to the Stack itself

_No response_

### CDK CLI Version

2.132.0

### Framework Version

_No response_

### Node.js Version

v20.11.1

### OS

MacOS

### Language

Python

### Language Version

_No response_

### Other information

_No response_

comcalvi:
### Issue # (if applicable)

Closes #29424.

### Reason for this change

Stack tags get applied to every resource they contain. Tokens must be resolved by CloudFormation itself and the CDK CLI has to apply stack tags to the stack before the deployment starts. This results in the string representation of the token (eg `Token[AWS.AccountId.6]`) making its way into the stack tags. CFN applies these tags automatically to every resource, but this string value contains characters that many resource handlers do not allow in tags. This causes a deployment error.

### Description of changes

Move this error to synthesis time by checking for tokens before adding the tags. 

Unit tests. 

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/core/lib/tag-manager.ts
+++ b/packages/aws-cdk-lib/core/lib/tag-manager.ts
@@ -2,6 +2,7 @@ import { TagType } from './cfn-resource';
 import { CfnTag } from './cfn-tag';
 import { Lazy } from './lazy';
 import { IResolvable } from './resolvable';
+import { Token } from './token';
 
 const TAG_MANAGER_SYM = Symbol.for('@aws-cdk/core.TagManager');
 
@@ -361,6 +362,9 @@ export class TagManager {
   public setTag(key: string, value: string, priority = 0, applyToLaunchedInstances = true): void {
     // This method mostly exists because we don't want to expose the 'Tag' type used (it will be confusing
     // to users).
+    if (Token.isUnresolved(key) || Token.isUnresolved(value)) {
+      throw new Error(`Stack tags cannot include unresolved tokens, found a tag with key '${key}' and value '${value}'`);
+    }
     this._setTag({ key, value, priority, applyToLaunchedInstances });
   }
 
--- a/packages/aws-cdk-lib/core/test/stack.test.ts
+++ b/packages/aws-cdk-lib/core/test/stack.test.ts
@@ -14,6 +14,7 @@ import {
   PERMISSIONS_BOUNDARY_CONTEXT_KEY,
   Aspects,
   Stage,
+  Token,
 } from '../lib';
 import { Intrinsic } from '../lib/private/intrinsic';
 import { resolveReferences } from '../lib/private/refs';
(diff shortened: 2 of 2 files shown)


comcalvi:
Closing this because it is possible to add a Stack tag containing a Token that gets resolved to the right value for all the resources in the stack, and this change would break that.

Issue #29781, opened 2024-04-10
apigateway: Can't set default authorization scopes when adding a proxy to a resource


David-Jacobsen:
### Describe the bug

When trying to add a Proxy to a Resource, you can't set the default authorization scopes because it attempts to set the authorization scopes to the OPTIONS method despite the OPTIONS method intentionally not having any authorizer.

```
    const api = new apigw.RestApi(this, "Actions-ApiGateway", {
        defaultCorsPreflightOptions: {
          allowOrigins: apigw.Cors.ALL_ORIGINS,
          maxAge: Duration.days(10)
        },        
      });
    const root = api.root;
    const sendResource = root.addResource(`InitiateAction`)
    const sendLambdaIntegration = new apigw.LambdaIntegration(this.initiateAction.lambda);
    const sendProxy = sendResource.addProxy({
      defaultIntegration: sendLambdaIntegration,
      anyMethod: true,
      defaultMethodOptions: {
        authorizer: this.authorizer,
        authorizationType: apigw.AuthorizationType.COGNITO,
        // authorizationScopes: [`scopes`],
    },
      defaultCorsPreflightOptions: {
        allowOrigins: apigw.Cors.ALL_ORIGINS,
        allowMethods: apigw.Cors.ALL_METHODS
      }
    })
```

This works, and creates an InitiateAction resource with a proxy using the Cognito Authorizer as the authorizer. But if I uncomment the authorizationScopes line then it fails because it tries to assign the authorizationScopes to the OPTIONS method.

### Expected Behavior

default authorization scopes are only applied to non-OPTIONS methods

### Current Behavior

default authorization scopes are applied to all methods, including OPTIONS, which causing the deployment to fail.

### Reproduction Steps

```
   const authorizer = authorizer = new apigw.CognitoUserPoolsAuthorizer(this, "Authorizer", {
        cognitoUserPools: [props.userPool],
      });
    const api = new apigw.RestApi(this, "Actions-ApiGateway", {
        defaultCorsPreflightOptions: {
          allowOrigins: apigw.Cors.ALL_ORIGINS,
          maxAge: Duration.days(10)
        },        
      });
    const root = api.root;
    const sendResource = root.addResource(`InitiateAction`)
    const sendLambdaIntegration = new apigw.LambdaIntegration(this.initiateAction.lambda);
    const sendProxy = sendResource.addProxy({
      defaultIntegration: sendLambdaIntegration,
      anyMethod: true,
      defaultMethodOptions: {
[...]

PR #30035, opened 2024-05-01
fix(apigateway): set authorization scope when authorization type is Cognito

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29781

### Reason for this change

Authorization scope is set even when the auth type is None. This will cause deployment failure 
>12:52:11 PM | CREATE_FAILED | AWS::ApiGateway::Method | ActionsApiGatewayI...oxyOPTIONS041B022A
Resource handler returned message: "Invalid Method authorization type specified. Authorization Scopes are only valid for COGNITO_USER_POOLS authorization type (Servic
e: ApiGateway, Status Code: 400, Request ID: f9c6357b-428e-42a8-884c-07b77939d165)" (RequestToken: bb8de2e9-37b7-ca15-9bd8-547bc7eea134, HandlerErrorCode: InvalidRequ
est)

### Description of changes

Check when auth type is not Cognito, set auth scope to none. Not a breaking change because original templates cannot deploy.

All existing and new tests pass.

### Checklist

----


packages/aws-cdk-lib/aws-apigateway/lib/method.ts
```diff
@@ -199,6 +199,11 @@ export class Method extends Resource {
         `which is different from what is required by the authorizer [${authorizer.authorizationType}]`);
     }
 
+    // When AuthorizationType is None, there shouldn't be any AuthorizationScope since AuthorizationScope should only
+    // be applied to COGNITO_USER_POOLS AuthorizationType.
```
nmussy, line 203: If I understand this correctly, `authorizationScopes` should not be exposed to users? Should we deprecate it and warn against its usage?

GavinZZ: I may miss something, why should we not expose it?

nmussy: I misunderstood the comment, apologies. Shouldn't we warn or throw instead of silently overriding the user's selection?

GavinZZ: Good call, will add a warning.

packages/aws-cdk-lib/aws-apigateway/test/method.test.ts
```diff
@@ -767,20 +769,137 @@ describe('method', () => {
...
-    Template.fromStack(stack).hasResourceProperties('AWS::ApiGateway::Method', {
-      ApiKeyRequired: true,
-      AuthorizationScopes: ['MethodAuthScope'],
+    expect(() => Template.fromStack(stack).hasResourceProperties('AWS::ApiGateway::Method', {
+      AuthorizationScopes: ['DefaultAuth'],
+    })).toThrow(/Template has 1 resources with type AWS::ApiGateway::Method, but none match as expected./);
+
+  });
+
+  test('Method options Auth Scopes should be None when Auth Type from Method is IAM', () => {
```
aaythapa, line 788: Can some of the tests you added for the different Auth Types be combined into 1 test?
```suggestion
  test.each([
  apigw.AuthorizationType.IAM,
  apigw.AuthorizationType.NONE,
  apigw.AuthorizationType.CUSTOM,
])('...', (authType) => {
    ...
    const api = new apigw.RestApi(stack, 'test-api', {
      cloudWatchRole: false,
      deploy: false,
      defaultMethodOptions: {
        authorizationType: authType,
        authorizationScopes: ['DefaultAuth'],
      },
    });
    ....
}
```

packages/aws-cdk-lib/aws-apigateway/lib/method.ts
```diff
@@ -199,6 +199,12 @@ export class Method extends Resource {
...
+    const defaultScopes = options.authorizationScopes ?? defaultMethodOptions.authorizationScopes;
+    const authorizationScopes = authorizationTypeOption === AuthorizationType.COGNITO ? defaultScopes : undefined;
+    Annotations.of(this).addWarningV2('@aws-cdk/aws-apigateway:invalidAuthScope', '\'AuthorizationScopes\' can only be set when \'AuthorizationType\' sets \'COGNITO_USER_POOLS\'. Default to ignore the values set in \'AuthorizationScopes\'.');
```
aaythapa, line 206: Why do we warn instead of throwing an error when AuthorizationType isn't Congnito but AuthorizationScopes is set?

packages/aws-cdk-lib/aws-apigateway/test/method.test.ts
```diff
@@ -767,20 +769,137 @@ describe('method', () => {
...
+      },
+    });
+
+    // WHEN
+    new apigw.Method(stack, 'MethodAuthScopeUsed', {
+      httpMethod: 'OPTIONS',
+      resource: api.root,
+      options: {
+        apiKeyRequired: true,
+        authorizationType: apigw.AuthorizationType.NONE,
```
aaythapa, line 806: A clarifying question, does this not override the `authorizationType` set in the `defaultMethodOptions`

packages/aws-cdk-lib/aws-apigateway/test/method.test.ts
```diff
@@ -767,20 +769,137 @@ describe('method', () => {
...
+    });
+
+    // THEN
+    expect(() => Template.fromStack(stack).hasResourceProperties('AWS::ApiGateway::Method', {
+      AuthorizationScopes: ['DefaultAuth'],
+    })).toThrow(/Template has 1 resources with type AWS::ApiGateway::Method, but none match as expected./);
+
+  });
+
+  test('Method options Auth Scopes should be None when Auth Type from RestApi is None', () => {
```
aaythapa, line 848: Duplicate test? I see another one named `Method options Auth Scopes should be None when Auth Type from Method is None` above which looks like it does the same thing

aaythapa (commented):
Generally lgtm, just some clarifying questions

packages/aws-cdk-lib/aws-apigateway/test/method.test.ts
GavinZZ: Good call, updated

packages/aws-cdk-lib/aws-apigateway/lib/method.ts
GavinZZ: We want to warn here because there's a subtle use case. When a API creates a OPTIONS method in addition to the POST/GET method (i.e. when authorizor is used), the POST/GET method is having the correct behaviour. However, the OPTIONS method will set auth scope regardless of auth type. The OPTIONS method is not explicitly managed by users. If we raise exception, then users would not be able to use the implicitly created OPTIONS method along with authorizor. Instead warn and continue seems like a better approach.

packages/aws-cdk-lib/aws-apigateway/test/method.test.ts
GavinZZ: Updated the test, no longer overriding.

packages/aws-cdk-lib/aws-apigateway/test/method.test.ts
GavinZZ: Removed. Good catch.

packages/aws-cdk-lib/aws-apigateway/lib/method.ts
aaythapa: mmm, i see. Thanks for the explanation

aaythapa (approved):
🚢

packages/aws-cdk-lib/aws-apigateway/lib/method.ts
```diff
@@ -199,6 +199,12 @@ export class Method extends Resource {
         `which is different from what is required by the authorizer [${authorizer.authorizationType}]`);
     }
 
+    // When AuthorizationType is None, there shouldn't be any AuthorizationScope since AuthorizationScope should only
+    // be applied to COGNITO_USER_POOLS AuthorizationType.
+    const defaultScopes = options.authorizationScopes ?? defaultMethodOptions.authorizationScopes;
+    const authorizationScopes = authorizationTypeOption === AuthorizationType.COGNITO ? defaultScopes : undefined;
```
alfie-bends, line 205: By utilizing authorizationTypeOption instead of authorizationType, you are now forcing people to implement an optional parameter for something was previously set by the authorizer, that was previously recommended to not be set explicitly. It is explicitly stated in the documentation, that the authorizer will set the authorizationType, see line 30 of this file. This is a breaking change.

tibbing: Yes, this just broke our prod environment after a dependabot update - it should be marked as a breaking change 🙁

PR #30042, opened 2024-05-02
feat(s3-deployment): switch CDKBucketDeployment Lambda architecture to ARM_64

### LINKED ISSUE #29996: (aws-s3-deployment): Switch Lambda architecture to ARM_64
### Describe the feature

To reduce costs and improve sustainability, it would be great to switch CDKBucketDeployment Lambda to ARM_64.

### Use Case

Reduced costs.

### Proposed Solution

The code looks compatible to ARM_64, switching architecture should be enough.

### Other Information

To give more information, it would be great to be able to have little more control about created Lambdas.
E.g. our environment enforces Lambdas to be deployed into a VPC. This is only possible using an Aspect.

### Acknowledgements

### CDK version used

2.102.0

independent

eruvanos:
### Issue # (if applicable)

Closes #29996

### Reason for this change

Reduce costs for s3 bucket deployments.

### Description of changes

Change LambdaFunction architecture

Run tests

### Checklist

----


eruvanos:
Exemption Request: This change does not need an entry in the README

daschaa (approved):

jfuss (changes requested):
How did we validate this works correctly? I have a couple concerns with this:

1) The Lambda Function that is getting updated from x86 to arm is using a Lambda Layer that is not built for arm. Since `AwsCliLayer` is using https://github.com/cdklabs/awscdk-asset-awscli which looks to only be built on an non arm version: https://github.com/cdklabs/awscdk-asset-awscli/blob/awscli-v1/main/layer/Dockerfile#L1. 
2) I also worry when ARM is added to Lambda for new regions and if they are in all the existing regions. I need to follow up on this internally but usually blanket things like this don't always work on. Which leads me to wanting to expose `architecture` as a property on the Construct and allowing customers to change this instead of us flipping under the hood.
3) It's possible someone is building on top of this construct and switching the arch without a way for these customers to keep the current arch is a breaking changing.

I think 1 is really important to get right. Which means we need to stage multiple changes here. We will need `AwsCliLayer` to be able to support both archs of Lambda before we can even look at 2/3. I will follow up internally but I think we need to be exposing this as a property on the Construct not just updating the value.

There also looks to be other changes in this PR that are outside of the Lambda Arch update. Please remove them as it makes it harder as a reviewer to understand the side effects of changes you are making. I can't tell if the appconfig tests are changing because of the intent of the PR or because you made other changes.

daschaa:
@jfuss Thank you for your review/reply! Actually I'm a bit confused with 1) because I deployed all the integration tests here so if the Layer does not support ARM, how is it then possible that the integration tests succeeded?

jfuss:
@daschaa Sorry for the delayed response, I had to take a couple unexpected days off last week.

> 1) because I deployed all the integration tests here so if the Layer does not support ARM, how is it then possible that the integration tests succeeded?

It's possible, so it's not surprising but my worry is at some point a change breaks things. Let me try to articulate things better here.

So the `AwsCliLayer` is a python package (AWS CLI v1). Lambda Layers are meant to target a Runtime[s] and Architecture[s]. So the part I am worried about is the `AwsCliLayer` is building on an `x86_64` image (`public.ecr.aws/sam/build-python3.8`). Docker will auto pull the `latest` tag, which is why I know it is an `x86_64` image (I was also a maintainer of SAM CLI which owns all the build images). It's very possible/likely that the the dependencies of the CLI don't have any architecture requirements today but that doesn't mean that can't change or you just didn't hit a case in the CLI that requires a different dependency for `arm`. I have been bitten by Python to many times to trust things will just always work in different architectures.

> Can you explain how I can see that this is only built for ARM?

What I would expect is, a change to `AwsCliLayer` repo that would add an `AwsCliLayerArm` (not locked on name). We can follow the same process as the current one but instead of building from `public.ecr.aws/sam/build-python3.8`, we would need to build from `public.ecr.aws/sam/build-python3.8@latest-arm64`. We can keep the `AwsCliLayer` the same for now, we might want to add `CompatibleArchitectures` but if it isn't on `AwsCliLayer` already, let's only add it to the `AwsCliLayerArm` to keep from unexpected breaks for existing customers. Once we have that, we can then do a `if arch == arm: use AwsCliLayerArm, else AwsCliLayer`. This just gets us into a better spot for maintaining things longer term. 

> Regarding the additonal changes: No, I didn't made any additional changes. I adjusted the appconfig test because the integration test was not working like it was implemented there (to be specific: the used bucket could not be deleted because it contained files).

Can you submit them in a different diff. I usually don't care that much but it would really help me as I don't have deep familiarity with the codebase. So I heavily rely on the diff to help me understand possible impact.

daschaa:
@jfuss Thanks for the clarification and the detailed writeup, now I understand the issue!
What are now the next steps to have a solution to the "problem"? 

Opening up the underlying lambda functions would be a really good idea, because for our enterprise context we also have to ensure that all lambdas are running inside a VPC (compliance reasons). Right now we solve this by using an aspect, but it would be a lot easier if the CDK just allows to put additional properties to the (internal) lambda functions.

eruvanos:
@jfuss @daschaa 

Looks like the proposed solution does not cover all aspects. I think it makes sense to close this PR and have the discussions on the issue.

PR #30050, opened 2024-05-03
feat(core): support Stage properties in App constructor

### LINKED ISSUE #24570: aws-cdk-lib(core): App Inherits From Stage, But Doesn't Allow Setting Region
### Describe the bug

I believe there's a fundamental "miss" in the code for `App`. `App` inherits from `Stage` yet does not allow setting any of the `StageProps`. As defined here https://github.com/aws/aws-cdk/blob/15cb919fab9d20d0e8f0485662131cbb10980269/packages/%40aws-cdk/core/lib/stage.ts#L13-L43, `env` allows setting the region from which all other resources in the stage will use.

Even when casting `as any` to force `env` into `App`s constructor, `App` ignores `props` as a whole and only passes `outdir` as can be seen here: https://github.com/aws/aws-cdk/blob/15cb919fab9d20d0e8f0485662131cbb10980269/packages/%40aws-cdk/core/lib/app.ts#L160-L162

Also of note, not passing a region to a `Stack` constructor's props will default the stack to `us-east-1` even though `--verbose` output shows that `CDK_DEFAULT_REGION` has been correctly pulled from `~/.aws/config`. 

### Expected Behavior

I would expect a number of things here:

1. We can set the region for the entire stage, via `App` in `AppProps`
2. The stage and stacks within it would use the value present in `CDK_DEFAULT_REGION` if nothing was specified manually in props
3. There would be a warning that `CDK_DEFAULT_REGION` diverged from what the default value of `us-east-1` with instructions or hints as to how to proceed. 

### Current Behavior

The stage silently defaults to `us-east-1`

### Reproduction Steps

Please reference code permalinks above

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.68.0

### Framework Version

2.68.0

### Node.js Version

18

### OS

Mac Vetura

### Language

Typescript

### Language Version

4.9

### Other information

_No response_

soucema9:
### Issue #24570

Closes #24570.

### Reason for this change

`App` construct inherits from `Stage` but does not allow to pass `Stage` parameters to constructor.

### Description of changes

Two parameters `env` and `permissionsBoundary` are added to `AppProps`. 

`stageName` is now only unsupported parameter. That is because `App` is special case of `Stage` acting as root for construct tree and as such it is synthesised to root of output directory. 

I copied unit tests for `env` propagation from `Stage` to `App`. 

### Checklist

----


soucema9:
Clarification Request

I didn't find any integration tests validating behaviour  of `App` or `Stage`.  And I'm not sure if it's even possible or make sense to write integration tests for this PR. As integration tests usually create stack, deploy it and assert on deployed resources but changes in this PR will affect where the stack is deployed.

PR #30051, opened 2024-05-03
feat(ses): allow VDM settings at the configuration set level

### LINKED ISSUE #30041: (ses): enable setting vdmOptions in `ConfigurationSet`
### Describe the feature

`CfnConfigurationSet` supported `vdmOptions`, but `ConfigurationSet` does not support vdmOptions.

https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-resource-ses-configurationset.html#cfn-ses-configurationset-vdmoptions

### Use Case

It's useful when you want to setting vdmOptions to the configuration set.

### Proposed Solution

Add  vdmOptions to `ConfigurationSet`.

```ts
    new ConfigurationSet(stack, 'ConfigurationSet', {
      customTrackingRedirectDomain: 'track.cdk.dev',
      suppressionReasons: SuppressionReasons.COMPLAINTS_ONLY,
      tlsPolicy: ConfigurationSetTlsPolicy.REQUIRE,
      dedicatedIpPool: new DedicatedIpPool(stack, 'Pool'),
      vdmOptions: { // Add
        engagementMetrics: true, // Add
        optimizedSharedDelivery: true, // Add
      }
    });
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.139.1

MacOS

mazyu36:
### Issue # (if applicable)

Closes #30041 .

### Reason for this change
As described in the issue.

### Description of changes
To allow VDM settings at the configuration set level, `vdmOptions` property has been added to the `ConfigurationSet` Construct.

```ts
new ses.ConfigurationSet(this, 'ConfigurationSetWithVdmOptions', {
  vdmOptions: { // Add
    engagementMetrics: true,
    optimizedSharedDelivery: true,
  },
});
```

I implemented unit tests and integration tests for the three cases.

1. Configuration set with both engagement metrics and optimized shared delivery enabled.
2. Configuration set with only engagement metrics enabled and optimized shared delivery not configured.
3. Configuration set with only optimized shared delivery enabled and engagement metrics not configured.

### Checklist

----


packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
```diff
@@ -73,6 +73,32 @@ export interface ConfigurationSetProps {
    * @default - use the default awstrack.me domain
    */
   readonly customTrackingRedirectDomain?: string;
+
+  /**
+   * The Virtual Deliverability Manager (VDM) options that apply to the configuration set
+   *
+   * @default - use account level settings
```
nmussy, line 80: We should probably indicate what the default values are first, and mention that this can be overridden by account level settings

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
```diff
@@ -73,6 +73,32 @@ export interface ConfigurationSetProps {
...
+}
+
+/**
+ * Properties for the Virtual Deliverability Manager (VDM) options that apply to the configuration set
+ */
+export interface VdmOptions {
+  /**
+   * Whether engagement metrics are enabled for the configuration set
+   *
+   * @default - use account level settings
```
nmussy, line 92: See above

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
```diff
@@ -73,6 +73,32 @@ export interface ConfigurationSetProps {
...
+   */
+  readonly engagementMetrics?: boolean;
+
+  /**
+   * Whether optimized shared delivery is enabled for the configuration set
+   *
+   * @default - use account level settings
```
nmussy, line 99: See above

packages/aws-cdk-lib/aws-ses/test/configuration-set.test.ts
```diff
@@ -40,3 +40,57 @@ test('configuration set with options', () => {
...
+
+  Template.fromStack(stack).hasResourceProperties('AWS::SES::ConfigurationSet', {
+    VdmOptions: {
+      DashboardOptions: Match.absent(),
+      GuardianOptions: {
+        OptimizedSharedDelivery: 'ENABLED',
+      },
+    },
+  });
+});
```
nmussy, line 96: Missing new line:
```suggestion
});

```

nmussy, line 96: These tests seem a bit superfluous, given they're already being covered by the first one. It'd be more appropriate to test `vdmOptions: {(false, false}` and `vdmOptions: {}`,

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
```diff
@@ -73,6 +73,32 @@ export interface ConfigurationSetProps {
...
+   */
+  readonly vdmOptions?: VdmOptions;
+}
+
+/**
+ * Properties for the Virtual Deliverability Manager (VDM) options that apply to the configuration set
+ */
+export interface VdmOptions {
+  /**
+   * Whether engagement metrics are enabled for the configuration set
```
nmussy, line 90: I find it more explicit to say exactly what the `true` value is. This way if the variable is something like `isFeatureDisabled`, you know exactly what `true` means
```suggestion
   * If true, engagement metrics are enabled for the configuration set
```

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
```diff
@@ -73,6 +73,32 @@ export interface ConfigurationSetProps {
...
+   *
+   * @default - use account level settings
+   */
+  readonly engagementMetrics?: boolean;
+
+  /**
+   * Whether optimized shared delivery is enabled for the configuration set
```
nmussy, line 97:
```suggestion
   * If true, optimized shared delivery is enabled for the configuration set
```

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
mazyu36: Thank you for the review.
I've stated that "default" indicates the absence of settings at the Configuration level, and in such cases, settings at the account level will be applied.

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
mazyu36: It's the same as the comment above.

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
mazyu36: It's the same as the comment above.

packages/aws-cdk-lib/aws-ses/test/configuration-set.test.ts
mazyu36: I've added test cases for when both options in vdmOptions are set to "DISABLED" and when vdmOptions itself is not configured. 

Additionally, I'm sorry to say that I've realized from the results of these tests that the "DISABLED" setting cannot be implemented. Therefore, I'm revisiting the implementation of vdmOptions.

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
```diff
@@ -150,6 +176,15 @@ export class ConfigurationSet extends Resource implements IConfigurationSet {
       trackingOptions: undefinedIfNoKeys({
         customRedirectDomain: props.customTrackingRedirectDomain,
       }),
+      vdmOptions: undefinedIfNoKeys({
+        dashboardOptions: props.vdmOptions?.engagementMetrics !== undefined ? {
```
mazyu36, line 180: In the previous implementation, if props.vdmOptions?.engagementMetrics was false, it resulted in undefined.

```ts
        dashboardOptions: props.vdmOptions?.engagementMetrics ? {
          engagementMetrics: booleanToEnabledDisabled(props.vdmOptions?.engagementMetrics),
        } : undefined,
```

Therefore, I'm modifying it so that the condition evaluates to true when props.vdmOptions?.engagementMetrics is set to false.

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
```diff
@@ -150,6 +176,15 @@ export class ConfigurationSet extends Resource implements IConfigurationSet {
...
+          engagementMetrics: booleanToEnabledDisabled(props.vdmOptions?.engagementMetrics),
+        } : undefined,
+        guardianOptions: props.vdmOptions?.optimizedSharedDelivery !== undefined ? {
```
mazyu36, line 183: It's the same as the comment above.

mazyu36:
@nmussy 
Thank you for your review! I have addressed the comments you provided.

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
```diff
@@ -73,6 +73,32 @@ export interface ConfigurationSetProps {
    * @default - use the default awstrack.me domain
    */
   readonly customTrackingRedirectDomain?: string;
+
+  /**
+   * The Virtual Deliverability Manager (VDM) options that apply to the configuration set
+   *
+   * @default - VDM options not configured at the configuration set level. In this case, use account level settings.
```
nmussy, line 80: Sorry, I should have made myself clearer. I think it's more important to tell the users that by default, VDM options are disabled, unless there are account level settings. Your current documentation only states that if they happen to have account level settings, those will be used. I would expect most of the CDK users will not be using account level settings.

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
nmussy: Sorry, I should have made myself clearer. I think it's more important to tell the users that by default, VDM options are disabled, unless there are account level settings. Your current documentation only states that if they happen to have account level settings, those will be used. I would expect most of the CDK users will not be using account level settings.

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
```diff
@@ -180,3 +215,9 @@ function renderSuppressedReasons(suppressionReasons?: SuppressionReasons): strin
       return ['COMPLAINT'];
   }
 }
+
+function booleanToEnabledDisabled(value: boolean): 'ENABLED' | 'DISABLED' {
+  return value === true
+    ? 'ENABLED'
+    : 'DISABLED';
+}
```
nmussy, line 223: Missing new line
```suggestion
}

```

packages/aws-cdk-lib/aws-ses/test/configuration-set.test.ts
nmussy: Have you tried setting it to `DISABLED` with `ENABLED` account wide settings? That might be the only allowed use case

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
mazyu36: I understand. Sorry for the lack of explanation.

However, my understanding is that the configuration set level settings need to be used together with the account level settings. (Deploying with only the configuration set level settings enabled is possible, but VDM itself cannot be used until the account level settings are enabled.)

That's why I tried to include the above as a Note in the README, but I would like your opinion.
I'm a little unsure about how it should be written...

packages/aws-cdk-lib/aws-ses/test/configuration-set.test.ts
mazyu36: Yes, I did and I could.

As I mentioned in a separate comment, the configuration set level settings can be deployed on their own, but to enable VDM, it is appropriate to configure them together with the account level settings.
Therefore, I also added the account level settings to the integration tests.

packages/aws-cdk-lib/aws-ses/README.md
```diff
@@ -239,3 +239,18 @@ Use the `VdmAttributes` construct to configure the Virtual Deliverability Manage
 // Enables engagement tracking and optimized shared delivery by default
 new ses.VdmAttributes(this, 'Vdm');
 ```
+
+If you want to override the VDM settings in the specified configuration set, use `vdmOptions` in the `ConfigurationSet` construct.
+
+> *Note* The configuration set level settings need to be used together with the account level settings.
+> If you enable only the configuration set level settings, VDM will not be enabled until the account level settings are configured.
+> For more information, see [Virtual Deliverability Manager settings](https://docs.aws.amazon.com/ses/latest/dg/vdm-settings.html).
```
nmussy, line 247: Formatting suggestion:

```suggestion
> **Note:** The configuration set level settings need to be used together with the account level settings.
If you enable only the configuration set level settings, VDM will not be enabled until the account level settings are configured.
For more information, see [Virtual Deliverability Manager settings](https://docs.aws.amazon.com/ses/latest/dg/vdm-settings.html).
```

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
nmussy: Oh, got it. The AWS documentation doesn't really make it explicit that they need to be enabled account wide first:
> Virtual Deliverability Manager options are also provided at the configuration set level so you can define custom settings for how a configuration set will use engagement tracking and optimized shared delivery by overriding how they’ve been defined in Virtual Deliverability Manager.

I would really emphasize it in the TSDoc as well, the README seems to have it covered

packages/@aws-cdk-testing/framework-integ/test/aws-ses/test/integ.configuration-set-vdmoptions.ts
```diff
@@ -0,0 +1,40 @@
...
+      },
+    });
+
+    new ses.ConfigurationSet(this, 'VdmOptionsNotConfigured', {
+    });
+
+    new ses.VdmAttributes(this, 'VdmAccountLevelSettings', {
+      engagementMetrics: true,
+      optimizedSharedDelivery: true,
+    });
```
nmussy, line 30: So there are the account-wide settings? I would indicate the construct name in the README and the TSDoc if that's the case

packages/@aws-cdk-testing/framework-integ/test/aws-ses/test/integ.configuration-set-vdmoptions.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+
+const app = new App();
+
+new integ.IntegTest(app, 'ConfigurationSetVdmOptionsInteg', {
+  testCases: [new TestStack(app, 'cdk-ses-configuration-set-vdmoptions-integ')],
+});
```
nmussy, line 38: It's probably a good idea to run some assertions to make sure that the configuration sets and VdmAttributes are created with the expected values, see [integ-tests](https://docs.aws.amazon.com/cdk/api/v2/docs/integ-tests-alpha-readme.html#assertions). Let me know if you need some guidance

mazyu36: @nmussy 
Thank you. I'm not used to it, but I'll give it a try. Please give me a little time.
I will try to implement assertions for account level and configuration set level settings using awsapicall.​​​​​​​​​​​​​​​​

mazyu36: @nmussy 
Thank you for your comments. I have added assertions using AwsApiCall for integration testing. Please let me know if there are any inappropriate parts.

packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
mazyu36: @nmussy 
Thank you. I have added a note to the README and docs that when setting Account Level settings with CDK, VdmAttributes should be used.

packages/@aws-cdk-testing/framework-integ/test/aws-ses/test/integ.configuration-set-vdmoptions.ts
mazyu36: As per the other comments, the README and docs have been updated.

nmussy (approved):
LGTM, thanks for the changes 👍

GavinZZ (approved):
Nice work, thanks for contributing!

PR #30060, opened 2024-05-04
chore(lambda): hide warning if skipPermissions is set

### LINKED ISSUE #29887: aws-lambda: addPermission() warning should not be shown
### Describe the bug

Incorrect warning is logged when synthesizing a `HttpLambdaAuthorizer` that points to a Lambda in another AWS account.

### Expected Behavior

No warning message as I'm not intentionally setting any permissions on the handler.

### Current Behavior

Consider a setup where a http lambda authorizer is used from another AWS account. In this case, the permissions to access the authorizer are already managed in the other account. When creating a lambda authorizer, a warning message is logged:
```
[Warning at /AuthorizerLambda] addPermission() has no effect on a Lambda Function with region=eu-west-1, account=12345678990, in a Stack with region=eu-west-1, account=09876543210. Suppress this warning if this is is intentional, or pass sameEnvironment=true to fromFunctionAttributes() if you would like to add the permissions. [ack: UnclearLambdaEnvironment]
```

### Reproduction Steps

Create a `HttpLambdaAuthorizer` with an ARN pointing to another account:
```ts
new HttpLambdaAuthorizer('LambdaAuthorizer',
    lambda.Function.fromFunctionAttributes(this, 'AuthorizerLambda', {
         functionArn: 'arn...',
     }
),
```

### Possible Solution

Looking at the [skipPermissions](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.FunctionAttributes.html#skippermissions) option, I would actually expect that this option would make sure this warning is not logged as we do not care about permissions not being added (it's already handled). So maybe something like this could be the fix in `function-base.ts`:
```ts
  public addPermission(id: string, permission: Permission) {
    if (!this.canCreatePermissions) {
      if (!this._skipPermissions) {
        Annotations.of(this).addWarningV2('UnclearLambdaEnvironment', `addPermission() has no effect on a Lambda Function with region=${this.env.region}, account=${this.env.account}, in a Stack with region=${Stack.of(this).region}, account=${Stack.of(this).account}. Suppress this warning if this is is intentional, or pass sameEnvironment=true to fromFunctionAttributes() if you would like to add the permissions.`);
      }
      return;
    }
  }
```

I also looked at where `addPermission()` is actually triggered for this use-case, which is [here](https://github.com/aws/aws-cdk/blame/main/packages/aws-cdk-lib/aws-apigatewayv2-authorizers/lib/http/lambda.ts#L106). Maybe we should have an option in `HttpLambdaAuthorizer` for not adding this permission in the first place? 

### CDK CLI Version

2.137.0

### Framework Version

_No response_

### Node.js Version

v18.19.0

### OS

OSX

### Language

TypeScript

### Language Version

5.2.0

### Other information

Ticket which is slightly related: https://github.com/aws/aws-cdk/issues/28936

daschaa:
### Issue #29887

Closes #29887

### Reason for this change

If an user imports a lambda and wants to add permissions a warning is show. This warning should be skippable with the skipPermissions flag.

Unit tests for checking if the warning is shown/not shown depending on the value of `skipPermissions` are added.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-rds/test/integ.instance-with-parameter-group.js.snapshot/aws-cdk-rds-instance-with-rds-parameter-group.assets.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-rds/test/integ.instance-with-parameter-group.js.snapshot/aws-cdk-rds-instance-with-rds-parameter-group.assets.json
@@ -1,7 +1,7 @@
 {
   "version": "36.0.0",
   "files": {
-    "27d09999310ee1193509c78be177319a1946d9d28632d2362b5f6168fd2078f1": {
+    "239d9f5d8cff7634197e9978eaf8133a6904b71cd0f07fbb239f839c7ce32886": {
       "source": {
         "path": "aws-cdk-rds-instance-with-rds-parameter-group.template.json",
         "packaging": "file"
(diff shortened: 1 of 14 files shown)


daschaa:
Exemption request - In my opinion this change neither needs an integration test nor an entry to the README.

comcalvi (approved):

PR #30061, opened 2024-05-04
feat(stepfunctions-tasks): step functions task for cross-region AWS API call

### LINKED ISSUE #29918: stepfunctions-tasks: Step Functions task for cross-region AWS API call
### Describe the feature

It would be useful if we could call AWS API across regions from a Step Functions state machine.

Currently it is not officially supported even with AWS SDK integration tasks.

> Currently, cross-Region AWS SDK integration and cross-Region AWS resource access aren't available in Step Functions.
> https://docs.aws.amazon.com/step-functions/latest/dg/concepts-access-cross-acct-resources.html

### Use Case

My usecase is to automate cross-region failover in a multi-region application. This requires you to orchestrate multiple API calls for both active and standby regions (e.g. failover Aurora DB cluster, rewrite AppConfig parameter, etc), and it would be great if we can manage these operations in a single state machine.

### Proposed Solution

Add a construct like `CallAwsServiceAcrossRegion` in the `aws-stepfunctions-tasks` module. The API would be similar to the existing [`CallAwsService` task](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions_tasks.CallAwsService.html), with an additional property `region`.

In the new construct, we create a Lambda function to proxy the API request to the target region, use [`LambdaInvoke` task](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions_tasks.LambdaInvoke.html) internally, and call AWS API from the function.

I wrote an example Lambda function to proxy AWS API requests: https://gist.github.com/tmokmss/aeb85fc8c3076cbc5d9ae1c6698969d1

The alternative option would be adding a property like `region` to `CallAwsService` task, but it could potentially make a breaking change when SFn would officially start to support cross-region AWS API call. That is why I think the above approach (creating a brand new construct) is better.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.138.0

n/a

tmokmss:
### Issue # (if applicable)

Closes #29918.

### Reason for this change

It would be useful if we could call AWS API across regions from a Step Functions state machine. Currently it is not officially supported even with AWS SDK integration tasks.

Our usecase is to automate a cross-region failover scenario in a multi-region application. This requires you to orchestrate multiple API calls for both active and standby regions (e.g. failover Aurora DB cluster, rewrite AppConfig parameter, etc), and it would be great if we can manage these operations in a single state machine.

### Description of changes

This PR adds a new construct `CallAwsServiceCrossRegion` that deploys 1. a Lambda function to call AWS API in different regions 2. SFn task to call the function.

Because most properties are compatible with the existing `CallAwsService` construct, you can use the new construct by just adding the `region` property.

Additionally, it also allows to set `endpoint` to override AWS API endpoint, because some AWS APIs requires you to override it. (e.g. [Route53 ARC](https://docs.aws.amazon.com/r53recovery/latest/dg/getting-started-cli-routing.control-state.html))

Here's example code to define this new SFn task:

```ts
    const pollTable = new tasks.CallAwsServiceCrossRegion(this, 'DescribeTable', {
      service: 'dynamodb',
      action: 'describeTable',
      parameters: {
        TableName: tableName,
      },
      // You can set any region in the same partition
      region: 'ap-northeast-1',
    });
```

Added unit tests and integ tests.

### Checklist

----


packages/@aws-cdk/custom-resource-handlers/lib/aws-stepfunctions-tasks/cross-region-aws-sdk-handler/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * API Action name (must be in camelCase or PascalCase)
+   */
+  action: string;
+
+  /**
+   * request parameters to call the API (must be in AWS SDK for JS v3 style)
+   */
+  parameters?: any;
```
nmussy, line 28: I couldn't find a common interface or union for all `InputTypes` across services, but this should be a safe bet?

```suggestion
  parameters?: Record<string, any>;
```

tmokmss: Hi @nmussy, thanks. Given that we cannot validate it in compile time nor synth time, I think it will not add much value if we narrow its type definition. It will result in a runtime error anyway.

packages/@aws-cdk/custom-resource-handlers/lib/aws-stepfunctions-tasks/cross-region-aws-sdk-handler/index.ts
```diff
@@ -0,0 +1,79 @@
...
+};
+
+// reference: code from CDK AwsCustomResource
+function findV3ClientConstructor(pkg: object) {
+  const [_clientName, ServiceClient] = Object.entries(pkg).find(([name]) => {
```
nmussy, line 33: It might be a good idea to `filter` the entries instead, and throw if `length !== 1`, just in case multiple `Client`s are present in a single package

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/lambda/integ.call-aws-service-cross-region.ts
nmussy, line 1: Have you confirmed that this integration is also able to run in `eu-west-1`, i.e. the same region as the "cross-region" target?

packages/@aws-cdk/custom-resource-handlers/lib/aws-stepfunctions-tasks/cross-region-aws-sdk-handler/index.ts
```diff
@@ -0,0 +1,79 @@
...
+        send: (command: any) => Promise<any>;
+        config: any;
+      };
+    },
+  ];
+  return ServiceClient;
+}
+
+function findCommandClass(pkg: object, action: string) {
+  const camelToPascal = (str: string) => str.charAt(0).toUpperCase() + str.slice(1);
```
nmussy, line 49: Small nit, but you can move the `camelToPascal` function outside the `findCommandClass` scope, to prevent it being re-declared on every call

packages/@aws-cdk/custom-resource-handlers/lib/aws-stepfunctions-tasks/cross-region-aws-sdk-handler/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  return Command;
+}
+
+export const handler = async (event: Event) => {
+  // Log the event to understand the incoming request structure
+  console.log('Event: ', event);
+
+  try {
+    // esbuild-disable unsupported-require-call -- not esbuildable but that's fine
+    const pkg = require(`@aws-sdk/client-${event.service}`);
```
nmussy, line 66: Correct me if I'm wrong, but `normalizeServiceName` is not being ran for your handler? It would allow us to keep the same pattern as other `service`s props, see [AwsSdkCall](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.custom_resources.AwsSdkCall.html#service):

https://github.com/aws/aws-cdk/blob/fff9cf694b14811682c8671a1e55afa53151df8b/packages/@aws-cdk/aws-custom-resource-sdk-adapter/lib/api-call.ts#L65

https://github.com/aws/aws-cdk/blob/fff9cf694b14811682c8671a1e55afa53151df8b/packages/@aws-cdk/aws-custom-resource-sdk-adapter/lib/sdk-info.ts#L10-L15

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
```diff
@@ -0,0 +1,182 @@
...
+import { Token, Duration } from '../../../core';
+import { CrossRegionAwsSdkSingletonFunction } from '../../../custom-resource-handlers/dist/aws-stepfunctions-tasks/cross-region-aws-sdk-provider.generated';
+
+/**
+ * Properties for calling an AWS service's API action from your
+ * state machine across regions.
+ */
+export interface CallAwsServiceCrossRegionProps extends sfn.TaskStateBaseProps {
+  /**
+   * The AWS service to call in AWS SDK for JavaScript v3 style.
```
nmussy, line 14: See comment about `normalizeServiceName`

packages/@aws-cdk/custom-resource-handlers/lib/aws-stepfunctions-tasks/cross-region-aws-sdk-handler/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    const Client = findV3ClientConstructor(pkg);
+    const Command = findCommandClass(pkg, event.action);
```
nmussy, line 68: Similarly, you should be able to use `normalizeActionName`, see [AwsSdkCall](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.custom_resources.AwsSdkCall.html#action):
https://github.com/aws/aws-cdk/blob/fff9cf694b14811682c8671a1e55afa53151df8b/packages/@aws-cdk/aws-custom-resource-sdk-adapter/lib/sdk-info.ts#L10-L16

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
```diff
@@ -0,0 +1,182 @@
...
+   *
+   * @see https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/
+   */
+  readonly service: string;
+
+  /**
+   * The API action to call.
+   *
+   * Use camelCase.
```
nmussy, line 23: See comment about `normalizeActionName`

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
```diff
@@ -0,0 +1,182 @@
...
+      });
+    }
+  }
+
+  /**
+   * Provides the Lambda Invoke service integration task configuration
+   */
+  /**
+   * @internal
+   */
```
nmussy, line 168: Any reason to split the TSDoc?
```suggestion
  /**
   * Provides the Lambda Invoke service integration task configuration
   *
   * @internal
   */
```

packages/@aws-cdk/custom-resource-handlers/lib/aws-stepfunctions-tasks/cross-region-aws-sdk-handler/index.ts
tmokmss: I'd like to keep this feature and API as simple as possible to lower the maintenance cost. `AwsSdkCall` code has become such complicated for backward compatibility reason. Since this feature is new, we don't have to introduce the complexity at all.

Also the API is almost compatible with the existing `CallAwsService` construct, which also does not support normalizing service names.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/lambda/integ.call-aws-service-cross-region.ts
tmokmss: I added an integ-test case for the scenario.

tmokmss:
Hi @nmussy, thanks for the comments.

> reuse the existing CallAwsService, and just to add an optional crossRegionTarget prop

I considered this approach as well but chose the current approach, because this approach can easily avoid from future breaking changes when SFn officially supports cross-region API call; we can just deprecate `CallAwsServiceCrossRegion` construct and add the official functionality to `CallAwsService` construct.

packages/@aws-cdk/custom-resource-handlers/lib/aws-stepfunctions-tasks/cross-region-aws-sdk-handler/index.ts
tmokmss: Same as service name, I'd like to intentionally only support action names in camelCase.

packages/@aws-cdk/custom-resource-handlers/lib/aws-stepfunctions-tasks/cross-region-aws-sdk-handler/index.ts
nmussy: I'll let the maintainer weigh in, I think it's a nice piece of QoL, but I understand your maintainability concern

nmussy (approved):
LGTM, just a couple of pending comments

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  constructor(scope: Construct, id: string, private readonly props: CallAwsServiceCrossRegionProps) {
+    super(scope, id, props);
+
+    if (props.integrationPattern === sfn.IntegrationPattern.RUN_JOB) {
+      throw new Error('The RUN_JOB integration pattern is not supported for CallAwsService');
+    }
+    if (!Token.isUnresolved(props.action) && !props.action.startsWith(props.action[0]?.toLowerCase())) {
+      throw new Error(`action must be camelCase, got: ${props.action}`);
+    }
```
GavinZZ, line 114: I know you mentioned that it's similar to `CallAwsService` where no validation/normalization is done. I would still recommend using the `normalizeServiceName` and `normalizeActionName`. If users input something like `getobject` we wouldn't catch it. While they would eventually know from the logs/execution results, why not normalize it ahead?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
```diff
@@ -0,0 +1,184 @@
...
+import { IFunction } from '../../../aws-lambda';
+import * as sfn from '../../../aws-stepfunctions';
+import { Token, Duration } from '../../../core';
+import { CrossRegionAwsSdkSingletonFunction } from '../../../custom-resource-handlers/dist/aws-stepfunctions-tasks/cross-region-aws-sdk-provider.generated';
+
+/**
+ * Properties for calling an AWS service's API action from your
+ * state machine across regions.
+ */
+export interface CallAwsServiceCrossRegionProps extends sfn.TaskStateBaseProps {
```
GavinZZ, line 12: This seems to be re-defining a lot of props from `CallAwsServiceProps`. Any reason we're not extending `CallAwsServiceProps` instead?

packages/@aws-cdk/custom-resource-handlers/lib/aws-stepfunctions-tasks/cross-region-aws-sdk-handler/index.ts
GavinZZ: To help me understand, can you elaborate on your concern? What maintenance cost are you concerned with using AwsSdkCall code.

GavinZZ (commented):
Left some minor feedbacks. Also have a high-level questions I want to clarify on. There's already a stepfunctions task for `CallAwsService`. Is the reason of building this `cross-region-aws-sdk-handler` using custom resource lambda function because the AWS SDK service integrations do not support cross region invocation?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * The action for the IAM statement that will be added to the Lambda
+   * function role's policy to allow the state machine to make the API call.
+   *
+   * Use in the case where the IAM action name does not match with the
+   * API service/action name, e.g. `s3:ListBuckets` requires `s3:ListAllMyBuckets`.
+   *
+   * @default - service:action
+   */
+  readonly iamAction?: string;
```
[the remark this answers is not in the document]
tmokmss: In that case, you can use `additionalIamStatements`. I agree that this should ideally be `iamActions?: string[]`, but then we lose API compatibility with `CallAwsService` construct. I think it's okay to leave this as-is because most API only requires a single IAM action anyway . wdyt?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
```diff
@@ -0,0 +1,184 @@
...
+   *
+   * @default - no parameters
+   */
+  readonly parameters?: { [key: string]: any };
+
+  /**
+   * The resources for the IAM statement that will be added to the Lambda
+   * function role's policy to allow the state machine to make the API call.
+   *
+   * By default the action for this IAM statement will be `service:action`.
```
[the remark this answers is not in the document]
tmokmss: Thanks. I moved the line to iamAction.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
tmokmss: I think your example `getobject` is much more difficult to catch. Afaik there has been no complaint about `CallAwsService` only supporting camelCase, so I don't think it's worth supporting other formats in this construct either, at least at this moment.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
tmokmss: They look similar but are slightly different, e.g. `service` property's name format. Also I'd like to completely decouple them to avoid unwanted side effect to the stable CallAwsService construct.

tmokmss:
Hi @GavinZZ, thank you so much for the review!

>  Is the reason of building this cross-region-aws-sdk-handler using custom resource lambda function because the AWS SDK service integrations do not support cross region invocation

Yes. According to the doc:

> Currently, cross-Region AWS SDK integration and cross-Region AWS resource access aren't available in Step Functions.
> https://docs.aws.amazon.com/step-functions/latest/dg/concepts-access-cross-acct-resources.html

tmokmss:
> To help me understand, can you elaborate on your concern? What maintenance cost are you concerned with using AwsSdkCall code.

AwsSdkCall's code is mainly to solve the v2-v3 compatibility problem. Since this construct does not require such compatibility, I think there is little advantage to using the existing code. After all, the new Lambda is only code under 100 lines. Also, if we want to reuse the existing code, we need to refactor it first (for example, currently it's only for CFn custom resource handler). Given the ROI, I'd rather keep them independent.

packages/@aws-cdk/custom-resource-handlers/lib/aws-stepfunctions-tasks/cross-region-aws-sdk-handler/index.ts
tmokmss: AwsSdkCall's code is mainly to solve the v2-v3 compatibility problem. Since this construct does not require such compatibility, I think there is little advantage to using the existing code. After all, the new Lambda is only code under 100 lines. Also, if we want to reuse the existing code, we need to refactor it first (for example, currently it's only for CFn custom resource handler, but this one is for invocation from SFn). Given the ROI, I'd rather keep them independent.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
GavinZZ: Fair point, decoupling makes sense to me

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/lambda/call-aws-service-cross-region.ts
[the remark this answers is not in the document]
GavinZZ: Can you share an example how to use `additionalIamStatements` with another action? I also think it would be nice to add this example to the readme file.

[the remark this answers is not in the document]
GavinZZ: Sounds good, thank you!

GavinZZ (approved):

tmokmss:
@GavinZZ Integ test failed due to a recent mainline change. Looking forward to your re-approval 🙏 Thanks!

GavinZZ (approved):

PR #30064, opened 2024-05-05
feat(codebuild): add missing types for webhook filters

### LINKED ISSUE #30062: (codebuild): Add support new newly introduced event filter
### Describe the feature

A new feature has been recently introduced in AWS CodeBuild: https://aws.amazon.com/about-aws/whats-new/2024/04/aws-codebuild-managed-github-action-runners/. This feature enables to use AWS CodeBuild as GitHub Actions runner. However, using this feature requires to configure a new event filter which is currently not yet available in CDK.

### Use Case

To be able to use 'aws-cdk-lib/aws-codebuild.Project' L2 Construct to define a runner for Github Action I need to be able to pass in `FilterGroup.inEventOf(EventAction.WORKFLOW_JOB_QUEUED)` but its not available so for now I'm force to use L1 Construct which requires setting up IAM roles etc which would be automatically created by the Project construct.

### Proposed Solution

Introduce new value to `aws-cdk-lib/aws-codebuild.EventAction` enum: "WORKFLOW_JOB_QUEUED"

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.140.0

MacOS 14.3.1 (23D60)

msambol:
The full list is found [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codebuild-project-webhookfilter.html#cfn-codebuild-project-webhookfilter-type).

Closes #30062.


msambol:
Exemption request: I need a GitHub access token to update the integration test: `Could not find access token for server type github`. I think it's OK with no integration test update as I'm only adding enum values.

msambol:
Exemption Request: I need a GitHub access token to update the integration test: `Could not find access token for server type github`. I think it's OK with no integration test update as I'm only adding enum values.

nmussy (approved):

msambol:
Exemption Request: I need a GitHub access token to update the integration test: Could not find access token for server type github. I think it's OK with no integration test update as I'm only adding enum values.

msambol:
> Could you add an integ test for these enums? The integ tests helps prevent regressions.

See my exemption request above. I think I need a access token? Happy to, but CDK core team might need to run it?

msambol:
@comcalvi The webhooks are [only available](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_codebuild.Source.html) on GitHub and BitBucket. I looked at doing an integration test with CodeCommit, but not possible here. Let me know how you want to proceed.

aseppala:
fyi @msambol & @comcalvi: I opened https://github.com/aws/aws-cdk/pull/30204 that extends this PR and adds the requested integration test. Please have a look.

packages/aws-cdk-lib/aws-codebuild/test/codebuild.test.ts
```diff
@@ -567,6 +567,9 @@ describe('default properties', () => {
         webhookFilters: [
           codebuild.FilterGroup.inEventOf(codebuild.EventAction.PUSH).andTagIsNot('stable'),
           codebuild.FilterGroup.inEventOf(codebuild.EventAction.PULL_REQUEST_REOPENED).andBaseBranchIs('main'),
+          codebuild.FilterGroup.inEventOf(codebuild.EventAction.RELEASED).andBaseBranchIs('main'),
```
comcalvi, line 570: could we add the `CLOSED` to this test?

comcalvi (changes requested):

comcalvi:
If the only integration test we can add is one that can't be deployed reliably, then I think it's safe to skip the integ test for this one. Thanks both, I think we can leave the integ test off of this one.

msambol:
@comcalvi changes made as requested 👍🏼

msambol:
@comcalvi gentle nudge 😄

comcalvi (approved):

msambol:
@comcalvi Do I need to do something further to push this through?

msambol:
@comcalvi Could we push this one through before it gets closed?

PR #30084, opened 2024-05-07
fix(pipelines): pipeline asset role trust policy has account root principal

xazhao:
### Note
This may be a **breaking change** for some users. If the `AssetFileRole` is being assumed by other service/Users/Roles, they will not be able to assume the role after this change by default.

To revert back to the old behavior, setting the following feature flag to false in the `context` section of `cdk.json`
```
"@aws-cdk/pipelines:reduceAssetRoleTrustScope": true
```

The is enabled by default to address the overly broad trust policy of `AssetFileRole`
### Reason for this change

CDK Pipeline will create a `AssetFileRole` which has trust policy including the root account principal. The root account principal is not needed in this use case and should be removed to scope down trust policy.

### Description of changes
Adding a new feature flag `PIPELINE_REDUCE_ASSET_ROLE_TRUST_SCOPE` with default value `true`.
When the feature flag is enabled, remove the root account principal from the trust policy.
When the feature flag is disabled, keep the old behavior.

Using the feature flag here in case of customers are using the root account principal and it will allow them to turn off this change.

Unit test/Integration Test
Manually tested in cross-account pipeline

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/cx-api/test/features.test.ts
+++ b/packages/aws-cdk-lib/cx-api/test/features.test.ts
@@ -38,7 +38,7 @@ test('feature flag defaults may not be changed anymore', () => {
     [feats.LAMBDA_RECOGNIZE_VERSION_PROPS]: true,
     [feats.CLOUDFRONT_DEFAULT_SECURITY_POLICY_TLS_V1_2_2021]: true,
     // Add new disabling feature flags below this line
-    // ...
+    [feats.PIPELINE_REDUCE_ASSET_ROLE_TRUST_SCOPE]: true,
 
   });
 });


packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1037,6 +1038,21 @@ export const FLAGS: Record<string, FlagInfo> = {
...
+  [PIPELINE_REDUCE_ASSET_ROLE_TRUST_SCOPE]: {
+    type: FlagType.ApiDefault,
+    summary: 'Remove the root account principal from PipelineAssetsFileRole trust policy',
+    detailsMd: `
+      Remove the root account principal from the generated PipelineAssetsFileRole trust policy,
+      because it's only assumed by the codebuild.
+      Use a feature flag to make sure existing customers who might be relying
+      on the overly-broad trust policy are not broken.
+    `,
+    introducedIn: { v2: '2.140.0' },
```
GavinZZ, line 1051: This should be `V2NEXT` according to the [documentation](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#L1201) instead of explicitly specifying the version.

packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
```diff
@@ -69,6 +69,7 @@ Flags come in three types:
 | [@aws-cdk/aws-kms:reduceCrossAccountRegionPolicyScope](#aws-cdkaws-kmsreducecrossaccountregionpolicyscope) | When enabled, IAM Policy created from KMS key grant will reduce the resource scope to this key only. | 2.134.0 | (fix) |
 | [@aws-cdk/aws-eks:nodegroupNameAttribute](#aws-cdkaws-eksnodegroupnameattribute) | When enabled, nodegroupName attribute of the provisioned EKS NodeGroup will not have the cluster name prefix. | 2.139.0 | (fix) |
 | [@aws-cdk/aws-ec2:ebsDefaultGp3Volume](#aws-cdkaws-ec2ebsdefaultgp3volume) | When enabled, the default volume type of the EBS volume will be GP3 | 2.140.0 | (default) |
+| [@aws-cdk/pipelines:reduceAssetRoleTrustScope](#aws-cdkpipelinesreduceassetroletrustscope) | Remove the root account principal from PipelineAssetsFileRole trust policy | 2.140.0 | (default) |
```
GavinZZ, line 72: This should be `V2NEXT` as well.

packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
```diff
@@ -1298,4 +1300,22 @@ When this featuer flag is enabled, the default volume type of the EBS volume wil
...
 
 
+### @aws-cdk/pipelines:reduceAssetRoleTrustScope
+
+*Remove the root account principal from PipelineAssetsFileRole trust policy* (default)
+
+Remove the root account principal from the generated PipelineAssetsFileRole trust policy,
+because it's only assumed by the codebuild.
+Use a feature flag to make sure existing customers who might be relying
+on the overly-broad trust policy are not broken.
```
GavinZZ, line 1310: Instead of explaining why we add a feature flag here, can we explain the behaviour of the flag when it's true vs false, i.e. `When this feature flag is enabled, the it will not add root account principal ...`

packages/aws-cdk-lib/pipelines/test/codepipeline/codepipeline.test.ts
```diff
@@ -197,6 +198,28 @@ test('CodeBuild action role has the right AssumeRolePolicyDocument', () => {
   });
 });
 
+test('CodeBuild asset role has the right Principal', () => {
```
GavinZZ, line 201: nit: can we have another test to strictly test when the flag is false and the role has the root account principal?

GavinZZ (commented):
Some doc changes comment.

packages/aws-cdk-lib/pipelines/test/compliance/assets.test.ts
```diff
@@ -396,17 +396,6 @@ describe('basic pipeline', () => {
...
-                Action: 'sts:AssumeRole',
-                Effect: 'Allow',
-                Principal: {
-                  AWS: {
-                    'Fn::Join': ['', [
-                      'arn:', { Ref: 'AWS::Partition' }, `:iam::${PIPELINE_ENV.account}:root`,
-                    ]],
-                  },
-                },
-              },
```
GavinZZ, line 409: Why does this test change when no feature flag is set?

packages/aws-cdk-lib/cx-api/lib/features.ts
xazhao: Good catch. Updated.

packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
xazhao: Good catch. Updated.

packages/aws-cdk-lib/pipelines/test/codepipeline/codepipeline.test.ts
xazhao: yep added another test to cover that

packages/aws-cdk-lib/pipelines/test/compliance/assets.test.ts
xazhao: Since the behavior has been tested with feature flag in the other test, we can use this to test default value.

GavinZZ (approved):
LGTM as long as we update the description with more explanation of the security issue and the solution with the feature flag (how and why it defaults to true)

PR #30093, opened 2024-05-07
fix(diff): properties from ChangeSet diff were ignored

### LINKED ISSUE #29731: (cli): changeset-based `cdk diff` not showing changes caused by SSM parameters, but appear in CloudFormation UI changeset info
### Describe the bug

Currently if you use a Systems Manager StringParameter as a value to another resource, the `cdk diff` does not identify a diff when the underlying parameter changes, even when using the new behavior that produces a CloudFormation changeset (vs. template-only diffs). The new behavior to produce a real changeset is designed to identify these types of changes.

If I manually create a changeset in the console, it _does_ identify the change, so this feels like an issue specific to the CDK changeset diffing behavior.

### Expected Behavior

I expect to be notified that my stack will change, just like I am if I upload the template to the CloudFormation UI. This bug is especially concerning if the change could trigger an unexpected resource replacement (which makes me think this should be a P1 issue).

### Current Behavior

`cdk diff` says `There were no differences` when, really, CloudFormation generates a changeset that shows a diff.

### Reproduction Steps

Consider the following straightforward stack:

```ts
import { Queue } from 'aws-cdk-lib/aws-sqs';
import { StringParameter } from 'aws-cdk-lib/aws-ssm';

export class CdkBugReportStack extends cdk.Stack {
    super(scope, id, props);

    const queueNameFromParameter = StringParameter.valueForStringParameter(this, '/cdk/test/queue-name-param');
    new Queue(this, "Queue", {
      queueName: queueNameFromParameter
    })
  }
}
```

It imports an SSM StringParameter and uses the resolved value to name the queue.

1. Create an SSM StringParameter with an initial value. For this repro, I'll call it `blimmer-test-1`.

Screenshot_2024-04-04_at_14_31_24

2. Deploy the example stack above (`cdk deploy CdkBugReportStack`). You'll see your queue is cre
ated with the name of your StringParameter from step 1.

Screenshot_2024-04-04_at_14_32_49

3. Now, edit the StringParameter from step 1. I updated the value from `blimmer-test-1` to `blimmer-test-2`.

Screenshot_2024-04-04_at_14_33_04

4. Run `cdk diff CdkBugReportStack`. Make sure you're using the latest CDK (at time of writing v2.135.0) and that you're generating a changeset for the diff (e.g., not passing `--no-change-set`).

	```shell
	> npx cdk diff CdkBugReportStack
	Stack CdkBugReportStack
	There were no differences
	
	```

	As you can see, no diffs are identified (even though the underlying parameter _did_ change).

5. Generate the CloudFormation stack for use in the console in the next steps:

	```
	> npx cdk synth -j CdkBugReportStack > stack.json
	```

6. Visit the CloudFormation service page in the AWS Console. Select the stack, `CdkBugReportStack`.
7. Choose "Stack Actions" -> "Create change set for current stack".

Screenshot_2024-04-04_at_14_49_59

8. Replace the template with the one you generated in step 5.

Screenshot_2024-04-04_at_14_51_24

9. Keep clicking "next" and, finally, submit, without changing any other values.
10. Now, once the changeset is rendered, you'll see that it identifies that the SSM parameter has changed and will result in a replacement of the queue. This is what I expect aws-cdk to report, too.

Screenshot_2024-04-04_at_14_56_05

11. Now you can also run `npx cdk deploy CdkBugReportStack` and you'll see that the queue is replaced, even though the `diff` said no changes were detected.

```shell
> npx cdk deploy

✨  Synthesis time: 1.84s

CdkBugReportStack:  start: Building 3514dbc73317309e7175dd9ffc618c6a7b4f814c0383edf38308e14ddaf77d81:current_account-current_region
CdkBugReportStack:  success: Built 3514dbc73317309e7175dd9ffc618c6a7b4f814c0383edf38308e14ddaf77d81:current_account-current_region
CdkBugReportStack:  start: Publishing 3514dbc73317309e7175dd9ffc618c6a7b4f814c0383edf38308e14ddaf77d81:current_account-current_region
CdkBugReportStack:  success: Published 3514dbc73317309e7175dd9ffc618c6a7b4f814c0383edf38308e14ddaf77d81:current_account-current_region
CdkBugReportStack: deploying... [1/1]
CdkBugReportStack: creating CloudFormation changeset...

 ✅  CdkBugReportStack

✨  Deployment time: 43.33s

Stack ARN:
arn:aws:cloudformation:us-west-2:REDACTED:stack/CdkBugReportStack/6b883590-f2c2-11ee-afe3-025d26e8baa1

✨  Total time: 45.17s
```

### Possible Solution

It feels like if CloudFormation can identify this change, CDK should also be able to identify the change when running the more accurate changeset-based diff.

I was confused why my services were sometimes redeploying when no diffs were shown via `cdk diff`. It turned out that my problem was with `obtainDefaultFluentBitECRImage`, which obtains the fluent bit image via an SSM parameter ([docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/firelens-using-fluentbit.html#firelens-image-ecr)). When the underlying parameter changed, it caused my task definitions and services to be updated.

Linking this up with https://github.com/aws/aws-cdk/issues/7366 and https://github.com/aws/aws-cdk/issues/23288, which are related to the specific issue I ran into here.

### CDK CLI Version

2.135.0 (build d46c474)

### Framework Version

_No response_

### Node.js Version

20 LTS

### OS

MacOS

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

bergjaak:
### Issue # (if applicable)

Closes aws/aws-cdk-cli#641

### Reason for this change

* Diffs that are only present in the change set are ignored

### Description of changes

* Include changes to properties that are only present in the changeset

* Here is an image of what the change looks like, with the fix on the right and the old behavior on the left. I did this with a CDK app that is the same as given in the example from the customer who opened the issue (29731), but the app also includes a new queue (which is in the left and right).

29731Fix

* manual, unit, and integration tested. Ran all integration tests that mention `cdk diff`:

```
[INTEG TEST::cdk diff] Starting (pid 34550)...
[INTEG TEST::cdk diff] Done (13725 ms).
[INTEG TEST::cdk diff --fail on multiple stacks exits with error if any of the stacks contains a diff] Starting (pid 34550)...
[INTEG TEST::cdk diff --fail on multiple stacks exits with error if any of the stacks contains a diff] Done (80833 ms).
[INTEG TEST::cdk diff --fail with multiple stack exits with if any of the stacks contains a diff] Starting (pid 34550)...
[INTEG TEST::cdk diff --fail with multiple stack exits with if any of the stacks contains a diff] Done (81796 ms).
[INTEG TEST::cdk diff --security-only successfully outputs sso-permission-set-without-managed-policy information] Starting (pid 34550)...
[INTEG TEST::cdk diff --security-only successfully outputs sso-permission-set-without-managed-policy information] Done (7394 ms).
[INTEG TEST::cdk diff --security-only successfully outputs sso-permission-set-with-managed-policy information] Starting (pid 34550)...
[INTEG TEST::cdk diff --security-only successfully outputs sso-permission-set-with-managed-policy information] Done (7043 ms).
[INTEG TEST::cdk diff --security-only successfully outputs sso-assignment information] Starting (pid 34550)...
[INTEG TEST::cdk diff --security-only successfully outputs sso-assignment information] Done (6930 ms).
[INTEG TEST::cdk diff --security-only successfully outputs sso-access-control information] Starting (pid 34550)...
[INTEG TEST::cdk diff --security-only successfully outputs sso-access-control information] Done (6958 ms).
[INTEG TEST::cdk diff --security-only --fail exits when security diff for sso access control config] Starting (pid 34550)...
[INTEG TEST::cdk diff --security-only --fail exits when security diff for sso access control config] Done (6945 ms).
[INTEG TEST::cdk diff --security-only --fail exits when security diff for sso-perm-set-without-managed-policy] Starting (pid 34550)...
[INTEG TEST::cdk diff --security-only --fail exits when security diff for sso-perm-set-without-managed-policy] Done (7042 ms).
[INTEG TEST::cdk diff --security-only --fail exits when security diff for sso-perm-set-with-managed-policy] Starting (pid 34550)...
[INTEG TEST::cdk diff --security-only --fail exits when security diff for sso-perm-set-with-managed-policy] Done (7131 ms).
[INTEG TEST::cdk diff --security-only --fail exits when security diff for sso-assignment] Starting (pid 34550)...
[INTEG TEST::cdk diff --security-only --fail exits when security diff for sso-assignment] Done (7225 ms).
[INTEG TEST::cdk diff --security-only --fail exits when security changes are present] Starting (pid 34550)...
[INTEG TEST::cdk diff --security-only --fail exits when security changes are present] Done (7124 ms).
[INTEG TEST::cdk diff --quiet does not print 'There were no differences' message for stacks which have no differences] Starting (pid 34550)...
[INTEG TEST::cdk diff --quiet does not print 'There were no differences' message for stacks which have no differences] Done (75387 ms).
[INTEG TEST::cdk diff picks up changes that are only present in changeset] Starting (pid 34550)...
[INTEG TEST::cdk diff picks up changes that are only present in changeset] Done (98624 ms).
 PASS  tests/cli-integ-tests/cli.integtest.js (414.667 s)
  ● Console

    console.log
      Using regions: us-east-1

      at log (../../lib/with-aws.ts:36:11)


Test Suites: 2 skipped, 1 passed, 1 of 3 total
Tests:       90 skipped, 14 passed, 104 total
Snapshots:   0 total
Time:        414.86 s
Ran all test suites with tests matching "cdk diff".
```

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
+++ b/packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
@@ -1,5 +1,6 @@
 import * as fc from 'fast-check';
 import { arbitraryTemplate } from './test-arbitraries';
+import { sqsQueue, sqsQueueWithAargs, ssmParam } from './util';
 import { fullDiff, ResourceImpact } from '../lib/diff-template';
 
 const POLICY_DOCUMENT = { foo: 'Bar' }; // Obviously a fake one!
@@ -1242,16 +1243,7 @@ describe('changeset', () => {
         },
       },
       Resources: {
-        mySsmParameter: {
-          Type: 'AWS::SSM::Parameter',
-          Properties: {
-            Name: 'mySsmParameterFromStack',
-            Type: 'String',
-            Value: {
-              Ref: 'SsmParameterValuetestbugreportC9',
-            },
-          },
-        },
+        mySsmParameter: ssmParam,
       },
     };
 
@@ -1293,26 +1285,8 @@ describe('changeset', () => {
     expect(diffWithChangeSet.resources.changes).toEqual(
       {
         mySsmParameter: {
-          oldValue: {
-            Type: 'AWS::SSM::Parameter',
-            Properties: {
-              Name: 'mySsmParameterFromStack',
-              Type: 'String',
-              Value: {
-                Ref: 'SsmParameterValuetestbugreportC9',
-              },
-            },
-          },
-          newValue: {
-            Type: 'AWS::SSM::Parameter',
-            Properties: {
-              Name: 'mySsmParameterFromStack',
-              Type: 'String',
-              Value: {
-                Ref: 'SsmParameterValuetestbugreportC9',
-              },
-            },
-          },
+          oldValue: ssmParam,
+          newValue: ssmParam,
           resourceTypes: {
             oldType: 'AWS::SSM::Parameter',
             newType: 'AWS::SSM::Parameter',
@@ -1331,10 +1305,8 @@ describe('changeset', () => {
               changeImpact: 'NO_CHANGE',
             },
             Value: {
-              oldValue: {
-              },
-              newValue: {
-              },
+              oldValue: {},
+              newValue: {},
               isDifferent: true,
               changeImpact: 'WILL_UPDATE', // this is what changed!
             },
@@ -1414,32 +1386,16 @@ describe('changeset', () => {
     expect(diffWithChangeSet.resources.changes).toEqual(
       {
         Queue: {
-          oldValue: {
-            Type: 'AWS::SQS::Queue',
-            Properties: {
-              QueueName: {
-                Ref: 'SsmParameterValuetestbugreportC9',
-              },
-            },
-          },
-          newValue: {
-            Type: 'AWS::SQS::Queue',
-            Properties: {
-              QueueName: {
-                Ref: 'SsmParameterValuetestbugreportC9',
-              },
-            },
-          },
+          oldValue: sqsQueue,
+          newValue: sqsQueue,
           resourceTypes: {
             oldType: 'AWS::SQS::Queue',
             newType: 'AWS::SQS::Queue',
           },
           propertyDiffs: {
             QueueName: {
-              oldValue: {
-              },
-              newValue: {
-              },
+              oldValue: {},
+              newValue: {},
               isDifferent: true,
               changeImpact: 'WILL_REPLACE', // this is what changed!
             },
@@ -1471,23 +1427,7 @@ describe('changeset', () => {
         },
       },
       Resources: {
-        Queue: {
-          Type: 'AWS::SQS::Queue',
-          Properties: {
-            QueueName: {
-              'Fn::Join': [
-                '',
-                [
-                  'newValue',
-                  {
-                    Ref: 'SsmParameterValuetestbugreportC9',
-                  },
-                ],
-              ],
-            },
-            ReceiveMessageWaitTimeSeconds: 10,
-          },
-        },
+        Queue: sqsQueueWithAargs({ waitTime: 10 }),
       },
     };
 
@@ -1499,23 +1439,7 @@ describe('changeset', () => {
         },
       },
       Resources: {
-        Queue: {
-          Type: 'AWS::SQS::Queue',
-          Properties: {
-            QueueName: {
-              'Fn::Join': [
-                '',
-                [
-                  'newValue',
-                  {
-                    Ref: 'SsmParameterValuetestbugreportC9',
-                  },
-                ],
-              ],
-            },
-            ReceiveMessageWaitTimeSeconds: 20,
-          },
-        },
+        Queue: sqsQueueWithAargs({ waitTime: 20 }),
       },
     };
 
@@ -1562,40 +1486,8 @@ describe('changeset', () => {
     expect(diffWithoutChangeSet.resources.changes).toEqual(
       {
         Queue: {
-          oldValue: {
-            Type: 'AWS::SQS::Queue',
-            Properties: {
-              QueueName: {
-                'Fn::Join': [
-                  '',
-                  [
-                    'newValue',
-                    {
-                      Ref: 'SsmParameterValuetestbugreportC9',
-                    },
-                  ],
-                ],
-              },
-              ReceiveMessageWaitTimeSeconds: 10,
-            },
-          },
-          newValue: {
-            Type: 'AWS::SQS::Queue',
-            Properties: {
-              QueueName: {
-                'Fn::Join': [
-                  '',
-                  [
-                    'newValue',
-                    {
-                      Ref: 'SsmParameterValuetestbugreportC9',
-                    },
-                  ],
-                ],
-              },
-              ReceiveMessageWaitTimeSeconds: 20,
-            },
-          },
+          oldValue: sqsQueueWithAargs({ waitTime: 10 }),
+          newValue: sqsQueueWithAargs({ waitTime: 20 }),
           resourceTypes: {
             oldType: 'AWS::SQS::Queue',
             newType: 'AWS::SQS::Queue',
@@ -1603,26 +1495,10 @@ describe('changeset', () => {
           propertyDiffs: {
             QueueName: {
               oldValue: {
-                'Fn::Join': [
-                  '',
-                  [
-                    'newValue',
-                    {
-                      Ref: 'SsmParameterValuetestbugreportC9',
-                    },
-                  ],
-                ],
+                Ref: 'SsmParameterValuetestbugreportC9',
               },
               newValue: {
-                'Fn::Join': [
-                  '',
-                  [
-                    'newValue',
-                    {
-                      Ref: 'SsmParameterValuetestbugreportC9',
-                    },
-                  ],
-                ],
+                Ref: 'SsmParameterValuetestbugreportC9',
               },
               isDifferent: false,
               changeImpact: 'NO_CHANGE',
@@ -1649,61 +1525,11 @@ describe('changeset', () => {
     );
 
     expect(diffWithChangeSet.differenceCount).toBe(1); // this is the count of how many resources have changed
-    expect(diffWithChangeSet.resources.get('Queue').propertyUpdates).toEqual(
-      {
-        QueueName: {
-          oldValue: {
-          },
-          newValue: {
-          },
-          isDifferent: true,
-          changeImpact: 'WILL_REPLACE',
-        },
-        ReceiveMessageWaitTimeSeconds: {
-          oldValue: 10,
-          newValue: 20,
-          isDifferent: true,
-          changeImpact: 'WILL_UPDATE',
-        },
-      },
-    );
     expect(diffWithChangeSet.resources.changes).toEqual(
       {
         Queue: {
-          oldValue: {
-            Type: 'AWS::SQS::Queue',
-            Properties: {
-              QueueName: {
-                'Fn::Join': [
-                  '',
-                  [
-                    'newValue',
-                    {
-                      Ref: 'SsmParameterValuetestbugreportC9',
-                    },
-                  ],
-                ],
-              },
-              ReceiveMessageWaitTimeSeconds: 10,
-            },
-          },
-          newValue: {
-            Type: 'AWS::SQS::Queue',
-            Properties: {
-              QueueName: {
-                'Fn::Join': [
-                  '',
-                  [
-                    'newValue',
-                    {
-                      Ref: 'SsmParameterValuetestbugreportC9',
-                    },
-                  ],
-                ],
-              },
-              ReceiveMessageWaitTimeSeconds: 20,
-            },
-          },
+          oldValue: sqsQueueWithAargs({ waitTime: 10 }),
+          newValue: sqsQueueWithAargs({ waitTime: 20 }),
           resourceTypes: {
             oldType: 'AWS::SQS::Queue',
             newType: 'AWS::SQS::Queue',
--- a/packages/@aws-cdk/cloudformation-diff/test/util.ts
+++ b/packages/@aws-cdk/cloudformation-diff/test/util.ts
@@ -73,3 +73,35 @@ export function largeSsoPermissionSet() {
     ),
   });
 }
+
+export const ssmParam = {
+  Type: 'AWS::SSM::Parameter',
+  Properties: {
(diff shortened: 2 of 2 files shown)


packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
```diff
@@ -143,13 +143,6 @@ function calculateTemplateDiff(currentTemplate: { [key: string]: any }, newTempl
   return new types.TemplateDiff(differences);
 }
 
-/**
- * Compare two CloudFormation resources and return semantic differences between them
- */
-export function diffResource(oldValue: types.Resource, newValue: types.Resource): types.ResourceDifference {
-  return impl.diffResource(oldValue, newValue);
-}
-
```
bergjaak, line 152: This was never used, so I removed it

packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
```diff
@@ -1231,4 +1231,517 @@ describe('changeset', () => {
...
+          ParameterKey: 'SsmParameterValuetestbugreportC9',
+          ParameterValue: 'goodJob',
+          ResolvedValue: 'changedVal',
+        }],
+      },
+    );
+
+    // THEN
+    expect(diffWithChangeSet.differenceCount).toBe(1);
+    expect(diffWithChangeSet.resources.changes).toEqual(
```
comcalvi, line 1294: we don't need to assert against all of this, can we just assert against `diffWithChangeSet.resources.changes.mySssmParameter.propertyUpdates`?

packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
```diff
@@ -1231,4 +1231,517 @@ describe('changeset', () => {
...
+      },
+    };
+
+    // Template diff misses the change:
+    const diffWithoutChangeSet = fullDiff(currentTemplate, currentTemplate);
+    // THEN
+    expect(diffWithoutChangeSet.differenceCount).toBe(0);
+    expect(diffWithoutChangeSet.resources.changes).toEqual({});
+
+    // AND THEN
```
comcalvi, line 1385: can we do 

```
// GIVEN
// WHEN
// THEN
```

for these tests?

comcalvi (changes requested):
This is a great change! I only have a few comments.

Make sure the title [describes the bug, not the solution](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md?plain=1#L602). 

Can we make the various tests use `propertyUpdates` where possible, like we do [elsewhere in this file](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/cloudformation-diff/test/diff-template.test.ts#L844)? The assertions as written are very long and hard to parse.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -944,6 +945,57 @@ integTest('cdk diff --quiet does not print \'There were no differences\' message
...
+
+    // WHEN
+    const diff = await fixture.cdk(['diff', fixture.fullStackName('queue-name-defined-by-ssm-param')]);
+    `
+      Resources
+      [~] AWS::SQS::Queue DiffFromChangeSetQueue DiffFromChangeSetQueue06622C07 replace
+       └─ [~] QueueName (requires replacement)
+      [~] AWS::SSM::Parameter DiffFromChangeSetSSMParam DiffFromChangeSetSSMParam92A9A723
+       └─ [~] Value
+    `;
```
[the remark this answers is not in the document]
bergjaak: nah, just a comment to show what the output looks like

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -944,6 +945,57 @@ integTest('cdk diff --quiet does not print \'There were no differences\' message
...
+
+    // THEN
+    // the reason these aren't just 1 line is because the terminal output includes colors, which comes up like \u001b[4m\u001b[1mResources\u001b
```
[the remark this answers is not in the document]
bergjaak: I'll give that a try, thanks!

packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
bergjaak: What's the benefit of asserting less? On my previous team, we always asserted as much as possible information

packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
bergjaak: Will do

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
```diff
@@ -229,8 +222,33 @@ function makeAllResourceChangesImports(diff: types.TemplateDiff) {
   });
 }
 
-function filterFalsePositives(diff: types.TemplateDiff, changeSet: DescribeChangeSetOutput) {
+function refineDiffWithChangeSet(diff: types.TemplateDiff, changeSet: DescribeChangeSetOutput, newTemplateResources: {[logicalId: string]: any}) {
```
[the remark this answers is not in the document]
bergjaak: also, I think that's a great article ^. A book that endorses this opinion and that explicitly disagrees with Uncle Bob clean code style (which argues for functions no longer than 4-6 lines) is A Philosophy of Software Design

packages/@aws-cdk/cloudformation-diff/test/diff-template.test.ts
comcalvi: The test assertions are long enough that they're difficult to follow; I have a hard time telling what part of that assertion you care about for this test. There's a lot of noise in each assertion

bergjaak: I think that's fair, I'm willing to make that change

piradeepk: I agree with @bergjaak, I'd rather that we asserted against more than less.

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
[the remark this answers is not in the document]
comcalvi: This is an interesting perspective. That post is generally agreeable; in particular, this:

> if a lot of operations are supposed to happen in a sequential fashion, their code should follow sequentially. 

The rest of the post makes a solid argument that one function called in 20 places is more likely to lead to bugs that are hard to fix than if you just inlined some code in that one spot you need it. That seems to be the general, overarching theme of the whole post; a function written in one place, with one set of assumptions, can do subtle yet catastrophic damage when called in a place with different assumptions. 

This is fair. For the code here, you could argue that it's very unclear if `addMissingProperties` should be called before or after `refineChangeImpact`, since the names give no indication of the correct order. This is a bad contract, since if you call it in the wrong order, it might crash or give you wrong output. 

The only thing I'm not convinced of is this:

> Using large comment blocks inside the major function to delimit the minor functions is a good idea for quick scanning

There's two sides of this for me:

1. Comments explaining what the code is doing encourage poor variable and function naming. 

This is from my personal experience, so I'm curious what your perspective is on this. I've reviewed code that uses very obscure variable names, but had was a brief comment above explaining what was really happening. This isn't about functions or inlining anymore; this is just a short section of code that made an API call and did something with the response. Without the comment, there would be no way of knowing what that code was supposed to do because the variable names chosen communicated what the property was in AWS, but not at all what it was being used for. 

2. Comments get stale. Code that communicates clear intent is far faster to read, understand, and fix a bug in than code that communicates no intent, but even undocumented, uncommented code is easier to understand than code with *wrong* comments. CDK CLI code changes frequently, with little small parts modified here and there. It's easy for a comment at the top of a long code block to become invalid farther down. This is why I always tend to prefer breaking things up into functions.

Changing gears a bit, I think the clearest way to organize functions is based on their input and output types. For this function, we need changesets and the diff the entire time, and there's no clear way to break this by type. 

For me, a quick glance through both halves of this function is not very readable, but this was true both before and after this PR; it's a consequence of mutating the diff object in this way. 

I see the value of keeping it all in one function. I can see one other way that combines both worlds a bit, which we sometimes use in the CDK; create some local-scoped functions that only exist within this function. This means that they can't be called anywhere else, but we can still call them in here to indicate the two halves of this function at a glance. 

Either way (leaving it as you have it, or creating the local helper functions) is fine with me, I'll leave that up to you.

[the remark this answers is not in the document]
bergjaak: for 1 `Comments explaining what the code is doing encourage poor variable and function naming.` I think a team should hold a high standard of not allowing poor naming, or encouraging clarifying and consistent names when possible. I think having comments is a bad excuse for having poor names, especially because of your second point, that `Comments get stale.` I do agree with that; I think that's also a team culture thing, namely that holding PRs accountable for updating comments is important. 

It also makes sense to have a centralized location for comments/documentation, so that you don't have multiple "comment states" to synchronize -- so ideally comments on implementation go right above what's being commented on or on a function/class signature that describes high level (say, business logic) details that are unlikely to change. Also, comments shouldn't explain implementation details but rather business logic that cannot be discerned from the code -- information that can be captured in code ought to be captured in the code itself, since code has lower odds of going stale. And I do think comment readers should always keep in mind the possibility that a comment they're reading is stale. Nonetheless, I am of the opinion that comments can communicate context that otherwise cannot be captured in code, and its precisely that information (cannot be captured in code) that should be commented.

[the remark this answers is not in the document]
bergjaak: I ended up doing the private methods suggestion, as I think that makes both of us happy :)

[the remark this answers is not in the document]
comcalvi: > Also, comments shouldn't explain implementation details but rather business logic that cannot be discerned from the code

I agree with this sentiment, which runs a bit counter to the idea of comment blocks explaining what is happening; but I also see the danger of exposing too many functions that can make non-obvious state mutations.

comcalvi (approved):

blimmer:
It's awesome to get this fixed. Thank you for this PR! 🏆

Issue #29873, opened 2024-04-17
(aws-s3): "s3: PutBucketPolicy" is added when autoDeleteObjects=True, in CDK Version 2.100


khushail:
### Describe the bug

SIM Ticket - V1347352006

This [PR](https://github.com/aws/aws-cdk/pull/26875) added "s3:PutBucketPolicy" when autoDeleteObjects is set to True in CDK Version 2.100

### Expected Behavior

the policy should not be added by default

### Current Behavior

creating a bucket with "autoDeleteObject: true" automatically added "s3:PutBucketPolicy".

### Reproduction Steps

1. Run below  code with CDK 2.99 

```
    const bucket100 = new s3.Bucket(this, 'Bucket100', {
      versioned: true,
      removalPolicy: cdk.RemovalPolicy.DESTROY,
      autoDeleteObjects: true
    });
```
This policy generated -

 ```
   "Bucket": {
     "Ref": "Bucket9970D96479"
    },
    "PolicyDocument": {
     "Statement": [
      {
       "Action": [
        "s3:DeleteObject*",
        "s3:GetBucket*",
        "s3:List*"
       ],
       "Effect": "Allow",
       "Principal": {
        "AWS": {
         "Fn::GetAtt": [
          "CustomS3AutoDeleteObjectsCustomResourceProviderRole3B1BD092",
          "Arn"
         ]
        }
       },
```

2. npm uninstall -g aws-cdk
3. npm cache clean --force
4. npm install -g aws-cdk@2.100
5. Run above code and do 'cdk synth'
6. Policy generated -

```
"Bucket": {
     "Ref": "Bucket1000EAAE573"
    },
    "PolicyDocument": {
     "Statement": [
      {
       "Action": [
        "s3:DeleteObject*",
        "s3:GetBucket*",
        "s3:List*",
        "s3:PutBucketPolicy"
       ],
       "Effect": "Allow",
       "Principal": {
        "AWS": {
         "Fn::GetAtt": [
          "CustomS3AutoDeleteObjectsCustomResourceProviderRole3B1BD092",
          "Arn"
         ]
        }
```

### CDK CLI Version

2.100

### Node.js Version

na

### OS

mac

### Language

TypeScript

SankyRed:
[...]

PR #30096, opened 2024-05-07
docs(s3): update documentation for more context around `autoDeleteObjects`

SankyRed:
### Issue # (if applicable)
[29873](https://github.com/aws/aws-cdk/issues/29873)

### Description of changes
This the documentation update to notify users that upon setting the `autoDeleteObjects` to `true` we would also add a new policy `s3:PutBucketPolicy` to handle race conditions.

### Checklist
-  [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/lib/aws.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/aws.ts
@@ -20,6 +20,7 @@ export class AwsClients {
   public readonly ecr: AwsCaller<AWS.ECR>;
   public readonly ecs: AwsCaller<AWS.ECS>;
   public readonly sso: AwsCaller<AWS.SSO>;
+  public readonly ssm: AwsCaller<AWS.SSM>;
   public readonly sns: AwsCaller<AWS.SNS>;
   public readonly iam: AwsCaller<AWS.IAM>;
   public readonly lambda: AwsCaller<AWS.Lambda>;
@@ -39,6 +40,7 @@ export class AwsClients {
     this.ecs = makeAwsCaller(AWS.ECS, this.config);
     this.sso = makeAwsCaller(AWS.SSO, this.config);
     this.sns = makeAwsCaller(AWS.SNS, this.config);
+    this.ssm = makeAwsCaller(AWS.SSM, this.config);
     this.iam = makeAwsCaller(AWS.IAM, this.config);
     this.lambda = makeAwsCaller(AWS.Lambda, this.config);
     this.sts = makeAwsCaller(AWS.STS, this.config);
--- a/packages/@aws-cdk-testing/cli-integ/package.json
+++ b/packages/@aws-cdk-testing/cli-integ/package.json
@@ -30,12 +30,12 @@
   "license": "Apache-2.0",
   "devDependencies": {
     "@aws-cdk/cdk-build-tools": "0.0.0",
-    "@types/semver": "^7.5.8",
-    "@types/yargs": "^15.0.19",
+    "@aws-cdk/pkglint": "0.0.0",
     "@types/fs-extra": "^9.0.13",
     "@types/glob": "^7.2.0",
     "@types/npm": "^7.19.3",
-    "@aws-cdk/pkglint": "0.0.0"
+    "@types/semver": "^7.5.8",
+    "@types/yargs": "^15.0.19"
   },
   "dependencies": {
     "@octokit/rest": "^18.12.0",
--- a/packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
+++ b/packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
@@ -126,6 +126,22 @@ class SsoInstanceAccessControlConfig extends Stack {
   }
 }
 
+class DiffFromChangeSetStack extends Stack {
+  constructor(scope, id) {
+    super(scope, id);
+
+    const queueNameFromParameter = ssm.StringParameter.valueForStringParameter(this, 'for-queue-name-defined-by-ssm-param');
+    new sqs.Queue(this, "DiffFromChangeSetQueue", {
+      queueName: queueNameFromParameter,
+    })
+
+    new ssm.StringParameter(this, 'DiffFromChangeSetSSMParam', {
+      parameterName: 'DiffFromChangeSetSSMParamName',
+      stringValue: queueNameFromParameter,
+    });
+  }
+}
+
 class ListMultipleDependentStack extends Stack {
   constructor(scope, id) {
     super(scope, id);
@@ -657,6 +673,8 @@ switch (stackSet) {
     new SsoPermissionSetNoPolicy(app, `${stackPrefix}-sso-perm-set-without-managed-policy`);
 
     const failed = new FailedStack(app, `${stackPrefix}-failed`)
+
+    new DiffFromChangeSetStack(app, `${stackPrefix}-queue-name-defined-by-ssm-param`)
 
     // A stack that depends on the failed stack -- used to test that '-e' does not deploy the failing stack
     const dependsOnFailed = new OutputsStack(app, `${stackPrefix}-depends-on-failed`);
--- a/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
+++ b/packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
@@ -1,7 +1,7 @@
 import { promises as fs, existsSync } from 'fs';
 import * as os from 'os';
 import * as path from 'path';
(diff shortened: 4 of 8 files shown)


packages/aws-cdk-lib/aws-s3/README.md
```diff
@@ -622,6 +622,8 @@ as it does not contain any objects.
 To override this and force all objects to get deleted during bucket deletion,
 enable the`autoDeleteObjects` option.
 
+The auto-deletion for objects within S3 buckets is handled by a custom resource. Enabling this feature by setting it to true would additionally include the imposition of the `s3:PutBucketPolicy` to restrict external write access during cleanup, thus mitigating potential race conditions.
```
colifran, line 625: Let's just let this be a continuation of the above text block rather than having it be a standalone. When you do this, we should provide more context about why this being handled by a custom resource is relevant to the policy being added. Right now you mention it being handled by a custom resource but quickly move away from that point. A reader would probably ask themselves how this being handled by a custom resource is related to the `s3:PutBucketPolicy` action. Also, your second sentence can just read, "Enabling this feature would include...". The "setting it true" is already implied by the feature being enabled.

We need to make sure we explain why a `s3:PutBucketPolicy` is being added since this isn't entirely clear upfront especially give that the PR this originated from claims that it, "Adds a DENY policy for S3:PutObject on buckets to be auto-deleted to prevent a race condition on emptying with external bucket writers". This is confusing because `s3:PutBucketPolicy` is not `s3:PutObject` AND `s3:PutBucketPolicy` was added to a policy document with an `Allow` effect, not a `Deny`. I would start with this to understand why `s3:PutBucketPolicy` was added here - https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketPolicy.html

colifran (changes requested):
See my comment. We need to give more context behind why the `s3:PutBucketPolicy` is being added.

SankyRed:
> See my comment. We need to give more context behind why the `s3:PutBucketPolicy` is being added.

Updated the readme to provide more context around why the policy is being added. Please take a look when you get some time @colifran

packages/aws-cdk-lib/aws-s3/README.md
```diff
@@ -622,6 +622,8 @@ as it does not contain any objects.
 To override this and force all objects to get deleted during bucket deletion,
 enable the`autoDeleteObjects` option.
 
+In order to facilitate bucket deletion, we add the `s3:PutBucketPolicy` action from AWS Identity and Access Management (IAM) policies. This action authorizes users to manage bucket policies for Amazon S3 buckets, enabling them to adjust access control settings by either establishing new policies or modifying existing ones. As part of this we update the policy to **Deny** `s3:PutObject` action within the specified bucket preventing new objects from being created in the bucket. Consequently, this measure ensures that during the bucket emptying phase, the onset of potential race conditions is mitigated, as no new objects can be generated within the bucket.
```
colifran, line 625: What I'm still not seeing is why we only added `s3:PutBucketPolicy`. As far as I can tell, there was no `Deny` for `s3:PutObject` added which is the claim in the original PR. Here is one of the snapshot diffs:
image
You can see the only difference is that `s3:PutBucketPolicy` was added with an `Allow` effect. Customers are going to be wondering why a new action was added to the bucket policy that grants more actions to be taken on the bucket. I'm even wondering why that's the case. This is confusing because they claimed that their change, "Adds a DENY policy for S3:PutObject on buckets to be auto-deleted to prevent a race condition on emptying with external bucket writers.", but that isn't at all what changed. My guess is that this was done because if you don't have `PutBucketPolicy` permissions, Amazon S3 returns a 403 Access Denied error which is essentially blocking any writes to the bucket from external bucket writers. This is what needs to be investigated a bit more and explained.

SankyRed: @colifran Yes, the only difference is that `s3:PutBucketPolicy` was added with an `Allow` effect but, that policy is used to grant permissions for putting or updating bucket policies for Amazon S3 buckets. As part of this [code change](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/custom-resource-handlers/lib/aws-s3/auto-delete-objects-handler/index.ts#L46) we retrieve the existing bucket policy for the specified bucket using `getBucketPolicy` method and then add a new statement to `Deny` the `s3:PutObject` as part of this [code](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/custom-resource-handlers/lib/aws-s3/auto-delete-objects-handler/index.ts#L52). We then use the `s3:PutBucketPolicy` to update the policy while cleanup. This runs in the background while enabling the `autoDeleteObjects`. 

But, do we want to specify why that new policy is getting added and what happens if not?

colifran: Ahh okay awesome. That part was missing for me. I would assume this would be confusing for customers too because an update for a customer using `autoDeleteObjects` will just show the `s3:PutBucketPolicy` action being added to the bucket policy in the diff. Without understanding that this is backed by a custom resource they'll likely not understand the point of that action being added. So what we want to clearly convey is that when `autoDeleteObjects` is enabled we add the `s3:PutBucketPolicy` action to the bucket policy. We do this because on bucket deletion the custom resource provider needs to update the bucket policy by adding a DENY policy for s3:PutObject to prevent a race condition on emptying with external bucket writers.

packages/aws-cdk-lib/aws-s3/README.md
```diff
@@ -622,6 +622,8 @@ as it does not contain any objects.
 To override this and force all objects to get deleted during bucket deletion,
 enable the`autoDeleteObjects` option.
 
+When `autoDeleteObjects` is enabled, we include the `s3:PutBucketPolicy` action in the Bucket Policy. This allows users to manage bucket policies, creating new ones or adjusting existing ones for Amazon S3 buckets. We do this because during bucket deletion, the custom resource provider updates the bucket policy by adding a **DENY** effect for `s3:PutObject` action, preventing new objects from being created in the bucket. This helps avoid any race conditions during the deletion process with external bucket writers.
```
colifran, line 625:
```suggestion
When `autoDeleteObjects` is enabled, `s3:PutBucketPolicy` is added to the bucket policy. This is done to allow the custom resource this feature is built on to add a deny policy by for `s3:PutObject` to the bucket policy when a delete stack event occurs. Adding this deny policy prevents new objects from being written to the bucket. Doing this prevents race conditions with external bucket writers during the deletion process.
```

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1471,6 +1471,11 @@ export interface BucketProps {
    * all objects in the bucket being deleted. Be sure to update your bucket resources
    * by deploying with CDK version `1.126.0` or later **before** switching this value to `false`.
    *
+   * Setting the `autoDeleteObjects: true` on a bucket will include `s3:PutBucketPolicy` action
+   * to the Bucket Policy. The reason being upon bucket deletion, the custom resource provider
+   * needs to update the bucket policy by adding a `DENY` policy for `s3:PutObject` to
+   * prevent a race condition on emptying with external bucket writers.
+   *
```
colifran, line 1478:
```suggestion
   * Setting `autoDeleteObjects` to true on a bucket will add `s3:PutBucketPolicy` to the
   * bucket policy. This is because during bucket deletion, the custom resource provider
   * needs to update the bucket policy by adding a deny policy for `s3:PutObject` to
   * prevent race conditions with external bucket writers.
   *
```

colifran:
A few verbiage suggestions and then this will be good to go.

SankyRed:
> A few verbiage suggestions and then this will be good to go.

Updated the docs, thanks @colifran

packages/aws-cdk-lib/aws-s3/README.md
```diff
@@ -622,6 +622,8 @@ as it does not contain any objects.
 To override this and force all objects to get deleted during bucket deletion,
 enable the`autoDeleteObjects` option.
 
+When `autoDeleteObjects` is enabled, `s3:PutBucketPolicy` is added to the bucket policy. This is done to allow the custom resource this feature is built on to add a deny policy by for `s3:PutObject` to the bucket policy when a delete stack event occurs. Adding this deny policy prevents new objects from being written to the bucket. Doing this prevents race conditions with external bucket writers during the deletion process.
```
colifran, line 625:
```suggestion
When `autoDeleteObjects` is enabled, `s3:PutBucketPolicy` is added to the bucket policy. This is done to allow the custom resource this feature is built on to add a deny policy for `s3:PutObject` to the bucket policy when a delete stack event occurs. Adding this deny policy prevents new objects from being written to the bucket. Doing this prevents race conditions with external bucket writers during the deletion process.
```

colifran (approved):

PR #30099, opened 2024-05-08
chore(lambda-nodejs): clean up sdk v2 references and warn about runtime updates related to sdk bundling

### LINKED ISSUE #29836: SDKv3-02: Update `aws-lambda-nodejs` in `aws-cdk-lib`
### Context
The AWS SDK V2 is being deprecated. We need to remove all uses of it from the CDK.

### Considerations
- Should be completed after SDKv3-01 and follow the same/similar plan for deprecation of v2. Consider adding a warning for now that we will remove use of SDKv2 in a few weeks and then make this change then.
- Be aware of the node 20 upgrade happening in this package as well.

### Task Description
We have custom logic to allow users to either use sdk v2 or sdk v3. All custom resources using this package will need to be tested, not just for hash changes, but for deployability. Some manual validation may be required.

The deprecation date for node16 Lambda runtimes is June 12, 2024. Lambda will block function creation on July 16, 2024 and will block function updates on August 15, 2024.

### Acceptance Criteria
1. Remove all other uses of SDKv2.
2. Update relevant integ tests and verify that the impacted custom resources successfully deploy and behave as expected.

scanlonp:
Closes #29836.

While sdk v2 was not being used directly in the construct, we had some remnants of its use. Added a warning for users of Node 16 who do not bundle the sdk on what it will take to update to newer node versions. Also updated integ tests to make actual sdk v3 calls as well as clean up false references to v2.

----


packages/aws-cdk-lib/aws-lambda-nodejs/lib/bundling.ts
```diff
@@ -133,8 +133,15 @@ export class Bundling implements cdk.BundlingOptions {
     // Don't automatically externalize aws sdk if `bundleAwsSDK` is true so it can be
     // include in the bundle asset
     const defaultExternals = props.runtime?.isVariable || props.bundleAwsSDK ? [] : versionedExternals;
+
     const externals = props.externalModules ?? defaultExternals;
 
+    // warn users if they are using a runtime that does not support sdk v2
+    // and the sdk is not explicitly bundled
+    if (externals.length && isV2Runtime) {
```
bergjaak, line 141: what is `externals.length` supposed to mean/indicate?

scanlonp: `externals` are the packages we are explicitly not bundling. `externals.length` should check if that list has anything in it. `externals` will have the sdk package correlated to the runtime by default. Users can override or provide their own external packages.

The thought process behind this is that if a user is setting `bundleAwsSDK = true` or `props.externalModules = []`, then they likely are aware of how the bundling in the module works, and they are including the sdk in their handler code. Both of these will cause `externals.length` to be falsy, and not give the warning since they are avoiding sdk problems by bundling it in.

If they have not set anything, or set certain externals, we give them this warning.

bergjaak (approved):

PR #30108, opened 2024-05-08
feat: implement node runtime region awareness for cdk vended custom resources

colifran:
### Reason for this change

This PR introduces node runtime region awareness into the custom resource provider framework by doing the following:

- Adding a `LATEST_NODE_RUNTIME_MAP` fact table used to maintain the latest node runtime available per AWS partition
- Introducing `determineLatestNodeRuntime` and `determineLatestNodeRuntimeName` functions which determine the latest Lambda node `Runtime` and the latest Lambda node `Runtime` name, respectively
- Updating the custom resource provider framework to utilize these two functions when code generating the runtime property for providers using node runtimes

### Description of changes

This PR can be segmented into the following code changes:

- A `LATEST_NODE_RUNTIME_MAP` fact table was added which maintains the latest Lambda node runtime available per AWS partition.
- Introduced `determineLatestNodeRuntime` and `determineLatestNodeRuntimeName` functions which determine the latest Lambda node `Runtime` and the latest Lambda node `Runtime` name, respectively.
- The existing runtime property being code generated via the custom resource provider framework has been altered to now use the appropriate runtime determiner function, i.e., `determineLatestNodeRuntimeName` for `CustomResourceProvider` or `determineLatestNodeRuntime` for `Function` or `SingletonFunction`. Any custom resource provider using a python runtime will not use either of these functions.
- To consolidate and manage importing of external modules a `ModuleImporter` class has been created. This class allows external modules to be registered as an import for a target module and prevents duplicate imports for modules that contain multiple framework components. This class also provides the ability to specify different import paths which allows all external modules to be consolidated into a single class rather than having duplicate external modules defined for different import paths. Lastly, this class determines whether or not external modules should be imported selectively or if all targets in the external module should be imported under an alias, i.e., `import { Function } from 'aws-lambda'` vs. `import * as lambda from 'aws-lambda'`
- A `CallableExpr` class was created to allow expression proxies to be created from a specified expression name. This allows the new runtime determiner functions and other module specific functions to be called from their specified module and built into a JavaScript object that will mirror the JavaScript operations done to it in an expression tree.

Manually tested all individual custom resource handlers for create, update, and delete. I verified that the behavior seen was what was expected based off of the handler code. All impacted integ tests were updated. New unit tests were added to the custom resource handler framework to test that generated code correctly included `determineLatestNodeRuntime` for Lambda based handlers and `determineLatestNodeRuntimeName` for `CfnResource` based handlers. Added new unit tests to test the functionality of `determineLatestNodeRuntime` and `determineLatestNodeRuntimeName` for region agnostic stacks, stacks in commercial region, stacks in China regions, stacks in ADC regions, and stacks in GovCloud regions.

### Checklist

----


packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -215,9 +227,7 @@ export abstract class HandlerFrameworkClass extends ClassType {
         });
         getOrCreateMethod.addParameter({
           name: 'props',
-          type: scope.coreInternal
-            ? CORE_INTERNAL_CR_PROVIDER.CustomResourceProviderOptions
-            : CORE_MODULE.CustomResourceProviderOptions,
```
colifran, line 220: This is one example of what the `ModuleImporter` class helps with. We can maintain one module defined, i.e., `CORE_MODULE`, with all of it's types, expressions, etc. rather than needing to define a duplicate module based on where the module would be imported from. Instead, when registering an import we can just register the same module with a different `fromLocation` specified. Now the code looks a lot cleaner.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -271,57 +277,15 @@ export abstract class HandlerFrameworkClass extends ClassType {
...
-      }
-      case CORE_INTERNAL_STACK.fqn: {
-        CORE_INTERNAL_STACK.importSelective(scope, ['Stack']);
-        return;
-      }
-      case LAMBDA_MODULE.fqn: {
-        LAMBDA_MODULE.import(scope, 'lambda');
-        return;
-      }
-    }
```
colifran, line 321: Lots of reduced code here due to `ModuleImporter` class managing imports

colifran, line 321: Lots of reduced code here due to `ModuleImporter` class managing imports

packages/aws-cdk-lib/region-info/lib/fact.ts
```diff
@@ -195,6 +195,11 @@ export class FactName {
    */
   public static readonly SAML_SIGN_ON_URL = 'samlSignOnUrl';
 
+  /**
+   * The latest Lambda NodeJS runtime available in a given region.
+   */
+  public static readonly LATEST_NODE_RUNTIME = 'latestNodeRuntime';
```
rix0rrr, line 201: I also would expect this version to be used for `lambda.Runtime.LATEST_NODEJS`

packages/aws-cdk-lib/custom-resources/lib/provider-framework/provider.ts
```diff
@@ -8,6 +8,7 @@ import * as ec2 from '../../../aws-ec2';
 import * as iam from '../../../aws-iam';
 import * as kms from '../../../aws-kms';
 import * as lambda from '../../../aws-lambda';
+import { determineLatestNodeRuntime } from '../../../aws-lambda/lib/runtime-determiner-lambda.generated';
```
rix0rrr, line 11: Explain to me why this code needed to be generated and couldn't just be a single handwritten function?

As far as I can tell, there are no parameters that go into the generation?

rix0rrr (commented):
> The overall name of the framework was changed from HandlerFramework to ProviderFramework. This is because the framework is being used to code generate custom resource providers. The custom resource providers execute handler code that this framework has the responsibility of bundling, but calling this a HandlerFramework isn't accurate. 

Not in fact true, as far as I'm aware. We call it "custom resource handlers" etc, but I'm pretty sure it also contains Lambda Handlers that don't go into custom resources. Yes, 90% of them are custom resources, but not 100%. So calling this all "custom resources" is a mental shortcut.

packages/aws-cdk-lib/custom-resources/lib/provider-framework/provider.ts
colifran: Going to hand write the two functions `determineLatestNodeRuntime` and `determineLatestNodeRuntimeName` as discussed offline.

MrArnoldPalmer (commented):
Lot's of good improvements to the codegen here that makes it more robust. I'll leave others to approve but I think this handles the move back to a "deploy time" computed runtime value well. 👍🏻

scanlonp (commented):
Looking good, would like to go over it with you in more depth. Only small thing I noticed is we should add `isVariable: true` to the runtime created by the `determineLatestNodeRuntime()` function in `runtime.ts`.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -373,4 +356,30 @@ export abstract class HandlerFrameworkClass extends ClassType {
...
+
+    if (runtime) {
+      return isCustomResourceProvider ? expr.lit(runtime) : expr.directCode(toLambdaRuntime(runtime));
+    }
+
+    if (isCustomResourceProvider) {
+      scope.registerImport(CORE_MODULE, {
+        targets: [CORE_MODULE.determineLatestNodeRuntimeName],
+        fromLocation: scope.isCoreInternal
+          ? '../../custom-resource-provider'
```
TheRealAmazonKendra, line 371: I'm seeing paths hardcoded a lot. Can we turn most of these into a const?

TheRealAmazonKendra: We could unintentionally break ourselves if we ever move packages around.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/modules.ts
```diff
@@ -1,61 +1,80 @@
...
+  public readonly determineLatestNodeRuntime = CallableExpr.fromName(this, 'determineLatestNodeRuntime');
 
-class CoreInternalCustomResourceProvider extends ExternalModule {
-  public readonly CustomResourceProviderBase = Type.fromName(this, 'CustomResourceProviderBase');
-  public readonly CustomResourceProviderOptions = Type.fromName(this, 'CustomResourceProviderOptions');
+  public readonly importAs = 'lambda';
 
   public constructor() {
-    super('../../custom-resource-provider');
+    super('../../../aws-lambda');
```
TheRealAmazonKendra, line 62: All the hardcoded paths continue to worry me.

packages/@aws-cdk/custom-resource-handlers/test/custom-resources-framework/expected/node-runtime/custom-resource-provider-core.ts
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable prettier/prettier,max-len */
```
TheRealAmazonKendra, line 1: NO! BAD! Obey prettier.

TheRealAmazonKendra: Wait, are these generated?

packages/@aws-cdk/custom-resource-handlers/test/custom-resources-framework/expected/python-runtime/custom-resource-provider-core.ts
```diff
@@ -26,7 +26,7 @@ export class TestProvider extends CustomResourceProviderBase {
     super(scope, id, {
       ...props,
       "codeDirectory": path.join(__dirname, 'my-handler'),
-      "runtimeName": "nodejs18.x"
+      "runtimeName": "python3.10"
```
TheRealAmazonKendra, line 29: node to python? Seems sketch.

TheRealAmazonKendra: Oh wait, this is a test for python runtimes. How did we end up with node here? Ooooops.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/callable-expr.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import { $E, Expression, IScope, ThingSymbol, expr } from '@cdklabs/typewriter';
+
+/**
+ * A class representing an expression proxy which builds a JavaScript object that
+ * will mirror the JavaScript operations done to it in an expression tree.
+ */
+export class CallableExpr {
```
TheRealAmazonKendra, line 8: This is a comment for almost all the files in this folder. I'm not seeing a lot of tests here. I'd like to see more unit tests added.

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3979,6 +3979,199 @@ describe('VPC configuration', () => {
...
+
+    // THEN
+    Template.fromStack(stack).hasResource('AWS::Lambda::Function', {
+      Properties: {
+        Runtime: 'nodejs18.x',
+      },
+    });
+  });
+
+  test('with stack in govcloud region', () => {
```
TheRealAmazonKendra, line 4155: A test case for ADC is missing

TheRealAmazonKendra: Actually below I see that this is actually for ADC. GovCloud is the one that is actually missing here.

packages/aws-cdk-lib/core/lib/custom-resource-provider/custom-resource-provider.ts
```diff
@@ -132,5 +138,14 @@ function customResourceProviderRuntimeToString(x: CustomResourceProviderRuntime)
       return 'nodejs16.x';
     case CustomResourceProviderRuntime.NODEJS_18_X:
       return 'nodejs18.x';
+    case CustomResourceProviderRuntime.NODEJS_20_X:
```
TheRealAmazonKendra, line 141: Can we either add deprecation to the older runtimes or remove them completely as you can no longer create or update functions that contain them (at least for 12)? https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-deprecated

packages/aws-cdk-lib/core/test/custom-resource-provider/custom-resource-provider.test.ts
```diff
@@ -459,3 +462,201 @@ describe('custom resource provider', () => {
...
+
+    // THEN
+    Template.fromStack(stack).hasResource('AWS::Lambda::Function', {
+      Properties: {
+        Runtime: 'nodejs18.x',
+      },
+    });
+  });
+
+  test('with stack in govcloud region', () => {
```
TheRealAmazonKendra, line 627: ADC !== GovCloud

packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
```diff
@@ -3453,3 +3453,11 @@ export const PARTITION_SAML_SIGN_ON_URL: Record<Partition, string> = {
   [Partition.UsIso]: 'https://signin.c2shome.ic.gov/saml',
   [Partition.UsIsoB]: 'https://signin.sc2shome.sgov.gov/saml',
 };
+
+export const LATEST_NODE_RUNTIME_MAP: Record<Partition, string> = {
+  [Partition.Default]: 'nodejs20.x',
```
TheRealAmazonKendra, line 3458: Default cannot be node20, it must be node18

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
TheRealAmazonKendra: I would also like to see a test for a region that we don't currently have in the map to show that we are properly defaulting to node18 on those.

TheRealAmazonKendra (changes requested):
On the first pass through this is looking pretty good. Please see my comments inline. Also a had a whole bunch of questions inline as well.

packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
colifran: This is not named well but it is not the default. `Partition.Default` is for commercial regions. Our true default is set as node18 in the `regionalMap` API.

packages/@aws-cdk/custom-resource-handlers/test/custom-resources-framework/expected/node-runtime/custom-resource-provider-core.ts
colifran: Yes, these are generated by typewriter..

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
colifran: Agreed. Will move these into constants. Probably as part of the defined modules.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -86,22 +93,26 @@ export abstract class HandlerFrameworkClass extends ClassType {
...
       public constructor() {
         super(scope, {
           name: props.name,
           extends: LAMBDA_MODULE.Function,
           export: true,
         });
 
-        this.importExternalModulesInto(scope);
+        if (scope.isAlphaModule) {
+          scope.registerImport(LAMBDA_MODULE, { fromLocation: 'aws-cdk-lib/aws-lambda' });
```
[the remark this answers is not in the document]
colifran: Correct. The import path changes based on where the file will be compiled. Since alpha modules have a dependency on aws-cdk-lib we need to use that as the import path.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/callable-expr.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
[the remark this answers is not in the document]
colifran: Extraneous because they are `devDependencies` and eslint will be mad at us.

packages/@aws-cdk/aws-amplify-alpha/test/integ.app-asset-deployment.js.snapshot/cdkamplifyappassetdeploymentcomamazonawscdkcustomresourcesamplifyassetdeploymentprovider02396C99.nested.template.json
```diff
@@ -1,4 +1,113 @@
 {
+ "Mappings": {
+  "LatestNodeRuntimeMap": {
+   "af-south-1": {
+    "value": "nodejs20.x"
```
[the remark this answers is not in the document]
colifran: Correct. If the region is known then we do the lookup and hardcode the runtime. Otherwise, we defer to deploy time with this mapping.

packages/@aws-cdk/custom-resource-handlers/test/custom-resources-framework/expected/python-runtime/custom-resource-provider-core.ts
colifran: Yeah I also was confused but this is just GitHub showing the diff weird.

packages/aws-cdk-lib/core/lib/custom-resource-provider/custom-resource-provider.ts
colifran: They're marked as deprecated on the `CustomResourceProviderRuntime` enum.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/modules.ts
colifran: Agree with the ones used in the `classes.ts`, but these are just used to initialize the `fqn` property in the super class for `ExternalModules`. This would then be used as `this.fqn` for the specific module. That said, I think paths used to initialize the super are okay. I moved everything else into a constant.

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
colifran: Yes. Good suggestions. Updated with new unit tests. We now have unit tests for commercial regions, adc regions, china regions, govcloud regions, and unsupported regions to test that the default runtime is used.

packages/aws-cdk-lib/core/test/custom-resource-provider/custom-resource-provider.test.ts
colifran: Yeah. Didn't fully understand the difference in region formatting. Updated.

packages/aws-cdk-lib/region-info/build-tools/fact-tables.ts
colifran: `Partition.Default` refers to the default partition not the default runtime.

packages/@aws-cdk/custom-resource-handlers/test/custom-resources-framework/expected/python-runtime/custom-resource-provider-core.ts
colifran: I created new tests specific to python runtimes and node runtimes. I'm not sure why the diff is showing up like this.

packages/@aws-cdk/aws-amplify-alpha/scripts/airlift-custom-resource-handlers.sh
```diff
@@ -4,20 +4,32 @@ scriptdir=$(cd $(dirname $0) && pwd)
 customresourcedir=$(node -p "path.dirname(require.resolve('@aws-cdk/custom-resource-handlers/package.json'))")
 awscdklibdir=${scriptdir}/..
 
-list_custom_resources() {
-  for file in $customresourcedir/dist/aws-amplify-alpha/*/index.js; do
-    echo $file | rev | cut -d "/" -f 2-4 | rev
-  done
+function airlift() {
```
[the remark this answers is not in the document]
colifran: Added docstrings and a comment at the top of the file explaining what it does

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/callable-expr.ts
colifran: Additional unit tests added.

packages/@aws-cdk/aws-amplify-alpha/test/integ.app-asset-deployment.js.snapshot/cdkamplifyappassetdeploymentcomamazonawscdkcustomresourcesamplifyassetdeploymentprovider02396C99.nested.template.json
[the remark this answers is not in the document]
TheRealAmazonKendra: Excellent

TheRealAmazonKendra (changes requested):
What's still missing here is a default in the map so that if someone attempts to deploy to a region we have not yet added and they are not hardcoding that region in their stack configuration, they are not blocked by us.

colifran:
@TheRealAmazonKendra Looks like CloudFormation support for default values for `FindInMap` is supported by us. I've updated the code to support the default in the deploy time mapping if a region isn't detected.

TheRealAmazonKendra (approved):

Issue #25114, opened 2023-04-13
core: CrossRegionExportReader: Response object is too long


antoncohen:
### Describe the bug

When using `crossRegionReferences: true` the `cdk deploy` fails in the reader region with `Response object is too long`.

Similar to #23958, but on the reader side, likely with a different root cause.

### Expected Behavior

I expect `crossRegionReferences: true` to work, and it be possible to reference variables across stacks in different regions.

### Current Behavior

When referencing a variable in a stack in a different region, it fails in the reader region with this error:

```
testing-cdk-cross-region-references-as-example-for-github-issue-rootstack-us-west-2 |  3/14 | 12:28:01 PM | CREATE_COMPLETE      | AWS::Lambda::Function           | Custom::CrossRegionExportReaderCustomResourceProvider/Handler (CustomCrossRegionExportReaderCustomResourceProviderHandler46647B68)
[12:28:03] Stack testing-cdk-cross-region-references-as-example-for-github-issue-rootstack-us-west-2 has an ongoing operation in progress and is not stable (CREATE_IN_PROGRESS)
testing-cdk-cross-region-references-as-example-for-github-issue-rootstack-us-west-2 |  3/14 | 12:28:06 PM | CREATE_IN_PROGRESS   | Custom::CrossRegionExportReader | ExportsReader/Resource/Default (ExportsReader8B249524)
[12:28:08] Stack testing-cdk-cross-region-references-as-example-for-github-issue-rootstack-us-west-2 has an ongoing operation in progress and is not stable (CREATE_IN_PROGRESS)
[12:28:13] Stack testing-cdk-cross-region-references-as-example-for-github-issue-rootstack-us-west-2 has an ongoing operation in progress and is not stable (CREATE_IN_PROGRESS)
testing-cdk-cross-region-references-as-example-for-github-issue-rootstack-us-west-2 |  3/14 | 12:28:15 PM | CREATE_FAILED        | Custom::CrossRegionExportReader | ExportsReader/Resource/Default (ExportsReader8B249524) Response object is too long.
	new CustomResource (/path/to/code/node_modules/aws-cdk-lib/core/lib/custom-resource.ts:135:21)
	\_ new ExportReader (/path/to/code/node_modules/aws-cdk-lib/core/lib/custom-resource-provider/cross-region-export-providers/export-reader-provider.ts:62:27)
[...]

Issue #23958, opened 2023-02-01
CrossRegionExportWriter: Response object is too long.


joshbalfour:
### Describe the bug

When using cross-region references between stacks I get this error:
`(The following resource(s) failed to update: [ExportsWritereuwest25F52B32F9AB3BFFD]. ): Response object is too long.`

This might be because I'm exporting resources from a nested stack so the generated keys are really long.

I'm aware cross-region references are very alpha, and I have a workaround I'm going to try that basically replicates the functionality but does one resource reference at a time, not ideal as it will be much slower.

### Expected Behavior

The reference to work as expected.

### Current Behavior

The stack fails to deploy.

### Reproduction Steps

The resource when synthed looks like this:

```json
"ExportsWritereuwest25F52B32F9AB3BFFD": {
   "Type": "Custom::CrossRegionExportWriter",
   "Properties": {
    "ServiceToken": {
     "Fn::GetAtt": [
      "CustomCrossRegionExportWriterCustomResourceProviderHandlerD8786E8A",
      "Arn"
     ]
    },
    "WriterProps": {
     "region": "eu-west-2",
     "exports": {
      "/cdk/exports/companyproductjosh22acc66077BB2/companyproductjosh22apsoutheast2A61AF646apsoutheast2FnGetAttoryNestedStackoryNestedStackResource908720F1Outputscompanyproductjosh22apsoutheast2oryhydrahydrapublicB75FC865RefEA95D68D": {
       "Fn::GetAtt": [
        "oryNestedStackoryNestedStackResource908720F1",
        "Outputs.companyproductjosh22apsoutheast2oryhydrahydrapublicB75FC865Ref"
       ]
      },
      "/cdk/exports/companyproductjosh22acc66077BB2/companyproductjosh22apsoutheast2A61AF646apsoutheast2FnGetAttoryNestedStackoryNestedStackResource908720F1Outputscompanyproductjosh22apsoutheast2oryhydraadhydraadpublic8B011CFARef0C11F132": {
       "Fn::GetAtt": [
        "oryNestedStackoryNestedStackResource908720F1",
        "Outputs.companyproductjosh22apsoutheast2oryhydraadhydraadpublic8B011CFARef"
       ]
      },
[...]

PR #30115, opened 2024-05-08
chore: document reason and mitigation for cross region reference failure from response objects being too large

colifran:
### Issue

#23958, #25114

### Reason for this change

Using the `crossRegionReference` flag on `StackProps` can produce deployment errors associated with response objects being too large. The root cause of this is a CloudFormation limitation that restricts the total size of a custom resource provider response body to [4096 bytes](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html). Using `crossRegionReference` will create a custom resource in both the producing stack and the consuming stack. The custom resource provider in the producing stack will respond back to CloudFormation with a response object that includes all exported parameter names and values. Similarly, the consuming stack will respond back to CloudFormation with a response object that includes all imported parameter names and values. The parameter names are created with the following naming format: `/cdk/exports/${consumingStackName}/${export-name}`. Users need to be careful about stack names and the use of nested stacks to limit the overall length of parameter names which will limit the size of the response body. This PR adds documentation to give more context about this issue and ways to help mitigate it.

### Description of changes

Added documentation.

### Checklist

----


packages/aws-cdk-lib/README.md
```diff
@@ -268,7 +268,15 @@ In order to mimic strong references, a Custom Resource is also created in the co
...
-See the [adr](https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/core/adr/cross-region-stack-references)
+> [!NOTE]
+> As a consequence of this feature being built on a Custom Resource, we are restricted to a
+> CloudFormation response body size limitation of [4096 bytes](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html).
+> To prevent deployment errors related to the Custom Resource Provider response body being too
+> large, we recommend limiting the use of nested stacks and minimizing the length of stack names.
+> Doing this will prevent SSM parameter names from becoming too long which will reduce the size of the
+> response body.
+
+See the [adr](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/adr/cross-region-stack-references)
```
scanlonp, line 279:
```suggestion
See the [adr](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/adr/cross-region-stack-references.md)
```

packages/aws-cdk-lib/README.md
```diff
@@ -268,7 +268,15 @@ In order to mimic strong references, a Custom Resource is also created in the co
 stack which marks the SSM parameters as being "imported". When a parameter has been successfully
 imported, the producing stack cannot update the value.
 
-See the [adr](https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/core/adr/cross-region-stack-references)
+> [!NOTE]
+> As a consequence of this feature being built on a Custom Resource, we are restricted to a
```
scanlonp, line 272:
```suggestion
> As a big consequence of this feature being built on a Custom Resource, we are restricted to a
```

scanlonp (approved):

PR #30140, opened 2024-05-10
feat(stepfunctions-tasks): add cpu and memory parameters to EcsRunTask

### LINKED ISSUE #30027: aws-stepfunctions-tasks: Add cpu and memory parameters to EcsRunTask
### Describe the feature

Add cpu and memory parameters to EcsRunTask in StepFunctions to allow flexible allocation of Ecs resources according to the situation.

### Use Case

This is useful when you want to allocate resources flexibly according to the data set used in EcsTask, or when you want to increase memory when retrying from errors such as OOM.

### Proposed Solution

Add cpu and memory parameters to EcsRunTask

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.139.1

client:  macOS,Windows,Linux

kackyt:
### Issue # (if applicable)

Closes #30027 .

### Reason for this change

As described in the issue.

### Description of changes

Add cpu and memoryMiB property to EcsRunTaskProps.

Add unit tests.

### Checklist

----


kackyt:
Exemption Request : Integration testing is not required as the changes are minor.
> Features must contain a change to an integration test file and the resulting snapshot.

kackyt:
Review please!

pahud:
Hi

This PR has been pending for community review for a while. Please consider posting it on the #contributing channel in cdk.dev. Community members will be on the lookout there as well for possible reviews.

Check [How to get your P2 PR community reviewed](https://github.com/aws/aws-cdk/wiki/CDK-Community-PR-Reviews#how-to-get-your-p2-pr-community-reviewed) for more details.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
```diff
@@ -91,6 +91,20 @@ export interface EcsRunTaskProps extends sfn.TaskStateBaseProps {
    * @default false
    */
   readonly enableExecuteCommand?: boolean;
+
+  /**
+   * Cpu setting override
+   *
+   * @default - No override
+   */
+  readonly cpu?: string;
```
tmokmss, line 100: Why not use number type to be consistent with [other similar props](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions_tasks.ContainerOverride.html#cpu)?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
```diff
@@ -436,26 +450,33 @@ export class EcsRunTask extends sfn.TaskStateBase implements ec2.IConnectable {
   }
 }
 
-function renderOverrides(containerOverrides?: ContainerOverride[]) {
-  if (!containerOverrides || containerOverrides.length === 0) {
+function renderOverrides(cpu?: string, memoryMiB?: string, containerOverrides?: ContainerOverride[]) {
```
tmokmss, line 453: Given that there are [many more props](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskOverride.html) to be overriden, shouldn't we use an interface to pass these arguments instead of positional arguments?

e.g. 
```ts
function renderOverrides(props: OverrideProps) 

interface OverrideProps {
 cpu?: number;
 memoriMib?: number;
 containerOverrides?: ContainerOverride[];
}
```

tmokmss (commented):
Thanks for the PR! I left some comments but otherwise LGTM.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
kackyt: @tmokmss 
In the AWS API and task definitions, cpu is string type, and changing it to number type may cause unexpected behavior.

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts#L157

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
kackyt: Thank you. I'll fix it.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
tmokmss: > may cause unexpected behavior.
Not really, we can just stringfy it after receiving the value. All the existing code will remain unchanged.

Then, the problem is whether we should use string or number. I'm not sure why the current code is inconsistent (e.g. TaskDefinition uses string, while FargateTaskDefinition uses number). Maybe looking through git history will help to find why. If there is no particular reason, I'd prefer number type because the existing ContainerOverride interface also uses number, which is closer to this interface than TaskDefinition construct.

kackyt: All right. I'll fix ix.

kackyt: @tmokmss 
I tried to fix it, but simply changing to number type failed the integration test. 
```
Failed resources:
aws-sfn-tasks-ecs-fargate-run-task | 5:48:49 PM | UPDATE_FAILED        | AWS::StepFunctions::StateMachine      | StateMachine (StateMachine2E01A3A5) Resource handler returned message: "Invalid State Machine Definition: 'SCHEMA_VALIDATION_FAILED: The value for the field 'Cpu' must be a STRING at /States/FargateTask/Parameters' (Service: AWSStepFunctions; Status Code: 400; Error Code: InvalidDefinition; Request ID: ed6dadc7-f05a-474e-abfa-bc5291934ab8; Proxy: null)" (RequestToken: e1181314-190c-640a-43ca-ec6ccd123530, HandlerErrorCode: InvalidRequest)
```

Where in the code do you think we should stringfy this value?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
kackyt: done 8bb2d45

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
kackyt: done 7061480

tmokmss: renderOverrides is where CFn template is actually rendered, so you can stringify the values there. To stringify the value, I'd recommend to look at the existing code such as FargateContainerDefinition. You have to take care of CDK tokens.

tmokmss: Sorry, I now find that they are already inconsistent at CFn level... [ContainerOverride](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerOverride.html) uses integer as cpu while [TaskOverride](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskOverride.html) uses string as cpu. Given that, I think it's better to just follow CFn type definitions, because in the future CFn may start to use an actual string that cannot be parsed to a number. What do you think? The below API is so weird but we won't override two cpu props at once anyway...

```ts
const runTask = new tasks.EcsRunTask(this, 'Run', {
  // ...
  cpu: '1024',
  containerOverrides: [
    cpu: 1024,
  ]
});
```

tmokmss: We cannot change the existing type definition, so we have to choose which one is better:

```ts
// option 1: weird but maybe safer
const runTask = new tasks.EcsRunTask(this, 'Run', {
  // ...
  cpu: '1024',
  containerOverrides: [
    cpu: 1024,
  ]
});

// option 2: more natural, but prone to future breaking changes
const runTask = new tasks.EcsRunTask(this, 'Run', {
  // ...
  cpu: 1024,
  containerOverrides: [
    cpu: 1024,
  ]
});
```

I now prefer option 1, but anyway I'll approve the review once the all the PR check passes, and let the core team decide it. Any thoughts?

kackyt: Done (option 1) [8c79da9](https://github.com/aws/aws-cdk/pull/30140/commits/8c79da968b328cf4429668afccfd5ed416edbb36)

tmokmss (approved):
To the maintaners: can you confirm the decision made [here](https://github.com/aws/aws-cdk/pull/30140#discussion_r1705206227)? Or let us know if there are other options 🙏

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
```diff
@@ -436,26 +455,40 @@ export class EcsRunTask extends sfn.TaskStateBase implements ec2.IConnectable {
...
-      Cpu: override.cpu,
-      Memory: override.memoryLimit,
-      MemoryReservation: override.memoryReservation,
-      Environment:
-        override.environment?.map((e) => ({
-          Name: e.name,
-          Value: e.value,
-        })),
-    });
+  if (!noContainerOverrides) {
```
shikha372, line 472: can we keep this same as before `!containerOverrides || containerOverrides.length === 0`,  it was more readable than current negation of negation.

shikha372:
Thanks @kackyt for submitting the PR, want to be double sure that we're not adding the functionality that can already be achieved with the help of container override. Could you help in providing a specific use case of ECS run task where this was not working for you. If you've a code snippet that you're using that would help too.

kackyt:
@shikha372 

This is useful when you want to allocate resources flexibly according to the data set used in EcsTask, or when you want to increase memory when retrying from errors such as OOM.Furthermore, in Fargate Task, this parameter is required to specify the size. We cannot change the size of the Fargate with container override.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
kackyt: @shikha372 
I am sorry. However, it is a complex branch and we do not want to clone the same expression. Any suggestions on how to write better code would be appreciated.

shikha372: Thanks @kackyt , looked at it my initial thought was it would be `!containerOverrides || containerOverrides.length === 0` but seems like not, it would be something like `if (containerOverrides || containerOverrides.length !== 0)` in which case its a negation anyways, i would go with the current one.

shikha372:
Thanks @kackyt, checked the differences with the deployment, i'm not sure i completely agree with statement that it is not supported in Fargate Cluster, I see this getting populated in the state machine definition for both Ec2 and Fargate, let me know if i'm missing something here. Here is the code snippet: 

```
        "TaskDefinition": "ECREc2StackTD10AFFC0F",
        "Overrides": {
          "ContainerOverrides": [
            {
              "Name": "TheContainer",
              "Cpu": 512,
              "Memory": 1024,
              "Environment": [
                {
                  "Name": "SOME_KEY",
                  "Value.$": "$.SomeKey"
                }
              ]
            }
          ]
        },
        "LaunchType": "EC2"
```

```
        "Overrides": {
          "ContainerOverrides": [
            {
              "Name": "TheContainer",
              "Cpu": 512,
              "Memory": 1024,
              "Environment": [
                {
                  "Name": "SOME_KEY",
                  "Value.$": "$.SomeKey"
                }
              ]
            }
          ]
        },
        "LaunchType": "FARGATE",
        "PlatformVersion": "1.4.0"
```
That being said, adding these values help to override the cpu and memory set at task definition level which i don't think exist today, and would review PR for that requirement.

Task override: 
```
        "Overrides": {
          "Cpu": "1024",
          "Memory": "2048",
          "ContainerOverrides": [
            {
              "Name": "TheContainer",
              "Environment": [
                {
                  "Name": "SOME_KEY",
                  "Value.$": "$.SomeKey"
                }
              ]
            }
          ]
        },
        "LaunchType": "FARGATE",
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
```diff
@@ -90,6 +90,20 @@ export interface EcsRunTaskProps extends sfn.TaskStateBaseProps {
    * @default false
    */
   readonly enableExecuteCommand?: boolean;
+
+  /**
+   * Cpu setting override
+   *
```
shikha372, line 96:
```suggestion
   * @see https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskOverride.html
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
```diff
@@ -90,6 +90,20 @@ export interface EcsRunTaskProps extends sfn.TaskStateBaseProps {
...
+   * @default - No override
+   */
+  readonly cpu?: string;
+
+  /**
+   * Memory setting override
+   *
```
shikha372, line 103:
```suggestion
   * @see https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskOverride.html
```

shikha372 (approved):

PR #30157, opened 2024-05-11
feat(lambda-python-alpha): cache python lambda dependencies usig lambda layer

### LINKED ISSUE #23829: (aws-lambda-python): cache Docker layer with dependencies
### Describe the feature

Bundling Python Lambdas that contain `requirements.txt`, `Pipfile`, or `poetry.lock` file happens in a Docker container. Firstly, the `requirements.txt` file is generated (for pipenv and Poetry), and then dependencies are installed.

However, this happens each time any code change is made. Each time you change Lambda code (not its dependencies), the Docker build performs the above steps, downloading libraries from the internet.

This is the responsible code:

https://github.com/aws/aws-cdk/blob/4613886eda2c06f75f17768bf47b4d7f14d52e4a/packages/%40aws-cdk/aws-lambda-python/lib/bundling.ts#L109-L122

Dependencies change far less often than the code. Best practices for building in Docker are to firstly download dependencies and only then copy the code. This allows the dependencies layer to be cached, and on consecutive runs, only the code is updated while the dependencies layer is cached.

### Use Case

This will greatly reduce consecutive Lambda bundling times, as dependencies will be fetched from the internet only when they change. When only the code changes, a cached Docker layer with dependencies will be used.

### Proposed Solution

In short, bundling Python Lambda should be changed from:

1. Copy all source files (line 113)
2. Generate `requirements.txt` (line 115)
3. Install dependencies (line 117)

to:

1. Copy dependencies files (i.e. `poetry.lock`)
2. Generate `requirements.txt`
3. Install dependencies
4. Copy the rest of the files

### Other Information

I am willing to implement it after greenlighting by the CDK team.

### Acknowledgements

### CDK version used

2.61.1

any

orshemtov:
### Issue # (if applicable)

Closes #23829

### Reason for this change

cdk synth takes very long to synthesize a `PythonFunction` construct, because the dependencies are getting installed regardless if there was any changes made to them.

The dependencies, whether they are specified in a `requirements.txt` file, a `poetry.lock` file or a `pipenv.lock` file,  get installed as part of the `CMD` in the bundling phase, meaning that we don't use docker's cache or any other caching mechanism and re-install the dependencies from the internet every time.

Trying to compute a custom `assetHash` based on the dependencies file won't work, because of where the call for install is currently placed (in the `CMD` of the bundling phase container)

This causes deployment times to rise significantly, in some cases from minutes to hours.

### Description of changes

The PythonFunction construct in `function.ts` will introduce a new prop: `layer` which is defined as:

```ts
/**
   * Whether or not to create a layer for the function's dependencies.
   * @default - No layer is created.
   */
  readonly layer?: boolean;
```

If `layer` is true, we create the layer before the constructor of `PythonFunction` is called, and attach the layer to PythonFunction in the constructor's `layers` prop.

To control whether dependencies are installed during the Bundling phase, a new prop is introduced to  `BundlingOptions` and is defined as:

```ts
  /**
   * Whether or not to install the dependencies
   * @default true
   */
  readonly installDependencies?: boolean;
```

Since bundling is used for the PythonLayerVersion and PythonFunction props, this prop defaults to true, to not alter any behavior of PythonLayerVersion.

When initializing a PythonFunction, if `layer` is true, and we are creating a layer, the `installDependencies` is set to true for the layer creation, and false for the constructor of PythonFunction, so dependencies will only be installed in the layer.

To compute the assetHash for the layer, I added a `dependenciesHash(entry)` method inside the `Packaging` class, with this we can compute the hash of the lock file, and in the created layer we can use `AssetHashType.CUSTOM` with this hash.

```ts
bundling: {
          ...props.bundling,
          installDependencies: true,
          // assetExcludes: ["TODO: exclude everything except the dependencies file"]
          assetHashType: AssetHashType.CUSTOM,
          assetHash: Packaging.dependenciesHash(entry),
        },
```

This is my first PR here, and in open source in general, I am working on adding and validating the changes with unit tests, I would appreciate some help with that, if possible.

### Checklist

### TODOs

- assetExcludes for the created layer, exclude all files besides the lock file
- Add tests
- Edit `README.md`

----


orshemtov:
Exemption Request

This is a first PR for me in this codebase, I could use some help with the PR checks, i have the general idea of the design and could use some help with the impl.

Still a draft, supposed to close this issue #23829

PR #30166, opened 2024-05-12
feat(ecs): add validation checks to memory cpu combinations of FARGATE compatible task definitions

### LINKED ISSUE #22216: (ecs): Incorrect ECS Fargate cpu/memory settings are not detected
### Describe the bug

Now this setting succeeds, but this is the wrong cpu/memory ratio for a Fargate setting.
It may be necessary to add a `checkFargateWindowsBasedTasksSize` type of determination method to Fargate as well!

```ts
    const taskDefinitiongraviton2 = new ecs.FargateTaskDefinition(this, 'TaskDefGraviton2', {
      runtimePlatform: {
        operatingSystemFamily: ecs.OperatingSystemFamily.LINUX,
        cpuArchitecture: ecs.CpuArchitecture.ARM64,
      },
      cpu: 256,
      memoryLimitMiB: 122880
    });
```

### Expected Behavior

Before `cdk synth`, CDK will determine the wrong configuration.

### Current Behavior

Generate unsuccessful task definitions as CFn templates

### Reproduction Steps

Here's the sample code I wrote above

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.41.1

### Framework Version

_No response_

### Node.js Version

any

### OS

any

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

mrlikl:
### Issue # (if applicable)

Closes #22216

### Reason for this change

Adds a validation that would point out invalid CPU - Memory limits for a FARGATE compatible task definition. This is based on the document https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html

### Description of changes

Added a `validateFargateTaskDefinitionMemoryCpu` method in the base task definition and run this validation when a task definition has FARGATE in requiresCompatibilities. 

This is in line with the ECS requirement that even if a task definition has EC2 compatibility (along with FARGATE), FARGATE validations apply

Validated the changes by providing incorrect as well as correct values, fixed all the invalid combinations that are present in the unit tests

### Checklist

----


packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
```diff
@@ -899,6 +905,35 @@ export class TaskDefinition extends TaskDefinitionBase {
...
+    const isValidCombination = validCpuMemoryCombinations.some((combo) => {
+      return combo.cpu === Number(cpu) && combo.memory.includes(Number(memory));
+    });
+
+    if (!isValidCombination) {
+      throw new Error('Invalid CPU and memory combinations for FARGATE compatible task definition - https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html');
+    }
+  }
+
+  private range(start: number, end: number, step: number): number[] {
```
lpizzinidev, line 929:
```suggestion
    const validCpuMemoryCombinations = [
      { cpu: '256', memory: ['512', '1024', '2048'] },
      { cpu: '512', memory: this.range(1024, 4096, 1024) },
      { cpu: '1024', memory: this.range(2048, 8192, 1024) },
      { cpu: '2048', memory: this.range(4096, 16384, 1024) },
      { cpu: '4096', memory: this.range(8192, 30720, 1024) },
      { cpu: '8192', memory: this.range(16384, 61440, 4096) },
      { cpu: '16384', memory: this.range(32768, 122880, 8192) },
    ];

    const isValidCombination = validCpuMemoryCombinations.some((combo) => {
      return combo.cpu === cpu && combo.memory.includes(memory);
    });

    if (!isValidCombination) {
      throw new Error('Invalid CPU and memory combinations for FARGATE compatible task definition - https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html');
    }
  }

  private range(start: number, end: number, step: number): string[] {
```
What about avoiding `Number -> String` casting?

packages/aws-cdk-lib/aws-ecs/test/fargate/fargate-task-definition.test.ts
```diff
@@ -21,28 +21,24 @@ describe('fargate task definition', () => {
 
     });
 
-    test('support lazy cpu and memory values', () => {
+    test('does not support lazy cpu and memory values', () => {
```
lpizzinidev, line 24: We should be able to support this by adding validation on the `node` (like [this](https://github.com/aws/aws-cdk/blob/4e9ef15a0cf8b1bc664b5d22b34af9eaaa17bec8/packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts#L524)):
```ts
this.node.addValidation({ validate: () => this. validateFargateTaskDefinitionMemoryCpu(...);
```

lpizzinidev (changes requested):
Thanks 👍 Left some comments for adjustments

packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
```diff
@@ -458,6 +444,33 @@ export class TaskDefinition extends TaskDefinitionBase {
...
+        throw new Error('Cannot use inference accelerators on tasks that run on Fargate');
+      }
+
+      // Validate the CPU and memory size for the Windows operation system family.
+      if (props.runtimePlatform?.operatingSystemFamily?.isWindows()) {
+        this.checkFargateWindowsBasedTasksSize(props.cpu, props.memoryMiB, props.runtimePlatform);
+      }
+
+      // Check the combination as per doc https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html
+      this.validateFargateTaskDefinitionMemoryCpu(props.cpu, props.memoryMiB);
```
lpizzinidev, line 471:
```suggestion
     this.node.addValidation({ 
       validate: () => this.validateFargateTaskDefinitionMemoryCpu(props.cpu, props.memoryMiB)
     });
```
Please note https://github.com/aws/aws-cdk/pull/30166/files#r1641289793

lpizzinidev (changes requested):

packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
mrlikl: I have made all the suggested changes @lpizzinidev

mrlikl: Please take a look once

lpizzinidev (approved):
Thanks 👍

packages/@aws-cdk-testing/framework-integ/test/aws-ecs/test/integ.fargate-task-def.ts
```diff
@@ -0,0 +1,26 @@
...
+import { IntegTest } from '@aws-cdk/integ-tests-alpha';
+
+const app = new cdk.App();
+const stack = new cdk.Stack(app, 'aws-ecs-fargate-task-def');
+
+const taskDefinition = new ecs.TaskDefinition(stack, 'TaskDef', {
+  compatibility: ecs.Compatibility.EC2_AND_FARGATE,
+  cpu: '256',
+  memoryMiB: '512',
+});
```
Leo10Gama, line 12: Since these changes will also affect `ecs.FargateTaskDefinition`, can we include it in an integration test as well? Either in this test or as a separate file should be fine, since we're testing both the regular and Fargate tasks in unit tests already.

Leo10Gama (changes requested):
Thanks for the PR! Left a few comments and nits to be addressed.

packages/aws-cdk-lib/aws-ecs/test/fargate/fargate-task-definition.test.ts
```diff
@@ -26,23 +26,21 @@ describe('fargate task definition', () => {
...
       });
 
-      // THEN
-      Template.fromStack(stack).hasResourceProperties('AWS::ECS::TaskDefinition', {
-        Cpu: '128',
-        Memory: '1024',
-      });
+      expect(() => {
+        Template.fromStack(stack);
+      }).toThrow(/Invalid CPU and memory combinations for FARGATE compatible task definition/);
```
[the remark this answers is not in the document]
mrlikl: The test was reverted to previous version and added a check in `validateFargateTaskDefinitionMemoryCpu` to resolve the cpu and mem values to validate lazy values.

packages/aws-cdk-lib/aws-ecs/test/fargate/fargate-task-definition.test.ts
```diff
@@ -216,6 +214,32 @@ describe('fargate task definition', () => {
...
+        cpu: 256,
+        memoryLimitMiB: 125,
+      });
+
+      expect(() => {
+        Template.fromStack(stack);
+      }).toThrow(/Invalid CPU and memory combinations for FARGATE compatible task definition/);
+    });
+
+    test('succesfull when valid CPU and memory combination is provided', () => {
```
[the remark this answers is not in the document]
mrlikl: Changed this alongside the results for misspelled "successful" that showed up in search

packages/@aws-cdk-testing/framework-integ/test/aws-ecs/test/integ.fargate-task-def.ts
mrlikl: I have added one more integ test for TaskDefinition

packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
```diff
@@ -458,6 +444,35 @@ export class TaskDefinition extends TaskDefinitionBase {
...
+      }
+
+      if (props.inferenceAccelerators && props.inferenceAccelerators.length > 0) {
+        throw new Error('Cannot use inference accelerators on tasks that run on Fargate');
+      }
+
+      // Validate the CPU and memory size for the Windows operation system family.
+      if (props.runtimePlatform?.operatingSystemFamily?.isWindows()) {
+        this.checkFargateWindowsBasedTasksSize(props.cpu, props.memoryMiB, props.runtimePlatform);
+      }
```
Leo10Gama, line 468: Might've missed this before, but this check was moved inside the `(this.isFargateCompatible)` block, when before it was outside of it. Just wondering if this check is only supposed to happen for Fargate-related things and if it's safe to have moved behind this check.

Leo10Gama (commented):
Thanks for the changes, and fixing a few other typos around the repo! Just one minor concern about some of the new logic-shuffling.

packages/aws-cdk-lib/aws-ecs/lib/base/task-definition.ts
mrlikl: If the task def does not have FARGATE capability then task level CPU / mem does not apply for windows containers. 
Screenshot 2024-07-30 at 5 22 30 PM
Plus, 
> For task definitions that specify FARGATE for the requiresCompatibilities parameter (even if EC2 is also specified), you must use one of the values in the following table. These values determines your range of supported values for the CPU and memory parameter.

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html

Hence the if condition would be good

Leo10Gama: Can we move it back outside of the `(this.isFargateCompatible)` check anyway? Adding the additional logic is still technically changing the behaviour of the class, and there's a risk that it might affect something further downstream.

mrlikl: @Leo10Gama Moved it outside the fargate checks

PR #30196, opened 2024-05-14
feat(lambda): allow running a build file

### LINKED ISSUE #18470: (aws-lambda-nodejs): Use esbuild API instead of CLI for bundling
### What is the problem?

If a dependency has a `.node` dependency, for example `ssh2` in `node_modules/ssh2/lib/protocol/crypto.js` has 

`binding = require('./crypto/build/Release/sshcrypto.node');`,

this will have esbuild fail the bundling with 

`error: No loader is configured for ".node" files: ../node_modules/ssh2/lib/protocol/crypto/build/Release/sshcrypto.node`.

In https://github.com/evanw/esbuild/issues/1051 they provide a plugin, but sadly plugins can't be used with binary esbuild.

### Reproduction Steps

Run `npx cdk init app --language=typescript` to create sample app

Add `new NodejsFunction(this, 'testfunc', { entry: 'lib/testfunc/index.ts' });` to test stack

Add to file `lib/testfunc/index.ts`:

```typescript
import { Client } from "ssh2";

const handler = async () => {
  const conn = new Client();

  conn.end();
};

export { handler };
```

Add to file `lib/testfunc/package.json`:

```json
{
  "dependencies": {
    "ssh2": "^1.5.0"
  }
}
```

Run `npm run cdk synth`

It bundling the example code

### What actually happened?

```
$ npm run cdk synth

> cdktest@0.1.0 cdk
> cdk "synth"

Bundling asset CdktestStack/testfunc/Code/Stage...
✘ [ERROR] No loader is configured for ".node" files: lib/testfunc/node_modules/cpu-features/build/Release/cpufeatures.node

    lib/testfunc/node_modules/cpu-features/lib/index.js:1:24:
      1 │ const binding = require('../build/Release/cpufeatures.node');
        ╵                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

✘ [ERROR] No loader is configured for ".node" files: lib/testfunc/node_modules/ssh2/lib/protocol/crypto/build/Release/sshcrypto.node

    lib/testfunc/node_modules/ssh2/lib/protocol/crypto.js:30:20:
      30 │   binding = require('./crypto/build/Release/sshcrypto.node');
         ╵                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

2 errors

/home/azatoth/tmp/cdktest/node_modules/aws-cdk-lib/core/lib/asset-staging.ts:395
      throw new Error(`Failed to bundle asset ${this.node.path}, bundle output is located at ${bundleErrorDir}: ${err}`);
            ^
Error: Failed to bundle asset CdktestStack/testfunc/Code/Stage, bundle output is located at /home/azatoth/tmp/cdktest/cdk.out/bundling-temp-43706a6374f4884c6246dca414315bc291a0eff7496c6eda3da7968ee92f4c47-error: Error: bash -c npx --no-install esbuild --bundle "/home/azatoth/tmp/cdktest/lib/testfunc/index.ts" --target=node14 --platform=node --outfile="/home/azatoth/tmp/cdktest/cdk.out/bundling-temp-43706a6374f4884c6246dca414315bc291a0eff7496c6eda3da7968ee92f4c47/index.js" --external:aws-sdk run in directory /home/azatoth/tmp/cdktest exited with status 1
    at AssetStaging.bundle (/home/azatoth/tmp/cdktest/node_modules/aws-cdk-lib/core/lib/asset-staging.ts:395:13)
    at AssetStaging.stageByBundling (/home/azatoth/tmp/cdktest/node_modules/aws-cdk-lib/core/lib/asset-staging.ts:243:10)
    at stageThisAsset (/home/azatoth/tmp/cdktest/node_modules/aws-cdk-lib/core/lib/asset-staging.ts:134:35)
    at Cache.obtain (/home/azatoth/tmp/cdktest/node_modules/aws-cdk-lib/core/lib/private/cache.ts:24:13)
    at new AssetStaging (/home/azatoth/tmp/cdktest/node_modules/aws-cdk-lib/core/lib/asset-staging.ts:159:44)
    at new Asset (/home/azatoth/tmp/cdktest/node_modules/aws-cdk-lib/aws-s3-assets/lib/asset.ts:72:21)
    at AssetCode.bind (/home/azatoth/tmp/cdktest/node_modules/aws-cdk-lib/aws-lambda/lib/code.ts:180:20)
    at new Function (/home/azatoth/tmp/cdktest/node_modules/aws-cdk-lib/aws-lambda/lib/function.ts:350:29)
    at new NodejsFunction (/home/azatoth/tmp/cdktest/node_modules/aws-cdk-lib/aws-lambda-nodejs/lib/function.ts:50:5)
    at new CdktestStack (/home/azatoth/tmp/cdktest/lib/cdktest-stack.ts:9:5)
```

### CDK CLI Version

2.8.0 (build 8a5eb49)

### Framework Version

_No response_

### Node.js Version

v14.18.3

### OS

Linux

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

bergjaak:
### Issue # (if applicable)

Closes #18470

### Reason for this change

This allows customers to execute an arbitrary build script as part of cdk synth, which will enable customer to use esbuild plugins. The rationale for this decision is given the issue that is linked above.

### Description of changes

1. Expose the code field on the `aws-lambda-nodejs` construct, so that customers can specify code in ways other than bundling, which was the default and abstracted away from customers before this change.
2. Add a new static method on Code, namely `Code.fromCustomCommand`. This method takes in the commands to run an arbitrary script during cdk synthesis that the customer provides. The customer also provides the location of the output from the buildscript. Then this output is supplied to a lambda function.

manual testing (involving inspecting output in the AWS Lambda console and invoking the function), integration tests, and full unit test coverage of new changes.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-lambda-nodejs/README.md
+++ b/packages/aws-cdk-lib/aws-lambda-nodejs/README.md
@@ -427,7 +427,7 @@ await esbuild
 
 then you could use `build.mjs` in a cdk construct as follows:
 
-```ts
+```
 export class ExampleStack extends Stack {
   public constructor(scope: Construct, id: string, props?: StackProps) {
     super(scope, id, props);


mrgrain:
Thanks for getting this started @bergjaak !

I'd rather see this implemented as outlined by @colifran in Option 3 here:
https://github.com/aws/aws-cdk/issues/18470#issuecomment-2027675962

Reason being that attaching this to the Asset will benefit ALL users of Lambda Functions  PLUS ALL other users of Code Assets. If we feel that it's valuable to make this additionally available to on the Lambda Function directly, we can still do that.

What do you think? Any reasons we can't go with what Option 3 proposed?

packages/aws-cdk-lib/aws-lambda-nodejs/lib/function.ts
```diff
@@ -94,27 +109,41 @@ export class NodejsFunction extends lambda.Function {
...
-        architecture,
-        depsLockFilePath,
-        projectRoot,
-      }),
-      handler: handler.indexOf('.') !== -1 ? `${handler}` : `index.${handler}`,
-    });
+        code: props.code,
+        handler: props.handler!,
+      });
+    } else {
```
colifran, line 125: Can you help me understand why we need this new code block here? Why can't we just determine the handler in the same way we are doing it right now? Couldn't we just keep the existing code and do something where if `code` is supplied we use it, otherwise we do `Bundling.bundle`? So something like:
```ts
  // Entry and defaults
  const entry = path.resolve(findEntry(id, props.entry));
  const architecture = props.architecture ?? Architecture.X86_64;
  const depsLockFilePath = findLockFile(props.depsLockFilePath);
  const projectRoot = props.projectRoot ?? path.dirname(depsLockFilePath);
  const handler = props.handler ?? 'handler';

  super(scope, id, {
    ...props,
    runtime,
    code: props.code ?? Bundling.bundle(scope, {
      ...props.bundling ?? {},
      entry,
      runtime,
      architecture,
      depsLockFilePath,
      projectRoot,
    }),
    handler: handler.indexOf('.') !== -1 ? `${handler}` : `index.${handler}`,
  });
}
```

.vscode/launch.json
```diff
@@ -5,19 +5,15 @@
...
-      // Saves some button-pressing latency on attaching
-      "stopOnEntry": false
+      "request": "launch",
+      "args": ["jest", "--runTestsByPath", "/Users/bergjak/workplace/CDK/aws-cdk/packages/aws-cdk-lib/aws-lambda-nodejs/test/function.test.ts"],
+      "name": "debug unit test",
+      "program": "/opt/homebrew/bin/yarn",
+      "cwd": "/Users/bergjak/workplace/CDK/aws-cdk/packages/aws-cdk-lib",
+      "console": "integratedTerminal",
+      "sourceMaps": true,
+      "skipFiles": [ "<node_internals>/**/*" ],
```
colifran, line 16: leftover?

bergjaak: yeah, I'll remove this in my final revision

packages/aws-cdk-lib/aws-lambda-nodejs/lib/function.ts
bergjaak: That's what I originally had, but I think it's a worse customer experience. Because there's nothing forcing the customer to name the file containing the exported handler `index`. It's entirely up to the customer to decide how they want to name that file, if they're using the `Code.fromCustomCommand`. Therefore, I think this is a better customer experience, even if it's a few more lines of code

colifran: I see where you're coming from but I'm not sure I agree. I think this construct has already established that the expected behavior is we will default the handler if it isn't supplied. Consider a customer that is using the nodejs function and they currently don't supply a handler and just use the default. Now, once this feature is released, they decide to use the new `code` property to run a custom build script. However, suddenly they're encountering errors since they don't set a handler. I recognize that it is trivial to just specify a handler, but it is a change from the expectation for the construct as a whole. Additionally, we generally try to avoid properties where we have to do a bunch of validation around if a property is set then another one must also be set. We consider it a bit of an anti-pattern. Our VPC construct is a more extreme example of this, but it shows where this can get to be hard to deal with when we have to start doing a ton of property validation - https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts

bergjaak: basically, to  avoid having an outage in production, if a customer is switching to using this Code.fromCustomCommand, the customer would need to know that their build file must output a file called index.js that exports a function called handler. That's not something that I think should be hidden from the customer. By requiring handler, we don't abstract away a detail that could potentially lead to an outage

bergjaak: I agree that this is more complicated, but the alternative is to push this complexity onto the customer after they've deployed their lambda

packages/aws-cdk-lib/aws-lambda/lib/code.ts
```diff
@@ -54,6 +55,44 @@ export abstract class Code {
...
+   * @param command The command which will be executed to generate the output, for example, [ 'node', 'bundle_code.js' ]
+   * @param assetOptions The same options that are available for `Code.fromAsset` -- but bundling options are not allowed
+   * @param commandOptions Options that are passed to the spawned process, which determine the characteristics of the spawned process.
+   * * See `child_process.SpawnSyncOptions` for possible inputs and defaults (https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options).
+   */
+  public static fromCustomCommand(
+    output: string,
+    command: string[],
+    commandOptions?: {[option: string]: any}, // jsii build fails if the type is SpawnSyncOptions... so best we can do is point user to those options.
+    assetOptions?: s3_assets.AssetOptions,
```
colifran, line 73: I generally prefer to use interfaces where we can so that if there is ever a chance that a new argument is introduced we can just add it to the interface. What do you think about creating a new interface that just extends `AssetOptions`? We could call it `CustomCommandOptions`. Something like this:

```ts
export interface CustomCommandOptions extends AssetOptions {
	readonly commandOptions?: { [options: string]: any },
}
```

Then the function definition could just be:

```ts
public static fromCustomCommand(output: string, command: string[], options: CustomCommandOptions = {})
```

packages/aws-cdk-lib/aws-lambda-nodejs/lib/function.ts
colifran: Happy to change my mind on this, and with the way you've explained it I agree. Doesn't make sense to let the customer fail during deploy time.

packages/aws-cdk-lib/aws-lambda/lib/code.ts
bergjaak: I agree with you, that's an improvement

packages/aws-cdk-lib/aws-lambda-nodejs/lib/function.ts
```diff
@@ -101,27 +116,45 @@ export class NodejsFunction extends lambda.Function {
...
+          + 'The handler should be the name of the exported function to be invoked and the file containing that function.\n'
+          + 'For example, handler should be specified in the form `myFile.myFunction`',
+        );
+      }
+
+      super(scope, id, {
+        ...props,
+        runtime,
+        code: props.code,
+        handler: props.handler!,
```
colifran, line 134: Can we remove the `!`? We're already checking and throwing an error if `props.handler` is undefined so we don't need to use `!` here.

packages/aws-cdk-lib/aws-lambda-nodejs/lib/function.ts
```diff
@@ -90,6 +94,17 @@ export interface NodejsFunctionProps extends lambda.FunctionOptions {
    * @default - the directory containing the `depsLockFilePath`
    */
   readonly projectRoot?: string;
+
+  /**
+   * The code that will be deployed to the Lambda Handler. If included, then properties related to
+   * bundling of the code are ignored. In the constructor of NodeJsFunction, where the Super is called,
+   * you can see which bundling properties are ignored.
```
colifran, line 101: Let's re-phrase this. Maybe just leave it to "If included, then properties related to bundling of the code are ignored". The "In the constructor of the NodeJsFunction, where the Super is called you can see which bundling properties are ignored" isn't helpful since I wouldn't necessarily expect a user to have to go into our source code.

colifran (changes requested):
This looks really good! I have one last request, could we add an entry to the `aws-lambda-nodejs` `README` that shows an example of how to use it with `Code.fromCustomCommand`? We could start with something that tells the user that if they want more granular control over bundling they can utilize `Code.fromCustomCommand` with a custom build script. An example of a build script using esbuild with plugins is this (`build.mjs`):

```ts
import * as path from 'path';
import { fileURLToPath } from 'url';
import esbuild from "esbuild";
import { cache } from "esbuild-plugin-cache";
import time from "esbuild-plugin-time";

const __filename = fileURLToPath(import.meta.url); // get the resolved path to the file
const __dirname = path.dirname(__filename); // get the name of the directory

await esbuild
  .build({
    entryPoints: [path.join(__dirname, 'handler', 'index.ts')],
    outfile: path.join(__dirname, 'build-output', 'index.js'),
    external: ['@aws-sdk/*', 'aws-sdk'],
    format: 'cjs',
    platform: 'node',
    target: 'node18',
    bundle: true,
    minify: true,
    plugins: [time(), cache({ directory: ".cache" })],
  })
  .catch((error) => {
    console.log(error);
    process.exit(1)
  });
```
To then execute this build script in a stack with a `NodejsFunction` they write the following code:
```ts
export class ExampleStack extends Stack {
  public constructor(scope: Construct, id: string, props?: StackProps) {
    super(scope, id, props);

    const pathToBuildFile = path.join(__dirname, 'build.mjs');
    const pathToOutputFile = path.join(__dirname, 'build-output');

    const code = Code.fromCustomCommand(
      pathToOutputFile,
      ['node', pathToBuildFile],
    );

    new NodejsFunction(this, 'NodejsFunctionBuild', {
      code,
      handler: 'index.handler',
    });
  }
}
```

Just something so that the user doesn't need to dig around through our source code to figure out what an example of this could look like.

packages/aws-cdk-lib/aws-lambda-nodejs/lib/function.ts
bergjaak: that makes sense, agreed

packages/aws-cdk-lib/aws-lambda-nodejs/README.md
```diff
@@ -392,3 +392,65 @@ In situtations where this does not work, like Docker-in-Docker setups or when us
   },
 });
 ```
+
+## Running a custom build script as part of cdk synthesis
```
colifran, line 396: Nice README update!

bergjaak: yay

colifran (approved):
Nice job with this!

PR #30197, opened 2024-05-15
feat(synthetics): add `artifactS3Encryption` property to the Canary Construct.

### LINKED ISSUE #30190: (synthetics): support SSE-S3 and SSE-KMS to encrypt canary artifacts
### Describe the feature

CloudWatch Synthetics supports SSE-S3 and SSE-KMS to encrypt artifacts.
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_artifact_encryption.html

But in the current Canary Construct, we cannot choose encryption options.

### Use Case

To use the SSE-S3 or SSE-KMS option.

### Proposed Solution

Add artifactConfig property to the Canary Construct.

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_synthetics.CfnCanary.ArtifactConfigProperty.html

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.141.0

MacOS

mazyu36:
### Issue # (if applicable)

Closes #30190.

### Reason for this change
To select encryption options.

### Description of changes
Add `artifactS3Encryption` property to the Canary Construct.

### Checklist

----


packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -296,6 +344,31 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
       this._connections = new ec2.Connections({});
     }
 
+    if (props.runtime.family !== RuntimeFamily.NODEJS && props.artifactS3Encryption) {
```
mazyu36, line 347: Artifact encryption functionality is available only for canaries that use Synthetics runtime version syn-nodejs-puppeteer-3.3 or later. 
However, versions prior to 3.3 are deprecated and can no longer be configured, which is why I implemented it this way.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -229,6 +230,51 @@ export interface CanaryProps {
    * @default - no rules applied to the generated bucket.
    */
   readonly artifactsBucketLifecycleRules?: Array<s3.LifecycleRule>;
+
+  /**
+   * Canary Artifacts in S3 encryption configuration.
+   * Artifact encryption is only supported for Node.js runtime.
```
lpizzinidev, line 236:
```suggestion
   *
   * Artifact encryption is only supported for canaries that use Synthetics runtime
   * version `syn-nodejs-puppeteer-3.3` or later.
```
It seems more appropriate to keep it consistent with CloudFormation [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-synthetics-canary-artifactconfig.html).

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -296,6 +347,31 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
       this._connections = new ec2.Connections({});
     }
 
+    if (props.runtime.family !== RuntimeFamily.NODEJS && props.artifactS3Encryption) {
+      throw new Error('Artifact encryption is only supported for Node.js runtime.');
```
lpizzinidev, line 351:
```suggestion
      throw new Error(`Artifact encryption is only supported for canaries that use Synthetics runtime version syn-nodejs-puppeteer-3.3 or later, got ${props.runtime.name}.`);
```
For consistency with CFN [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-synthetics-canary-artifactconfig.html).

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
lpizzinidev: Makes sense 👍 Can you please add a short comment about it before the check?

lpizzinidev (changes requested):
Thanks 👍 Left some comments for some initial adjustments

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -229,6 +230,51 @@ export interface CanaryProps {
...
+
+  /**
+   * Canary Artifacts in S3 encryption configuration.
+   * Artifact encryption is only supported for Node.js runtime.
+   *
+   * @default - Artifacts are encrypted at rest using an AWS managed key
+   *
+   * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_artifact_encryption.html
+   */
+  readonly artifactS3Encryption?: ArtifactS3Encryption;
```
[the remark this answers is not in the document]
mazyu36: Thank you. Canary Construct is implemented to flatten other properties as well, so I agree that it would be better to flatten it in this case as you suggested. I will modify it to make it flat.​​​​​​​​​​​​​​​​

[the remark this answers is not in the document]
mazyu36: I've changed it to flat.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
mazyu36: Thanks. I've added a comment.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -296,6 +335,18 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
       this._connections = new ec2.Connections({});
     }
 
+    if (!cdk.Token.isUnresolved(props.artifactS3EncryptionMode) &&
```
mazyu36, line 338: I refactored the creation of artifactConfig into createArtifactConfig.
However, since kmsKey needs to be set as a property, I made it so that checking and setting are done here.

Please provide your opinion if there is a better way to do this.​​​​​​​​​​​​​​​​

mazyu36:
@lpizzinidev 

Thank you for the review. I incorporated the comments and refactored.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -229,6 +230,39 @@ export interface CanaryProps {
...
+   * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_artifact_encryption.html
+   */
+  readonly artifactS3EncryptionMode?: ArtifactsEncryptionMode;
+
+  /**
+   * The KMS key to be used to encrypt the data.
+   *
+   * @default - no kms key if mode = S3_MANAGED. A key will be created if one is not provided and mode = KMS.
+   */
+  readonly kmsKey?: kms.IKey;
```
lpizzinidev, line 250:
```suggestion
   * The KMS key used to encrypt canary artifacts.
   *
   * @default - no kms key if `artifactS3EncryptionMode` is set to `S3_MANAGED`. A key will be created if one is not provided and `artifactS3EncryptionMode` is set to `KMS`.
   */
  readonly artifactS3KmsKey?: kms.IKey;
```
What about a more meaningful variable name?

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
lpizzinidev: I like the approach, just a couple of adjustments:
* Why not move all validations inside the method?
* No need to declare the `public readonly encryptionKey?: kms.IKey;` variable
* Watch out for the typo in the function name
Suggestion:
```ts
  private createArtifactConfig(props: CanaryProps): CfnCanary.ArtifactConfigProperty | undefined {
    if (!props.artifactS3EncryptionMode) {
      return undefined;
    }

    const isNodeRuntime = !cdk.Token.isUnresolved(props.runtime) && props.runtime.family === RuntimeFamily.NODEJS;
    const isArtifactS3EncryptionModeDefined = !cdk.Token.isUnresolved(props.artifactS3EncryptionMode) && props.artifactS3EncryptionMode;
    const isArtifactS3KmsKeyDefined = !cdk.Token.isUnresolved(props.artifactS3KmsKey) && props.artifactS3KmsKey;

    if (isArtifactS3EncryptionModeDefined &&
      props.artifactS3EncryptionMode !== ArtifactsEncryptionMode.KMS &&
      isArtifactS3KmsKeyDefined
    ) {
      throw new Error('A customer-managed KMS key was provided, but the encryption mode is not set to SSE-KMS.');
    }

    // Only check runtime family is nodejs because versions prior to syn-nodejs-puppeteer-3.3 are deprecated and can no longer be configured.
    if (!isNodeRuntime && isArtifactS3EncryptionModeDefined) {
      throw new Error(`Artifact encryption is only supported for canaries that use Synthetics runtime version syn-nodejs-puppeteer-3.3 or later, got ${props.runtime.name}.`);
    }

    let encryptionKey: kms.IKey | undefined;
    if (props.artifactS3EncryptionMode === ArtifactsEncryptionMode.KMS) {
      // Kms Key is set or generated for using `createArtifactConfig`
      encryptionKey = props.artifactS3KmsKey ?? new kms.Key(this, 'Key', { description: `Created by ${this.node.path}` });
    }

    encryptionKey?.grantEncryptDecrypt(this.role);

    return {
      s3Encryption: {
        encryptionMode: props.artifactS3EncryptionMode,
        artifactS3KmsKeyArn: encryptionKey?.keyArn,
      },
    };
  }
```

lpizzinidev (changes requested):
Thanks 👍 Left comments for some final adjustments

mazyu36:
@lpizzinidev 
Thank you for the review. I have addressed the comments.
(For some reason, the PR linter is showing an error...)

lpizzinidev (approved):
Thanks 👍
Not sure about the failing build. It seems like an issue with automation scripts:
```
error /home/runner/work/aws-cdk/aws-cdk/node_modules/@lerna/create/node_modules/nx, /home/runner/work/aws-cdk/aws-cdk/node_modules/@nx/devkit/node_modules/nx, /home/runner/work/aws-cdk/aws-cdk/node_modules/lerna/node_modules/nx: Command failed.
```

lpizzinidev (approved):
Thanks 👍

comcalvi (dismissed):
Minor changes only, looks great!

packages/@aws-cdk-testing/framework-integ/test/aws-synthetics/test/integ.canary-artifact-s3-encryption.ts
```diff
@@ -0,0 +1,70 @@
...
+  cleanup: Cleanup.LAMBDA,
+  artifactS3EncryptionMode: ArtifactsEncryptionMode.KMS,
+  artifactS3KmsKey: encryptKey,
+});
+
+new IntegTest(app, 'IntegCanaryTest', {
+  testCases: [stack],
+});
+
+app.synth();
```
comcalvi, line 70: You don't need `app.synth()` at the end of new-style (eg, using `IntegTest`) integ tests. Old-style tests required it, but new-style ones should not use it.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -564,6 +599,43 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
     };
   }
 
+  private createArtifactConfig(props: CanaryProps): CfnCanary.ArtifactConfigProperty | undefined {
+    if (!props.artifactS3EncryptionMode) {
+      return undefined;
+    }
+
+    const isNodeRuntime = !cdk.Token.isUnresolved(props.runtime) && props.runtime.family === RuntimeFamily.NODEJS;
```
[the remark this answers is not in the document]
mazyu36: Thanks. This check is redundant.
I've removed it.

packages/@aws-cdk-testing/framework-integ/test/aws-synthetics/test/integ.canary-artifact-s3-encryption.ts
mazyu36: Thanks. removed.

mazyu36:
@paulhcsun 
Thank you for the review.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -633,6 +668,43 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
     };
   }
 
+  private createArtifactConfig(props: CanaryProps): CfnCanary.ArtifactConfigProperty | undefined {
+    if (!props.artifactS3EncryptionMode) {
+      return undefined;
+    }
```
Leo10Gama, line 674: What would be the behaviour if the user provides `props.artifactS3KmsKey` but not `props.artifactS3EncryptionMode`?

Leo10Gama (commented):
Thanks for the contribution! Overall looks good, just one minor comment about the combination of the new props.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
mazyu36: Thanks.
I tried this case and Canary uses SSE-S3 (CMK is ignored).

But, I think this case should throw an error, so I've added a validation.

Leo10Gama: I don't think throwing an error is the best route here, since if the user provides a KMS key, it's likely they'll want to use SSE-KMS, and it would be more intuitive for us to set it for them (see: [Avoid Errors if Possible](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#avoid-errors-if-possible)).

mazyu36: Thanks.
In that case, I have modified it to automatically use SSE KMS.​​​​​​​​​​​​​​​​

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -269,7 +269,7 @@ export interface CanaryProps {
    * Artifact encryption is only supported for canaries that use Synthetics runtime
    * version `syn-nodejs-puppeteer-3.3` or later.
    *
-   * @default - Artifacts are encrypted at rest using an AWS managed key
+   * @default - `ArtifactsEncryptionMode.KMS` is set if you specify `artifactS3KmsKey`, otherwise artifacts are encrypted at rest using an AWS managed key
```
Leo10Gama, line 272: nit: I think stating the normal default "Artifacts encrypted at rest with AWS-managed key" first might be better phrasing, mentioning the `artifactS3KmsKey` afterwards, just to avoid confusion.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -669,37 +669,44 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
     const isNodeRuntime = props.runtime.family === RuntimeFamily.NODEJS;
-    const isArtifactS3EncryptionModeDefined = !cdk.Token.isUnresolved(props.artifactS3EncryptionMode) && props.artifactS3EncryptionMode;
-    const isArtifactS3KmsKeyDefined = !cdk.Token.isUnresolved(props.artifactS3KmsKey) && props.artifactS3KmsKey;
 
     if (
-      isArtifactS3EncryptionModeDefined &&
-      props.artifactS3EncryptionMode !== ArtifactsEncryptionMode.KMS &&
-      isArtifactS3KmsKeyDefined
+      props.artifactS3EncryptionMode === ArtifactsEncryptionMode.S3_MANAGED &&
+      props.artifactS3KmsKey
```
Leo10Gama, line 680: I can see we're no longer checking if these props are tokens. Is it possible for users to provide them as tokens? I don't think it would change much of the logic here, but if they can be tokens, we ought to have unit tests for them.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -669,37 +669,44 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
     }
 
     let encryptionKey: kms.IKey | undefined;
-    if (props.artifactS3EncryptionMode === ArtifactsEncryptionMode.KMS) {
-      encryptionKey = props.artifactS3KmsKey ?? new kms.Key(this, 'Key', { description: `Created by ${this.node.path}` });
+    if (props.artifactS3EncryptionMode === ArtifactsEncryptionMode.KMS && !props.artifactS3KmsKey) {
+      encryptionKey = new kms.Key(this, 'Key', { description: `Created by ${this.node.path}` });
+    } else {
+      encryptionKey = props.artifactS3KmsKey;
     }
```
Leo10Gama, line 695: I'm curious about why we made this change. The previous lines seemed more intuitive in my opinion, whereas the second isn't quite as clear (e.g. `encryptionKey` will still be set if encryption mode is S3, since it'll fail the first part of the condition). If we need the default behaviour to be that the encryption mode is set to KMS when a key is provided, maybe we could instead set the encryption mode before this block? (i.e. move lines L699-L706 before this)

Leo10Gama (dismissed):
Thanks for addressing the changes! Just a few more comments about some of the recent changes.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
mazyu36: Thank you. I reviewed it to set the `encryptionMode` first, followed by setting the `encryptionKey`.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
mazyu36: As I understand it, these are not expected to become tokens. Therefore, the token check is unnecessary, so I removed it.

Leo10Gama (dismissed):
LGTM, thanks for the contribution!

Leo10Gama (approved):

Issue #23005, opened 2022-11-21
(lambda): Add `domain` helper prop


diesal11:
### Describe the feature

The lambda.FunctionUrl construct currently returns the Function URL as a full URL including the scheme + trailing `/`.

`https://*******.lambda-url.ap-southeast-2.on.aws/`

In many cases, it's preferable to have just the domain name.

`*******.lambda-url.ap-southeast-2.on.aws`

### Use Case

When serving a Function URL through a Cloudfront Distribution. The following workaround is required to remove the URL Scheme and path from `fn_url.url` so it can be consumed by `Cloudfront.HttpOrigin`

```python
# In future it would be nice to have some level of auth between CF & lambda
fn_url = function.add_function_url(
    auth_type=lmbda.FunctionUrlAuthType.NONE,
)

# This removes the URL Scheme and trailing `/` from the Function URL`
# leaving: "*******.lambda-url.ap-southeast-2.on.aws"
fn_url_domain = Fn.select(
    0, Fn.split("/", Fn.select(1, Fn.split("https://", fn_url.url)))
)

distribution = cf.Distribution(
    self,
    "distribution",
    default_behavior=cf.BehaviorOptions(
        origin=cfo.HttpOrigin(fn_url_domain),
    )
)
```

### Proposed Solution

The above code be added as a helper function on FunctionUrl construct:

```
fn_url_domain = Fn.select(
    0, Fn.split("/", Fn.select(1, Fn.split("https://", fn_url.url)))
)
```

### Other Information

Happy to implement this one if it's a wanted change :) 

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.50.0

### Environment details (OS name and version, etc.)

MacOS 13

peterwoodworth:
The return value here comes from [CloudFormation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-url.html) - https://github.com/aws/aws-cdk/blob/3d30cdb55620a5f2a34944d21edca770e90ff7c5/packages/%40aws-cdk/aws-lambda/lib/function-url.ts#L206

We could add the domain prop to `FunctionUrl` using the same workaround you're using with intrinsic functions, thanks for the feature request 🙂

diesal11:
I've opened #23046 which adds this + a unit test.

PR #30198, opened 2024-05-15
feat(lambda): add domainName prop to FunctionUrl

mbfreder:
### Issue # (if applicable)

Closes <https://github.com/aws/aws-cdk/issues/23005>.

### Reason for this change
This change is to address this [issue](https://github.com/aws/aws-cdk/issues/23005)

### Description of changes
This PR adds a new property called domainName to the FunctionUrl construct. The goal is to make it easy for customers to get the domain name directly (`FunctionUrl.domainName`) instead of first getting the url (`FunctionUrl.url`) and then removing the scheme and the trailing slash. 
The README has been updated.

I Added some unit tests. I Don't think integration tests are necessary in this case.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-lambda/README.md
+++ b/packages/aws-cdk-lib/aws-lambda/README.md
@@ -480,6 +480,15 @@ new CfnOutput(this, 'TheUrl', {
   value: fnUrl.url,
 });
 ```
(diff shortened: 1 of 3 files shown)


mbfreder:
I don't think integration tests are needed in this case.

PR #30237, opened 2024-05-16
chore(stepfunction-tasks): eventbridge aws. event source prefix check is more strict than it should be

### LINKED ISSUE #30191: aws-stepfunctions-tasks: EventBridge `aws.` event source prefix check is more strict than it should be
### Describe the bug

We discovered that the [`EventBridgePutEvents`](
https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/eventbridge/put-events.ts#L136-L137) is throwing an error even in cases when it's not necessary to throw. 

For example our `source` looks like this `aws-some-feature.this.that` and according to the [docs](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events-structure.html) this should be valid.

> All events that come from AWS services begin with "aws." 

Whereas in the CDK code of [`EventBridgePutEvents`](
https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/eventbridge/put-events.ts#L136-L137) is this check

```ts
if (entry.source?.startsWith('aws')) {
```

The correct implementation should be following

```ts
if (entry.source?.startsWith('aws.')) {
```

### Expected Behavior

`EventBridgePutEvents` should not throw an error when `source` starts with `aws.` _(but not just `aws`)_

### Current Behavior

`EventBridgePutEvents` throws an error when `source` starts with `aws`

### Reproduction Steps

Use this task in a state machine.

```ts
import { EventBridgePutEvents } from 'aws-cdk-lib/aws-stepfunctions-tasks'

new EventBridgePutEvents(
  this,
  'SendEvent',
  {
    entries: [
      {
        eventBus: 'some-event-bus',
        source: `aws-feature-name-whatever-you-want`,
        detailType: 'some-detail',
        detail: TaskInput.fromJsonPathAt('$')
      }
    ]
  }
)
```

### Possible Solution

Change this line: https://github.com/aws/aws-cdk/blob/205163fc0d2cac84d3d746a98c393e137f0e2388/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/eventbridge/put-events.ts#L136C7-L136C45

to:

```diff
- if (entry.source?.startsWith('aws')) {
+ if (entry.source?.startsWith('aws.')) { 
```

_No response_

### CDK CLI Version

v2.141.0

### Framework Version

_No response_

### Node.js Version

20.x

### OS

MacOS (personal) / Ubuntu (in CI)

### Language

TypeScript

### Language Version

5.4.5

### Other information

_No response_

pahud:
1. fix the event source validation
2. move the validation from renderEntries() to validateEntries() 
3. add unit tests

### Issue # (if applicable)

Closes #30191

### Checklist

----


pahud:
Exemption Request

pahud:
Exemption Request

pahud:
Exemption Request

paulhcsun (approved):

paulhcsun:
@mergifyio update

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/eventbridge/put-events.ts
```diff
@@ -132,23 +132,23 @@ export class EventBridgePutEvents extends sfn.TaskStateBase {
   }
 
   private renderEntries(): Object[] {
+    // we should have validated all entries in validateEntries()
```
paulhcsun, line 135: Is it guaranteed that `validateEntries()` will always be called before `renderEntires()`? I just wonder if there's any cases where `renderEntries()` is called on it's own and removing the check from this method would mean the check isn't run at all.

PR #30240, opened 2024-05-16
fix: ignored platform property on bundling an asset with docker

### LINKED ISSUE #30239: asset-staging.ts: target platform ignored when bundling asset
### Describe the bug

When bundling an asset e.g. in an s3_deployment using docker, the platform property gets ignored. The bundling always takes place on the host's platform. 

### Expected Behavior

Using 
`platform: 'linux/amd64' `
should bundle for amd64 on M1 and 
`platform: 'linux/arm64' `
should bundle for ARM on amd64 architecture as described in https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.BundlingOptions.html#platform

### Current Behavior

`platform: 'something'` is silently ignored.

### Reproduction Steps

Bundle for a s3_deployment like this:
```
new s3_deployment.BucketDeployment(this, 'platform-bug', {
      sources: [s3_deployment.Source.asset('.', {
        bundling: {
          image: lambda.Runtime.PYTHON_3_9.bundlingImage,
          platform: 'linux/amd64',
          //platform: 'linux/arm64',
          command: [
            'bash', '-c',
            'uname -m > /asset-output/uname-m.out.txt'
          ],
        },
      })],
      destinationBucket: bucket,
      destinationKeyPrefix: 'test',
    })
```
Do a `cdk synthesize` and look into the created output file 'uname-m.out.txt' asset in cdk.out. Whatever you put into 'platform', it always contains your host machine's platform string.

### Possible Solution

The reason is that the platform given does not get propagated in aws-cdk-lib/core/lib/private/asset-staging.ts when calling image.run().
adding `platform: this.options.platform` at two places fixes the issue.
Pull request is on it's way.

_No response_

### CDK CLI Version

2.141.0 (build 3d1c06e)

### Framework Version

_No response_

### Node.js Version

v20.11.0

### OS

macos 14.4.1 (23E224)

### Language

TypeScript

### Language Version

_No response_

### Other information

As a sidenote: before changing the platform, the local docker build image, e.g. `public.ecr.aws/sam/build-python3.9:latest` has to be deleted manually using `docker image rm public.ecr.aws/sam/build-python3.9:latest`. Otherwise docker would use the same image for a different/wrong platform with unwanted behavior.
Fixing that is not within the scope of this issue.

t3r:
This fix makes the 'platform' property work as described in https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.BundlingOptions.html#platform

The bundling-properties value was not copied from the AssetBundling properties to the image.run properties so the --platform option for docker run was never added.  This patch copies the platform property for AssetBundlingBindMount and AssetBundlingVolumeCopy.

fixes #30239

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/core/lib/private/asset-staging.ts
+++ b/packages/aws-cdk-lib/core/lib/private/asset-staging.ts
@@ -52,6 +52,7 @@ export class AssetBundlingBindMount extends AssetBundlingBase {
   public run() {
     this.options.image.run({
       command: this.options.command,
+      platform: this.options.platform,
       user: this.determineUser(),
       environment: this.options.environment,
       entrypoint: this.options.entrypoint,
@@ -178,6 +179,7 @@ export class AssetBundlingVolumeCopy extends AssetBundlingBase {
 
     this.options.image.run({
       command: this.options.command,
+      platform: this.options.platform,
       user: user,
       environment: this.options.environment,
       entrypoint: this.options.entrypoint,
--- a/packages/aws-cdk-lib/core/test/private/asset-staging.test.ts
+++ b/packages/aws-cdk-lib/core/test/private/asset-staging.test.ts
@@ -113,4 +113,33 @@ describe('bundling', () => {
       'public.ecr.aws/docker/library/alpine',
     ]), { encoding: 'utf-8', stdio: ['ignore', process.stderr, 'inherit'] })).toEqual(true);
   });
+
+  test('AssetBundlingBindMount bundles with given platform', () => {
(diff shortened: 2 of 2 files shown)


t3r:
The PR Linter complains about missing changes to test/integration test files.  There is currently no existing test in `staging.test.ts` for building on different platforms. Adding it might lead to failed tests on machines without working docker buildx. As this is a trivial fix, I'd suggest to not add a test for platform and ask for a Exemption Request for this PR.

PR #30245, opened 2024-05-16
chore(cli): skipping old cli integ test `cdk diff changes in change set` for new changes

SankyRed:
Fixes: [30093](https://github.com/aws/aws-cdk/pull/30093)

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/resources/cli-regression-patches/v2.142.0/skip-tests.txt
+++ b/packages/@aws-cdk-testing/cli-integ/resources/cli-regression-patches/v2.142.0/skip-tests.txt
@@ -1,4 +1,4 @@
 # Skipping the test to fix issue https://github.com/aws/aws-cdk/issues/30241. 
-# cli-integ tests failing for the old tests with the new cli changes for list stacks.
+# cli-integ tests failing for the old tests with the new cli changes for diff stacks.
 
 cdk diff picks up changes that are only present in changeset


packages/@aws-cdk-testing/cli-integ/resources/cli-regression-patches/v2.142.0/skip-tests.txt
```diff
@@ -0,0 +1,4 @@
+# Skipping the test to fix issue https://github.com/aws/aws-cdk/issues/30241. 
+# cli-integ tests failing for the old tests with the new cli changes for list stacks.
```
bergjaak, line 2: `cli-integ tests failing for the old tests with the new cli changes for list stacks.` Should that line be removed?

bergjaak: I think just the line above is sufficient and true, whereas that line isn't true?

bergjaak (approved):

packages/@aws-cdk-testing/cli-integ/resources/cli-regression-patches/v2.142.0/skip-tests.txt
SankyRed: No, I've cleaned it up but, these are just comments I've added. The test we'd want to skip is `cdk diff picks up changes that are only present in changeset`

SankyRed:
> 
> ```
> ```
> 
> 

The fix is to skip old tests being tested against the new cli changes for diff stacks and fix the issue over the pipeline.

colifran:
Can you improve the title so we know what cli-integ test is being added? That way if we search our commits we will be able to easily identify which PR added the integ test to the regression suite.

SankyRed:
> Can you improve the title so we know what cli-integ test is being added? That way if we search our commits we will be able to easily identify which PR added the integ test to the regression suite.

This change is related skipping the old integ test but, updated the title with the test.

Issue #29600, opened 2024-03-25
aws-cdk-lib.aws_dynamodb: Resource Policy property


clouddev-code:
### Describe the feature

DynamoDB Resource Policy was supported recently.
https://aws.amazon.com/about-aws/whats-new/2024/03/amazon-dynamodb-resource-based-policies/

L2 Constructs donot have  blocker   for ResourcePolicy is supported in CloudFormation.
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-table-resourcepolicy.html

### Use Case

DynamoDB ResourcePolicy set by L2 Constructs. 

### Acknowledgements

- [X] I may be able to implement this feature request
- [X] This feature might incur a breaking change

### CDK version used

2.133.0

### Environment details (OS name and version, etc.)

M1 Mac, node18

tim-finnigan:
Thanks for the feature request!

LeeroyHannigan:
You can assign this one to me.

PR #30251, opened 2024-05-17
feat(dynamodb): add resource polices for table

LeeroyHannigan:
Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29600.

https://github.com/aws/aws-cdk/issues/29600
Reason for this change

Adding a new feature
Description of changes

Add resourcePolicy for DynamoDB Table component in aws-dynamodb
Description of how you validated changes

integration test integ.dynamodb.policy.ts
Checklist

    [X ] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----


scanlonp:
@Mergifyio update

packages/aws-cdk-lib/aws-dynamodb/test/table-v2.test.ts
```diff
@@ -2805,3 +2806,81 @@ describe('imports', () => {
...
+            ],
+            Version: '2012-10-17',
+          },
+        },
+      },
+    ],
+  });
+});
+
+test('throws if trying to add a resource policy to a region other than local region', () => {
```
scanlonp, line 2857: This is an odd test. It is not actually testing what is says its testing. There is no error added related to policy statements, so we cannot throw in a unit test for that scenario. 

This error is already covered by a test in this section, so I suggest we delete this test unless it is testing something related to policy documents.

scanlonp (dismissed):
@LeeroyHannigan, this looks amazing! Just a bit of clean up and some implementation questions. I said in the comments, but I think we can take all of the doc strings and code style from table-v2.ts and copy it to the corresponding places in table.ts.

I see that we are slightly changing the grant implementation; I assume it should not affect its current behavior. I wish we had some grant coverage in our integ tests, but it looks like we do not currently test it beyond a couple `grantReadData/grantWriteData` calls. However, I see it is thoroughly unit tested, and none were effected, so I feel good.

Lastly, you call out that adding a replica and adding a resource-based policy to that replica in the same deployment (update) is not possible. Is this an error that is thrown on the CloudFormation side? Can you detail how this affects a user creating a TableV2? It sounds to me like a user needs to define all their replicas, deploy, then add all the resource policies, the deploy again. Is that correct?
I do not believe we can validate this at synth time since we are stateless. Calling it out may be the only option.

Again, thanks for submitting this PR and excited for customers being able to use this soon!

scanlonp:
@Mergifyio update

LeeroyHannigan:
> Lastly, you call out that adding a replica and adding a resource-based policy to that replica in the same deployment (update) is not possible. Is this an error that is thrown on the CloudFormation side? Can you detail how this affects a user creating a TableV2? It sounds to me like a user needs to define all their replicas, deploy, then add all the resource policies, the deploy again. Is that correct? I do not believe we can validate this at synth time since we are stateless. Calling it out may be the only option.
> 
Yes, precisely. You first would need to create the replica and then update with the resource based policy. This limitation comes from the CFN implementation.

packages/aws-cdk-lib/aws-dynamodb/README.md
```diff
@@ -644,6 +644,35 @@ If you intend to use the `tableStreamArn` (including indirectly, for example by
...
+      partitionKey: {
+        name: 'id',
+        type: dynamodb.AttributeType.STRING,
+      },
+      removalPolicy: RemovalPolicy.DESTROY,
+      resourcePolicy: policy,
+    });
+```
+
+TableV2 doesn’t support creating a replica and adding a resource-based policy to that replica in the same stack update in Regions other than the Region where you deploy the stack update. To incorporate a resource-based policy into a replica, you'll need to initially deploy the replica without the policy, followed by a subsequent update to include the desired policy.
```
scanlonp, line 674:
```suggestion
TableV2 doesn’t support creating a replica and adding a resource-based policy to that replica in the same stack update in Regions other than the Region where you deploy the stack update.
To incorporate a resource-based policy into a replica, you'll need to initially deploy the replica without the policy, followed by a subsequent update to include the desired policy.
```

scanlonp (dismissed):
Looks good. Thanks for your hard work on this @LeeroyHannigan

scanlonp (approved):

PR #30255, opened 2024-05-17
feat(cognito): support SecretValue for UserPoolIdentityProviderOidc

### LINKED ISSUE #29851: aws_cognito: UserPoolIdentityProviderOidc doesn't support SecretValue for client secret
### Describe the feature

Hello,

Small FR that would make the use of SecretValue more straightforward.

`UserPoolIdentityProviderOidc` needs a client secret value for deployment. As of now it only supports `string` value. 
Compared to the generic one like `UserPoolIdentityProviderGoogle` who support `clientSecretValue`.

### Use Case

It would make the DX a bit better, simplifying the resolution. And also align with the other providers from the cognito package.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.131.0

MacOS

jwoehrle:
### Issue # (if applicable)

Closes #29851

### Reason for this change

Support SecretValue for UserPoolIdentityProviderOidc.
### Description of changes

Added optional support for SecretValue to UserPoolIdentityProviderOidc.

Yes, see packages/aws-cdk-lib/aws-cognito/test/user-pool-idps/oidc.test.ts#L229

### Checklist

----


paulhcsun (changes requested):
Hi @jwoehrle, thanks for making this contribution! 

This change definitely brings this contract more in line with what we prefer in terms of accepting objects instead of strings, however it also makes the contract more complicated because we would need to maintain both of these paths until cdk v3 and having both properties as optional becomes a confusing DX. 

This is not an incorrect change by any means and we would definitely want to update this property when cdk v3 eventually happens, but after discussing with the team I don't think we are willing to accept this improvement due due to the tradeoffs that come with it at this time. 

I can't think of any other ways to make this change without making both properties optional other than using a feature flag but we try to avoid adding new feature flags as much as possible as it makes our code messier and harder to maintain. I'm open to any other alternative suggestions if you have any, but otherwise we will close this PR and I'll update the original issue.

clayrosenthal:
@paulhcsun 
If this is messy here, why isn't it messy in the google provider? It's the same logic in both. Would refactoring the logic into a helper function so its DRY make this an acceptable solution?

PR #30268, opened 2024-05-19
fix(diff): properties from ChangeSet diff were ignored

### LINKED ISSUE #29731: (cli): changeset-based `cdk diff` not showing changes caused by SSM parameters, but appear in CloudFormation UI changeset info

bergjaak:
### Issue # (if applicable)

Closes aws/aws-cdk-cli#641

### Reason for this change

This is a second attempt at https://github.com/aws/aws-cdk/pull/30093/files, after that PR had to be reverted due to causing issues. (That pr made it possible for ResourceTypes to be undefined, which caused `includes` to be invoked on undefined.)

### Description of changes

* Since the changeset logic is significant enough, I moved out the existing changeset functions to its own file, to allow for better separation of concerns. Further, I wrote each function to be concerned with a single idea and tested each function directly, with full coverage.
* The resourceType getter that caused `includes` to be called on `undefined` in the previous PR attempting this change -- I changed the resourceType getter to state that it can return undefined, so that callers easily know to handle undefined.
* I add `include-property-values` to our describeChangeSet API call so that the changeset description includes the before and after values of the changed properties, which is a pretty new CloudFormation feature https://aws.amazon.com/about-aws/whats-new/2024/04/aws-cloudformation-changesets-enhanced-change-visibility-deployments/
* Also, changes from the changeset will now show up in the security diff (e.g., `cdk diff --security-only`).

* full unit test coverage on the individual functions of the new class, an integration test, and manually testing

### Checklist

----


packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
```diff
@@ -53,10 +54,16 @@ export function fullDiff(
 
   normalize(currentTemplate);
   normalize(newTemplate);
-  const theDiff = diffTemplate(currentTemplate, newTemplate);
+  let theDiff = diffTemplate(currentTemplate, newTemplate);
   if (changeSet) {
-    filterFalsePositives(theDiff, changeSet);
-    addImportInformation(theDiff, changeSet);
+    // These methods mutate the state of theDiff, using the changeSet.
+    const changeSetDiff = new TemplateAndChangeSetDiffMerger({ changeSet: changeSet });
```
colifran, line 60: Nit: You can consolidate key-value pairs with the same name down to a single value. For example, this could just be written as:
```ts
const changeSetDiff = new TemplateAndChangeSetDiffMerger({ changeSet });
```

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+  findResourceImports(): string[] {
+    const importedResourceLogicalIds = [];
+    for (const resourceChange of this.changeSet?.Changes ?? []) {
+      if (resourceChange.ResourceChange?.Action === 'Import') {
+        importedResourceLogicalIds.push(resourceChange.ResourceChange.LogicalResourceId!);
+      }
+    }
+
+    return importedResourceLogicalIds;
+  }
```
bergjaak, line 210: these are just copy and pasted from diff-template.ts

packages/@aws-cdk/cloudformation-diff/lib/diff-template.ts
```diff
@@ -214,100 +214,12 @@ function deepCopy(x: any): any {
...
-      }
-    }
-    replacements[resourceChange.ResourceChange?.LogicalResourceId!] = {
-      resourceReplaced: resourceChange.ResourceChange?.Replacement === 'True',
-      propertiesReplaced,
-    };
-  }
-
-  return replacements;
-}
```
bergjaak, line 309: this was refactored into multiple functions in the new class

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+ * The purpose of this class is to include differences from the ChangeSet to differences in the TemplateDiff.
+ */
+export class TemplateAndChangeSetDiffMerger {
+  // If we somehow cannot find the resourceType, then we'll mark it as UNKNOWN, so that can be seen in the diff.
+  static UNKNOWN_RESOURCE_TYPE = 'UNKNOWN';
+
+  /**
+   * @param parseOldOrNewValues These enum values come from DescribeChangeSetOutput, which indicate if the property resolves to that value after the change or before the change.
+   */
+  static convertResourceFromChangesetToResourceForDiff(
```
colifran, line 20: Seems like this is only ever used by other class methods so `private static convertResourceFromChangesetToResourceForDiff`?

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+
+export type DescribeChangeSetOutput = DescribeChangeSet;
+export type ChangeSetResourceChangeDetail = RCD;
+
+/**
+ * The purpose of this class is to include differences from the ChangeSet to differences in the TemplateDiff.
+ */
+export class TemplateAndChangeSetDiffMerger {
+  // If we somehow cannot find the resourceType, then we'll mark it as UNKNOWN, so that can be seen in the diff.
+  static UNKNOWN_RESOURCE_TYPE = 'UNKNOWN';
```
colifran, line 15: `private static UNKNOWN_RESOURCE_TYPE`. Also, just curious, what will an unknown resourceType look like in the diff?

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+
+        // This property diff will be hydrated when hydrateChangeImpactFromChangeSet is called.
+        const emptyPropertyDiff = new types.PropertyDifference({}, {}, {});
+        emptyPropertyDiff.isDifferent = true;
+        resourceDiffs.get(logicalId).setPropertyChange(propertyName, emptyPropertyDiff);
+      }
+    }
+  }
+
+  hydrateChangeImpactFromChangeSet(logicalId: string, change: types.ResourceDifference) {
```
colifran: `public hydrateChangeImpactFromChangeSet`

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+import { diffResource } from '.';
+import * as types from '../diff/types';
+
+export type DescribeChangeSetOutput = DescribeChangeSet;
+export type ChangeSetResourceChangeDetail = RCD;
+
+/**
+ * The purpose of this class is to include differences from the ChangeSet to differences in the TemplateDiff.
+ */
+export class TemplateAndChangeSetDiffMerger {
```
colifran, line 13: I love this. This is a great refactor!

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+          case 'Metadata':
+            // we want to ignore metadata changes in the diff, so compare newValue against newValue.
+            change.setOtherChange('Metadata', new types.Difference<string>(value.newValue, value.newValue));
+            break;
+        }
+      }
+    });
+  }
+
+  addImportInformation(resourceDiffs: types.DifferenceCollection<types.Resource, types.ResourceDifference>) {
```
colifran: `public addImportInformation`

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
bergjaak:
```
[~] UNKNOWN nice187 nice1878D68AE26 replace
 └─ [~] QueueName (requires replacement)
     ├─ [-] nice187ANDnewAndDifferent
     └─ [+] nice187ANDnewAndDifferentsdgf
[~] UNKNOWN nice188 nice188D555BE27 replace
 └─ [~] QueueName (requires replacement)
     ├─ [-] nice188ANDnewAndDifferent
     └─ [+] nice188ANDnewAndDifferentsdgf
```

To produce this, I actually just changed the code to always set UNKNOWN. However, this should only happen if something has gone wrong... but rather than crash the program, I think this is better. Maybe we could change it to `UNKNOWN_RESOURCE_TYPE`, though

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+    const imports = this.findResourceImports();
+    resourceDiffs.forEachDifference((logicalId: string, change: types.ResourceDifference) => {
+      if (imports.includes(logicalId)) {
+        change.isImport = true;
+      }
+    });
+  }
+
+  findResourceImports(): string[] {
```
colifran: `private findResourceImports`

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
bergjaak: I will add the public and private modifiers. However, some of these functions, while only used in this class for application code, are directly unit tested. So I will set those methods as public so that they can be unit tested

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+export class TemplateAndChangeSetDiffMerger {
+  // If we somehow cannot find the resourceType, then we'll mark it as UNKNOWN, so that can be seen in the diff.
+  static UNKNOWN_RESOURCE_TYPE = 'UNKNOWN';
+
+  /**
+   * @param parseOldOrNewValues These enum values come from DescribeChangeSetOutput, which indicate if the property resolves to that value after the change or before the change.
+   */
+  static convertResourceFromChangesetToResourceForDiff(
+    resourceInfoFromChangeset: types.ChangeSetResource,
+    parseOldOrNewValues: 'BEFORE_VALUES' | 'AFTER_VALUES',
```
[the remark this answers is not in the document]
bergjaak: sounds good :)

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+    },
+  ) {
+    this.changeSet = args.changeSet;
+    this.changeSetResources = args.changeSetResources ?? this.createChangeSetResources(this.changeSet);
+  }
+
+  /**
+   * use information from the changeset to populate details that are missing in the templateDiff
+   */
+  createChangeSetResources(changeSet: DescribeChangeSetOutput): types.ChangeSetResources {
```
colifran: `private createChangeSetResources`

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+      }
+    }
+
+    return {
+      Type: resourceInfoFromChangeset.resourceType ?? TemplateAndChangeSetDiffMerger.UNKNOWN_RESOURCE_TYPE,
+      Properties: props,
+    };
+  }
+
+  static determineChangeSetReplacementMode(propertyChange: ChangeSetResourceChangeDetail): types.ChangeSetReplacementMode {
```
colifran: `private static determineChangeSetReplacementMode`

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
```diff
@@ -0,0 +1,211 @@
...
+          propertiesReplaced[propertyChange.Target.Name] = {
+            changeSetReplacementMode: TemplateAndChangeSetDiffMerger.determineChangeSetReplacementMode(propertyChange),
+            beforeValue: _maybeJsonParse(propertyChange.Target.BeforeValue),
+            afterValue: _maybeJsonParse(propertyChange.Target.AfterValue),
+          };
+        }
+      }
+
+      changeSetResources[resourceChange.ResourceChange.LogicalResourceId] = {
+        resourceWasReplaced: resourceChange.ResourceChange.Replacement === 'True',
```
colifran, line 101: Are we sure `ResourceChange` will always exist? I'm seeing that the old code always had `resourceChange.ResourceChange?.<property>`.

colifran: I guess line 80 would be our check for this?

bergjaak: yeah, I'm trying my best to be super paranoid in this PR about running into anything that's undefined... part of me wants to wrap the entire changeset operation in try-catch and just continue with template diff if the changeset operations fail... but that would probably not be good in that it would mask many issues.

packages/@aws-cdk/cloudformation-diff/lib/diff/types.ts
```diff
@@ -367,6 +380,10 @@ export class DifferenceCollection<V, T extends IDifference<V>> {
     delete this.diffs[logicalId];
   }
 
+  public set(logicalId: string, diff: T): void {
+    this.diffs[logicalId] = diff;
```
colifran, line 384: Do you think it would be an error if `diffs` already had a `diff` set for a specific `logicalId`? Asking because it could help us catch our own errors easier. So something like: `if (this.diffs.hasOwnProperty(logicalId) => throw an error`. Not sure if this makes sense or not. What do you think?

packages/@aws-cdk/cloudformation-diff/lib/diff/template-and-changeset-diff-merger.ts
bergjaak: Here's what it looks like now: 

```
[~] UNKNOWN_RESOURCE_TYPE nice8 nice828888BA3 replace
 └─ [~] QueueName (requires replacement)
     ├─ [-] nice8ANDnewAndDifferentsdgf
     └─ [+] nice8ANDdiff
```

packages/aws-cdk/lib/api/util/cloudformation.ts
```diff
@@ -199,12 +199,35 @@ async function describeChangeSet(
...
+    debug('Failed to describeChangeSet with IncludePropertyValues. Diff will not include property values from ChangeSet. Error Message: %s', e?.message);
+  }
+
   // If fetchAll is true, traverse all pages from the change set description.
   while (fetchAll && response.NextToken != null) {
     const nextPage = await cfn.describeChangeSet({
       StackName: stackName,
       ChangeSetName: response.ChangeSetId ?? changeSetName,
       NextToken: response.NextToken,
+      IncludePropertyValues: true,
```
bergjaak, line 230: I need to handle this for regions where this parameter isn't supported

packages/aws-cdk/lib/api/util/cloudformation.ts
```diff
@@ -191,21 +191,31 @@ export class CloudFormationStack {
  *
  * @returns       CloudFormation information about the ChangeSet
  */
-async function describeChangeSet(
+export async function describeChangeSet(
```
bergjaak, line 194: this is where we get the changeset for the diff

PR #30277, opened 2024-05-20
fix(rds): readers not always be created after the writer

### LINKED ISSUE #30260: rds: Bug in writer/reader definition of clusters
### Describe the bug

When creating a fresh cluster with readers and writer, one of the reader instances is effectively created before the writer and becomes the actual writer. So the opposite of what's written in the CDK code happens. This is a bug and it's reproducible.

### Expected Behavior

The expected behavior is that the writer specified in CDK is indeed the writer when the cluster is created.

### Current Behavior

The current behavior is that the writer becomes the reader, and viceversa, on a fresh cluster creation, which is the opposite of what's specified in CDK.

The CDK puts the writer instance first in CloudFormation, but that's not enough to guarantee that the first instance being created in the cluster because CloudFormation creates resources in parallel.

### Reproduction Steps

Code:

```
import { App, RemovalPolicy, Stack, StackProps } from 'aws-cdk-lib';
import { InstanceClass, InstanceSize, InstanceType, SubnetType, Vpc } from 'aws-cdk-lib/aws-ec2';
import {
  AuroraMysqlEngineVersion,
  ClusterInstance,
  Credentials,
  DatabaseCluster,
  DatabaseClusterEngine,
} from 'aws-cdk-lib/aws-rds';

export class RDSReproduceStack extends Stack {
  constructor(parent: App, id: string, props: StackProps, vpc: Vpc) {
    super(parent, id, props);

    const cluster = new DatabaseCluster(this, 'testcluster', {
      clusterIdentifier: 'testcluster',
      engine: DatabaseClusterEngine.auroraMysql({ version: AuroraMysqlEngineVersion.VER_3_04_1 }),
      credentials: Credentials.fromGeneratedSecret('myuser', {}),
      writer: ClusterInstance.provisioned('writer', {
        instanceIdentifier: 'writer',
        instanceType: InstanceType.of(InstanceClass.BURSTABLE4_GRAVITON, InstanceSize.MEDIUM),
        publiclyAccessible: false,
      }),
      serverlessV2MinCapacity: 0.5,
      serverlessV2MaxCapacity: 1.0,
      readers: [
        ClusterInstance.serverlessV2('reader', {
          scaleWithWriter: true,
          publiclyAccessible: false,
        }),
      ],
      vpc: vpc,
      vpcSubnets: {
        subnetType: SubnetType.PRIVATE_WITH_EGRESS,
      },
    });
  }
}
```

Result, see in the image that the writer is the reader and viceversa, on a freshly created cluster:

snap 240508161922 wxXVOsXL

### Possible Solution

The possible solution is to add a dependency on the writer instance to all the reader instances here: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-rds/lib/cluster.ts , to guarantee that the writer instance is the first instance being created in the cluster and thus become the writer.

_No response_

### CDK CLI Version

2.130.0

### Framework Version

_No response_

### Node.js Version

18

### OS

Ubuntu

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

pahud:
This PR ensures the dependency on the readers always to be created after the writer. This might not be the best solution as all the readers will not start provisioning until the writer is completed. Another solution is to build a custom resource to check if the writer has started provisioning, if yes, return success and let all the dependent readers start provisioning. But that would require a new custom resource.

- [x] unit tests
- [x] update integ tests
  - fixed the integ error `"Cannot find version 8.0.mysql_aurora.3.01.0 for aurora-mysql` for `integ.cluster-instance-id`

### Issue # (if applicable)

Closes #30260

### Checklist

----


godwingrs22 (approved):
LGTM. Thanks @pahud for your contribution.

pahud:
Exemption Request

PR #30298, opened 2024-05-22
feat(stepfunctions-tasks): allow BedrockInvokeModel to use JsonPath

### LINKED ISSUE #29229: (stepfunctions-tasks): BedrockInvokeModel cannot use JsonPath to specify input/output S3 URIs
### Describe the bug

If you try and use JsonPath to specify the S3 URIs that BedrockInvokeModel will read from and write from, you get an error.

### Expected Behavior

I expect to be able to provide S3 URIs to BedrockInvokeModel using JsonPath, so I can have dynamic paths for different state machine executions.

### Current Behavior

I get the error

```
jsii.errors.JavaScriptError:
  Error: Field references must be the entire string, cannot concatenate them (found 's3://${Token[prompt_bucket.348]}/${Token[prompt_key.349]}')
```

### Reproduction Steps

```
tasks.BedrockInvokeModel(
    scope=self,
    id="Call LLM",
    model=aws_bedrock.FoundationModel.from_foundation_model_id(
        scope=self,
        _id="Model",
        foundation_model_id=aws_bedrock.FoundationModelIdentifier.ANTHROPIC_CLAUDE_V2_1,
    ),
    input=tasks.BedrockInvokeModelInputProps(
        s3_location=aws_s3.Location(
            bucket_name=aws_stepfunctions.JsonPath.string_at("$.prompt_bucket"),
            object_key=aws_stepfunctions.JsonPath.string_at("$.prompt_key"),
        ),
    ),
    output=tasks.BedrockInvokeModelOutputProps(
        s3_location=aws_s3.Location(
            bucket_name=aws_stepfunctions.JsonPath.string_at("$.response_bucket"),
            object_key=aws_stepfunctions.JsonPath.string_at("$.response_key"),
        ),
    ),
)
```

### Possible Solution

The issue here is that the S3 locations are specified as an `aws_s3.Location`. This object expects a separate `bucket_name` and `object_key`. However, BedrockInvokeModel expects an S3 URI. This mismatch is handled in the code by building an S3 URI from the S3 Location 

https://github.com/aws/aws-cdk/blob/83aa3956ad035f534a7d96374844085919be3e8a/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts#L208-L213

The problem is that field references cannot be concatenated in this way, hence the error I see.

This **makes the S3 integration basically unusable from the CDK**, because the S3 locations must be hardcoded, which cannot work for any real world workload.

My suggestion is to extend `BedrockInvokeModelInputProps` and `BedrockInvokeModelOutputProps` to allow either the existing `s3Location` field to be used (for backwards compatibility), or a `string` field called `s3Uri`. If `s3Uri` were provided, it would be used directly, allowing for the use of JsonPaths for the URI.

_No response_

### CDK CLI Version

2.129.0 (build d5ab0df)

### Framework Version

2.129.0

### Node.js Version

v18.17.1

### OS

Ubuntu (Windows Subsystem for Linux)

### Language

### Language Version

Python 3.11.6

### Other information

_No response_

shikha372:
### Issue # (if applicable)

Closes #29229.

### Reason for this change

When trying to use JsonPath to specify the S3 URIs that BedrockInvokeModel will read from and write from, you get an error.

Example of the Error message:

`jsii.errors.JavaScriptError: Error: Field references must be the entire string, cannot concatenate them (found 's3://${Token[prompt_bucket.348]}/${Token[prompt_key.349]}')`

### Description of changes

Extended the inputPath property to be allowed as an input value for the task state.
Instead of adding a new S3Uri props in current `BedrockInvokeModelProps` as proposed in the original issue, leveraged the `inputPath` property that is already defined in   `sfn.TaskStateBaseProps` and being extended by `BedrockInvokeModelInputProps` and `BedrockInvokeModelOutputProps`

**Limitation:**  We cannot limit the resource policy to specific input token for which the value might be coming from the prompt, so had to keep it as [*] here. 

Added unit tests.
Successful deployment of integration tests in the account.

### Checklist
- [x] Unit Tests
- [x] Integration Tests
- [x] Updated ReadMe

----


packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -124,12 +124,14 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
...
     super(scope, id, props);
+
     this.integrationPattern = props.integrationPattern ?? sfn.IntegrationPattern.REQUEST_RESPONSE;
 
     validatePatternSupported(this.integrationPattern, BedrockInvokeModel.SUPPORTED_INTEGRATION_PATTERNS);
 
     const isBodySpecified = props.body !== undefined;
-    const isInputSpecified = props.input !== undefined && props.input.s3Location !== undefined;
+    //Either specific props.input with bucket name and object key or input s3 path
+    const isInputSpecified = (props.input !== undefined && props.input.s3Location !== undefined) || (props.inputPath !== undefined);
```
aaythapa, line 134: if `props.input != undefined` is true why do we need to check for `props.input.s3Location` (should it be `||` there instead of `&&`)?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -155,7 +157,21 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
...
 
-    if (this.props.input !== undefined && this.props.input.s3Location !== undefined) {
+    if (this.props.inputPath !== undefined) {
+      policyStatements.push(
+        new iam.PolicyStatement({
+          actions: ['s3:GetObject'],
+          resources: [
+            Stack.of(this).formatArn({
+              region: '',
+              account: '',
```
aaythapa, line 167: for my own understanding, why are region and account `''`?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -207,11 +237,13 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
         Body: this.props.body?.value,
         Input: this.props.input?.s3Location ? {
           S3Uri: `s3://${this.props.input.s3Location.bucketName}/${this.props.input.s3Location.objectKey}`,
-        } : undefined,
+        } : this.props.inputPath ? { S3Uri: this.props.inputPath } : undefined,
```
aaythapa, line 240: i see the `S3Uri` has a specific format, should we check if `inputPath` and `outputPath` matches that format?

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/bedrock/invoke-model.test.ts
```diff
@@ -203,6 +203,52 @@ describe('Invoke Model', () => {
...
+      inputPath: sfn.JsonPath.stringAt('$.prompt'),
+      outputPath: sfn.JsonPath.stringAt('$.prompt'),
+    });
+
+    new sfn.StateMachine(stack, 'StateMachine', {
+      definitionBody: sfn.DefinitionBody.fromChainable(task),
+    });
+
+    // THEN
+    expect(stack.resolve(task.toStateJson())).toEqual({
```
aaythapa, line 221: Can we test that the right policies are added to the `policyStatement` whenever `inputPath` or `outputPath` are set

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
shikha372: This is how customer can specify an input : 
1. either using props.input, if they are using this then input.s3Location is required which is defined under the input type (BedrockInvokeModelInputProps), so we need to check for both, hence an `AND` condition. 

2. use inputPath in which case input.props is not needed 
Result should be either of this condition, hence an or

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
shikha372: afaik, this is for global bucket.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
shikha372: current s3 uri is limiting and enforcing that format., this props is to allow user to break out of that specific format and allow value usage from prompt which would be at step-function runtime i suppose which might be difficult to check here same way we can't limit our policy to that specific input hence used `*` as above.

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/bedrock/invoke-model.test.ts
shikha372: sure, will add a test for the policy statement check.

shikha372: added test along with squashish commits

aaythapa (approved):

shikha372:
@mergifyio update

shikha372:
updating manually since mergifyio update is failing

PR #30302, opened 2024-05-22
fix(stepfunctions-tasks): runtime language used to evaluate expressions is ignored

colifran:
### Reason for this change

`EvaluateExpression` exposes a runtime property that can be used to configure the runtime language used to evaluate an expression. When the handler for this was migrated into the handler framework we hid the runtime property and didn't make it configurable. As a result, when the runtime property is specified as part of `EvaluateExpressionProps` it ends up being dropped in place of the code generated runtime.

### Description of changes

Added a configurable runtime property to the generated `EvalNodejsSingletonFunctionProps` interface and set this property using runtime property on `EvaluateExpressionProps` if one was provided. Otherwise, the current node 18 default is used.

Unit test for codegen with eval-nodejs-provider. Integ test for default `EvaluateExpression` runtime (we already test a configurable runtime, unfortunately this was the same as the default so this bug was not caught).

### Checklist

----


colifran:
Note to the reviewer. We already have an integ test for this [here](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/integ.evaluate-expression.ts). This bug was missed because the default runtime is the same as the runtime that was being passed in as part of the runtime which is node 18. I don't want to create a new integ test using node 16 because that is on a path to deprecation. This construct doesn't current support node 20 which means we can't create a new integ test using that either. We could add support for node 20 in this PR, but I don't want to intermix a fix and a feat. Therefore, I'm marking this as integ test exempt.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/integ.evaluate-expression-default-runtime.ts
```diff
@@ -0,0 +1,56 @@
...
+      resultPath: '$.now',
+    });
+
+    const statemachine = new sfn.StateMachine(this, 'StateMachine', {
+      definition: sum
+        .next(multiply)
+        .next(
+          new sfn.Wait(this, 'Wait', {
+            time: sfn.WaitTime.secondsPath('$.d'),
+          }),
```
bergjaak, line 40: is there a reason to add a wait? Just wondering since it will make the test slower

packages/@aws-cdk/custom-resource-handlers/test/custom-resources-framework/expected/singleton-function-eval-nodejs.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable prettier/prettier,max-len */
+import * as path from "path";
+import { Construct } from "constructs";
+import * as lambda from "../../../aws-lambda";
+
+export class EvalNodejsSingletonFunction extends lambda.SingletonFunction {
```
bergjaak, line 6: Just for my understanding, is this used in `packages/@aws-cdk/custom-resource-handlers/test/custom-resources-framework/expected/singleton-function-eval-nodejs.ts`? If not, what's the purpose of this?

colifran: Yeah. It's just a template to use as "expected" to compare against a result in a unit test.

bergjaak (approved):

PR #30333, opened 2024-05-24
feat(s3): add `autoDeleteObjectsLogGroup` to pass log group to custom resource lambda

### LINKED ISSUE #24815: (aws-s3): retention on S3 autoDeleteObjects lambda CloudWatch log group is Never expire

samson-keung:
### Issue # (if applicable)

Closes #24815.

### Reason for this change

To allow log group customization on the custom resource lambda created for the `autoDeleteObjects` feature.

### Description of changes

There are 2 changes to implement the `autoDeleteObjectsLogGroup` prop:

1. `CustomResourceProviderBase` is updated to accept an optional `logGroupName` prop. If it is defined, it will be passed to the `AWS::Lambda::Function` resource definition (namely, setting the [`LoggingConfig` prop](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-loggingconfig)).
2. `s3.Bucket` is updated to accept an optional `autoDeleteObjectsLogGroup` prop. The idea is to make the API experience similar to the [`logGroup` prop](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.Function.html#loggroup) on `lambda.Function`. 

Unit tests have been added for both the `s3.Bucket` class and `core.CustomResourceProvider` class

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-s3/README.md
+++ b/packages/aws-cdk-lib/aws-s3/README.md
@@ -627,15 +627,14 @@ When `autoDeleteObjects` is enabled, `s3:PutBucketPolicy` is added to the bucket
 Pass a custom log group via the `autoDeleteObjectsLogGroup` option, which will be used by the custom resource lambda.
 
 ```ts
-import { LogGroup, RetentionDays } from 'aws-cdk-lib/aws-logs';
+import { ILogGroup } from 'aws-cdk-lib/aws-logs';
+
+declare const logGroup: ILogGroup;
 
 const bucket = new s3.Bucket(this, 'MyTempFileBucket', {
   removalPolicy: cdk.RemovalPolicy.DESTROY,
   autoDeleteObjects: true,
-  autoDeleteObjectsLogGroup: new LogGroup(this, 'LogGroup', {
-    logGroupName: 'MyLogGroup',
-    retention: RetentionDays.TWO_YEARS
-  })
+  autoDeleteObjectsLogGroup: logGroup
 });
 ```
 
--- a/packages/aws-cdk-lib/aws-s3/lib/bucket.ts
+++ b/packages/aws-cdk-lib/aws-s3/lib/bucket.ts
@@ -10,7 +10,7 @@ import { parseBucketArn, parseBucketName } from './util';
 import * as events from '../../aws-events';
 import * as iam from '../../aws-iam';
 import * as kms from '../../aws-kms';
-import * as logs from '../../aws-logs/index';
+import * as logs from '../../aws-logs';
 import {
   CustomResource,
   Duration,
--- a/packages/aws-cdk-lib/aws-s3/test/bucket.test.ts
+++ b/packages/aws-cdk-lib/aws-s3/test/bucket.test.ts
@@ -3,7 +3,7 @@ import { testDeprecated } from '@aws-cdk/cdk-build-tools';
 import { Annotations, Match, Template } from '../../assertions';
 import * as iam from '../../aws-iam';
 import * as kms from '../../aws-kms';
-import { LogGroup } from '../../aws-logs';
+import * as logs from '../../aws-logs';
 import * as cdk from '../../core';
 import * as s3 from '../lib';
 
@@ -3470,7 +3470,7 @@ describe('bucket', () => {
     new s3.Bucket(stack, 'MyBucket', {
       removalPolicy: cdk.RemovalPolicy.DESTROY,
       autoDeleteObjects: true,
-      autoDeleteObjectsLogGroup: new LogGroup(stack, 'LogGroup1', {
+      autoDeleteObjectsLogGroup: new logs.LogGroup(stack, 'LogGroup1', {
         logGroupName: 'MyLogGroup',
       }),
     });
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-s3/test/integ.bucket-auto-delete-objects.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-s3/test/integ.bucket-auto-delete-objects.ts
@@ -4,7 +4,6 @@ import { IntegTest } from '@aws-cdk/integ-tests-alpha';
 import { Construct } from 'constructs';
 import * as s3 from 'aws-cdk-lib/aws-s3';
 import * as logs from 'aws-cdk-lib/aws-logs';
-import { RetentionDays } from 'aws-cdk-lib/aws-logs';
 import { AwsCustomResource, AwsCustomResourcePolicy, PhysicalResourceId } from 'aws-cdk-lib/custom-resources';
 import { STANDARD_CUSTOM_RESOURCE_PROVIDER_RUNTIME } from '../../config';
 
@@ -42,7 +41,7 @@ class TestStack extends Stack {
       autoDeleteObjects: true,
       autoDeleteObjectsLogGroup: new logs.LogGroup(this, 'LogGroup', {
         logGroupName: 'AutoDeleteObjectsLambdaLogs',
-        retention: RetentionDays.THREE_DAYS,
+        retention: logs.RetentionDays.THREE_DAYS,
       }),
     });
 


packages/aws-cdk-lib/aws-s3/README.md
```diff
@@ -624,10 +624,17 @@ enable the`autoDeleteObjects` option.
 
 When `autoDeleteObjects` is enabled, `s3:PutBucketPolicy` is added to the bucket policy. This is done to allow the custom resource this feature is built on to add a deny policy for `s3:PutObject` to the bucket policy when a delete stack event occurs. Adding this deny policy prevents new objects from being written to the bucket. Doing this prevents race conditions with external bucket writers during the deletion process.
 
+Pass a custom log group via the `autoDeleteObjectsLogGroup` option, which will be used by the custom resource lambda.
+
 ```ts
+import { ILogGroup } from 'aws-cdk-lib/aws-logs';
+
+declare const logGroup: ILogGroup;
```
aaythapa, line 632: nit: I think its more valuable if you actually define a log group here. The log group from the integ test should do.

```suggestion
const logGroup = new logs.LogGroup(this, 'LogGroup', {
        logGroupName: 'AutoDeleteObjectsLambdaLogs',
        retention: logs.RetentionDays.THREE_DAYS,
      }),
```

packages/aws-cdk-lib/aws-s3/test/bucket.test.ts
```diff
@@ -3463,6 +3464,26 @@ describe('bucket', () => {
...
+    });
+
+    Template.fromStack(stack).hasResourceProperties('AWS::Lambda::Function', {
+      LoggingConfig: {
+        LogGroup: {
+          'Ref': 'LogGroup106AAD846',
+        },
+      },
+    });
+  });
```
aaythapa, line 3485: nit: should we also be checking if the log group resource was created?

aaythapa (commented):
thanks for making this change!

I see from this https://github.com/aws/aws-cdk/issues/24815#issuecomment-2021514101 comment from the linked issue that someone made a similar prototype and they faced this error.

```
An unexpected behavior was that the provider lambda was re-creating the log group after I got it to successfully delete itself on stack deletion, but the new log group came back with infinite duration (argh!). This is because the default role for the lambda has permission to create a log group (which was occurring when it issued a final log stream on delete). Fortunately policyStatements was already wired up so I just explicitly denied CreateLogGroup for the Lambda role.
```
Wondering if you faced something similar (I see similarities between the prototype and you're change so just checking)

samson-keung:
> thanks for making this change!
> 
> I see from this [#24815 (comment)](https://github.com/aws/aws-cdk/issues/24815#issuecomment-2021514101) comment from the linked issue that someone made a similar prototype and they faced this error.
> 
> ```
> An unexpected behavior was that the provider lambda was re-creating the log group after I got it to successfully delete itself on stack deletion, but the new log group came back with infinite duration (argh!). This is because the default role for the lambda has permission to create a log group (which was occurring when it issued a final log stream on delete). Fortunately policyStatements was already wired up so I just explicitly denied CreateLogGroup for the Lambda role.
> ```
> 
> Wondering if you faced something similar (I see similarities between the prototype and you're change so just checking)

I did not. 

> thanks for making this change!
> 
> I see from this [#24815 (comment)](https://github.com/aws/aws-cdk/issues/24815#issuecomment-2021514101) comment from the linked issue that someone made a similar prototype and they faced this error.
> 
> ```
> An unexpected behavior was that the provider lambda was re-creating the log group after I got it to successfully delete itself on stack deletion, but the new log group came back with infinite duration (argh!). This is because the default role for the lambda has permission to create a log group (which was occurring when it issued a final log stream on delete). Fortunately policyStatements was already wired up so I just explicitly denied CreateLogGroup for the Lambda role.
> ```
> 
> Wondering if you faced something similar (I see similarities between the prototype and you're change so just checking)

Let me try reproducing the issue and get back to you. From your call out, I realized the integ test is not generating expected template, namely, the `LoggingConfig` on the lambda isn't set.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1480,6 +1483,16 @@ export interface BucketProps {
...
+  /**
+   * The log group to use for the custom resource lambda backing the `autoDeleteObjects` feature
+   *
+   * When `autoDeleteObjects` is set to true, a customer resource backed by a Lambda function
+   * is created. The lambda will use the log group passed in to this prop if defined.
+   *
+   * @default the default log group created by Lambda
+   */
+  readonly autoDeleteObjectsLogGroup?: logs.ILogGroup;
+
```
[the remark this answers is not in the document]
samson-keung: It is ignored.

packages/aws-cdk-lib/aws-s3/README.md
samson-keung: My first revision was defining the log group explicitly. But I found the [Lambda REAME](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda-readme.html#advanced-logging) that uses this style. But I agree defining the log group is more valuable to the reader. Will update

packages/aws-cdk-lib/aws-s3/test/bucket.test.ts
samson-keung: Good idea. Will update

samson-keung:
This PR has a problem with the integ test. Looking at [the snapshot](https://github.com/aws/aws-cdk/pull/30333/files#diff-a2ffcbe0e7a8d5da673e46dbb29c0b26f7f3ff2784313a2937f8a1009a14f6c5), it is apparent that the log group is created but it is not referenced anywhere in the template. The expected template should have the custom resource lambda referencing the log group like so:
```
  "CustomS3AutoDeleteObjectsCustomResourceProviderHandler9D90184F": {
   "Type": "AWS::Lambda::Function",
   "Properties": {
    // ...
    "LoggingConfig": {
     "LogGroup": {
      "Ref": "LogGroupF5B46931"
     }
    }
   }
  }
```

Upon further investigation, I have learned that the `autoDeleteObjects` prop creates a stack wide singleton provider lambda. Therefore, with this PR, the following happens:
```
const myLogGroupA = new LogGroup(...)
const bucket = new s3.Bucket(this, 'BucketA', {
  autoDeleteObjects: true, 
  autoDeleteObjectsLogGroup: myLogGroupA
});

// The above creates a singleton AutoDeleteObjectsProvider lambda and set the LoggingConfig on the lambda to use "myLogGroupA"

const myLogGroupB = new LogGroup(...)
const bucketThatWillBeRemoved = new s3.Bucket(this, 'BucketB', {
  autoDeleteObjects: true,
  autoDeleteObjectsLogGroup: myLogGroupB
});

// This second bucket DOES NOT create a new AutoDeleteObjectsProvider lambda. But rather it will retrieve and re-use the existing one. Hence, "myLogGroupB" is ignored
```

-----

With the above, the approach in this PR creates a poor UX as user must remember to set the `autoDeleteObjectsLogGroup` on the very first bucket with `autoDeleteObjects: true`. And all subsequent buckets with `autoDeleteObjectsLogGroup` will be ignored quietly. Due to this, I am abandoning this approach and closing this PR. Please refer to [the issue](https://github.com/aws/aws-cdk/issues/24815) for further updates.

PR #30346, opened 2024-05-27
fix(globalaccelerator-endpoints): add preserveClientIp option for net…

sj5n:
…work loadbalancer

### Reason for this change

preserveClientIp was missing for GlobalAccelerator Endpoints when using a network loadbalancer.

### Description of changes

* add missing network load balancer endpoint prop.

Added unit tests.

### Checklist


SaschaJanssen:
Exemption Request: No integration but unit tests added. fix just contains a new property to the global accelerator endpoint.

packages/aws-cdk-lib/aws-globalaccelerator-endpoints/lib/nlb.ts
```diff
@@ -14,6 +14,19 @@ export interface NetworkLoadBalancerEndpointProps {
...
+  /**
+   * Forward the client IP address in an `X-Forwarded-For` header
+   *
+   * GlobalAccelerator will create Network Interfaces in your VPC in order
+   * to preserve the client IP address.
+   *
+   * Client IP address preservation is supported only in specific AWS Regions.
+   * See the GlobalAccelerator Developer Guide for a list.
+   *
+   * @default true if available
```
scorbiere, line 27: According to the documentation, my understanding is that the default value is `false` : 
https://docs.aws.amazon.com/global-accelerator/latest/dg/preserve-client-ip-address.html
> When you add an Network Load Balancer with security groups as an endpoint in Global Accelerator, client IP address preservation is not enabled by default.

```suggestion
   * @default false
```

scorbiere (changes requested):
Thank you for your contribution. Would you mind updating the default value in the comment? Then I can approve the PR.

SaschaJanssen:
@scorbiere thanks, updated the doc.

scorbiere (approved):
LGTM, thanks for the update.

scorbiere:
@mergify update

scorbiere (approved):
Re-approving after updating the branch.

PR #30352, opened 2024-05-27
feat(apprunner): add kmsKey property for the AppRunner Service class

### LINKED ISSUE #30365: (aws-apprunner): support a customer managed key
### Describe the feature

Support a customer managed key to encrypt all stored copies of your application source image or source bundle.
At the moment,  L2 Construct (alpha module) cannot use a customer managed key.

### Use Case

In App Runner, the source code and images that are copied are typically encrypted using AWS-managed keys. 
However, there may be cases where you want to use a customer-managed key for encryption to enhance security or meet auditing requirements.

For this purpose, App Runner supports encryption using customer-managed keys.

### Proposed Solution

Add a property to use a KMS key

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.143.0

all

mazyu36:
### Issue # (if applicable)

Close #30365.

### Reason for this change
AppRunner supports for using a customer managed key to encrypt  all stored copies of your application source image or source bundle.

https://docs.aws.amazon.com/apprunner/latest/dg/security-data-protection-encryption.html

But L2 Construct (alpha module) cannot use a customer managed key.

### Description of changes
Add kmsKey property to the Service class.

Add unit tests and integ tests

### Checklist

----


moelasmar:
can you create an issue to explain more your use case.

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -715,6 +716,13 @@ export interface ServiceProps {
    * @default - no health check configuration
    */
   readonly healthCheck?: HealthCheck;
+
+  /**
+   * The customer managed key that AWS App Runner uses to encrypt copies of the source repository and service logs.
+   *
+   * @default - Use an AWS managed key
```
moelasmar, line 723: I think the default value will be `no managed key` or `undefined`

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -1239,6 +1247,9 @@ export class Service extends cdk.Resource implements iam.IGrantable {
           this.renderCodeConfiguration(this.source.codeRepository!.codeConfiguration.configurationValues!) :
           undefined,
       },
+      encryptionConfiguration: this.props.kmsKey ? {
+        kmsKey: this.props.kmsKey.keyArn,
+      } : undefined,
```
moelasmar, line 1252: I think we need to update the role assigned to the App runner to use the KMS key. I will run some test to verify this.

mazyu36: Thank you for the review!

I also thought the same, but when I actually tried deploying using a CMK, it was able to host without any issues, so I had concluded that no additional permissions were required.

I may have made a mistake somewhere, so I will double check again. If you happen to find out anything else, please let me know as well.

moelasmar: did some testing, and it looks like during the App Runner service creation, CreateGrant method got created on the passed KMS that should allow this action for the account Principal or the role that create the CFN stack. so, I think this grant is not required.

moelasmar (approved):
it looks good, thanks for creating this PR

mazyu36:
@moelasmar 
Thanks for the review and testing!

PR #30358, opened 2024-05-28
feat(apprunner): add AutoScalingConfiguration for AppRunner Service

### LINKED ISSUE #30353: apprunner: support auto-scaling configuration management
### Describe the feature

AppRunner supports Auto-Scaling configuration management.
https://aws.amazon.com/about-aws/whats-new/2023/09/aws-app-runner-auto-scaling-management/

[L1 Construct ](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apprunner.CfnAutoScalingConfiguration.html)already supported, but current L2 Construct does not support.

### Use Case

To use auto-scaling for AppRunner Service.

### Proposed Solution

* Add `AutoScalingConfiguration` class.
* Add autoscaling property to the `Service` class.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.143.0

All

mazyu36:
### Issue # (if applicable)

Closes #30353 .

### Reason for this change
At the moment, L2 Construct does not support a custom auto scaling configuration for the AppRunner Service.

### Description of changes
* Add `AutoScalingConfiguration` Class
* Add `autoScalingConfiguration` property to the `Service` Class

### Checklist

----


packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -656,6 +657,18 @@ export interface ServiceProps {
...
 
+  /**
+   * Specifies an App Runner Auto Scaling Configuration.
+   *
+   * A default configuration is either the AWS recommended configuration,
+   * or the configuration you set as the default.
+   *
+   * @see https://docs.aws.amazon.com/apprunner/latest/dg/manage-autoscaling.html
+   *
+   * @default - use a default Auto Scaling Configuration.
```
lpizzinidev, line 668:
```suggestion
   * @default - the latest revision of a default auto scaling configuration is used.
```
[Docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apprunner-service.html#cfn-apprunner-service-autoscalingconfigurationarn)

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,181 @@
+import * as cdk from 'aws-cdk-lib/core';
+import { Construct } from 'constructs';
+import { CfnAutoScalingConfiguration } from 'aws-cdk-lib/aws-apprunner';
+
+/**
+ * Properties of the App Runner Auto Scaling Configuration.
+ */
+export interface AutoScalingConfigurationProps {
```
lpizzinidev, line 8: Shouldn't we allow to specify [`tags`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apprunner-autoscalingconfiguration.html#cfn-apprunner-autoscalingconfiguration-tags) as well?

packages/@aws-cdk/aws-apprunner-alpha/test/auto-scaling-configuration.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    MaxConcurrency: 150,
+    MaxSize: 20,
+    MinSize: 5,
+  });
+});
+
+test('create an Auto scaling Configuration without all properties', () => {
+  // WHEN
+  new AutoScalingConfiguration(stack, 'AutoScalingConfiguration', {
+  });
```
lpizzinidev, line 31:
```suggestion
  new AutoScalingConfiguration(stack, 'AutoScalingConfiguration');
```
Already defaulting to `{}`

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,181 @@
...
+   * The revision of the Auto Scaling Configuration.
+   * @attribute
+   */
+  readonly autoScalingConfigurationRevision: number;
+
+  public constructor(scope: Construct, id: string, props: AutoScalingConfigurationProps = {}) {
+    super(scope, id, {
+      physicalName: props.autoScalingConfigurationName,
+    });
+
```
lpizzinidev, line 156: Can you also add validation for the `autoScalingConfigurationName` property ([docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apprunner-autoscalingconfiguration.html#cfn-apprunner-autoscalingconfiguration-autoscalingconfigurationname))?

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,181 @@
...
+    if (props.minSize !== undefined && (props.minSize < 1 || props.minSize > 25)) {
```
lpizzinidev, line 157:
```suggestion
    if (props.minSize !== undefined && !cdk.Token.isUnresolved(props.minSize) && (props.minSize < 1 || props.minSize > 25)) {
```
We should also verify that the attribute value is resolved before applying validation.

lpizzinidev (changes requested):
Thanks 👍 Left some comments for minor adjustments

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
mazyu36: Thanks. I added validation and unit tests.

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
mazyu36: Looking at the other L2 Constructs, it seemed like there were few resources adding tags as properties, so I didn't add them.
Since tags can also be added through aspects, I thought they might be unnecessary like other resources. What do you think?
I would appreciate your opinion on the criteria for adding tags, as I'm not sure about it.

mazyu36:
@lpizzinidev 
Thank you for the review!

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,181 @@
...
+   *
+   * @default 1
+   */
+  readonly maxSize?: number;
+
+  /**
+   * The minimum number of instances that App Runner provisions for a service.
+   * The service always has at least minSize provisioned instances.
+   *
+   * @default 25
```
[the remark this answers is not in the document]
mazyu36: Sorry, you are right.

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
lpizzinidev: Makes sense 👍 Thanks for clarifying!

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    if (props.autoScalingConfigurationName !== undefined) {
+      if (props.autoScalingConfigurationName.length < 4 || props.autoScalingConfigurationName.length > 32) {
+        throw new Error(`autoScalingConfigurationName must be between 4 and 32 characters long, but it has ${props.autoScalingConfigurationName.length} characters.`);
+      }
+
+      if (!/^[A-Za-z0-9][A-Za-z0-9\-_]{3,31}$/.test(props.autoScalingConfigurationName)) {
+        throw new Error(`autoScalingConfigurationName ${props.autoScalingConfigurationName} must start with a letter or number, and can contain only letters, numbers, hyphens, and underscores.`);
+      }
+    }
```
lpizzinidev, line 168:
```suggestion
    if (
      props.autoScalingConfigurationName !== undefined && 
      !cdk.Token.isUnresolved(props.autoScalingConfigurationName) &&
      !/^[A-Za-z0-9][A-Za-z0-9\-_]{3,31}$/.test(props.autoScalingConfigurationName)
    ) {
      throw new Error(`autoScalingConfigurationName must match the ^[A-Za-z0-9][A-Za-z0-9\-_]{3,31}$ pattern, got ${props.autoScalingConfigurationName}`);
    }
```
We could simplify as the check on the length is redundant.

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,194 @@
...
+    if (props.maxSize !== undefined && (props.maxSize < 1 || props.maxSize > 25)) {
+      throw new Error(`maxSize must be between 1 and 25, got ${props.maxSize}`);
+    }
+    if (props.minSize !== undefined && props.maxSize !== undefined && !(props.minSize < props.maxSize)) {
+      throw new Error('maxSize must be greater than minSize');
+    }
+    if (props.maxConcurrency !== undefined && (props.maxConcurrency < 1 || props.maxConcurrency > 200)) {
+      throw new Error(`maxConcurrency must be between 1 and 200, got ${props.maxConcurrency}`);
+    }
```
lpizzinidev, line 181:
```suggestion
    const isMinSizeDefined = props.minSize !== undefined && !cdk.Token.isUnresolved(props.minSize)
    const isMaxSizeDefined = props.maxSize !== undefined && !cdk.Token.isUnresolved(props.maxSize)
    if (isMinSizeDefined && (props.minSize < 1 || props.minSize > 25)) {
      throw new Error(`minSize must be between 1 and 25, got ${props.minSize}`);
    }
    if (isMaxSizeDefined && (props.maxSize < 1 || props.maxSize > 25)) {
      throw new Error(`maxSize must be between 1 and 25, got ${props.maxSize}`);
    }
    if (isMinSizeDefined && isMaxSizeDefined && !(props.minSize < props.maxSize)) {
      throw new Error('maxSize must be greater than minSize');
    }
    if (
      props.maxConcurrency !== undefined && 
      !cdk.Token.isUnresolved(props.maxConcurrency) &&
      (props.maxConcurrency < 1 || props.maxConcurrency > 200)
    ) {
      throw new Error(`maxConcurrency must be between 1 and 200, got ${props.maxConcurrency}`);
    }
```
We should check if resolved for all validated props

lpizzinidev (changes requested):
Thanks 👍 Left suggestions for minor adjustments on validation

mazyu36:
@lpizzinidev 
Thanks. I've addressed your comments and fix unit tests.

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,198 @@
...
+   */
+  readonly maxConcurrency?: number;
+
+  /**
+   * The maximum number of instances that a service scales up to.
+   * At most maxSize instances actively serve traffic for your service.
+   *
+   * Must be between 1 and 25.
+   *
+   * @default 1
```
lpizzinidev, line 32:
```suggestion
   * @default 25
```

lpizzinidev (approved):
Thanks 👍

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * Imports an App Runner Auto Scaling Configuration from attributes
+   */
+  public static fromAutoScalingConfigurationAttributes(scope: Construct, id: string,
+    attrs: AutoScalingConfigurationAttributes): IAutoScalingConfiguration {
+    const autoScalingConfigurationArn = attrs.autoScalingConfigurationArn;
+    const autoScalingConfigurationName = attrs.autoScalingConfigurationName;
+    const autoScalingConfigurationRevision = attrs.autoScalingConfigurationRevision;
+
+    class Import extends cdk.Resource {
```
pahud, line 106:
```suggestion
    class Import extends cdk.Resource implements  IAutoScalingConfiguration  {
```

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,198 @@
...
+      public readonly autoScalingConfigurationRevision = autoScalingConfigurationRevision
+    }
+
+    return new Import(scope, id);
+  }
+
+  /**
+   * Imports an App Runner Auto Scaling Configuration from its ARN
+   */
+  public static fromArn(scope: Construct, id: string, autoScalingConfigurationArn: string): IAutoScalingConfiguration {
```
pahud, line 118: I guess `fromArn` won't pass the linter but I am not 100% sure. I guess `fromAutoScalingConfigurationArn` would be expected. But I might be wrong.

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,198 @@
...
+    const arn = cdk.Stack.of(scope).splitArn(autoScalingConfigurationArn, cdk.ArnFormat.SLASH_RESOURCE_NAME);
+
+    const resourceParts = arn.resourceName?.split('/');
```
pahud, line 121: I think `arn.resourceName?.split('/');` won't work when `autoScalingConfigurationArn` is a Token.

[splitArn](https://github.com/aws/aws-cdk/blob/c1e1b4d777a65ba3fcb6a108c43b10b4e7b9ace6/packages/aws-cdk-lib/core/lib/stack.ts#L804) can handle token but split only handles string. I guess we need Fn.Split and Fn.Select.

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -656,6 +657,18 @@ export interface ServiceProps {
...
+   * Specifies an App Runner Auto Scaling Configuration.
+   *
+   * A default configuration is either the AWS recommended configuration,
+   * or the configuration you set as the default.
+   *
+   * @see https://docs.aws.amazon.com/apprunner/latest/dg/manage-autoscaling.html
+   *
+   * @default - the latest revision of a default auto scaling configuration is used.
+   */
+  readonly autoScalingConfiguration?: IAutoScalingConfiguration;
```
pahud, line 670: I would prefer just `autoScaling?: IAutoScalingConfiguration;`

pahud (commented):
added some of my comments

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
pahud: In AWS CDK L2s. If the resource name is `Foo`, the construct class of `Foo` should have at least one of fromFooName(), fromFooAttributes() or fromFooArn() and those methods should return `IFoo`. To make it easier, we have a convention to define a tiny `Import` class in those fromXxxx methods that implements IFoo and so we just need to `return new Import()` from those fromXxxx construct methods.

mazyu36: Thank you for the review. I understand that while the `fromArn` method has been implemented, additional `fromXXX` methods are needed. After rechecking the design guidelines, it seems better to have a `fromAttributes` method, so I will add that.​​​​​​​​​​​​​​​​

https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#from-attributes

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    class Import extends cdk.Resource {
+      public readonly autoScalingConfigurationArn = autoScalingConfigurationArn
+      public readonly autoScalingConfigurationName = autoScalingConfigurationName
+      public readonly autoScalingConfigurationRevision = autoScalingConfigurationRevision
+    }
+
+    return new Import(scope, id);
+  }
+
```
pahud, line 114: If we just know the Name and Revision, we probably can formate the Arn string.

How about just have Name and Revision in the interface:

```ts
export interface AutoScalingConfigurationAttributes {
  /**
   * The name of the Auto Scaling Configuration.
   */
  readonly autoScalingConfigurationName: string;

  /**
   * The revision of the Auto Scaling Configuration.
   */
  readonly autoScalingConfigurationRevision: number;
}
```

And in the fromXxx method:

```ts
public static fromAutoScalingConfigurationAttributes(scope: Construct, id: string,
    attrs: AutoScalingConfigurationAttributes): IAutoScalingConfiguration {

    class Import extends cdk.Resource implements IAutoScalingConfiguration {
       public readonly autoScalingConfigurationName = attrs.autoScalingConfigurationName;
       public readonly autoScalingConfigurationRevision = attrs.autoScalingConfigurationRevision;
       public readonly autoScalingConfigurationArn = Stack.of(this).formatArn({
           resource: 'autoscalingconfiguration',
           service: 'apprunner',
           resourceName: `${attrs.autoScalingConfigurationName}/${attrs.autoScalingConfigurationRevision}`,
      })
    }
    
    return new Import(scope, id);
};
```

If we can generate the Arn from given Name and Revision then Arn would not be required in the Attributes?

pahud (changes requested):

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
mazyu36: Thanks. I've fixed it.

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
mazyu36: Thank you. I will keep it as is, in the form of aligning with the class names or property names of the L1 construct.

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
mazyu36: Thank you. You are certainly right. I have removed arn from the attributes and made the correction.

mazyu36:
@pahud 
Thank you for the review. 
I've incorporated all your comments.

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,196 @@
...
+      throw new Error('maxSize must be greater than minSize');
+    }
+    if (
+      props.maxConcurrency !== undefined &&
+      !cdk.Token.isUnresolved(props.maxConcurrency) &&
+      (props.maxConcurrency < 1 || props.maxConcurrency > 200)
+    ) {
+      throw new Error(`maxConcurrency must be between 1 and 200, got ${props.maxConcurrency}`);
+    }
+
```
pahud, line 184: Consider the code like this

```ts
function validateAutoScalingConfiguration(props: {
  minSize?: number;
  maxSize?: number;
  maxConcurrency?: number;
}) {
  const isMinSizeDefined = typeof props.minSize === 'number';
  const isMaxSizeDefined = typeof props.maxSize === 'number';
  const isMaxConcurrencyDefined = typeof props.maxConcurrency === 'number';

  if (isMinSizeDefined && (props.minSize < 1 || props.minSize > 25)) {
    throw new Error(`minSize must be between 1 and 25, got ${props.minSize}`);
  }

  if (isMaxSizeDefined && (props.maxSize < 1 || props.maxSize > 25)) {
    throw new Error(`maxSize must be between 1 and 25, got ${props.maxSize}`);
  }

  if (isMinSizeDefined && isMaxSizeDefined && !(props.minSize < props.maxSize)) {
    throw new Error('maxSize must be greater than minSize');
  }

  if (isMaxConcurrencyDefined && (props.maxConcurrency < 1 || props.maxConcurrency > 200)) {
    throw new Error(`maxConcurrency must be between 1 and 200, got ${props.maxConcurrency}`);
  }
}
```

The main changes are:

1. Extracted the validation logic into a separate function `validateAutoScalingConfiguration` to make the code more modular and easier to test.

2. Used `typeof` to check if the properties are defined and a number as well as  `!cdk.Token.isUnresolved` because if it is unresolved it would not be number? I am not 100% sure here maybe you still need

```ts
typeof props.minSize === 'number' && !cdk.Token.isUnresolved(props.minSize);
```

3. Moved the `isMaxConcurrencyDefined` check into a separate condition to make the code more readable.

There are a few benefits to using `typeof` to check if the properties are defined and a number, instead of using 
`undefined` and `cdk.Token.isUnresolved`:

1. Simplicity and Readability: The `typeof` checks are more straightforward and easier to understand at a glance. They directly check the type of the value, which makes the code more self-documenting and easier to maintain.

2. Handling Unresolved Tokens: The `cdk.Token.isUnresolved` check is specifically for handling unresolved tokens in the AWS CDK. However, in this case, we're not just checking for unresolved tokens, but also ensuring the property is defined and a number. Using `typeof` handles all of these cases in a single, simple check. [[1]](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.TokenComparison.html)

3. Robustness: The `typeof` checks are more robust because they handle a wider range of cases. For example, if the property is set to `null`, the `typeof` check will still correctly identify it as not a number, whereas the 
`undefined` check would not catch this case.

4. Performance: The `typeof` checks are generally faster than using `undefined` and `cdk.Token.isUnresolved`
, as they are simpler operations. This can be especially beneficial in performance-critical code.

5. Consistency: Using `typeof` to check the type of a value is a common and widely-accepted pattern in JavaScript/TypeScript development. It aligns with the language's standard practices and makes the code more familiar to other developers.

In summary, the `typeof` checks provide a more straightforward, robust, and performant way to validate the properties, while also making the code more readable and consistent with standard JavaScript/TypeScript practices. This can lead to better maintainability and easier understanding of the validation logic.

This optimized version of the code is more concise, easier to read, and easier to maintain. It also follows best practices for input validation, such as using clear error messages and keeping the validation logic separate from the main application logic.

pahud (commented):
some minor suggested changes

packages/@aws-cdk/aws-apprunner-alpha/test/auto-scaling-configuration.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  // THEN
+  Template.fromStack(stack).hasResourceProperties('AWS::AppRunner::AutoScalingConfiguration', {
+    AutoScalingConfigurationName: Match.absent(),
+    MaxConcurrency: Match.absent(),
+    MaxSize: Match.absent(),
+    MinSize: Match.absent(),
+  });
+});
+
+test.each([0, 26])('invalid minSize', (minSize: number) => {
```
pahud, line 44:
```suggestion
test.each([-1, 0, 26])('invalid minSize', (minSize: number) => {
```

let's check -1 as well

packages/@aws-cdk/aws-apprunner-alpha/test/auto-scaling-configuration.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  ['test-autoscaling-configuration-name-over-limitation'],
+  ['-test'],
+  ['test-?'],
+])('invalid autoScalingConfigurationName (name: %s)', (autoScalingConfigurationName: string) => {
+  expect(() => {
+    new AutoScalingConfiguration(stack, 'AutoScalingConfiguration', {
+      autoScalingConfigurationName,
+    });
+  }).toThrow(`autoScalingConfigurationName must match the ^[A-Za-z0-9][A-Za-z0-9\-_]{3,31}$ pattern, got ${autoScalingConfigurationName}`);
+});
```
pahud, line 88:
```suggestion
});

test('create an Auto scaling Configuration with tags', () => {
  // WHEN
  const autoScalingConfiguration = new AutoScalingConfiguration(stack, 'AutoScalingConfiguration', {
    autoScalingConfigurationName: 'my-autoscaling-config',
    maxConcurrency: 150,
    maxSize: 20,
    minSize: 5,
  });

  new cdk.Tags.of(autoScalingConfiguration).add('Environment', 'production');

  // THEN
  Template.fromStack(stack).hasResourceProperties('AWS::AppRunner::AutoScalingConfiguration', {
    Tags: [
      {
        Key: 'Environment',
        Value: 'production',
      },
    ],
  });
});
```

Let's test tags as well?

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
mazyu36: Thanks. I like your suggested your approach.
I've create a private method including autoScalingConfigurationName validation.

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
```diff
@@ -0,0 +1,196 @@
...
+    * The name of the Auto Scaling Configuration.
+    * @attribute
+    */
+  readonly autoScalingConfigurationName: string;
+
+  /**
+   * The revision of the Auto Scaling Configuration.
+   * @attribute
+   */
+  readonly autoScalingConfigurationRevision: number;
```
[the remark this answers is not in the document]
mazyu36: Thank you. Personally, I'm in favor of making the revision a string. 
However, I have some concerns.

The revision in the Cfn (L1 construct) is a number, but would it be appropriate to change it to a string in the L2 construct? I'm not sure if there's a possibility of it changing, so if you know anything about it, please let me know.

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apprunner.CfnAutoScalingConfiguration.html#attrautoscalingconfigurationrevision

Additionally, since the already implemented VpcConnector treats the revision as a number, I'm also concerned about having a different approach from that.

https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-apprunner-alpha.VpcConnector.html#vpcconnectorrevisionspan-classapi-icon-api-icon-experimental-titlethis-api-element-is-experimental-it-may-change-without-noticespan

mazyu36:
@pahud 
Thank you.
I have replied to the comments and refactored the code.​​​​​​​​​​​​​​​​

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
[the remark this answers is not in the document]
pahud: Thank you for the callout.

I checked the cfnspec and yes the type is Integer, which is number in TS.

```
"Attributes": {
        "AutoScalingConfigurationRevision": {
          "PrimitiveType": "Integer"
        },
```

I am fine having it as `number` for consistency here but I kind of feel if apprunner supports `latest` revision someday, which would be a breaking change to CFN then CDK would have breaking change as well.

@GavinZZ @paulhcsun  what do you think?

pahud:
LGTM now! I'll let other maintainers to take a final review.

packages/@aws-cdk/aws-apprunner-alpha/lib/auto-scaling-configuration.ts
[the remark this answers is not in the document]
GavinZZ: Discussed with team internally. I think we don't need to over-optimize the type here. CFN schema type changes are not supposed to happen especially after it's already in CloudFormation resource schema registry.

GavinZZ:
Thanks Pahud for the thorough reviews and thanks @mazyu36 for the PR and quick responses.

GavinZZ (approved):
LGTM thanks for contributing

mazyu36:
@Mergifyio update

mazyu36:
@Mergifyio refresh

PR #30371, opened 2024-05-29
feat(cli): add ability to configure hotswap properties for ECS

atanaspam:
### Reason for this change

We aim to speed up deployment times in our development environment by using the hotswap feature. However, our services have dependencies on each other, and the current hotswap behavior is too disruptive. 

### Description of changes

We modified the ECS service hotswap implementation to accept minimumHealthyPercent and maximumHealthyPercent as configurable parameters, with default values of 0 and 200, respectively. These parameters are exposed to the cli and can be passed as `--hotswap-ecs-minimum-healthy-percent <number>` and `--hotswap-ecs-maximum-healthy-percent <number>`

We added a unit test validating that the correct values are passed to the task definition. We also executed using the locally built version of cdk validating that the behaviour is as expected: the parameters are respected during hotswap deployments, and the existing API is maintained.

### Checklist

----


atanaspam:
Superseded by https://github.com/aws/aws-cdk/pull/30511

PR #30374, opened 2024-05-29
fix(cli): error message for destroy action on non-existent stack

### LINKED ISSUE #22240: cdk: Destroy reports success when there is nothing to destroy
### Describe the bug

Configure AWS profile to an account that has not been bootstrapped with cdk.
Run cdk destroy - cdk reports success.

### Expected Behavior

cdk should report an error indicating that the selected stack does not exist.

### Current Behavior

cdk incorrectly reports success in destroying the stack.

### Reproduction Steps

Create a stack, switch accounts. execute: cdk destroy

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.43.1 (build c1ebb85)

### Framework Version

_No response_

### Node.js Version

v16.16.0

### OS

MacOS Monterey

### Language

Typescript

### Language Version

Actually, javascript but drop down does not offer that option

### Other information

_No response_

slxsh:
Closes #22240

### Reason for this change

- Destroy command reports success even if a stack doesn't exist.

### Description of changes

While running destroy command if a stack is not deployed then an error message will be displayed but the operation won't be interrupted. 
- In deploy-stack.ts added an exception if the stack doesn't exist.
- In cdk-toolkit.ts added error message if the above exception is thrown.

I need some help with testing. I couldn't find a way to test the changes using real stacks/cli. Also, not sure if I need to add integration test too (didn't see any existing relevant test where I could make some changes).

- Have added unit tests.

### Checklist

----


slxsh:
Clarification Request - regarding integration tests. Also, change in CLI code.

packages/aws-cdk/lib/api/deploy-stack.ts
```diff
@@ -578,7 +578,7 @@ export async function destroyStack(options: DestroyStackOptions) {
 
   const currentStack = await CloudFormationStack.lookup(cfn, deployName);
   if (!currentStack.exists) {
-    return;
+    throw new StackNotFoundError(`Failed to destroy ${deployName}. The stack is not deployed.`);
```
comcalvi, line 581: This is a breaking change, because we are changing the exit code for this command. We will not accept breaking changes to any CLI commands. Instead please a new flag, something `--fail-on-missing`, to `destroy`, and in this `if` ensure we check both `currentStack.exists` and `options.failOnMissing`, or a different name if you think a different one would be better.

packages/aws-cdk/lib/api/deploy-stack.ts
```diff
@@ -694,3 +694,9 @@ function suffixWithErrors(msg: string, errors?: string[]) {
     ? `${msg}: ${errors.join(', ')}`
     : msg;
 }
+
+export class StackNotFoundError extends Error {
+  constructor(message: string) {
+    super(message);
+  }
+}
```
comcalvi, line 702: We don't need this error type here, please use `throw new Error('message');`

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -620,8 +620,12 @@ export class CdkToolkit {
         });
         success(`\n ✅  %s: ${action}ed`, chalk.blue(stack.displayName));
       } catch (e) {
-        error(`\n ❌  %s: ${action} failed`, chalk.blue(stack.displayName), e);
-        throw e;
+        if (e instanceof StackNotFoundError) {
```
comcalvi, line 623: Don't use `instanceof`, instead check with `e.message.includes()`.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -1525,4 +1529,4 @@ function buildParameterMap(parameters: {
   }
 
   return parameterMap;
-}
+}
```
comcalvi, line 1532: Please keep the newline at the end of the file.

comcalvi (changes requested):
You will need a CLI integration test for this feature. The CLI integration tests all live here: https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/cli-integ. The readme provides guidance in authoring and running them locally.

packages/aws-cdk/lib/api/deploy-stack.ts
slxsh: Would this be considered a breaking change even though the [exception is being caught](https://github.com/aws/aws-cdk/pull/30374/files/bc1db8f5b91f06d876110f3afcc83b3f806bb5da#diff-8cff125f4b53b025add8740ab74f39ee6c20221c0d3ac5ab7db184a6bb077398R623-R625)? The exit code of destroy command is not changed. The command doesn't fail if the stack doesn't exist.

The issue states `destroy` returns success even if the stack doesn't exist, if we add the flag this problem will persist with the `destroy` command.

packages/aws-cdk/lib/api/deploy-stack.ts
slxsh: Sure. Any reason why you feel this is unnecessary? Saw some custom errors in the code. Thought it would be make it clear which exception is being thrown.

PR #30389, opened 2024-05-30
fix(stepfunctions-tasks): run task permission is too broad (under feature flag)

### LINKED ISSUE #30368: (stepfunctions-tasks): Security action for EcsRunTask
### Describe the feature

The following message has been received by AWS users and needs to be addressed to be compliant in granting permissions when using EcsRunTask.

```
We are contacting you due to an issue with the Amazon Elastic Container Service (Amazon ECS) API that requires action: RunTask , StartTask, CreateService, and CreateTaskSet, When calling the RunTask, StartTask, CreateService, CreateTaskSet, or UpdateService APIs, users can specify the revision number of a task definition to launch a specific version of that task. We have identified an inconsistency in the way Identity and Access Management (IAM) policies are applied during request authentication for the above APIs. Specifically, when the API is invoked without the revision number of the task definition, it could be interpreted differently if the revision number of the task definition family is not specified in the resource condition key. As a result, the latest version of the task was selected. We have implemented a fix and have verified that the service is working as expected.
We have verified that your account sent a request to one or more of the affected ECS APIs. We recommend that you review the policies listed in the “Affected Resources” tab to ensure that the resource condition key specifying the task definition family contains the revision number. To allow time to review and make any necessary changes, we are adding your account to the permit list until October 15, 2024; if you wish to remove your account from the permit list prior to October 15, 2024, please create an AWS Support Case [see below]. If you do not perform the action by that date, any calls to the affected APIs will return an AccessDeniedException error message.
After October 15, 2024, to properly apply IAM policy-based decisions, you must specify the revision number or task definition family wildcard ('*') when using a task definition ARN as the resource type.
```

### Use Case

When executing ECS task from StepFunctions using EcsRunTask in stepfunctions-tasks.

### Proposed Solution

Remove the Ecs Task Family from the IAMResource permission policy.

https://github.com/aws/aws-cdk/blob/c8f2646fc903114c7819e4167b8f38436af91ee1/packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts#L355

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": "ecs:RunTask",
      "Resource": [
          "arn:aws:ecs:*:*:task-definition/sleep360",   // here
          "arn:aws:ecs:*:*:task-definition/sleep360:*",
       ]
    }
  ]
}
```

### Other Information

related issue #12080 #27891

### Acknowledgements

### CDK version used

2.143.0

macOS,Linux,Windows

msambol:
Closes #30368. 

Doc: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/security_iam_id-based-policy-examples.html#IAM_run_policies

### DIFF
--- a/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts
+++ b/packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/shared/base-load-balancer.ts
@@ -54,7 +54,7 @@ export interface BaseLoadBalancerProps {
    *
    * @default - false for Network Load Balancers and true for Application Load Balancers.
    * This can not be `false` for Application Load Balancers.
-   * @see - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattribute.html
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattribute.html
    */
   readonly crossZoneEnabled?: boolean;
 
--- a/packages/aws-cdk-lib/aws-rds/lib/instance-engine.ts
+++ b/packages/aws-cdk-lib/aws-rds/lib/instance-engine.ts
@@ -882,6 +882,8 @@ export class MysqlEngineVersion {
   public static readonly VER_8_0_35 = MysqlEngineVersion.of('8.0.35', '8.0');
   /** Version "8.0.36". */
   public static readonly VER_8_0_36 = MysqlEngineVersion.of('8.0.36', '8.0');
+  /** Version "8.0.37". */
+  public static readonly VER_8_0_37 = MysqlEngineVersion.of('8.0.37', '8.0');
 
   /**
    * Create a new MysqlEngineVersion with an arbitrary version.


msambol:
Exemption Request: I updated integration snapshots, but a change in the integration test file isn't needed.

GavinZZ (commented):
I agree that we were granting too much permission in the current solution based on the documentation you provided. However, I'm a bit worried that this would introduce breaking changes for existing users (who incorrectly rely on the broad permissions) and accepting this would break their application.

lpizzinidev (commented):
@GavinZZ https://github.com/aws/aws-cdk/pull/30389#pullrequestreview-2100172756
Feature flag?

GavinZZ:
> @GavinZZ [#30389 (review)](https://github.com/aws/aws-cdk/pull/30389#pullrequestreview-2100172756) Feature flag?

Yes, thanks for leaving a comment. I think feature flag is needed in this case to guarantee no breaking change.

msambol:
> > @GavinZZ [#30389 (review)](https://github.com/aws/aws-cdk/pull/30389#pullrequestreview-2100172756) Feature flag?
> 
> Yes, thanks for leaving a comment. I think feature flag is needed in this case to guarantee no breaking change. IMO I think it's likely that only few users would experience any breaking changes from this change, but I don't have the data on me and it's our promise that stable module don't bring breaking changes.

I don't even think this was working previously. See my PR description here: https://github.com/aws/aws-cdk/pull/27891.

GavinZZ:
@msambol Sorry for the delayed response. A bit confused here. It seems there was an issue with too little permission and you've fixed it half a year ago. So I'm assuming this is working now except we've giving too much permission in the resource field, meaning that this is currently working right and it may be a breaking change?

msambol:
> @msambol Sorry for the delayed response. A bit confused here. It seems there was an issue with too little permission and you've fixed it half a year ago. So I'm assuming this is working now except we've giving too much permission in the resource field, meaning that this is currently working right and it may be a breaking change?

This https://github.com/aws/aws-cdk/pull/27891/files#diff-9d5f521b191b7a78f695205cd6cdbf39fce4ffcae0de03b4c4a92d3c59c06f3cL343 permission was initially there, but it didn't actually work or allow the integration tests to run properly. Adding https://github.com/aws/aws-cdk/pull/27891/files#diff-9d5f521b191b7a78f695205cd6cdbf39fce4ffcae0de03b4c4a92d3c59c06f3cR346 permission fixed it. I should have removed the initial condition when I added the other one. But if you feel we should do a feature flag just in case, we can add that.

GavinZZ:
I would like to ask to add a feature flag just in case. Reason being is that my understanding is that currently we allow `runTask` on both the task itself `taskDefinitionFamilyArn` and all revisions of the task `taskDefinitionFamilyArn/*`. I'm worried that this could be a breaking change. Feel free to correct me if my understanding is incorrect.

msambol:
@GavinZZ Do you have an example on where we use a FF like this?

GavinZZ:
This one seems to be similar https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/pipelines/lib/codepipeline/codepipeline.ts#L987

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/ecs/run-tasks-feature-flag.test.ts
```diff
@@ -0,0 +1,488 @@
...
+import * as sfn from '../../../aws-stepfunctions';
+import { Stack } from '../../../core';
+import { ECS_REDUCE_RUN_TASK_PERMISSIONS } from '../../../cx-api';
+import * as tasks from '../../lib';
+
+let stack: Stack;
+let vpc: ec2.Vpc;
+let cluster: ecs.Cluster;
+
+/* eslint-disable quote-props */
```
msambol, line 14: given how `run-tasks.test.ts` was implemented with the `beforeEach`, I think it makes more sense to test these in a new file.

msambol:
@GavinZZ Let me know what you think of this. Default is the extra permission gets added.

GavinZZ:
@msambol Looks good, please also add an entry to README [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/cx-api/README.md)

msambol:
> @msambol Looks good, please also add an entry to README [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/cx-api/README.md)

Done!

GavinZZ (approved):

GavinZZ (approved):

Issue #15523, opened 2021-07-13
(aws-cloudfront): Unexpected diffs caused by cloudfront.Function


iliapolo:
Non deterministic auto-generated function name causing unexpected diffs.

### Reproduction Steps

Using this code:

```ts
new cloudfront.Function(stack, 'Function', {
  code: cloudfront.FunctionCode.fromInline('function handler(event) { return event.request }'),
});
```

1. synth and extract the function name
2. synth again and extract the function name

The function name is different for every synth.

### What did you expect to happen?

The function name should be the same.

### What actually happened?

Function name changes per synth.

### Environment

  - **CDK CLI Version  : ALL**
  - **Framework Version: 1.109.0**
  - **Node.js Version: ALL** 
  - **OS               : ALL**
  - **Language (Version): ALL** 

### Other

This happens because of:

https://github.com/aws/aws-cdk/blob/94f81c441f9e2cb8dc70eb2e772d2cd75e468b67/packages/%40aws-cdk/aws-cloudfront/lib/function.ts#L178-L185

Where `Stack.of(this).region` is an unresolved token that can have a different sequence number every time the program is executed.

A workaround right now would be to provide a name explicitly and not rely on this logic.

### Workaround

Provide an explicit name for `cloudfront.Function`. Using `this.node.addr` or `Node.of(this).addr` will return a stable fixed length unique id. Example:

```ts
const functionId = `MyFunction${this.node.addr}`;

new cloudfront.Function(stack, functionId, {
  code: cloudfront.FunctionCode.fromInline('function handler(event) { return event.request }'),
  functionName: functionId
});
```

--- 

This is :bug: Bug Report

eladb:
I would recommend using `Node.of(this).addr` which returns a stable fixed length output

hoegertn:
@njlynch 's wish was to add the region information as CF is a global service. But if this is no longer needed happy to remove it. This would be a breaking change for the implementation afaik.

eladb:
> @njlynch 's wish was to add the region information as CF is a global service. But if this is no longer needed happy to remove it. This would be a breaking change for the implementation afaik.

What kind of breaking change?

hoegertn:
If we change the generation of the name it will recreate the CF functions on the next deployment. The name is the physicalName of the function.

eladb:
Will it cause outage or availability risk?

hoegertn:
[...]

PR #30392, opened 2024-05-30
fix(cloudfront): make long function name deterministic

### LINKED ISSUE #20017: (cloudfront): (cloudfront function generated name is not deterministic and will change between synthesis)
### Describe the bug

When using `cloudfront.Function` without specifying the function name, the construct `cloudfront.Function` will generate a name automatically.  However this generated function name will change from time to time between synthesis due to the name truncated to a fixed length with a region token prefix in it. This change of name caused deployment failure when deploying onto an existing stack. 

For example, with below code in a stack named `CdkIssueWithALongNameStack`
```
    new cloudfront.Function(this, "MyCloudFrontFunction", {
      code: cloudfront.FunctionCode.fromInline(""),
    });
```

Running `cdk synth` multiple times, It can be observed that during some runs, in the synthesised CloudFormation template, we can find below. Please note the function name is  `CdkIssueWtackMyCloudFrontFunction6157953B`.

```
  "MyCloudFrontFunction18D9B9FD": {
   "Type": "AWS::CloudFront::Function",
   "Properties": {
    "Name": {
     "Fn::Join": [
      "",
      [
       {
        "Ref": "AWS::Region"
       },
       "CdkIssueWtackMyCloudFrontFunction6157953B"
      ]
     ]
    },
```
However, some other runs, the name will change to `CdkIssueWitackMyCloudFrontFunction6157953B`. Note the additional `i` after `CdkIssueW`. 

This name change of the cloudfront function will cause failure of deployment on an existing stack as CDK will report error like:  
```
Resource of type ‘AWS::CloudFront::Function’ with identifier ‘CdkIssueWtackMyCloudFrontFunction6157953B’ was not found.
```

### Expected Behavior

The automatically generated function name should remain the same during multiple cdk synth.

### Current Behavior

The name changes due to the following reason:

The following code snippet is how `cloudfront.Function` generate its function name if the name is not specified in the props. 

```
  private generateName(): string {
    const name = Stack.of(this).region + Names.uniqueId(this);
    if (name.length > 64) {
      return name.substring(0, 32) + name.substring(name.length - 32);
    }
    return name;
  }
}
```
https://github.com/aws/aws-cdk/blob/9487b39d6da4ab7efd884c3a8379b03e317786b9/packages/%40aws-cdk/aws-cloudfront/lib/function.ts#L178

it added `Stack.of(this).region` in front of the unique id of this construct and then took the first 32 chars and then last 32 chars if the string is longer than 64 chars, and concatenated these two parts together to form the name of the function. However, the region added in front of the name is still a token at this time, so the name string became: 
 
`${Token[AWS.Region.7]}CdkIssueWithALongNameStackMyCloudFrontFunction6157953B`

This string is then sent through `name.substring(0, 32) + name.substring(name.length - 32);` and resulted in two parts:  

`${Token[AWS.Region.7]}CdkIssueWi` and `tackMyCloudFrontFunction6157953B` and then they are concatenated together to make the name  `${Token[AWS.Region.7]}CdkIssueWitackMyCloudFrontFunction6157953B`, which eventually became the name in the CloudFormation template.

However, the region token string could be different every time cdk synth runs, sometimes, the region token will have a two digits value like 11 instead of 7, then the name above will be like:

`${Token[AWS.Region.11]}CdkIssueWithALongNameStackMyCloudFrontFunction6157953B`, which run through `name.substring(0, 32) + name.substring(name.length - 32);` and resulted in two parts: 

`${Token[AWS.Region.11]}CdkIssueW` and `tackMyCloudFrontFunction6157953B` and then they are concatenated together make the name `${Token[AWS.Region.11]}CdkIssueWtackMyCloudFrontFunction6157953B`.  This name is different from the name when the number in region token is one digit, and missed that `i`. 

Since the number in the region token is unpredictable so after a few rounds of cdk synth command, the name will change between those two results.

### Reproduction Steps

1. Create a cdk sample app  with command `cdk init sample-app --language=typescript`.
2. Add below into `lib/<app name>-stack.ts` 
```
import * as cloudfront from "aws-cdk-lib/aws-cloudfront";

...
// in the constructor, add below at the end

    new cloudfront.Function(this, "MyCloudFrontFunction", {
      code: cloudfront.FunctionCode.fromInline(""),
    });
```

3. add below into `bin/<app name>.ts`. Make sure the stack name is long enough
```
#!/usr/bin/env node
import { Construct } from "constructs";
import * as cloudfront from "aws-cdk-lib/aws-cloudfront";
import { Aspects } from "aws-cdk-lib";

class TestAspect implements cdk.IAspect {
  public visit(construct: Construct) {
    if (construct instanceof cloudfront.CfnFunction) {
      console.log(construct.name);
    }
  }
}

const app = new cdk.App();
Aspects.of(app).add(new TestAspect());
new CdkissueStack(app, "CdkIssueWithALongNameStack");
```
4. build with `npm run build` and run cdk synth with `npx cdk synth -q`, the following content will be printed out during the `cdk synth`
```
$ npx cdk synth -q
${Token[AWS.Region.10]}CdkIssueWtackMyCloudFrontFunction6157953B
```

5. when seeing this, check the generated CloudFormation Template in `cdk.out` and search for `AWS::CloudFront::Function` resource and there should be something like below:  

```
  "MyCloudFrontFunction18D9B9FD": {
   "Type": "AWS::CloudFront::Function",
   "Properties": {
    "Name": {
     "Fn::Join": [
      "",
      [
       {
        "Ref": "AWS::Region"
       },
       "CdkIssueWtackMyCloudFrontFunction6157953B"
      ]
     ]
    },
```
Write down the name

6. run `npx cdk synth -q` multiple times until you can see the number in the region token is one digit
```
$ npx cdk synth -q
${Token[AWS.Region.8]}CdkIssueWitackMyCloudFrontFunction6157953B
```
then check the generated CloudFormation template in `cdk.out`, and search for `AWS::CloudFront::Function` resource and you can see the name of the resource became `CdkIssueWitackMyCloudFrontFunction6157953B` with an additional `i`.

### Possible Solution

The quoted code snippet below are from  `cloudfront.Function` construct, instead of adding the region token into the name string first, it could perform the name truncating and concatenation without adding the region token in front of it, and then add the region as prefix after to make the new name. So the concatenated name will be stable and won't change due to the different length of the region token (the digits of the number in it) 

Below is one possible fix and it might need further validation:

```
  private generateName(): string {
    // currently the longest region name is ap-southeast-2 which is 14 chars
    const maxLengthOfRegionName = 14;
    // CloudFront Function has maximum 64 chars limit on function name
    const maxLengthOfName = 64 -  maxLengthOfRegionName;
    const longName = Names.uniqueId(this);
    if (longName.length > maxLengthOfName)) {
      const partLength = Math.floor(maxLengthOfName/2)
      return Stack.of(this).region  + longName.substring(0, partLength) + longName.substring(name.length - partLength);
    }
    return Stack.of(this).region  + longName;
  }
```

Current code in `cloudfront.Function`:
```
  private generateName(): string {
    const name = Stack.of(this).region + Names.uniqueId(this);
    if (name.length > 64) {
      return name.substring(0, 32) + name.substring(name.length - 32);
    }
    return name;
  }
}
```
https://github.com/aws/aws-cdk/blob/9487b39d6da4ab7efd884c3a8379b03e317786b9/packages/%40aws-cdk/aws-cloudfront/lib/function.ts#L178

_No response_

### CDK CLI Version

2.20.0 (build 738ef49)

### Framework Version

_No response_

### Node.js Version

v14.18.2

### OS

macOS Monterey 12.2.1

### Language

Typescript

### Language Version

Typescript(3.9.7)

### Other information

_No response_

Leo10Gama:
### Issue # (if applicable)

Closes #20017 as well as #15523 and #28629 

### Reason for this change

Due to the way function names are generated using token strings with either single- or double-digit numbers, longer function names can be truncated differently, leading to inconsistency in generated CloudFormation templates.

### Description of changes

To ensure backwards compatibility, if names are longer than 64 characters and use region tokens, if the token uses a single-digit region number, it takes the first **31** characters + the last 32 characters; if the token uses a double-digit region number or otherwise, it takes the first **32** characters + the last 32 characters. This ensures it will always take the same first chunk of the actual function's name.

A new unit test was added to verify the consistency of function names in the template.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-codebuild/README.md
+++ b/packages/aws-cdk-lib/aws-codebuild/README.md
@@ -69,6 +69,9 @@ const gitHubSource = codebuild.Source.gitHub({
       .inEventOf(codebuild.EventAction.PUSH)
       .andBranchIs('main')
       .andCommitMessageIs('the commit message'),
+    codebuild.FilterGroup
+      .inEventOf(codebuild.EventAction.RELEASED)
+      .andBranchIs('main')
   ], // optional, by default all pushes and Pull Requests will trigger a build
 });
 ```
--- a/packages/aws-cdk-lib/aws-codebuild/lib/source.ts
+++ b/packages/aws-cdk-lib/aws-codebuild/lib/source.ts
@@ -179,6 +179,11 @@ export enum EventAction {
   PULL_REQUEST_UPDATED = 'PULL_REQUEST_UPDATED',
 
   /**
+   * Closing a Pull Request.
+   */
+  PULL_REQUEST_CLOSED = 'PULL_REQUEST_CLOSED',
+
+  /**
    * Merging a Pull Request.
    */
   PULL_REQUEST_MERGED = 'PULL_REQUEST_MERGED',
@@ -188,6 +193,24 @@ export enum EventAction {
    * Note that this event is only supported for GitHub and GitHubEnterprise sources.
    */
   PULL_REQUEST_REOPENED = 'PULL_REQUEST_REOPENED',
+
+  /**
+   * A release is created in the repository.
+   * Works with GitHub only.
+   */
+  RELEASED = 'RELEASED',
+
+  /**
+   * A prerelease is created in the repository.
+   * Works with GitHub only.
+   */
+  PRERELEASED = 'PRERELEASED',
+
+  /**
+   * A workflow job is queued in the repository.
+   * Works with GitHub only.
+   */
+  WORKFLOW_JOB_QUEUED = 'WORKFLOW_JOB_QUEUED',
 }
 
 enum WebhookFilterTypes {
--- a/packages/aws-cdk-lib/aws-codebuild/test/codebuild.test.ts
+++ b/packages/aws-cdk-lib/aws-codebuild/test/codebuild.test.ts
@@ -567,6 +567,10 @@ describe('default properties', () => {
         webhookFilters: [
           codebuild.FilterGroup.inEventOf(codebuild.EventAction.PUSH).andTagIsNot('stable'),
           codebuild.FilterGroup.inEventOf(codebuild.EventAction.PULL_REQUEST_REOPENED).andBaseBranchIs('main'),
+          codebuild.FilterGroup.inEventOf(codebuild.EventAction.RELEASED).andBaseBranchIs('main'),
+          codebuild.FilterGroup.inEventOf(codebuild.EventAction.PRERELEASED).andBaseBranchIs('main'),
+          codebuild.FilterGroup.inEventOf(codebuild.EventAction.WORKFLOW_JOB_QUEUED).andBaseBranchIs('main'),
+          codebuild.FilterGroup.inEventOf(codebuild.EventAction.PULL_REQUEST_CLOSED).andBaseBranchIs('main'),
         ],
       }),
     });
@@ -593,6 +597,22 @@ describe('default properties', () => {
           ],
           [
             { Type: 'EVENT', Pattern: 'PULL_REQUEST_REOPENED' },
+            { Type: 'BASE_REF', Pattern: 'refs/heads/main' },
+          ],
+          [
+            { Type: 'EVENT', Pattern: 'RELEASED' },
+            { Type: 'BASE_REF', Pattern: 'refs/heads/main' },
+          ],
+          [
+            { Type: 'EVENT', Pattern: 'PRERELEASED' },
+            { Type: 'BASE_REF', Pattern: 'refs/heads/main' },
+          ],
+          [
+            { Type: 'EVENT', Pattern: 'WORKFLOW_JOB_QUEUED' },
+            { Type: 'BASE_REF', Pattern: 'refs/heads/main' },
+          ],
+          [
+            { Type: 'EVENT', Pattern: 'PULL_REQUEST_CLOSED' },
             { Type: 'BASE_REF', Pattern: 'refs/heads/main' },
           ],
         ],
@@ -664,6 +684,7 @@ describe('default properties', () => {
             codebuild.EventAction.PULL_REQUEST_CREATED,
             codebuild.EventAction.PULL_REQUEST_UPDATED,
             codebuild.EventAction.PULL_REQUEST_MERGED,
+            codebuild.EventAction.PULL_REQUEST_CLOSED,
           ).andTagIs('v.*'),
           // duplicate event actions are fine
           codebuild.FilterGroup.inEventOf(codebuild.EventAction.PUSH, codebuild.EventAction.PUSH).andActorAccountIsNot('aws-cdk-dev'),
@@ -685,7 +706,7 @@ describe('default properties', () => {
         Webhook: true,
         FilterGroups: [
           [
-            { Type: 'EVENT', Pattern: 'PULL_REQUEST_CREATED, PULL_REQUEST_UPDATED, PULL_REQUEST_MERGED' },
+            { Type: 'EVENT', Pattern: 'PULL_REQUEST_CREATED, PULL_REQUEST_UPDATED, PULL_REQUEST_MERGED, PULL_REQUEST_CLOSED' },
             { Type: 'HEAD_REF', Pattern: 'refs/tags/v.*' },
           ],
           [


scanlonp (changes requested):
Hey @Leo10Gama, I like the nature of the changes. A couple things to think about.

First, this looks like it may be a breaking change. From the comments in #15523 and #28641, changing this name will replace the CloudFront function. We may need to put this behind a feature flag.

Second, I think we can solve the consistency problems as well keeping the region in the name by adding the region to the name separately from the truncation. Say we do
```
const uniqueName = Lazy.string({
        produce: () => Names.uniqueResourceName(this, { maxLength: 40, allowedSpecialCharacters: '-_' }),
return Stack.of(this).region + uniqueName;
```
If the current longest region name is 14 characters (ap-southeast-1), we need to make sure the uniqueName is max 50 characters. Why not bake in some wiggleroom for future region names an make the max length 40.

#28641 used `Names.uniqueResourceName` in order to control the length, but truncating `Names.uniqueId` is also an option.

Leo10Gama:
Hey @scanlonp, thanks for the feedback!

With regards to whether this would be a breaking change, I tried to find a solution that would only affect users facing this issue, hence why I wrote the if-else logic the way I did. My idea was to preserve the original logic, but rework it so that the newer logic only happens for affected users, hence the `.startsWith()` check at the beginning. In any other situation, the original name generation is preserved, so users who weren't affected won't see a difference, thus I don't see the need for using a flag.

The `Lazy.string()` approach seems like it would also work, since it would replace the name with a token like the region, but that would definitely introduce breaking changes and would need to go behind a feature flag as you said. Since this is a bug that's been affecting users for a while (given a few issues opened about it), I think it would be good to try prioritizing a fix that isn't hidden behind a flag, for ease of use for users. 

I wouldn't be opposed to making the lazy string change if the flag is necessary, since it does seem like a better way to generate names as well compared to the existing logic! But if we can find a way to maintain the original logic while only fixing the issue for affected users, we ought to do that. If breaking changes are unavoidable, then the flag and lazy string approach would definitely be the way to go, but I'd be open to trying to rework this logic if you know of any cases this would break.

If we go with the current implementation, though, I can also replace the magic number indicating how many digits are in the region token with a computed value for clarity lol

scanlonp:
@Leo10Gama, I certainly agree with preserving the logic for users who do not run into this bug.

To what extent we can "break" customers depends on who is facing this bug right now. The non-deterministic behavior occurs when the function name is sufficiently long. Since we have non-deterministic behavior, we are possibly forcing replacements on every deploy, so it should be ok to just make the change without a feature flag in those cases. For customers whose function name is not that long, they have never faced the issue. I appreciate your concern for not changing the behavior for this second group of users.

With that in mind, I did some testing, and found that if the user's stack and function id's are 34 or more characters combined, we see the non-determinism. 
It appears that `Names.uniqueResourceName()` and `Names.uniqueId()` essentially have the same output - `uniqueResourceName()` just allows some configuration, such as the max length. Therefore, for users whose stack and function id's are 33 or less characters combined, `Names.uniqueResourceName()`with max characters >33 will generate the same name as the old composition, so those users will see no difference. (`Lazy.string()` also resolves the token at synth time, so there is not a concern there about introducing more tokens to the template or breaking users in that way.)

For users with stack + FunctionName >= 34, they are already broken, so the new generation will make their names standard. I also like using a function specialized for accommodating resource names with a max length rather than hacking it together with logic in the construct itself.

This is to say I believe we are free to change the name generation in this way without a feature flag at all!

scanlonp (changes requested):
Do we need to check the length at all? I think we can just generate the name consistently from now on. I believe that the behavior will not change for users whose function name was <64 characters.

Leo10Gama:
Yes, behaviour should be consistent now thanks to doing the name-splitting before appending the region. As you said, the current longest region name is 14 characters, so leaving the name at 40 shouldn't cause any issues, but I could add a quick check before the new logic's `return` to verify it's <64 characters, if that's preferred.

scanlonp:
I meant that we would not even have to create the initial name and verify if it is less than 64 characters at all. We could simply use the new logic every time.

I am trying to test if this would make any difference. So far, it has not, but I need to look at the impact for when the Function is nested in the construct tree - defined within another construct or stack.

packages/aws-cdk-lib/aws-cloudfront/lib/function.ts
```diff
@@ -222,11 +222,13 @@ export class Function extends Resource implements IFunction {
   }
 
   private generateName(): string {
-    const name = Stack.of(this).region + Names.uniqueId(this);
-    if (name.length > 64) {
-      return name.substring(0, 32) + name.substring(name.length - 32);
+    const maxNameLength = 64 - '${Token[AWS.Region.00]}'.length;
```
scanlonp, line 225: nit: this variable is not really the max name length
nit2: lets add a doc string here to explain what & why we are doing this. It is not clear since this was an edge case we are trying to avoid.

packages/aws-cdk-lib/aws-cloudfront/lib/function.ts
```diff
@@ -222,11 +222,13 @@ export class Function extends Resource implements IFunction {
...
+    if (Names.uniqueId(this).length < maxNameLength) {
```
scanlonp, line 226:
```suggestion
    if (Names.uniqueId(this).length <= maxNameLength) {
```
Off by one I believe. Users are fine if their name came out to be this length as the non determinism comes when the token length plus the name is greater than 64.

scanlonp (changes requested):
One small change, and a couple small nits. Otherwise, we are good to go.

Also lets untrack the Feature Flags file. Even if that is what needs to be in there, this PR is not the best place for it.

scanlonp (approved):
Nice work @Leo10Gama, looks great!

Pretty funny that the two integ tests straddle the character limit exactly.

Issue #23909, opened 2023-01-30
core: add global `logRetentionDays` setting


fargito:
### Describe the feature

add a global `logRetentionDays` context setting in `cdk.json` and make it the default for all log retention settings

### Use Case

Hello,

Setting the retention period on all resources is quite frustrating.

Moreover, the value is often identical between all resources in the same stack, so it would be awesome to be able to define it in a centralized place.

Also, since the default is `Infinity`, having this setting would potentially have a big impact to help reduce the number of useless logs.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.62.2

### Environment details (OS name and version, etc.)

Ubuntu 22.04

peterwoodworth:
Have you looked into aspects? [Aspects](https://docs.aws.amazon.com/cdk/v2/guide/aspects.html) are a generic way you could implement anything like this in your own CDK app

I'm not sure that this particular request would be easy to implement however, some of our log retention implementations use custom resources. Additionally, I'm not sure we'd want to implement this feature due to the precedent it could set and the maintenance overhead it would create. I can leave this open to discuss workarounds, and to gather community feedback on if this is something we should do

fargito:
Thanks for your response! I understand this constraint, I will check if aspects can help with my use case.

anentropic:
I have a similar requirement, but more a frustration with resources which are apparently out of my control which create logs with "never expire" retention policy.

For example I find there are a bunch of "never expire" log groups like:
```
/aws/lambda/MyStack-CustomVpcRestrictDefaultSGCustomRes-KeArPuyYcxtf
/aws/lambda/MyStack-CustomVpcRestrictDefaultSGCustomRes-VeSnkFXYTboC
...
```
...presumably one per deployment, as mentioned in this related issue https://github.com/aws/aws-cdk/issues/13250

`CustomVpcRestrictDefaultSGCustomRes` is generated by something in CDK under the hood. I can see this resource in the generated template.json after synth, but have been unable to find any way to target it via Aspects.

Also note how the log group name changes each time, making it impossible(?) to target with an override via `LogRotation`  construct as some people are doing in the issue above.

e.g. I made this aspect:
```python
import jsii
from aws_cdk import Annotations, IAspect
from constructs import IConstruct, MetadataEntry

def _get_logical_id(metadata: list[MetadataEntry]) -> str | None:
    for entry in metadata:
[...]

PR #30394, opened 2024-05-31
feat(s3): allow user to set Log Group on S3 Bucket autoDeleteObjects provider lambda

### LINKED ISSUE #24815: (aws-s3): retention on S3 autoDeleteObjects lambda CloudWatch log group is Never expire

samson-keung:
### Issue # (if applicable)

Closes #24815.

### Reason for this change

To allow log group customization on the custom resource lambda created for the `autoDeleteObjects` feature.

### Description of changes

At the highest level overview, a static method `setAutoDeleteObjectsLogGroup` is added to the `Bucket` class. When it is called, it will set the log group on the `AutoDeleteObjectsProvider` lambda (i.e. setting the [`LoggingConfig.LogGroup`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-loggingconfig.html#cfn-lambda-function-loggingconfig-loggroup).

In order to support the above change, 2 underlying changes had to be made:
1. `setAutoDeleteObjectsLogGroup(..)` needs to have a way to find the singleton `AutoDeleteObjectsProvider` lambda. This means a method needs to be added in the `AutoDeleteObjectsProvider` class that returns the singleton. Note that the `AutoDeleteObjectsProvider` class itself is code generated. So I have modified the code gen logic to generate the `getProvider(..)` method, which returns the singleton.
2. With a handle of the singleton of type `AutoDeleteObjectsProvider`, which wraps the actual `AWS::Lambda::Function`, we need a way to set the log group on the lambda. With `AutoDeleteObjectsProvider` extending the `CustomResourceProviderBase` type, a method is added to `CustomResourceProviderBase` class to set the log group.

Updated the integ test and ran it against my own AWS account

### Checklist

----


packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1882,6 +1886,26 @@ export class Bucket extends BucketBase {
...
+   * @param scope the stack with bucket(s) with `autoDeleteObjects: true`.
+   * @param logGroup the log group to use on the lambda.
+   */
+  public static setAutoDeleteObjectsLogGroup(scope: Construct, logGroup: logs.ILogGroup): void {
+    const provider = AutoDeleteObjectsProvider.getProvider(scope, AUTO_DELETE_OBJECTS_RESOURCE_TYPE);
+
+    if (provider === undefined) {
+      throw new Error('Requires at least one bucket with \'autoDeleteObjects: true\'. None is found.');
+    }
+    provider.configureLambdaLogGroup(logGroup.logGroupName);
```
GavinZZ, line 1906: Would this code be problematic eventually? Consider the example where I have the following setups:
```
  autoDeleteObjects: true,
});

s3.Bucket.setAutoDeleteObjectsLogGroup(this, new logs.LogGroup(this, 'MyLogGroup', {
  logGroupName: 'MyLogGroup',
  retention: logs.RetentionDays.FIVE_YEARS
}))

const bucket2 = new s3.Bucket(this, 'MyTempFileBucket2', {
  autoDeleteObjects: true,
});
```

The first s3 bucket creation and `setAutoDeleteObjectsLogGroup` will work as expected. However, during the second bucket creation, it will use the existing singleton custom resource function (since it already exists) and a lambda log group is configured already.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1882,6 +1886,26 @@ export class Bucket extends BucketBase {
...
+   * this will throw an exception as no AutoDeleteObjects provider lambda will exist
+   * in the stack.
+   *
+   * @param scope the stack with bucket(s) with `autoDeleteObjects: true`.
+   * @param logGroup the log group to use on the lambda.
+   */
+  public static setAutoDeleteObjectsLogGroup(scope: Construct, logGroup: logs.ILogGroup): void {
+    const provider = AutoDeleteObjectsProvider.getProvider(scope, AUTO_DELETE_OBJECTS_RESOURCE_TYPE);
+
+    if (provider === undefined) {
```
GavinZZ, line 1903: This feels weird to me where I would expect users to do something 
```
s3.Bucket.setAutoDeleteObjectsLogGroup(this, new logs.LogGroup(this, 'MyLogGroup', {
  logGroupName: 'MyLogGroup',
  retention: logs.RetentionDays.FIVE_YEARS
}))

  autoDeleteObjects: true,
});
```

GavinZZ (changes requested):
Left some comments

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1882,6 +1886,26 @@ export class Bucket extends BucketBase {
...
+   * Note: Calling this multiple times will override the previously set log group
+   *
+   * In the case of no buckets with `autoDeleteObjects: true` exist in the stack,
+   * this will throw an exception as no AutoDeleteObjects provider lambda will exist
+   * in the stack.
+   *
+   * @param scope the stack with bucket(s) with `autoDeleteObjects: true`.
+   * @param logGroup the log group to use on the lambda.
+   */
+  public static setAutoDeleteObjectsLogGroup(scope: Construct, logGroup: logs.ILogGroup): void {
```
TheRealAmazonKendra, line 1900: I don't think this is the right place for this function. Given that the lambda is a singleton for the entire stack, such a function should be at the stack level. Having it exist on the Bucket class and taking in a scope that can be set at the construct level is misleading as to the actual scope of this function.

TheRealAmazonKendra: You could argue that you've explained that in the docstring, but I think this will still mislead customers.

TheRealAmazonKendra: Also, as mentioned last week, I think that a solution here needs to be implemented so that it is consistent across custom resources.

TheRealAmazonKendra (dismissed):
As discussed last week, I don't believe that this is the right place for this functionality to exist.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
samson-keung: I have updated the code to use a different implementation. But to answer your question:

The `bucket2` will re-use to the same singleton custom resource function which has been configured to use `MyLogGroup`. This is expected.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
samson-keung: Implementation is updated such that the order of the method calls no longer matters.

packages/@aws-cdk/cx-api/FEATURE_FLAGS.md
```diff
@@ -128,7 +130,8 @@ The following json shows the current recommended set of flags, as `cdk init` wou
     "@aws-cdk/aws-codepipeline:defaultPipelineTypeToV2": true,
     "@aws-cdk/aws-kms:reduceCrossAccountRegionPolicyScope": true,
     "@aws-cdk/aws-eks:nodegroupNameAttribute": true,
-    "@aws-cdk/aws-ec2:ebsDefaultGp3Volume": true
+    "@aws-cdk/aws-ec2:ebsDefaultGp3Volume": true,
+    "@aws-cdk/aws-ecs:removeDefaultDeploymentAlarm": true
```
GavinZZ, line 134: Can you update from main to get rid of these changes (which I believe is not part of ur changes).

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -274,6 +274,59 @@ export abstract class HandlerFrameworkClass extends ClassType {
...
+          stmt.ret(expr.directCode(`stack.node.tryFindChild(id) as ${this.type}`)),
+        );
+
+        const useLogGroupMethod = this.addMethod({
+          name: 'useLogGroup',
+          static: true,
+          docs: {
+            summary: 'Set the log group to be used by the singleton provider',
+          },
+        });
```
GavinZZ, line 308: Technically you're adding `useLogGroup` functionality to all custom resource handlers not only the s3 ones right? But the functionality is only exposed to the s3 auto delete handler. It would be nice to expose `useLogGroup` to other handlers as I believe this issue doesn't only happen on s3 auto delete but also on all custom resource handlers.

GavinZZ (approved):

GavinZZ:
Discussed with Samson on this issue. Since there are over 27 upvotes, will merge the fix for s3 custom resource handler first. There’s already a tracking issue for the underlying issue for all custom resources. Will handle that separately.

GavinZZ:
@mergify update

samson-keung:
> Discussed with Samson on this issue. Since there are over 27 upvotes, will merge the fix for s3 custom resource handler first. There’s already a tracking issue for the underlying issue for all custom resources. Will handle that separately.

Linking the issue tracking the ability to set log retention across all custom resource: https://github.com/aws/aws-cdk/issues/23909

GavinZZ:
@mergify update

Issue #29949, opened 2024-04-24
:exclamation: NOTICE (custom-resources): attribute error - vendor response doesn't contain `endpointAddress`


grant-d:
Please add your +1 👍 to let us know you have encountered this

---

### Describe the bug

If I upgrade `cdk` & `cdk-lib` to `2.138.0` from `2.137.0`, the following error occurs at deployment. 
If I rollback, the error goes away.

```bash
CustomResource attribute error: Vendor response doesn't contain endpointAddress attribute in object ...|...IoTEndpoint
```

This was previously working for quite some time.

The specific resource is an `AwsCustomResource` related to `IoT`.
(I could not find a way to get the endpoint using `aws-cdk-lib/aws-iot`, so had to use a custom resource)

Code repro below.

### Expected Behavior

Deployment succeeds

### Current Behavior

Deployment fails with the following error:

```bash
8:34:21 AM | UPDATE_FAILED        | AWS::SSM::Parameter             | fooservice/Stage2...IotEndpointAddress
CustomResource attribute error: Vendor response doesn't contain endpointAddress attribute in object
arn:aws:cloudformation:eu-central-1:<account>:stack/<...>|...IoTEndpoint...
```

### Reproduction Steps

```typescript    
    import * as cus from 'aws-cdk-lib/custom-resources'

    // https://stackoverflow.com/a/77092236
    const getIoTEndpoint = new cus.AwsCustomResource(this, 'IoTEndpoint', {
      onCreate: {
        service: 'Iot',
        action: 'describeEndpoint',
        // I am guessing the failure is on the following line (and/or below)
        physicalResourceId: cus.PhysicalResourceId.fromResponse('endpointAddress'),
        parameters: {
          endpointType: 'iot:Data-ATS'
        }
      },
      policy: cus.AwsCustomResourcePolicy.fromSdkCalls({ resources: cus.AwsCustomResourcePolicy.ANY_RESOURCE })
    })

    // I am guessing the failure is on the following line (and/or above)
    const IOT_ENDPOINT = getIoTEndpoint.getResponseField('endpointAddress') // eg, abc...xyz-ats.iot.us-west-2.amazonaws.com
```

### Possible Solution

Rollback `cdk-lib` to `2.137.0`

### CDK CLI Version

2.119.0 (build 0392e71)

### Framework Version

2.138.0

### Node.js Version

v20.7.0

### OS

MacOS

### Language

TypeScript

### Language Version

5.4.5

[...]

PR #30418, opened 2024-06-01
fix(custom-resources): `logApiResponseData` is added to custom resource event properties by default

### LINKED ISSUE #30121: custom-resources: PR:29648 caused Custom resources to be replaced
### Describe the bug

the following PR - https://github.com/aws/aws-cdk/pull/29648 added an additional "logApiResponseData":true to the custom resource properties.

This triggered a replacement of existing custom resources which is not a good outcome for "onCreate" resources that are only meant to be executed once

### Expected Behavior

Not have existing custom-resources replaced with this update.

### Current Behavior

"logApiResponseData":true was not present prior to this PR and the custom resource was not triggered on stack update

### Reproduction Steps

will trigger after an update of aws-cdk-lib

### Possible Solution

allow this value to be suppressed so it doesn't trigger a resource replacement ?

_No response_

### CDK CLI Version

2.138.0

### Framework Version

2.138.0

### Node.js Version

18.16.1

### OS

linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

colifran:
Closes #30121, #29949

### Reason for this change

PR #29648 introduced a new resource property `logApiResponseData`. This resource property is `true` by default which forces an update for `AwsCustomResource`. For users without `onUpdate` configured an empty data object is returned if no SDK call is configured. This can cause an attribute error if the user is depending on a data from a specific SDK call.

### Description of changes

Made `logApiResponseData` undefined by default which will not trigger `onUpdate`. To maintain backwards compatibility with the original PR introducing `logApiResponseData` as true by default, I've also introduced a feature flag that will allow users to keep the current behavior so they aren't now forced into another `onUpdate` event.

Updated unit tests where `logApiResponseData` was added as a resource property. Added new unit test to verify that `logApiResponseData` could be added to the event. Updated unit tests that test `_render()` to ensure that the default case will result in an empty object. Updated integ tests.

### Checklist

----


packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1088,6 +1089,20 @@ export const FLAGS: Record<string, FlagInfo> = {
     recommendedValue: true,
     compatibilityWithOldBehaviorMd: 'Set AWS::ECS::Service \'DeploymentAlarms\' manually to restore the previous behavior.',
   },
+
+  //////////////////////////////////////////////////////////////////////
+  [LOG_API_RESPONSE_DATA_PROPERTY_TRUE_DEFAULT]: {
```
colifran, line 1094: Does it make sense to add this feature flag? My thinking is that by remove `logApiResponseData` as a default event property we're now forcing another update for people who may not want it. For those people, they can enable this feature flag to maintain the existing behavior. For everyone else, they can keep it set to `false` and `logApiResponseData` will be undefined by default.

comcalvi: I think the feature flag makes sense, but why is the recommended value `false`?

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1088,6 +1089,20 @@ export const FLAGS: Record<string, FlagInfo> = {
...
+    type: FlagType.BugFix,
+    summary: 'When enabled, the custom resource used for AwsCustomResource will configure the logApiResponseData property as true by default',
+    detailsMd: `
```
comcalvi, line 1097: This is a bit hard to follow. Particularly, 

> Unlike most feature flags, we don't recommend setting this feature flag
      to true unless you're using the AwsCustomResource construct with logApiResponseData set to true as a custom resource event property by default and do not
      want to trigger an update by letting the default logApiResponseData value go back to undefined.

The "this will trigger an update event to your custom resource" should be earlier in the statement, not in the front. How about something like this?

```
This results in `logApiResponseData` being passed as true to the custom resource provider. This will cause the custom resource handler to receive an `Update` event. If you not have a handler for the `Update` event, you will see this error from CFN:

CustomResource attribute error: Vendor response doesn't contain endpointAddress attribute in object ...

see https://github.com/aws/aws-cdk/issues/29949) for more details.

 Unlike most feature flags, we don't recommend setting this feature flag
      to true unless you're using the AwsCustomResource construct with logApiResponseData set to true as a custom resource event property by default and do not
      want to trigger an update by letting the default logApiResponseData value go back to undefined.
```

comcalvi (changes requested):
Only doc improvement comments from me, otherwise looks good.

packages/aws-cdk-lib/cx-api/lib/features.ts
colifran: `false` because it isn't the recommended behavior, just a convenience to anyone that has adopted it or adapted to it since CDK v2.138.0

packages/aws-cdk-lib/cx-api/lib/features.ts
mjgoble: Does this mean that those of us who have already upgraded to the latest version of cdk and dealt with the fallout the AwsCustomResources updating unexpectedly now have to set a feature flag just to avoid another AwsCustomResource update?

jnawk: Given general assurances associated with CDKv2 that there would never be breaking changes, I appreciate that "just don't release breaking changes" is a bit of an impossible ask - bugs happen - but this does make me wonder what measures have been taken to prevent this kind of breaking change from occurring again in the future - specifically in this case, the unexpected change to the arguments given to otherwise unchanged custom resource resources. 

---

While we're musing about things, the custom resource implementation has a choice of what to do when given a request to update, yet no SDK call to make in such a case - it must either:

a) signal failure, causing a stack update rollback - consequence is this app cannot be updated without pinning a specific CDK version, or until the CDK bug is fixed, **or** 

b) do what it currently does - return _something_ - an empty response, thereby breaking any downstream resources referring to outputs - presumably also causing an update rollback - but that update rollback would itself cause an update of the custom resource, since no physical resource replacement has occurred, so again we will return an empty response - so even if we somehow get the stack rolled back, we _still_ no longer have those outputs - any future reference to them must forever fail..  This stack (or at least the downstream resource(s)) has been bricked.

Surely option (a) is _much_ safer.

colifran: @jnawk Thanks for sharing your thoughts on this. I guess my bigger questions is why we aren't enforcing an SDK call to be configured for both `onCreate` and `onUpdate` if someone is depending on response data. Resource updates in and of themself aren't necessarily a bug. However, in this case relying on specific SDK call response data and it not being there due to not having a configured SDK call for `onUpdate` leads to a deploy time error. The documentation isn't great, but we mention this exact scenario when using `getResponseField` and `getResponseFieldReference` as part of their docstrings:
https://github.com/aws/aws-cdk/blob/3ee5ce03905e5d12d569bcc58b8134c81a97d7c7/packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/aws-custom-resource.ts#L596-L598
https://github.com/aws/aws-cdk/blob/3ee5ce03905e5d12d569bcc58b8134c81a97d7c7/packages/aws-cdk-lib/custom-resources/lib/aws-custom-resource/aws-custom-resource.ts#L580-L582

This scenario is also mentioned in the README here as a limitation: https://github.com/aws/aws-cdk/blob/3ee5ce03905e5d12d569bcc58b8134c81a97d7c7/packages/aws-cdk-lib/custom-resources/README.md?plain=1#L539

Any thoughts on this?

jnawk: Hi @colifran - I sure do have some thoughts..  

Firstly - resource lookup type lookups - perhaps you have the name of a subnet and want the subnet ID - this is the kind of thing that in theory will not change - the point here is the application doesn't need to know the ugly details like subnet ID, rather just needs to know a more friendly value, like "apps-subnet" or "loadbalancer-subnet" or whatever.   These kinds of resources are certainly safe to "update" - just do the lookup again, job done, and for this we can specify only an `on_update` and be done.  Not everyone necessarily knows this however, so they instead specify only an `on_create` - now we've got unexpected problems - because something we didn't expect to be updated, something that for reasons _can't_ be updated, none the less **is** being updated.  I agree it's technically possible to enforce an `on_update` being configured _if_ the outputs are referenced, but to change this now in v2.x would be yet another breaking change, and populating an on_update would necessarily cause CFn to update the resource next time the stack was deployed!

Secondly - resources that legitimately can not be updated, nor replaced (by a single AWS SDK call) - for instance named resources - eg an S3 bucket - ignoring for now why one would create an S3 bucket this way - that's not the point - the immutability of the property that also uniquely identifies the resource being the point - CFn will try to update the resource, and if given a new physical ID, delete the old, since the resource has been replaced - but for this class of resource, the old needs to be deleted first, & the new created.  This actually isn't a problem unique to these custom resources - if I have a named AWS::S3::Bucket in my template and I try to rename it, my stack fails and I need to change the logical ID in the template.  For this class of resource, I am not sure we can be so blase about them being updated just coz someone changed the framework - if these resources are updated because framework is changed (as is the case for this bug), then we are in big trouble, yet these are examples of resources which would legitimately not have an `on_create` defined.  If these resources were all hand-rolled, the wouldn't be updated - CFn doesn't have a track record of doing this kind of thing.


That all said, and I apologise, it's a little hard to follow, I know, having such a limitation in AwsCustomResource wouldn't be a deal-breaker per se, since one could always roll a bespoke custom resource the old way.  (Of course, imposing such a restriction in aws-cdk-lib v2.x would constitute a breaking change, so please don't!  Leave it for v3.

jnawk: Also I agree that documentation isn't great - it's pretty clear with the problem well understood, but without that understanding, it doesn't really call out all the foot guns.

comcalvi (approved):

PR #30319, opened 2024-05-23
feat(stepfunctions-tasks): add workerType and numberOfWorkers to GlueStartJobRun class

mazyu36:
### Issue # (if applicable)

Closes #12757.

### Reason for this change
Missing property

### Description of changes
Add workerType and numberOfWorkers to GlueStartJobRun class.

The reasons for this change are as follows:

* AllocatedCapacity is deprecated.
* MaxCapacity can only be used with Glue version 1 and earlier, which have already reached end of support (EOS).
* Glue version 2 and later use WorkerType and NumberOfWorkers.

For mor information, see also the documents below.

https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-jobs-runs.html#aws-glue-api-jobs-runs-StartJobRun

https://docs.aws.amazon.com/glue/latest/dg/glue-version-support-policy.html

### Description of how you validated changes
Add unit tests and integ tests.

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30426, opened 2024-06-02
feat(stepfunctions-tasks): add guardrailConfiguration and trace property to the BedrockInvokeModel

### LINKED ISSUE #30425: stepfunctions-tasks: enable guardrail and trace settings for BedrockInvokeModel
### Describe the feature

Bedrock [InvokeModel](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html) API supports guardrail configuration and trace.

But in [BedrockInvokeModel](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions_tasks.BedrockInvokeModel.html), both options are not supported.

### Use Case

To use guardrail and trace when invoke the Bedrock model.

### Proposed Solution

Add guardrailConfiguration and trace property to the BedrockInvokeModel class.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.143.0

all

mazyu36:
### Issue # (if applicable)

Closes #30425 

### Reason for this change
In the current [BedrockInvokeModel](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions_tasks.BedrockInvokeModel.html), guardrail configuration and trace for the invocation are not supported.

### Description of changes
Add `gurdrailConfiguration` and `trace` property to the `BedrockInvokeModel`

### Checklist

----


packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -102,9 +132,23 @@ export interface BedrockInvokeModelProps extends sfn.TaskStateBaseProps {
...
+   * @default - No guardrail is applied to the invocation.
+   */
+  readonly guardrailConfiguration?: GuardrailConfiguration;
+
+  /**
+   * Specifies whether to enable or disable the Bedrock trace.
+   *
+   * @default - Trace is not enabled for the invocation.
+   */
+  readonly trace?: Trace;
```
lpizzinidev, line 151:
```suggestion
  readonly traceEnabled?: boolean;
```
We could avoid declaring the extra `enum`.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -211,6 +274,9 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
         Output: this.props.output?.s3Location ? {
           S3Uri: `s3://${this.props.output.s3Location.bucketName}/${this.props.output.s3Location.objectKey}`,
         } : undefined,
+        GuardrailIdentifier: this.props.guardrailConfiguration?.guardrailIdentifier,
+        GuardrailVersion: this.props.guardrailConfiguration?.guardrailVersion,
+        Trace: this.props.trace,
```
lpizzinidev, line 279:
```suggestion
        Trace: this.props.traceEnabled ? 'ENABLED' : 'DISABLED',
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -102,9 +132,23 @@ export interface BedrockInvokeModelProps extends sfn.TaskStateBaseProps {
...
+   * @default - The API response body is returned in the result.
    */
   readonly output?: BedrockInvokeModelOutputProps;
+
+  /**
+   * The guardrail is applied to the invocation
+   *
+   * @default - No guardrail is applied to the invocation.
+   */
+  readonly guardrailConfiguration?: GuardrailConfiguration;
```
lpizzinidev, line 144:
```suggestion
  /**
   * The unique identifier of the guardrail that you want to use.
   * If you don't provide a value, no guardrail is applied to the invocation.
   */
  readonly guardrailIdentifier?: string;

  /**
   * The version number for the guardrail.
   * The value can also be `DRAFT`.
   */
  readonly guardrailVersion?: string;
```
To provide a simpler interface, I'd rather expose the two attributes directly and apply the `guardrailVersion must be specified if guardrailIdentifier` is specified validation.
What do you think?
Also, we should apply validation for the accepted values according to the [docs](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html#API_runtime_InvokeModel_RequestSyntax)

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -144,6 +188,10 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
       throw new Error('Output S3 object version is not supported.');
     }
 
+    if (props.guardrailConfiguration && props.contentType && (props.contentType !== 'application/json')) {
```
lpizzinidev, line 191: > You don't provide a guardrail identifier but you specify the amazon-bedrock-guardrailConfig field in the request body.

Can we check this as well?

Also, don't forget to check if the attributes are resolved before applying validation.
Otherwise, we might block valid deployments.

lpizzinidev (changes requested):
Thanks 👍 Left some comments for adjustments on the interface.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
mazyu36: @lpizzinidev 
Thank you. 

Since guardrailIdentifier and guardrailVersion need to be used as a set, I thought it would be better to have them as required fields in the current standalone interface. What do you think?

For your reference, there was a similar discussion [here](https://github.com/aws/aws-cdk/pull/30319#discussion_r1612489270).

I also like the flat approach, but it could lead to an increase in validation patterns, and since bedrock is a growing service, there's a possibility of more guardrail fields being added in the future.

Adding accepted values validation is agree.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
mazyu36: @lpizzinidev 
Thank you. 
Regarding `amazon-bedrock-guardrailConfig`, after reviewing the [documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters.html), it doesn't seem to be supported in any of the foundation models yet.

I tried configuring it in the management console and running InvokeModel, but it was treated as an invalid value and couldn't be executed.

Since it's not supported yet, I thought it might be better to avoid adding validation for it. What do you think?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
mazyu36: Thank you. 
I'm concerned that with the proposed approach, if `traceEnabled` is undefined, `DISABLED` would be set in the template, potentially modifying existing templates. 
To account for `undefined` case, I'm considering the following approach. 
If you have any better suggestions, please let me know.

```ts
        Trace: this.props.traceEnabled === undefined
          ? undefined
          : this.props.traceEnabled
            ? 'ENABLED'
            : 'DISABLED',
```

lpizzinidev: That's a valid concern and I agree with your proposed and improved solution.
Thanks for the feedback!

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
lpizzinidev: I remember having similar discussions in the past in favor of the plain interface approach for other scenarios (as suggested by the [design guidelines](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#flat)).
Keeping the nested structure saves us from implementing explicit validation.
However, providing a clean interface for other programming languages is also a concern.

> since bedrock is a growing service, there's a possibility of more guardrail fields being added in the future

That's a good point. Let's keep it like this and see what the core team thinks about it 👍

mazyu36: Thank you. For now, we've decided to keep the property design approach as is, without any modifications.

I'd like to wait for the maintainer's review.

I have added the validation.​​​​​​​​​​​​​​​​

mazyu36:
@lpizzinidev 
Thank you for the review!
I’v addressed all your comments.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -189,6 +220,21 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
       );
     }
 
+    if (this.props.guardrailConfiguration !== undefined) {
```
lpizzinidev, line 223:
```suggestion
    if (!this.props.guardrailConfiguration) {
```
Cleaner

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -211,7 +257,33 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
...
+      throw new Error(`guardrailIdentifier must match the ^(([a-z0-9]+)|(arn:aws(-[^:]+)?:bedrock:[a-z0-9-]{1,20}:[0-9]{12}:guardrail/[a-z0-9]+))$ pattern, got ${props.guardrailConfiguration.guardrailIdentifier}`);
+    }
+
+    if (props.guardrailConfiguration.guardrailVersion && !/^(([1-9][0-9]{0,7})|(DRAFT))$/.test(props.guardrailConfiguration.guardrailVersion)) {
+      throw new Error(`guardrailVersion must match the ^(([1-9][0-9]{0,7})|(DRAFT))$ pattern, got ${props.guardrailConfiguration.guardrailVersion}`);
+    }
+
+    if (props.guardrailConfiguration.guardrailIdentifier && (props.contentType !== 'application/json')) {
+      throw new Error('You must set contentType to \'application/json\' when use guardrailConfiguration');
+    }
```
lpizzinidev, line 286:
```suggestion
  private validateGuardrailConfiguration(props: BedrockInvokeModelProps) {
    if (!props.guardrailConfiguration) return;

    const { guardrailIdentifier, guardrailVersion } = props.guardrailConfiguration;

    if (!Token.isUnresolved(guardrailIdentifier)) {
      const guardrailConfigurationPattern = /^(([a-z0-9]+)|(arn:aws(-[^:]+)?:bedrock:[a-z0-9-]{1,20}:[0-9]{12}:guardrail\/[a-z0-9]+))$/;
      if (!guardrailConfigurationPattern.test(guardrailIdentifier)) {
        throw new Error(`guardrailIdentifier must match the \`${guardrailConfigurationPattern}\` pattern, got ${guardrailIdentifier}`);
      }
      if (props.contentType !== 'application/json') {
        throw new Error(`You must set contentType to \'application/json\' when using guardrailConfiguration, got ${props.contentType}.`);
      }
    }

    const guardrailVersionPattern = /^(([1-9][0-9]{0,7})|(DRAFT))$/;
    if (!Token.isUnresolved(guardrailVersion) && !guardrailVersionPattern.test(guardrailVersion)) {
      throw new Error(`guardrailVersion must match the \`${guardrailVersionPattern}\` pattern, got ${guardrailVersion}`);
    }
  }
```
We should validate the two props independently

lpizzinidev (changes requested):
Thanks 👍 Left some notes for some final adjustments

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
mazyu36: Thank you. 
I think `!` is unnecessary, so I have removed it.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
mazyu36: Thank you. 
The method you suggested includes `/` around the regular expression in the error message, so I wrote it directly in the error message instead. 
If you have a better way, please let me know.

mazyu36:
@lpizzinidev 
Thanks. I addressed all your comments.

lpizzinidev (approved):
Nice 👍

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
paulhcsun: Hi both, the teams current stance is that we would prefer to have this interface containing the two required properties rather than exposing two separate attributes as it allows us to avoid adding more error handling and just enforce that both attributes are specified when defining the resource. 

While the section linked from the Design Guidelines does state that flat property structure is preferred over a nested structure, that is only the case when the two properties are not required to be specified together. I can update that to clarify the distinction.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -211,7 +257,36 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
...
+
+  private validateGuardrailConfiguration(props: BedrockInvokeModelProps) {
+    if (!props.guardrailConfiguration) return;
+
+    const { guardrailIdentifier, guardrailVersion } = props.guardrailConfiguration;
+
+    if (!Token.isUnresolved(guardrailIdentifier)) {
+      const guardrailConfigurationPattern = /^(([a-z0-9]+)|(arn:aws(-[^:]+)?:bedrock:[a-z0-9-]{1,20}:[0-9]{12}:guardrail\/[a-z0-9]+))$/;
+      if (!guardrailConfigurationPattern.test(guardrailIdentifier)) {
+        throw new Error(`guardrailIdentifier must match the ^(([a-z0-9]+)|(arn:aws(-[^:]+)?:bedrock:[a-z0-9-]{1,20}:[0-9]{12}:guardrail\/[a-z0-9]+))$ pattern, got ${guardrailIdentifier}`);
```
paulhcsun, line 279: Could we also add validation on the [length constraints](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html#API_runtime_InvokeModel_RequestSyntax:~:text=Length%20Constraints%3A%20Minimum%20length%20of%200.%20Maximum%20length%20of%202048.)? 

Min. 0, Max. 2048.

paulhcsun (changes requested):
Hi @mazyu36, thanks for the contribution! Just left one comment about an additional piece of validation to add but otherwise this looks good!

Thanks for the review @lpizzinidev!

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
mazyu36: @paulhcsun 
Thanks. 
I've added a validation and a unit test.

mazyu36:
@paulhcsun 
Thank you for the review.
I've addressed your comment.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
TheRealAmazonKendra: Both sides are technically correct here. In this case, I'd suggest using `guardrail` and instead of a nested set of values, make it a class with a single function so that it would look something like `guardrail: Guardrail.apply('my-guardrail', 1)`, even better, make the second value optional and if it isn't supplied, we automatically set that value to `DRAFT` for the user.

But also, if you can think of better naming conventions than mine, that would probably be good. I'm terrible at naming anything.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
TheRealAmazonKendra: If you are going to provide this kind of error checking could you please make the error message descriptive of what the expected input is instead of regex? It's better readability.

mazyu36: @TheRealAmazonKendra 
Thanks. I've changed the error message.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
mazyu36: @TheRealAmazonKendra  @paulhcsun 
Thank you. I think that approach is good too.
However, in this case, there's a possibility of unintentionally applying the DRAFT version of the guardrail. 
Since guardrail is a security-related configuration, I thought it might be better to make the version a required field in the current implementation, as it is now. This would help prevent the application of unintended versions. 
What do you think about this?

Also, I've modified the property name to `guardrail` as I thought it would be better.

TheRealAmazonKendra: Totally fair concern, in that case, I would create an enum like class here for `guardrail`

```
public class Guardrail {
  public static enable(string: id, version: number);

  public static enableDraft(string: id);
}
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -221,15 +221,18 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
     }
 
     if (this.props.guardrail) {
+      const isArn = this.props.guardrail.guardrailIdentifier.startsWith('arn:');
```
mazyu36, line 224: The policy document was incorrect for the case when guardrailIdentifier is an ARN, so I added a condition to address this.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
mazyu36: Thank you. I apologize for the repeated requests, but I still can't agree that the proposed method is better.

The guardrail version is a string in the L1 Construct, so using a method to make it a number would require casting.
Also, there's [validation for 'contentType'](https://github.com/aws/aws-cdk/pull/30426/files#diff-a9f9ef1a1d3f4c941af934cf7943b7be05c856d20246c0674fdfa72dacceb15fR284) which is not included in the guardrail, and [generation of policy documents](https://github.com/aws/aws-cdk/pull/30426/files#diff-a9f9ef1a1d3f4c941af934cf7943b7be05c856d20246c0674fdfa72dacceb15fR225). Because of these factors, I think the current approach of not separating classes is more understandable.
The proposed approach seems over-optimized to me.

If you still think the proposed method is better despite these points, could you provide an example of where a similar approach is used elsewhere? 
This would help me understand the implementation concept better.

paulhcsun: Hi @mazyu36, one example where a similar approach is used is in the [`Schedule` class in `aws-events`](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-events/lib/schedule.ts#L11-L70).

I think that casting from number to string is not a major concern here, and avoids us having to validate the version if a string is passed in. 

As for generating the policy documents, I think those can be left where it is, but I agree there's not a super clean way to validate for `contentType` without passing in the `contentType` to the method which is not a straightforward DX. I will discuss with the team and get back to you on that tomorrow.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -211,7 +260,40 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
...
+      if (props.contentType !== 'application/json') {
+        throw new Error(`You must set contentType to \'application/json\' when using guardrail, got '${props.contentType}'.`);
+      }
+      if (guardrailIdentifier.length > 2048) {
+        throw new Error(`\`guardrailIdentifier\` length must be between 0 and 2048, got ${guardrailIdentifier.length}.`);
+      }
+
+    }
+
+    const guardrailVersionPattern = /^(([1-9][0-9]{0,7})|(DRAFT))$/;
```
paulhcsun, line 293: I think I would prefer a stricter check that just ensures the version is between 10 and 99999999 or is "DRAFT" instead of using regex here. Just to improve the readability of this code and make it easier to maintain in the future.

paulhcsun (dismissed):

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
paulhcsun: Hi @mazyu36, I believe the only allowed value we currently accept for `contentType` is `application/json`. I assume that when this module was first written it was made to be an optional property to reserve the right to support different `contentType`s in the future but for the time being I think that we should just deprecate the property and remove this check. If the day comes that we need to support additional `contentType`s then we can just undeprecate it.

mazyu36: @paulhcsun 
Thank you. I've added the `GuardRail` Class and made `contentType` deprecated.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
mazyu36: Thank you. I have implemented it as follows.

https://github.com/aws/aws-cdk/pull/30426/files#diff-254b356f1a1b98a7034dec79302bc664e46a4de89a1fe14aa5c62d8982109dc2R15

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -211,7 +260,40 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
...
     };
   }
+
+  private validateGuardrail(props: BedrockInvokeModelProps) {
+    if (!props.guardrail) return;
+
+    const { guardrailIdentifier, guardrailVersion } = props.guardrail;
+
+    if (!Token.isUnresolved(guardrailIdentifier)) {
+      const guardrailPattern = /^(([a-z0-9]+)|(arn:aws(-[^:]+)?:bedrock:[a-z0-9-]{1,20}:[0-9]{12}:guardrail\/[a-z0-9]+))$/;
```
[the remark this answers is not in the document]
mazyu36: Thank you. However, after trying various approaches and searching for other implementation examples within the package, I couldn't figure out how to check the ARN. If you have any implementation examples, could you please share them with me?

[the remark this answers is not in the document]
paulhcsun: Yes for sure, here's one example of using the `formatArn()` method: https://github.com/aws/aws-cdk/blob/b7f626b9c845dd4161c4d37fd32835438b05124a/packages/aws-cdk-lib/aws-route53/README.md?plain=1#L305

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
paulhcsun: Awesome! The implementation looks good, I'd be happy to approve after the ARN class change is updated.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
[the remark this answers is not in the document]
mazyu36: @paulhcsun 
Thank you.
I apologize for my poorly phrased question. While I was aware of the existence of the `formatArn` method, I didn't understand how to use it for validation.
After further investigation, I've implemented validation using `ArnFormat`. Is this approach on the right track?

https://github.com/aws/aws-cdk/pull/30426/commits/7f307c9ee009c630a889953c8182e3a3d33fdfdc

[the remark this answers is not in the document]
paulhcsun: Apologies, I realize now that `.formatArn()` wouldn't have helped here. Your implementation looks good to me and follows a [similar example](https://github.com/aws/aws-cdk/blob/3eef58928f77cf1b7a79c228446cc39be6dbd2f1/packages/aws-cdk-lib/aws-appconfig/lib/deployment-strategy.ts#L46-L50).

paulhcsun (approved):
Nice work @mazyu36! I've approved the PR but added a `do-not-merge` label as we're currently holding off on merging anything while fixing a custom resources related issue which touches a lot of integ tests. This change probably would not affect that but we're pausing all merges temporarily to be safe. We should hopefully be able to get this merged in tomorrow if not sometime next week. Thanks for your patience and understanding! :pray:

PR #30428, opened 2024-06-02
refactor(events-targets): replace custom resource with native grantWrite method for log group write permissions

### LINKED ISSUE #20855: (events): log group target
### Describe the feature

In #10598, the author added support for CloudWatch Logs log group as a target of an EventBridge event bus. The author noted that at that time, CloudFormation did not have support for CWL resource policy, they used a custom lambda to manipulate the policy. Since then, CFN added the resource support and another author added support in #17015.

Now the custom resource added in #10598 can be removed in favour of the native support.

### Use Case

Removing tech debt. Custom resource lambda functions require S3 bucket to provision the code.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.29.0

N/A

frankpengau:
### Issue # (if applicable)

Closes #20855

### Reason for this change

The custom resource lambda used to just add the log group resource policy seems like an unnecessary use of the lambda resource. It would be better to refactor it to use the native `grantWrite` method for CloudWatch Log Group to grant the write permissions necessary for eventbridge. 

### Description of changes

Remove custom cloudwatch log group rule target custom resource lambda to add write permissions and replace with `grantWrite`.

Similar to how sqs grants write permissions for eventbridge: https://github.com/aws/aws-cdk/blob/efbe6debaf1ccebbcd884912ccb38cb13a989061/packages/aws-cdk-lib/aws-events-targets/lib/sqs.ts#L72

Changed unit test, but might need some help to get the integration test to work

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-events-targets/lib/log-group.ts
+++ b/packages/aws-cdk-lib/aws-events-targets/lib/log-group.ts
@@ -1,4 +1,3 @@
-import { LogGroupResourcePolicy } from './log-group-resource-policy';
 import { TargetBaseProps, bindBaseTargetConfig } from './util';
 import * as events from '../../aws-events';
 import { RuleTargetInputProperties, RuleTargetInput, EventField, IRule } from '../../aws-events';
@@ -78,14 +77,6 @@ export interface LogGroupProps extends TargetBaseProps {
    * @default - the entire EventBridge event
    */
   readonly logEvent?: LogGroupTargetInput;
-
-  /**
-   * Whether the custom resource created wll default to
-   * install latest AWS SDK
-   *
-   * @default - install latest AWS SDK
-   */
-  readonly installLatestAwsSdk?: boolean;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


TheRealAmazonKendra:
This was used because a ResourcePolicy is needed here, which is what the grant function is creating under the hood. When this was first written, that resource wasn't available in all regions so we used a Custom Resource. 

Unfortunately, providing a migration path to not using said custom resource isn't completely straight forward due to the `onDelete` function added here and because there is a limit of 10 resource policies allowed per region/account. If we try to add a new policy to perform a replacement, it might cause users to hit that limit and then the deployment will fail. 

I'm actually working on this same issue for another package right now.

frankpengau:
@TheRealAmazonKendra If you're working on this in another package and are able to also add in this package as well, that would be great! 🙏 

Happy to close this PR off for now and wait for your changes to fix it for good.

PR #30440, opened 2024-06-03
fix: route53 CrossAccountZoneDelegationRecord fails at deployment time with imported `delegatedZone`

### LINKED ISSUE #28581: aws-cdk-lib: CrossAccountZoneDelegationRecord does not validate hostedZoneNameServers
### Describe the bug

Our deployment failed when trying to use `CrossAccountZoneDelegationRecord` with an `IHostedZone` that did not have `hostedZoneNameServers` set. In our particular case, `IHostedZone` came from `PublicHostedZone.fromHostedZoneAttributes`.

### Expected Behavior

Either an error should be thrown during the CDK build process, or the construct should support an `IHostedZone` without unset.

### Current Behavior

We received the error during our CF deployment:
```
[FAILED] from custom resource. Message returned: TypeError: Cannot read properties of undefined (reading 'map') at cfnEventHandler
```

### Reproduction Steps

Our internal CDK code looked something like:
```typescript
const hostedZone = route53.PublicHostedZone.fromHostedZoneAttributes(this, 'WebsiteHostedZone', {
  hostedZoneId,
  zoneName: 'same.example.com'',
});

new CrossAccountZoneDelegationRecord(this, 'CrossAccountZoneDelegation', {
  // ...
  parentHostedZoneName: 'same.example.com',
  childHostedZoneName: hostedZone,
});
```

### Possible Solution

This line forcefully asserts that `hostedZoneNameServers` is defined: https://github.com/aws/aws-cdk/blob/a40f2ec1508e9b1718854b4b7c4037d59f197903/packages/aws-cdk-lib/aws-route53/lib/record-set.ts#L793

Then the handler reads from that possibly undefined array: https://github.com/aws/aws-cdk/blob/d69c51a2e10864e69d4610b6df30d40a9bfec3ac/packages/aws-cdk-lib/aws-route53/lib/cross-account-zone-delegation-handler/index.ts#L58

_No response_

### CDK CLI Version

2.77.0

### Framework Version

_No response_

### Node.js Version

v18.12.1

### OS

Ubuntu 22.04.3

### Language

TypeScript

### Language Version

_No response_

### Other information

We are an internal AWS team using SuperNova.

samson-keung:
### Issue # (if applicable)

Closes #28581.

### Reason for this change

An imported `delegatedZone` will not have info about the Name Servers. When it is passed to `CrossAccountZoneDelegationRecord`, the handler will see `undefined` when trying to retrieve the Name Servers info on `delegatedZone`, then throw exception during deployment.

This change throws the exception at build time for a faster feedback loop.

### Description of changes

`CrossAccountZoneDelegationRecord` throws exception if `delegatedZone.hostedZoneNameServers` is undefined.

Add unit test to cover the case of passing an imported HostedZone to `CrossAccountZoneDelegationRecord`

### Checklist

----


samson-keung:
Exemption Request

Reason: Added build time validation. No changes to deployment behaviour, hence, no integ test update

packages/aws-cdk-lib/aws-route53/lib/record-set.ts
```diff
@@ -929,6 +929,10 @@ export class CrossAccountZoneDelegationRecord extends Construct {
       throw Error('Only one of parentHostedZoneName and parentHostedZoneId is supported');
     }
 
+    if (!props.delegatedZone.hostedZoneNameServers) {
```
aaythapa, line 932: nit: I think it'd be nice to have some read me changes [here](https://github.com/aws/aws-cdk/blob/ddbbd002da6298679500d4dd6b6c5b1487bd5f5d/packages/aws-cdk-lib/aws-route53/README.md?plain=1#L246)

packages/aws-cdk-lib/aws-route53/lib/record-set.ts
```diff
@@ -929,6 +929,10 @@ export class CrossAccountZoneDelegationRecord extends Construct {
...
+      throw Error(`Not able to retrieve Name Servers for ${props.delegatedZone.zoneName} due to it being imported.`);
```
aaythapa, line 933: will it always be imported for this construct?

packages/aws-cdk-lib/aws-route53/lib/record-set.ts
samson-keung: Thanks for the callout! I updated the README with a new commit.

packages/aws-cdk-lib/aws-route53/lib/record-set.ts
samson-keung: Are you asking about the "due to it being imported" part of the error message? And should we assume `delegatedZone` is imported?

If so, yes, afaik only imported Hosted Zone would have `hostedZoneNameServers` being undefined.

aaythapa: thats what i was wondering, thanks

aaythapa (approved):

vhelke:
This implementation is problematic for us. We are using the imported delegatedZone method, and since the nameservers are missing we are fetching them separately using a customResource and then by using the `add_property_override` method we are updating the nameserver info.

Is it possible to add an attribute which would bypass the `!props.delegatedZone.hostedZoneNameServers` validation?

samson-keung:
> This implementation is problematic for us. We are using the imported delegatedZone method, and since the nameservers are missing we are fetching them separately using a customResource and then by using the `add_property_override` method we are updating the nameserver info.
> 
> Is it possible to add an attribute which would bypass the `!props.delegatedZone.hostedZoneNameServers` validation?

Thanks for reporting the issue. I have created https://github.com/aws/aws-cdk/issues/30600 to track and will be pushing a fix to replace the error with a warning instead.

PR #29127, opened 2024-02-16
feat(codepipeline): `GitPushFilter` with branches and file paths for trigger

go-to-k:
### Issue # (if applicable)

Closes #29124

Related PR: https://github.com/aws/aws-cdk/pull/29128
Perhaps if one merges, the other will cause a conflict.

### Reason for this change

We would be good to trigger pipelines by GitPushFilter with branches and file paths.

- CFn docs
  - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codepipeline-pipeline-gitpushfilter.html
  - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codepipeline-pipeline-gitbranchfiltercriteria.html
  - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codepipeline-pipeline-gitfilepathfiltercriteria.html

### Description of changes

Add props:

- branchesExcludes
- branchesIncludes
- filePathsExcludes
- filePathsIncludes

### Description of how you validated changes

Both unit and integ tests.

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30462, opened 2024-06-05
revert: "feat(codepipeline): `GitPushFilter` with branches and file paths for trigger"

gracelu0:
Reverts aws/aws-cdk#29127


aaythapa (approved):

go-to-k:
@gracelu0 Why did you revert? I would like you to state reasons. Is it relevant to the following comment?

https://github.com/aws/aws-cdk/pull/30465#issuecomment-2153270981

gracelu0:
> @gracelu0 Why did you revert? I would like you to state reasons. Is it relevant to the following comment?
> 
> [#30465 (comment)](https://github.com/aws/aws-cdk/pull/30465#issuecomment-2153270981)

Hi @go-to-k , yes I believe so. @TheRealAmazonKendra is working on some CodePipeline changes and can provide more context here. Thank you for your effort and apologies as I know the community has been looking forward to this feature. I'll make sure to communicate more clearly next time beforehand.

TheRealAmazonKendra:
This was reverted based off my request because there were a couple bugs in the logic and because the contract here doesn't meet our design guidelines. Wherever possible (and I recognize that it is not possible. 100% of the time), we should not have optional props that are mutually exclusive or that rely on each other. If you require several statements validating the contract you have defined, that indicates that the contract is not quite right.

This is not perfect, this here is a PR that I have in progress to deprecate the triggers as they exist now and to improve the abstraction https://github.com/aws/aws-cdk/pull/30516. While we have limitations of how much we can perfect things here because of the changes that are already released, we should be using enum like classes and adding static functions to abstract away complexity and improve upon the CloudFormation contract.

go-to-k:
Thanks for your clarification. I got it.

PR #30470, opened 2024-06-06
chore(events): rule cannot have more than 5 targets

flexelem:
### Issue # (if applicable)

Closes #<issue number here>.

### Reason for this change
This is a minor refactor for preventing an Event Rule to have at most 5 associated targets defined during a synth. This is the main limit set by EventBridge https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-quota.html#eb-limits

### Description of changes
This is a very simple change in `aws-events` module to check if the `targets` array length of an `Event` is greater than 5. If it is then it throws an Error during build time.

Added a unit test by in `rule.test.ts` by verifying an Error is being thrown when a Rule has more than 5 associated targets.

### Checklist

----


packages/aws-cdk-lib/aws-events/lib/rule.ts
```diff
@@ -320,6 +320,8 @@ export class Rule extends Resource implements IRule {
   private renderTargets() {
     if (this.targets.length === 0) {
       return undefined;
+    } else if (this.targets.length > 5) {
```
lpizzinidev, line 323: What about adding the check inside the `validateRule` method instead?

lpizzinidev (changes requested):
Thanks 👍 Left a suggestion

packages/aws-cdk-lib/aws-events/lib/rule.ts
flexelem: thanks makes sense. I moved it into `validateRule` method.

xazhao (approved):
Thanks for contributing. The change looks good to me.

PR #17176, opened 2021-10-27
feat(cli): added `build` field to cdk.json

comcalvi:
Adds a `build` field to `cdk.json`. The command specified in the `build` will be executed before synthesis. This can be used to build any code that needs to be built before synthesis (for example, CDK App code or Lambda Function code).

This is part of the changes needed for the `cdk watch` command
(https://github.com/aws/aws-cdk-rfcs/pull/383).

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30511, opened 2024-06-10
feat(cli): add ability to configure hotswap properties for ECS

atanaspam:
### Reason for this change

We aim to speed up deployment times in our development environment by using the hotswap feature. However, our services have dependencies on each other, and the current hotswap behavior is too disruptive. 

### Description of changes

We modified the hotswap implementation for ECS services to pass the `minimumHealthyPercent` and `maximumHealthyPercent` configurable parameters. These parameters are exposed to the cli and can be passed as `--hotswap-ecs-minimum-healthy-percent <number>` and `--hotswap-ecs-maximum-healthy-percent <number>`

The implementation is careful to maintain the existing behaviour. That is, if none of the new flags is used, the current `minimumHealthyPercent = 0` and  `maximumHealthyPercent = undefined` values are used.

We added a unit test validating that the correct values are passed to the task definition. We also executed using the locally built version of cdk validating that the behavior is as expected: the parameters are respected during hotswap deployments, and the existing API is maintained.

### Checklist

----


atanaspam:
Clarification Request

> ❌ Features must contain a change to an integration test file and the resulting snapshot.

I was able to execute the integration tests I added locally successfully, but there doesn't seem to be any snapshot changes for me to report.

packages/aws-cdk/lib/cli.ts
```diff
@@ -158,6 +158,17 @@ async function parseCommandLineArguments(args: string[]) {
...
       })
+      .option('hotswap-ecs-minimum-healthy-percent', {
+        type: 'number',
+        default: 0,
+        desc: 'When using hotswap for ECS, set the minimum healthy percent' +
+          'for the updated task definition',
+      })
+      .option('hotswap-ecs-maximum-healthy-percent', {
+        type: 'number',
+        desc: 'When using hotswap for ECS, set the minimum healthy percent' +
```
ahammond, line 169:
```suggestion
        desc: 'When using hotswap for ECS, set the maximum healthy percent' +
```
typo?

ahammond (approved):
This would have made it easier / more likely for us to stay on ECS instead of moving to EKS.

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
```diff
@@ -1823,6 +1823,41 @@ integTest('hotswap deployment for ecs service detects failed deployment and erro
...
+  const stackArn = await fixture.cdkDeploy('ecs-hotswap', {
+    captureStderr: false,
+  });
+
+  // WHEN
+  await fixture.cdkDeploy('ecs-hotswap', {
+    options: [
+      '--hotswap',
+      '--hotswap-ecs-minimum-healthy-percent', '100',
+      '--hotswap-ecs-maximum-healthy-percent', '200',
```
comcalvi, line 1837: This should not be a CLI option. The CLI should not understand configuration about AWS resources, because that's what the CDK application is responsible for. Allowing configuration of hotswappable resources like this is not desirable. 

Would adding an entry to `cdk.json` solve this for you? This key should be a boolean; if set, hotswap will use the value configured on the construct, and otherwise it will use the default values. 

Here's an example of adding a key to `cdk.json`: https://github.com/aws/aws-cdk/pull/17176/files.

I could see expanding more hotswap keys here, so `hotswap` should be a top-level key like `watch`, and the min and max healthy percentages should be configurable keys within the `hotswap` entry.

comcalvi (changes requested):

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/cli.integtest.ts
rexrafa: Hi @comcalvi ,

We have updated the PR, changing the implementation to get the configuration from cdk.json
Here is an example of the configuration:
`hotswap: {
      ecs: {
        minimumHealthyPercent: ecsMinimumHealthyPercent,
        maximumHealthyPercent: ecsMaximumHealthyPercent,
      },
    },`
    
    Please let me know your thoughts and if we need to change something.
    
    Our change didn't affect tests that create snapshots, but the lint is failing.
    Can you review it please and let me know what the process is.

atanaspam:
Hey folks, can I have a review here please? I'd love to merge this soon as it has been hanging around for a while

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -245,6 +245,18 @@ export class CdkToolkit {
...
+    let hotswapPropertiesFromSettings = this.props.configuration.settings.get(['hotswap']) || {};
+
+    let hotswapProperties = new HotswapProperties();
+    hotswapProperties.ecsHotswapProperties = new EcsHotswapProperties(
+      hotswapPropertiesFromSettings.ecs?.minimumHealthyPercent,
+      hotswapPropertiesFromSettings.ecs?.maximumHealthyPercent,
+    );
+
+    if (!hotswapProperties.ecsHotswapProperties.isEmpty() && options.hotswap == HotswapMode.FULL_DEPLOYMENT) {
+      warning('⚠️ Hotswap properties defined while not in hotswap mode will be ignored.');
```
comcalvi, line 257: Can we remove this warning?

packages/aws-cdk/lib/api/hotswap/common.ts
```diff
@@ -98,6 +98,52 @@ export class HotswappableChangeCandidate {
 
 type Exclude = { [key: string]: Exclude | true }
 
+/**
+ * Represents configuration properties for hotswap deployments
+ */
+export class HotswapProperties {
```
comcalvi, line 104:
```suggestion
export class HotswapPropertyOverrides {
```

comcalvi (changes requested):
Sorry for the delay on this one, this is a great change!

It's unfortunate that we chose to put the entire SDK call into the promises array, and not separated out the arguments...that would've prevented us from passing this thing throughout the entire code path, but it's fine as-is. 

Minor comments only.

packages/aws-cdk/lib/api/hotswap/common.ts
atanaspam: I have renamed the class and also all the variables as we propagate the properties throughout the code path to reflect the new name.

comcalvi (dismissed):

atanaspam:
@comcalvi @kaizencc I had to update the PR as mergify failed to merge this automatically due to what I assume is an incoming change.
https://github.com/aws/aws-cdk/pull/30511/checks?check_run_id=31046011303

Can you please review again?

comcalvi (approved):

atanaspam:
@comcalvi  Same thing again 😕 Seems like `.github/workflows/sync-from-upstream.yml` was updated in the meantime.

This time I won't update the PR myself as to not dismiss your review.

comcalvi (dismissed):

atanaspam:
@comcalvi I raised #31805 to fix the failing integ-test pipeline.

comcalvi (approved):

PR #30515, opened 2024-06-11
fix(cloudformation-include): can't use CFN intrinsics in Tags

### LINKED ISSUE #27594: core: CfnParser can't correctly parse conditional Tags
### Describe the bug

The CfnParser class (used by the CfnInclude construct) cannot correctly parse Tag declarations that conditionally include tags.  Using intrinsic functions inside the `Key` or `Value` of a Tag is fine, but if you try to declare a tag using a top-level `Fn::If` (which is allowed by CloudFormation), parsing using CfnParser fails.

### Expected Behavior

CfnParser should be able to correctly parse Tag declarations that use `Fn::If`.

### Current Behavior

Tag declarations that use top-level `Fn::If` get turned into an empty `{}` by CfnParser, which causes failure during synthesis when `{}` is rejected as an invalid Tag by [validateCfnTag()](https://github.com/aws/aws-cdk/blob/75e56ea980590fcf8860f5a7a6fba5347e0ace2d/packages/aws-cdk-lib/core/lib/runtime.ts#L252)

The error message looks like this:
```
CfnSynthesisError: Resolution error: Supplied properties not correct for "CfnLoadBalancerProps"
  tags: element 0: {} should have a 'key' and a 'value' property.
    at ValidationResult.assertSuccess (/workplace/cdkApp/node_modules/aws-cdk-lib/core/lib/runtime.js:11:1271)
    at convertCfnLoadBalancerPropsToCloudFormation (/workplace/cdkApp/node_modules/aws-cdk-lib/aws-elasticloadbalancingv2/lib/elasticloadbalancingv2.generated.js:26:80)
    at CfnLoadBalancer.renderProperties (/workplace/cdkApp/node_modules/aws-cdk-lib/aws-elasticloadbalancingv2/lib/elasticloadbalancingv2.generated.js:20:2083)
    at PostResolveToken.Resources [as processor] (/workplace/cdkApp/node_modules/aws-cdk-lib/core/lib/cfn-resource.js:1:7100)
    at PostResolveToken.postProcess (/workplace/cdkApp/node_modules/aws-cdk-lib/core/lib/util.js:1:1565)
    at Object.postProcess (/workplace/cdkApp/node_modules/aws-cdk-lib/core/lib/private/resolve.js:1:1018)
    at DefaultTokenResolver.resolveToken (/workplace/cdkApp/node_modules/aws-cdk-lib/core/lib/resolvable.js:1:1320)
    at resolve (/workplace/cdkApp/node_modules/aws-cdk-lib/core/lib/private/resolve.js:1:2510)
    at Object.resolve (/workplace/cdkApp/node_modules/aws-cdk-lib/core/lib/private/resolve.js:1:892)
    at resolve (/workplace/cdkApp/node_modules/aws-cdk-lib/core/lib/private/resolve.js:1:2787) {
  type: 'CfnSynthesisError'
```

### Reproduction Steps

Create a CFN template with a resource containing a tag specification that looks like this:
```
"Tags" : [
  {
    "Fn::If" : [
      "ShouldIncludeTag",
      {"Key" : "TagName", "Value" : "TagValue"},
      {"Ref" :  "AWS::NoValue"}
    ]
  }
]
```

If you try to include that template into a CDK application using CfnInclude, it results in a synthesis failure.

### Possible Solution

There are two places that are causing problems here.  The first is the [FromCloudFormation.getCfnTag()](https://github.com/aws/aws-cdk/blob/v2.99.1/packages/aws-cdk-lib/core/lib/helpers-internal/cfn-parse.ts#L219) function from `cfn-parse.ts`.  It assumes that a parsed tag will always be an object with `Key` and `Value` properties.  Tags declared using `Fn::If` get turned into an `IResolvable` object instead.  Adding `if (isResolvableObject(tag)) { return tag }` to `getCfnTag()` should solve that part of the problem.

The second problem is the [cfnTagToCloudFormation()](https://github.com/aws/aws-cdk/blob/75e56ea980590fcf8860f5a7a6fba5347e0ace2d/packages/aws-cdk-lib/core/lib/runtime.ts#L51) function from `runtime.ts`.  It also assumes that tags are always objects with `Key` and `Value` properties.  My recommended solution here would be to add `if (!canInspect(x)) { return x; }`, similar to what the listMapper and hashMapper functions do.

_No response_

### CDK CLI Version

2.94

### Framework Version

_No response_

### Node.js Version

18.16.0

### OS

Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

comcalvi:
### Issue # (if applicable)

Closes #27594.

### Reason for this change

Templates that use intrinsics in resource Tags cannot be used with CFN Include. 

### Description of changes

Modifed the CFN Parser to not choke on Intrinsics found in resource Tags. 

Unit tests.

### Checklist

----


TheRealAmazonKendra (changes requested):
This looks good but could you add an additional test for an intrinsic that uses other intrinsics within it or just a more complex one in general?

KyleBS:
@TheRealAmazonKendra Any chance we get this merged as-is to fix the issue for the simple case at the very least?

rix0rrr (approved):

PR #30542, opened 2024-06-13
feat(elasticloadbalancingv2): support Weighted Random algorithm and Automatic Target Weights for alb

### LINKED ISSUE #29969: (aws-elasticloadbalancingv2): Add the weighted_random load balancing algorithm
### Describe the feature

In 2023, Automatic Target Weights (ATW), which includes anomaly detection and a load balancing algorithm called Weighted Random (per [this announcement](https://aws.amazon.com/blogs/networking-and-content-delivery/improving-availability-with-application-load-balancer-automatic-target-weights/)). This feature request is about adding support for specifying the `weighted_random` algorithm as well as enable anomaly mitigation.

The CFN documentation can be found [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-targetgroup-targetgroupattribute.html).

### Use Case

Enable CDK users to make use of the new routing strategy.

### Proposed Solution

_No response_

### Other Information

References:
* https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#automatic-target-weights
* https://aws.amazon.com/blogs/networking-and-content-delivery/improving-availability-with-application-load-balancer-automatic-target-weights/
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-targetgroup-targetgroupattribute.html

### Acknowledgements

### CDK version used

2.139.0

Linux

mazyu36:
### Issue # (if applicable)

Closes #29969 

### Reason for this change
Application Load Balancer supports Weight Random routing Algorithm and Automatic Target Weights. ([link](https://aws.amazon.com/about-aws/whats-new/2023/11/application-load-balancer-availability-target-weights/))

But current L2 construct does not support it.

### Description of changes
* Add `weighted_random` algorithm to ENUM.
* Add a flag for setting Automatic Target Weights anomaly mitigation.

### Checklist

----


packages/aws-cdk-lib/aws-elasticloadbalancingv2/README.md
```diff
@@ -458,6 +458,22 @@ const tg = new elbv2.ApplicationTargetGroup(this, 'TG', {
...
+
+```ts
+declare const vpc: ec2.Vpc;
+
+const tg = new elbv2.ApplicationTargetGroup(this, 'TargetGroup', {
+  loadBalancingAlgorithmType: elbv2.TargetGroupLoadBalancingAlgorithmType.WEIGHTED_RANDOM,
+  vpc,
+  enableAnomalyMitigation: true,
+});
+```
```
lpizzinidev, line 475:
```suggestion
### Weighted random routing algorithms and automatic target weights for your Application Load Balancer

You can use the `weighted_random` routing algorithms by setting the `loadBalancingAlgorithmType` property.

When using this algorithm, Automatic Target Weights (ATW) anomaly mitigation can be used by setting `enableAnomalyMitigation` to `true`.

For more information, see [Routing algorithms](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#modify-routing-algorithm) and [Automatic Target Weights (ATW)](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#automatic-target-weights).

```ts
declare const vpc: ec2.Vpc;

const tg = new elbv2.ApplicationTargetGroup(this, 'TargetGroup', {
  vpc,
  loadBalancingAlgorithmType: elbv2.TargetGroupLoadBalancingAlgorithmType.WEIGHTED_RANDOM,
  enableAnomalyMitigation: true,
});
```
```

**Review by lpizzinidev — CHANGES_REQUESTED** [review]
Thanks 👍 Left comments for minor adjustments

**Comment by mazyu36** [comment]
@lpizzinidev 
Thank you for the review. I have incorporated your suggestions and made the modifications.

File: packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-target-group.ts [hunk 2]
```diff
@@ -367,13 +367,11 @@ export class ApplicationTargetGroup extends TargetGroupBase implements IApplicat
       this.addTarget(...(props.targets || []));
 
       if (props.enableAnomalyMitigation !== undefined) {
-        if (props.enableAnomalyMitigation && isWeightedRandomAlgorithm) {
+        if (props.enableAnomalyMitigation && !isWeightedRandomAlgorithm) {
```
mazyu36, line 370: I want to throw an error when anomaly mitigation is enabled and the algorithm is not` weighted_random`, so I added a '!' before `isWeightedRandomAlgorithm`.

lpizzinidev (approved):
Thanks 👍

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-target-group.ts
paulhcsun: The codechanges make sense to me but the only thing I'm not so keen on is enforcing the restrictions on the properties:
- `weighted_random` algorithm cannot be used in slow start mode
- anomaly mitigation can only be enabled when the `weighted_random` algorithm is used 

with error handling in the constructor. For this current case it works fine because we're not adding much but this can slowly grow over time into a wall of error handling checks as more properties with restrictions are added.

What I had in mind here was to instead create another version of this class that specifically has the `weighted_random` algorithm set by default and does not allow the user to set other conflicting properties to remove the need for these error checks.

i.e. A class like `ApplicationTargetGroupWeightedRandomRouting` or something where the anomaly mitigation can be enabled or not without the need for error handling and slow start mode is disabled and cannot be enabled.

Although part of me worries that this may be a bit confusing as a DX to use because we're not providing the option through `ApplicationTargetGroup` and I'm not sure how many other routing algorithm options there are that we may not have done this for. What do you think?

paulhcsun (changes requested):
Hi @mazyu36, thank you for your contribution! I just left one comment regarding the error checks in the constructor to enforce a couple of the restrictions of using the `weighted_random` algorithm. Let me know what you think!

packages/aws-cdk-lib/aws-elasticloadbalancingv2/README.md
```diff
@@ -458,6 +458,24 @@ const tg = new elbv2.ApplicationTargetGroup(this, 'TG', {
 });
 ```
 
+### Weighted random routing algorithms and automatic target weights for your Application Load Balancer
```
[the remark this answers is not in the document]
mazyu36: Thanks. I've updated docs.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-target-group.ts
mazyu36: Thank you for your suggestion. 

Personally, I like the current consolidated form of ApplicationTargetGroup.

* While there are other routing algorithms such as Least outstanding requests, I think having only weighted random as a separate class would be confusing for users. 
* Additionally, I think that the current restriction is not complex enough to warrant a separate class.
* There are common items with other routing algorithms such as stickiness.enabled, and it's easier to understand when they are grouped together.

For the reasons mentioned above, I personally think the current format is preferable. 
However, as you've expressed concern, your suggestion might be better considering the possibility of increased constraints in the future. 
But honestly, I'm not sure if they will increase...

Given the above, I'd appreciate your thoughts on this.

mazyu36:
@paulhcsun 
Thank you for the review !

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-target-group.ts
paulhcsun: Thanks for sharing your thoughts. 

After hearing your reasonings I agree that separating out a new class just for this algorithm is probably not necessary. As you mentioned, the restriction is not super complex so I'm okay to accept this change as is. 

Thanks @mazyu36!

paulhcsun (approved):
Nice work @mazyu36!

And thanks for the review as always @lpizzinidev!

Issue #28152, opened 2023-11-27
aws_synthetics: The timeout from the RunConfig was missed so there is no way to change it


ignaciolarranaga:
### Describe the feature

Add the timeout, here is the reference:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-synthetics-canary-runconfig.html

As you can see here https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_synthetics.Canary.html it is not there.

### Use Case

Se the timeout

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.110.0

### Environment details (OS name and version, etc.)

NA

pahud:
Yes we can expose that in the construct prop and before that we can always use [escape hatches](https://docs.aws.amazon.com/cdk/v2/guide/cfn_layer.html) to addPropertyOverride on it.

mazyu36:
Hi.
This issue has been addressed in #30556  and released in [v2.151.0](https://github.com/aws/aws-cdk/releases/tag/v2.151.0).
I apologize that it wasn't automatically closed, possibly due to an error in my settings.

PR #30556, opened 2024-06-15
feat(synthetics): add activeTracing, memory and timeout property to Canary class

### LINKED ISSUE #9300: [synthetics] support runConfig properties
### Issue

AWS Synthetics Canary has an optional `runConfig` [property](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-synthetics-canary-runconfig.html) that is currently not supported by the Canary L2. This property exposes `timeoutInSeconds` and `memoryInMb`. 

The difficulty is that `timeoutInSeconds` is required if `runConfig` is set, so one cannot **only** specify `memoryInMb`. 

This does not seem like it should be a hard requirement (i.e. one should be able to set `memoryInMb` without `timeout`).

The requirements for `timeoutInSeconds` are that it is inside the range 3 - 840 seconds* and not greater than the amount specified in the `schedule` expression. 

&nbsp;

\* yes, this is different than specified in the docs but I tried to set it to be 900 seconds and received the error "Invalid request provided: Run timeout cannot be less than 3 seconds or greater than 840 seconds"

### Use Case

The use case for this property is unclear so this issue serves as a platform for discussion on whether it is needed and how to implement.

While the proposed solution is possible, this seems like extra work if the `runConfig` property is largely ignored anyway. And if `timeoutInSeconds` becomes optional, then we no longer need to find a default for the property.

### Proposed Solution

Add separate optional properties `timeout: cdk.Duration` and `memorySize: cdk.Size` to `CanaryProps`.

`timeout` should default to the minimum of 840 seconds and the amount in `expression`. This would require parsing through the expression string to extract the number/unit and convert to seconds.

`memorySize` is optional in `runConfig`, so no need to default and it is up to the user to specify a valid amount of memory.

* [x] :wave: I may be able to implement this feature request

---

### LINKED ISSUE #14086: (synthetics): add tracing
Enable active tracing (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-synthetics-canary-runconfig.html#cfn-synthetics-canary-runconfig-activetracing) for a canary. 

### Use Case

End-to-end tracing from Canary to API Gateway to Lambda.

### Proposed Solution

Add a new property to `synthetics.CanaryProps`: 

```
new synthetics.Canary(this, 'MyCanary', {
  ...,
  activeTracing: true,
})
```

By default, it will be `false`. When set to true, it will also append XRay permissions to the IAM role (e.g. the managed policy `arn:aws:iam::aws:policy/AWSXrayWriteOnlyAccess`). 

### Other

None.

* [X] :wave: I may be able to implement this feature request

---

mazyu36:
### Issue # (if applicable)

Closes #9300, #14086, #28152

### Reason for this change
In order to make the properties in the [runConfig](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_synthetics.CfnCanary.RunConfigProperty.html) of the L1 Construct configurable in the L2 Construct.

* [activeTracing](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_synthetics.CfnCanary.RunConfigProperty.html#activetracing)
* [memoryInMb](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_synthetics.CfnCanary.RunConfigProperty.html#memoryinmb)
* [timeoutInSeconds](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_synthetics.CfnCanary.RunConfigProperty.html#timeoutinseconds)

At the moment, the L2 [Canary](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_synthetics.Canary.html) Class only supports [environmentVariables](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_synthetics.CfnCanary.RunConfigProperty.html#environmentvariables) in the `runConfig`.

### Description of changes
Add missing properties.

* `activeTracing`
* `memory`
* `timeout`

Note:  The following is stated in #9300, but when tested, it was possible to set only `memoryInMb` (`memory`).

> The difficulty is that timeoutInSeconds is required if runConfig is set, so one cannot only specify memoryInMb.

Test code is here.

```ts
class TestStack extends Stack {
    super(scope, id, props);

    new synthetics.Canary(this, 'Canary', {
      canaryName: 'next',
      runtime: synthetics.Runtime.SYNTHETICS_NODEJS_PUPPETEER_7_0,
      test: synthetics.Test.custom({
        handler: 'index.handler',
        code: synthetics.Code.fromInline(`
          exports.handler = async () => {
            console.log(\'hello world\');
          };`),
      }),
      cleanup: synthetics.Cleanup.LAMBDA,
      memory: Size.mebibytes(2048),
    });
  }
}
```

### Checklist

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-events/test/integ.eventbus.js.snapshot/Stack.assets.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-events/test/integ.eventbus.js.snapshot/Stack.assets.json
@@ -1,7 +1,7 @@
 {
   "version": "36.0.0",
   "files": {
-    "1ac02a396bc49bcce2ab3c276dd4e234efe7e86f9d1a06c9ae16c054f1764bf1": {
+    "2671a34b39f9d75f10baaa810e7901cbcaa8901ff9852aa7a49bf43a87af0155": {
       "source": {
         "path": "Stack.template.json",
         "packaging": "file"
@@ -9,7 +9,7 @@
       "destinations": {
         "current_account-current_region": {
           "bucketName": "cdk-hnb659fds-assets-${AWS::AccountId}-${AWS::Region}",
-          "objectKey": "1ac02a396bc49bcce2ab3c276dd4e234efe7e86f9d1a06c9ae16c054f1764bf1.json",
+          "objectKey": "2671a34b39f9d75f10baaa810e7901cbcaa8901ff9852aa7a49bf43a87af0155.json",
           "assumeRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-file-publishing-role-${AWS::AccountId}-${AWS::Region}"
         }
       }
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-events/test/integ.eventbus.js.snapshot/Stack.template.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-events/test/integ.eventbus.js.snapshot/Stack.template.json
@@ -3,6 +3,7 @@
   "BusEA82B648": {
    "Type": "AWS::Events::EventBus",
    "Properties": {
+    "Description": "myEventBus",
     "Name": "StackBusAA0A1E4B"
    }
   },
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-events/test/integ.eventbus.js.snapshot/manifest.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-events/test/integ.eventbus.js.snapshot/manifest.json
@@ -18,7 +18,7 @@
         "validateOnSynth": false,
         "assumeRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-deploy-role-${AWS::AccountId}-${AWS::Region}",
         "cloudFormationExecutionRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-cfn-exec-role-${AWS::AccountId}-${AWS::Region}",
-        "stackTemplateAssetObjectUrl": "s3://cdk-hnb659fds-assets-${AWS::AccountId}-${AWS::Region}/1ac02a396bc49bcce2ab3c276dd4e234efe7e86f9d1a06c9ae16c054f1764bf1.json",
+        "stackTemplateAssetObjectUrl": "s3://cdk-hnb659fds-assets-${AWS::AccountId}-${AWS::Region}/2671a34b39f9d75f10baaa810e7901cbcaa8901ff9852aa7a49bf43a87af0155.json",
         "requiresBootstrapStackVersion": 6,
         "bootstrapStackVersionSsmParameter": "/cdk-bootstrap/hnb659fds/version",
         "additionalDependencies": [
@@ -62,24 +62,6 @@
           {
             "type": "aws:cdk:logicalId",
             "data": "CheckBootstrapVersion"
-          }
-        ],
-        "BusStatement1B4D0336C": [
-          {
-            "type": "aws:cdk:logicalId",
-            "data": "BusStatement1B4D0336C",
-            "trace": [
-              "!!DESTRUCTIVE_CHANGES: WILL_DESTROY"
-            ]
-          }
-        ],
-        "BusStatement2B5FB314B": [
-          {
-            "type": "aws:cdk:logicalId",
-            "data": "BusStatement2B5FB314B",
-            "trace": [
-              "!!DESTRUCTIVE_CHANGES: WILL_DESTROY"
-            ]
           }
         ]
       },
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-events/test/integ.eventbus.js.snapshot/tree.json
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-events/test/integ.eventbus.js.snapshot/tree.json
@@ -18,12 +18,13 @@
                 "attributes": {
                   "aws:cdk:cloudformation:type": "AWS::Events::EventBus",
                   "aws:cdk:cloudformation:props": {
+                    "description": "myEventBus",
                     "name": "StackBusAA0A1E4B"
                   }
                 },
                 "constructInfo": {
-                  "fqn": "aws-cdk-lib.aws_events.CfnEventBus",
-                  "version": "0.0.0"
+                  "fqn": "constructs.Construct",
+                  "version": "10.3.0"
                 }
               },
               "cdk-Statement1": {
@@ -72,14 +73,14 @@
                       }
                     },
                     "constructInfo": {
-                      "fqn": "aws-cdk-lib.aws_events.CfnEventBusPolicy",
-                      "version": "0.0.0"
+                      "fqn": "constructs.Construct",
+                      "version": "10.3.0"
                     }
                   }
                 },
                 "constructInfo": {
-                  "fqn": "aws-cdk-lib.aws_events.EventBusPolicy",
-                  "version": "0.0.0"
+                  "fqn": "constructs.Construct",
+                  "version": "10.3.0"
                 }
               },
               "cdk-Statement2": {
@@ -128,42 +129,42 @@
                       }
                     },
                     "constructInfo": {
-                      "fqn": "aws-cdk-lib.aws_events.CfnEventBusPolicy",
-                      "version": "0.0.0"
+                      "fqn": "constructs.Construct",
+                      "version": "10.3.0"
                     }
                   }
                 },
                 "constructInfo": {
-                  "fqn": "aws-cdk-lib.aws_events.EventBusPolicy",
-                  "version": "0.0.0"
+                  "fqn": "constructs.Construct",
+                  "version": "10.3.0"
                 }
               }
             },
             "constructInfo": {
-              "fqn": "aws-cdk-lib.aws_events.EventBus",
-              "version": "0.0.0"
+              "fqn": "constructs.Construct",
+              "version": "10.3.0"
             }
           },
           "BootstrapVersion": {
             "id": "BootstrapVersion",
             "path": "Stack/BootstrapVersion",
             "constructInfo": {
-              "fqn": "aws-cdk-lib.CfnParameter",
-              "version": "0.0.0"
+              "fqn": "constructs.Construct",
+              "version": "10.3.0"
             }
           },
           "CheckBootstrapVersion": {
             "id": "CheckBootstrapVersion",
             "path": "Stack/CheckBootstrapVersion",
             "constructInfo": {
-              "fqn": "aws-cdk-lib.CfnRule",
-              "version": "0.0.0"
+              "fqn": "constructs.Construct",
+              "version": "10.3.0"
             }
           }
         },
         "constructInfo": {
-          "fqn": "aws-cdk-lib.Stack",
-          "version": "0.0.0"
+          "fqn": "constructs.Construct",
+          "version": "10.3.0"
         }
       },
       "IntegTest-EventBusStack": {
@@ -190,22 +191,22 @@
                     "id": "BootstrapVersion",
                     "path": "IntegTest-EventBusStack/DefaultTest/DeployAssert/BootstrapVersion",
                     "constructInfo": {
-                      "fqn": "aws-cdk-lib.CfnParameter",
-                      "version": "0.0.0"
+                      "fqn": "constructs.Construct",
+                      "version": "10.3.0"
                     }
                   },
                   "CheckBootstrapVersion": {
                     "id": "CheckBootstrapVersion",
                     "path": "IntegTest-EventBusStack/DefaultTest/DeployAssert/CheckBootstrapVersion",
                     "constructInfo": {
-                      "fqn": "aws-cdk-lib.CfnRule",
-                      "version": "0.0.0"
+                      "fqn": "constructs.Construct",
+                      "version": "10.3.0"
                     }
                   }
                 },
                 "constructInfo": {
-                  "fqn": "aws-cdk-lib.Stack",
-                  "version": "0.0.0"
+                  "fqn": "constructs.Construct",
+                  "version": "10.3.0"
                 }
               }
             },
@@ -230,8 +231,8 @@
       }
     },
     "constructInfo": {
-      "fqn": "aws-cdk-lib.App",
-      "version": "0.0.0"
+      "fqn": "constructs.Construct",
+      "version": "10.3.0"
     }
   }
 }
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-events/test/integ.eventbus.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-events/test/integ.eventbus.ts
@@ -5,7 +5,9 @@ import { EventBus } from 'aws-cdk-lib/aws-events';
 
 const app = new App();
 const stack = new Stack(app, 'Stack');
-const bus = new EventBus(stack, 'Bus');
+const bus = new EventBus(stack, 'Bus', {
+  description: 'myEventBus',
+});
 
 bus.addToResourcePolicy(new iam.PolicyStatement({
   effect: iam.Effect.ALLOW,
--- a/packages/aws-cdk-lib/aws-events/README.md
+++ b/packages/aws-cdk-lib/aws-events/README.md
@@ -208,7 +208,8 @@ It is possible to archive all or some events sent to an event bus. It is then po
 
 ```ts
 const bus = new events.EventBus(this, 'bus', {
-  eventBusName: 'MyCustomEventBus'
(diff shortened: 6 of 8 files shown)


packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -513,11 +543,45 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
 
   private createRunConfig(props: CanaryProps): CfnCanary.RunConfigProperty | undefined {
-    if (!props.environmentVariables) {
+    if (props.activeTracing === undefined &&
+      !props.environmentVariables &&
+      !props.memory &&
+      !props.timeout) {
       return undefined;
     }
+    const activeTracingNotSupportedRuntime = [
```
mazyu36, line 552: The error messages are taken directly from the [documentation](https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-properties-synthetics-canary-runconfig.html#cfn-synthetics-canary-runconfig-activetracing).
However, since syn-nodejs-2.0 and earlier versions are no longer available, I am only setting it for the currently available runtimes that do not support active tracing (see also [docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_Canaries_Library.html#CloudWatch_Synthetics_Canaries_runtime_support)).

I was unsure about including this validation, so if you have any thoughts or feedback, please let me know.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -513,11 +543,45 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
+      if (memoryInMb % 64 !== 0) {
+        throw new Error(`\`memory\` must be a multiple of 64 MiB, got ${memoryInMb} MiB.`);
+      }
+      if (memoryInMb < 960 || memoryInMb > 3008) {
+        throw new Error(`\`memory\` must be between 960 MiB and 3008 MiB, got ${memoryInMb} MiB.`);
+      }
+    }
+
+    let timeoutInSeconds = undefined;
+    if (!cdk.Token.isUnresolved(props.timeout) && props.timeout) {
```
mazyu36, line 573: According to the [documentation](https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-properties-synthetics-canary-runconfig.html#cfn-synthetics-canary-runconfig-timeoutinseconds), the property can't set this time to be longer than the frequency of the runs of this canary.

However, there are various ways to set the `schedule` property, and I determined that implementing this validation would be difficult. Therefore, I decided not to implement it. If you have any better suggestions or methods, please let me know.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -179,6 +179,20 @@ export interface CanaryProps {
    */
   readonly test: Test;
 
+  /**
+   * Specifies whether this canary is to use active AWS X-Ray tracing when it runs.
+   * Active tracing enables this canary run to be displayed in the ServiceLens and X-Ray service maps even if the
+   * canary does not hit an endpoint that has X-Ray tracing enabled. Using X-Ray tracing incurs charges.
+   *
+   * You can enable active tracing only for canaries that use version syn-nodejs-2.0 or later for their canary runtime.
```
lpizzinidev, line 187:
```suggestion
   * You can enable active tracing only for canaries that use version `syn-nodejs-2.0` or later for their canary runtime.
```

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -179,6 +179,20 @@ export interface CanaryProps {
...
+   *
+   * @default false
+   *
+   * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_Canaries_tracing.html
+   */
+
```
lpizzinidev, line 193:
```suggestion
```
Extra space

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -188,6 +202,22 @@ export interface CanaryProps {
    */
   readonly environmentVariables?: { [key: string]: string };
 
+  /**
+   * The maximum amount of memory that the canary can use while running.
+   * This value must be a multiple of 64 Mib.
```
lpizzinidev, line 207:
```suggestion
   * This value must be a multiple of 64 Mib.
   * The range is 960 MiB to 3008 MiB.
```

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -188,6 +202,22 @@ export interface CanaryProps {
...
+   *
+   * @default Size.mebibytes(5)
+   */
+  readonly memory?: cdk.Size;
+
+  /**
+   * How long the canary is allowed to run before it must stop.
+   * You can't set this time to be longer than the frequency of the runs of this canary.
+   *
```
lpizzinidev, line 216:
```suggestion
   *
   * The minimum allowed value is 3 seconds.
   * The maximum allowed value is 840 seconds (14 minutes).
   *
```

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -513,11 +543,45 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
+      !props.timeout) {
       return undefined;
     }
+    const activeTracingNotSupportedRuntime = [
+      Runtime.SYNTHETICS_PYTHON_SELENIUM_2_1,
+      Runtime.SYNTHETICS_PYTHON_SELENIUM_3_0,
+    ];
+
+    if (props.activeTracing && activeTracingNotSupportedRuntime.includes(props.runtime)) {
+      throw new Error('You can enable active tracing only for canaries that use version syn-nodejs-2.0 or later for their canary runtime.');
```
lpizzinidev, line 558:
```suggestion
      throw new Error('You can enable active tracing only for canaries that use version `syn-nodejs-2.0` or later for their canary runtime.');
```

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -513,11 +543,45 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
+    }
+
+    let memoryInMb = undefined;
```
lpizzinidev, line 561:
```suggestion
    let memoryInMb: number | undefined;
```
Typing.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -513,11 +543,45 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
+    if (!cdk.Token.isUnresolved(props.memory) && props.memory) {
```
lpizzinidev, line 562:
```suggestion
    if (!cdk.Token.isUnresolved(props.memory) && props.memory !== undefined) {
```
Could be `0`.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -513,11 +543,45 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
+      memoryInMb = props.memory.toMebibytes();
+      if (memoryInMb % 64 !== 0) {
+        throw new Error(`\`memory\` must be a multiple of 64 MiB, got ${memoryInMb} MiB.`);
+      }
+      if (memoryInMb < 960 || memoryInMb > 3008) {
+        throw new Error(`\`memory\` must be between 960 MiB and 3008 MiB, got ${memoryInMb} MiB.`);
+      }
+    }
+
+    let timeoutInSeconds = undefined;
```
lpizzinidev, line 572:
```suggestion
    let timeoutInSeconds: number | undefined;
```
Typing.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
lpizzinidev, line 573:
```suggestion
    if (!cdk.Token.isUnresolved(props.timeout) && props.timeout !== undefined) {
```
Could be `0`

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -513,11 +543,45 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
+    if (!cdk.Token.isUnresolved(props.timeout) && props.timeout) {
+      timeoutInSeconds = props.timeout.toSeconds();
```
lpizzinidev, line 574:
```suggestion
      const timeoutError = '`timeout` must be between 3 seconds and 840 seconds';
      const timeoutInMillis = props.timeout.toMilliseconds();
      if (timeoutInMillis > 0 && timeoutInMillis < Duration.seconds(3).toMilliseconds()) {
        throw new Error(`${timeoutError}, got ${timeoutInMillis} milliseconds.`);
      }
      timeoutInSeconds = props.timeout.toSeconds();
```
Handle exceptions when the user specifies timeout between 0 and 1 sec.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -188,6 +202,22 @@ export interface CanaryProps {
    */
   readonly environmentVariables?: { [key: string]: string };
 
+  /**
+   * The maximum amount of memory that the canary can use while running.
+   * This value must be a multiple of 64 Mib.
+   *
+   * @default Size.mebibytes(5)
```
lpizzinidev, line 209: Are you sure about this default? 🤔

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
lpizzinidev: What about checking `props.runtime.family !== RuntimeFamily.NODEJS` like in the other [PR](https://github.com/aws/aws-cdk/pull/30197)?

packages/aws-cdk-lib/aws-synthetics/README.md
```diff
@@ -87,6 +87,62 @@ const schedule = synthetics.Schedule.cron({
 
 If you want the canary to run just once upon deployment, you can use `Schedule.once()`.
 
+### Active Tracing
+
+You can choose to enable active AWS X-Ray tracing on canaries that use the syn-nodejs-2.0 or later runtime by setting `activeTracing` to `true`.
```
lpizzinidev, line 92:
```suggestion
You can choose to enable active AWS X-Ray tracing on canaries that use the `syn-nodejs-2.0` or later runtime by setting `activeTracing` to `true`.
```

packages/aws-cdk-lib/aws-synthetics/README.md
```diff
@@ -87,6 +87,62 @@ const schedule = synthetics.Schedule.cron({
...
+    handler: 'index.handler',
+  }),
+  runtime: synthetics.Runtime.SYNTHETICS_NODEJS_PUPPETEER_6_2,
+  activeTracing: true, // active tracing enabled
+});
+```
+
+### Memory
+
+You can set the maximum amount of memory that the canary can use while running by setting `memory`.
```
lpizzinidev, line 112:
```suggestion
You can set the maximum amount of memory the canary can use while running with the `memory` property.
```

packages/aws-cdk-lib/aws-synthetics/README.md
```diff
@@ -87,6 +87,62 @@ const schedule = synthetics.Schedule.cron({
...
+    handler: 'index.handler',
+  }),
+  runtime: synthetics.Runtime.SYNTHETICS_NODEJS_PUPPETEER_6_2,
+  memory: cdk.Size.mebibytes(1024), // 1024 MiB
+});
+```
+
+### Timeout
+
+You can set how long the canary is allowed to run before it must stop by setting `timeout`.
```
lpizzinidev, line 130:
```suggestion
You can set how long the canary is allowed to run before it must stop with the `timeout` property.
```

lpizzinidev (changes requested):
Thanks 👍 Left some comments for adjustments

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
mazyu36: Thank you. I made the correction since checking the runtime is better. 
(Even though I did it myself in another PR...lol)

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
mazyu36: @lpizzinidev 
Thank you. However, I did not quite understand the need for this validation. 
Could you please explain it in more detail?

From my testing, millisecond values that cannot be converted to whole seconds (e.g. 100ms, 3100ms, etc.) resulted in an error with the following validation:

https://github.com/aws/aws-cdk/blob/4e9ef15a0cf8b1bc664b5d22b34af9eaaa17bec8/packages/aws-cdk-lib/core/lib/duration.ts#L326

Therefore, millisecond values between 0-1 second would also result in an error with the above validation. 
As such, I considered that an additional check is unnecessary. What do you think?

mazyu36:
@lpizzinidev 
Thank you for always reviewing. I have addressed all comments.

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
lpizzinidev: The point of adding the validation is to handle the error on the construct and provide a meaningful error message to the user.

lpizzinidev:
@mazyu36 Thanks for the changes 👍 Answered https://github.com/aws/aws-cdk/pull/30556/files#r1643336431

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
mazyu36: Thank you. I added validation to throw an error if the value is not an integer representing seconds. 
How about this? 
With this change, it will also throw an error if a value between 0 and 1 second is set.

mazyu36:
@lpizzinidev 
Thanks.
I've added a validation, and answered https://github.com/aws/aws-cdk/pull/30556/files#r1643641842

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -513,11 +546,47 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
+        throw new Error(`\`memory\` must be a multiple of 64 MiB, got ${memoryInMb} MiB.`);
+      }
+      if (memoryInMb < 960 || memoryInMb > 3008) {
+        throw new Error(`\`memory\` must be between 960 MiB and 3008 MiB, got ${memoryInMb} MiB.`);
+      }
+    }
+
+    let timeoutInSeconds: number | undefined;
+    if (!cdk.Token.isUnresolved(props.timeout) && props.timeout !== undefined) {
+      const timeoutInMillis = props.timeout?.toMilliseconds();
```
lpizzinidev, line 574:
```suggestion
      const timeoutInMillis = props.timeout.toMilliseconds();
```

lpizzinidev (approved):
Nice 👍

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
```diff
@@ -513,11 +546,47 @@ export class Canary extends cdk.Resource implements ec2.IConnectable {
...
+    if (props.activeTracing === undefined &&
+      !props.environmentVariables &&
+      !props.memory &&
+      !props.timeout) {
       return undefined;
     }
+
+    // Only check runtime family is nodejs because versions prior to syn-nodejs-2.0 are deprecated and can no longer be configured.
+    if (props.activeTracing && !cdk.Token.isUnresolved(props.runtime.family) && props.runtime.family !== RuntimeFamily.NODEJS) {
+      throw new Error('You can enable active tracing only for canaries that use version `syn-nodejs-2.0` or later for their canary runtime.');
```
paulhcsun, line 558:
```suggestion
      throw new Error('You can only enable active tracing for canaries that use canary runtime version `syn-nodejs-2.0` or later.');
```

paulhcsun (commented):
Nice work @mazyu36! Thanks for the contribution!

packages/aws-cdk-lib/aws-synthetics/lib/canary.ts
mazyu36: @paulhcsun 
Thanks.
I've updated error message and unit tests.

paulhcsun (approved):

paulhcsun:
@mergifyio update

PR #30557, opened 2024-06-15
feat: add service timeout to custom resources

### LINKED ISSUE #30517: custom-resource: Support of ServiceTimeout from Cloudformation
### Describe the feature

Hi, everyone.

Cloudformation just release ServiceTimeout attribute support in Custom Resources, so now we do not need to wait 1h to notice faulty deployment. 

Announcement: https://aws.amazon.com/about-aws/whats-new/2024/06/aws-cloudformation-dev-test-cycle-timeouts-custom-resources/

Docs: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html#response-timeout

### Use Case

Mostly development cycle should be improved when you actively develop custom resources. I will be able to set timeout to 5 min max and see it fail faster.

### Proposed Solution

New optional property added to custom resource construct (not sure to which part, most likely it is framework)

### Other Information

_No response_

### Acknowledgements

### CDK version used

all

AL2

prazian:
### Issue # (if applicable)

Closes #30517.

### Reason for this change

CloudFormation recently added the `ServiceTimeout` property to adjustable timeouts for custom resources:
https://aws.amazon.com/about-aws/whats-new/2024/06/aws-cloudformation-dev-test-cycle-timeouts-custom-resources/

The Cfn documentation can be found [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudformation-customresource.html).

Currently, we have the [timeout](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.custom_resources.AwsCustomResource.html#timeout) field for the lambda function that creates the resource (here is the [cfn docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html#response-timeout)), but this new field sets timeout on a higher level (I guess).

### Description of changes

Even though now it is possible to pass PascalCase properties, adding an explicit property would be more intuitive.

A new test called "custom timeout" has been added to check the availability of the ServiceTimeout property with the correct value.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/custom-resources/test/aws-custom-resource/integ.aws-custom-resource-vpc.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/custom-resources/test/aws-custom-resource/integ.aws-custom-resource-vpc.ts
@@ -26,6 +26,7 @@ new AwsCustomResource(stack, 'DescribeVpcAttribute', {
   },
   policy: AwsCustomResourcePolicy.fromSdkCalls({ resources: AwsCustomResourcePolicy.ANY_RESOURCE }),
   timeout: cdk.Duration.minutes(3),
+  serviceTimeout: cdk.Duration.minutes(15),
   vpc: vpc,
   vpcSubnets: { subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS },
 });
--- a/packages/aws-cdk-lib/core/lib/custom-resource.ts
+++ b/packages/aws-cdk-lib/core/lib/custom-resource.ts
@@ -1,5 +1,6 @@
 import { Construct } from 'constructs';
 import { CfnResource } from './cfn-resource';
+import { Duration } from './duration';
 import { RemovalPolicy } from './removal-policy';
 import { Resource } from './resource';
 import { Token } from './token';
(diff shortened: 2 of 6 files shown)


prazian:
If this is the way to go, please let me know if:
- I need to create a new integration test or can I can just for example add this field to the VPC integration test?
- A maintainer needs to run the snapshot test (with credentials to your test account), right?
- Which docs should be updated?

GavinZZ:
Hi @prazian, thanks for making contribution to CDK repo. 
1. You will need to create an integration test following this guide https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md#L1. Which VPC test are you talking about? Would you be able to share it so I can help you decide if it's a good place to add test?
2. You'll need to deploy the newly created integration test. Once it deploys successfully, it will save a snapshot of the Cloud Assembly to local and you can commit that along with your changes. More detail can be found in the guide provided
3. Generally each module you worked on has a corresponding README file where you can update the doc. In this case, I think this is a pretty good place https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/README.md#custom-resources

PR #30558, opened 2024-06-15
chore(ec2): choose NAT instance v2 machineImage cpuType automatically from instanceType

tmokmss:
### Issue # (if applicable)
n/a

### Reason for this change

When configuring NAT instance v2, currently we have to set machineImage manually when we want to use a graviton instance.

Like this:

```ts
const vpc = new Vpc(this, 'Vpc', {
  natGatewayProvider: NatProvider.instanceV2({
    instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.NANO),
    // we should be able to omit this line!
    machineImage: MachineImage.latestAmazonLinux2023({ cpuType: AmazonLinuxCpuType.ARM_64 }),
  }),
});
```

This can be easily avoided if Nat instance v2 construct decides which cpu type to use for the given instance type.

### Description of changes

Use `instanceType.architecture` to choose cpu type of a machine image.

Now we can remove the redundant code:

```ts
const vpc = new Vpc(this, 'Vpc', {
  natGatewayProvider: NatProvider.instanceV2({
    instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.NANO),
  }),
});
```

Added an integ test.
### Checklist

----


GavinZZ (changes requested):
I agree this is something we should have thought of on the intial implementation. But updating it now would potentially be a breaking change due to x86 and arm compatibilities issues. Correct me if wrong but existing users who specified arm_64 instance type are actually using x86 type. Now if we allow this change, their instance is going to be updated with arm_64 instance type and may be breaking change. 

If this is correct, I would recommend you putting this fix under feature flag.

GavinZZ:
> @GavinZZ Fortunately, no, it is not a breaking change, because we cannot deploy arm instance with x86 AMIs due to the mismatch of architecture. We get the following error during a deployment:
> 
> > Resource handler returned message: "The architecture 'arm64' of the specified instance type does not match the architecture 'x86_64' of the specified AMI. Specify an instance type and an AMI
> > that have matching architectures, and try again. You can use 'describe-instance-types' or 'describe-images' to discover the architecture of the instance type or AMI. (Service: Ec2, Status Co
> > de: 400,

Thanks for clarification,I'm good to approve in this case.

GavinZZ (approved):

PR #30563, opened 2024-06-16
feat(redshift): supports excludeCharacters settings for DatabaseSecret

### LINKED ISSUE #26847: aws-redshift-alpha: remove backtick ` from the user passwords
### Describe the feature

As per https://docs.aws.amazon.com/cdk/api/v2/docs/aws-redshift-alpha-readme.html#creating-users

> The user password is generated by AWS Secrets Manager using the default configuration found in [secretsmanager.SecretStringGenerator](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-secretsmanager.SecretStringGenerator.html), except with password length 30 and some SQL-incompliant characters excluded. The plaintext for the password will never be present in the CDK application; instead, a [CloudFormation Dynamic Reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html) will be used wherever the password value is required.

Could you also please exclude backtick ` or expose the secret configuration?

### Use Case

When integrating with DBT airflow dags that use redshift, we have a generated password with backtick.  The password has to be written as an environment variable to keep it out of dbt profiles yaml.  The backtick are problematic.  This had to be fixed manually.

### Proposed Solution

Simplest, add more characters to the list excluded from the password

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.85.0 (build 4e0d726)

Mac Os

mazyu36:
### Issue # (if applicable)

Closes #26847.

### Reason for this change
In the case of passwords generated by [DatabaseSecret](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-redshift-alpha.DatabaseSecret.html), there may be a need to exclude certain characters. 

The original issue was to exclude the backtick character from passwords. 
However, the current default value of `excludeCharacters`, `'"@/\ ''`, matches the characters that are not supported in Redshift ([docs](https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_USER.html#r_CREATE_USER-parameters)).

> It can use any ASCII characters with ASCII codes 33–126, except ' (single quotation mark), " (double quotation mark), \, /, or @.

Instead of including the backtick in the default value of `excludeCharacters`, it was considered appropriate to make it configurable.

### Description of changes
Add `excludeCharacters` property to specify characters to not include in generated passwords.

### Checklist

----


packages/@aws-cdk/aws-redshift-alpha/test/integ.cluster-exclude-characters.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+const app = new App();
+
+new integ.IntegTest(app, 'ExcludeCharactersInteg', {
+  testCases: [new RedshiftEnv(app, 'redshift-exclude-characters-integ')],
+});
+
+app.synth();
```
badmintoncryer, line 45: nit: it is not essential
```suggestion
```

badmintoncryer (approved):
Thank you for your PR! I added only a nit comment.

packages/@aws-cdk/aws-redshift-alpha/test/integ.cluster-exclude-characters.ts
mazyu36: Thanks.
I've removed.

sumupitchayan (approved):

PR #30567, opened 2024-06-17
feat(cloudfront): add `attachWebAclId` method for Distribution

go-to-k:
### Reason for this change

I often create a custom construct for a WAF only. I also create resources (such as API Gateway, ALB, etc...) that attach the WAF in separate constructs. Instead of attaching the WAF in the target resource's construct, I create a method for attaching it in the WAF's construct.

In this way, the constructs can be loosely coupled, and the target resource's constructs can be more simply. The WAF can also be attached to multiple resources at once later.

```ts
export class Waf extends Construct {
  public readonly webAcl: CfnWebACL;

  constructor(scope: Construct, id: string, props: WafProps) {
    super(scope, id);
    this.webAcl = new CfnWebACL(this, 'WebAcl', {
      // ...
    },
  });

  public attachToRegionalResource(id: string, resourceArn: string) {
    new CfnWebACLAssociation(this, `${id}Association`, {
      resourceArn: resourceArn,
      webAclArn: this.webAcl.attrArn,
    });
  }
}

const waf = new Waf(this, 'Waf', { /* props */ });
waf.attachToRegionalResource('A', resourceA);
waf.attachToRegionalResource('B', resourceB);
waf.attachToRegionalResource('C', resourceC);
```

However, when attaching a WAF to a CloudFront, the WAF attaching configuration needs to be defined through CloudFront props, rather than using CfnWebACLAssociation.
To do this with the above WAF construct, a method is needed to pass a pre-defined CloudFront and override the properties of that definition with an escape hatch. This is a bit complicated.

```ts
  public attachToCloudFront(distribution: Distribution) {
    // override the webAcl using escape hatch
  }
```

In other words, it would be good if CloudFront also had a mechanism (like CfnWebACLAssociation) to attach WAF after defining resources.
This would allow WAF custom constructs to be more generic.

### Description of changes

Add `attachWebAclId` method for Distribution.

```ts
declare const bucketOrigin: origins.S3Origin;
declare const webAcl: wafv2.CfnWebACL;

const distribution = new cloudfront.Distribution(stack, 'Distribution', {
  defaultBehavior: { origin: bucketOrigin },
});

distribution.attachWebAclId(webAcl.attrArn);
```

Both of unit and integ tests.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/cloudformation-diff/package.json
+++ b/packages/@aws-cdk/cloudformation-diff/package.json
@@ -23,8 +23,8 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@aws-cdk/aws-service-spec": "^0.1.33",
-    "@aws-cdk/service-spec-types": "^0.0.100",
+    "@aws-cdk/aws-service-spec": "^0.1.34",
+    "@aws-cdk/service-spec-types": "^0.0.101",
     "chalk": "^4",
     "diff": "^5.2.0",
     "fast-deep-equal": "^3.1.3",
--- a/packages/@aws-cdk/integ-runner/package.json
+++ b/packages/@aws-cdk/integ-runner/package.json
@@ -74,7 +74,7 @@
     "@aws-cdk/cloud-assembly-schema": "^38.0.0",
     "@aws-cdk/cloudformation-diff": "0.0.0",
     "@aws-cdk/cx-api": "0.0.0",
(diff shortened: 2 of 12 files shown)


badmintoncryer (approved):
Thank you for your contribution. I can't wait to try out this feature myself!

GavinZZ (approved):
Thanks for contributing!

go-to-k:
@mergifyio update

go-to-k:
@Mergifyio update

Issue #1633, opened 2019-01-29
Depend on bucket and policy before configuring ELB logging


schof:
I tried adding logging support to my VPC using the following:

```javascript
    const alb = new Alb.ApplicationLoadBalancer(this, 'LB', {
      vpc,
      internetFacing: true
    })

    const bucket = new Bucket(this, 'Bucket', {
      encryption: BucketEncryption.KmsManaged
    })
    alb.logAccessLogs(bucket)
```

When running `cdk deploy` I got an expected warning about IAM changes and the permission looks to be correct.

> \+ │ ${Bucket.Arn}/* │ Allow  │ s3:PutObject │ AWS:arn:${AWS::Partition}:iam::127311923021:root

It appears that there is a bug in the CF template which is not waiting on the bucket policy to finish completion before it attempts to add the logging in the VPC.

> Access Denied for bucket: [BUCKET NAME]. Please check S3bucket permission (Service: AmazonElasticLoadBalancingV2; Status Code: 400; Error Code: InvalidConfigurationRequest; Request ID: 657ff61c-23dd-11e9-94e9-c57251c19c33)

I confirmed this was the case by checking the CF events in the console.

schof:
I found a temporary workaround but it's not perfect since I couldn't figure out a way to add `DependsOn` for the BucketPolicy that is also being created.

```javascript
// Temporary Hack (https://github.com/awslabs/aws-cdk/issues/1633)
const albResource = alb.node.findChild('Resource') as Alb.CfnLoadBalancer
const bucketResource = bucket.node.findChild('Resource') as cdk.Resource
albResource.addDependency(bucketResource)
```
It only works because the ALB takes so long to create (from scratch) that the bucket stuff is generally done in time. I tried working up a fix but I ran into some difficulties with `test-region` and my general unfamiliarity with the source.

RomainMuller:
This will be much easier to fix (and much cleaner, too) once #1583 has landed. At this point we'll be able to simply register the dependency properly in the `logAccesLogs` method.

gbooth27:
I just ran into this, any ETA on a fix?

pepito-j:
Hey team. Wanted to chime in as I think I found an issue with how the dependency is being placed. If the ApplicationLoadBalancer Construct gets extended and the user configures bucket access logs, this is going to create a circular dependency because of this line: 
[...]

PR #30599, opened 2024-06-20
feat(elasticloadbalancingv2): connection logs for ALB

sakurai-ryo:
### Reason for this change
ALB can output connection logs as well as access logs to the S3 bucket, but this is not yet supported by L2 Construct.
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-connection-logs.html

### Description of changes
The implementation is almost the same as for access logs.
However, since connection logs are not supported by NLB, but only by ALB, the `logConnectionLogs` method is added to the `ApplicationLoadBalancer` instead of the `BaseLoadBalancer`.

The needed BucketPolicy is described in the documentation only as follows, but to support buckets that still use the ACL, it is necessary to set the same policy that is currently set in the access logs.
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::elb-account-id:root"
      },
      "Action": "s3:PutObject",
       "Resource": "arn:aws:s3:::bucket-name/prefix/AWSLogs/aws-account-id/*"
    }
  ]
}
```
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-connection-logging.html#attach-bucket-policy-connection

add unit tests and integ tests.

### Checklist

----


packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-load-balancer.ts
```diff
@@ -265,15 +265,33 @@ export class ApplicationLoadBalancer extends BaseLoadBalancer implements IApplic
     * See https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html#bucket-permissions-troubleshooting
     */
 
+    if (bucket.encryptionKey) {
+      throw new Error('Encryption key detected. Bucket encryption using KMS keys is unsupported');
+    }
+    super.logAccessLogs(bucket, prefix);
```
sakurai-ryo, line 271: This PR adds a method with the same content as `BaseLoadBalancer.logAccessLogs` to `ApplicationLoadBalancer` to perform validation for ALB only.
However, modifying both methods would be necessary when changing the access log settings, so I changed it to call the base class implementation except for the validation.
https://github.com/aws/aws-cdk/pull/29382

go-to-k: Isn't this change fundamentally not related to this PR for the connection logs? If so, it would be better to use a separated PR.

If not, please let me know.

go-to-k (changes requested):
Thanks for this PR. I just made a minor comment.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-load-balancer.ts
```diff
@@ -265,15 +265,33 @@ export class ApplicationLoadBalancer extends BaseLoadBalancer implements IApplic
...
+  /**
+   * Enable connection logging for this load balancer.
+   *
+   * A region must be specified on the stack containing the load balancer; you cannot enable logging on
+   * environment-agnostic stacks. See https://docs.aws.amazon.com/cdk/latest/guide/environments.html
+   */
+  public logConnectionLogs(bucket: s3.IBucket, prefix?: string) {
+    /**
+    * KMS key encryption is not supported on Connection Log bucket for ALB, the bucket must use Amazon S3-managed keys (SSE-S3).
+    * See https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html#bucket-permissions-troubleshooting
```
go-to-k, line 283: Is there the description of the connection log in the troubleshooting section? If not, is there elsewhere?

sakurai-ryo:
@go-to-k 
Thanks for the review!
I fixed the wrong link and deleted unrelated changes for the connection logs feat.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-load-balancer.ts
```diff
@@ -312,6 +312,66 @@ export class ApplicationLoadBalancer extends BaseLoadBalancer implements IApplic
     }
   }
 
+  /**
+   * Enable connection logging for this load balancer.
+   *
+   * A region must be specified on the stack containing the load balancer; you cannot enable logging on
+   * environment-agnostic stacks. See https://docs.aws.amazon.com/cdk/latest/guide/environments.html
```
go-to-k, line 319: How about using the `@see` label?

sakurai-ryo: @go-to-k 
Yes, add the `@see` label.

go-to-k (approved):
Thanks for the changes.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-load-balancer.ts
```diff
@@ -312,6 +312,68 @@ export class ApplicationLoadBalancer extends BaseLoadBalancer implements IApplic
...
+        bucket.arnForObjects(`${prefix ? prefix + '/' : ''}AWSLogs/${Stack.of(this).account}/*`),
+      ],
+    }));
+    // We still need this policy for the bucket using the ACL
+    bucket.addToResourcePolicy(
+      new PolicyStatement({
+        actions: ['s3:PutObject'],
+        principals: [logsDeliveryServicePrincipal],
+        resources: [
+          bucket.arnForObjects(`${prefix ? prefix + '/' : ''}AWSLogs/${this.env.account}/*`),
```
shikha372, line 352: for my own understanding, is there a difference in setting this value to `this.env.account` instead of `Stack.of(this).account` and purpose of using this instead of how we used it in line 343.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/test/alb/load-balancer.test.ts
```diff
@@ -645,6 +645,344 @@ describe('tests', () => {
...
+        },
+      });
+
+      // verify the ALB depends on the bucket policy
+      Template.fromStack(stack).hasResource('AWS::ElasticLoadBalancingV2::LoadBalancer', {
+        DependsOn: ['ImportedConnectionLoggingBucketPolicy548EEC12'],
+      });
+    });
+
+    test('does not add circular dependency on bucket with extended load balancer', () => {
```
shikha372, line 938: are we checking that s3 bucket creation doesn't depend on ELB here?

packages/@aws-cdk-testing/framework-integ/test/aws-elasticloadbalancingv2/test/integ.alb.log.imported-bucket.js.snapshot/aws-cdk-alb-log-imported-bucket-integ.template.json
```diff
@@ -438,28 +438,64 @@
...
         ]
        }
       },
       {
-       "Action": "s3:GetBucketAcl",
+       "Action": "s3:PutObject",
        "Effect": "Allow",
        "Principal": {
-        "Service": "delivery.logs.amazonaws.com"
+        "AWS": "arn:aws:iam::797873946194:root"
```
shikha372, line 498: is this expected ?

packages/aws-cdk-lib/aws-elasticloadbalancingv2/lib/alb/application-load-balancer.ts
sakurai-ryo: Replaced `this.env.account` with `Stack.of(this).account`.
Since there is no `logConnectionLogs` method in `IApplicationLoadBalancer`, in the `logConnectionLogs` method, `Stack.of(this).account` and `this.env.account` will point to the same region.
So either one is fine, but `Stack.of(this).account` is clearer.

packages/aws-cdk-lib/aws-elasticloadbalancingv2/test/alb/load-balancer.test.ts
sakurai-ryo: Added [test code](https://github.com/aws/aws-cdk/pull/30599/files#diff-8de1f35a0483f429c861e21e5353123bc5e610f98602b0cccc7f6f877b3a6848R982) to verify that there is no `DependsOn` on the bucket to ensure that the bucket storing the logs does not set an explicit dependency on the bucket policy.

Here is the cause for this test case to be added for reference.
https://github.com/aws/aws-cdk/issues/1633#issuecomment-1758171559
https://github.com/aws/aws-cdk/pull/27558

packages/@aws-cdk-testing/framework-integ/test/aws-elasticloadbalancingv2/test/integ.alb.log.imported-bucket.js.snapshot/aws-cdk-alb-log-imported-bucket-integ.template.json
sakurai-ryo: The diff is hard to review, but this is what I expected.
Originally, the policy allowing s3:GetBucketAcl was the final element in the Statement array, but the policy allowing s3:PutObject from the ALB AWS account is now the final element.

sakurai-ryo:
@shikha372 
Thanks for the review!
I've made the changes based on your feedback

shikha372:
Thank you for your contribution @sakurai-ryo !!

shikha372 (approved):

PR #30610, opened 2024-06-21
fix(s3-notifications): unable to delete the existing S3 event notifications  (under feature flag)

### LINKED ISSUE #28915: @aws-cdk/aws-s3: Unable to delete the existing S3 event notifications
### Describe the bug

I'm facing an issue while trying to delete the existing S3 event notifications.

The problem arises with S3 bucket event notification handler Lambda(AWS CloudFormation handler for "Custom::S3BucketNotifications" resources (@aws-cdk/aws-s3)), where it fails and reports an error: when calling the PutBucketNotificationConfiguration operation: Configuration is ambiguously defined. Cannot have overlapping suffixes in two rules if the prefixes are overlapping for the same event type.

Custom event notification handler reference :  https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py

The issue is likely related to the change: https://github.com/aws/aws-cdk/commit/37be7b9b5d2abe04aa5f632a7b78aed3e9c233d8

The offending line is Line 39 - we are updating the ID of a JSON object, but the ID is inside of that object when it comes from outside. (https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py#L39)

### Expected Behavior

Should delete the existing event notifications

### Current Behavior

Unable to delete the existing filters: 
Error: when calling the PutBucketNotificationConfiguration operation: Configuration is ambiguously defined. Cannot have overlapping suffixes in two rules if the prefixes are overlapping for the same event type.

### Reproduction Steps

1. Add few event notifications
2. Remove some of the the added notifications.
Please find the example stack attached to reproduce the issue.
[s3-event-notifications.txt](https://github.com/user-attachments/files/15928937/s3-event-notifications.txt)

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.124.0

### Framework Version

_No response_

### Node.js Version

v20.11.0

### OS

Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

sarangarav:
### Issue # (if applicable)

Closes #28915 

### Reason for this change

Fix to address the issues deleting the existing S3 event notifications and adding new event notifications on top of existing notifications.

### Description of changes
We fixed the hashing logic used to identify old vs external S3 event notifications

Manually tested, integration and unit tested the changes

yes
### Checklist

----

### DIFF
--- a/.github/workflows/issue-label-assign.yml
+++ b/.github/workflows/issue-label-assign.yml
@@ -58,7 +58,7 @@ jobs:
 env:
  OSDS_DEVS: >
            {
-             "assignees":["indrora","khushail","hunhsieh"]
+             "assignees":["ashishdhingra","khushail","hunhsieh"]
            }
 
  AREA_AFFIXES: >
@@ -283,5 +283,6 @@ env:
             {"area":"@aws-cdk/triggers","keywords":["trigger","triggers"],"labels":["@aws-cdk/triggers"]},
             {"area":"@aws-cdk/integ-tests","keywords":["integ-tests", "integ"],"labels":["@aws-cdk/integ-tests"]},
             {"area":"@aws-cdk/integ-runner","keywords":["integ-runner"],"labels":["@aws-cdk/integ-runner"]},
-            {"area":"@aws-cdk/app-staging-synthesizer-alpha", "keywords":["app-staging-synthesizer", "app-staging-synthesizer-alpha"],"labels":["@aws-cdk/app-staging-synthesizer-alpha"]}
+            {"area":"@aws-cdk/app-staging-synthesizer-alpha", "keywords":["app-staging-synthesizer", "app-staging-synthesizer-alpha"],"labels":["@aws-cdk/app-staging-synthesizer-alpha"]},
+            {"area":"@aws-cdk/aws-location-alpha", "keywords":["aws-location", "aws-location-alpha"],"labels":["@aws-cdk/aws-location-alpha"]}
           ]
--- a/packages/aws-cdk-lib/aws-rds/lib/cluster-engine.ts
+++ b/packages/aws-cdk-lib/aws-rds/lib/cluster-engine.ts
@@ -549,6 +549,8 @@ export class AuroraMysqlEngineVersion {
   public static readonly VER_2_12_0 = AuroraMysqlEngineVersion.builtIn_5_7('2.12.0');
   /** Version "5.7.mysql_aurora.2.12.1". */
   public static readonly VER_2_12_1 = AuroraMysqlEngineVersion.builtIn_5_7('2.12.1');
+  /** Version "5.7.mysql_aurora.2.12.2". */
+  public static readonly VER_2_12_2 = AuroraMysqlEngineVersion.builtIn_5_7('2.12.2');
   /**
    * Version "8.0.mysql_aurora.3.01.0"
    * @deprecated Aurora MySQL 8.0.mysql_aurora.3.01.0 is no longer supported by Amazon RDS.
@@ -607,6 +609,8 @@ export class AuroraMysqlEngineVersion {
   public static readonly VER_3_05_2 = AuroraMysqlEngineVersion.builtIn_8_0('3.05.2');
   /** Version "8.0.mysql_aurora.3.06.0". */
   public static readonly VER_3_06_0 = AuroraMysqlEngineVersion.builtIn_8_0('3.06.0');
+  /** Version "8.0.mysql_aurora.3.06.1". */
+  public static readonly VER_3_06_1 = AuroraMysqlEngineVersion.builtIn_8_0('3.06.1');
   /** Version "8.0.mysql_aurora.3.07.0". */
   public static readonly VER_3_07_0 = AuroraMysqlEngineVersion.builtIn_8_0('3.07.0');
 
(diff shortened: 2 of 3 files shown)


packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource-handler.ts
```diff
@@ -84,13 +84,9 @@ export class NotificationsResourceHandler extends Construct {
 
     const handlerSource = fs.readFileSync(path.join(__dirname, '..', '..', '..', 'custom-resource-handlers', 'dist', 'aws-s3', 'notifications-resource-handler', 'index.py'), 'utf8');
 
-    // Removing lines that starts with '#' (comment lines) in order to fit the 4096 limit
+    // Removing lines that starts with '#' (comment lines)
     const handlerSourceWithoutComments = handlerSource.replace(/^ *#.*\n?/gm, '');
 
-    if (handlerSourceWithoutComments.length > 4096) {
```
sarangarav, line 90: Inline function code can now exceed 4096 bytes. It can support up to 4MB now.

Latest public documentation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html#cfn-lambda-function-code-zipfile

packages/@aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py
```diff
@@ -34,20 +35,31 @@ def handle_managed(request_type, notification_configuration):
     return {}
   return notification_configuration
 
-def handle_unmanaged(bucket, stack_id, request_type, notification_configuration, old):
+def handle_unmanaged(bucket, stack_id, request_type, notification_configuration, old, s3NotificationsDeleteFeatureFlagEnabled):
+  def get_id(n):
+    n['Id'] = ''
```
GavinZZ, line 40: Why do we need to set `n['Id'] = ''` here?

packages/@aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py
```diff
@@ -34,20 +35,31 @@ def handle_managed(request_type, notification_configuration):
...
+    strToHash=json.dumps(n, sort_keys=True).replace('"Name": "prefix"', '"Name": "Prefix"').replace('"Name": "suffix"', '"Name": "Suffix"')
+    return f"{stack_id}-{hash(strToHash)}"
   def with_id(n):
-    n['Id'] = f"{stack_id}-{hash(json.dumps(n, sort_keys=True))}"
+    if s3NotificationsDeleteFeatureFlagEnabled:
+      n['Id'] = get_id(n)
```
GavinZZ, line 45: Can the `get_id` uses the same naming format where it starts with `stack-id` so the `Delete` event can handle it correctly since it relies on `{stack-id}` prefix.

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource.ts
```diff
@@ -133,6 +135,7 @@ export class BucketNotifications extends Construct {
           BucketName: this.bucket.bucketName,
           NotificationConfiguration: cdk.Lazy.any({ produce: () => this.renderNotificationConfiguration() }),
           Managed: managed,
+          S3NotificationsDeleteFeatureFlagEnabled: FeatureFlags.of(this).isEnabled(S3_EXISTING_NOTIFICATIONS_DELETE_ENABLED),
```
GavinZZ, line 138: I think this name is a bit misleading. We're not actually changing the `Delete` event but more of the `Update` event, and it's not something that users can modified so we don't need to mention it's a feature flag. Can we change the name to something like `ApplyNameTransformations` or anything better that you can think of.

GavinZZ (changes requested):
Left some feedbacks

packages/@aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py
sarangarav: To exclude the "Id" field while computing the hash values for existing vs old notifications. Despite the content of the old and existing notifications being the same, their hash values are being generated differently due to a mismatch in the capitalization of the "Prefix" field (e.g., "Prefix" vs "prefix") and id field in existing notifications.

packages/@aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py
sarangarav: Yes, it's using the same format.

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource.ts
sarangarav: How about ApplyS3EventNotificationNameTransformations?

GavinZZ: IMO S3EventNotification is a bit redundant since this custom resource handler is for s3 notification already.

packages/@aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py
GavinZZ: Oh yeah, my bad for not noticing it's using the same format. Thanks for addressing this.

GavinZZ:
@mergify update

sarangarav:
@mergify update

GavinZZ (dismissed):

GavinZZ (dismissed):

GavinZZ (dismissed):

sarangarav:
@mergifyio refresh

GavinZZ (approved):

PR #29754, opened 2024-04-07
feat(codebuild): `Fleet` L2

nmussy:
### Issue # (if applicable)

Closes #29617.

### Reason for this change

Implementation of reserved capacity CodeBuild projects 

### Description of changes

* Add Fleet Construct
* Add `EnvironmentType` enum
  * Refactor existing type strings to use the new enum	
* Validate that Windows 2022 build images can only be used in fleet projects

Changes merged from #29616:
* Added missing build images
* Updated JSDoc comments to indicate AL2023 based images, see [docs](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html)
  * It might be a good idea to deprecate and rename `AMAZON_LINUX_2_STANDARD_3_0` to `AMAZON_LINUX_2023_STANDARD_3_0`, despite how the images are named. I'll leave it up to the maintainers
* Added `{@link}` tags where missing

### Description of how you validated changes

Unit and integ tests

The images were retrieved using the [codebuild:ListCuratedEnvironmentImages](https://docs.aws.amazon.com/codebuild/latest/APIReference/API_ListCuratedEnvironmentImages.html) API command, and comparing it to the CDK.

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30615, opened 2024-06-21
fix(codebuild): prevent fleet and batch build combined usage

nmussy:
### Issue # (if applicable)

None, initially reported by @michaelbwebb, see #30596

### Reason for this change

The following stack would cause a CloudFormation deployment error:

```ts
const app = new cdk.App();
const stack = new cdk.Stack(app, 'aws-cdk-pipeline-project-fleet');

const repository = new codecommit.Repository(stack, 'MyRepo', {
  repositoryName: 'integration-repo',
});

const fleet = new codebuild.Fleet(stack, 'MyFleet', {
  fleetName: 'MyFleet',
  baseCapacity: 1,
  computeType: codebuild.FleetComputeType.SMALL,
  environmentType: codebuild.EnvironmentType.LINUX_CONTAINER,
});

const project = new codebuild.PipelineProject(stack, 'MyProject', {
  environment: {
    fleet,
    buildImage: codebuild.LinuxBuildImage.STANDARD_7_0,
  },
});

const artifactBucket = new s3.Bucket(stack, 'MyBucket', {
  versioned: true,
});
const pipeline = new codepipeline.Pipeline(stack, 'Pipeline', {
  artifactBucket,
});

const sourceOutput = new codepipeline.Artifact();
pipeline.addStage({
  stageName: 'Source',
  actions: [new cpactions.CodeCommitSourceAction({
    actionName: 'Source',
    repository,
    output: sourceOutput,
    role: pipeline.role,
  })],
});

// project.enableBatchBuilds() is invoked here:
pipeline.addStage({
  stageName: 'Build',
  actions: [new cpactions.CodeBuildAction({
    actionName: 'Build',
    project,
    executeBatchBuild: true,
    input: sourceOutput,
    role: pipeline.role,
  })],
});
```

I initially thought that the `fleet` property was not correctly set for `PipelineProject`s, but this is only tangentially related CodePipeline, see [Working with reserved capacity in AWS CodeBuild](https://docs.aws.amazon.com/codebuild/latest/userguide/fleets.html):

> Reserved capacity fleets don't support batch builds

This is reflected by the CloudFormation error thrown by the above stack:
```
Build batch is not supported for project using reserved capacity
```

While I understand the CDK does not need to double check every CloudFormation exception, fleets can be slow to deploy, and particularly slow to destroy. This process can be costly depending on the compute resource allocation, which in my opinion warrants a synth time check

### Description of changes

* Added check and error if a project has both a fleet and batch builds enabled
	* Added public `Project.isBatchBuildEnabled` property, giving the ability to check whether `enableBatchBuilds()` has been invoked 
	* Added private `Project._environment` property, to retrieve the environment value synthesized in the `constructor` method 

Added a unit test. The CloudFormation error was found whilst attempting to add an integration test for a reserved capacity `PipelineProject`

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-codebuild/lib/project.ts
+++ b/packages/aws-cdk-lib/aws-codebuild/lib/project.ts
@@ -276,12 +276,6 @@ abstract class ProjectBase extends Resource implements IProject {
   protected _connections: ec2.Connections | undefined;
 
   /**
-   * Actual value will be determined for a Project
-   * using a getter depending on the effect of enableBatchBuilds
-   */
-  public readonly isBatchBuildEnabled = false;
-
-  /**
    * Access the Connections object.
    * Will fail if this Project does not have a VPC set.
    */
@@ -290,6 +284,14 @@ abstract class ProjectBase extends Resource implements IProject {
       throw new Error('Only VPC-associated Projects have security groups to manage. Supply the "vpc" parameter when creating the Project');
     }
     return this._connections;
+  }
+
+  /**
+   * Actual value will be determined for a Project
+   * using a getter depending on the effect of enableBatchBuilds
+   */
+  public get isBatchBuildEnabled(): boolean {
+    return false;
   }
 
   public enableBatchBuilds(): BatchBuildConfig | undefined {
@@ -1220,7 +1222,7 @@ export class Project extends ProjectBase {
     this.node.addValidation({ validate: () => this.validateProject() });
   }
 
-  public get isBatchBuildsEnabled(): boolean {
+  public get isBatchBuildEnabled(): boolean {
     return !!this._batchServiceRole;
   }
 
--- a/packages/aws-cdk-lib/aws-codebuild/test/codebuild.test.ts
+++ b/packages/aws-cdk-lib/aws-codebuild/test/codebuild.test.ts
@@ -2104,13 +2104,13 @@ test('enableBatchBuilds()', () => {
       repo: 'testrepo',
     }),
   });
-  expect(project.isBatchBuildsEnabled).toBeFalsy();
+  expect(project.isBatchBuildEnabled).toBeFalsy();
 
   const returnVal = project.enableBatchBuilds();
   if (!returnVal?.role) {
     throw new Error('Expecting return value with role');
   }
-  expect(project.isBatchBuildsEnabled).toBeTruthy();
+  expect(project.isBatchBuildEnabled).toBeTruthy();
 
   Template.fromStack(stack).hasResourceProperties('AWS::CodeBuild::Project', {
     BuildBatchConfig: {


packages/aws-cdk-lib/aws-codebuild/lib/project.ts
```diff
@@ -1426,6 +1450,13 @@ export class Project extends ProjectBase {
       return undefined;
     }
 
+    // !! Failsafe !! This should not occur with the current methods, given:
+    // * The fleet can only be set with the constructor
+    // * The batch builds can only be enabled with the `enableBatchBuilds` method
+    if (this.isBatchBuildEnabled) {
+      throw new Error('Build batch is not supported for project using reserved capacity');
+    }
```
nmussy, line 1458: As mentioned in the comment, this condition should never be true. I've added it just in case a future CDK API change enables this behavior

scanlonp (commented):
Overall I like the idea here. This case seems fairly easy to run in to, and the error from CloudFormation is not immediately clear on how you have messed up your CDK configuration, so I support the synth-time check. 

We have ran into a couple issues with synth-time checks being overly restrictive, so a couple questions on hypothetical cases where a user could run into this.

- Is there any way for a user to set the fleet property but not use reserved capacity? (This may be the definition of the [CFN fleet resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-codebuild-fleet.html), but wanted to make sure theres no way to set on-demand fleets with this property)
- Is there a way for a user to pass in a "dummy" fleet for their environment, which would not result in reserved capacity, but would still trigger this check?

I'd expect no to both, but still worth a consideration.

Lastly, I don't quite understand the boolean variable and the getter method. I do not believe both are needed. The variable is never set, and the getter method evaluates a different variable. Could you set the variable to true in the `enableBatchBuilds()` method, and then reference it directly in `configureFleet()`? Or could the function be the evaluation without the corresponding variable?

Of the two, I would prefer the direct set of the variable, but up to you.

packages/aws-cdk-lib/aws-codebuild/lib/project.ts
scanlonp: Agree, this is a good regression check.

nmussy:
Hey @scanlonp, thanks for the review and sorry for the late reply.

> Is there any way for a user to set the fleet property but not use reserved capacity? (This may be the definition of the [CFN fleet resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-codebuild-fleet.html), but wanted to make sure theres no way to set on-demand fleets with this property)

Not as far as I can tell, setting a fleet to a CodeBuild project makes it become a reserved capacity project, see [Working with reserved capacity in AWS CodeBuild](https://docs.aws.amazon.com/codebuild/latest/userguide/fleets.html)

> Is there a way for a user to pass in a "dummy" fleet for their environment, which would not result in reserved capacity, but would still trigger this check?

I'm not sure how they would do that. The `BaseCapacity` must be greater than 0, and they must either be EC2 managed on-demand instances, or user managed reserved instances. I suppose you could use trashed reserved instances, but from what I've seen when I implemented #29754, if there are no instances able or available to run a reserved build project, it just doesn't build.

> Lastly, I don't quite understand the boolean variable and the getter method.

The boolean value and the getter method are one and the same. `isBatchBuildEnabled` is defined as a boolean value in the interface, and implemented as getter in the two classes. But you're right, I could switch to a primitive boolean member variable and set it in `enableBatchBuilds()`, or remove it altogether and derive its value on the fly. Let me know which you would prefer.

kumvprat:
@nmussy Thank you for the effort on this PR
But I think this PR is now obsolete with : https://aws.amazon.com/about-aws/whats-new/2025/01/aws-codebuild-batch-builds-reserved-capacity-lambda-compute/

Batch builds are now supported with reserved capacities/fleets

Screenshot 2025-06-11 at 17 04 54

Will close the PR as the support for this now there

PR #30621, opened 2024-06-22
feat(cloudfront): find and replace text in the code of CloudFront Functions during synth

### LINKED ISSUE #30492: feat(cloudfront): parameterise CloudFront Functions via find/replace
### Describe the feature

CloudFront Functions (not Lambda@Edge) don't provide a way to parameterise the code, in the way environment variables can with Lambda@Edge, for example. Allow code to be replaced with values from the CDK app during synth.

### Use Case

- Inject the Key Value Store (KVS) ID into the function when the KVS is deployed as part of the CDK app.
- Change the function for different environments, dev vs prod for example.

### Proposed Solution

In aws-cdk-lib/aws-cloudfront/lib/function.ts:
- Add these interfaces.
  ```typescript
  export interface FindReplace {
    readonly find: string
    readonly replace: string | cdk.Token
    readonly all?: boolean
  }

  export interface FunctionCodeOptions {
    readonly findReplace?: FindReplace[]
  }
   ```
- Add a protected findReplace(code) method to the abstract class FunctionCode, which returns the code after doing the find/replaces of the options. 
  - If replace is a Token, the code is replace with the result of cdk.Token.asString(replace)
  - If all is truthy, all occurrences are replaced. If all is falsey only the first occurrence is replaced.
- Change the render methods of InlineCode and FileCode to call findReplace before returning their respective values.
- Change FileCodeOptions to extend FunctionCodeOptions.
- Change InlineCode's constructor to take a second options parameter, being a FunctionCodeOptions object.
  - The options parameter is a private property.
- Change FunctionCode.fromInline(code) to FunctionCode.fromInline(code, options).
  - The new second parameter is optional, defaulting to an empty object.
  - Pass options to InlineCode's constructor.

This is a naïve find (all) and replace. It does not attempt to parse the function code to find appropriate tokens. It does not attempt to replace code meaningfully. There is no guarantee that syntactically valid code remains syntactically valid after the find/replace.

<details><summary>Proof of concept</summary>
<p>
This is the code I'm currently using to get the KVS ID into a CloudFront Function. The proposed solution is a generalisation of this. 

```typescript
interface FileCodeWithKvsIdSubstitutionOptions extends cloudfront.FileCodeOptions {
  readonly keyValueStore: cloudfront.IKeyValueStore
}

class FileCodeWithKvsIdSubstitution extends cloudfront.FunctionCode {

  constructor(private options: FileCodeWithKvsIdSubstitutionOptions) {
    super()
  }

  public render(): string {
    return (
      fs.readFileSync(this.options.filePath, { encoding: 'utf-8' })
      .replace('«KVS_ID»', cdk.Token.asString(this.options.keyValueStore.keyValueStoreId))
    )
  }

}
``` 
</p>
</details> 

### Example of using the proposed solution

Given this code for a CloudFront Function.

```javascript
import cf from 'cloudfront'

// __IS_PROD__ anywhere in the code is replaced with true or false during CDK synth 

const kvs = cf.kvs('__KVS_ID__')  // __KVS_ID__ is replaced during CDK synth

async function handler(event) {
  console.log('is production: __IS_PROD__')
  // …
  if (__IS_PROD__) {
   // …
  }
  // …
  const v = __IS_PROD__ ? x : y
  // …
}
```

> The use of double underscore delimiters is not part of the proposed solution. It would up to the CDK builder to create code that can be unambiguously found. If any delimiters are used, those delimiters would be part of the value for find.

\_\_KVS_ID__ and \_\_IS_PROD__ could be replaced in the code with this. 

```typescript
new cloudfront.Function(this, 'Function', {
  code: FunctionCode.fromFile({
    filePath: 'cloudfront-function.js',
    findReplace: [
      { find: '__KVS_ID__',  replace: keyValueStore.keyValueStoreId },
      { find: '__IS_PROD__', replace: `${ !!isProd }`, all: true },            // replace has to be string
    ],
  }),
})
```

If the KVS id is `61736184-1ad9-4b1b-9466-6ca1fb629685` and isProd is `false`, the code that would be sent to CloudFront would be this.

```javascript
import cf from 'cloudfront'

// false anywhere in the code is replaced with true or false during CDK synth 

const kvs = cf.kvs('61736184-1ad9-4b1b-9466-6ca1fb629685')  // 61736184-1ad9-4b1b-9466-6ca1fb629685 is replaced during CDK synth

async function handler(event) {
  console.log('is production: false')
  // …
  if (false) {
   // …
  }
  // …
  const v = false ? x : y
  // …
}
```

### Other Information

The Key Value Store can be used for parameterisation. But there's costs associated with that. And there's the conundrum of how to get the KVS ID into the CloudFront Function to begin with.

### Acknowledgements

### CDK version used

2.145.0

NodeJS/iron

SydneyUni-Jim:
### Issue # (if applicable)
Closes #30492

### Reason for this change

~~Primarily to allow deploying a CloudFront Function and its Key/Value Store in the same stack, without having to deploy the Key/Value Store first, manually retrieve its ID, and then hardcode the ID into the Function's code.~~

Secondarily to allow other values in a CloudFront Function's code to be changed when deploying into different environments, dev verses prod for example.

### Description of changes

* aws_cloudfront/README.md
  * Update the CloudFront Functions and Key Value Store sections.
* aws_cloudfront/function.ts
  * Add struct FunctionCodeFindReplace.
  * Add struct FunctionCodeOptions, with the *optional* findReplace property.
  * Change struct FileCodeOptions to extend FunctionCodeOptions.
  * Add struct InlineCodeOptions, which extends FunctionCodeOptions with no additional properties.
    * For symmetry with FileCodeOptions and as a future extension point.
  * Add an optional options parameter to FunctionCode.fromInline.
  * Add private method FunctionCode.findReplace.
  * Change InlineCode.render to call FunctionCode.findReplace.
  * Change FileCode.render to call FunctionCode.findReplace.

#### Backwards compatible API changes

* FunctionCode.fromInline gains a second *optional* parameter.
* FileCodeOptions gains an *optional* findReplace property.

* Add `describe('find/replace in function code'…` to function.test.ts.
* Add integ.function-find-replace.ts
  * Deploys a CloudFront distribution, function, and key-value store.
  * Use find/replace to get the ID of the key-value store into the function.
  * Exercises the function to test the function can access the key-value store.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudfront/README.md
+++ b/packages/aws-cdk-lib/aws-cloudfront/README.md
@@ -589,7 +589,7 @@ In which case you use `FunctionCode.fromFile()` instead of `FunctionCode.fromInl
 
 ```ts
 new cloudfront.Function(this, 'Function', {
-  code: cloudfront.FunctionCode.fromFile('function-code.js'),
+  code: cloudfront.FunctionCode.fromFile({ filePath: 'function-code.js' }),
   runtime: cloudfront.FunctionRuntime.JS_2_0
 });
 ```
@@ -606,7 +606,7 @@ Consider this CloudFront function in `function-code.js`. The intention to
 replace `%DEFAULT_REDIRECT%` with a URL that is different for different
 environments.
 
-```ts
+```js
 function handler(event) {
   // …
   return {
@@ -616,33 +616,37 @@ function handler(event) {
 }
 ```
 
-You can achieve this in your CDK app with this.
+You can have CDK replace `%DEFAULT_REDIRECT%` by adding a `findReplace` 
+option to `FunctionCode.fromFile`. The `findReplace` option is also
+available with `FunctionCode.fromInline`.
 
 ```ts
 declare const defaultRedirect: string // The actual URL depends on the environment, dev vs prod for example
 new cloudfront.Function(this, 'Function', {
-  code: cloudfront.FunctionCode.fromFile('function-code.js'),
   runtime: cloudfront.FunctionRuntime.JS_2_0,
-  findReplace: [
-    { find: '%DEFAULT_REDIRECT%', replace: defaultRedirect, all: true }, // all defaults to false
-  ],
+  code: cloudfront.FunctionCode.fromFile({
+    filePath: 'function-code.js',
+    findReplace: [
+      { find: '%DEFAULT_REDIRECT%', replace: defaultRedirect, all: true }, // all defaults to false
+    ],
+  }),
 });
 ```
 
 If `defaultRedirect` were `https://www.example.com/` during synthesis, the CDK would
 update your CloudFront function with this code.
 
-```ts
+```js
 function handler(event) {
   // …
   return {
     statusCode: 302, statusDescription: 'Found', 
-    headers: { location: { value: 'https://www.example.com/' } }, // %DEFAULT_REDIRECT% is replaced during CDK synth
+    headers: { location: { value: 'https://www.example.com/' } }, // https://www.example.com/ is replaced during CDK synth
   }
 }
 ```
 
-> **Note:** This is simple find and replace, like that of a non-intelligent text 
+> **Note:** This is a simple find and replace, like that of a non-intelligent text 
 > editor. The CDK does not look for whole identifiers, strings, or words. The
 > CDK replaces anything in your code that matches `find` regardless of its context.
 >
@@ -715,12 +719,12 @@ new cloudfront.Function(this, 'Function', {
 
 In your CloudFront function, put `%KVS_ID%` where you would normally put the key value store ID.
 
-```ts
+```js
 import cf from 'cloudfront';
 const kvsId = "%KVS_ID%";
 const kvsHandle = cf.kvs(kvsId);
 
-function handler(event) {
+async function handler(event) {
   // …
 }
 ```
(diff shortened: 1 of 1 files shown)


SydneyUni-Jim:
In the last PR Linter / validate-pr, node core dumped from a bus error. The prior validate-pr had succeeded.

packages/aws-cdk-lib/aws-cloudfront/lib/function.ts
```diff
@@ -31,12 +32,93 @@ export abstract class FunctionCode {
...
+export interface FunctionCodeFindReplace {
+  /**
+   * The text to find.
+   */
+  readonly find: string;
+
+  /**
+   * The text to replace the find text with.
+   */
+  readonly replace: string | Token;
```
gshpychka, line 103: Be careful here - deploy-time tokens will not be resolved with this, will they?

SydneyUni-Jim: I think they do. In the case of KVS ID, it's not known until the KVS is deployed.

aaythapa:
Hi @SydneyUni-Jim sorry about the super-late response on this PR. Thank you for contributing. Looks like the PR was closed because it was in merge conflict stage for some time. I do see that this PR was open for some time and no one got to reviewing it so that is on us. I'll add needs-maintainer-review on this so that its on our radar. If you could fix the conflict in the PR that'd be great. Thanks again for contributing!

QuantumNeuralCoder (changes requested):
Hi Can you please help look at the merge conflict?

PR #30623, opened 2024-06-22
feat(apprunner): support vpc ingress connection

### LINKED ISSUE #22850: aws-apprunner: support private access from VPC
### Describe the feature

App Runner now supports privately accessible services within VPC

https://aws.amazon.com/about-aws/whats-new/2022/11/aws-app-runner-supports-privately-accessible-services-amazon-vpc/

To reference the VPC interface endpoint it's needed:
- `ingressConfiguration` property in `NetworkConfigurationProperty`. At the moment there is only `egressConfiguration`. (https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apprunner.CfnService.NetworkConfigurationProperty.html)
- VP Ingress Connection for App Runner to associate it with the VPC interface endpoint 

### Use Case

I have an app running on App Runner and I need it to be available only via VPC and not publicly accessible.

Cloudformation already supports this.

### Proposed Solution

_No response_

### Other Information

Here details how it's implemented using CLI

https://docs.aws.amazon.com/apprunner/latest/dg/network-pl-manage.html

### Acknowledgements

### CDK version used

2.50.0

Linux

mazyu36:
### Issue # (if applicable)

Closes #22850.

### Reason for this change
To support VPC Ingress Connection  for making App Runner Service private and only accessible from within a VPC.

### Description of changes
* Add `isPubliclyAccessible` property to the `Service` class
* Add `VpcIngressConnection` class

Add unit tests and integ tests

### Checklist

----


mazyu36:
@Merfigyio update

mazyu36:
@Mergifyio update

packages/@aws-cdk/aws-apprunner-alpha/README.md
```diff
@@ -201,6 +201,42 @@ new apprunner.Service(this, 'Service', {
 });
 ```
 
+## VPC Ingress Connection
+
+You need to create a VPC Ingress Connection to make your App Runner service private and only accessible from within a VPC.
+
+To set up a VPC Ingress Connection, you need to associate a VPC, a VPC Interface Endpoint, and an App Runner Service.
```
lpizzinidev, line 208:
```suggestion
To make your App Runner service private and only accessible from within a VPC use the `isPubliclyAccessible` property and associate it to a `VpcIngressConnection` resource.

To set up a `VpcIngressConnection`, specify a VPC, a VPC Interface Endpoint, and the App Runner service.
```

packages/@aws-cdk/aws-apprunner-alpha/lib/vpc-ingress-connection.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * The name of the VPC Ingress Connection.
+   * @attribute
+   */
+  readonly vpcIngressConnectionName: string;
+
+  public constructor(scope: Construct, id: string, props: VpcIngressConnectionProps) {
+    super(scope, id, {
+      physicalName: props.vpcIngressConnectionName,
```
lpizzinidev, line 143: Did you check if it deploys with `vpcIngressConnectionName: undefined`?
If not, we should lazily generate one.

lpizzinidev (changes requested):

packages/@aws-cdk/aws-apprunner-alpha/lib/vpc-ingress-connection.ts
mazyu36: Thank you. I tried deploying with `vpcIngressConnectionName: undefined`, and I was able to deploy without any issues.

image

image

mazyu36:
@lpizzinidev 
Thanks.
I've addressed your comments.

lpizzinidev (approved):
Thanks 👍

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
```diff
@@ -732,6 +732,13 @@ export interface ServiceProps {
    */
   readonly vpcConnector?: IVpcConnector;
 
+  /**
+   * Specifies whether your App Runner service is publicly accessible.
+   *
+   * @default true
+   */
+  readonly isPubliclyAccessible?: boolean;
```
paulhcsun, line 740: I'm not too familiar with restrictions for creating a VPC ingress connection but is it still possible to create an ingress connection if `isPubliclyAccessible` is set to `true`? Just trying to figure out why this property is needed, and if that is the case that it must be set to false to create an ingress connection, we should check that somewhere when creating the `VpcIngressConnection` class constructor.

paulhcsun (changes requested):
Thanks for the contribution! Design overall looks good, just have a couple questions below.

packages/@aws-cdk/aws-apprunner-alpha/lib/vpc-ingress-connection.ts
```diff
@@ -0,0 +1,168 @@
...
+    });
+
+    if (
+      props.vpcIngressConnectionName !== undefined &&
+      !cdk.Token.isUnresolved(props.vpcIngressConnectionName) &&
+      !/^[A-Za-z0-9][A-Za-z0-9\-_]{3,39}$/.test(props.vpcIngressConnectionName)
```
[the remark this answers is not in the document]
mazyu36: This pattern is quoted from CFn docs.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apprunner-vpcingressconnection.html#cfn-apprunner-vpcingressconnection-vpcingressconnectionname

packages/@aws-cdk/aws-apprunner-alpha/lib/vpc-ingress-connection.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * The domain name associated with the VPC Ingress Connection resource.
+   */
+  readonly domainName: string;
+
+  /**
+   * The current status of the VPC Ingress Connection.
+   */
+  readonly status: string;
```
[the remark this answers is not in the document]
mazyu36: The followings are status. 

`AVAILABLE | PENDING_CREATION | PENDING_UPDATE | PENDING_DELETION | FAILED_CREATION | FAILED_UPDATE | FAILED_DELETION | DELETED`

https://docs.aws.amazon.com/apprunner/latest/api/API_VpcIngressConnection.html

I think this item should not be enum because this item is an attribute not props.
Please let me know if I misunderstood.

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
mazyu36: To use VPC Ingress Connection, the Service's `isPubliclyAccessible` needs to be set to false.
However, since the flow is to create the VPC Ingress Connection after creating the Service, we can only set it to false in the Service configuration.

I've added this restriction to README and property docs.

I considered checking the Service's i`isPubliclyAccessible` in the `VpcIngressConnection`, but I thought it wasn't possible to retrieve this since it doesn't exist as a [CFn attribute](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apprunner.CfnService.html#properties).
Is there any good way to check this?

mazyu36:
@paulhcsun 
Thank you for the review!
I've addressed the comments.

packages/@aws-cdk/aws-apprunner-alpha/lib/vpc-ingress-connection.ts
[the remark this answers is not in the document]
paulhcsun: Good point, ya let's keep it as a string to match with the CloudFormation type.

packages/@aws-cdk/aws-apprunner-alpha/lib/service.ts
paulhcsun: Hmmm ya I can't really think of a good way to check it either. Will CloudFormation throw an error if you try to create a `VpcIngressConnection` with `isPubliclyAcessible` set to true? If so then I think just documenting the restriction is the best we can do on the CDK side.

mazyu36: @paulhcsun 
Thanks.
Yeas, I tried to create `VpcIngressConnection` with `isPubliclyAcessible` set to `true` and got the following an error:

> Resource handler returned message: "Failed to create a VpcIngressConnection for a Service because it is publicly accessible.

paulhcsun: Ok in that case I think it is okay to leave it as is then. Thanks for checking!

paulhcsun (approved):
Nice work @mazyu36! Thanks for reviewing @lpizzinidev!

paulhcsun:
@mergifyio update

PR #30646, opened 2024-06-24
feat(ssm): support cross-account sharing

### LINKED ISSUE #29292: aws_ssm: StringListParameter.ValueForTypedListParameter to support shared parameters
### Describe the feature

Now that parameter sharing between accounts is a feature, the CDK constructs should support ARN parameter names so that a shared parameter from a different account can be retrieved.

### Use Case

I'm attempting to create the required Route53 resources in a multi-account organization. The DNS architecture we're using is:

- `domain-account` : contains the registered domain and a hosted zone.
- `workload-account` : contains workloads and has a sub-domain of the main domain in `domain-account`

In order for the CDK to create the required resources, I'm trying to use the following stacks:

`WorkloadAccountStack` - this stack creates a hosted zone, e.g. dev.domain.com, and creates a `StringListParameter` with a value of the NS entry of the new hosted zone. The parameter name is `_dns_DevNameServers` and is in the ADVANCED tier. Additionally, a `CfnResourceShare` is created to share the parameter with the `domain-account`.

`DomainAccountStack` - this stack creates a `NS` record in the `domain.com` hosted zone. To do this, the stack must get the value of the `_dns_DevNameServers` parameter from the `workload-account`.

[AWS docs](https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-shared-parameters.html#accessing) state that accessing a parameter is done by passing the ARN as the parameter name.

In 2.130 of CDK, I'm using the following code:

```
    public static string[] ValueForTypedListParameter(Construct scope, string id, string parameterName, Environment environment)
    {
        var parameterArn = Arn.Format(new ArnComponents
        {
            ArnFormat = ArnFormat.COLON_RESOURCE_NAME,
            Partition = AwsArnPartitions.Aws,
            Account = environment.Account,
            Region = environment.Region,
            Service = AwsArnServices.Ssm,
            Resource = AwsArnResources.Parameter,
            ResourceName = parameterName
        });

        var parameter = StringListParameter.FromListParameterAttributes(scope, id, new ListParameterAttributes
        {
            ParameterName = parameterArn
        });

        return parameter.StringListValue;
    }

```

When attempting to deploy the stack, I get the following error:

```
Could not determine the resource name from arn: arn:aws:ssm:ap-southeast-2:xxxxxxxxxxxxxx:parameter:_dns_DevNameServers
```

I believe this is due to using `COLON_RESOURCE_NAME` as the `ArnFormat`. I would like to use `SLASH_RESOURCE_NAME`, however, when using that, the `synth` process throws an error:

```
Error: Parameter names must be fully qualified (if they include "/" they must also begin with a "/")
```

### Proposed Solution

At this stage I'm not sure where to start with a solution for this problem.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.130

Windows 11

pahud:
AWS [introduced](https://aws.amazon.com/about-aws/whats-new/2024/02/aws-systems-manager-parameter-store-cross-account-sharing/) SSM Parameter Store cross-account sharing in Feb 2024. Under the hood, the sharing account has to create an AWS RAM ResourceShare for the principal of the consuming account and the consuming account has to accept the sharing invite. The only approach to access that sharing parameter is through CfnParameter. Dynamic Reference is NOT supported.

This PR adds `StringParameter.fromStringParameterArn()` method so we can use the API like

```ts
const remoteParameterArn = 'arn:aws:ssm:us-east-1:123456789012:parameter/dummyName';

const sharedParam = StringParameter.fromStringParameterArn(scope, id, remoteParameterArn);

new cdk.CfnOutput(this, 'ParamValue', { value: sharedParam.stringValue });
```

Note:

1. The only option to consume sharing parameters in CFN is template parameter. `StringParameter.fromStringParameterArn()` would synthesize CfnParameter like:

```yaml
SharedParameter:
    Description: a shared golden AMI from centralised accounts
    Type: "AWS::SSM::Parameter::Value<String>"
    Default: "arn:aws:ssm:us-east-1:123456789012:parameter/sharedParameterName"
```
And the `Default` value has to be a static string. It can't be an unresolved token.

2. For full use case(sharing, accepting and consuming), check out `integ.parameter-store-string-sharing.ts` for more details.

### Issue # (if applicable)

Closes #29292 

Debugger
```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "type": "node",
      "request": "launch",
      "name": "Jest",
      "program": "${workspaceFolder}/node_modules/jest/bin/jest.js",
      "cwd": "${workspaceFolder}/packages/aws-cdk-lib",
      "args": [
        "--verbose",
        "-i",
        "--no-cache",
        "test/parameter.test.ts",
      ],
      "console": "integratedTerminal",
      "internalConsoleOptions": "neverOpen",
      "skipFiles": [
        "<node_internals>/**"
      ],
      "outFiles": [
        "${workspaceFolder}/**/*.(m|c|)js",
        "!**/node_modules/**"
      ],
    }
  ]
}
```

### Checklist

----


packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
```diff
@@ -472,6 +473,39 @@ export class StringParameter extends ParameterBase implements IStringParameter {
...
+    }
+
+    // has to be the same region
+    // split the arn string to get the region string
+    // arn sample: arn:aws:ssm:us-east-1:123456789012:parameter/dummyName
+    const arnParts = stringParameterArn.split(':');
+    const stackRegion = Stack.of(scope).region;
+    if (arnParts.length !== 6) {
+      throw new Error('unexpected StringParameterArn format');
+    } else if (!Token.isUnresolved(stackRegion) && arnParts[3] !== stackRegion) {
```
shikha372, line 491: can we enforce user to set a region in order to use this shared parameter or add something like a warning which states that same region is expected for this shared parameter arn, this is to ensure they don't see the failure directly at runtime.

packages/@aws-cdk-testing/framework-integ/test/aws-ssm/test/integ.parameter-store-string-sharing.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    this.resourceShareArn = share.attrArn;
+    new cdk.CfnOutput(this, 'ResourceShareArn', { value: this.resourceShareArn });
+    new cdk.CfnOutput(this, 'ParameterArn', { value: this.parameterArn });
+
+    share.node.addDependency(publicParameter);
+
+  }
+}
+class AcceptingStack extends cdk.Stack {
```
shikha372, line 68: slightly confused as to why we have accepting stack and consuming stack as different, can we combine both into one, i.e. keep the accepting API call in consuming stack.

shikha372:
Thanks @pahud for submitting the PR, left minor comment, also curious whether we are looking to any further implementation in future probably with help of custom resources etc. to achieve resource sharing feature.

packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
pahud: Yes we could. But if we do that, those who prefer stack environment-agnostic and is aware of this limitation would be forced to configure stack env vars. They would lose their freedom to be environment-agnostic.

Failing early is great. But if the cost is losing the freedom of environment-agnostic I would be hesitant to enforce that. wdyt?

packages/@aws-cdk-testing/framework-integ/test/aws-ssm/test/integ.parameter-store-string-sharing.ts
pahud: We need to ensure the order:

1. producer Account to create a ssm parameter
2. producer Account to create a RAM share with the consumer Account
3. consumer Account to list/select the invitation via AwsApiCall
4. consumer Account to accept the invitation via  yet another AwsApiCall
5. consumer Account to consume this shared parameter


In the `ConsumingStack` there's only one statement

```ts
const remoteParam = ssm.StringParameter.fromStringParameterArn(this, 'RemoteParam', props.parameterArn);
```

As this is not a resource or construct, it won't be able to depend on the `AwsApiCall` which is a Construct. But we still have to wait before the invitation is accepted.

The workaround here is to group them into stacks and ensure stack dependency.

packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
shikha372: would adding a warning be forcing? i thought warning would just let them know what's about to happen in case they've not configured it and it will not fail synth or deploy, one can choose to ignore or take care of it.

packages/@aws-cdk-testing/framework-integ/test/aws-ssm/test/integ.parameter-store-string-sharing.ts
shikha372: Got it so the dependency is to make sure the invite has been accepted before consumption. Thank you for explanation.

packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
pahud: Sure. Just added an annotation warning when stack.region is an unresolved token. Let me know if it looks good to you.

shikha372 (approved):

PR #30653, opened 2024-06-24
chore(ssm): update simple name description and documentation

### LINKED ISSUE #28778: aws_ssm: When creating or consuming an SSM Parameter w/ a simple name, the arn attribute of Parameter/IParameter construct contains an extra slash after 'parameter' in the arn.
### Scenario 1: I am creating an ssm parameter in CDK. I have to use `simple_name=True`  because i need to use a token for the prefix. I pass the parameter path to an IAM policy using the parameter method `parameter_arn `.

```python
my_parameter = aws_ssm.StringParameter(
            construct,
            "ssm-my-parameter",
            string_value="SOME_VALUE",
            description="Some Description",
            parameter_name=f"/{ssm_prefix}/path/for/my/parameter",
            simple_name=True,
        )
```
After creating the `StringParameter`, the `parameter_arn` property is used in an IAM role to grant access to read the parameter:

```python
#Inside of IAM role
inline_policies={
    "ssm-policy": aws_iam.PolicyDocument(
        statements=[
            aws_iam.PolicyStatement(
                effect=aws_iam.Effect.ALLOW,
                actions=["ssm:GetParameter"],
                resources=[my_parameter.parameter_arn],
           ),
      ]
}
```

The resulting policy in the synth'd template.json looks like this:

```json
"Policies": [
    {
        "PolicyDocument": {
            "Statement": [
                {
                    "Action": [
                        "ssm:GetParameter",
                        "ssm:PutParameter"
                    ],
                    "Effect": "Allow",
                    "Resource": {
                        "Fn::Join": [
                            "",
                            [
                                "arn:",
                                {
                                    "Ref": "AWS::Partition"
                                },
                                ":ssm:",
                                {
                                    "Ref": "AWS::Region"
                                },
                                ":",
                                {
                                    "Ref": "AWS::AccountId"
                                },
                                ":parameter/",
                                {
                                    "Ref": "MyParameter"
                                }
                            ]
                        ]
                    }
                }
            ],
            "Version": "2012-10-17"
        },
        "PolicyName": "ssm-policy"
    }
]
```

Note that within the join, the `parameter` string has a `/` so it gets concatenated  with the `Ref MyParameter` (which contains a `/` already)
The resultant ARN gets an extra slash so it looks like this:

`arn:aws:ssm:us-east-1:1111111111111:parameter//prefix/for/parameter/path/for/my/parameter`

-----

### Scenario 2: I want to dynamically consume an existing SSM Parameter at deploy time of my synth'd CFN template.

Example: Use the value of a CfnParameter in the path of an SSM String Parameter lookup:

```python
id_property = CfnParameter(
    Stack.of(self),
    "IdProperty",
    description="User input specifying id of deployment",
    default="value",
    type="String",
).value_as_string

my_parameter = aws_ssm.StringParameter.from_string_parameter_attributes(
    self,
    "ssm-param",
    parameter_name=f"/prefix/{id_property}/path/to/my/parameter"
    simple_name=True,
    force_dynamic_reference=True,
)
```

In this scenario, two differing results are observed depending on how the IParameter is used.

the resulting CloudFormation Template gets a `resolve` string that contains a double `/` if the FormatArn function is used or if the IParameter is passed to another construct such as an ECS Task for use in Secrets.

When using `parameter.string_value`, the correct resolve string is observed:
```json
"Value": {
    "Fn::Join": [
        "",
        [
            "{{resolve:ssm:/prefix/",
            {
                "Ref": "IdProperty"
            },
            "/path/to/my/parameter}}"
        ]
    ]
}
```
This effectively looks like this:
`{{resolve:ssm://prefix/some-id/path/to/my/parameter}}`

However, when the IParameter is passed to the task_definition.add_container function, a second slash is observed in the resultant secret property in CFN.

```Python
task_definition.add_container(
...
secrets={
    "MY_SECRET": aws_ecs.Secret.from_ssm_parameter(my_parameter),
}
...
```

Note that the second slash is being inserted during synth.

The resultant CFN Template resolve looks like this:
```json
...
    "Secrets": [
    {
        "Name": "BACKSTAGE_NAME",
        "ValueFrom": {
            "Fn::Join": [
                "",
                [
                    "arn:aws:ssm:us-east-1:11111111111:parameter//prefix/",
                    {
                         "Ref": "IdProperty"
                    },
                    "/path/to/my/parameter"
               ]
           ]
       }
    }
]
```

There are two issues here, the first is that the account and region are being hardcoded despite the string being dynamic.  The second is that there is an extra slash after `parameter` in the arn.
In addition, the biggest issue is that there is a difference in behavior between the resolve string from`string_value` and how the ARN is generated.

### Expected Behavior

When using SSM Parameter/IParameter constructs with `simple_name=True`, the construct should not add an additional `/` before the ARN/resolve strings.

### Current Behavior

Included in bug description.

### Reproduction Steps

Included in bug description

### Possible Solution

For backwards compatibility, there could be a new additional property used in conjunction with `simple_name=True` to allow setting the construct to not add the slash.

Another solution could be removing the validation that there are no tokens in SSM parameters when not using the 'simple_name' parameter

_No response_

### CDK CLI Version

2.117.0

### Framework Version

_No response_

### Node.js Version

v18.17.1

### OS

OSX Sonoma

### Language

Python

### Language Version

3.10.11

### Other information

_No response_

GavinZZ:
### Issue # (if applicable)

Closes #28778.

### Reason for this change

There are issues with SSM `StringParameter` where the parameter ARN would have missing `/` or duplicate `/` depending on the setup of `simpleName` with unresolved tokens in the parameter name.

### Description of changes

Update README and docstring to explain to users when and how to correctly use `simpleName` parameter.

No code changes made.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-bedrock/lib/foundation-model.ts
+++ b/packages/aws-cdk-lib/aws-bedrock/lib/foundation-model.ts
@@ -41,6 +41,9 @@ export class FoundationModelIdentifier {
   /** Base model "amazon.titan-embed-text-v2:0". */
   public static readonly AMAZON_TITAN_EMBED_TEXT_V2_0 = new FoundationModelIdentifier('amazon.titan-embed-text-v2:0');
 
+  /** Base model "amazon.titan-embed-text-v2:0:8k". */
+  public static readonly AMAZON_TITAN_EMBED_TEXT_V2_0_8K = new FoundationModelIdentifier('amazon.titan-embed-text-v2:0:8k');
+
   /** Base model "amazon.titan-image-generator-v1". */
   public static readonly AMAZON_TITAN_IMAGE_GENERATOR_G1_V1 = new FoundationModelIdentifier('amazon.titan-image-generator-v1');
 
@@ -65,11 +68,17 @@ export class FoundationModelIdentifier {
   /** Base model "ai21.j2-ultra-v1". */
   public static readonly AI21_LABS_JURASSIC_2_ULTRA_V1 = new FoundationModelIdentifier('ai21.j2-ultra-v1');
 
+  /** Base model "ai21.j2-ultra-v1:0:8k". */
+  public static readonly AI21_LABS_JURASSIC_2_ULTRA_V1_0_8K = new FoundationModelIdentifier('ai21.j2-ultra-v1:0:8k');
+
   /** Base model "ai21.j2-grande-instruct". */
   public static readonly AI21_J2_GRANDE_INSTRUCT = new FoundationModelIdentifier('ai21.j2-grande-instruct');
 
   /** Base model "ai21.j2-jumbo-instruct". */
   public static readonly AI21_J2_JUMBO_INSTRUCT = new FoundationModelIdentifier('ai21.j2-jumbo-instruct');
+
+  /** Base model "ai21.jamba-instruct-v1:0". */
+  public static readonly AI21_J2_JAMBA_INSTRUCT_V1_0 = new FoundationModelIdentifier('ai21.jamba-instruct-v1:0');
 
   /**
    * Base model "anthropic.claude-v1".
@@ -138,7 +147,7 @@ export class FoundationModelIdentifier {
   public static readonly COHERE_COMMAND_LIGHT_TEXT_V14_7_4K = new FoundationModelIdentifier('cohere.command-light-text-v14:7:4k');
 
   /** Base model "cohere.command-r-v1:0". */
-  public static readonly COHERE_COMMAND_R_V1 = new FoundationModelIdentifier('ccohere.command-r-v1:0');
+  public static readonly COHERE_COMMAND_R_V1 = new FoundationModelIdentifier('cohere.command-r-v1:0');
 
   /** Base model "cohere.command-r-v1:0". */
   public static readonly COHERE_COMMAND_R_PLUS_V1 = new FoundationModelIdentifier('cohere.command-r-plus-v1:0');
@@ -146,8 +155,14 @@ export class FoundationModelIdentifier {
   /** Base model "cohere.embed-english-v3". */
   public static readonly COHERE_EMBED_ENGLISH_V3 = new FoundationModelIdentifier('cohere.embed-english-v3');
 
+  /** Base model "cohere.embed-english-v3:0:512". */
+  public static readonly COHERE_EMBED_ENGLISH_V3_0_512 = new FoundationModelIdentifier('cohere.embed-english-v3:0:512');
+
   /** Base model "cohere.embed-multilingual-v3". */
   public static readonly COHERE_EMBED_MULTILINGUAL_V3 = new FoundationModelIdentifier('cohere.embed-multilingual-v3');
+
+  /** Base model "cohere.embed-multilingual-v3:0:512". */
+  public static readonly COHERE_EMBED_MULTILINGUAL_V3_0_512 = new FoundationModelIdentifier('cohere.embed-multilingual-v3:0:512');
 
   /** Base model "meta.llama2-13b-v1". */
   public static readonly META_LLAMA_2_13B_V1 = new FoundationModelIdentifier('meta.llama2-13b-v1');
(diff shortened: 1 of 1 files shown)


packages/aws-cdk-lib/aws-ssm/README.md
```diff
@@ -143,3 +142,32 @@ When specifying an `allowedPattern`, the values provided as string literals
 are validated against the pattern and an exception is raised if a value
 provided does not comply.
 
+## Using Tokens in parameter name
+
+When using [CDK Tokens](https://docs.aws.amazon.com/cdk/v2/guide/tokens.html) in parameter name,
+you need to explicitly set `simpleName` property. Setting `simpleName` to incorrect boolean
+value may result in unexpected behaviours, such as getting duplicate '/' in the parameter ARN
+or missing '/' in the parameter ARN.
```
paulhcsun, line 150:
```suggestion
you need to explicitly set the `simpleName` property. Setting `simpleName` to an incorrect boolean
value may result in unexpected behaviours such as having a duplicate '/' in the parameter ARN
or missing a '/' in the parameter ARN.
```

packages/aws-cdk-lib/aws-ssm/README.md
```diff
@@ -143,3 +142,32 @@ When specifying an `allowedPattern`, the values provided as string literals
...
+
+`simpleName` is used to indicates whether the parameter name is a simple name. A parameter name
+without any '/' is considered a simple name, thus you should set `simpleName` to `true`.
+If the parameter name includes '/', set `simpleName` to `false`.
+
+```ts
+import * as lambda from 'aws-cdk-lib/aws-lambda';
+
+const simpleParameter = new ssm.StringParameter(this, 'StringParameter', {
+  // the parameter name is simple
```
paulhcsun, line 160:
```suggestion
  // the parameter name does not contain any '/'
```

packages/aws-cdk-lib/aws-ssm/README.md
```diff
@@ -143,3 +142,32 @@ When specifying an `allowedPattern`, the values provided as string literals
...
+  parameterName: 'parameter',
+  stringValue: 'SOME_VALUE',
+  simpleName: true, // set `simpleName` to true
+});
+
+declare const func: lambda.IFunction;
+const nonSimpleParameter = new ssm.StringParameter(this, 'StringParameter', {
+  // the parameter name is not simple
```
paulhcsun, line 168:
```suggestion
  // the parameter name contains '/'
```

packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
```diff
@@ -99,8 +99,10 @@ export interface ParameterOptions {
   readonly parameterName?: string;
 
   /**
-   * Indicates if the parameter name is a simple name (i.e. does not include "/"
-   * separators).
+   * Indicates whether the parameter name is a simple name. A parameter name
+   * without any "/" is considered a simple name. If the parameter name includes
+   * "/", setting simpleName to true might cause unintended issues, such
```
paulhcsun, line 104:
```suggestion
   * "/", setting simpleName to true might cause unintended issues such
```

packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
```diff
@@ -337,8 +339,10 @@ export interface CommonStringParameterAttributes {
   readonly parameterName: string;
 
   /**
-   * Indicates if the parameter name is a simple name (i.e. does not include "/"
-   * separators).
+   * Indicates whether the parameter name is a simple name. A parameter name
+   * without any "/" is considered a simple name. If the parameter name includes
+   * "/", setting simpleName to true might cause unintended issues, such
```
paulhcsun, line 344:
```suggestion
   * "/", setting simpleName to true might cause unintended issues such
```

paulhcsun (commented):
Just a few minor changes. Is there any validation present that checks if the `simpleName` boolean value is appropriate for the given `parameterName`? If not should we add some to enforce it or are there cases where we want to allow it?

GavinZZ:
> Just a few minor changes. Is there any validation present that checks if the `simpleName` boolean value is appropriate for the given `parameterName`? If not should we add some to enforce it or are there cases where we want to allow it?

There're codes to verify is `simpleName` is appropriate to use. If not will raise exceptions. But this is to handle the case of unresolved token so we can't do much verifications.

paulhcsun:
> Ideally it would be nice to throw a warning at synthesize time, but that's not possible because this issue only happens with unresolved tokens in parameter name. We can't determine whether the unresolved token contains `/` or not until later stage in the app life cycle. Thus, updating the docstring and README.

Ohh gotcha, ok that should be fine then.

paulhcsun (approved):

GavinZZ:
@mergify update

PR #30660, opened 2024-06-25
feat(eks): support CUSTOM amiType

### LINKED ISSUE #30641: aws-eks: Support `CUSTOM` AMI type on Nodegroups
### Describe the feature

CloudFormation supports passing `CUSTOM` as value for `AmiType`: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-eks-nodegroup.html#cfn-eks-nodegroup-amitype

This is not available in CDK, making the use of custom AMIs impossible.

### Use Case

When building custom AMIs for a cluster (e.g. based on the EKS optimized images) you should be able to use them.

### Proposed Solution

Rough idea:
- add the `CUSTOM` type to the enum: https://github.com/aws/aws-cdk/blob/3b95777ee5dae32fd41481d81922d07c804a2c6b/packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts#L29
- modify https://github.com/aws/aws-cdk/blob/3b95777ee5dae32fd41481d81922d07c804a2c6b/packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts#L635 to append `CUSTOM` as the last option for all architectures

It looks like this should already be enough, but I didn't actually attempt doing this.

### Other Information

I tried forcing the issue by bypassing TypeScript checks: `amiType: ('CUSTOM' as any)` - internal check rejected it. After investigating a little - the `Nodegroup` constructor tries to do the right thing and validates the value, but completely ignores the existence of `CUSTOM` in CloudFormation.

### Acknowledgements

### CDK version used

2.133.0

Ubuntu 22.04.4 LTS

Artemigos:
### Issue # (if applicable)

Closes #30641.

### Reason for this change

Node groups with custom AMI are supported by CloudFormation, but not by CDK. This adds that support.

### Description of changes

- added `CUSTOM` to the `NodegroupAmiType` enum
- always add `CUSTOM` to `possibleAmiTypes` when validating `instanceTypes` CPU architecture against `amiType`
- verify that `launchTemplateSpec` is defined when using `CUSTOM`

Added unit tests checking valid deployment as well as failure when `CUSTOM` is used, but `launchTemplateSpec` isn't defined.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts
+++ b/packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts
@@ -432,7 +432,7 @@ export class Nodegroup extends Resource implements INodegroup {
        * 1. instance types of different CPU architectures are not mixed(e.g. X86 with ARM).
        * 2. user-specified amiType should be included in `possibleAmiTypes`.
        */
-      possibleAmiTypes = getPossibleAmiTypes(instanceTypes).concat(NodegroupAmiType.CUSTOM);
+      possibleAmiTypes = getPossibleAmiTypes(instanceTypes);
 
       // if the user explicitly configured an ami type, make sure it's included in the possibleAmiTypes
       if (props.amiType && !possibleAmiTypes.includes(props.amiType)) {
@@ -662,5 +662,5 @@ function getPossibleAmiTypes(instanceTypes: InstanceType[]): NodegroupAmiType[] 
     throw new Error('instanceTypes of different architectures is not allowed');
   }
 
-  return archAmiMap.get(Array.from(architectures)[0])!;
+  return archAmiMap.get(Array.from(architectures)[0])!.concat(NodegroupAmiType.CUSTOM);
 }


packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts
```diff
@@ -426,7 +430,7 @@ export class Nodegroup extends Resource implements INodegroup {
        * 1. instance types of different CPU architectures are not mixed(e.g. X86 with ARM).
        * 2. user-specified amiType should be included in `possibleAmiTypes`.
        */
-      possibleAmiTypes = getPossibleAmiTypes(instanceTypes);
+      possibleAmiTypes = getPossibleAmiTypes(instanceTypes).concat(NodegroupAmiType.CUSTOM);
```
lpizzinidev, line 433: Can you please concatenate inside the method for reusability?

packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts
```diff
@@ -442,6 +446,11 @@ export class Nodegroup extends Resource implements INodegroup {
       }
     }
 
+    // custom AMI type can be used only when there's a launch template that picks an AMI
+    if (props.amiType === NodegroupAmiType.CUSTOM && !props.launchTemplateSpec) {
```
lpizzinidev, line 450: Can you please update the comments [here](https://github.com/aws/aws-cdk/blob/da2ec75f6bc8f2e28b07cdf1307b22c83bb652d5/packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts#L196-L197) and [here](https://github.com/aws/aws-cdk/blob/da2ec75f6bc8f2e28b07cdf1307b22c83bb652d5/packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts#L295) accordingly?

packages/aws-cdk-lib/aws-eks/README.md
```diff
@@ -227,6 +227,27 @@ cluster.addNodegroupCapacity('custom-node-group', {
 });
 ```
 
+To use a custom AMI for the node group, you can set `amiType` to `eks.NodegroupAmiType.CUSTOM` and provide a launch template.
```
lpizzinidev, line 230:
```suggestion
To use a custom AMI for the node group, you can set `amiType` to `CUSTOM`.
A launch template must be provided.
```

lpizzinidev (changes requested):
Thanks 👍 Left some comments for minor adjustments

packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts
Artemigos: Moved :)

packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts
Artemigos: Added a little bit of clarification in both places.

packages/aws-cdk-lib/aws-eks/README.md
Artemigos: change applied

lpizzinidev (approved):
Thanks 👍

packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts
```diff
@@ -651,5 +662,5 @@ function getPossibleAmiTypes(instanceTypes: InstanceType[]): NodegroupAmiType[]
     throw new Error('instanceTypes of different architectures is not allowed');
   }
 
-  return archAmiMap.get(Array.from(architectures)[0])!;
+  return archAmiMap.get(Array.from(architectures)[0])!.concat(NodegroupAmiType.CUSTOM);
```
GavinZZ, line 665: Instead of concat here, should we just add it directly to the x8664AmiTypes, arm64AmiTypes, etc. Otherwise this line of code would be easily missed and maintainers and other contributors may not realize those amiTypes are not accurate and complete.

packages/aws-cdk-lib/aws-eks/lib/managed-nodegroup.ts
```diff
@@ -193,8 +197,8 @@ export interface NodegroupOptions {
    */
   readonly subnets?: SubnetSelection;
   /**
-   * The AMI type for your node group. If you explicitly specify the launchTemplate with custom AMI, do not specify this property, or
-   * the node group deployment will fail. In other cases, you will need to specify correct amiType for the nodegroup.
+   * The AMI type for your node group. If you explicitly specify the launchTemplate with custom AMI, either set this property to `CUSTOM` or leave
+   * it undefined, otherwise the node group deployment will fail. In other cases, you will need to specify correct amiType for the nodegroup.
```
GavinZZ, line 201: Just so for my own understanding, `CUSTOM` ami type is supported now if we leave it unset?

packages/aws-cdk-lib/aws-eks/README.md
```diff
@@ -227,6 +227,28 @@ cluster.addNodegroupCapacity('custom-node-group', {
 });
 ```
 
+To use a custom AMI for the node group, you can set `amiType` to `CUSTOM`.
+A launch template must be provided.
```
GavinZZ, line 231: Related to the comment above, if custom AMI is supported now leaving the `amiType` property unset, we should call that out in the README too.

GavinZZ (commented):
Thank you for making the changes. Left some comments and questions.

shikha372 (changes requested):
marking it as requested changes based on previous feedback provided

PR #30661, opened 2024-06-25
feat: enable pnpx with cdk init

### LINKED ISSUE #23205: (cdk init): (Does't use pnpm to install dependencies when invoked with pnpx)
### Describe the bug

When I try to init CDK with `pnpm` it starts installing dependencies using `npm`

screenshot showing npm install getting executed

### Expected Behavior

CDK should use `pnpm` to install dependencies instead of just going for `npm`

### Current Behavior

Installs package using npm

### Reproduction Steps

Create a new project with 

```sh
pnpx cdk init app --language typescript
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.53.0

### Framework Version

_No response_

### Node.js Version

16

### OS

Arch Linux

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

IanKonlog:
### Issue # (if applicable)

Closes #23205.

### Reason for this change

`cdk init` does not use `pnpm install` when creating a new project with `pnpx cdk init app --language typescript`

### Description of changes

- Add support for `pnpm install` by resolving the package manager tool used with the `cdk init` command.
- Update `cdk.json` to use a place holder for the app command depending on the package manager tool used
- Keep the default option as `npm install` in case nothing is passed

Manual testing

### Checklist

----

### DIFF
--- a/packages/aws-cdk/lib/init.ts
+++ b/packages/aws-cdk/lib/init.ts
@@ -153,10 +153,14 @@ export class InitTemplate {
 
   private async installProcessed(templatePath: string, toFile: string, language: string, project: ProjectInfo) {
     const template = await fs.readFile(templatePath, { encoding: 'utf-8' });
-    await fs.writeFile(toFile, this.expand(template, language, project));
+    let appCommandTool = 'npx';
+    if (toFile.includes('cdk.json') && await getCommand() === 'pnpm') {
+      appCommandTool = 'pnpx';
+    }
+    await fs.writeFile(toFile, this.expand(template, language, project, appCommandTool));
   }
 
-  private expand(template: string, language: string, project: ProjectInfo) {
+  private expand(template: string, language: string, project: ProjectInfo, appCommandTool: string = 'npx') {
     // eslint-disable-next-line @typescript-eslint/no-require-imports
     const manifest = require(path.join(rootDir(), 'package.json'));
     const MATCH_VER_BUILD = /\+[a-f0-9]+$/; // Matches "+BUILD" in "x.y.z-beta+BUILD"
@@ -173,6 +177,8 @@ export class InitTemplate {
         break;
     }
     return template.replace(/%name%/g, project.name)
+      // the appcommand is only added to the cdk.json file when initializing a Typescript app.
+      .replace(/%appcommandtool%/g, appCommandTool)
       .replace(/%stackname%/, project.stackName ?? '%name.PascalCased%Stack')
       .replace(/%PascalNameSpace%/, project.stackName ? camelCase(project.stackName + 'Stack', { pascalCase: true }) : '%name.PascalCased%')
       .replace(/%PascalStackProps%/, project.stackName ? (camelCase(project.stackName, { pascalCase: true }) + 'StackProps') : 'StackProps')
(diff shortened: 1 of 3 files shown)


packages/aws-cdk/lib/init.ts
```diff
@@ -359,6 +364,16 @@ async function postInstallTypescript(canUseNetwork: boolean, cwd: string) {
   }
 }
 
+async function getCommand(): Promise<string> {
+  print('Getting command...');
+  if (process.env.npm_execpath && process.env.npm_execpath.includes('pnpm')) {
+    print(`Got exec path ${chalk.green(`${process.env.npm_execpath}`)}...`);
+    print(`Got command ${chalk.green(`${'pnpm'}`)}...`);
+    return 'pnpm';
```
comcalvi, line 372:
```suggestion
    return 'pnpm';
```

comcalvi (changes requested):
Can you add a CLI integ test? It would be something like `init-typescript-app-npx`, they live here: https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/cli-integ

packages/aws-cdk/lib/init.ts
```diff
@@ -153,10 +153,14 @@ export class InitTemplate {
 
   private async installProcessed(templatePath: string, toFile: string, language: string, project: ProjectInfo) {
     const template = await fs.readFile(templatePath, { encoding: 'utf-8' });
-    await fs.writeFile(toFile, this.expand(template, language, project));
+    let appCommandTool = 'npx';
+    if (toFile.includes('cdk.json') && await getCommand() === 'pnpm') {
```
scanlonp, line 157: This is hard-coded for pnpm, but we wont support yarn or other package managers etc. Is that the approach we want?

scanlonp (commented):
So we are supporting pnpm. Where is the line for other package managers? We have custom logic for calling pnpx, but what about the next tool?

Would it be better to draw the line somewhere and create custom logic for each manager, or is there a way to map the command we get in to the command we call in init?

packages/aws-cdk/lib/init.ts
IanKonlog: per my understanding we do not want to extend this feature pass `pnpm` following the comments in the original issue I think we should just expand the support to `pnpm`. Otherwise we expose ourselves into making changes every time there is a possibility of having a new package manager. Any thoughts?

packages/aws-cdk/lib/init.ts
IanKonlog: Adding it and testing as we speak

packages/aws-cdk/lib/init.ts
```diff
@@ -359,6 +364,16 @@ async function postInstallTypescript(canUseNetwork: boolean, cwd: string) {
   }
 }
 
+async function getCommand(): Promise<string> {
+  print('Getting command...');
+  if (process.env.npm_execpath && process.env.npm_execpath.includes('pnpm')) {
```
mrgrain, line 369:
```suggestion
  if (process.env.npm_execpath?.includes('pnpm')) {
```

PR #30663, opened 2024-06-25
feat(cloudfront): s3 origin access control L2 construct

### LINKED ISSUE #21771: cloudfront: Support Origin Access Control

gracelu0:
### Issue # (if applicable)

Closes #21771 .

### Reason for this change

See [RFC](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0617-cloudfront-oac-l2.md) 

### Description of changes

See [RFC](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0617-cloudfront-oac-l2.md) 

Unit tests and integration tests

### Checklist

----

### DIFF
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-cloudfront-origins/test/integ.s3-origin-oac-ssekms.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-cloudfront-origins/test/integ.s3-origin-oac-ssekms.ts
@@ -18,8 +18,8 @@ const originAccessControl = new cloudfront.OriginAccessControl(stack, 'OriginAcc
 new cloudfront.Distribution(stack, 'Distribution', {
   defaultBehavior: {
     origin: new origins.S3Origin(bucket, {
-      originAccessControl: originAccessControl
-    })
+      originAccessControl: originAccessControl,
+    }),
   },
 });
 
--- a/packages/@aws-cdk-testing/framework-integ/test/aws-cloudfront-origins/test/integ.s3-origin-oac.ts
+++ b/packages/@aws-cdk-testing/framework-integ/test/aws-cloudfront-origins/test/integ.s3-origin-oac.ts
@@ -11,10 +11,10 @@ const stack = new cdk.Stack(app, 'cloudfront-s3-origin-oac');
 const bucket = new s3.Bucket(stack, 'Bucket');
 const originAccessControl = new cloudfront.OriginAccessControl(stack, 'OriginAccessControl');
 new cloudfront.Distribution(stack, 'Distribution', {
-  defaultBehavior: { 
+  defaultBehavior: {
     origin: new origins.S3Origin(bucket, {
-      originAccessControl: originAccessControl
-    })
+      originAccessControl: originAccessControl,
+    }),
   },
 });
 
--- a/packages/@aws-cdk/custom-resource-handlers/test/aws-cloudfront-origins/s3-origin-access-control-bucket-policy-handler.test.ts
+++ b/packages/@aws-cdk/custom-resource-handlers/test/aws-cloudfront-origins/s3-origin-access-control-bucket-policy-handler.test.ts
@@ -46,12 +46,12 @@ describe('S3 OAC bucket policy handler', () => {
   })
 });
 
-describe('updatePolicy', () => {
+describe('appendStatementToPolicy', () => {
   it('should add a new policy statement if it does not exist', () => {
     const currentPolicy = { Statement: [] };
     const policyStatementToAdd = { Sid: 'NewStatement', Effect: 'Allow', Action: 's3:GetObject', Resource: 'arn:aws:s3:::bucket/*' };
 
-    const updatedPolicy = updatePolicy(currentPolicy, policyStatementToAdd);
+    const updatedPolicy = appendStatementToPolicy(currentPolicy, policyStatementToAdd);
 
     expect(updatedPolicy.Statement).toContainEqual(policyStatementToAdd);
   });
--- a/packages/@aws-cdk/custom-resource-handlers/test/aws-cloudfront-origins/s3-origin-access-control-key-policy-handler.test.ts
+++ b/packages/@aws-cdk/custom-resource-handlers/test/aws-cloudfront-origins/s3-origin-access-control-key-policy-handler.test.ts
@@ -149,7 +149,7 @@ describe('getActions', () => {
   });
 });
 
-describe('updatePolicy', () => {
+describe('appendStatementToPolicy', () => {
   it('should add a new policy statement if it does not exist', () => {
     const currentPolicy = { Statement: [] };
     const policyStatementToAdd = { Sid: 'NewStatement', Effect: 'Allow', Action: ['kms:Decrypt'], Resource: '*' };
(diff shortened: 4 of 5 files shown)


packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-origin.ts
```diff
@@ -44,46 +108,210 @@ export class S3Origin implements cloudfront.IOrigin {
...
  * An Origin specific to a S3 bucket (not configured for website hosting).
  *
- * Contains additional logic around bucket permissions and origin access identities.
+ * Contains additional logic around bucket permissions and origin access control (via OAI or OAC).
  */
-class S3BucketOrigin extends cloudfront.OriginBase {
-  private originAccessIdentity!: cloudfront.IOriginAccessIdentity;
+abstract class S3BucketOrigin extends cloudfront.OriginBase {
+  public static withAccessIdentity(bucket: s3.IBucket, props: S3OriginProps = {}): S3BucketOrigin {
+    return new (class OriginAccessIdentity extends S3BucketOrigin {
```
GavinZZ, line 115: This nested class looks weird to me. Can you instead split them out?
```
abstract class S3BucketOrigin extends cloudfront.OriginBase {
   ...
}

class OriginAccessIdentity extends S3BucketOrigin {
   ...
}

class OriginAccessControl extends S3BucketOrigin {
  ...
}
```
And on line 90~100, you can do 
```
    else if (props.originAccessControl) {
      this.origin = new OriginAccessControl(bucket, props);
    } else if (props.originAccessIdentity) {
      this.origin = new OriginAccessIdentity(bucket, props);
    } else {
      this.origin = FeatureFlags.of(bucket.stack)
        .isEnabled(cxapi.CLOUDFRONT_USE_ORIGIN_ACCESS_CONTROL_BY_DEFAULT) ?
        OriginAccessControl(bucket, props) :
        new OriginAccessIdentity(bucket, props);
    }
```

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-origin.ts
```diff
@@ -28,12 +79,25 @@ export class S3Origin implements cloudfront.IOrigin {
...
         ...props,
-      }) :
-      new S3BucketOrigin(bucket, props);
+      });
+    } else if (props.originAccessControl) {
+      this.origin = S3BucketOrigin.withAccessControl(bucket, props);
+    } else if (props.originAccessIdentity) {
+      this.origin = S3BucketOrigin.withAccessIdentity(bucket, props);
+    } else {
+      this.origin = FeatureFlags.of(bucket.stack)
```
GavinZZ, line 96: Should you check current stack or bucket stack for cross stack usage? IMO ideally it doesn't make a difference since the Feature flag is app-level not stack-level. Just want to confirm.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-origin.ts
```diff
@@ -44,46 +108,210 @@ export class S3Origin implements cloudfront.IOrigin {
...
+        const oacReadOnlyBucketPolicyStatement = new iam.PolicyStatement(
+          {
+            sid: 'GrantOACAccessToS3',
+            effect: iam.Effect.ALLOW,
+            principals: [new iam.ServicePrincipal('cloudfront.amazonaws.com')],
+            actions,
+            resources: [this.bucket.arnForObjects('*')],
+            conditions: {
+              StringEquals: {
+                'AWS:SourceArn': `arn:${Aws.PARTITION}:cloudfront::${Aws.ACCOUNT_ID}:distribution/${distributionId}`,
```
GavinZZ, line 239: For my learning purposes, is CloudFront arn not including region?

packages/aws-cdk-lib/aws-cloudfront/README.md
```diff
@@ -38,9 +38,108 @@ new cloudfront.Distribution(this, 'myDist', {
...
+            "StringEquals": {
+                "AWS:SourceArn": "arn:aws:cloudfront::111122223333:distribution/<CloudFront distribution ID>"
+            }
+        }
+    }
+}
+```
+
+If your bucket previously used OAI, there will be a best-effort attempt to remove both the policy statement
+that allows access to the OAI and the origin access identity itself.
```
GavinZZ, line 123: Curious if the same bucket can be used by two differnt CloudFront distributions?

GavinZZ (commented):
Reviewed half of the changes. Left some questions.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-origin.ts
```diff
@@ -15,6 +28,44 @@ export interface S3OriginProps extends cloudfront.OriginProps {
    * @default - An Origin Access Identity will be created.
```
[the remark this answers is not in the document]
gracelu0: Yes, will specify the feature flag behaviour for creating OAI by default

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-origin.ts
gracelu0: This was one of the suggested changes from the RFC (see [feedback](https://github.com/aws/aws-cdk-rfcs/pull/619#discussion_r1638661763)). I agree the structure looks a bit unconventional, but what I like is that the API is very clear on lines 90-100: 
```
else if (props.originAccessControl) {
      this.origin = S3BucketOrigin.withAccessControl(bucket, props);
    } else if (props.originAccessIdentity) {
      this.origin = S3BucketOrigin.withAccessIdentity(bucket, props);
    } else {
      this.origin = FeatureFlags.of(bucket.stack)
        .isEnabled(cxapi.CLOUDFRONT_USE_ORIGIN_ACCESS_CONTROL_BY_DEFAULT) ?
        S3BucketOrigin.withAccessControl(bucket, props) :
        S3BucketOrigin.withAccessIdentity(bucket, props);
    }
```
since it is a S3 origin with either OAI or OAC configured

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-origin.ts
gracelu0: Correct, CloudFront Distribution arns don't have the region since they are considered global resources

packages/aws-cdk-lib/aws-cloudfront/README.md
gracelu0: Yes, the same S3 origin can be used by multiple CloudFront distributions. Hmm might be tricky then to only remove the OAI associated with the distribution in the current stack if we don't have the OAI ID, is that your concern?

watany-dev:
@gracelu0 @GavinZZ 
I am one of the developers eagerly awaiting the implementation of this Construct. If you are too busy to continue, I would be delighted to fork this and take over the work.

PR #30665, opened 2024-06-25
feat(pipes-targets): add CloudWatch Logs

msambol:
Add CloudWatch Logs as a Pipe target.


packages/@aws-cdk/aws-pipes-targets-alpha/README.md
```diff
@@ -122,3 +122,19 @@ const pipe = new pipes.Pipe(this, 'Pipe', {
     target: pipeTarget
 });
 ```
+
+### Amazon CloudWatch Logs Log Group
```
nmussy, line 126: I understand that CloudWatch Logs is the product name, but it sounds pretty redundant here
```suggestion
### Amazon CloudWatch Log Group
```

packages/@aws-cdk/aws-pipes-targets-alpha/lib/cloudwatch-logs.ts
```diff
@@ -0,0 +1,88 @@
...
+   */
+  readonly logStreamName?: string;
+
+  /**
+   * The time the event occurred, expressed as the number of milliseconds after Jan 1, 1970 00:00:00 UTC.
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetcloudwatchlogsparameters.html#cfn-pipes-pipe-pipetargetcloudwatchlogsparameters-timestamp
+   * @default - none
+   */
+  readonly timestamp?: string;
```
nmussy, line 31: Any reason not to have a `number` here? You can convert it back to a string in the constructor, and it reduces the chance that a user enters something like a ISO date

nmussy (changes requested):
LGTM, just a couple of nitpicks

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
nmussy: I guess there's already some precedence:
image

Logs is missing more often than not, but I think that the product name change from "CloudWatch" "CloudWatch Logs" is somewhat recent?

image

I'm not dead set on changing it, I'll leave it up to you

msambol: Definitely redundant but the service is `Amazon CloudWatch Logs` and the product is a `log group`, usually lowercase but since it's in a title I capitalized it here. I'll leave this as-is.

packages/@aws-cdk/aws-pipes-targets-alpha/lib/cloudwatch-logs.ts
msambol: I can't get this to work in the Console even:

```
[ECMA 262 regex "^\$(\.[\w/_-]+(\[(\d+|\*)\])*)*$" does not match input string "1719416220650"]
```

msambol: I'm no regex expert but I can't make sense of that 🤔

msambol: @redwheeler3 is there a disconnect between the docs and the API? or am I missing something?

redwheeler3: For this one, we expect a JSONPath expression that references the timestamp in your payload. If you try the string "$.data.timestamp" it will validate.

redwheeler3: Leaving this blank defaults to the current time.

msambol: @redwheeler3 I updated the above and added an example. I also confirmed `$.data.timestamp` worked in the Console. We should update the [API docs](https://docs.aws.amazon.com/eventbridge/latest/pipes-reference/API_PipeTargetCloudWatchLogsParameters.html).

msambol:
@Leo10Gama Could we keep rolling on these ? 😄

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.cloudwatch-logs.ts
Leo10Gama, line 1: Correct me if I'm wrong, but my understanding of this test is that we are using the Pipe to connect an SQS source with the new CloudWatch Log group target, and validating that they're connected by sending a message through the queue and logging the event on the other side. Can we add some documentation to this test so it's clear what we are doing, similar to some of the other pipe target integ tests?

Leo10Gama (changes requested):
Thanks for another contribution to the pipes targets! Just a few minor requests from my end, but overall looks good

msambol:
@Leo10Gama Fixed! Thanks for the review.

Leo10Gama (commented):
Thanks for the changes! I'm still only in disagreement on the current phrasing of the timestamp parameter, since we don't want to give the impression that it supports a value that might be a number. My understanding of how it looks now, we should also aim to describe the value itself, as opposed to exclusively defining it as being referenced by another value, if that makes sense.

packages/@aws-cdk/aws-pipes-targets-alpha/lib/cloudwatch-logs.ts
```diff
@@ -0,0 +1,93 @@
...
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetcloudwatchlogsparameters.html#cfn-pipes-pipe-pipetargetcloudwatchlogsparameters-logstreamname
+   * @default - none
+   */
+  readonly logStreamName?: string;
+
+  /**
+   * The JSON path expression that references the timestamp in the payload.
+   *
+   * This is the time the event occurred, expressed as the number of milliseconds after Jan 1, 1970 00:00:00 UTC.
```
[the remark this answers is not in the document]
Leo10Gama: It seems like the main source of confusion is that the timestamp in the payload itself is the number of milliseconds, but the parameter we provide is the JSON path expression.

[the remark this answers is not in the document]
msambol: Updated. How's that?

Leo10Gama (approved):
LGTM, thanks again for the contribution!

Leo10Gama:
@mergifyio update

PR #30674, opened 2024-06-26
feat(pipes-targets): add Redshift

msambol:
Add Redshift as a Pipes target


packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.redshift.ts
```diff
@@ -0,0 +1,74 @@
...
+});
+// not recommended, only for integration test
+sg.addIngressRule(ec2.Peer.anyIpv4(), ec2.Port.tcp(5439), 'Ingress from Internet');
+
+const targetCluster = new Cluster(stack, 'MyCluster', {
+  clusterName: 'my-cluster',
+  masterUser: { masterUsername: 'admin' },
+  nodeType: NodeType.DC2_LARGE,
+  clusterType: ClusterType.SINGLE_NODE,
+  publiclyAccessible: true, // not recommended, only for integration test
```
msambol, line 26: @redwheeler3 I assume the Redshift cluster needs to be publicly accessible in order for Pipes to hit it ?

msambol: @redwheeler3, any insight on this one? I can get the pipe to execute (I see the `BatchExecuteStatement` API call in CloudTrail), but I don't see the query run in Redshift. I think my networking/security group look good but the Redshift cluster has to be publicly accessible for Pipes to hit it, right?

redwheeler3: Yes. For now, the Redshift cluster needs to be publicly accessible.

PR #30725, opened 2024-07-01
feat(dynamodb): adding on-demand-throughput to table

### LINKED ISSUE #30091: aws-dynamodb: OnDemandThroughput configuration
### Describe the feature

Sets the maximum number of read and write units for the specified on-demand table. If you use this property, you must specify MaxReadRequestUnits, MaxWriteRequestUnits, or both.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html#cfn-dynamodb-table-ondemandthroughput

### Use Case

New DynamoDB feature not yet implemented in CDK

### Proposed Solution

Include the new parameters as soon as CFN is updated to the current version

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.122.0

MacOS

LeeroyHannigan:
### Issue # (if applicable)

Closes #30091

### Reason for this change

New DynamoDB feature that was just released

### Description of changes

Added `maxReadRequestUnits` and `maxWriteRequestUnits` for PAY_PER_REQUEST tables

Tests on `integ.dynamodb.ondemand.ts`

### Checklist

----


shikha372:
Thank you for submitting this pull request, will take a look at the implementation for any updates this week

packages/aws-cdk-lib/aws-dynamodb/lib/table-v2.ts
```diff
@@ -485,9 +525,15 @@ export class TableV2 extends TableBaseV2 {
       this.addKey(props.sortKey, RANGE_KEY_TYPE);
     }
 
-    this.billingMode = props.billing?.mode ?? BillingMode.PAY_PER_REQUEST;
-    this.readProvisioning = props.billing?._renderReadCapacity();
-    this.writeProvisioning = props.billing?._renderWriteCapacity();
+    if (props.billing?.mode === BillingMode.PAY_PER_REQUEST || props.billing?.mode === undefined) {
```
shikha372, line 528: for clarification, what we're trying to do here is check if customer has set billing mode, even if its undefined by they have the maxReadRequestUnits and maxWriteRequestUnits, billing mode should be set to `PAY_PER_REQUEST`.
Let me know if my understanding is not correct.

From the code it looks like, what will happen is even if you get undefined values from renderCapacity functions your billing mode will be set to `PAY_PER_REQUEST`. If this is not intended, instead of simply checking for undefined billing mode we should validate and throw an error that props.billing.mode should be set in order to use new props.

packages/aws-cdk-lib/aws-dynamodb/lib/table-v2.ts
```diff
@@ -636,6 +685,11 @@ export class TableV2 extends TableBaseV2 {
         ? props.readCapacity._renderReadCapacity()
         : this.readProvisioning,
       tags: props.tags,
+      readOnDemandThroughputSettings: props.maxReadRequestUnits
```
shikha372, line 688: not sure what we're trying to do here

packages/aws-cdk-lib/aws-dynamodb/lib/table-v2.ts
shikha372, line 528: To understand it better, is it intended to set billing mode to `PAY_PER_REQUEST` in case of an undefined value?

packages/aws-cdk-lib/aws-dynamodb/lib/table-v2.ts
```diff
@@ -636,6 +685,11 @@ export class TableV2 extends TableBaseV2 {
...
+        ? { maxReadRequestUnits: props.maxReadRequestUnits }
```
shikha372, line 689: why it can't be the same as `writeOnDemandThroughputSettings: this.maxWriteRequestUnits
        ? { maxWriteRequestUnits: this.maxWriteRequestUnits }
        : undefined` ?

packages/aws-cdk-lib/aws-dynamodb/lib/table-v2.ts
LeeroyHannigan: It's intended yes. PAY_PER_REQUEST is known as OnDemand mode and is the default capacity mode for DynamoDB.

packages/aws-cdk-lib/aws-dynamodb/lib/table-v2.ts
LeeroyHannigan: This is because writes limits must be synced across all replicas. Reads do not follow this rule and can be defined individually per table replica.

shikha372:
Thank you @LeeroyHannigan for the pull request, changes looks okay to me but I see no unit tests added for this new feature, can we cover that as well?

shikha372 (approved):

PR #30664, opened 2024-06-25
feat(route53): added EvaluateTargetHealth to Route53 Alias targets (#9481)

wladyslawczyzewski:
### Issue # (if applicable)

Part of a bigger piece of work for #9481 
Another PR for this ticket #30739 

### Reason for this change

At this moment constructing the Route53 health checks is available only via L1 construct, this PR is one of the series to introduce the R53 health checks via higher-level constructs.

### Description of changes

The underlying L1 construct of a few L2 constructs for Alias targets already has the `EvaluateTargetHealth` property, so by adding this prop to the L2 construct we get the quick win to introduce the health checks for Alias records.

### Description of how you validated changes

- unit tests
- integration tests

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

Issue #9481, opened 2020-08-06
[route53] Support for health checks


justin8:
Currently the only way to define a health check or even to enable health checks for alias targets appears to be using `Cfn` resources instead of the higher level constructs.

### Use Case
Route53 health checks are commonly used for multi-region architectures, and not being able to create them is a little limiting

### Other

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

shivlaks:
@justin8 I agree this is a pretty important missing feature. marking it as a `p1` to help us prioritize getting it implemented and added in.

elhedran:
I see this has been opened for a while now.  Can someone at put the work around using `Cfn` resources while we wait for a fix?

elhedran:
Trying to work around this open ticket... not sure this is right though.  Deployment seems to get stuck.

```Typescript
const alias = new route53_targets.ApiGatewayv2DomainProperties(
    domainName.regionalDomainName, domainName.regionalHostedZoneId
)
const target = route53.RecordTarget.fromAlias(alias)

// latency based routing with health check
const gwStack = core.Stack.of(gw)
const healthCheck = new route53.CfnHealthCheck(this, 'healthCheck', {
    healthCheckConfig: {
        fullyQualifiedDomainName:
            `${gw.apiId}.execute-api.${gwStack.region}.${gwStack.urlSuffix}`,
        type: 'HTTPS',
        measureLatency: true
    }
})

new route53.CfnRecordSet(this, 'aRecordSet', {
    aliasTarget: {
        dnsName: domainName.regionalDomainName,
        hostedZoneId: domainName.regionalHostedZoneId,
        evaluateTargetHealth: true
    },
    name: siteName,
    hostedZoneId: hostedZone.hostedZoneId,
    resourceRecords: target.values,
    type: route53.RecordType.A,
    region: gwStack.region,
    setIdentifier: id,
    healthCheckId: healthCheck.attrHealthCheckId
})
```

I'll update if I work out something better, but wanted to put what I had so far down given this has been open a while now

Fitmavincent:
Before this feature is available in the new release, do we have any way to associate a Cloudwatch Alarm with the Route53 Health Check in CDK?

OGoodness:
Any update on this?

pflorek:
[...]

PR #30739, opened 2024-07-02
feat(route53): added L2 construct for Route53's health checks

wladyslawczyzewski:
### Issue # (if applicable)

Part of #9481 
Another PR for this ticket #30664 

### Reason for this change

Added L2 construct for `AWS::Route53::HealthCheck` resource

### Description of changes

The changes only introduces the L2 construct for Route53 health check resources. Except the L2 construct itself, I added basic validations for the input props.

- unit tests
- integration tests

### Checklist

----


packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+      rulePortIsNotAllowed(props);
+      ruleEnableSNIIsNotAllowed(props);
+      ruleSearchStringIsNotAllowed(props);
+      ruleChildHealthChecksIsNotAllowed(props);
+      ruleRoutingControlIsNotAllowed(props);
+      ruleHealthThresholdIsNotAllowed(props);
+      break;
+    }
+    default:
+      throw new Error(`Unsupported health check type: ${props.type}`);
```
nmussy, line 305: Switching to a map would also remove the need for this exception, since the `rules` type will require that all possible `HealthCheckType` values are covered

packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -220,6 +220,37 @@ Constructs are available for A, AAAA, CAA, CNAME, MX, NS, SRV and TXT records.
 Use the `CaaAmazonRecord` construct to easily restrict certificate authorities
 allowed to issue certificates for a domain to Amazon only.
 
+### Health Checks
+
+You can associate health checks with records:
```
nmussy, line 225: A brief description of what health checks are and what their purpose is would be ideal here, even if failovers aren't supported yet

packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -220,6 +220,37 @@ Constructs are available for A, AAAA, CAA, CNAME, MX, NS, SRV and TXT records.
...
+  weight: 100,
+});
+new route53.ARecord(this, 'ARecord2', {
+  zone: myZone,
+  target: route53.RecordTarget.fromIpAddresses('5.6.7.8'),
+  weight: 0,
+});
+```
+
+Possible values for `type` are `HTTP`, `HTTPS`, `TCP`, `CLOUDWATCH_METRIC`, `CALCULATED` and `RECOVERY_CONTROL`.
```
nmussy, line 252: I would just add a link to the `HealthCheckType` enum doc page. It will remain up to date, and have the corresponding description for each type:

```suggestion
See [`HealthCheckType`](aws-cdk-lib.aws_route53.HealthCheckType.html) for possible values.
```

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+  constructor(scope: Construct, id: string, props: HealthCheckProps) {
+    super(scope, id);
+
+    validateProperties(props);
+
+    const resource = new CfnHealthCheck(this, 'Resource', {
+      healthCheckConfig: {
+        type: props.type,
+        alarmIdentifier: props.alarmIdentifier,
+        childHealthChecks: props.childHealthChecks,
```
nmussy, line 197: Even though this is not well documented, I'm assuming that `childHealthChecks` is a list of `Ref`s/`healthCheckId`s. I would add an integration test to verify this assumption, and change this type from `string[]` to `IHealthCheck[]`. CDK users will then be able to create multiple health checks in the same stack, or import them with `fromHealthCheckId`.

This should also be able to handle dependency resolution (e.g. if you create both `hc1` and `hc2`, with `hc2` being a child of `hc1`,) but again, that's something you can verify with an integration test

packages/@aws-cdk-testing/framework-integ/test/aws-route53/test/integ.health-check.ts
```diff
@@ -0,0 +1,75 @@
...
+  recordName: '_bar',
+  target: route53.RecordTarget.fromValues('5.6.7.8'),
+  weight: 0,
+});
+
+new IntegTest(app, 'integ-test', {
+  testCases: [stack],
+  diffAssets: true,
+  enableLookups: true,
+});
```
nmussy, line 75: I would add more properties to your health checks, and make assertions to make sure all the health checks properties are correctly set, see [AwsApiCall](https://docs.aws.amazon.com/cdk/api/v2/docs/integ-tests-alpha-readme.html#api-calls). This way you'll be able to ensure that everything is able to be set properly

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+   * A complex type that contains one ChildHealthCheck element for each health check that you want to associate with a CALCULATED health check.
+   *
+   * @default - no child health checks
+   */
+  readonly childHealthChecks?: string[];
+
+  /**
+   * Specify whether you want Amazon Route 53 to send the value of FullyQualifiedDomainName to the endpoint in the client_hello message during TLS negotiation. This allows the endpoint to respond to HTTPS health check requests with the applicable SSL/TLS certificate.
+   *
+   * @default - not configured
```
nmussy, line 43: I strongly suspect the default is `false` here, but this is not made clear in either the Cfn nor the API documentation

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+   * The status of the health check when CloudWatch has insufficient data about the state of associated alarm.
+   *
+   * @default - not configured
+   */
+  readonly insufficientDataHealthStatus?: InsufficientDataHealthStatusEnum;
+
+  /**
+   * Specify whether you want Amazon Route 53 to invert the status of a health check, so a health check that would normally be considered unhealthy is considered healthy, and vice versa.
+   *
+   * @default - not configured
```
nmussy, line 91:
```suggestion
   * @default false
```

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+   * The IPv4 or IPv6 IP address for the endpoint that you want Amazon Route 53 to perform health checks on. If you don't specify a value for IPAddress, Route 53 sends a DNS request to resolve the domain name that you specify in FullyQualifiedDomainName at the interval that you specify in RequestInterval. Using an IPv4 address that DNS returns, Route 53 then checks the health of the endpoint.
+   *
+   * @default - not configured
+   */
+  readonly ipAddress?: string;
+
+  /**
+   * Specify whether you want Amazon Route 53 to measure the latency between health checkers in multiple AWS regions and your endpoint, and to display CloudWatch latency graphs on the Health Checks page in the Route 53 console.
+   *
+   * @default - not configured
```
nmussy, line 105: Again, not explicitly said, but I'm going to assume the default is `false`

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+
+  /**
+   * CloudWatch alarm that you want Amazon Route 53 health checkers to use to determine whether the specified health check is healthy.
+   *
+   * @default - no alarm identifier
+   */
+  readonly alarmIdentifier?: AlarmIdentifier;
+
+  /**
+   * A complex type that contains one ChildHealthCheck element for each health check that you want to associate with a CALCULATED health check.
```
nmussy, line 34: "Complex types" don't make a ton of sense in the CDK, this is just an array

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+   * The port on the endpoint that you want Amazon Route 53 to perform health checks on.
+   *
+   * @default - not configured
+   */
+  readonly port?: number;
+
+  /**
+   * A complex type that contains one Region element for each region from which you want Amazon Route 53 health checkers to check the specified endpoint.
+   *
+   * @default - not configured
```
nmussy, line 119: The default value is documented in the Cfn docs. Also, "complex type" and "Region element" sound a bit off here. I assume it's just "an array of region identifiers", but you can add this to your integration tests to make sure

```suggestion
   * An array of region identifiers from which you want Amazon Route 53 health checkers to check the specified endpoint.
   *
   * @default - performs checks from all of the regions that are listed under Valid Values.
```

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+   */
+  readonly regions?: string[];
+
+  /**
+   * The number of seconds between the time that Amazon Route 53 gets a response from your endpoint and the time that it sends the next health check request. Each Route 53 health checker makes requests at this interval.
+   *
+   * @default 30
```
nmussy, line 126: "30" is not a valid `Duration`, and we do ask CDK users for a number of seconds here, just a duration

```suggestion
   * The duration between the time that Amazon Route 53 gets a response from your endpoint and the time that it sends the next health check request. Each Route 53 health checker makes requests at this interval.
   *
   * @default Duration.seconds(30)
```

nmussy (changes requested):
Apologies for the number of requested changes, but it's a really good first pass. I would just be mindful of copy-pasting CloudFormation docs for the property JSDocs, some of them don't make as much sense in the context of the CDK 👍

packages/aws-cdk-lib/aws-route53/test/record-set.test.ts
```diff
@@ -188,6 +188,36 @@ describe('record set', () => {
...
+    // WHEN
+    new route53.ARecord(stack, 'Alias', {
+      zone,
+      recordName: '_foo',
+      target: route53.RecordTarget.fromValues('1.2.3.4'),
+      healthCheck,
+    });
+
+    // THEN
+    Template.fromStack(stack).hasResourceProperties('AWS::Route53::RecordSet', {
```
nmussy, line 214: I would also add an assertion for a `AWS::Route53::HealthCheck`

packages/aws-cdk-lib/aws-route53/test/record-set.test.ts
```diff
@@ -188,6 +188,36 @@ describe('record set', () => {
     });
   });
 
+  test('A record with health check', () => {
```
nmussy, line 191: I would also add a test to make sure that imported health checks can be properly mapped to Record sets

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+
+    validateProperties(props);
+
+    const resource = new CfnHealthCheck(this, 'Resource', {
+      healthCheckConfig: {
+        type: props.type,
+        alarmIdentifier: props.alarmIdentifier,
+        childHealthChecks: props.childHealthChecks,
+        enableSni: props.enableSNI,
+        failureThreshold: props.failureThreshold ?? 3,
```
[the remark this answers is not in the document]
nmussy: My main issue is that it's an optional Cfn parameter, and by having a default value here you are preventing CDK users from being able to pass a null-ish value. I'll let the maintainer weigh in here, I don't think it's a huge issue either way

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+        fullyQualifiedDomainName: props.fqdn,
+        healthThreshold: props.healthThreshold,
+        insufficientDataHealthStatus: props.insufficientDataHealthStatus,
+        inverted: props.inverted,
+        ipAddress: props.ipAddress,
+        measureLatency: props.measureLatency,
+        port: props.port,
+        regions: props.regions,
+        requestInterval: (props.requestInterval && props.requestInterval.toSeconds()) ?? 30,
```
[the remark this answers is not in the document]
nmussy: See https://github.com/aws/aws-cdk/pull/30739#discussion_r1671724994

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  if (
+    !new RegExp(
+      '^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]).){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))$|^(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]).){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]).){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))$',
+    ).test(props.ipAddress)
+  ) {
+    throw new Error('IpAddress must be a valid IPv4 or IPv6 address');
+  }
+}
```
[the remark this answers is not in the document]
nmussy: Again, I'll leave it up to the maintainer review

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,517 @@
...
+    throw new Error('FQDN must be between 0 and 255 characters long');
+  }
+}
+
+function validateSearchStringForStringMatch(props: HealthCheckProps) {
+  if (props.searchString === null || props.searchString === undefined) {
+    throw new Error(`SearchString is required for health check type: ${props.type}`);
+  }
+
+  if (props.searchString.length === 0 || props.searchString.length > 255) {
```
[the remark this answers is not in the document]
nmussy: You could simplify the tests by doing the following, the first error message also implying a non-empty string value:

```ts
  if (!props.searchString {
    throw new Error(`SearchString is required for health check type: ${props.type}`);
  }

  if (props.searchString.length > 255) {
    throw new Error('SearchString length cannot be greater than 255 characters');
  }
```

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,670 @@
...
+  readonly port?: number;
+
+  /**
+   * An array of region identifiers that you want Amazon Route 53 health checkers to check the health of the endpoint from.
+   *
+   * Valid values are: 'us-east-1', 'us-west-1', 'us-west-2', 'eu-west-1', 'ap-southeast-1', 'ap-southeast-2', 'ap-northeast-1', 'sa-east-1'
+   * @default - if the type is CALCULATED, CLOUDWATCH_METRIC, or RECOVERY_CONTROL, this property is not configured.
+   * - otherwise, the default value is all valid regions: 'us-east-1', 'us-west-1', 'us-west-2', 'eu-west-1', 'ap-southeast-1', 'ap-southeast-2', 'ap-northeast-1', 'sa-east-1'
+   */
+  readonly regions?: ('us-east-1' | 'us-west-1' | 'us-west-2' | 'eu-west-1' | 'ap-southeast-1' | 'ap-southeast-2' | 'ap-northeast-1' | 'sa-east-1')[];
```
nmussy, line 192: I have the same maintainability concerns for this JSDoc and this type, I would just revert this type to a `string[]` and add a link to the Cfn documentation.

Just as an FYI, you could have used a `const` assertion to derive this type from the `VALID_REGIONS` array, see this [example Playground](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAagQQDIEkAiB9ASgUQOIoDyAcgMowC8MA2gOQCuEAtAKYCG0TAjLQDQwNmAdxaceAXRgcYoSFADcAKGUBLMFBYAnAGZtgLGAAVNIAA4QYAb0UxbMTSwDmK8BABcMABRQAnqZYg2vDI6Nj4RGTUYPQAtgBGWuIAlNTiSgC+yrLQMABubAA2KgAmHsZmFlSWDs6uHnSMrBxQ3LTi6UrZsGr5RaVGJuaUVjUukPWCTZwATG0dQA)

packages/@aws-cdk-testing/framework-integ/test/aws-route53/test/integ.health-check.ts
nmussy: The integ looks great, thanks for being thorough 👍

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
nmussy: Again, I don't think setting a default value in the CDK is that useful

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
nmussy: See comment about setting the default value: https://github.com/aws/aws-cdk/pull/30739#discussion_r1671755028

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
nmussy: See comment about setting the default value: https://github.com/aws/aws-cdk/pull/30739#discussion_r1671755028

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
nmussy: Thanks for the doc update, I've added some comments about the typing and validation though:
https://github.com/aws/aws-cdk/pull/30739/files#r1671746172
https://github.com/aws/aws-cdk/pull/30739/files#r1671739388

nmussy (changes requested):
Thanks for the changes, I had a few more comments

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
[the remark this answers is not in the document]
wladyslawczyzewski: @nmussy thanks for this, updated the implementation ✅

wladyslawczyzewski:
hi @nmussy , thanks for you review!

I made the regions weak-typed `string[]` property and as result passing all the validation responsibility to CloudFormation itself, however, I think having this prop strong-typed will be better option even if this will require more maintainer work in the future – not all AWS' regions are used for health checks, therefore user may provide the wrong value and will get the error only on deployment phase.

On the default values stuff – you're having a valid points why to not provide default values in the CDK codebase, however, each of that properties have implicit default values, imho, it's better to let users know what the default value is and if they indeed want to have the resource configured differently, they will need to provide their configuration explicitly in the code rather then relying on CloudFormation hidden decisions.

I pushed few changes according to your last review, could you please take a look and let me know if you see anything else that could be improved?

packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -220,6 +220,42 @@ Constructs are available for A, AAAA, CAA, CNAME, MX, NS, SRV and TXT records.
...
+new route53.ARecord(this, 'ARecord', {
+  zone: myZone,
+  target: route53.RecordTarget.fromIpAddresses('1.2.3.4'),
+  healthCheck,
+  weight: 100,
+});
+new route53.ARecord(this, 'ARecord2', {
+  zone: myZone,
+  target: route53.RecordTarget.fromIpAddresses('5.6.7.8'),
+  weight: 0,
```
nmussy, line 253: I'm not sure using a zero weight here is the best example, it's a pretty complex behavior with a few pitfalls, see [Weight](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-weighted.html#rrsets-values-weighted-weight) and [How Amazon Route 53 chooses records when health checking is configured](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/health-checks-how-route-53-chooses-records.html)

```suggestion
  weight: 50,
```

nmussy (approved):
LGTM, I'll let the maintainer make a decision regarding the default values. I'm also not sure using `weight: 0` is a great example, see comment

packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -220,6 +220,42 @@ Constructs are available for A, AAAA, CAA, CNAME, MX, NS, SRV and TXT records.
 Use the `CaaAmazonRecord` construct to easily restrict certificate authorities
 allowed to issue certificates for a domain to Amazon only.
 
+### Health Checks
+
+Route 53 has the ability to monitor the health of your application and only return records for healthy endpoints.
+This is done using a `HealthCheck` construct.
+
+In the following example, the `ARecord` will be returned by Route 53in response to DNS queries only if the HTTP requests to the `example.com/health` endpoint return a 2XX or 3XX status code.
```
shikha372, line 228:
```suggestion
In the following example, the `ARecord` will be returned by Route 53 in response to DNS queries only if the HTTP requests to the `example.com/health` endpoint return a 2XX or 3XX status code.
```

packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -220,6 +220,42 @@ Constructs are available for A, AAAA, CAA, CNAME, MX, NS, SRV and TXT records.
 Use the `CaaAmazonRecord` construct to easily restrict certificate authorities
 allowed to issue certificates for a domain to Amazon only.
 
+### Health Checks
+
+Route 53 has the ability to monitor the health of your application and only return records for healthy endpoints.
+This is done using a `HealthCheck` construct.
```
shikha372, line 226:
```suggestion
In CDK, this can be done using `HealthCheck` construct.
```

packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -220,6 +220,42 @@ Constructs are available for A, AAAA, CAA, CNAME, MX, NS, SRV and TXT records.
 Use the `CaaAmazonRecord` construct to easily restrict certificate authorities
 allowed to issue certificates for a domain to Amazon only.
 
+### Health Checks
+
+Route 53 has the ability to monitor the health of your application and only return records for healthy endpoints.
```
shikha372, line 225: Please link the documentation to route53 health check

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,658 @@
...
+        routingControlArn: props.routingControl,
+        searchString: props.searchString,
+      },
+    });
+
+    this.healthCheckId = resource.ref;
+  }
+}
+
+const validationRules: Record<HealthCheckType, ((props: HealthCheckProps) => void)[]> = {
```
shikha372, line 292: instead of having all these functions to validate these scenarios, can we include something like a enum based properties or class to define these props, i think one of the [design principles](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#property-documentation) we follow is to prevent many check cases.

The fewer “if statements” the better.

packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -220,6 +220,42 @@ Constructs are available for A, AAAA, CAA, CNAME, MX, NS, SRV and TXT records.
...
+  weight: 100,
+});
+new route53.ARecord(this, 'ARecord2', {
+  zone: myZone,
+  target: route53.RecordTarget.fromIpAddresses('5.6.7.8'),
+  weight: 0,
+});
+```
+
+See the [Route 53 Health Checks documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-healthcheck-healthcheckconfig.html#cfn-route53-healthcheck-healthcheckconfig-type) for possible types of health checks.
```
shikha372, line 257: I would suggest adding these documentation links upward so that its easier to get the overview before looking at the implementation in CDK

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
wladyslawczyzewski: sorry @shikha372 could you please explain this further?

packages/aws-cdk-lib/aws-route53/README.md
wladyslawczyzewski: @shikha372 moved this link to the top of the Health Checks section

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
shikha372: referring to all the validation functions defined with each healthcheck type eg. 
```
function ruleAlarmIdentifierIsNotAllowed(props: HealthCheckProps) {
  if (props.alarmIdentifier) {
    throw new Error(`AlarmIdentifier is not supported for health check type: ${props.type}`);
  }
}
```
instead of validating these checks with if-else, can we introduce a enum type class or method which will allow users to set only the required fields possible under a healthcheck, let me know if you think this is not feasible

packages/aws-cdk-lib/aws-route53/README.md
```diff
@@ -223,6 +223,42 @@ Constructs are available for A, AAAA, CAA, CNAME, MX, NS, SRV and TXT records.
...
 allowed to issue certificates for a domain to Amazon only.
 
+### Health Checks
+
+See the [Route 53 Health Checks documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-healthcheck-healthcheckconfig.html#cfn-route53-healthcheck-healthcheckconfig-type) for possible types of health checks.
+
+Route 53 has the ability to monitor the health of your application and only return records for healthy endpoints.
+This is done using a `HealthCheck` construct.
+
+In the following example, the `ARecord` will be returned by Route 53in response to DNS queries only if the HTTP requests to the `example.com/health` endpoint return a 2XX or 3XX status code.
```
GavinZZ, line 233:
```suggestion
In the following example, the `ARecord` will be returned by Route 53 in response to DNS queries only if the HTTP requests to the `example.com/health` endpoint return a 2XX or 3XX status code.
```

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,658 @@
...
+  return undefined;
+}
+
+function getDefaultFailureThresholdForType(type: HealthCheckType): number | undefined {
+  switch (type) {
+    case HealthCheckType.CALCULATED:
+    case HealthCheckType.CLOUDWATCH_METRIC:
+      return undefined;
+    default:
+      return 3;
```
GavinZZ, line 442: Is the 3 here cloudformation default? i.e. even if you set it to `undefined` for this type of health check, CFN will default it to 3?

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,658 @@
...
+  }
+}
+
+function getDefaultRequestIntervalForType(type: HealthCheckType): Duration | undefined {
+  switch (type) {
+    case HealthCheckType.CALCULATED:
+    case HealthCheckType.CLOUDWATCH_METRIC:
+      return undefined;
+    default:
+      return Duration.seconds(30);
```
GavinZZ, line 452: same as above.

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,658 @@
...
+    ruleRoutingControlIsNotAllowed,
+    ruleHealthThresholdIsNotAllowed,
+    validateFqdn,
+    validateIpAddress,
+    validateEitherFqdnOrIpAddressRequired,
+    validateRequestInterval,
+    validateFailureThreshold,
+    validateRegions,
+  ],
+  [HealthCheckType.HTTP_STR_MATCH]: [
```
GavinZZ, line 320: I appreciate the fact that you added all these validations on CDK side, however, this also makes me a bit worried at the same time. If CloudFormation decides to change one or some of the validation rules, we need to make changes accordingly and given there are so many, it's not straightforward to update this.

How many of these validations rule are essential? Can we only validate on the essential ones and leave the rest to CFN?

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
```diff
@@ -0,0 +1,658 @@
...
+    return;
+  }
+
+  if (
+    !new RegExp(
+      '^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]).){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))$|^(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]).){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]).){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))$',
+    ).test(props.ipAddress)
+  ) {
+    throw new Error('IpAddress must be a valid IPv4 or IPv6 address');
+  }
```
GavinZZ, line 604: This regex is not human readable IMO. We should leave it out and let CFN to do the validation or please add documentations that link to the CFN page and describe the Regex.

GavinZZ (changes requested):
Hi sorry for keep you waiting on this PR. It's great work and it's a lot of work done and I'll try to help you to get this merged ASAP.

Left some comments to get your thoughts.

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
wladyslawczyzewski: hi @GavinZZ , yes, this is information from the official CFN's docs for this resource type https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-healthcheck-healthcheckconfig.html

packages/aws-cdk-lib/aws-route53/lib/health-check.ts
wladyslawczyzewski: hi @GavinZZ , please take a look on my another response to similar comment

wladyslawczyzewski:
hi @GavinZZ , could you please re-review the PR – I've deleted the props validation piece as it was too opinionated, I prefer to have the validation to be in synthesis phase instead deployment as this reduces the feedback time by a lot, but I also see the arguments against this.

GavinZZ:
> hi @GavinZZ , could you please re-review the PR – I've deleted the props validation piece as it was too opinionated, I prefer to have the validation to be in synthesis phase instead deployment as this reduces the feedback time by a lot, but I also see the arguments against this.

IMO I am usually in favour of synthesis validation rather than deployment validation. However, the validation was too complicated and I think we should only include the essential/most obvious ones in the synthesis validation. 

Either way, I think it's a two way door to release this feature without the validation first and we can always add more validation in the future.

GavinZZ (approved):

PR #30741, opened 2024-07-03
feat(logs): support regex patterns for JSON Metrics filters

### LINKED ISSUE #30451: aws-logs: Allow JSON filter patterns to use regex by removing automatic double quotes around pattern strings
### Describe the feature

[FilterPattern.stringValue](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-logs.FilterPattern.html) should allow the user to specify if the `value` input field is a regex string or not.

### Use Case

According to [CloudWatch filter pattern documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html#regex-expressions), regex expressions must be surrounded by `%`. 

I'm trying to use cdk's [FilterPattern.stringValue](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-logs.FilterPattern.html) to match a json document. 

CDK automatically adds double quotes in the final filter pattern, which prevents Cloudwatch from recognizing any string as a regex pattern. For example:
- Correct: { $.foo = %bar% }
- Incorrect: { $.foo = “%bar%” }, what is produced by CDK and does not match the document : {“foo”:“bar”}

### Proposed Solution

Reason for this behavior may be this line: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-logs/lib/pattern.ts#L399. Perhaps we need another input to `stringValue` that allows the user to specify if the input string is a regex pattern or not.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.144.0

Linux

jan-xyz:
### Issue # (if applicable)

Closes #30451 

### Reason for this change

Support Regex in filter functions for JSON

### Description of changes

Adding a new JSONPattern factory that uses `%` instead of `"` to support the regex pattern.

I tried it in our own CDK code. 

### Checklist

----

### DIFF
--- a/packages/@aws-cdk-testing/cli-integ/lib/with-cli-lib.ts
+++ b/packages/@aws-cdk-testing/cli-integ/lib/with-cli-lib.ts
@@ -37,9 +37,16 @@ export function withCliLibIntegrationCdkApp<A extends TestContext & AwsContext &
       }
 
       const alphaInstallationVersion = fixture.packages.requestedAlphaVersion();
+
+      // cli-lib-alpha has a magic alpha version in the old release pipeline,
+      // but will just mirror the CLI version in the new pipeline.
+      const cliLibVersion = process.env.CLI_LIB_VERSION_MIRRORS_CLI
+        ? `${fixture.packages.requestedCliVersion()}-alpha.0`
+        : alphaInstallationVersion;
+
       await installNpmPackages(fixture, {
         'aws-cdk-lib': installationVersion,
-        '@aws-cdk/cli-lib-alpha': alphaInstallationVersion,
+        '@aws-cdk/cli-lib-alpha': cliLibVersion,
         '@aws-cdk/aws-lambda-go-alpha': alphaInstallationVersion,
         '@aws-cdk/aws-lambda-python-alpha': alphaInstallationVersion,
         'constructs': '^10',
--- a/packages/@aws-cdk/aws-lambda-python-alpha/lib/function.ts
+++ b/packages/@aws-cdk/aws-lambda-python-alpha/lib/function.ts
@@ -62,7 +62,7 @@ export class PythonFunction extends Function {
       throw new Error(`Cannot find index file at ${resolvedIndex}`);
     }
 
-    const resolvedHandler =`${index.slice(0, -3)}.${handler}`.replace(/\//g, '.');
+    const resolvedHandler = `${index.slice(0, -3)}.${handler}`.replace(/\//g, '.');
 
     if (props.runtime && props.runtime.family !== RuntimeFamily.PYTHON) {
       throw new Error('Only `PYTHON` runtimes are supported.');
--- a/packages/@aws-cdk/cloudformation-diff/package.json
+++ b/packages/@aws-cdk/cloudformation-diff/package.json
@@ -23,8 +23,8 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@aws-cdk/aws-service-spec": "^0.1.52",
-    "@aws-cdk/service-spec-types": "^0.0.118",
+    "@aws-cdk/aws-service-spec": "^0.1.53",
+    "@aws-cdk/service-spec-types": "^0.0.119",
     "chalk": "^4",
     "diff": "^5.2.0",
     "fast-deep-equal": "^3.1.3",
--- a/packages/@aws-cdk/integ-runner/package.json
+++ b/packages/@aws-cdk/integ-runner/package.json
@@ -75,7 +75,7 @@
     "@aws-cdk/cloud-assembly-schema": "^39.2.0",
     "@aws-cdk/cloudformation-diff": "0.0.0",
     "@aws-cdk/cx-api": "0.0.0",
-    "@aws-cdk/aws-service-spec": "^0.1.52",
+    "@aws-cdk/aws-service-spec": "^0.1.53",
     "cdk-assets": "3.0.0-rc.143",
     "@aws-cdk/cdk-cli-wrapper": "0.0.0",
     "aws-cdk": "0.0.0",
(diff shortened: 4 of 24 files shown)


jan-xyz:
Clarification Request: How should the snapshot look like? It seems that this package doesn't have any yet? According to INTEGRATION_TEST.md there should be a `integ.mtericfilter.js.snapshot/` directory but I couldn't find it.

jan-xyz:
I had a question how to do the integration test, and I am still waiting on the response

paulhcsun:
Hi @jan-xyz, apologies for missing this clarification request. The integration tests for modules in `aws-cdk-lib` are located in the `aws-cdk/packages/@aws-cdk-testing/framework-integ/test/aws-logs/test` directory: https://github.com/aws/aws-cdk/tree/main/packages/%40aws-cdk-testing/framework-integ/test/aws-logs/test

packages/aws-cdk-lib/aws-logs/lib/pattern.ts
```diff
@@ -116,6 +116,27 @@ export class FilterPattern {
...
+   *
+   * For more information, see:
+   *
+   * https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html
+   *
+   * @param jsonField Field inside JSON. Example: "$.myField"
+   * @param comparison Comparison to carry out. Either = or !=.
+   * @param value The regex value to compare to.
+   */
+  public static regexValue(jsonField: string, comparison: '=' | '!=', value: string): JsonPattern {
```
paulhcsun, line 136: Due to JSII language translation limitations, we're not able to support a union for the `comparison` type. As a workaround, you could provide two methods for each case: `regexValueEquals()` and `regexValueNotEquals()`.

packages/aws-cdk-lib/aws-logs/lib/pattern.ts
```diff
@@ -225,6 +246,16 @@ class JSONStringPattern extends JsonPattern {
   }
 }
 
+/**
+ * A regex comparison for JSON patterns
+ */
+class JSONRegexPattern extends JsonPattern {
+  public constructor(jsonField: string, comparison: '=' | '!=', value: string) {
```
paulhcsun, line 253: Same as above, we're not able to support a union type due to JSII limitations. You could have two classes `JsonRegexPatternEquals` and `JsonRegexPatternNotEquals` for the `=` and `!=` case instead of having it passed in through the constructor.

packages/aws-cdk-lib/aws-logs/lib/pattern.ts
```diff
@@ -225,6 +246,16 @@ class JSONStringPattern extends JsonPattern {
   }
 }
 
+/**
+ * A regex comparison for JSON patterns
+ */
+class JSONRegexPattern extends JsonPattern {
```
paulhcsun, line 252:
```suggestion
class JsonRegexPattern extends JsonPattern {
```

paulhcsun (dismissed):
@jan-xyz Thanks for the contribution and for getting back to this PR! I've left a couple comments about some refactoring that will be needed since unions are not supported in public APIs/methods in the CDK. I will pass this review off to another team member to review once the changes are made.

packages/aws-cdk-lib/aws-logs/lib/pattern.ts
jan-xyz: Changing this breaks the pattern with the other classes, I can change it, but I am not sure if that is intended?

```
JSONStringPattern
JSONNumberPattern
JSONPostfixPattern
JSONAggregatePattern
```

packages/aws-cdk-lib/aws-logs/lib/pattern.ts
jan-xyz: I would keep the scope for now small, I think getting this in already unblocks a lot of use-cases.

packages/aws-cdk-lib/aws-logs/lib/pattern.ts
```diff
@@ -225,6 +246,16 @@ class JSONStringPattern extends JsonPattern {
...
+  public constructor(jsonField: string, comparison: string, value: string) {
+    // No validation, we assume these are generated by trusted factory functions
+    super(`${jsonField} ${comparison} %${value}%`);
```
aaythapa, line 255: Wondering if we should keep this is as `${value}` and throw an error if `value` isn't surrounded by `%`? The original issue complained that extra quotes were added and this will add extra `%`.

Personally haven't used log patterns too much so not sure if its more likely that people would pass in value with encapsulated within `%` or not

aaythapa (commented):
Thanks for contributing! Small comment

packages/aws-cdk-lib/aws-logs/lib/pattern.ts
aaythapa: We can throw an error if the comparison isn't `=` or `!=`. Better to throw error during synth time rather than the deployment failing

packages/aws-cdk-lib/aws-logs/lib/pattern.ts
jan-xyz: The problem with the quotes was that it didn't allow for %, which is the only other option, denoting regex. Since this adds the regex support with %, I don't think there is a reason to have the version without any surrounding markers. People who want literal matching use the " mark Version and people who want the regex pattern take this new % version.

packages/aws-cdk-lib/aws-logs/lib/pattern.ts
paulhcsun: Ah ok, in that case let's leave it as is.

packages/aws-cdk-lib/aws-logs/lib/pattern.ts
aaythapa: Fair enough

aaythapa (approved):
thank you for the contribution

paulhcsun:
@mergifyio update

aaythapa:
@mergify update

aaythapa (approved):

PR #30691, opened 2024-06-27
chore: use published version of cdk-assets

TheRealAmazonKendra:
This change reflects the move of cdk-assets from the monorep to a standalone library.

The build also pulled in some unrelated changes to the THIRD_PARTY_LICENSES files. This is to be expected.

### Checklist
- [ ] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30750, opened 2024-07-04
fix: preserve original docker config if credentialhelpers exists

### LINKED ISSUE #30743: cdk-assets: private ecr registry support overwrite any additional docker configuration
### Describe the bug

- We have a proxy configuration for docker in the ./docker/config.json file. 
- We are using private ecr to fetch our images which configuration is placed into the ~/.cdk/cdk-docker-creds.json file

The cdk-assets creates a new docker config file but it is only placing the contents of the ~/.cdk/cdk-docker-creds.json file and only the `credHelpers` property, so it is not even possible to include the proxy settings.

### Expected Behavior

Both configuration should be respected.

### Current Behavior

Only credentials are kept.

### Reproduction Steps

see before.

### Possible Solution

Change this line: https://github.com/aws/aws-cdk/blob/1fcef6309e54d7ec1e179c22a47a235b38fa1673/packages/cdk-assets/lib/private/docker.ts#L168C81-L168C92

Options:
1. Write out the whole ~/.cdk/cdk-docker-creds.json  file not just the credHelpers
2. Add support to include additional ~/.cdk/cdk-docker*.json files to plug in additional docker configurations
3. Check whenever a ~./docker/config.json file exists and merge both files.

_No response_

### CDK CLI Version

2.147.0

### Framework Version

_No response_

### Node.js Version

Any

### OS

Any

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

gyalai-aws:
### Issue # (if applicable)

Closes #30743.

### Reason for this change
cdk-assets docker credential configuration overwrite the original docker configurations , etc proxy

### Description of changes

Check whenever a ~./docker/config.json file exists and merge both files.

Unit tests are added

### Checklist

----


gyalai-aws:
Exemption Request

packages/cdk-assets/lib/private/docker.ts
```diff
@@ -158,18 +158,49 @@ export class Docker {
     const config = cdkCredentialsConfig();
     if (!config) { return false; }
 
-    this.configDir = fs.mkdtempSync(path.join(os.tmpdir(), 'cdkDockerConfig'));
+    // Condition configured for supporting testing, it is not expected the CDK_DOCKER_TMP_CONFIG_DIR to be set in production
+    if (process.env.CDK_DOCKER_TMP_CONFIG_DIR) {
```
otaviomacedo, line 162: I advise against having separate code paths for test and production. It's very easy for someone in the future to make a mistake and change code in one path and not the other. Is there any reason you're not mocking the `fs.mkdtempSync` method?

otaviomacedo (changes requested):

packages/cdk-assets/lib/private/docker.ts
gyalai-aws: let me try that

packages/cdk-assets/test/private/docker.test.ts
```diff
@@ -91,4 +97,164 @@ describe('Docker', () => {
...
+    });
+
+    afterEach(() => {
+      mockfs.restore();
+      process.env = { ..._ENV };
+    });
+
+    test('returns false if no cred config exists', () => {
+      expect(docker.configureCdkCredentials()).toBeFalsy();
+    });
```
scanlonp, line 195: Was this expected behavior before, or is this new?

If not new, would this have been covered by one of the above tests - i.e. would this have resulted in a shell error?

packages/cdk-assets/lib/private/docker.ts
```diff
@@ -165,11 +165,36 @@ export class Docker {
...
+      if (fs.existsSync(this.dockerConfigFile())) {
+        return JSON.parse(fs.readFileSync(this.dockerConfigFile(), { encoding: 'utf-8' }));
+      } else {
+        return {};
+      }
+    } catch (e: any) {
+      if (e instanceof SyntaxError) {
+        throw new Error(`Unable to parse \'${this.dockerConfigFile()}\' in order to determine the configuration. Please ensure \'${this.dockerConfigFile()}\' is a valid JSON.`);
+      }
+      throw e;
```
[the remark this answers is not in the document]
scanlonp: Maybe this is not a big concern since it is only when the config file exists and is invalid. That would indicate the user wants to use the config file, but it is not written correctly. This case also seems rare.

Still something to think about for backwards compatability.

scanlonp (commented):
As a heads up, we are currently doing work on re-designing the package structure of cdk-assets: #30691. This PR may need to be remade / reflected in the standalone library.
____

Thanks for this contribution! I think appending a local docker config to the credentials is an improvement.

My only concern is introducing errors where users previously did not encounter them. 
Can you give some context on docker config files?

PR #30756, opened 2024-07-05
feat(pipes-targets): add API destination

msambol:
Add EventBridge API destination as a Pipes target.

CloudFormation groups EventBridge API destination with API Gateway REST API
as [PipeTargetHttpParameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargethttpparameters.html#cfn-pipes-pipe-pipetargethttpparameters-pathparametervalues), but I think separating them here similar to [aws-event-targets](https://github.com/aws/aws-cdk/tree/main/packages/aws-cdk-lib/aws-events-targets/lib) 
makes more sense, as API Gateway requires `stage`, `path`, and `method` (see [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-events-targets/lib/api-gateway.ts#L11-L32)).


packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-destination.ts
```diff
@@ -0,0 +1,64 @@
...
+  testCases: [stack],
+});
+
+test.assertions.awsApiCall('SQS', 'sendMessage', {
+  QueueUrl: sourceQueue.queueUrl,
+  MessageBody: 'USA',
+});
+
+// This is difficult to validate as we're invoking an external API.
+// A manual check was done to ensure the pipe executed successfully.
```
nmussy, line 62: You could spin up an HTTP API Gateway with a Lambda to handle the requests, and store the request object in something like a Bucket

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-destination.ts
```diff
@@ -0,0 +1,64 @@
...
+});
+
+const destination = new cdk.aws_events.ApiDestination(stack, 'MyDestination', {
+  connection,
+  endpoint: 'https://httpbin.org/headers',
+  httpMethod: cdk.aws_events.HttpMethod.GET,
+  apiDestinationName: 'MyDestination',
+  rateLimitPerSecond: 1,
+  description: 'Calling example.com with API key x-api-key',
+});
```
nmussy, line 44: Could you add all properties? I'm particularly curious to see if `queryStringParameters: {"key": ""}` is converted to `&key=` and `headerParameters: {"X-Header": ""}` to `X-Header: ""`

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-destination.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.queue.grantConsumeMessages(pipeRole);
+  }
+}
+
+const secret = new cdk.aws_secretsmanager.Secret(stack, 'MySecret', {
+  secretStringValue: cdk.SecretValue.unsafePlainText('abc123'),
+});
+
+const connection = new cdk.aws_events.Connection(stack, 'MyConnection', {
+  authorization: cdk.aws_events.Authorization.apiKey('x-api-key', secret.secretValue),
```
nmussy, line 32: I'm also curious to see what happens when you also set the same header in `ApiDestination.headerParameters`. It's either ignored or overrides the `authorization`, but I think we should probably warn the user either way

nmussy (changes requested):
Mostly integ coverage, LGTM overall

msambol:
@nmussy Here's the result from the Lambda logs. The `x-api-key` defined in `Connection` has precedence over the one defined in `headerParameters`.

Screenshot 2024-07-06 at 10 35 18 AM

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-destination.ts
msambol: See comment below with screenshot.

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-destination.ts
```diff
@@ -0,0 +1,80 @@
...
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetparameters.html#cfn-pipes-pipe-pipetargetparameters-inputtemplate
+   * @default - none
+   */
+  readonly inputTransformation?: IInputTransformation;
+
+  /**
+   * The headers to send as part of the request invoking the EventBridge API destination.
+   *
+   * The headers are merged with the headers from the API destination.
+   * If there are conflicts, the headers from the API desintation take precedence.
```
msambol, line 21: @nmussy I added this note about precedence. From the logs we can see that `x-api-key` is `abc123`, not `apiKeyFromHeaderParams`.

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-destination.ts
```diff
@@ -0,0 +1,80 @@
...
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargetparameters.html#cfn-pipes-pipe-pipetargetparameters-inputtemplate
+   * @default - none
+   */
+  readonly inputTransformation?: IInputTransformation;
+
+  /**
+   * The headers to send as part of the request invoking the EventBridge API destination.
+   *
+   * The headers are merged with the headers from the API destination.
+   * If there are conflicts, the headers from the API destination take precedence.
```
msambol, line 21: @nmussy I added this note about precedence. From the logs we can see that x-api-key is abc123, not apiKeyFromHeaderParams.

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
```diff
@@ -171,3 +171,20 @@ const pipe = new pipes.Pipe(this, 'Pipe', {
     target: pipeTarget
 });
 ```
+
+### Amazon EventBridge API Destination
```
Leo10Gama, line 175: Similar to the previous target types (SQS, Step Functions, Lambda), can we include an example for `inputTransformation`?

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-destination.ts
Leo10Gama: I'm not sure whether this is enough to indicate to users which value will take precedence. Is there somewhere we can print a warning when there are two identical headers provided, which explicitly lets the user know that the value is being set twice, and which of those values it's using?

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-destination.ts
```diff
@@ -0,0 +1,104 @@
...
+  limit: 1,
+});
+
+const message = putMessageOnQueue.next(logEvents);
+
+// Check that the Lambda was invoked successfully from API GW
+message.assertAtPath('events.0.message', ExpectedResult.stringLikeRegexp('INIT_START')).waitForAssertions({
+  totalTimeout: cdk.Duration.minutes(1),
+  interval: cdk.Duration.seconds(15),
+});
```
Leo10Gama, line 102: Correct me if I'm wrong, but I'm not sure whether this check is explicit enough to verify that the Lambda was invoked. Since the code is throwing a 200 return code, could we check for that instead? Or can we verify that the SQS message was sent by checking for the message body?

Leo10Gama (changes requested):
Thanks for your contribution, and sorry for the delay! Overall seems good to me, just a few things to note before moving forward with these changes.

msambol:
>I'm not sure whether this is enough to indicate to users which value will take precedence. Is there somewhere we can print a warning when there are two identical headers provided, which explicitly lets the user know that the value is being set twice, and which of those values it's using?

I'm not quite sure how to do this. There could be any number of custom headers, no?

Leo10Gama:
> > I'm not sure whether this is enough to indicate to users which value will take precedence. Is there somewhere we can print a warning when there are two identical headers provided, which explicitly lets the user know that the value is being set twice, and which of those values it's using?
> 
> I'm not quite sure how to do this. There could be any number of custom headers, no?

I believe that's the case, yeah. I was thinking some check like iterating through the header keys to see which overlap, like turning them into a set and taking the intersection type of operation. Then if we know the `Connection` key has precedence, like your logs show, we print a warning along the lines of:
```
"Key <so-and-so> is defined in both Connection [connection] and ApiDestinationTarget [api destination target]. Value <my value> from Connection will be used."
```

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-destination.ts
msambol: I updated the integration test to check for the payload from SQS, which is sent as follows:

`queryStringParameters: { key: '$.body' },`

msambol:
> > > I'm not sure whether this is enough to indicate to users which value will take precedence. Is there somewhere we can print a warning when there are two identical headers provided, which explicitly lets the user know that the value is being set twice, and which of those values it's using?
> > 
> > 
> > I'm not quite sure how to do this. There could be any number of custom headers, no?
> 
> I believe that's the case, yeah. I was thinking some check like iterating through the header keys to see which overlap, like turning them into a set and taking the intersection type of operation. Then if we know the `Connection` key has precedence, like your logs show, we print a warning along the lines of:
> 
> ```
> "Key <so-and-so> is defined in both Connection [connection] and ApiDestinationTarget [api destination target]. Value <my value> from Connection will be used."
> ```

I'd prefer to do this in a follow-up PR. What do you think? I think the logic could get messy/complicated. I think the documentation note is sufficient for now.

Leo10Gama:
> I'd prefer to do this in a follow-up PR. What do you think? I think the logic could get messy/complicated. I think the documentation note is sufficient for now.

I do think we should include these changes in this PR, unless there are any other extra changes you wanted to add that might make the logic more confusing to follow. But ideally, including them in this PR means the logic is bundled together from the get-go.

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
```diff
@@ -181,7 +182,9 @@ The API destination will receive the (enriched/filtered) source payload.
 declare const sourceQueue: sqs.Queue;
 declare const dest: events.ApiDestination;
 
-const apiTarget = new targets.ApiDestinationTarget(dest);
+const apiTarget = new targets.ApiDestinationTarget(dest, {
+  inputTransformation: pipes.InputTransformation.fromObject({ body: "👀" }),
+});
```
Leo10Gama, line 187: Ah, when I said to include an `inputTransformation` example, I meant on top of the existing default example 😅 So for instance, for Lambda Function (L127-L173), there's the default example (L131-L141) and the input transformation example (L161-L173). I was hoping we could replicate that structure for the API destination as well.

Leo10Gama (changes requested):
Thanks for the changes! Aside from my comment above regarding including the warning, just one more minor nit in the README.

msambol:
> > I'd prefer to do this in a follow-up PR. What do you think? I think the logic could get messy/complicated. I think the documentation note is sufficient for now.
> 
> I do think we should include these changes in this PR, unless there are any other extra changes you wanted to add that might make the logic more confusing to follow. But ideally, including them in this PR means the logic is bundled together from the get-go.

I'm a little confused on what this would look like. An `ApiDestination` doesn't have headers associated with it. In my integration test example, the `Connection` has an API key (`x-api-key`). Happy to add the warning but I need some direction as there is nothing to iterate through on `ApiDestination` ?

Leo10Gama:
> > > I'd prefer to do this in a follow-up PR. What do you think? I think the logic could get messy/complicated. I think the documentation note is sufficient for now.
> > 
> > 
> > I do think we should include these changes in this PR, unless there are any other extra changes you wanted to add that might make the logic more confusing to follow. But ideally, including them in this PR means the logic is bundled together from the get-go.
> 
> I'm a little confused on what this would look like. An `ApiDestination` doesn't have headers associated with it. In my integration test example, the `Connection` has an API key (`x-api-key`). Happy to add the warning but I need some direction as there is nothing to iterate through on `ApiDestination` ?

If I'm understanding the code correctly, the area of conflict is between `ApiDestination.headerParameters` and the API key, so we'd iterate on the header parameters. We should be able to compare the headers since the target has access to the `Connection` via `ApiDestinationTarget.destination.connection`.

msambol:
> > > > I'd prefer to do this in a follow-up PR. What do you think? I think the logic could get messy/complicated. I think the documentation note is sufficient for now.
> > > 
> > > 
> > > I do think we should include these changes in this PR, unless there are any other extra changes you wanted to add that might make the logic more confusing to follow. But ideally, including them in this PR means the logic is bundled together from the get-go.
> > 
> > 
> > I'm a little confused on what this would look like. An `ApiDestination` doesn't have headers associated with it. In my integration test example, the `Connection` has an API key (`x-api-key`). Happy to add the warning but I need some direction as there is nothing to iterate through on `ApiDestination` ?
> 
> If I'm understanding the code correctly, the area of conflict is between `ApiDestination.headerParameters` and the API key, so we'd iterate on the header parameters. We should be able to compare the headers since the target has access to the `Connection` via `ApiDestinationTarget.destination.connection`.

I think I would need to change 

```
private destination: IApiDestination;
```
to 
```
private destination: ApiDestination;
```

Excuse my naïveté. Good with that?

msambol:
> > > > > I'd prefer to do this in a follow-up PR. What do you think? I think the logic could get messy/complicated. I think the documentation note is sufficient for now.
> > > > 
> > > > 
> > > > I do think we should include these changes in this PR, unless there are any other extra changes you wanted to add that might make the logic more confusing to follow. But ideally, including them in this PR means the logic is bundled together from the get-go.
> > > 
> > > 
> > > I'm a little confused on what this would look like. An `ApiDestination` doesn't have headers associated with it. In my integration test example, the `Connection` has an API key (`x-api-key`). Happy to add the warning but I need some direction as there is nothing to iterate through on `ApiDestination` ?
> > 
> > 
> > If I'm understanding the code correctly, the area of conflict is between `ApiDestination.headerParameters` and the API key, so we'd iterate on the header parameters. We should be able to compare the headers since the target has access to the `Connection` via `ApiDestinationTarget.destination.connection`.
> 
> I think I would need to change
> 
> ```
> private destination: IApiDestination;
> ```
> 
> to
> 
> ```
> private destination: ApiDestination;
> ```
> 
> Excuse my naïveté. Good with that?

Also, we'd have to import the API right?

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
```diff
@@ -171,3 +172,36 @@ const pipe = new pipes.Pipe(this, 'Pipe', {
...
+The API destination will receive the (enriched/filtered) source payload.
+
+```ts
+declare const sourceQueue: sqs.Queue;
+declare const dest: events.ApiDestination;
+
+const apiTarget = new targets.ApiDestinationTarget(dest);
+
+const pipe = new pipes.Pipe(this, 'Pipe', {
+    source: new SomeSource(sourceQueue),
```
moelasmar, line 188: could you please use the correct Class name, this should be `SqsSource` instead of `SomeSource`

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-destination.ts
```diff
@@ -0,0 +1,80 @@
...
+   *
+   * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargethttpparameters.html#cfn-pipes-pipe-pipetargethttpparameters-headerparameters
+   * @default - none
+   */
+  readonly headerParameters?: Record<string, string>;
```
moelasmar, line 26: could you please change the Record type to be something like `{ [key: string]: string }` .. we used this pattern before for representing Json Object of  Key, Value

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-destination.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+  bind(_pipe: IPipe): SourceConfig {
+    return {
+      sourceParameters: this.sourceParameters,
+    };
+  }
+  grantRead(pipeRole: cdk.aws_iam.IRole): void {
+    this.queue.grantConsumeMessages(pipeRole);
+  }
+}
```
moelasmar, line 36: why do you create this Source class, and not to use the `SqsSource` available in the pipes-sources module

Leo10Gama:
> > 
> > 
> > I think I would need to change
> > ```
> > private destination: IApiDestination;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to
> > ```
> > private destination: ApiDestination;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Excuse my naïveté. Good with that?
> 
> Also, we'd have to import the API right?

The best practice would be to use the interface. If there's no other way to get ahold of the value, I think we can omit the warning in that case and just leave the docstring. It seems like the service team's implementation is what controls the priority, so I can reach out to them separately to look into including the warning from our side. In the meantime, leaving the changes with `IApiDestination` seems fine, just the other comments mentioned above left to be addressed.

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
msambol: The rest of the README has `SomeSource` but happy to change it.

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-destination.ts
msambol: I was reusing the pattern in other integration tests. I think it's so we don't need to import `SqsSource` and instead use the interface?

[1] https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-pipes-targets-alpha/test/integ.stepfunctions.ts#L39
[2] https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-pipes-targets-alpha/test/integ.sqs.ts#L12
[3] https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-pipes-targets-alpha/test/integ.lambda.ts#L46

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-destination.ts
msambol: @redwheeler3 jfyi. I think it's worth mentioning in Docs and/or Console about precedence regarding conflict headers.

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-destination.ts
moelasmar: IMO, in integration test cases we should use CDK Apis the same way the customer will use them, so I prefer we also change the other integration test cases, but this is out of this PR scope. But at least for this PR, let's use the SqsSource

msambol: Fixed in https://github.com/aws/aws-cdk/pull/30756/commits/a4f52b2454ae2be4666288b20d4b53ced0bedfe3

packages/@aws-cdk/aws-pipes-targets-alpha/README.md
msambol: Fixed in https://github.com/aws/aws-cdk/pull/30756/commits/a4f52b2454ae2be4666288b20d4b53ced0bedfe3.

msambol:
@moelasmar updated with your feedback—thanks!

Leo10Gama (commented):
Seems all good to me, thanks for the quick responses on feedback! Just adding in one quick typehint that I missed before, and will wait to make sure the tests pass.

Leo10Gama (approved):

msambol:
@mergify update

PR #30772, opened 2024-07-06
feat(pipes-targets): add API Gateway

msambol:
Add API Gateway REST API as a Pipes target.

### DIFF
--- a/packages/@aws-cdk/aws-pipes-targets-alpha/README.md
+++ b/packages/@aws-cdk/aws-pipes-targets-alpha/README.md
@@ -218,7 +218,7 @@ const fn = new lambda.Function( this, 'MyFunc', {
   handler: 'index.handler',
   runtime: lambda.Runtime.NODEJS_LATEST,
   code: lambda.Code.fromInline( 'exports.handler = e => {}' ),
-} );
+});
 
 const restApi = new api.LambdaRestApi( this, 'MyRestAPI', { handler: fn } );
 const apiTarget = new targets.ApiGatewayTarget(restApi);
@@ -238,12 +238,12 @@ const fn = new lambda.Function( this, 'MyFunc', {
   handler: 'index.handler',
   runtime: lambda.Runtime.NODEJS_LATEST,
   code: lambda.Code.fromInline( 'exports.handler = e => {}' ),
-} );
+});
 
 const restApi = new api.LambdaRestApi( this, 'MyRestAPI', { handler: fn } );
-const apiTarget = new targets.ApiGatewayTarget(restApi), {
+const apiTarget = new targets.ApiGatewayTarget(restApi, {
   inputTransformation: pipes.InputTransformation.fromObject({ body: "👀" }),
-};
+});
 
 const pipe = new pipes.Pipe(this, 'Pipe', {
     source: new SqsSource(sourceQueue),


packages/@aws-cdk/aws-pipes-targets-alpha/README.md
```diff
@@ -171,3 +171,26 @@ const pipe = new pipes.Pipe(this, 'Pipe', {
     target: pipeTarget
 });
 ```
+
+### Amazon API Gateway Rest API
+
+A REST API can be used as a target for a pipe. 
+The REST API will receive the (enriched/filtered) source payload.
```
nmussy, line 178: Consistent casing:

```suggestion
### Amazon API Gateway REST API

A REST API can be used as a target for a pipe. 
The REST API will receive the (enriched/filtered) source payload.
```

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-gateway.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    this.targetArn = this.restApiArn;
+  }
+
+  grantPush(grantee: IRole): void {
+    grantee.addToPrincipalPolicy(new PolicyStatement({
+      resources: [this.restApiArn],
+      actions: [
+        'execute-api:Invoke',
+        'execute-api:ManageConnections',
```
nmussy, line 90: Isn't `ManageConnections` exclusively used for WebSocket APIs? See [permissions.cloud](https://aws.permissions.cloud/iam/execute-api)

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-gateway.ts
```diff
@@ -0,0 +1,109 @@
...
+  public readonly targetArn;
+
+  constructor(restApi: IRestApi, parameters?: ApiGatewayTargetParameters) {
+    this.restApi = restApi;
+    this.restApiParameters = parameters;
+
+    this.restApiArn = this.restApi.arnForExecuteApi(
+      this.restApiParameters?.method,
+      this.restApiParameters?.path || '/',
+      this.restApiParameters?.stage || this.restApi.deploymentStage.stageName,
```
nmussy, line 79: I think you might have an issue with `this.restApi.deploymentStage?.stageName` being undefined:

https://github.com/aws/aws-cdk/blob/ddbf57bd3bf4f8937a63c94d7188792ed37814b3/packages/aws-cdk-lib/aws-apigateway/lib/restapi.ts#L355-L361

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-gateway.ts
```diff
@@ -0,0 +1,73 @@
...
+restApi.root.addMethod('GET', new apigw.LambdaIntegration(fn));
+
+new Pipe(stack, 'Pipe', {
+  source: new TestSource(sourceQueue),
+  target: new ApiGatewayTarget(restApi, {
+    method: 'GET',
+    headerParameters: {
+      'x-header': 'myheader',
+    },
+    queryStringParameters: { key: 'USA' },
```
nmussy, line 47: Could you add a `pathParameterValues` here as well?

nmussy (changes requested):
The [docs](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-api-gateway-target.html) don't explicitly mention that the target can only be a REST API, it might work with WebSockets? If that's the case, could you add an integ and the likes?

msambol:
> The [docs](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-api-gateway-target.html) don't explicitly mention that the target can only be a REST API, it might work with WebSockets? If that's the case, could you add an integ and the likes?

Good shout about the docs. I did create a WebSocket API and it wasn't available in the Console when I tried to create a Pipe target. @redwheeler3, can you confirm if REST APIs are the only type supported here?

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-gateway.ts
msambol: I created a pipe manually in the Console and these were the perms on the IAM role created by the Console. It does appear this permission is specific to WebSocket APIs. Let's get clarity on whether WebSocket APIs are permitted as a Pipe target.

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-gateway.ts
msambol: wait doesn't `!` mean it can't be `null` or `undefined` ?

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-gateway.ts
msambol: I'm actually not sure how these are used? Do I need this since I specified `path` ?

msambol: It works like this:

```
new Pipe(stack, 'Pipe', {
  source: new TestSource(sourceQueue),
  target: new ApiGatewayTarget(restApi, {
    method: 'GET',
    path: '/*',
    headerParameters: {
      'x-header': 'myheader',
    },
    queryStringParameters: { key: 'USA' },
    pathParameterValues: ['books'],
  }),
});
```

But I'd prefer to leave it as I have now, I think it's more straightforward.

packages/@aws-cdk/aws-pipes-targets-alpha/test/api-gateway.test.ts
```diff
@@ -0,0 +1,212 @@
...
+            { Ref: 'MyLambdaRestApiDeploymentStageprodA127C527' },
+            '/*/',
+          ],
+        ],
+      },
+      TargetParameters: {},
+    });
+  });
+
+  it('should have only target arn with stage/method/path', () => {
```
msambol, line 57: I added this additional test.

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-gateway.ts
nmussy: The non-null assertion operator is just a compile time override. Say you have this class:

```ts
class Example {
  value: string;
}
```

By default, TS will (rightfully) complain that `value` cannot be not a string: `Property 'value' has no initializer and is not definitely assigned in the constructor.(2564)`

You can override this check with the `!` operator, and `new Example().value` will be typed as a string, but will remain `undefined` at execution time

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-gateway.ts
nmussy: I think that either way, we should only grant `execute-api:Invoke` here. If WebSockets are usable, we can conditionally grant `execute-api:ManageConnections` instead

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-gateway.ts
nmussy: Yeah your example couldn't have used a `pathParameterValues`, I just like to see as much coverage as possible, even on integ tests

msambol: I could add '/books/*' and add 'fiction' for pathParameterValues

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-gateway.ts
msambol: I think the CDK automatically creates a stage though? So we should be good?

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-gateway.ts
```diff
@@ -0,0 +1,106 @@
...
+    );
+
+    this.targetArn = this.restApiArn;
+  }
+
+  grantPush(grantee: IRole): void {
+    grantee.addToPrincipalPolicy(new PolicyStatement({
+      resources: [this.restApiArn],
+      actions: ['execute-api:Invoke'],
```
msambol, line 88: @nmussy updated this to just `Invoke`

packages/@aws-cdk/aws-pipes-targets-alpha/test/integ.api-gateway.ts
```diff
@@ -0,0 +1,75 @@
...
+new Pipe(stack, 'Pipe', {
+  source: new TestSource(sourceQueue),
+  target: new ApiGatewayTarget(restApi, {
+    method: 'GET',
+    path: '/books/*',
+    headerParameters: {
+      'x-header': 'myheader',
+    },
+    queryStringParameters: { key: 'USA' },
+    pathParameterValues: ['fiction'],
```
msambol, line 49: @nmussy I think this is what you're after? `fiction` replaces the `*` in `path` defined above. More coverage..

redwheeler3:
> Good shout about the docs. I did create a WebSocket API and it wasn't available in the Console when I tried to create a Pipe target. The [CFN docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargethttpparameters.html) do state only REST APIs. @redwheeler3, can you confirm if REST APIs are the only type supported here?

Pipes does not support WebSocket APIs.

msambol:
> > Good shout about the docs. I did create a WebSocket API and it wasn't available in the Console when I tried to create a Pipe target. The [CFN docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-pipes-pipe-pipetargethttpparameters.html) do state only REST APIs. @redwheeler3, can you confirm if REST APIs are the only type supported here?
> 
> Pipes does not support WebSocket APIs.

The `ManageConnections` permission can be removed from the role created by the Console.

packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-gateway.ts
nmussy: I can see a couple of issues off the top of my head:

Given you're allowing `IRestApi` to construct `ApiGatewayTarget`, imported constructs can be passed. The `deploymentStage` will not be imported, even if the API has one:

```ts
const api = RestApi.fromRestApiId(this, 'RestApi', 'xxx');
assert(api.deploymentStage, 'Deployment stage is not defined');
```

You can also create a `RestApi` without a deployment with the `deploy` prop:

```ts
const api = new RestApi(this, 'api', {
  // Toggling this line will cause the assertion to fail  
  // deploy: false,
});
api.root.addMethod('GET');

assert(api.deploymentStage, 'Deployment stage is not defined');
```

PR #30773, opened 2024-07-07
chore(lambda): inline code size limit is 4MB

### LINKED ISSUE #30761: aws_cdk.aws_lambda.Code: from_inline method code size limit incorrect
### Describe the issue

The CDK documentation for `from_inline` states the `code` parameter has a size limit of **4KiB**. However, this limit is actually **4MB** according to the CloudFormation documentation for ZipFile size, which is what `from_inline` will synthesize to.

I tested this by providing a string that was larger than 4KB, and CDK successfully synthesized a template without error.

### Links

https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda/Code.html#aws_cdk.aws_lambda.Code.from_inline
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html

msambol:
### Issue # (if applicable)

Closes #30761.

### Checklist

----


scanlonp (commented):
This seems like a massive in-line code size, but not out of the question. 

Looks like there used to be a character restriction on inline code based on some old stackoverflow answers, but that may have been removed. 

I would like to see if an inline lambda will deploy with more than 4kb. The template being synthesized is not a good indicator of whether it works or not since it would be a CFN limit that we do not validate currently.

msambol:
> This seems like a massive in-line code size, but not out of the question.
> 
> Looks like there used to be a character restriction on inline code based on some old stackoverflow answers, but that may have been removed.
> 
> I would like to see if an inline lambda will deploy with more than 4kb. The template being synthesized is not a good indicator of whether it works or not since it would be a CFN limit that we do not validate currently.

I changed it to be `the resulting zip file cannot exceed 4MB`. I think that's the only limit we can deduce from the docs?

scanlonp:
Hey @msambol, I am doing some work with lambda assets, so I will give this a spin. I still think this may be an error in their documentation. The limit for a CFN template as a whole uploaded to s3 is 1 MB ([CFN limits](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html)).

I'd love to be wrong here, but I am feeling lucky with this prediction! I'll get back to you shortly.

msambol:
@scanlonp Any update on this one?

msambol:
@scanlonp Thoughts here? Should I close this?

GavinZZ (approved):

GavinZZ:
Hello, sorry for the wait. This change LGTM and approved it. Appreciate the contribution.

msambol:
@mergify update

PR #30490, opened 2024-06-07
fix(cloudtrail): do not attach s3 bucket permission when orgId is not set for organization trail

sarisia:
### Issue # (if applicable)

no open issue

### Reason for this change

Organization trail without `orgId` attaches improper s3 bucket policy
which allows cloudtrail to write to `/AWSLogs/undefined/*` prefix.

```json
                        {
                            "Action": "s3:PutObject",
                            "Condition": {
                                "StringEquals": {
                                    "s3:x-amz-acl": "bucket-owner-full-control",
                                    "aws:SourceArn": {
                                        "Fn::Join": [
                                            "",
                                            [
                                                "arn:",
                                                {
                                                    "Ref": "AWS::Partition"
                                                },
                                                ":cloudtrail:us-east-1:123456789012:trail/undefined"
                                            ]
                                        ]
                                    }
                                }
                            },
                            "Effect": "Allow",
                            "Principal": {
                                "Service": "cloudtrail.amazonaws.com"
                            },
                            "Resource": {
                                "Fn::Join": [
                                    "",
                                    [
                                        {
                                            "Fn::GetAtt": [
                                                "TrailS30071F172",
                                                "Arn"
                                            ]
                                        },
                                        "/AWSLogs/undefined/*"
                                    ]
                                ]
                            }
                        }
```

(there is also another bug which trail name becomes `undefined`, which is addressed in https://github.com/aws/aws-cdk/pull/30489 based on this PR)

### Description of changes

- do not attach s3 bucket policy if `isOrganizationTrail` is set and `orgId` is not set
  - generate warning if attaching s3 bucket policy is skipped

### Description of how you validated changes

- added unit test

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30489, opened 2024-06-07
fix(cloudtrail): emit error if trailName is not set for organization trail

sarisia:
This PR is based on #30490 and I'll rebase this once it is merged

### Issue # (if applicable)

no open issue

### Reason for this change

Organization trail without `trailName` attaches improper s3 bucket policy
which allows trail name `undefined` to send logs.

```json
                        {
                            "Action": "s3:PutObject",
                            "Condition": {
                                "StringEquals": {
                                    "s3:x-amz-acl": "bucket-owner-full-control",
                                    "aws:SourceArn": {
                                        "Fn::Join": [
                                            "",
                                            [
                                                "arn:",
                                                {
                                                    "Ref": "AWS::Partition"
                                                },
                                                ":cloudtrail:us-east-1:123456789012:trail/undefined"
                                            ]
                                        ]
                                    }
                                }
                            },
                            "Effect": "Allow",
                            "Principal": {
                                "Service": "cloudtrail.amazonaws.com"
                            },
                            "Resource": {
                                "Fn::Join": [
                                    "",
                                    [
                                        {
                                            "Fn::GetAtt": [
                                                "TrailS30071F172",
                                                "Arn"
                                            ]
                                        },
                                        "/AWSLogs/o-xxxxxxxxx/*"
                                    ]
                                ]
                            }
                        }

```

### Description of changes

- make `trailName` required if `isOrganizationTrail` is set

### Description of how you validated changes

- Added unit test

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30778, opened 2024-07-07
fix(cloudtrail): do not attach s3 bucket permission when orgId is not set for organization trail

sarisia:
Recreate #30490 

### Issue # (if applicable)

no open issue

### Reason for this change

Organization trail without `orgId` attaches improper s3 bucket policy
which allows cloudtrail to write to `/AWSLogs/undefined/*` prefix.

```json
                        {
                            "Action": "s3:PutObject",
                            "Condition": {
                                "StringEquals": {
                                    "s3:x-amz-acl": "bucket-owner-full-control",
                                    "aws:SourceArn": {
                                        "Fn::Join": [
                                            "",
                                            [
                                                "arn:",
                                                {
                                                    "Ref": "AWS::Partition"
                                                },
                                                ":cloudtrail:us-east-1:123456789012:trail/undefined"
                                            ]
                                        ]
                                    }
                                }
                            },
                            "Effect": "Allow",
                            "Principal": {
                                "Service": "cloudtrail.amazonaws.com"
                            },
                            "Resource": {
                                "Fn::Join": [
                                    "",
                                    [
                                        {
                                            "Fn::GetAtt": [
                                                "TrailS30071F172",
                                                "Arn"
                                            ]
                                        },
                                        "/AWSLogs/undefined/*"
                                    ]
                                ]
                            }
                        }
```

### Description of changes

- do not attach s3 bucket policy if `isOrganizationTrail` is set and `orgId` is not set
  - generate warning if attaching s3 bucket policy is skipped

- added unit test

### Checklist

----


sarisia:
Exemption Request: deploying this requires AWS organization management account

sarisia:
could someone review this please? I'm waiting this for 4 months, thanks!

GavinZZ:
@sarisia Hello, apologize for the extraa long wait time on the PR review. This seems to be an accident when we introduced `orgId` field and somehow missed the validation. I think this PR makes sense to me, but just curious what's your use case? Are you using organization trail without `orgId`?

sarisia:
@GavinZZ Thank you for reviewing!

> I think this PR makes sense to me, but just curious what's your use case? Are you using organization trail without `orgId`?

Actually yes. Back in 2023 there were no `orgId` prop in cloudtrail.Trail so I have to write the code which manually attaches insufficient permissions, like this:

```python
        # trail
        trail = cloudtrail.Trail(
            self,
            "Trail",
            bucket=bucket,
            is_organization_trail=True
        )

        # add bucket policy
        # https://docs.aws.amazon.com/ja_jp/awscloudtrail/latest/userguide/create-s3-bucket-policy-for-cloudtrail.html
        cloudtrail_principal = iam.ServicePrincipal(
            "cloudtrail.amazonaws.com"
        )
        bucket.add_to_resource_policy(
            iam.PolicyStatement(
                actions=["s3:GetBucketAcl"],
                principals=[cloudtrail_principal],
                resources=[bucket.bucket_arn],
            )
        )
        bucket.add_to_resource_policy(
            iam.PolicyStatement(
                actions=["s3:PutObject"],
                principals=[cloudtrail_principal],
                resources=[
                    bucket.arn_for_objects(f"AWSLogs/{Stack.of(self).account}/*"),
                    bucket.arn_for_objects(
                        f"AWSLogs/{organization_principal.organization_id}/*"
                    ),
                ],
                conditions={
                    "StringEquals": {"s3:x-amz-acl": "bucket-owner-full-control"}
                },
            )
        )
```

Recently I upgraded CDK and found there's a diff with new weird permission attached to the bucket, so I'm trying to get rid of them without breaking deploys and forcing code changes.

sarisia:
Actually there's another issue which `props.trailName` in generated policy becomes `undefined` too.

Initially I tried to fix them in separated PR (https://github.com/aws/aws-cdk/pull/30489) which makes `props.trailName` required but now I think it's better to fix this PR to check both `props.orgId` and `props.trailName` without making them required. WDYT?

GavinZZ:
@sarisia sorry for the delayed response. I think that makes sense. Let's fix both issue here.

GavinZZ:
@sarisia I see that this PR will be closed soon. Do you plan to work on this same PR or open a new PR for the changes we discussed?

sarisia:
@GavinZZ sorry for delayed progress, been busy recentry. I'm planning to work on this PR.

pahud:
@sarisia Hi I am following up here. Are you still working on this PR?

sarisia:
@pahud sorry for taking so so long to work on this. Just pushed the fixes and rebased on the current main. Now it's ready for review.

pahud:
@sarisia 

The team is ready to review your PR but we'll need you to fix the CI errors from [Codebuild PR Build / build (pull_request)](https://github.com/aws/aws-cdk/actions/runs/17530203342/job/49786151614?pr=30778)

When that CodeBuild CI passes, I believe the team will start the review and feedback very soon.

pahud:
Hi @sarisia  

The team is now reviewing all community PRs and this one seems to be stale for a while. Are you still planning to continue the work? If yes, you will want to check the CI error make sure CI pass. Then feel free to tag me and I'll reach out to the maintainer for review. 

Thanks.

sarisia:
hi @pahud 
sorry for this PR taking so long,
I'll fix CI errors by next week.
Thank you for checking!

sarisia:
@pahud I think CI errors are gone now!

pahud:
Hey @sarisia thank you. We just want to make sure this PR is still active. I am always happy to help and make sure the PR is ready for review but I am not privileged to review any PR. Our maintainer should be looking at this PR very soon. Feel free to tag me if you need any further help.

Abogical:
FYI, this will be accepted to be merged after 1 more review from our team, as indicated in the `pr/needs-further-review` label.

Abogical (dismissed):

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
```diff
@@ -280,19 +284,25 @@ export class Trail extends Resource {
...
-          `AWSLogs/${props.orgId}/*`,
-        )],
-        actions: ['s3:PutObject'],
-        principals: [cloudTrailPrincipal],
-        conditions: {
-          StringEquals: {
-            's3:x-amz-acl': 'bucket-owner-full-control',
-            'aws:SourceArn': `arn:${this.stack.partition}:cloudtrail:${this.s3bucket.stack.region}:${this.s3bucket.stack.account}:trail/${props.trailName}`,
+      if (props.orgId === undefined) {
+        Annotations.of(this).addWarningV2('@aws-cdk/aws-cloudtrail:missingOrgIdForOrganizationTrail', 'Skipped attaching a policy to the bucket which allows organization trail to write because of missing orgId. Consider specifying orgId to add missing permissions');
```
gasolima, line 288: Can you explain to me the reason for choosing it warning vs throwing an error?
Is there any such use case that the user intentionally want to have `isOrganizationTrail` true but not send `orgId` or `trailName`?

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
```diff
@@ -280,19 +284,25 @@ export class Trail extends Resource {
...
+      } else if (props.trailName === undefined) {
```
gasolima, line 289: is there any reason for this if condition? 
I believe in L301
You can use something like 
```
'aws:SourceArn': `arn:${this.stack.partition}:cloudtrail:${this.s3bucket.stack.region}:${this.s3bucket.stack.account}:trail/${this.physicalName}`,
```

Abogical: The trail name is not necessarily equal to the physical name. How the trail name is generated if its not defined can be different than the physicalName.

gasolima: I can expect the majority users are under this category of not defining `trailName`.
I believe we can get the generated `trailName` from the [arn](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudtrail.CfnTrail.html#attrarn), then we don't have this if condition at all.

But with introducing this PR we're fixing the bug, but introducing another bug

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
sarisia: For backward compatibility and avoiding breaking changes. See https://github.com/aws/aws-cdk/pull/30778#issuecomment-2541850019

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
sarisia: Referencing trail's cloudformation attributes in bucket policy makes circular dependency which cannot be deployed so we need hard-coded value here.

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
gasolima: My question was a bit unclear, is there a real use case for having `isOrganizationTrail` and not send `orgId`?
According to @Abogical for ppl who want to have the policy by themselves.
So if that's the argument is valid i'm happy with it.

gasolima: but for other issues if we're going to break ppl cause we were not doing what they're expecting and now we're doing what they're expecting but failing we're happy to go this way, since it's better than failing silently.

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
gasolima: very good point thanks!

gasolima: Can you write a code comment in line 301 stating this fact? cause otherwise ppl will try to do what i'm asking in the future

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
```diff
@@ -280,19 +284,25 @@ export class Trail extends Resource {
...
+        this.s3bucket.addToResourcePolicy(new iam.PolicyStatement({
+          resources: [this.s3bucket.arnForObjects(
+            `AWSLogs/${props.orgId}/*`,
+          )],
+          actions: ['s3:PutObject'],
+          principals: [cloudTrailPrincipal],
+          conditions: {
+            StringEquals: {
+              's3:x-amz-acl': 'bucket-owner-full-control',
+              'aws:SourceArn': `arn:${this.stack.partition}:cloudtrail:${this.s3bucket.stack.region}:${this.s3bucket.stack.account}:trail/${props.trailName}`,
```
gasolima, line 302:
```suggestion
// trailName can't stripped out from arn attribute, since it will make circular dependency
              'aws:SourceArn': `arn:${this.stack.partition}:cloudtrail:${this.s3bucket.stack.region}:${this.s3bucket.stack.account}:trail/${props.trailName}`,
```

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
sarisia: I don't think there's a real use case too so I'happy to change these warnings to throwing error if introducing breaking changes is acceptable here!

packages/aws-cdk-lib/aws-cloudtrail/lib/cloudtrail.ts
sarisia: Thank you!

Abogical (approved):

Abogical:
@mergifyio update

PR #30788, opened 2024-07-08
fix(stepfunctions-tasks): run task perm no longer valid

### LINKED ISSUE #30751: aws-cdk-lib/aws-stepfunctions: incorrect resource in ECS ecs:RunTask for State Machines
### Describe the bug

The policy generated in case of ECS tasks trigger in a state machine are of this type:
```
{
    "Action": "ecs:RunTask",
	"Resource": [
		"arn:aws:ecs:eu-central-1:xxxx:task-definition/my-task:*",
		"arn:aws:ecs:eu-central-1:xxxx:task-definition/my-task"
	],
	"Effect": "Allow"
},
```
the Resource `"arn:aws:ecs:eu-central-1:xxxx:task-definition/my-task"` is not a valid one, the policy validator fail in the UI (even if I'm able to deploy) and there is an AWS notification about my state machine role.

### Expected Behavior

The resource used for ecs:RunTask is simply:
`"arn:aws:ecs:eu-central-1:xxxx:task-definition/my-task:*",`

### Current Behavior

the resources for ecs:RunTask  are:
- "arn:aws:ecs:eu-central-1:xxxx:task-definition/my-task:*",
- "arn:aws:ecs:eu-central-1:xxxx:task-definition/my-task"

### Reproduction Steps

Create a state machine invoking an ecs task

### Possible Solution

Simply remove the not necessary resource from then policy attach to the IAM role used by the statemachine.
Creating a role, and passing to sfn.StateMachine doesn't help, because an inline policy with the wrong inline policy is attach to the custom role.

_No response_

### CDK CLI Version

2.147.3

### Framework Version

_No response_

### Node.js Version

v18.0.0

### OS

MacOs

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

msambol:
### Issue # (if applicable)

Closes #30751.

### Reason for this change

`runTask` on `${taskDefinitionFamilyArn}` is no longer relevant (see validation errors in the linked issue.
This was currently disabled with a FF. This PR removes the permission entirely, and removes the FF.

### Checklist

----


msambol:
@GavinZZ JFYI since you have the most context on this..

msambol:
Exemption request: The related integrations tests already had this permission removed.

GavinZZ (commented):
I think this looks good to me. I left a question in the relevant issue [here](https://github.com/aws/aws-cdk/issues/30751#issuecomment-2244028567) to help me understand the issue better. I just want to confirm this does not work for the current users to make sure it's non-breaking.

GavinZZ (approved):
I think this is good to merge since the removed resource in IAM policy is not a valid arn and thus removing this should not create any breaking changes.

GavinZZ (approved):

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/ecs/run-task.ts
```diff
@@ -347,31 +346,11 @@ export class EcsRunTask extends sfn.TaskStateBase implements ec2.IConnectable {
...
-          ],
-        }),
-      );
-    }
-
-    policyStatements.push(
+    const policyStatements = [
+      new iam.PolicyStatement({
+        actions: ['ecs:RunTask'],
+        resources: [`${this.getTaskDefinitionFamilyArn()}:*`],
```
samson-keung, line 352: If I understand correctly, `this.getTaskDefinitionFamilyArn()` strips away the revision number from `this.props.taskDefinition.taskDefinitionArn`. I am curious, would the CDK user be able to specify specific revision of the task definition if they would like to scope this down for strictest security?

GavinZZ: Yes, that's correct understanding. 

> I am curious, would the CDK user be able to specify specific revision of the task definition if they would like to scope this down for strictest security

Not easily, they could use escape hatch to scope it down further to a specific revision number.

PR #30817, opened 2024-07-10
chore(ec2): add `mac2-m1ultra` instance type

badmintoncryer:
### Issue # (if applicable)

None

### Reason for this change

EC2 instance supports `mac2-m1ultra` instance type but `InstanceType` class does not support it.

### Description of changes

Add `MAC2_M1ULTRA` instance type

None

### Checklist

----


wladyslawczyzewski (approved):
Looks good to me.

Am I right to think, that there's no integration tests because of minimum 24h allocation of Mac instance?

badmintoncryer:
@wladyslawczyzewski I think it is unnecessary to create integ test for a chore PR🤨

packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
```diff
@@ -1632,6 +1642,8 @@ export class InstanceType {
       [InstanceClass.MAC2_M2]: 'mac2-m2',
       [InstanceClass.MACINTOSH2_M2_PRO]: 'mac2-m2pro',
       [InstanceClass.MAC2_M2PRO]: 'mac2-m2pro',
+      [InstanceClass.MACINTOSH2_M1_ULTRA]: 'mac2-m1ultra',
```
moelasmar, line 1645: should the value be `macintosh2-m1-ultra` ?

moelasmar:
can you add an integration test case

TheRealAmazonKendra (changes requested):
Putting this into changes requested per @moelasmar's comments.

packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
badmintoncryer: I've fixed!

badmintoncryer:
@moelasmar 
> can you add an integration test case

I've added integ test. Could you confirm it?

packages/aws-cdk-lib/aws-ec2/lib/instance-types.ts
```diff
@@ -1632,6 +1642,8 @@ export class InstanceType {
       [InstanceClass.MAC2_M2]: 'mac2-m2',
       [InstanceClass.MACINTOSH2_M2_PRO]: 'mac2-m2pro',
       [InstanceClass.MAC2_M2PRO]: 'mac2-m2pro',
+      [InstanceClass.MACINTOSH2_M1_ULTRA]: 'macintosh2-m1-ultra',
```
tmokmss, line 1645: Where does this string come from? I cannot find [instance type](https://aws.amazon.com/ec2/instance-types/) named macintosh2-m1-ultra.

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.instance-m1-ultra.ts
```diff
@@ -0,0 +1,60 @@
...
+      availabilityZones: [az],
+      subnetType: ec2.SubnetType.PRIVATE_ISOLATED,
+    }).subnets[0];
+
+    const instanceType = ec2.InstanceType.of(ec2.InstanceClass.MAC2_M1ULTRA, ec2.InstanceSize.METAL);
+
+    const dedicatedHost = new ec2.CfnHost(this, 'MyDedicatedHost', {
+      availabilityZone: az,
+      instanceType: instanceType.toString(),
+    });
```
tmokmss, line 36: Does this integ-test even work? An EC2 Mac dedicated host cannot be deleted until 24 hours after the creation, so I guess integ-runner will fail when it tries to delete the stack.

> Mac instances are available only as bare metal instances on [Dedicated Hosts](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/dedicated-hosts-overview.html), with a minimum allocation period of 24 hours before you can release the Dedicated Host. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-mac-instances.html

Given this restriction, I don't think it's a good idea to add an integration test for EC2 Mac. Even if we set the host's removalPolicy to retain, we have to manually remove the host 24 hours later, which is easily forgotten and not ideal. Also, it [costs](https://aws.amazon.com/jp/ec2/dedicated-hosts/pricing/) at least 120USD (5USD/h * 24h for mac2-m1ultra) every time we update the snapshot.

Since CDK currently only provides the enum for EC2 Mac instance type, I think there's little benefit to testing it with the expensive integ test. Any thoughts? @moelasmar @badmintoncryer

badmintoncryer: @tmokmss Thank you for your comment. I completely agree with your opinion. I also feel that the benefits of conducting an integration test in this case are minimal.

Separately, as you mentioned, the integration test is failing. However, since the deployment was successful and the snapshot was generated, I was considering including it in the PR as it is.

I have encountered a similar situation before where a test failed due to resource deletion issues, but it’s good to know that this can be avoided by setting the RemovalPolicy. This is very helpful information.

moelasmar: thanks @tmokmss, it is a good point. I was thinking that even it is some new enum value, but some how we need a testing coverage, so incase if it got removed by mistake some how, we still have some testing to cover us. So, I think some unit testing is enough. we can add a unit test case for it, and it will be enough.

tmokmss: Thanks. Yes I agree that unit test will suffice for that purpose.

packages/aws-cdk-lib/aws-ec2/test/instance-type.test.ts
```diff
@@ -0,0 +1,8 @@
+import { InstanceClass, InstanceSize, InstanceType } from '../lib';
```
badmintoncryer, line 1: @moelasmar I've added simple unit test for `InstanceType` class. Is it enough?

moelasmar: it looks good

moelasmar (approved):

PR #30821, opened 2024-07-10
fix(ec2): prevent deduplication of init command args

### LINKED ISSUE #26221: ec2: InitCommand.argvCommand dedupes its args, causing the wrong command to be generated)
### Describe the bug

using `ec2.InitCommand.argvCommand()` can, in some circumstances, remove duplicate strings from the input array, causing unexpected output.

### Expected Behavior

Expected the following

```typescript
ec2.InitCommand.argvCommand([
  'useradd', '-u', '1001', '-g', '1001', 'eguser',
]),
ec2.InitCommand.argvCommand([
  'useradd', '-a', '-u', '1001', '-g', '1001', 'eguser',
]),
```

To generate the following YAML:

```yaml
          commands:
            "000":
              command:
                - useradd
                - -u
                - "1001"
                - -g
                - "1001"
                - eguser
            "001":
              command:
                - useradd
                - -a
                - -u
                - "1001"
                - -g
                - "1001"
                - eguser
```

### Current Behavior

The following YAML was generated. Note the omission of the second occurrence of '1001' in the second command:

```yaml
          commands:
            "000":
              command:
                - useradd
                - -u
                - "1001"
                - -g
                - "1001"
                - eguser
            "001":
              command:
                - useradd
                - -a
                - -u
                - "1001"
                - -g
                - eguser
```

### Reproduction Steps

The CDK App in the provided repo contains the minimal resources needed to demonstrate the bug when running `cdk synth`:

https://github.com/fidothe/cdk-ec2-construct-bug-repro

`git clone https://github.com/fidothe/cdk-ec2-construct-bug-repro && cd cdk-ec2-construct-bug-repro && cdk synth`

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.85.0 (build 4e0d726)

### Framework Version

_No response_

### Node.js Version

v18.16.1

### OS

macOS 13.4.1

### Language

Typescript

### Language Version

Typescript (5.1.5)

### Other information

_No response_

Leo10Gama:
### Issue # (if applicable)

Closes #26221 

### Reason for this change

Previously, using `ec2.InitCommand.argvCommand()` would remove some duplicate strings in the input array. This produces an incorrect command in the template, leading to unexpected behaviour.

### Description of changes

An additional line was added to the `deepMerge` function that is called in the `InitConfig.bindForType()` method, which checks the key of the input array, preventing it from becoming a set (removing duplicates) if it is a list of commands.

A unit test was added to generate an `AWS::CloudFormation::Init` resource identical to the one reproduced in the issue. The test was run and failed before the changes were made, and following the changes in `cfn-init.ts`, the test passed.

### Checklist

----


packages/aws-cdk-lib/aws-ec2/lib/cfn-init.ts
```diff
@@ -311,10 +311,14 @@ function deepMerge(target?: Record<string, any>, src?: Record<string, any>) {
...
-        ...value,
-      ]));
+      if (key != 'command') { // don't deduplicate command arguments
+        target[key] = Array.from(new Set([
+          ...target[key] ?? [],
+          ...value,
+        ]));
+      } else {
+        target[key] = value;
+      }
```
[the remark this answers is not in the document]
Leo10Gama: I'm not entirely sure, but it may be worth updating so nothing slips through the cracks. I can change the logic on line 320 to be similar to line 315, though instead of creating the array from a new `Set`, I can use a new `Array`.

packages/aws-cdk-lib/aws-ec2/lib/cfn-init.ts
```diff
@@ -311,10 +311,17 @@ function deepMerge(target?: Record<string, any>, src?: Record<string, any>) {
...
+        target[key] = Array.from(new Set([
+          ...target[key] ?? [],
+          ...value,
+        ]));
+      } else {
+        target[key] = new Array(
+          ...target[key] ?? [],
+          ...value,
+        );
+      }
```
scanlonp, line 324: I would like the logic to have positive paths with the `else` as the catch all. So if key === 'command', follow the new behavior (don't dedup), otherwise keep the behavior the same.

scanlonp (changes requested):
Looking good with a couple small comments:

1. Let's switch the logic flow. It will be easier to understand with the specific cases handled first.
2. Do we have any current unit tests that test for de-duplication on non-command uses of the deep merge? If yes then we are all set. If not, we should probably put that case in to verify that this change stays scoped to the individual function we are targeting.

Leo10Gama:
> 2. Do we have any current unit tests that test for de-duplication on non-command uses of the deep merge? If yes then we are all set. If not, we should probably put that case in to verify that this change stays scoped to the individual function we are targeting.

From what it seems like, this specific `deepMerge` is only used in this file by the `bind` functions, and it doesn't seem like it's used outside of this context. The only other `deepMerge` tests I can find are for a separate util that doesn't have to do with this function. But that is a good catch, I'll include tests to make sure the de-duping stays consistent.

scanlonp (commented):
My comments are addressed. I am good with the PR. Going to wait for a second review for approval.

packages/aws-cdk-lib/aws-ec2/test/cfn-init.test.ts
```diff
@@ -136,6 +136,60 @@ test('empty configs are not rendered', () => {
...
+    },
+  });
+});
+
+test('deepMerge properly deduplicates non-command arguments', () => {
+  // WHEN
+  const config = new ec2.InitConfig([
+    ec2.InitSource.fromUrl('/tmp/foo', 'https://amazon.com/foo.zip'),
+    ec2.InitSource.fromUrl('/tmp/foo', 'https://amazon.com/foo.zip'),
+    ec2.InitSource.fromUrl('/tmp/foo', 'https://amazon.com/foo.zip'),
```
comcalvi, line 182: can you make this:

```
ec2.InitSource.fromUrl('/tmp/foo', 'https://amazon.com/foo.zip'),
ec2.InitSource.fromUrl('/tmp/foo', 'https://amazon.com/foo.zip'),
ec2.InitSource.fromUrl('/tmp/foo', 'https://amazon.com/bar.zip'),
ec2.InitSource.fromUrl('/tmp/foo', 'https://amazon.com/bar.zip'),
ec2.InitSource.fromUrl('/tmp/bar', 'https://amazon.com/bar.zip'),
ec2.InitSource.fromUrl('/tmp/bar', 'https://amazon.com/bar.zip'),
ec2.InitSource.fromUrl('/tmp/bar', 'https://amazon.com/foobar.zip'),
ec2.InitSource.fromUrl('/tmp/bar', 'https://amazon.com/foobar.zip'),
```

or something like this so we can see whether the dedeup logic applies to both the first parameter and the second, or if it's just the second?

comcalvi (changes requested):

packages/aws-cdk-lib/aws-ec2/test/cfn-init.test.ts
Leo10Gama: Good catch, will include some more dedup test values.

comcalvi (approved):
I approve of the 👻

scanlonp (approved):

PR #30822, opened 2024-07-10
fix(apigateway): authorization scope is not added when not explicitly defining authorization type in method or root api

### LINKED ISSUE #30444: aws-apigateway: Authorization Scopes not rendered with CognitoUserPoolsAuthorizer
Please add your +1 👍 to let us know you have encountered this
---

#### Describe the bug

When using the `CognitoUserPoolsAuthorizer` with `authorizationScope`, the scopes are not rendered to the CloudFormation template, if `authorizationType` is not set explicitly. This worked before version `2.142.0`.

#### Expected Behavior

Scopes are rendered to the CloudFormation template when using CognitoUserPoolsAuthorizer without setting authorizationType explicitly. The authorizationType of the authorizer should be used implicitly.

#### Current Behavior

Scopes are not rendered to the CloudFormation template when using CognitoUserPoolsAuthorizer without setting authorizationType explicitly. They are only rendered, when authorizationType is set explicitly.

A warning is printed during CDK synth:
> 'AuthorizationScopes' can only be set when 'AuthorizationType' sets 'COGNITO_USER_POOLS'. Default to ignore the values set in 'AuthorizationScopes'.

#### Reproduction Steps

```typescript
import { Stack } from "aws-cdk-lib";
import { UserPool } from "aws-cdk-lib/aws-cognito";
import { CognitoUserPoolsAuthorizer, RestApi } from "aws-cdk-lib/aws-apigateway";
import { Template } from "aws-cdk-lib/assertions";

const stack = new Stack(undefined, "Stack");
const userPool = UserPool.fromUserPoolId(stack, "UserPool", "userPoolId");
const authorizer = new CognitoUserPoolsAuthorizer(stack, "Authorizer", {
  cognitoUserPools: [userPool],
});
const restApi = new RestApi(stack, "RestApi", {
  deploy: true,
  defaultMethodOptions: {
    authorizer,
    // here we need to add the authorizationType to make it work
  },
});
restApi.root.resourceForPath("/user/profile").addMethod("GET", undefined, {
  authorizationScopes: [OAuthScope.PROFILE.scopeName], // this scope is missing
});
restApi.root.resourceForPath("/any/other").addMethod("POST");

console.log(Template.fromStack(stack).toJSON());
```

### Workaround:
There are 2 workarounds:
 - you can pin the aws-cdk-lib version to `2.141.0`.
 - You can update the RestApi definition as following
 ```
const restApi = new RestApi(stack, "RestApi", {
  deploy: true,
  defaultMethodOptions: {
    authorizer,
    // setting it explicit:
    authorizationType: authorizer.authorizationType, 
  },
});
```

### Solution:
We are reverting [this PR](https://github.com/aws/aws-cdk/pull/30035) that introduces the breaking change.

GavinZZ:
### Issue # (if applicable)

Closes #30444

### Reason for this change

The original PR caused a breaking change, we can't rollback because it was released in v2.142.0 and it fixes customers issues (partially). Simply doing a revert will be breaking for those customers again.

### Description of changes

Identified the root cause and we should use `AuthorizationType` instead of `AuthorizationTypeOption`. `AuthorizationType` defaults to find the authorization type from the authorizer, falling back to use the auth type defined in the `Method` construct's options property and falling back to `None`.

`AuthorizationTypeOptions` on the other hand tries to find the auth type from `Method` construct's options property which can be None because it's optional.

New unit tests covering the changes and new integration tests covering it.

### Checklist

----


moelasmar (approved):

moelasmar (approved):

GavinZZ:
@mergify update

PR #30830, opened 2024-07-11
feat(sns): add delivery policy to sns subscriptions

### LINKED ISSUE #17576: (sns): Allow delivery retry policy configuration of HTTPS subscription on SNS topic
### Description

While creating HTTPS Subscription on SNS topic, there's no way to configure the retry policy.

### Use Case

I am creating SNS topic and HTTPS subscription using CDK and want to provide custom delivery retry policy.

### Proposed Solution

Update UrlSubscriptionProps interface to provide deliveryRetryPolicy as well.

### Other information

_No response_

### Acknowledge

LaurenceWarne:
### Issue # (if applicable)

Closes #17576.

### Reason for this change

To allow setting of delivery policy for appropriate sns subscriptions.

### Description of changes

Subscriptions can now take a new parameter `deliveryPolicy` which consists of a `healthyRetryPolicy`, a `throttlePolicy`, and a `requestPolicy` (each having the same parameters described here: https://docs.aws.amazon.com/sns/latest/dg/sns-message-delivery-retries.html#creating-delivery-policy).

New integration test along with unit tests.

### Checklist

----


packages/aws-cdk-lib/aws-sns/lib/delivery-policy.ts
```diff
@@ -0,0 +1,100 @@
+/**
+ * Algorithms which can be used by SNS to calculate the delays associated with all of the retry attempts between the first and last retries in the backoff phase.
+ */
+export enum BackoffFunction {
+  /** Arithmetic.
```
LaurenceWarne, line 5: Not the best doc I've ever written :sweat_smile: ,  the best indicator of to what these each of these do that I found was an image at the very bottom of the doc page here: https://docs.aws.amazon.com/sns/latest/dg/sns-message-delivery-retries.html - suggestions welcome.

packages/aws-cdk-lib/aws-sns/lib/delivery-policy.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * Options for customising the retry policy of the delivery of SNS messages to HTTP/S endpoints.
+ */
+export interface HealthyRetryPolicy {
+  /** The minimum delay for a retry (in seconds).  Must be at least 1 and not exceed `maxDelayTarget`.
+   * @default 20
+   */
+  readonly minDelayTarget: number;
```
LaurenceWarne, line 46: Only, this and the next two appear to be mandatory if `healthyRetryPolicy` is specified, thought I can't find any specific doc for this.

pahud:
Hi

This PR has been pending for community review for a while. Please consider posting it on the #contributing channel in cdk.dev. Community members will be on the lookout there as well for possible reviews.

Check [How to get your P2 PR community reviewed](https://github.com/aws/aws-cdk/wiki/CDK-Community-PR-Reviews#how-to-get-your-p2-pr-community-reviewed) for more details.

LaurenceWarne:
Thanks, I've posted on the channel.

packages/aws-cdk-lib/aws-sns/lib/delivery-policy.ts
badmintoncryer: Would it be correct to assume that if these are not set, it will result in an error during deployment? If so, I am curious when the default values would be set.

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -136,6 +144,43 @@ export class Subscription extends Resource {
...
+    if (props.deliveryPolicy) {
+      if (props.protocol !== SubscriptionProtocol.HTTP && props.protocol !== SubscriptionProtocol.HTTPS) {
+        throw new Error('Delivery policy is only supported for HTTP and HTTPS subscriptions');
+      }
+      const { healthyRetryPolicy, throttlePolicy } = props.deliveryPolicy;
+      if (healthyRetryPolicy) {
+        const delayTargetLimit = 3600;
+        if (healthyRetryPolicy.minDelayTarget < 1 || healthyRetryPolicy.minDelayTarget > delayTargetLimit) {
+          throw new Error(`minDelayTarget must be between 1 and ${delayTargetLimit} inclusive`);
+        }
```
badmintoncryer, line 156: In order to support Duration, please make the following modifications.

```suggestion
        const minDelayTarget = healthyRetryPolicy.minDelayTarget
        const maxDelayTarget = healthyRetryPolicy.maxDelayTarget
        if (minDelayTarget.toMilliseconds() < Duration.seconds(1).toMilliseconds()) {
          throw new Error(`minDelayTarget must be at least 1 second, got ${minDelayTarget.toMilliseconds()}ms.`);
        }
        if (maxDelayTarget.toMilliseconds() < Duration.seconds(1).toMilliseconds()) {
          throw new Error(`maxDelayTarget must be at least 1 second, got ${maxDelayTarget.toMilliseconds()}ms.`);
        }
        if (minDelayTarget.toSeconds() < 1 || minDelayTarget.toSeconds() > delayTargetLimit) {
          throw new Error(`minDelayTarget must be between 1 and ${delayTargetLimit} seconds inclusive`);
        }
```

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -136,6 +144,43 @@ export class Subscription extends Resource {
...
+        if (healthyRetryPolicy.maxDelayTarget < 1 || healthyRetryPolicy.maxDelayTarget > delayTargetLimit) {
+          throw new Error(`maxDelayTarget must be between 1 and ${delayTargetLimit} inclusive`);
+        }
+        if (healthyRetryPolicy.minDelayTarget > healthyRetryPolicy.maxDelayTarget) {
+          throw new Error('minDelayTarget must not exceed maxDelayTarget');
+        }
```
badmintoncryer, line 162: Could you also adjust these if statements to support Duration?

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -136,6 +144,43 @@ export class Subscription extends Resource {
...
+        const numRetriesLimit = 100;
+        if (healthyRetryPolicy.numRetries < 0 || healthyRetryPolicy.numRetries > numRetriesLimit) {
```
badmintoncryer, line 164: Is it necessary to validate that numRetries is an Integer?

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -136,6 +144,43 @@ export class Subscription extends Resource {
       throw new Error('Subscription role arn is required field for subscriptions with a firehose protocol.');
     }
 
+    if (props.deliveryPolicy) {
```
badmintoncryer, line 147: How about performing these validations inside the `renderDeliveryPolicy()`?

badmintoncryer (changes requested):
Thank you for creating such a wonderful PR! I’ve quickly reviewed it!

LaurenceWarne:
> Thank you for creating such a wonderful PR! I’ve quickly reviewed it!

Thanks for the review!  Hopefully I've addressed everything

packages/aws-cdk-lib/aws-sns/lib/delivery-policy.ts
LaurenceWarne: If I recall correctly from testing, it's required if and only if `healthyRetryPolicy` itself is specified, and so the default values would be set in the case `healthyRetryPolicy` is not specified (perhaps it's confusing to mention a default in the doc :thinking:)

packages/aws-cdk-lib/aws-sns/lib/delivery-policy.ts
LaurenceWarne: > I would appreciate it if you could consult with the maintainer again.

As in a CDK maintainer?

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
LaurenceWarne: Yeah good catch

packages/aws-cdk-lib/aws-sns/lib/delivery-policy.ts
badmintoncryer: Yes! After my approval, you need to get approval from the CDK maintainer. I would appreciate it if you could kindly check it at that time!

packages/aws-cdk-lib/aws-sns/lib/delivery-policy.ts
badmintoncryer: How about making these two props optional and explicitly setting the default values documented within the L2 construct? These properties may be required in CloudFormation, but I felt there is no need to make them mandatory in the L2 construct.

```ts
  readonly minDelayTarget?: Duration;
  readonly maxDelayTarget?: Duration;
```

```ts
const minDelayTarget = healthyRetryPolicy.minDelayTarget ?? Duration.seconds(20);
const maxDelayTarget = healthyRetryPolicy.maxDelayTarget ?? Duration.seconds(20);
```

packages/aws-cdk-lib/aws-sns/lib/delivery-policy.ts
```diff
@@ -0,0 +1,127 @@
+import { Duration } from '../../core';
```
badmintoncryer, line 1: Could the contents of `delivery-policy.ts` be used elsewhere? If it’s only going to be used in `subscription.ts`, I thought it might be better to include everything in `subscriptions.ts`.

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -152,10 +160,75 @@ export class Subscription extends Resource {
...
+      }
+      if (healthyRetryPolicy.numNoDelayRetries && healthyRetryPolicy.numNoDelayRetries < 0) {
+        throw new Error('numNoDelayRetries must be zero or greater');
+      }
+      if (healthyRetryPolicy.numMinDelayRetries && healthyRetryPolicy.numMinDelayRetries < 0) {
+        throw new Error('numMinDelayRetries must be zero or greater');
+      }
+      if (healthyRetryPolicy.numMaxDelayRetries && healthyRetryPolicy.numMaxDelayRetries < 0) {
+        throw new Error('numMaxDelayRetries must be zero or greater');
+      }
```
badmintoncryer, line 206: Cloud you please show the props value like below?

```ts
        throw new Error(`numRetries must be between 0 and ${numRetriesLimit} inclusive, got: ${healthyRetryPolicy.numRetries}`);
```

badmintoncryer (changes requested):
I Thank you for the revisions. I’ve added a few more comments. Once you address those, I plan to approve.

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
badmintoncryer: Has the integer check been added?

LaurenceWarne: Sorry, forgot to push a commit :sweat_smile:

packages/aws-cdk-lib/aws-sns/lib/delivery-policy.ts
LaurenceWarne: Ah right np, I thought you might have meant someone in the SNS team or something :+1:

badmintoncryer: Ref: https://github.com/aws/aws-cdk/wiki/CDK-Community-PR-Reviews

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -152,10 +160,85 @@ export class Subscription extends Resource {
...
+    }
+    return {
+      healthyRetryPolicy: healthyRetryPolicy ? {
+        // minDelayTarget, maxDelayTarget and numRetries are (empirically) mandatory when healthyRetryPolicy is specified,
+        // but for user-friendliness we allow them to be undefined and set them here instead.
+        // The defaults we use here are the same used in the event healthyRetryPolicy is not specified, see https://docs.aws.amazon.com/sns/latest/dg/creating-delivery-policy.html.
+        minDelayTarget: (healthyRetryPolicy.minDelayTarget === undefined)? 20 : healthyRetryPolicy.minDelayTarget.toSeconds(),
+        maxDelayTarget: (healthyRetryPolicy.maxDelayTarget === undefined)? 20 : healthyRetryPolicy.maxDelayTarget.toSeconds(),
+        numRetries: (healthyRetryPolicy.numRetries === undefined)? 3: healthyRetryPolicy.numRetries,
```
badmintoncryer, line 227:
```suggestion
        minDelayTarget: (healthyRetryPolicy.minDelayTarget === undefined) ? 20 : healthyRetryPolicy.minDelayTarget.toSeconds(),
        maxDelayTarget: (healthyRetryPolicy.maxDelayTarget === undefined) ? 20 : healthyRetryPolicy.maxDelayTarget.toSeconds(),
        numRetries: (healthyRetryPolicy.numRetries === undefined) ? 3: healthyRetryPolicy.numRetries,
```

badmintoncryer (approved):
Thank you for you contribution! I've approved this, but I kindly request a minor correction.

LaurenceWarne:
> Thank you for you contribution! I've approved this, but I kindly request a minor correction.

:+1: And thanks again, I really appreciate you taking the time :slightly_smiling_face:

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -152,10 +160,85 @@ export class Subscription extends Resource {
       region: props.region,
       redrivePolicy: this.buildDeadLetterConfig(this.deadLetterQueue),
       subscriptionRoleArn: props.subscriptionRoleArn,
+      deliveryPolicy: props.deliveryPolicy ? this.renderDeliveryPolicy(props.deliveryPolicy, props.protocol): undefined,
     });
 
   }
 
+  private renderDeliveryPolicy(deliveryPolicy: DeliveryPolicy, protocol: SubscriptionProtocol): any {
+    if (protocol !== SubscriptionProtocol.HTTP && protocol !== SubscriptionProtocol.HTTPS) {
```
GavinZZ, line 169: nit: 
```suggestion
    if (![SubscriptionProtocol.HTTP, SubscriptionProtocol.HTTPS].includes(protocol)) {
```

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -152,10 +160,85 @@ export class Subscription extends Resource {
...
+      throw new Error(`Delivery policy is only supported for HTTP and HTTPS subscriptions, got: ${protocol}`);
+    }
+    const { healthyRetryPolicy, throttlePolicy } = deliveryPolicy;
+    if (healthyRetryPolicy) {
+      const delayTargetLimitSecs = 3600;
+      const minDelayTarget = healthyRetryPolicy.minDelayTarget;
+      const maxDelayTarget = healthyRetryPolicy.maxDelayTarget;
+      if (minDelayTarget !== undefined) {
+        if (minDelayTarget.toMilliseconds() % 1000 !== 0) {
```
GavinZZ, line 178: I think you can directly use `toSeconds()` which will check if it's a integer second

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -152,10 +160,85 @@ export class Subscription extends Resource {
...
+        const maxDelayTargetSecs = maxDelayTarget.toSeconds();
+        if (maxDelayTargetSecs < 1 || maxDelayTargetSecs > delayTargetLimitSecs) {
+          throw new Error(`maxDelayTarget must be between 1 and ${delayTargetLimitSecs} seconds inclusive, got: ${maxDelayTargetSecs}s`);
+        }
+        if ((minDelayTarget !== undefined) && minDelayTarget.toSeconds() > maxDelayTargetSecs) {
+          throw new Error('minDelayTarget must not exceed maxDelayTarget');
+        }
+      }
+
+      const numRetriesLimit = 100;
```
GavinZZ, line 199: It's good intention to not use the magic value and define it explicitly. But given the number of validations, I actually think it's more readable to directly use the value like 100 and 3600 in the if condition checks. What do you think?

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -152,10 +160,85 @@ export class Subscription extends Resource {
...
+    }
+    if (throttlePolicy) {
+      const maxReceivesPerSecond = throttlePolicy.maxReceivesPerSecond;
+      if (maxReceivesPerSecond !== undefined && (maxReceivesPerSecond < 1 || !Number.isInteger(maxReceivesPerSecond))) {
+        throw new Error(`maxReceivesPerSecond must be an integer greater than zero, got: ${maxReceivesPerSecond}`);
+      }
+    }
+    return {
+      healthyRetryPolicy: healthyRetryPolicy ? {
+        // minDelayTarget, maxDelayTarget and numRetries are (empirically) mandatory when healthyRetryPolicy is specified,
```
GavinZZ, line 222: can you explain what you meant by `empirically mandatory)`? I do see that they do have default values set.

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
```diff
@@ -152,10 +160,85 @@ export class Subscription extends Resource {
...
+        // but for user-friendliness we allow them to be undefined and set them here instead.
+        // The defaults we use here are the same used in the event healthyRetryPolicy is not specified, see https://docs.aws.amazon.com/sns/latest/dg/creating-delivery-policy.html.
+        minDelayTarget: (healthyRetryPolicy.minDelayTarget === undefined) ? 20 : healthyRetryPolicy.minDelayTarget.toSeconds(),
+        maxDelayTarget: (healthyRetryPolicy.maxDelayTarget === undefined) ? 20 : healthyRetryPolicy.maxDelayTarget.toSeconds(),
+        numRetries: (healthyRetryPolicy.numRetries === undefined) ? 3: healthyRetryPolicy.numRetries,
+        numNoDelayRetries: healthyRetryPolicy.numNoDelayRetries,
+        numMinDelayRetries: healthyRetryPolicy.numMinDelayRetries,
+        numMaxDelayRetries: healthyRetryPolicy.numMaxDelayRetries,
```
GavinZZ, line 230: Should these have default values?

GavinZZ (dismissed):
Thanks for contributing! Left some feedback.

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
LaurenceWarne: I thought about it, but I thought this would be better as the error `toSeconds` raises doesn't have the same context we can add here?

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
LaurenceWarne: I think I prefer it how it is now, I agree it might be more readable, but I'd prefer keeping hard-coding to a minimum to make refactors safer/simplier, etc

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
LaurenceWarne: Does https://github.com/aws/aws-cdk/pull/30830#discussion_r1770405277 answer your question (also see last comment in that thread as well as to why they're optional)?

packages/aws-cdk-lib/aws-sns/lib/subscription.ts
LaurenceWarne: Similar to https://github.com/aws/aws-cdk/pull/30830#discussion_r1796169589 question, I guess it would depend on your view on that :sweat_smile:?

My thinking was that if Cloudformation doesn't require these to be specified when `healthyRetryPolicy` is specified, we can omit them here and have Cloudformation figure the defaults for us?

GavinZZ (approved):

badmintoncryer:
@mergify update

adoblas:
@GavinZZ do we have an ETA for this merge?

badmintoncryer:
@GavinZZ Could you please merge main branch and approve this PR again?

GavinZZ (approved):

mwebber:
@LaurenceWarne , @GavinZZ it looks like the reason this PR hasn't proceeded to the next is that we have 2 code coverage test failures (see https://app.codecov.io/gh/aws/aws-cdk/pull/30830/indirect-changes). It looks like the 2 failures are a single problem. 

The failures just look like some strange side-effect of the change, not an actual problem with the PR. However, I don't think the system will let you move forward until they are addressed.

Is someone able to resolve those failures? Sorry I have no idea how to fix them myself.

LaurenceWarne:
> The failures just look like some strange side-effect of the change, not an actual problem with the PR. However, I don't think the system will let you move forward until they are addressed.
> 
> Is someone able to resolve those failures? Sorry I have no idea how to fix them myself.

Merged master and it looks to have fixed itself

GavinZZ:
Yes, apologize that I didn't get back to you sooner. We've fixed issue with codecov now. Will approve and merge again.

GavinZZ (approved):

PR #30832, opened 2024-07-11
chore: remove use of deprecated ServicePrincipal Mapping

TheRealAmazonKendra:
They have now been standardized for a few years. We did not initially remove the old mappings out of caution and because we were unsure that the changes has made it to all regions yet. It is long past that happening at this point.

Because we never removed this or marked it as deprecated, we still have a not insignificant amount of customers who believe the individual mapping is necessary and cut tickets because it is not up-to-date.

### Issue # (if applicable)

Closes #<issue number here>.

### Checklist

----


packages/aws-cdk-lib/region-info/lib/fact.ts
```diff
@@ -226,9 +226,11 @@ export class FactName {
    * @param service the service name, either simple (e.g: `s3`, `codedeploy`) or qualified (e.g: `s3.amazonaws.com`).
    *                The `.amazonaws.com` and `.amazonaws.com.cn` domains are stripped from service names, so they are
    *                canonicalized in that respect.
+   *
+   * @deprecated - Use `iam.ServicePrincipal.servicePrincipalName()` instead.
    */
   public static servicePrincipal(service: string): string {
-    return `service-principal:${service.replace(/\.amazonaws\.com(\.cn)?$/, '')}`;
+    return `${service.replace(/\.amazonaws\.com(\.cn)?$/, '')}.amazonaws.com`;
```
comcalvi, line 233: isn't this breaking? I know you've deprecated it, but you're changing the behavior of this method

packages/aws-cdk-lib/region-info/lib/region-info.ts
```diff
@@ -124,9 +124,11 @@ export class RegionInfo {
   /**
    * The name of the service principal for a given service in this region.
    * @param service the service name (e.g: s3.amazonaws.com)
+   *
+   * @deprecated - Use `iam.ServicePrincipal.servicePrincipalName()` instead.
    */
   public servicePrincipal(service: string): string | undefined {
-    return Fact.find(this.name, FactName.servicePrincipal(service));
+    return `${service.replace(/\.amazonaws\.com(\.cn)?$/, '')}.amazonaws.com`;
```
comcalvi, line 131: breaking change, no?

comcalvi (changes requested):

packages/aws-cdk-lib/region-info/lib/fact.ts
TheRealAmazonKendra: This formatting was to do a lookup in the database, which resulted in returning what I have now changed it to.

packages/aws-cdk-lib/region-info/lib/region-info.ts
TheRealAmazonKendra: Same as above.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/glue/integ.glue-task.js.snapshot/aws-stepfunctions-integ.template.json
```diff
@@ -1,253 +1,253 @@
...
+    "GlueJobRole1CD031E0": {
+      "Type": "AWS::IAM::Role",
+      "Properties": {
+        "AssumeRolePolicyDocument": {
+          "Statement": [
+            {
+              "Action": "sts:AssumeRole",
+              "Effect": "Allow",
+              "Principal": {
+                "Service": "glue.amazonaws.com"
```
scanlonp, line 12: Were these glue service principals simply wrong before?

packages/@aws-cdk/aws-pipes-enrichments-alpha/test/__snapshots__/stepfunctions.test.ts.snap
```diff
@@ -10,15 +10,7 @@ exports[`stepfunctions should grant pipe role invoke access 1`] = `
...
-              "Service": {
-                "Fn::FindInMap": [
-                  "ServiceprincipalMap",
-                  {
-                    "Ref": "AWS::Region",
-                  },
-                  "states",
-                ],
-              },
+              "Service": "states.amazonaws.com",
```
scanlonp, line 13: I believe this service principal should have the region in it, i.e. `states.us-west-1.amazonaws.com`. This could be right since the principal is sans region some of the time, but looking through the regional maps we had, a majority of the regions are included in the principal.

scanlonp: Or at least it should not be static if it can be either. The test should be environment agnostic and be a mapping, correct?

scanlonp (changes requested):
How do the principals work for services that conditionally have the region in the principal? Are those still consistent after this change?

packages/aws-cdk-lib/region-info/lib/default.ts
```diff
@@ -1,12 +1,15 @@
...
 export class Default {
 
   /**
    * The default value for a VPC Endpoint Service name prefix, useful if you do
    * not have a synthesize-time region literal available (all you have is
    * `{ "Ref": "AWS::Region" }`)
+   *
+   * @deprecated - Use VpceEndpointService.VPC_ENDPOINT_SERVICE_NAME_PREFIX instead
    */
   public static readonly VPC_ENDPOINT_SERVICE_NAME_PREFIX = 'com.amazonaws.vpce';
```
comcalvi, line 14: You gave it a different name above, it's just `DEFAULT_PREFIX` now.

packages/aws-cdk-lib/region-info/lib/aws-entities.ts
```diff
@@ -78,26 +78,6 @@ export const AWS_REGIONS = AWS_REGIONS_AND_RULES
   .filter((x) => typeof x === 'string')
   .sort() as readonly string[];
 
-/**
- * Possibly non-exhaustive list of all service names, used to locate service principals.
- *
- * Not in the list ==> default service principal mappings.
- */
-export const AWS_SERVICES: readonly string[] = [
```
comcalvi, line 86: yes! Deleting this is a win.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/glue/integ.glue-task.js.snapshot/aws-stepfunctions-integ.template.json
TheRealAmazonKendra: Correct, this policy wasn't actually in use in the test so it wasn't caught but it was invalid altogether.

scripts/check-region-info-compatibility.ts
```diff
@@ -21,6 +21,12 @@ function main(oldPackage: string, newPackage: string) {
   const disappearedFacts = oldFacts
     .filter((oldFact) => !newFacts.some((newFact) => factEq(oldFact, newFact)))
     .map((fact) => ({ fact, key: `${fact[0]}:${fact[1]}` }))
+    // These aren't accessed directly and we've just updated our handling of them,
+    // not removed this functionality. The mapping is unnecessary.
```
comcalvi, line 25: Can you clarify which mapping this is referring to?

packages/aws-cdk-lib/region-info/lib/default.ts
```diff
@@ -1,12 +1,15 @@
 /**
  * Provides default values for certain regional information points.
+ * @deprecated - Service principals are now globally `<SERVICE>.amazonaws.com`
```
comcalvi, line 3: What's the direct replacement class / object / functionality that users should use instead of `Default`?

comcalvi (changes requested):
This is great work, nothing but some minor comments.

packages/@aws-cdk/aws-pipes-enrichments-alpha/test/__snapshots__/stepfunctions.test.ts.snap
TheRealAmazonKendra: The only context in which any region should be included is in opt in regions when the permissions are cross region. This is not that case.

packages/aws-cdk-lib/region-info/lib/default.ts
TheRealAmazonKendra: OOOOOPS. Missed updating this.

scripts/check-region-info-compatibility.ts
TheRealAmazonKendra: We generate it at build time. I'll clarify.

packages/aws-cdk-lib/region-info/lib/default.ts
TheRealAmazonKendra: I forgot to update this. Will change.

packages/aws-cdk-lib/region-info/lib/aws-entities.ts
TheRealAmazonKendra: DELETE ALL THE THINGS

scanlonp (approved):
Good by me

comcalvi (approved):

Issue #30831, opened 2024-07-11
(integ-tests-alpha): Account ID not redacted in integration test


jdukewich:
### Describe the bug

In snapshots from integration tests, you will notice that any account IDs get redacted and replaced with `${AWS::AccountId}`. However, if you try to write an integration test that creates a security group, it will not remove the account ID from the *.assets.json or manifest.json files.

### Expected Behavior

I expected my AWS Account ID to be replaced by `${AWS::AccountId}` in the generated snapshot files.

### Current Behavior

My AWS Account ID is not replaced by `${AWS::AccountId}` in the generated snapshot files.

### Reproduction Steps

1. Create a new integration test `packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.security-group.ts`
2. Make sure your AWS credentials are configured
3. Copy the following code into the test file
```
import * as cdk from 'aws-cdk-lib';
import * as ec2 from 'aws-cdk-lib/aws-ec2';

import { EC2_RESTRICT_DEFAULT_SECURITY_GROUP } from 'aws-cdk-lib/cx-api';
import { IntegTest } from '@aws-cdk/integ-tests-alpha';

const app = new cdk.App();

const env = {
  account: process.env.CDK_INTEG_ACCOUNT || process.env.CDK_DEFAULT_ACCOUNT,
  region: process.env.CDK_INTEG_REGION || process.env.CDK_DEFAULT_REGION,
};

class SgLookupStack extends cdk.Stack {
  constructor(scope: cdk.App, id: string, props?: cdk.StackProps) {
    super(scope, id, props);
    this.node.setContext(EC2_RESTRICT_DEFAULT_SECURITY_GROUP, false);

    const testVpc = new ec2.Vpc(this, 'MyVpc', {
      vpcName: 'my-vpc-name',
      ipAddresses: ec2.IpAddresses.cidr('10.0.0.0/16'),
      subnetConfiguration: [],
      natGateways: 0,
    });
    new ec2.SecurityGroup(this, 'MySgA', { vpc: testVpc, securityGroupName: 'my-sg' });
  }
}

const stack = new SgLookupStack(app, 'StackWithSg', { env });
new IntegTest(app, 'SgLookupTest', {
  testCases: [stack],
});

app.synth();
```
4. Run the test `yarn build && yarn integ --update-on-failed aws-ec2/test/integ.security-group`
[...]

PR #30880, opened 2024-07-17
chore(doc): update INTEGRATION_TESTS.md

pahud:
callout the `env.account` should not be defined in most cases, which might have a security concern discussed in https://github.com/aws/aws-cdk/issues/30831

### Issue # (if applicable)

Closes #<issue number here>.

### Checklist

----

### DIFF
--- a/INTEGRATION_TESTS.md
+++ b/INTEGRATION_TESTS.md
@@ -119,8 +119,7 @@ This will:
 Now when you run `npm test` it will synth the integ app and compare the result with the snapshot.
 If the snapshot has changed the same process must be followed to update the snapshot.
 
-*Note - When running integration tests, refrain from defining the `env.account` property for stacks. This ensures that the synthesized
-CloudFormation template from the snapshots does not contain your current AWS account information, promoting better security and portability.*
+*Note - When running integration tests, `env.account` is generally not required unless some special cases like cross-account tests.*
 
 ### New L2 Constructs
 


INTEGRATION_TESTS.md
```diff
@@ -119,6 +119,9 @@ This will:
 Now when you run `npm test` it will synth the integ app and compare the result with the snapshot.
 If the snapshot has changed the same process must be followed to update the snapshot.
 
+*Note - When running integration tests, refrain from defining the `env.account` property for stacks. This ensures that the synthesized
```
ashishdhingra, line 122: `Note` should be **bold** 😊

comcalvi, line 122: I don't think this is the answer to this issue. I don't think account numbers are inherently considered sensitive information, but even if we decide to them as such, we should write a linter rule that explicitly prevents users from including them in snapshots. 

This might not apply to all tests, some of them need real account numbers (like cross account tests)

comcalvi (changes requested):

INTEGRATION_TESTS.md
pahud: @comcalvi Agree. I have rephrased it this way

*Note - When running integration tests, `env.account` is generally not required unless some special cases like cross-account tests.*

Feel free to suggest changes. I think we still have to remind users that `env.account` is generally not required and not recommended. Thoughts?

pahud:
Have discussed with the team. Closing in favor of a linter tool.

PR #29506, opened 2024-03-15
chore(s3): upgrade S3BucketNotifications Lambda to python3.11

yerzhan7:
### Issue # (if applicable)

Closes D122674750.

### Reason for this change

This Lambda does not work in new regions.

### Description of changes

Current S3BucketNotifications Custom Resource Lambda `python3.9` uses `botocore` version `1.29.90` which does not have S3 endpoint information for new regions.

Upgrading to `python3.11` which has `botocore` version `1.30.1`. See more in D122674750.

Note that we cannot upgrade to `python3.12` because it's not available in all regions yet.

### Description of how you validated changes

N/A.

### Checklist
- [X] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30916, opened 2024-07-22
feat(s3): add skip destination validation property

### LINKED ISSUE #30914: aws-s3: Expose SkipDestinationValidation option when setting bucket notifications
### Describe the feature

S3 API: `PutBucketNotificationConfiguration`

https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketNotificationConfiguration.html

When customers call this API to setup S3 notification configuration for SQS/SNS/Lambda S3 sends `s3:TestEvent` in order to validate permissions. (For Lambda it does `dryrun` function invocation instead)

However, some customers do not want to do that and test permissions during CDK deployment.

1. Internal reference: `49359101-0e5e-43f3-99eb-3c6c5ed68db1`

### Use Case

One customer does not want these test events because they have alarm on unconsumed messages in SQS and they do not have any SQS consumers. And they update notification configuration frequently, which leads to many test events in the queue.

See internal ticket: `P142186522`

### Proposed Solution

Add new Bucket property with this flag.

### Other Information

_No response_

### Acknowledgements

### CDK version used

latest

MacOS

yerzhan7:
### Issue #30914

Closes #30914.

### Reason for this change

When customers call this API to setup S3 notification configuration for SQS/SNS/Lambda S3 sends s3:TestEvent in order to validate permissions. (For Lambda it does dryrun function invocation instead)

However, some customers do not want to do that and test permissions during CDK deployment.

Internal reference: `49359101-0e5e-43f3-99eb-3c6c5ed68db1`

For example, one customer does not want these test events because they have alarm on unconsumed messages in SQS and they do not have any SQS consumers. And they update notification configuration frequently, which leads to many test events in the queue. See internal ticket: `P142186522`

### Description of changes

Expose skip destination validation property when calling PutBucketNotification API in Bucket props.

Unit test updated.

Updated integration tests. Note that 2 integration tests I had to fix and run them with `--disable-update-workflow` flag because they were failing:
 - `integ.s3.imported-bucket.js` test failed because someone already created bucket `cdk-integration-test-s3-imported-bucket-name`
 - `integ.bucket-notifications.js` test failed because of overlapping suffix error (not sure how it was passing previously):
 
 ```
❌  cdk-integ-lambda-bucket-s3-notifications failed: Error: The stack named cdk-integ-lambda-bucket-s3-notifications failed to deploy: UPDATE_FAILED 
(The following resource(s) failed to update: [Construct1IntegUnmanagedBucket1Notifications4A1599D7]. ): Received response status [FAILED] from custom resource. 
Message returned: Error: An error occurred (InvalidArgument) when calling the PutBucketNotificationConfiguration operation: Configuration is ambiguously defined. 
Cannot have overlapping suffixes in two rules if the prefixes are overlapping for the same event type.. 
See the details in CloudWatch Log Stream: 2024/07/22/[$LATEST]e6a16cf979dd4671998e7d911769ff42 (RequestId: 19f6fcd7-d31d-4fbf-9f4a-e3b7cba1cd2b), 
Rolling back the failed resource only., Received response status [FAILED] from custom resource. Message returned: Error: An error occurred 
(InvalidArgument) when calling the PutBucketNotificationConfiguration operation: Configuration is ambiguously defined. 
Cannot have overlapping suffixes in two rules if the prefixes are overlapping for the same event type.. 
See the details in CloudWatch Log Stream: 2024/07/22/[$LATEST]c30efd0375d64b8088e0ee64d63ce4db (RequestId: 19f6fcd7-d31d-4fbf-9f4a-e3b7cba1cd2b)
```

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-applicationautoscaling/lib/target-tracking-scaling-policy.ts
+++ b/packages/aws-cdk-lib/aws-applicationautoscaling/lib/target-tracking-scaling-policy.ts
@@ -304,4 +304,19 @@ export enum PredefinedMetric {
    * @see https://docs.aws.amazon.com/autoscaling/application/APIReference/API_PredefinedMetricSpecification.html
    */
   ELASTICACHE_DATABASE_CAPACITY_USAGE_COUNTED_FOR_EVICT_PERCENTAGE = 'ElastiCacheDatabaseCapacityUsageCountedForEvictPercentage',
+  /**
+   * SAGEMAKER_INFERENCE_COMPONENT_CONCURRENT_REQUESTS_PER_COPY_HIGH_RESOLUTION
+   * @see https://docs.aws.amazon.com/autoscaling/application/APIReference/API_PredefinedMetricSpecification.html
+   */
+  SAGEMAKER_INFERENCE_COMPONENT_CONCURRENT_REQUESTS_PER_COPY_HIGH_RESOLUTION = 'SageMakerInferenceComponentConcurrentRequestsPerCopyHighResolution',
+  /**
+   * SAGEMAKER_VARIANT_CONCURRENT_REQUESTS_PER_MODEL_HIGH_RESOLUTION
+    * @see https://docs.aws.amazon.com/autoscaling/application/APIReference/API_PredefinedMetricSpecification.html
+    */
+  SAGEMAKER_VARIANT_CONCURRENT_REQUESTS_PER_MODEL_HIGH_RESOLUTION = 'SageMakerVariantConcurrentRequestsPerModelHighResolution',
+  /**
+   * WORKSPACES_AVERAGE_USER_SESSIONS_CAPACITY_UTILIZATION
+   * @see https://docs.aws.amazon.com/autoscaling/application/APIReference/API_PredefinedMetricSpecification.html
+   */
+  WORKSPACES_AVERAGE_USER_SESSIONS_CAPACITY_UTILIZATION = 'WorkSpacesAverageUserSessionsCapacityUtilization',
 }
--- a/packages/aws-cdk-lib/aws-applicationautoscaling/test/target-tracking.test.ts
+++ b/packages/aws-cdk-lib/aws-applicationautoscaling/test/target-tracking.test.ts
@@ -91,6 +91,66 @@ describe('target tracking', () => {
     });
   });
 
+  test('test setup target tracking on predefined metric for SAGEMAKER_VARIANT_CONCURRENT_REQUESTS_PER_MODEL_HIGH_RESOLUTION', () => {
+    // GIVEN
+    const stack = new cdk.Stack();
+    const target = createScalableTarget(stack);
+
+    // WHEN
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: packages/@aws-cdk/custom-resource-handlers/test/aws-s3/notifications-resource-handler/Dockerfile [hunk 1]
```diff
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/lambda/python:3.7
+FROM public.ecr.aws/lambda/python:3.11
```
gracelu0, line 1: I see integration tests were updated to use python3.11 previously here: https://github.com/aws/aws-cdk/pull/29506

For my understanding is this dockerfile is for running the integration tests? Also curious how there were no issues if it was using python3.7 before

gracelu0: I will double check with the team but I believe we will also need to update the runtime in the  custom resource config file here: https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/custom-resource-handlers/lib/custom-resources-framework/config.ts#L265

gracelu0 (dismissed):
Hi @yerzhan7 , thanks for adding support for this feature! I have some questions about the runtime update before approving but most of the changes LGTM.

packages/@aws-cdk/custom-resource-handlers/test/aws-s3/notifications-resource-handler/Dockerfile
yerzhan7: > For my understanding is this dockerfile is for running the integration tests?

This is for unit tests only AFAIK.

> Also curious how there were no issues if it was using python3.7 before

Python 3.7 and 3.11 are pretty much the same in this case. I am updating it in this PR in order to make unit test Lambda runtime to be the same version as in actual notification construct/Lambda.

Let me know if I need to update anything else. (Also, I am happy to revert this change if you think it's not needed in this PR)

Thanks for reviewing!

gracelu0: Got it, thanks for the explanation. For the sake of parity (and to prevent future confusion), could you update the runtime in the custom-resources-framework `config.ts` file please? (will also need to add Python 3.11 enum [here](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/custom-resource-handlers/lib/custom-resources-framework/config.ts#L7-L21))

yerzhan7: Done, thanks.

gracelu0 (approved):

gracelu0 (approved):

gracelu0:
@Mergifyio update

gracelu0 (approved):

PR #30920, opened 2024-07-22
feat(ecs-patterns): add `containerCpu` and `containerMemoryLimitMiB` property to `ApplicationLoadBalancedFargateService`

### LINKED ISSUE #20638: (aws_ecs_patterns): container-level cpu & memory props for ApplicationLoadBalancedFargateService
### Describe the feature

Surface the container-level `cpu` and container-level `memoryLimitMiB` property of `ContainerDefinitionOptions` as part of the `ApplicationLoadBalancedFargateService` construct.

Currently those two properties are not accessible when creating an `ApplicationLoadBalancedFargateService`. (see [code](https://github.com/aws/aws-cdk/blob/v2-release/packages/%40aws-cdk/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts#L145))

### Use Case

I want to deploy a sidecar container to the service container, within the same task definition. Right now the issue I observe is, the service container is set a cpu value that equals to the task cpu value (named `cpu` in ApplicationLoadBalancedFargateService construct prop), thus I get error because the sum of sidecar container cpu and service container cpu exceeds the task cpu limit.

### Proposed Solution

```
const service = new ApplicationLoadBalancedFargateService(this, 'Service',  {
    cpu: 4096 ,
    memoryLimitMiB: 7168, 
    containerCpu: 2048,
    containerMemoryLimitMiB: 6144,
    ...
}
```
or maybe passing in an object
```
const service = new ApplicationLoadBalancedFargateService(this, 'Service',  {
    cpu: 4096 ,
    memoryLimitMiB: 7168, 
    containerDefinitionOptions : {
        containerCpu: 2048,
        containerMemoryLimitMiB: 6144,
        ...
    }
    ...
}
```

### Other Information

_No response_

### Acknowledgements

### CDK version used

2

Linux

ren-yamanashi:
### Issue #20638 

Closes #20638 

### Reason for this change

ApplicationLoadBalancedFargateService did not allow you to specify the CPU and memory of the container.

### Description of changes

- Add `containerCpu` property to `ApplicationLoadBalancedFargateServiceProps`
- Add `containerMemoryLimitMiB` property to `ApplicationLoadBalancedFargateServiceProps`

Added both unit and integration tests.

### Checklist

----


go-to-k (changes requested):
Could you change the PR title from "feat(application-load-balanced-fargate-service): ..." to "feat(ecs-patterns): add `containerCpu` and `containerMemoryLimitMiB` property to `ApplicationLoadBalancedFargateService`"? Because we should specify the module name, not the construct name.

ren-yamanashi:
> Could you change the PR title from "feat(application-load-balanced-fargate-service): ..." to "feat(ecs-patterns): add `containerCpu` and `containerMemoryLimitMiB` property to `ApplicationLoadBalancedFargateService`"? Because we should specify the module name, not the construct name.

Thank you for the suggestion. I have updated the PR title. Please check the changes at your convenience.

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
```diff
@@ -2066,4 +2066,32 @@ describe('NetworkLoadBalancedFargateService', () => {
       },
     });
   });
+
+  test('specify containerCpu And containerMemoryLimitMiB', () => {
```
go-to-k, line 2070: It seems that this test is included in the `NetworkLoadBalancedFargateService` test, so it needs to be moved somewhere within the `ApplicationLoadBalancedFargateService` test.

https://github.com/go-to-k/aws-cdk/blob/ffd9d9ce510a4c820b1437cce93c4772cd7c14c0/packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts#L15

and nit.

```suggestion
  test('specify containerCpu and containerMemoryLimitMiB', () => {
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -36,6 +36,23 @@ export interface ApplicationLoadBalancedFargateServiceProps extends ApplicationL
...
+
+  /**
+   * The amount (in MiB) of memory to present to the container.
+   *
+   * If your container attempts to exceed the allocated memory, the container
+   * is terminated.
+   *
+   * @default - No memory limit.
+   */
+  readonly containerMemoryLimitMiB?: number;
```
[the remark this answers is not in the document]
go-to-k: I am not sure, but I think the current implementation is good. Because putting it in `FargateServiceBaseProps` would be risky because it would affect all constructs that use all props that implement it.

Also, a property such as `healthCheck`, for example, could be included as a common item in the Base Construct, but right now it is only in the `ApplicationLoadedBalancedFargateServiceProps` (and `QueueProcessingFargateServiceProps`). From this point of view, it would make sense to include them only in the props.

(However, to be honest, I also think that cpu and memory are common properties for containers. But if it became necessary, it could be modified later.)

packages/aws-cdk-lib/aws-ecs-patterns/README.md
```diff
@@ -70,6 +70,10 @@ By setting `redirectHTTP` to true, CDK will automatically create a listener on p
 
 If you specify the option `recordType` you can decide if you want the construct to use CNAME or Route53-Aliases as record sets.
 
+If you specify the option `containerCpu` allows you to set the minimum number of CPU units to reserve for the container.
+
+If you specify the option `containerMemoryLimitMiB` allows you to set the amount of memory (in MiB) to provide to the container.
+
```
go-to-k, line 76: Could you add the properties to the sample code above?

```ts
declare const cluster: ecs.Cluster;
const loadBalancedFargateService = new ecsPatterns.ApplicationLoadBalancedFargateService(this, 'Service', {
  cluster,
  memoryLimitMiB: 1024,
  cpu: 512,
  taskImageOptions: {
    // ...
```

packages/aws-cdk-lib/aws-ecs-patterns/README.md
```diff
@@ -70,6 +70,10 @@ By setting `redirectHTTP` to true, CDK will automatically create a listener on p
 
 If you specify the option `recordType` you can decide if you want the construct to use CNAME or Route53-Aliases as record sets.
 
+If you specify the option `containerCpu` allows you to set the minimum number of CPU units to reserve for the container.
+
+If you specify the option `containerMemoryLimitMiB` allows you to set the amount of memory (in MiB) to provide to the container.
```
go-to-k, line 75: Made some changes:

```suggestion
To set the minimum number of CPU units to reserve for the container, you can use the `containerCpu` property.

To set the amount of memory (in MiB) to provide to the container, you can use the `containerMemoryLimitMiB` property.
```

go-to-k (changes requested):
Thanks for this PR! I made some comments.

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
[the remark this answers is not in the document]
ren-yamanashi: Understood, thank you for the detailed explanation!

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/59b2aa1fe39180ed22d597c26e07c881f1c6f610

packages/aws-cdk-lib/aws-ecs-patterns/README.md
ren-yamanashi: For no apparent reason, I set the value of each property to half the value of 'cpu','memoryLimitMiB'. Is there a problem?

https://github.com/aws/aws-cdk/pull/30920/commits/3f2f3daf007ec28077913a9ca3f76b20b9b09480

ren-yamanashi:
@go-to-k 

Thanks for the review! We have fixed it, so please check it out!

(A link to the fix commit is provided in reply to the thread.)

packages/aws-cdk-lib/aws-ecs-patterns/README.md
go-to-k: No problem.

go-to-k (changes requested):
If the `containerCpu` and the `containerMemoryLimitMiB` are larger than (or equal to) the `cpu` and the `memoryLimitMiB`, will CloudFormation deploy errors occur?
If so, it is good to add validations for them. (and also good to add corresponding unit tests)

for example:
```ts
      if (
        props.containerCpu &&
        !Token.isUnresolved(props.containerCpu) &&
        props.cpu &&
        !Token.isUnresolved(props.cpu) &&
        props.containerCpu > props.cpu // or `props.containerCpu >= props.cpu`
      ) {
        throw new Error('containerCpu must be less than or equal to cpu');
      }
      // The same applies to memory.
      // ...
```

(It may be possible to check for values like this (the `validateHealthyPercentage` method) that are not negative.)

see: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts#L108-L119

ren-yamanashi:
> If the `containerCpu` and the `containerMemoryLimitMiB` are larger than (or equal to) the `cpu` and the `memoryLimitMiB`, will CloudFormation deploy errors occur? If so, it is good to add validations for them. (and also good to add corresponding unit tests)
> 
> for example:
> 
> ```ts
>       if (
>         props.containerCpu &&
>         !Token.isUnresolved(props.containerCpu) &&
>         props.cpu &&
>         !Token.isUnresolved(props.cpu) &&
>         props.containerCpu > props.cpu // or `props.containerCpu >= props.cpu`
>       ) {
>         throw new Error('containerCpu must be less than or equal to cpu');
>       }
>       // The same applies to memory.
>       // ...
> ```
> 
> see: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts#L108-L119
> 
> As with the `validateHealthyPercentage` method, it may also be possible to check for values that are not negative. (If CFn fails for negative values.)

@go-to-k 

Thank you for your comment!

We have confirmed that the deployment fails in the following cases, so we have added validation.

- If containerCpu is greater than cpu
- If containerCpu is not positive integer
- If containerMemoryLimitMiB is greater than memoryLimitMiB
- if containerMemoryLimitMiB is not positive integer

Please Check it out 🙇

ren-yamanashi:
> If the `containerCpu` and the `containerMemoryLimitMiB` are larger than (or equal to) the `cpu` and the `memoryLimitMiB`, will CloudFormation deploy errors occur? If so, it is good to add validations for them. (and also good to add corresponding unit tests)
> 
> for example:
> 
> ```ts
>       if (
>         props.containerCpu &&
>         !Token.isUnresolved(props.containerCpu) &&
>         props.cpu &&
>         !Token.isUnresolved(props.cpu) &&
>         props.containerCpu > props.cpu // or `props.containerCpu >= props.cpu`
>       ) {
>         throw new Error('containerCpu must be less than or equal to cpu');
>       }
>       // The same applies to memory.
>       // ...
> ```
> 
> see: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts#L108-L119
> 
> As with the `validateHealthyPercentage` method, it may also be possible to check for values that are not negative. (If CFn fails for negative values.)

@go-to-k

Thank you for your comment!

We have confirmed that the deployment fails in the following cases, so we have add validation.

- If containerCpu is greater than cpu
- If containerCpu is not positive integer
- If containerMemoryLimitMiB is greater than memoryLimitMiB
- if containerMemoryLimitMiB is not positive integer

Please confirm.

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -70,6 +87,12 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
     } else if (props.taskDefinition) {
       this.taskDefinition = props.taskDefinition;
     } else if (props.taskImageOptions) {
+      this.validateHealthyPercentage('containerCpu', props.containerCpu);
+      this.validateHealthyPercentage('containerMemoryLimitMiB', props.containerMemoryLimitMiB);
```
go-to-k, line 91: The method `validateHealthyPercentage` is for healthy percentage, so I don't think it is good to use it here. Actually the method can be used here because the validation process is same, but could you add another methods for each cpu and memory?

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -70,6 +87,12 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
...
+
+      this.validateNotExceeding('containerCpu', props.cpu ?? 256, props.containerCpu);
+      this.validateNotExceeding('containerMemoryLimitMiB', props.memoryLimitMiB ?? 512, props.containerMemoryLimitMiB);
```
go-to-k, line 94: Why do you specify 256 for cpu and 512 for memory?

It is true that the minimum values for cpu and memory are those values, but it is a bit confusing to pass those values when props.cpu and props.memory are not specified. If you want to validate the minimum value, it would be good to implement it more directly.

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +177,14 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
...
+
+  /**
+   * Throws an error if the specified value is greater than the limit.
+   */
+  private validateNotExceeding(name: string, limit: number, value?: number) {
+    if (value === undefined || Token.isUnresolved(value)) { return; }
+    if (value > limit) {
+      throw new Error(`${name}: Must be less than or equal to ${limit}; received ${value}`);
+    }
+  }
```
go-to-k, line 189: Since this is abstract (I don't think it is good to use it for properties other than cpu and memory), it would be better to make the methods specific to cpu and memory, as described above.

Also, if specifying containerCpu without cpu, what will happen in Cfn deploy?

go-to-k (changes requested):

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: It specified default values ​​for props.cpu and props.memoryLimitMiB...

https://github.com/aws/aws-cdk/blob/3670d83f40d22b72d883515b14c819ceb69f81df/packages/aws-cdk-lib/aws-ecs-patterns/lib/base/fargate-service-base.ts#L34

https://github.com/aws/aws-cdk/blob/3670d83f40d22b72d883515b14c819ceb69f81df/packages/aws-cdk-lib/aws-ecs-patterns/lib/base/fargate-service-base.ts#L60


In the modified code, 'JsDoc' indicates that it is the default value, as shown below, but I feel that it is difficult to understand.

What is the best way to implement it?

```ts
  /**
   * Throws an error if containerCpu is greater than cpu or is a not positive integer.
   * @param containerCpu The minimum number of CPU units to reserve for the container.
   * @param cpu The number of cpu units used by the task. (default: 256)
   */
  private validateContainerCpu(containerCpu?: number, cpu: number = 256) {
      // ...
 }

  /**
   * Throws an error if containerMemoryLimitMiB is greater than memoryLimitMiB or is a not positive integer.
   * @param containerMemoryLimitMiB The amount (in MiB) of memory to present to the container.
   * @param memoryLimitMiB The amount (in MiB) of memory used by the task. (default: 512)
   */
  private validateContainerMemoryLimitMiB(containerMemoryLimitMiB?: number, memoryLimitMiB: number = 512) {
    // ...
}
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: Thank you for pointing this out. We will correct it.

ren-yamanashi:
@go-to-k 

We have responded to comments and fixed the implementation, so please confirm!

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -70,6 +87,9 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
     } else if (props.taskDefinition) {
       this.taskDefinition = props.taskDefinition;
     } else if (props.taskImageOptions) {
+      this.validateContainerCpu(props.containerCpu, props.cpu);
+      this.validateContainerMemoryLimitMiB(props.containerMemoryLimitMiB, props.memoryLimitMiB);
```
go-to-k, line 91: Could you move them to the next?

```ts
      const enableLogging = taskImageOptions.enableLogging ?? true;
      const logDriver = taskImageOptions.logDriver !== undefined
        ? taskImageOptions.logDriver : enableLogging
          ? this.createAWSLogDriver(this.node.id) : undefined;

      // here
      this.validateContainerCpu(props.containerCpu, props.cpu);
      this.validateContainerMemoryLimitMiB(props.containerMemoryLimitMiB, props.memoryLimitMiB);

      const containerName = taskImageOptions.containerName ?? 'web';
      const container = this.taskDefinition.addContainer(containerName, {
        // ...
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +174,38 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
       throw new Error(`${name}: Must be a non-negative integer; received ${value}`);
     }
   }
+
+  /**
+   * Throws an error if containerCpu is greater than cpu or is a not positive integer.
+   * @param containerCpu The minimum number of CPU units to reserve for the container.
+   * @param cpu The number of cpu units used by the task. (default: 256)
+   */
```
go-to-k, line 182: Thanks, but JSDoc is not required for private methods.

```suggestion
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +174,38 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
...
+    if (containerCpu < 0 || !Number.isInteger(containerCpu)) {
+      throw new Error('containerCpu must be positive integer');
+    }
+  }
+
+  /**
+   * Throws an error if containerMemoryLimitMiB is greater than memoryLimitMiB or is a not positive integer.
+   * @param containerMemoryLimitMiB The amount (in MiB) of memory to present to the container.
+   * @param memoryLimitMiB The amount (in MiB) of memory used by the task. (default: 512)
+   */
```
go-to-k, line 199: same
```suggestion
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +174,38 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
...
+  private validateContainerCpu(containerCpu?: number, cpu: number = 256) {
+    if (containerCpu === undefined || Token.isUnresolved(containerCpu) || Token.isUnresolved(cpu)) {
+      return;
+    }
+    if (containerCpu > cpu) {
+      throw new Error('containerCpu must be less than to cpu');
+    }
+    if (containerCpu < 0 || !Number.isInteger(containerCpu)) {
+      throw new Error('containerCpu must be positive integer');
```
go-to-k, line 191:
```suggestion
      throw new Error(`containerCpu must be a non-negative integer; received ${containerCpu}`);
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +174,38 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
...
+   */
+  private validateContainerMemoryLimitMiB(containerMemoryLimitMiB?: number, memoryLimitMiB: number = 512) {
+    if (containerMemoryLimitMiB === undefined || Token.isUnresolved(containerMemoryLimitMiB) || Token.isUnresolved(memoryLimitMiB)) {
+      return;
+    }
+    if (containerMemoryLimitMiB > memoryLimitMiB) {
+      throw new Error('containerMemoryLimitMiB must be less than to memoryLimitMiB');
+    }
+    if (containerMemoryLimitMiB < 0 || !Number.isInteger(containerMemoryLimitMiB)) {
+      throw new Error('containerMemoryLimitMiB must be positive integer');
```
go-to-k, line 208:
```suggestion
      throw new Error(`containerMemoryLimitMiB must be a non-negative integer; received ${containerMemoryLimitMiB}`);
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
go-to-k: If you specify 0 for containerCpu, will CFn get an error? Depending on that, the condition of the if statement and this error message should be changed.

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
go-to-k: When containerCpu is specified to the same value as cpu, does CFn not cause an error?

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +174,38 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
...
+   * Throws an error if containerCpu is greater than cpu or is a not positive integer.
+   * @param containerCpu The minimum number of CPU units to reserve for the container.
+   * @param cpu The number of cpu units used by the task. (default: 256)
+   */
+  private validateContainerCpu(containerCpu?: number, cpu: number = 256) {
+    if (containerCpu === undefined || Token.isUnresolved(containerCpu) || Token.isUnresolved(cpu)) {
+      return;
+    }
+    if (containerCpu > cpu) {
+      throw new Error('containerCpu must be less than to cpu');
```
go-to-k, line 188: It would be user friendly to display the value specified.

```suggestion
      throw new Error(`containerCpu must be less than to cpu; received containerCpu: ${containerCpu}, cpu: ${cpu}`);
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +174,38 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
...
+   * Throws an error if containerMemoryLimitMiB is greater than memoryLimitMiB or is a not positive integer.
+   * @param containerMemoryLimitMiB The amount (in MiB) of memory to present to the container.
+   * @param memoryLimitMiB The amount (in MiB) of memory used by the task. (default: 512)
+   */
+  private validateContainerMemoryLimitMiB(containerMemoryLimitMiB?: number, memoryLimitMiB: number = 512) {
+    if (containerMemoryLimitMiB === undefined || Token.isUnresolved(containerMemoryLimitMiB) || Token.isUnresolved(memoryLimitMiB)) {
+      return;
+    }
+    if (containerMemoryLimitMiB > memoryLimitMiB) {
+      throw new Error('containerMemoryLimitMiB must be less than to memoryLimitMiB');
```
go-to-k, line 205:
```suggestion
      throw new Error(`containerMemoryLimitMiB must be less than to memoryLimitMiB; received containerMemoryLimitMiB: ${containerMemoryLimitMiB}, memoryLimitMiB: ${memoryLimitMiB}`);
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +174,38 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
       throw new Error(`${name}: Must be a non-negative integer; received ${value}`);
     }
   }
+
+  /**
+   * Throws an error if containerCpu is greater than cpu or is a not positive integer.
+   * @param containerCpu The minimum number of CPU units to reserve for the container.
+   * @param cpu The number of cpu units used by the task. (default: 256)
+   */
+  private validateContainerCpu(containerCpu?: number, cpu: number = 256) {
```
go-to-k, line 183:
```suggestion
  private validateContainerCpu(containerCpu?: number, cpu: number = 256) { // default value for cpu is 256
```

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +174,38 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
...
+      throw new Error('containerCpu must be positive integer');
+    }
+  }
+
+  /**
+   * Throws an error if containerMemoryLimitMiB is greater than memoryLimitMiB or is a not positive integer.
+   * @param containerMemoryLimitMiB The amount (in MiB) of memory to present to the container.
+   * @param memoryLimitMiB The amount (in MiB) of memory used by the task. (default: 512)
+   */
+  private validateContainerMemoryLimitMiB(containerMemoryLimitMiB?: number, memoryLimitMiB: number = 512) {
```
go-to-k, line 200:
```suggestion
  private validateContainerMemoryLimitMiB(containerMemoryLimitMiB?: number, memoryLimitMiB: number = 512) { // default value for memoryLimitMiB is 512
```

go-to-k (changes requested):
Thanks for changes. I made some comments. It's getting better!

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
go-to-k: Thanks for clarifying. I commented at another thread, please take a look at it.

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
```diff
@@ -1433,6 +1433,158 @@ describe('ApplicationLoadBalancedFargateService', () => {
...
+      ContainerDefinitions: [
+        Match.objectLike({
+          Cpu: 128,
+          Memory: 256,
+        }),
+      ],
+    });
+  });
+
+  test('errors when containerCpu is greater than cpu', () => {
```
go-to-k, line 1465:
```suggestion
  test('throw when containerCpu is greater than cpu', () => {
```

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
```diff
@@ -1433,6 +1433,158 @@ describe('ApplicationLoadBalancedFargateService', () => {
...
+        },
+        cpu: 256,
+        loadBalancerName: 'alb-test-load-balancer',
+        containerCpu: 512,
+      });
+    }).toThrow('containerCpu must be less than to cpu');
+  });
+
+  describe('errors when containerCpu is not positive integer', () => {
+    test('when containerCpu is negative integer', () => {
```
go-to-k, line 1486:
```suggestion
  test('throw when containerCpu is negative integer', () => {
```

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
```diff
@@ -1433,6 +1433,158 @@ describe('ApplicationLoadBalancedFargateService', () => {
...
+            image: ecs.ContainerImage.fromRegistry('test'),
+          },
+          cpu: 256,
+          loadBalancerName: 'alb-test-load-balancer',
+          containerCpu: -1,
+        });
+      }).toThrow('containerCpu must be positive integer');
+    });
+
+    test('when containerCpu is float', () => {
```
go-to-k, line 1506:
```suggestion
  test('throw when containerCpu is float', () => {
```

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
```diff
@@ -1433,6 +1433,158 @@ describe('ApplicationLoadBalancedFargateService', () => {
...
+          },
+          cpu: 256,
+          loadBalancerName: 'alb-test-load-balancer',
+          containerCpu: 0.5,
+        });
+      }).toThrow('containerCpu must be positive integer');
+    });
+  });
+
+  test('errors when containerMemoryLimitMiB is greater than memoryLimitMiB', () => {
```
go-to-k, line 1527:
```suggestion
  test('throw when containerMemoryLimitMiB is greater than memoryLimitMiB', () => {
```

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
```diff
@@ -1433,6 +1433,158 @@ describe('ApplicationLoadBalancedFargateService', () => {
...
+        },
+        memoryLimitMiB: 256,
+        loadBalancerName: 'alb-test-load-balancer',
+        containerMemoryLimitMiB: 512,
+      });
+    }).toThrow('containerMemoryLimitMiB must be less than to memoryLimitMiB');
+  });
+
+  describe('errors when containerMemoryLimitMiB is not positive integer', () => {
+    test('when containerMemoryLimitMiB is negative integer', () => {
```
go-to-k, line 1548:
```suggestion
  test('throw when containerMemoryLimitMiB is negative integer', () => {
```

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
```diff
@@ -1433,6 +1433,158 @@ describe('ApplicationLoadBalancedFargateService', () => {
...
+            image: ecs.ContainerImage.fromRegistry('test'),
+          },
+          memoryLimitMiB: 256,
+          loadBalancerName: 'alb-test-load-balancer',
+          containerMemoryLimitMiB: -1,
+        });
+      }).toThrow('containerMemoryLimitMiB must be positive integer');
+    });
+
+    test('when containerMemoryLimitMiB is float', () => {
```
go-to-k, line 1568:
```suggestion
    test('throw when containerMemoryLimitMiB is float', () => {
```

go-to-k (changes requested):

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/1c5d8cf272b6a385dd7edd1e57bc51a768d0cedb

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/707a331261163204318ca833786c140fda4f77a1

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/707a331261163204318ca833786c140fda4f77a1

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/2b7a189843299d8a4386c906cd8edf5f5c779775

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/2b7a189843299d8a4386c906cd8edf5f5c779775

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/2b7a189843299d8a4386c906cd8edf5f5c779775

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: I changed the error message in the commit below!

https://github.com/aws/aws-cdk/commit/2b7a189843299d8a4386c906cd8edf5f5c779775

> If you specify 0 for containerCpu, will CFn get an error? Depending on that, the condition of the if statement and this error message should be changed.

I tried with the pattern below, but CFn didn't give me an error.

・When containerCPU is 0
・When containerMemoryLimitMiB is 0

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/f2243912fca179e576d1fb5d11274c9efcd17046

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/f2243912fca179e576d1fb5d11274c9efcd17046

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/d60a82c81569c47f43e975fb28a5de40c5a7d4f8

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/d60a82c81569c47f43e975fb28a5de40c5a7d4f8

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/d60a82c81569c47f43e975fb28a5de40c5a7d4f8

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/d60a82c81569c47f43e975fb28a5de40c5a7d4f8

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/d60a82c81569c47f43e975fb28a5de40c5a7d4f8

packages/aws-cdk-lib/aws-ecs-patterns/test/fargate/load-balanced-fargate-service.test.ts
ren-yamanashi: https://github.com/aws/aws-cdk/pull/30920/commits/d60a82c81569c47f43e975fb28a5de40c5a7d4f8

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: Yes, I didn't get an error in this case.
(The same was true for memory)

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +174,28 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
...
+  private validateContainerMemoryLimitMiB(containerMemoryLimitMiB?: number, memoryLimitMiB: number = 512) { // default value for memoryLimitMiB is 512
+    if (containerMemoryLimitMiB === undefined || Token.isUnresolved(containerMemoryLimitMiB) || Token.isUnresolved(memoryLimitMiB)) {
+      return;
+    }
+    if (containerMemoryLimitMiB > memoryLimitMiB) {
+      throw new Error(`containerMemoryLimitMiB must be less than to memoryLimitMiB; received containerMemoryLimitMiB: ${containerMemoryLimitMiB}, memoryLimitMiB: ${memoryLimitMiB}`);
+    }
+    if (containerMemoryLimitMiB <= 0 || !Number.isInteger(containerMemoryLimitMiB)) {
+      throw new Error(`containerMemoryLimitMiB must be a positive integer; received ${containerMemoryLimitMiB}`);
+    }
```
ren-yamanashi, line 199: containerMemoryLimitMiB must be an integer greater than 0, so the error message specifies "Must be a positive integer"

ren-yamanashi (commented):
@go-to-k 

Thanks for the review! I fixed code, so please confirm.

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
go-to-k: Yeah, that's good.

go-to-k (commented):
You can resolve threads that have been resolved, as the number of threads has increased (I don't have the permission to do that).

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: Thank you. I Added comment

https://github.com/aws/aws-cdk/pull/30920/commits/89290c648a2234479802e1218b6dd2e47f7b1b33

ren-yamanashi:
> You can resolve threads that have been resolved, as the number of threads has increased (I don't have the permission to do that).

@go-to-k 

I resolved threads that have been resolved!

go-to-k (approved):
Thanks, LGTM.

go-to-k:
It should have a `pr/needs-maintainer-review` label, but apparently it doesn't. Could you try to pull diff from the main branch?

ren-yamanashi:
@go-to-k 

When I pulled the main branch, CI build failed...

I will continue to comment as soon as CI build is successed

go-to-k:
This would have nothing to do with this PR. However, it would be curious to know if this is just a coincidence or if it happens with other PRs. It might be good to wait and see a bit more.

ren-yamanashi:
> This would have nothing to do with this PR. However, it would be curious to know if this is just a coincidence or if it happens with other PRs. It might be good to wait and see a bit more.

I pulled the main branch again and CI went passed. Thanks you!

go-to-k (changes requested):
This is a dummy comment for labeling tests.

go-to-k (approved):

go-to-k (changes requested):
> I approved it, so it should have a pr/needs-maintainer-review label, but apparently it doesn't. Could you try to pull diffs from the main branch?

[The result](https://github.com/aws/aws-cdk/actions/runs/10625116391/job/29454561862?pr=30920) in the [PR Linter / validate-pr](https://github.com/aws/aws-cdk/tree/v2.155.0/.github/workflows/pr-linter.yml#L80) action is apparently that CodeBuild is not successful. (see: https://github.com/go-to-k/aws-cdk/blob/a45230ff6c2115c33d9a2409997ef0e708b97f70/tools/@aws-cdk/prlint/lint.ts#L587)

```
Run ./tools/@aws-cdk/prlint
  env:
    GITHUB_TOKEN: ***
    PR_NUMBER: 
    PR_SHA: 
    REPO_ROOT: /home/runner/work/aws-cdk/aws-cdk
⌛  Fetching PR number 309[2](https://github.com/aws/aws-cdk/actions/runs/10625116391/job/29454561862?pr=30920#step:4:2)0
⌛  Fetching files for PR number [3](https://github.com/aws/aws-cdk/actions/runs/10625116391/job/29454561862?pr=30920#step:4:3)0920
⌛  Validating...
Deleting PR Linter Comment now
✅  Success
CodeBuild Commit Statuses:  []
PR code build job not yet successful
```

Could you try making some changes to the description of this PR first? (This is because the job will be run again [if the PR is edited](https://github.com/aws/aws-cdk/tree/v2.155.0/.github/workflows/pr-linter.yml#L10).)

If there are still no changes in labels, could you try running CodeBuild CI again?

ren-yamanashi:
@go-to-k 

PR Linter / validate-pr succeeded!

```sh
Run ./tools/@aws-cdk/prlint
  env:
    GITHUB_TOKEN: ***
    PR_NUMBER: 
    PR_SHA: 
    REPO_ROOT: /home/runner/work/aws-cdk/aws-cdk
⌛  Fetching PR number 309[2](https://github.com/aws/aws-cdk/actions/runs/10629878433/job/29467658164?pr=30920#step:4:2)0
⌛  Fetching files for PR number [3](https://github.com/aws/aws-cdk/actions/runs/10629878433/job/29467658164?pr=30920#step:4:3)0920
⌛  Validating...
Deleting PR Linter Comment now
✅  Success
CodeBuild Commit Statuses:  [ 'success', 'pending' ]
PR code build job SUCCESSFUL
```

Regarding the label change, it seems that it has not been approved, so could you please approve?

If the label is still not granted, I will try runnning PR Linter or CodeBuild CI again.

ren-yamanashi:
@go-to-k 

PR Linter / validate-pr succeeded!

```sh
Run ./tools/@aws-cdk/prlint
  env:
    GITHUB_TOKEN: ***
    PR_NUMBER: 
    PR_SHA: 
    REPO_ROOT: /home/runner/work/aws-cdk/aws-cdk
⌛  Fetching PR number 30920
⌛  Fetching files for PR number 30920
⌛  Validating...
Deleting PR Linter Comment now
✅  Success
CodeBuild Commit Statuses:  [ 'success', 'pending' ]
PR code build job SUCCESSFUL
```

Regarding the label change, it seems that it has not been approved, so could you please approve?

スクリーンショット 2024-08-30 18 13 26


If the label is still not granted, I will try runnning PR Linter or CodeBuild CI again.

go-to-k (approved):

go-to-k:
I approved, but the PR doesn't have the label.

Could you run CI again?

go-to-k:
Could you try making some changes to the description of this PR without running CodeBuild CI?

ren-yamanashi:
@go-to-k 

I ran PR Linter and CodeBuild again, but the PR doesn`t have the label...
Looking at other PRs, it was labeled successfully, but is there a problem with what I submitted?

go-to-k (approved):

ren-yamanashi:
@go-to-k

I ran PR Linter without runnning CodeBuild again, but the PR doesn`t have the label...
Looking at other PRs, it was labeled successfully, but is there a problem with what I submitted?

go-to-k:
It appears that the status is CHANGES_REQUESTED even though it has been approved: https://github.com/aws/aws-cdk/actions/runs/10669155243/job/29570426536

I will consult this with the CDK team.

ren-yamanashi:
Thank you very much!

go-to-k (approved):

go-to-k (approved):

go-to-k:
@ren-yamanashi 

Hi, the fix PR has been merged!

https://github.com/aws/aws-cdk/pull/31290

I approved it again and the review label was successfully attached.

ren-yamanashi:
Thank you for your quick response!

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
shikha372, line 55: To understand it better for review, seeing the original issue i think there was a suggestion that we could use task definition of fargate to spcify cpu and memory limits, is that not possible for this use case..?

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
```diff
@@ -152,4 +174,29 @@ export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalanc
       throw new Error(`${name}: Must be a non-negative integer; received ${value}`);
     }
   }
+
+  private validateContainerCpu(containerCpu?: number, cpu: number = 256) { // default value for cpu is 256
```
shikha372, line 178: where are these defaults defined ?

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: @shikha372 

When specifying `taskDefinition` as a props of `ApplicationLoadBalancedFargateService`, the newly added `containerCpu` and `containerMemoryLimitMiB` props in this MR do not work.

See: https://github.com/aws/aws-cdk/blob/89290c648a2234479802e1218b6dd2e47f7b1b33/packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts#L87-%23L99

so, it seems impossible.

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
shikha372: Thanks @ren-yamanashi , I see  the default values are in fargate task definition and not under being set under `props.cpu` provided, i would prefer if we keep the check without default value as we might miss the update in case it changes in future as well.

ren-yamanashi: @shikha372 

Thank you for your comment.

In the current `validateContainerCpu` method, it compares the values of `containerCpu` and `cpu` and returns an error if containerCpu exceeds cpu.

As you pointed out, referencing the default value of `FargateServiceBaseProps.cpu` in `ApplicationLoadBalancedFargateService`’s `validateContainerCpu` could indeed lead to an issue where we might miss the update in case it changes in future as well.

However, avoiding the use of the default value of `FargateServiceBaseProps.cpu` within `ApplicationLoadBalancedFargateService` to address this concern could lead to the following problem:

#### `validateContainerCpu` method might not validate properly

If don’t reference the default value of `FargateServiceBaseProps.cpu`, the `cpu` param in `validateContainerCpu` would need to be optional.
(because `props.cpu` is optional)

```typescript
  private validateContainerCpu(containerCpu?: number, cpu?: number) {
    /**
     * ... Implementation of the method
     */
  }
```

In this scenario, it becomes challenging to validate cases where:

- `containerCpu` is set, but cpu is not.

In this cases, an error should be returned if `containerCpu` exceeds the default value of `FargateServiceBaseProps.cpu` (256), but without referencing the default value within the method, determining if an error should be returned becomes difficult.

---

### Possible Approaches for Resolution

Taking these considerations into account, I’ve outlined the following possible approaches. What could you think?

Note: this approaches would result in referencing the default value of `FargateServiceBaseProps.cpu` in `ApplicationLoadBalancedFargateService`, but I thought it could be more appropriate than the current implementation, so I am suggesting them.

#### Override props.cpu in `ApplicationLoadBalancedFargateServiceProps`

override `props.cpu` in `ApplicationLoadBalancedFargateServiceProps` and pass it to the `FargateTaskDefinition` argument within `ApplicationLoadBalancedFargateService`, as shown below:

```typescript
export interface ApplicationLoadBalancedFargateServiceProps extends ApplicationLoadBalancedServiceBaseProps, FargateServiceBaseProps {
  /**
   * ... some comments
   *
   * @default 256
   */
  readonly cpu?: number;

 /**
  * ... some props
  */
}

export class ApplicationLoadBalancedFargateService extends ApplicationLoadBalancedServiceBase {
  // ... some implementation
  constructor(scope: Construct, id: string, props: ApplicationLoadBalancedFargateServiceProps = {}) {
    // ... some implementation

      this.taskDefinition = new FargateTaskDefinition(this, 'TaskDef', {
        cpu: props.cpu ?? 256,
        // ... some props
      });
  }
}   
```

shikha372: Thank you @ren-yamanashi for defining the alternate detail approach, I would modify the current behaviour of the construct by setting this default, I believe that if the value of containerCpu and cpu don't align this would result in a failed deployment, if the intention is to fail fast before the deployment itself, can we add a warning instead of an error in validation so that in case these values change in the future it will not block the customer deployments.

shikha372 (changes requested):
Thank you @ren-yamanashi for defining the alternate approach in detail, I would not modify the current behaviour of the construct by setting the default for props.cpu, I believe in case, value of containerCpu and cpu doesn't align with each other, this would result in a failed deployment (please let me know if that is not the case), given that if intention by adding this validation is to fail fast before the deployment itself, can we add a warning instead of an error in the function so that in case these values change in near future it will not block customer deployments.

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: @shikha372 

I apologize for the delay in responding to you.

If the values of `containerCpu` and `cpu` do not match, no error occurs, and the error occurs if the value specified for `containerCpu` is greater than `cpu`.


Considering the concerns you mentioned, I thought it would be better to allow `containerCpu` to be specified only if `cpu` is specified. What are your thoughts on this?

gracelu0: Hi @ren-yamanashi , discussed with @shikha372 and the suggestion is to validate using `this.taskDefinition.cpu` since FargateTaskDefinition sets a default for `cpu`. i.e. 
```
this.validateContainerCpu(props.containerCpu, this.taskDefinition.cpu);
```

It looks like we will need to surface `cpu` in the `FargateTaskDefinition` class as a public readonly member in order to do this. Let us know if you're still planning to work on this, otherwise we can take over to get it merged (we will make sure to credit you as a co-author :) ). Thanks!

gracelu0 (dismissed):

packages/aws-cdk-lib/aws-ecs-patterns/lib/fargate/application-load-balanced-fargate-service.ts
ren-yamanashi: @gracelu0 

Thank you for review and suggestions on a clear implementation policy!  
I will modify the implementation on my part!

ren-yamanashi: @gracelu0 

I have fixed it, so please confirm!

https://github.com/aws/aws-cdk/pull/30920/commits/40fb61fa2cdc35edc2a433542e156002f751a982

gracelu0 (dismissed):
Thank you for contributing!

gracelu0:
@Mergifyio update

gracelu0 (approved):

PR #30935, opened 2024-07-24
feat(events): add  `description` property for `eventBus`

yuppe-kyupeen:
### Issue # (if applicable)

N/A

### Reason for this change
The change introduces the `description` property to the `EventBus`

### Description of changes
- Add the `description` property for `EventBusProps`, which was missing in the L2 construct.
- Add validation for character limit 

I Added a unit test for eventBus and added the `description` property in the integration tests.

### Checklist

----


packages/aws-cdk-lib/aws-events/test/event-bus.test.ts
```diff
@@ -40,11 +40,13 @@ describe('event bus', () => {
     // WHEN
     new EventBus(stack, 'Bus', {
       eventBusName: 'myEventBus',
+      description: 'myEventBus',
```
go-to-k, line 43: Could you separate it and create a new test because the name of this test is `'named event bus'`?

packages/aws-cdk-lib/aws-events/test/event-bus.test.ts
```diff
@@ -281,6 +283,19 @@ describe('event bus', () => {
     }).toThrow(/'eventSourceName' must satisfy: /);
   });
 
+  test('throw error when description is too long', () => {
```
go-to-k, line 286: Usually this test name is fine, but in this file, many tests for other errors have names of the following form: `'event bus xx cannot ...'`, so could you please match them?

go-to-k (changes requested):
Nice, but I just made a few comments.

yuppe-kyupeen:
@go-to-k 
Thanks for always reviewing!
I've made the changes. Please check🙏

go-to-k (approved):
Thanks!

packages/aws-cdk-lib/aws-events/lib/event-bus.ts
```diff
@@ -325,9 +336,14 @@ export class EventBus extends EventBusBase {
 
     super(scope, id, { physicalName: eventBusName });
 
+    if (props?.description && !Token.isUnresolved(props.description) && props.description.length > 512) {
```
Leo10Gama, line 339: I'm curious why we're checking for a token in this field. If we're letting users pass in the description as a token, will this catch whether that token resolves to a value > 512 characters?

I think we should also include a unit test or two making sure tokenized values are resolved properly, and that tokenized strings > 512 characters throw the same expected error as a non-tokenized string.

Leo10Gama (changes requested):
Thanks for the PR! Just left one comment regarding the token check.

packages/aws-cdk-lib/aws-events/lib/event-bus.ts
yuppe-kyupeen: @Leo10Gama 
Thank you for your suggestion.

I believe we should not add the unit test you mentioned for the following two reasons:

- It is unlikely that the `description` will be passed as a token. Of course, this does not mean that such cases do not exist.

- Tokenized values are resolved during the CDK lifecycle, particularly during the deployment stage, not during the synthesis stage. Their validation is already handled by CFn and the respective AWS services, and I don't believe it's the responsibility of CDK unit tests. In fact, the CDK user guide (https://docs.aws.amazon.com/cdk/v2/guide/tokens.html#tokens_encoding) states that "If name is a token, validation isn't performed, and an error could still occur in a later stage in the lifecycle, such as during deployment," indicating that errors are expected during the deployment lifecycle when a token is passed.

@go-to-k, do you have any thoughts?

go-to-k: This process says that if the property is a token, no validation will be performed.
In other words, this validation is skipped in the CDK layer, so I think unit testing is not necessary.

Although it is unlikely that the token itself will exceed 512 characters, the developer may mimic this validation process when other properties are newly added. Considering such a case, I think it is worthwhile to include such a process (skip tokens during validation).

Leo10Gama: That makes sense, thanks for the clarification!

packages/aws-cdk-lib/aws-events/lib/event-bus.ts
```diff
@@ -80,6 +80,17 @@ export interface EventBusProps {
    */
   readonly eventSourceName?: string;
 
+  /**
+   * The event bus description.
+   *
+   * The description must be maximum 512 characters long.
```
Leo10Gama, line 86: This phrasing feels a little confusing to me. Can we change it to something more like:
```suggestion
   * The description can be up to 512 characters long.
```

Leo10Gama (commented):
Giving it a second look-over, everything seems fine! Just one minor nit about the phrasing of the docstring.

TheRealAmazonKendra (commented):
Committing the suggested change to the wording and approving per @Leo10Gama's review.

Leo10Gama (dismissed):

Leo10Gama (approved):

PR #30944, opened 2024-07-24
feat(s3-deployment): added property `outputObjectKeys` for BucketDeployment

### LINKED ISSUE #28579: (aws-s3-deployment): Fails to deploy when a large number of object keys are present
### Describe the bug

When using the BucketDeployment for a large number of individual source files, the custom resource correctly persists them to S3, but ultimately fails the deployment due the large response send back to CloudFormation, exceeding the 4k limit for response objects. 

### Expected Behavior

The Cfn deployment should succeed if the BucketDeployment's underlying custom resource succeeds. 

### Current Behavior

Cfn fails to deploy the stack with `Response object is too long` despite the BucketDeployment lambda succeeding and successfully deploying all resources as expected. 

### Reproduction Steps

Any bucket deployment where the number of sources is sufficient to exceed the 4k response count. In our case, 50 config files succeeded to persist to s3  but failed due to the response to Cfn. 

```
    new BucketDeployment(this, `Deployment`, {
      sources: configs.map((conf) => Source.jsonData("someKey", conf)), 
      destinationBucket: bucket,
      memoryLimit: 1024,
      prune: false,
    });
```

### Possible Solution

The underlying custom resource provides the ability to choose an output path in the response, this could be used to truncate the response to Cfn and avoid this issue when the number of keys is sufficiently large or if a user intentionally configures it to do so. 

_No response_

### CDK CLI Version

2.103.1

### Framework Version

_No response_

### Node.js Version

18

### OS

any

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

mrlikl:
### Issue # (if applicable)

Closes #28579

### Reason for this change

The [CR lambda](https://github.com/aws/aws-cdk/blob/597228c1552a21f8dc7250a0be62160f838bb776/packages/%40aws-cdk/custom-resource-handlers/lib/aws-s3-deployment/bucket-deployment-handler/index.py#L138C14-L138C30) is essentially sending back the same data in the response which is hitting the limit for close to 50 object uploads.

Particularly this is being a limitation when using servicecatalog.ProductStack, if there are local assets beyond a particular number, the Custom::CDKBucketDeployment would fail with the error Response object is too long which is a hard limit of 4096 bytes.

### Description of changes

1. Added a new property to control the custom resource sending large data and hitting the 4096 bytes limit even though the deployment operation is successful. 
2. The property `outputObjectKeys` has been set to false by default for the service catalog product so that the error does not occur. 

Validated using a sample stack with the property set and confirmed the behavior. Also, the existing deployments would be unaffected. 

### Checklist

----


packages/@aws-cdk/custom-resource-handlers/lib/aws-s3-deployment/bucket-deployment-handler/index.py
```diff
@@ -135,7 +136,7 @@ def cfn_error(message=None):
         cfn_send(event, context, CFN_SUCCESS, physicalResourceId=physical_id, responseData={
             # Passing through the ARN sequences dependencees on the deployment
             'DestinationBucketArn': props.get('DestinationBucketArn'),
-            'SourceObjectKeys': props.get('SourceObjectKeys'),
+            **({'SourceObjectKeys': props.get('SourceObjectKeys')} if output_object_keys else {})
```
Leo10Gama, line 139: I'm not too familiar on what the `**` operator does in this context?

packages/aws-cdk-lib/aws-servicecatalog/lib/private/product-stack-synthesizer.ts
```diff
@@ -104,6 +104,7 @@ export class ProductStackSynthesizer extends cdk.StackSynthesizer {
         serverSideEncryption: this.serverSideEncryption,
         serverSideEncryptionAwsKmsKeyId: this.serverSideEncryptionAwsKmsKeyId,
         memoryLimit: this.memoryLimit,
+        outputObjectKeys: false,
```
Leo10Gama, line 107: Is setting this behaviour to false for this one module the right call? If old users have existing setups that rely on this property being true and the object keys being sent, then this acts as a breaking change for them. We can default this to true to preserve existing behaviour, and if people are running into this issue, they can set this attribute themselves.

packages/@aws-cdk-testing/framework-integ/test/aws-s3-deployment/test/integ.bucket-deployment.ts
```diff
@@ -99,6 +99,19 @@ class TestBucketDeployment extends cdk.Stack {
...
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      autoDeleteObjects: true, // needed for integration test cleanup
+    });
+
+    new s3deploy.BucketDeployment(this, 'DeployMe6', {
+      sources: [s3deploy.Source.asset(path.join(__dirname, 'my-website-second'))],
+      destinationBucket: bucket6,
+      retainOnDelete: false, // default is true, which will block the integration test cleanup
+      outputObjectKeys: false,
+    });
```
Leo10Gama, line 114: Can we confirm that this test fails with `outputObjectKeys = true`? That is, will this `BucketDeployment` make a response object that exceeds the 4K character count, but deploys successfully when we do not include the object keys in the output?

Leo10Gama (changes requested):
Thanks for the PR! Just a few comments and concerns to be addressed.

packages/aws-cdk-lib/aws-servicecatalog/lib/private/product-stack-synthesizer.ts
moelasmar: `ProductStackSynthesizer ` construct is a private construct that is not available for the customers, and also the `bucketDeployment` is also is not accessible by the customers, so I think it is Ok to set this property to false.

packages/@aws-cdk/custom-resource-handlers/lib/aws-s3-deployment/bucket-deployment-handler/index.py
moelasmar: This operator is to merge the content of that dictionary with the parent dictionary

moelasmar:
I am not able to push large files to this PR, I copied this PR to a new one https://github.com/aws/aws-cdk/pull/31452 ... I will close this one in favor of the new one

packages/@aws-cdk-testing/framework-integ/test/aws-s3-deployment/test/integ.bucket-deployment.ts
moelasmar: I added a new integration Test case that test the large number of assets that should fail if the new property is not there

PR #30949, opened 2024-07-25
feat(sagemaker): add support uncompressed model

WinterYukky:
### Issue # (if applicable)

None.

### Reason for this change

Currently, `sagemaker.Model` only support `ModelDataUrl`. `ModelDataUrl` must be gzip the model artifact. 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sagemaker-model-containerdefinition.html#cfn-sagemaker-model-containerdefinition-modeldataurl

LLM, which is popular these days, is large and takes longer to compress. For example, the 70B parameter model could take 2 hours or more to compress. SageMaker has a way to use the uncompressed model, and this PR is needed to support the uncompressed model.

The uncompressed model is also recommended in the official documentation.
> When deploying ML models, one option is to archive and compress the model artifacts into a tar.gz format. Although this method works well for small models, compressing a large model artifact with hundreds of billions of parameters and then decompressing it on an endpoint can take a significant amount of time. For large model inference, we recommend that you deploy uncompressed ML model.  
https://docs.aws.amazon.com/sagemaker/latest/dg/large-model-inference-uncompressed.html

### Description of changes

Users can now specify the use of the uncompressed model by taking the model's compression type and S3 data type as options. When using an uncompressed model, use `ModelDataSource` instead of `ModelDataUrl`. Since `ModelDataSource` cannot be used in some workloads, the compression model uses `ModelDataUrl` as before.
> [!NOTE]
Currently you cannot use ModelDataSource in conjunction with SageMaker batch transform, SageMaker serverless endpoints, SageMaker multi-model endpoints, and SageMaker Marketplace.
https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_ContainerDefinition.html#sagemaker-Type-ContainerDefinition-ModelDataSource

Also, asset bundling can now be used as a by-product of being able to use something other than tar.gz due to support for uncompressed models. Until now, no model artifacts with extensions other than tar.gz were permitted, but bundled assets are now allowed if they are not directories.

Since it is not possible to upload a directory or extract a zip with `s3_asset.asset` alone, the directory cannot continue to be used as an asset. (If there is a way to support directories, please let me know)

I've added an integ test with assertions that invoke the model endpoint and verify the response.

### Checklist

### Other
sagemaker-alpha module document is not aligned contributing guide recommendation, so also I fix the it.  
https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#recommendations

----


packages/@aws-cdk/aws-sagemaker-alpha/lib/model-data.ts
```diff
@@ -70,9 +159,6 @@ class AssetModelData extends ModelData {
 
   constructor(private readonly path: string, private readonly options: assets.AssetOptions) {
     super();
-    if (!path.toLowerCase().endsWith(ARTIFACT_EXTENSION)) {
-      throw new Error(`Asset must be a gzipped tar file with extension ${ARTIFACT_EXTENSION} (${this.path})`);
-    }
```
WinterYukky, line 75: The asset's availability determination was moved after binding it as an asset.

packages/@aws-cdk/aws-sagemaker-alpha/lib/model-data.ts
```diff
@@ -83,11 +169,17 @@ class AssetModelData extends ModelData {
         ...this.options,
       });
     }
-
+    if (!this.asset.isFile) {
+      throw new Error(`Asset must be a file, if you want to use directory you can use 'ModelData.fromBucket()' with the 's3DataType' option to 'S3DataType.S3_PREFIX' and 'compressionType' option to 'CompressionType.NONE' (${this.path})`);
+    }
```
WinterYukky, line 174: The bundled assets are now allowed if they are not directories.

packages/@aws-cdk/aws-sagemaker-alpha/lib/model-data.ts
```diff
@@ -83,11 +169,17 @@ class AssetModelData extends ModelData {
...
     this.asset.grantRead(model);
 
     return {
       uri: this.asset.httpUrl,
+      compressionType: this.asset.assetPath.toLowerCase().endsWith(COMPRESSED_ARTIFACT_EXTENSION)
+        ? CompressionType.GZIP
+        : CompressionType.NONE,
+      s3DataType: S3DataType.S3_OBJECT,
```
WinterYukky, line 182: Uncompressed single files are also supported

packages/@aws-cdk/aws-sagemaker-alpha/lib/model-data.ts
```diff
@@ -51,8 +113,33 @@ export abstract class ModelData {
   public abstract bind(scope: Construct, model: IModel): ModelDataConfig;
 }
 
+/**
+ * The options for identifying model artifacts.
+ * When you choose `CompressionType.GZIP` and `S3DataType.S3_OBJECT` then use `ModelDataUrl` property.
+ * Otherwise, use `ModelDataSource` property.
```
Leo10Gama, line 119: Reading this, I interpret it as telling the user to use the `ModelDataUrl` or `ModelDataSource` properties, but looking through the rest of the changes, it seems like this is something handled internally. Is there a reason the user needs to know about what's going on under the hood?

packages/@aws-cdk/aws-sagemaker-alpha/lib/model-data.ts
```diff
@@ -6,7 +6,58 @@ import { hashcode } from './private/util';
...
+   * S3 URI identifies an object that is a gzip-compressed TAR archive.
+   * SageMaker will attempt to decompress and untar the object during model deployment.
+   */
+  GZIP = 'Gzip',
+  /**
+   * If you choose `CompressionType.NONE` and choose `S3DataType.S3_PREFIX` as the value of `s3DataType`,
+   * S3 URI identifies a key name prefix, under which all objects represents the uncompressed ML model to deploy.
+   *
+   * If you choose `CompressionType.NONE`, then SageMaker will follow rules below when creating model data files
+   * under `/opt/ml/model` directory for use by your inference code:
```
Leo10Gama, line 26: The `CompressionType` and `S3DataType` enums seem to be used a lot in conjunction with each other, and when their behaviour changes this drastically between combinations, it makes me wonder if there is any way we can combine the two? Right now, for instance, there doesn't seem to be anything documented about what will happen if I combine `CompressionType.GZIP` and `S3DataType.S3_PREFIX`, if that's a valid combination.

packages/@aws-cdk/aws-sagemaker-alpha/lib/model-data.ts
```diff
@@ -6,7 +6,58 @@ import { hashcode } from './private/util';
...
+   * - Do not use any of the following as file names or directory names:
+   *   - An empty or blank string
+   *   - A string which contains null bytes
+   *   - A string longer than 255 bytes
+   *   - A single dot (.)
+   *   - A double dot (..)
+   * - Ambiguous file names will result in model deployment failure.
+   *    For example, if your uncompressed ML model consists of two S3 objects `s3://mybucket/model/weights` and `s3://mybucket/model/weights/part1`
+   *    and you specify `s3://mybucket/model/` as the value of S3 URI and `S3DataType.S3_PREFIX` as the value of `s3DataType`,
+   *    then it will result in name clash between `/opt/ml/model/weights` (a regular file) and `/opt/ml/model/weights/` (a directory).
```
Leo10Gama, line 41: This docstring feels unnecessarily lengthy, and the mention of what file/directory names doesn't seem relevant, at least for the enum. Is there anyway we can add a `@see` tag and link to the documentation instead for all this extra information?

packages/@aws-cdk/aws-sagemaker-alpha/lib/model.ts
```diff
@@ -357,11 +357,22 @@ export class Model extends ModelBase {
...
       environment: container.environment,
-      modelDataUrl: container.modelData ? container.modelData.bind(this, this).uri : undefined,
+      modelDataSource: useModelDataSource ? {
+        s3DataSource: {
+          s3Uri: modelDataConfig.uri,
+          s3DataType: modelDataConfig.s3DataType!,
+          compressionType: modelDataConfig.compressionType!,
+        },
+      } : undefined,
+      modelDataUrl: !useModelDataSource ? modelDataConfig?.uri : undefined,
```
Leo10Gama, line 375: Would it be possible for both of these values to be set at the same time? And if so, what sort of behaviour would we expect?

Leo10Gama (dismissed):
Thanks for the contribution! This is a great start for this feature, but I have a few concerns with the level of abstraction we're providing, specifically with the two enums for `CompressionType` and `S3DataType`. We should generally avoid situations where we tell users "if you want to use this, you also have to set this variable that's somewhere else", since it leads to a lesser user experience. If there's some way to refactor this as one enum, I think that would be ideal.

PR #30966, opened 2024-07-28
feat(s3): replicating objects

### LINKED ISSUE #1680: s3: bucket replication
PR #184 has some initial work around S3 bucket replication

badmintoncryer:
### Issue # (if applicable)

Closes #1680.

### Reason for this change

AWS S3 supports configuring [object replication](https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication.html) , but the `s3.Bucket` construct does not support it.

### Description of changes

Added `replicationRules` to `BucketProps`.

#### Replication configuration version

There are two versions of [replication configuration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-add-config.html#replication-backward-compat-considerations). This PR uses only the V2 replication configuration to enable the specification of the Filter element and S3 Replication Time Control (S3 RTC).

To use V2 replication configuration, this PR explicitly specifies [Filter.Prefix](https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-replicationrulefilter.html#cfn-s3-bucket-replicationrulefilter-prefix) property.
```ts
        const prefix = rule.prefixFilter ?? '';
        const filter = isAndFilter ? {
          and: {
            prefix,
            tagFilters: rule.tagFilter,
          },
        } : {
          prefix,
        };
```

V2 replication configuration has some restriction:
- Must specify [DeleteMarkerReplication](https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-replicationrule.html#cfn-s3-bucket-replicationrule-deletemarkerreplication)
```sh
ReplicationStack | 4/7 | 9:22:08 PM | CREATE_FAILED        | AWS::S3::Bucket  | SourceBucket (SourceBucketDDD2130A) Resource handler returned message:
Delete marker replication is not supported if any Tag filter is specified. Please refer to S3 Developer Guide for more information. (Service: S3, Status Code: 400, Request ID: XXX, Extended Request ID: XXX)
```
- Must specify [Priority](https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-replicationrule.html#cfn-s3-bucket-replicationrule-priority)
```sh
ReplicationStack | 4/7 | 9:12:08 PM | CREATE_FAILED        | AWS::S3::Bucket  | SourceBucket (SourceBucketDDD2130A) Resource handler returned message:
Priority must be specified for this version of Cross Region Replication configuration schema. Please refer to S3 Developer Guide for more information. (Service: S3, Status Code: 400, Request ID: XXX, Extended Request ID: XXX)
```

These restriction is not documented but there are some posts about these points.
- https://repost.aws/questions/QUiEc8wFE_Q16fX5WG-YWnrA/cloudformation-support-for-s3-replication-to-multiple-destination-buckets

To resolve these problems,I made the `priority` required and explicitly set the `deleteMarkerReplication`.

```ts
       const prefix = rule.prefixFilter ?? ''; // set empty string to use V2 replication configuration
        const filter = isAndFilter ? {
          and: {
            prefix,
            tagFilters: rule.tagFilter,
          },
        } : {
          prefix,
        };

        return {
          id: rule.id,
          priority: rule.priority,
          status: 'Enabled',
          destination: {
            bucket: rule.destination.bucket.bucketArn,
            account: rule.destination.account,
            storageClass: rule.storageClass?.toString(),
            accessControlTranslation: rule.destination.accessControlTransition ? {
              owner: 'Destination',
            } : undefined,
            encryptionConfiguration: rule.kmsKey ? {
              replicaKmsKeyId: rule.kmsKey.keyArn,
            } : undefined,
            replicationTime: rule.replicationTimeControl !== undefined ? {
              status: rule.replicationTimeControl ? 'Enabled' : 'Disabled',
              time: {
                minutes: 15,
              },
            } : undefined,
            metrics: rule.replicationTimeControlMetrics !== undefined ? {
              status: rule.replicationTimeControlMetrics ? 'Enabled' : 'Disabled',
              eventThreshold: {
                minutes: 15,
              },
            } : undefined,
          },
          filter,
          // To avoid deploy error when there are multiple replication rules with undefined deleteMarkerReplication,
          // CDK explicitly set the deleteMarkerReplication if it is undefined.
          deleteMarkerReplication: {
            status: rule.deleteMarkerReplication ? 'Enabled' : 'Disabled',
          },
          sourceSelectionCriteria,
        };
```

#### IAM permission

There is a [documentation to setup IAM permissions for service role](https://docs.aws.amazon.com/AmazonS3/latest/userguide/setting-repl-config-perm-overview.html).

```json
{
   "Version":"2012-10-17",
   "Statement":[
      {
         "Effect":"Allow",
         "Action":[
            "s3:GetReplicationConfiguration",
            "s3:ListBucket"
         ],
         "Resource":[
            "arn:aws:s3:::SRC-BUCKET"
         ]
      },
      {
         "Effect":"Allow",
         "Action":[
            "s3:GetObjectVersionForReplication",
            "s3:GetObjectVersionAcl",
            "s3:GetObjectVersionTagging"
         ],
         "Resource":[
            "arn:aws:s3:::SRC-BUCKET/*"
         ]
      },
      {
         "Effect":"Allow",
         "Action":[
            "s3:ReplicateObject",
            "s3:ReplicateDelete",
            "s3:ReplicateTags"
         ],
         "Resource":"arn:aws:s3:::DST-BUCKET/*"
      }
   ]
}
```

However, there are discrepancies between the automatically generated IAM policies in the management console and the IAM policies in the documentation.

Generated Policy:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": [
                "s3:ListBucket",
                "s3:GetReplicationConfiguration",
                "s3:GetObjectVersionForReplication",
                "s3:GetObjectVersionAcl",
                "s3:GetObjectVersionTagging",
                "s3:GetObjectRetention",
                "s3:GetObjectLegalHold"
            ],
            "Effect": "Allow",
            "Resource": [
                "arn:aws:s3:::SRC-BUCKET",
                "arn:aws:s3:::SRC-BUCKET/*"
            ]
        },
        {
            "Action": [
                "s3:ReplicateObject",
                "s3:ReplicateDelete",
                "s3:ReplicateTags",
                "s3:GetObjectVersionTagging",
                "s3:ObjectOwnerOverrideToBucketOwner"
            ],
            "Effect": "Allow",
            "Condition": {
                "StringLikeIfExists": {
                    "s3:x-amz-server-side-encryption": [
                        "aws:kms",
                        "aws:kms:dsse",
                        "AES256"
                    ]
                }
            },
            "Resource": [
                "arn:aws:s3:::DST-BUCKET/*"
            ]
        },
        {
            "Action": [
                "kms:Decrypt"
            ],
            "Effect": "Allow",
            "Condition": {
                "StringLike": {
                    "kms:ViaService": "s3.ap-northeast-1.amazonaws.com",
                    "kms:EncryptionContext:aws:s3:arn": [
                        "arn:aws:s3:::SRC-BUCKET/*"
                    ]
                }
            },
            "Resource": [
                "arn:aws:kms:ap-northeast-1:123456789012:key/hogehuga"
            ]
        },
        {
            "Action": [
                "kms:Encrypt"
            ],
            "Effect": "Allow",
            "Condition": {
                "StringLike": {
                    "kms:ViaService": [
                        "s3.ap-northeast-1.amazonaws.com"
                    ],
                    "kms:EncryptionContext:aws:s3:arn": [
                        "arn:aws:s3:::DST-BUCKET*"
                    ]
                }
            },
            "Resource": [
                "arn:aws:kms:ap-northeast-1:123456789012:key/hogefuga"
            ]
        }
    ]
}
```

I adopted the policy from the document. I look forward to hearing your thoughts on this matter.

Added both unit and integ tests.

### Checklist

----


shikha372:
Hi @badmintoncryer , Thank you for submitting this PR, to let you know since this PR is adding new policies to the template for replication we need to go through an internal security review as well while we go through this PR, will keep you updated.

shikha372:
Hi @badmintoncryer , Thank you for submitting this PR, as this feature is adding some additional pemissions for s3, we'll need to have an internal security review while we review this PR. Will keep you posted !!

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -2460,6 +2609,122 @@ export class Bucket extends BucketBase {
...
+            storageClass: rule.storageClass?.toString(),
+            accessControlTranslation: rule.destination.accessControlTransition ? {
+              owner: 'Destination',
+            } : undefined,
+            encryptionConfiguration: rule.kmsKey ? {
+              replicaKmsKeyId: rule.kmsKey.keyArn,
+            } : undefined,
+            replicationTime: rule.replicationTimeControl !== undefined ? {
+              status: rule.replicationTimeControl ? 'Enabled' : 'Disabled',
+              time: {
```
shikha372, line 2705: curious we why thought of hardcoding the minute value here

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1400,6 +1400,145 @@ export abstract class TargetObjectKeyFormat {
...
+   */
+  readonly destination: ReplicationDestination;
+
+  /**
+   * Specifying S3 Replication Time Control (S3 RTC),
+   * including whether S3 RTC is enabled and the time when all objects and operations on objects must be replicated.
+   *
+   * @default - S3 Replication Time Control is not enabled
+   */
+  readonly replicationTimeControl?: boolean;
```
shikha372, line 1418: from the [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-replicationtime.html) it seems we should have the option of specifying time value as well so should we have an interface here that takes in timevalue ( also keeping the scope for additional values other than minute in case its introduced in future) ? 
Reading further, i see the only allowed value is 15, it makes sense to extend if we can have more value going forward in the future for that field(will check with the team and get back on this)

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1400,6 +1400,145 @@ export abstract class TargetObjectKeyFormat {
...
+
+  /**
+   * A container specifying replication metrics-related settings enabling replication metrics and events.
+   *
+   * @default - Replication metrics are not enabled
+   */
+  readonly metrics?: boolean;
```
shikha372, line 1425: above interface eg. `IReplicationTimeValue` should help here as well in case we have a modifiable value for minutes in the future

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1400,6 +1400,145 @@ export abstract class TargetObjectKeyFormat {
   public abstract _render(): CfnBucket.LoggingConfigurationProperty['targetObjectKeyFormat'];
 }
 
+/**
+ * Specifies which Amazon S3 objects to replicate and where to store the replicas.
+ */
+export interface ReplicationRule {
+  /**
+   * A container for information about the replication destination.
```
shikha372, line 1408: can we also document here that currently destination can be same account or cross acount

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -2460,6 +2609,122 @@ export class Bucket extends BucketBase {
...
+      effect: iam.Effect.ALLOW,
+    }));
+    if (destinationBuckets.length > 0) {
+      role.addToPolicy(new iam.PolicyStatement({
+        actions: ['s3:ReplicateObject', 's3:ReplicateDelete', 's3:ReplicateTags', 's3:ObjectOwnerOverrideToBucketOwner'],
+        resources: destinationBuckets.map(bucket => bucket.arnForObjects('*')),
+        effect: iam.Effect.ALLOW,
+      }));
+    }
+
```
scorbiere, line 2658: On the KMS topic, I think we could also use the encryption info from the bucket props/context (`encryptionKey`, `bucketKeyEnabled`) and grant decrypt permission. see: https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-config-for-kms-objects.html#replications

>- You must grant kms:Decrypt permissions for the AWS KMS key that's used to decrypt the source object.
>
>- You must grant kms:Encrypt permissions for the AWS KMS key that's used to encrypt the object replica.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1400,6 +1400,145 @@ export abstract class TargetObjectKeyFormat {
...
+  /**
+   * Replicate to another bucket in a different account.
+   *
+   * @param bucket the destination bucket
+   * @param account the account to replicate to
+   * @param accessControlTransition whether to want to change replica ownership to the AWS account that owns the destination bucket. The replicas are owned by same AWS account that owns the source object by default.
+   */
+  public static crossAccount(bucket: IBucket, account: string, accessControlTransition?: boolean): ReplicationDestination {
+    return new ReplicationDestination(bucket, account, accessControlTransition);
```
scorbiere, line 1529: I think there is a requirement that the destination bucket should add a statement in its bucket policy to give the permission to the replication role (example [here](https://repost.aws/knowledge-center/s3-cross-encrypted-replication)). As I don't see the specific logic in this PR I am wondering if this method should be there, throw an exception, or add the logic in the `renderReplicationConfiguration(...)` method.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1400,6 +1400,145 @@ export abstract class TargetObjectKeyFormat {
...
+   * @default - auto generated random ID
+   */
+  readonly id?: string;
+
+  /**
+   * An object key name prefix that identifies the object or objects to which the rule applies.
+   *
+   * @default - applies to all objects
+   */
+  readonly prefixFilter?: string;
```
shikha372, line 1496: can we combine these two filters into one and check that with prefixfilter being set `deleteMarkerReplication` is also enabled  
From docs: `If you specify a Filter in your replication configuration, you must also include a DeleteMarkerReplication element`

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -2460,6 +2609,122 @@ export class Bucket extends BucketBase {
...
+            status: rule.replicaModifications ? 'Enabled' : 'Disabled',
+          } : undefined,
+          sseKmsEncryptedObjects: rule.sseKmsEncryptedObjects !== undefined ? {
+            status: rule.sseKmsEncryptedObjects ? 'Enabled' : 'Disabled',
+          } : undefined,
+        } : undefined;
+
+        // Whether to configure filter settings by And property.
+        const isAndFilter = rule.tagFilter && rule.tagFilter.length > 0;
+        // To avoid deploy error when there are multiple replication rules with undefined prefix,
```
shikha372, line 2677: curious what deployment errors were observed in case of no prefix filter

badmintoncryer: Not specifying a prefix filter means using the V1 replication configuration, which does not allow setting multiple replication rules.

To confirm this, I tried setting the filter property to undefined.

```ts
private renderWebsiteConfiguration(props: BucketProps): CfnBucket.WebsiteConfigurationProperty | undefined {
  ...
  return {
          id: rule.id,
          ...,
          // comment out filter option
          // filter,
        };
};
```

Result (`yarn integ test/aws-s3/test/integ.bucket-replication-test.js --update-on-failed`):
```sh
ReplicationStack | 4/7 | 12:58:45 PM | CREATE_FAILED        | AWS::S3::Bucket  | SourceBucket (SourceBucketDDD2130A) Resource handler returned message: "Number of distinct destination bucket ARNs cannot exceed 1 (Service: S3, Status Code: 400, Request ID: 56SVQEYYFXGSMG66, Extended Request ID: tAt6eHUYI0Hk5LuIaQQizGZDUE3sCLDI0dbcp8EMFU43ZKwqfF44lg6Q878Pz4hp8yjLzZDhnyQ=)" (RequestToken: db7089eb-d101-e4b7-51b4-7244fdc76577, HandlerErrorCode: InvalidRequest)
```

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: The addition to the text is not a problem. I just want to confirm the intention: are there any other destinations for replication besides same-account or cross-account?
For example, if we want to set up replication to both the same account and a cross-account simultaneously, we can do this by configuring multiple ReplicationRules.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: As you mentioned, when dealing with cross-account access, it's necessary to grant a bucket policy on the destination bucket. I think it’s great that this can be done through CDK.

However, I have the following concerns. Would it be possible to implement the proposed solution?

- I believe CDK does not have the authority to grant bucket policies on buckets in a different account.
- Similarly, I think CDK cannot retrieve the bucket policy of a bucket in a different account.

scorbiere: >I believe CDK does not have the authority to grant bucket policies on buckets in a different account.

I think it is not a CDK constraint but an AWS profile permission constraint. 

> Similarly, I think CDK cannot retrieve the bucket policy of a bucket in a different account.

Correct. That's why I am proposing to throw an exception with a message saying that cross account replication setup is not supported by this construct. Which can let the community open a feature request. But if you keep this method as it is, you can already open a bug 😅

badmintoncryer: Thank you very much for your clarification. 
Could you please confirm your intent with one of the following options?

- Do not implement the `crossAccount()` method in this PR and only support replication within the same account.
- Implement the `crossAccount()` method but display a warning related to the destination bucket policy using `Annotations` within the method.
- Other.

There is a possibility that I might not be interpreting your message correctly, so I would appreciate it if you could provide more specific instructions on how to proceed.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: I've added grant decrypt:
```ts
    kmsKeys.forEach(kmsKey => {
      kmsKey.grantEncrypt(role);
    });
    this.encryptionKey?.grantDecrypt(role); // added
```
`this.encryptionKey` contains the return value of `parseEncryption`, which takes into account `bucketKeyEnabled` and `encryptionKey`.

And I've also updated integ tests to verify whether encrypted objects can be replicated.

```ts
const sourceKmsKey = new kms.Key(stack, 'SourceKmsKey', {
});

const sourceBucket = new s3.Bucket(stack, 'SourceBucket', {
  versioned: true,
  encryptionKey: sourceKmsKey, // added
  replicationRules: [...]
})
```

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: I believe the meaning of this document is that when setting a prefix filter, `deleteMarkerReplication` must be explicitly specified. Specifically, when using a prefix filter, replication configuration V2 is required, and in V2, `deleteMarkerReplication` must be explicitly set to either `Enabled` or `Disabled`.

As a separate issue, enabling `deleteMarkerReplication` is not possible when `tagFilter` is enabled. Therefore, I have included validation to address this.

```ts
      if (rule.deleteMarkerReplication && rule.tagFilter) {
        throw new Error('\'tagFilter\' cannot be specified when \'deleteMarkerReplication\' is enabled.');
      }
```

Given these conditions, is it okay to proceed with the integration of prefix filter and tag filter as you suggested?

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
scorbiere: Not sure this is what I initially meant but I think the best trade off would be to clearly document that method and explain that the user will have to configure the destination bucket policy to allow the replication role from the source account to replicate object in this destination bucket. And also add a warning message, which should ideally contain the full ARN of the replication role (this should simplify the customer experience).

---

If you want to put extra effort and  implement the destination bucket policy update if the type is `Bucket` it would be great as well.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
shikha372: i meant `currently` to keep the scope for any additional use cases that we might have under this, if not then same account or cross account should be okay.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
shikha372: checked this documentation [here for replication configuration V2](https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-add-config.html) and see that deletmarker is specified as `enabled ` in case of a tag and prefix filter.  But going with the schema documentation, this field must be set to disabled while specifying a tag filter. Given this i think the statement also could be the other way around that `when specifying `tagFilter` deleteMarkerReplication must be set to disabled`. I'm still leaning towards combining these two fields, let me know if you think otherwise.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: This is `ReplicationRule.Prefix` property and it is different from `ReplicationRule.Filter.Prefix`.
By using the former, it becomes V1, and by using the latter, it becomes V2.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: Understood. 
In that case, would it be acceptable to define an enum like the following and pass it as an argument of `replicationTimeControl`?

```ts
export TimeControl {
  FIFTEEN_MINUTES: 15
}
```

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: I've combined these two filters to `filter` prop!

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: I've added some descriptions about destination bucket's policy.

```ts
/**
   * Replicate to another bucket in a different account.
   *
   * When performing cross-account replication,
   * you need to configure access permissions for the replication role in the bucket policy of the destination bucket.
   *
   * @see https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-walkthrough-2.html
```

And I've added `Annotations` about that.

```ts
        if (rule.destination.account !== undefined && rule.destination.account !== Stack.of(this).account) {
          Annotations.of(this).addInfo(`Cross-account S3 replication is set up. In the destination bucket's bucket policy, please grant access permissions from ${this.stack.resolve(role.roleArn)}.`);
        }
```

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
shikha372: Thanks @badmintoncryer , we can go with enum like class.

badmintoncryer: I've implemented it!

packages/@aws-cdk-testing/framework-integ/test/aws-s3/test/integ.bucket-replication.ts
```diff
@@ -0,0 +1,118 @@
...
+const fourthAssertion = integ.assertions
+  .awsApiCall('S3', 'getObject', {
+    Bucket: destinationBucket2.bucketName,
+    Key: 'prefix-test-object',
+  })
+  .waitForAssertions({
+    totalTimeout: Duration.minutes(5),
+  });
+
+const fifthAssertion = integ.assertions
```
shikha372, line 109: This looks same as the fourth assertion.

badmintoncryer: Thanks. I've updated like below.

```ts
const fourthAssertion = integ.assertions
  .awsApiCall('S3', 'getObject', {
    Bucket: destinationBucket1.bucketName, // bucket1
    Key: 'prefix-test-object',
  })
  .waitForAssertions({
    totalTimeout: Duration.minutes(5),
  });

const fifthAssertion = integ.assertions
  .awsApiCall('S3', 'getObject', {
    Bucket: destinationBucket2.bucketName, // bucket2
    Key: 'prefix-test-object',
  })
  .waitForAssertions({
    totalTimeout: Duration.minutes(5),
  });
```

shikha372:
Hi @badmintoncryer, to update had the first security review for this PR, currently concern is around populating correct permissions for opt-in regions, will let you know if any changes are required to address it.

packages/aws-cdk-lib/aws-s3/README.md
```diff
@@ -782,3 +782,89 @@ new s3.Bucket(this, 'Bucket2', {
...
+declare const destinationBucket2: s3.IBucket;
+declare const kmsKey: kms.IKey;
+
+const sourceBucket = new s3.Bucket(this, 'SourceBucket', {
+  // Versioning must be enabled on both the source and destination bucket
+  versioned: true,
+  replicationRules: [
+    {
+      // The destination bucket for the replication rule.
+      destination: s3.ReplicationDestination.sameAccount(destinationBucket1),
```
rix0rrr, line 806: Why a `ReplicationDestination`? Shouldn't the destination just be an `IBucket` ?

All resources will have a `resourceEnv` property you can use to get account and region from.

packages/aws-cdk-lib/aws-s3/README.md
```diff
@@ -782,3 +782,89 @@ new s3.Bucket(this, 'Bucket2', {
...
+      // Amazon S3 will attempt to replicate objects according to all replication rules.
+      // However, if there are two or more rules with the same destination bucket, then objects will be replicated according to the rule with the highest priority.
+      // The higher the number, the higher the priority.
+      // It is essential to specify priority explicitly when the replication configuration has multiple rules.
+      priority: 1,
+    },
+    {
+      destination: s3.ReplicationDestination.sameAccount(destinationBucket2),
+      priority: 2,
+      // Whether to specify S3 Replication Time Control (S3 RTC).
```
rix0rrr, line 817: Please spare a sentence to explain what enabling RTC means.

According to the docs it has something to do with metrics, but metrics is a separate field here, so now I don't understand what this does.

packages/aws-cdk-lib/aws-s3/README.md
```diff
@@ -782,3 +782,89 @@ new s3.Bucket(this, 'Bucket2', {
...
+      replicationTimeControl: s3.ReplicationTimeValue.FIFTEEN_MINUTES,
+      // Wherther to enable replication metrics.
```
rix0rrr, line 819:
```suggestion
      // Whether to enable replication metrics.
```

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1403,6 +1403,179 @@ export abstract class TargetObjectKeyFormat {
...
+   * @default - S3 Replication Time Control is not enabled
+   */
+  readonly replicationTimeControl?: ReplicationTimeValue;
+
+  /**
+   * A container specifying replication metrics-related settings enabling replication metrics and events.
+   *
+   * @default - Replication metrics are not enabled
+   */
+  readonly metrics?: ReplicationTimeValue;
```
rix0rrr, line 1445: The 15 minute value is used to log events that say something like "SLA missed", right? Might want to explain that here.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
```diff
@@ -1403,6 +1403,179 @@ export abstract class TargetObjectKeyFormat {
...
+   * It is essential to specify priority explicitly when the replication configuration has multiple rules.
+   *
+   * @default 0
+   */
+  readonly priority?: number;
+
+  /**
+   * Specifies whether Amazon S3 replicates delete markers.
+   *
+   * @default false
```
rix0rrr, line 1498: What's the implication of this being `false` ? That replicated objects in versioned buckets don't get deleted?

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
rix0rrr: I think I'd also prefer the bucket policy be added automatically. It is the CDK thing to do.

We can add the annotation if `addToBucketPolicy(): boolean` returns `false`, indicating that we have a referenced bucket.

rix0rrr (dismissed):

packages/aws-cdk-lib/aws-s3/README.md
badmintoncryer: I have changed it to the `IBucket` type. Thank you for suggesting the use of `ResourceEnv`.

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: I've updated to add the bucket policy for the destination bucket.
I referenced [this document](https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-walkthrough-2.html).

In addition, when enabling `accessControlTransition`, I also add this bucket policy.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-change-owner.html

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: I've added below description.

`When a value is set, metrics will be output to indicate whether the replication took longer than the specified time.`

I'm not very confident in writing English, so please feel free to improve my sentences, not just in this context.

packages/aws-cdk-lib/aws-s3/README.md
badmintoncryer: I've updated like this.

```ts
      // Whether to specify S3 Replication Time Control (S3 RTC).
      // S3 RTC replicates most objects that you upload to Amazon S3 in seconds,
      // and 99.99 percent of those objects within specified time.
      replicationTimeControl: s3.ReplicationTimeValue.FIFTEEN_MINUTES,
      // Whether to enable replication metrics about S3 RTC.
      // If set, metrics will be output to indicate whether replication by S3 RTC took longer than the configured time.
      metrics: s3.ReplicationTimeValue.FIFTEEN_MINUTES,
```

packages/aws-cdk-lib/aws-s3/lib/bucket.ts
badmintoncryer: False means do not replicate delete marker to the destination bucket.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/delete-marker-replication.html

I've updated JSDoc.

```ts
  /**
   * Specifies whether Amazon S3 replicates delete markers.
   *
   * @see https://docs.aws.amazon.com/AmazonS3/latest/userguide/delete-marker-replication.html
   *
   * @default - delete markers in source bucket is not replicated to destination bucket
   */
  readonly deleteMarkerReplication?: boolean;
```

badmintoncryer:
@rix0rrr Thank you for your review! I've addressed all of your comments.

shikha372:
Hello @badmintoncryer , 
Apologies to keep you waiting, have completed the feature review for this PR. Current ask is if we can add integ test for bidirectional replication as well before final approval.

badmintoncryer:
> Hello @badmintoncryer , Apologies to keep you waiting, have completed the feature review for this PR. Current ask is if we can add integ test for bidirectional replication and for cross-account replication(if possible) as well before final approval.

@shikha372 Thank you for your information. 

> add integ test for bidirectional replication

What is  the meaning of bidirectional replication? Are you suggesting that we prepare Bucket A and Bucket B, and designate each other as their respective replication destinations?

> I don't have two AWS account...　Would it be better to create a new account and run the tests?

shikha372:
> What is  the meaning of bidirectional replication? Are you suggesting that we prepare Bucket A and Bucket B, and designate each other as their respective replication destinations?  
- Yes, since this can also be supported using this construct, we'll need a integ test for this case.

>  I don't have two AWS account...　Would it be better to create a new account and run the tests? 
- Got it, I will help adding integ test for this case

badmintoncryer:
I considered bidirectional replication. First, I added an addReplicationRules method to allow the destination bucket to be set lazily. However, it seems that the circular reference between the source and destination buckets cannot be resolved. 

```ts
  public addReplicationRules(rules: ReplicationRule[]): void {
    this.replicationRules.push(...rules);
  }
```

Do you have any suggestions on the implementation?

shikha372:
> I considered bidirectional replication. First, I added an `addReplicationRules()` method to allow the destination bucket to be set lazily.
> 
> ```ts
>   public addReplicationRules(rules: ReplicationRule[]): void {
>     this.replicationRules.push(...rules);
>   }
> ```
> 
> However, it seems that the circular reference between the source and destination buckets cannot be resolved.
> 
> Do you have any suggestions on the implementation?

Thanks @badmintoncryer , working on adding cross-account integ test will give it a try at my end for bi-directional.

badmintoncryer:
Thank you very much @shikha372

shikha372 (approved):

badmintoncryer:
@mergifyio update

badmintoncryer:
@mergify update

shikha372:
Hi @badmintoncryer , tried to merge it yesterday but I think there is unit test coverage missing for new addReplication method, I can get back to it next week

badmintoncryer:
Thank you very much @shikha372 . Looking forward to the updates!

shikha372 (approved):

badmintoncryer:
@mergify refresh

badmintoncryer:
@mergify update

badmintoncryer:
@mergify update

badmintoncryer:
@mergify update

badmintoncryer:
@mergify update

mrgrain:
@mergifyio dequeue

badmintoncryer:
@mrgrain Do you know why CodeQL and Codecov fail? I'm only merging the main branch, but these jobs sometimes succeed and sometimes fail.

mrgrain:
@Mergifyio requeue

mrgrain:
@badmintoncryer Just a intermittent issue. I'm re-queuing the PR now.

PR #30978, opened 2024-07-29
chore(apigateway): enhance code arounds method authorizers

GavinZZ:
### Reason for this change

Add comments and enhance the code around method authorizers.

### Description of changes

No behavioural change, should only include comments, variables name update and re-ordering.

All existing tests pass.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
+++ b/packages/aws-cdk-lib/aws-ses/lib/configuration-set.ts
@@ -47,7 +47,7 @@ export interface ConfigurationSetProps {
    * Whether to publish reputation metrics for the configuration set, such as
    * bounce and complaint rates, to Amazon CloudWatch
    *
-   * @default false
+   * @default true
    */
   readonly reputationMetrics?: boolean;
 


packages/aws-cdk-lib/aws-apigateway/lib/method.ts
```diff
@@ -185,22 +185,33 @@ export class Method extends Resource {
...
+     *    └── authorizer: Specifies the default authorizer for all methods created within this API unless custom options are specified.
+     *        └── authorizerType: The default authorization type of this authorizer.
+     *
+     * REST API Method
+     * └── options: Method options.
+     *    ├── authorizationType: Specifies the authorization type.
+     *    └── authorizer: Specifies an authorizer to use for this method.
+     *        └── authorizerType: The authorization type of this authorizer.
+     *
+     * "authorizationType" is recommended to not be specified. We should prioritize using "authorizer's authorizerType",
```
Leo10Gama, line 211: Can we change this comment to not use first-person pronouns like "we"? May be best practice to not refer to the collective "us" when documenting this sort of stuff lol

Leo10Gama (commented):
Just one small nit, otherwise LGTM!

xazhao (approved):
LGTM overall

packages/aws-cdk-lib/aws-apigateway/lib/method.ts
```diff
@@ -185,22 +185,33 @@ export class Method extends Resource {
...
+     *
+     * REST API Method
+     * └── options: Method options.
+     *    ├── authorizationType: Specifies the authorization type, recommended to not be specified.
+     *    └── authorizer: Specifies an authorizer to use for this method.
+     *        └── authorizerType: The authorization type of this authorizer.
+     *
+     * Authorization type is first set to "authorizer.authorizerType", falling back to method's "authorizationType",
+     * falling back to API resource's default "authorizationType", and lastly "Authorizer.NONE".
+     */
```
moelasmar, line 213: minor comment, can you add that the authorizer.authorizerType should match authorizationType if both values are exist.

packages/aws-cdk-lib/aws-apigateway/lib/method.ts
```diff
@@ -303,6 +314,19 @@ export class Method extends Resource {
     this.methodResponses.push(methodResponse);
   }
 
+  private getMethodAuthorizationType(options: MethodOptions, defaultMethodOptions: MethodOptions, authorizerType?: AuthorizationType): string {
```
moelasmar, line 317: add some comment to this method to explain the parameters,  to make it easier for the team to use it in future

moelasmar: also, I believe it is better to extract the authorizerType in this method instead of accepting it as a parameter. The logic of extracting this value is also confusing

packages/aws-cdk-lib/aws-apigateway/lib/method.ts
GavinZZ: Updated

packages/aws-cdk-lib/aws-apigateway/lib/method.ts
GavinZZ: Updated

Issue #18297, opened 2022-01-06
(aws-apigateway) MockIntegrations with defaulCorsPreflightOptions return statusCode 500


Molaire:
### What is the problem?

If an API Gateway Resource allow CORS with defaultCorsPreflightOptions in its options, hitting its child methods that are MockIntegrations will fail with a Internal Error message (statusCode: 500)

This can be manually fixed by setting PassthroughBehavior.NEVER for the Integration request in CDK or the console.

### Reproduction Steps

```typescript

import {
    MockIntegration,
    PassthroughBehavior,
    RestApi,
} from "aws-cdk-lib/aws-apigateway";

const restApi = new RestApi(this, "testApi", {
  defaultCorsPreflightOptions: {
              allowOrigins: Cors.ALL_ORIGINS
          }
    })

const mockIntegration = new MockIntegration({
          integrationResponses: [{
              statusCode: '200',
              responseTemplates: {
                  'application/json': '{"abc": "dfg"}'
              }
          }],
          passthroughBehavior: PassthroughBehavior.NEVER,
          requestTemplates: {
              'application/json': '{ "statusCode": 200 }',
          },
      })

restApi.addMethod(
          "GET",
          mockIntegration,
          {
              methodResponses: [
                  { statusCode: '200' }
              ]
          });
```

### What did you expect to happen?

I expected MockIntegration methods to work beside the LambdaIntegration methods, even with CORS enabled by default.

### What actually happened?

Only the LambdaIntegration works, not the MockIntegration methods.
They work when testing through the API Gateway test option in the console, but not when hitting them with my browser, implying it's a problem with CORS/OPTIONS.

### CDK CLI Version

2.4.0 (build 993f14d)

### Node.js Version

v16.13.1

### OS

Ubuntu 21.10

### Language

Typescript

### Other information

A workaround is to set defaultCorsPreflightOptions to undefined for the resources with MockIntegration and manually add the method OPTIONS with passthroughBehavior: PassthroughBehavior.NEVER.

peterwoodworth:
Hey @Molaire,

This doesn't seem to be an issue with the CDK, but rather CloudFormation or API Gateway. Do you find the same behavior when using the console?

takeshixx:
@peterwoodworth the issue is that `passthroughBehavior` is not set to `NEVER`. However, CloudFormation and API Gateway do not have any defaults regarding CORS. So I would argue that CDK is responsible to ensure that the CORS preflight request is handled properly.

[...]

Issue #906, opened 2018-10-11
apigateway: add explicit support for CORS


eladb:
## Requirements

- [x] `resource.addCorsPreflight(options)`
- [x] [`AllowOrigin`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin)
- [x] [`AllowHeaders`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers) with defaults for API Gateway
- [x] [`AllowMethods`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods)
- [x] [`AllowCredentials`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials)
- [x] [`MaxAge`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age)
- [x] [`ExposeHeaders`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers)
- [ ] Conditional [`Vary`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary) response header
- [x] Configure response status code (should default to 204)
- [x] Support proxy and non-proxy resources
- [x] Support multiple origins through velocity templates (like in [serverless framework](https://github.com/serverless/serverless/blob/master/lib/plugins/aws/package/compile/events/apiGateway/lib/cors.js)).
- [ ] Automatically discover allowed methods by default based on model.
- [x] Recursive (apply to all child resources)
- [x] `LambdaRestApi` (apply a CORS policy to all routes)
- [ ] CORS with custom authorizers (good [post](https://serverless.com/blog/cors-api-gateway-survival-guide/#cors-with-custom-authorizers)) through `AWS::ApiGateway::GatewayResponse`

## Nice to Have

- [ ] Dynamic CORS handler? (through a Lambda proxy and Access-Control-Max-Age=0)

## Non-Requirements

- Response headers in [simple requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Simple_requests) are out of scope.

## Resources

- [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)
[...]

PR #30985, opened 2024-07-31
fix(apigateway): set passthrough behavior to avoid errors on CORS preflight requests

takeshixx:
Set `passthroughBehavior` behavior to `NEVER` to prevent errors for CORS preflight requests with content-types other than `application/json`.

### Issue # (if applicable)

This was reported as #18297, but it was closed without a fix.

### Reason for this change

Using `addCorsPreflight()` will add a mock integration for `OPTIONS` requests and maps them to content-type `application/json`. `OPTIONS` requests with a content-type header other than `application/json` lead to HTTP 500 Internal Server Errors.

### Description of changes

Setting the `passthroughBehavior` to `NEVER` returns a mime type error instead of a internal server error, which is the appropriate response.

It should be noted that this setting was proposed in the initial implementation of `addCorsPreflight()` in #906 already. However, it looks like it didn't make it into the CDK. Instead the default configuration is use, which sets it to `WHEN_NO_MATCH`.

I tested the change by manually overriding the `passthroughBehavior` on the `LambdaRestApi` resource:

```typescript
lambdaRestApi.methods
      .filter((m) => m.httpMethod === 'OPTIONS')
      .forEach((om) => {
        om.node.children.forEach((c) => {
          const mm = c as apigw.CfnMethod;
          mm.addOverride('Properties.Integration.PassthroughBehavior', apigw.PassthroughBehavior.NEVER);
        });
      });
```

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-apigateway/test/cors.test.ts
+++ b/packages/aws-cdk-lib/aws-apigateway/test/cors.test.ts
@@ -733,4 +733,51 @@ describe('cors', () => {
       HttpMethod: 'OPTIONS',
     });
   });
+
+  test('CORS resource with passthroughBehavior set to NEVER', () => {
+    // GIVEN
+    const stack = new Stack();
+    const api = new apigw.RestApi(stack, 'api');
+    const resource = api.root.addResource('MyResource');
+
+    // WHEN
+    resource.addCorsPreflight({
+      allowOrigins: ['https://amazon.com'],
+    });
+
+    // THEN
+    Template.fromStack(stack).hasResourceProperties('AWS::ApiGateway::Method', {
+      HttpMethod: 'OPTIONS',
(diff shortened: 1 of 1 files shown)


takeshixx:
Clarification Request if test cases are required. No "code changes", just changed a mock integration setting.

paulhcsun:
Hi @takeshixx, apologies for the late response. As discussed, please add a unit test case for this change and run the integ tests to update any snapshots.

takeshixx:
I have added a unit test that calls `resource.addCorsPreflight()` and ensures `PassthroughBehavior` is set to `NEVER`. I ran:

```bash
cd packages/aws-cdk-list
yarn build
yarn test
```

And the test ran without errors. Is there anything else that is missing?

PR #31013, opened 2024-08-02
feat(lambda): added new property allowAllIpv6Outbound to FunctionOptions

### LINKED ISSUE #30994: aws-cdk-lib/aws-lambda: Add allowAllIpv6Outbound option to lambda function props
### Describe the feature

There is already an `allowAllOutbound` option. It confused my that this did not actually allow all outbound (only all IPv4 outbound), given that there was no corresponding IPv6 option. I think adding the option would make this more clear. It would also mirror the options on security group props.

### Use Case

It's nice to be able to define a lambda function without explicitly setting up security groups. This option would allow me to do this when I need to allow outbound IPv6 traffic.

### Proposed Solution

It looks like we'd just need to pipe this option through to the security group that gets created (it's already an option on security groups).

### Other Information

_No response_

### Acknowledgements

### CDK version used

3.616

Ubuntu 22.04.4

ashishdhingra:
### Issue # (if applicable)

Closes #30994.

### Reason for this change
[SecurityGroupProps](https://github.com/aws/aws-cdk/blob/9295a85a8fb893d7f5eae06108b68df864096c4c/packages/aws-cdk-lib/aws-ec2/lib/security-group.ts#L272) supports `allowAllIpv6Outbound` property. The existing Lambda [FunctionOptions](https://github.com/aws/aws-cdk/blob/9295a85a8fb893d7f5eae06108b68df864096c4c/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L146C18-L146C33) only supports [allowAllOutbound](https://github.com/aws/aws-cdk/blob/9295a85a8fb893d7f5eae06108b68df864096c4c/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L296C12-L296C28), which is used in [configureVpc()](https://github.com/aws/aws-cdk/blob/9295a85a8fb893d7f5eae06108b68df864096c4c/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L1464C11-L1464C23) while creating a new `SecurityGroup` [here](https://github.com/aws/aws-cdk/blob/9295a85a8fb893d7f5eae06108b68df864096c4c/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L1503).

### Description of changes
Added new property `allowAllIpv6Outbound` to `FunctionOptions`.

Added unit and integration tests.

### Checklist

----


packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -1495,6 +1512,15 @@ Environment variables can be marked for removal when used in Lambda@Edge by sett
       }
     }
 
+    if (props.allowAllIpv6Outbound !== undefined) {
+      if (props.securityGroup) {
+        throw new Error('Configure \'allowAllIpv6Outbound\' directly on the supplied SecurityGroup.');
+      }
+      if (hasSecurityGroups) {
+        throw new Error('Configure \'allowAllIpv6Outbound\' directly on the supplied SecurityGroups.');
+      }
```
lpizzinidev, line 1521:
```suggestion
      if (props.securityGroup || hasSecurityGroups) {
        throw new Error('Configure \'allowAllIpv6Outbound\' directly on the supplied SecurityGroup.');
      }
```
Cleaner.

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3649,6 +3649,50 @@ describe('function', () => {
...
+      const stack = new cdk.Stack();
+
+      expect(() => new lambda.Function(stack, 'MyLambda', {
+        code: new lambda.InlineCode('foo'),
+        handler: 'index.handler',
+        runtime: lambda.Runtime.NODEJS_LATEST,
+        allowAllIpv6Outbound: true,
+      })).toThrow(/Cannot configure \'allowAllIpv6Outbound\' without configuring a VPC/);
+    });
+  });
```
lpizzinidev, line 3695: Can you please add test coverage for the `props.allowAllIpv6Outbound !== undefined && hasSecurityGroup` scenario?

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -295,6 +295,20 @@ export interface FunctionOptions extends EventInvokeConfigOptions {
    */
   readonly allowAllOutbound?: boolean;
```
lpizzinidev, line 296: We should update the property documentation as this allows all traffic except IPv6.
It might not be very clear for the users.

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -1143,6 +1143,9 @@ const fn = new lambda.Function(this, 'Lambda_with_IPv6_VPC', {
 });
 ```
 
+To allow outbound IPv6 traffic by default, set `allowAllIpv6Outbound` to true.
```
lpizzinidev, line 1146: What about defining a `## Outbound traffic` section that explains the behavior for both `allowAllOutbound` and `allowAllIpv6Outbound`?

lpizzinidev (changes requested):

packages/aws-cdk-lib/aws-lambda/lib/function.ts
ashishdhingra: Thanks for the suggestion. The difference in the 2 `if` blocks is that the 1st condition is checking if there is a security group specified as part of function props, whereas, in the 2nd condition relies on `hasSecurityGroups` initialized from `props.securityGroups && props.securityGroups.length > 0;`.

This follows the existing pattern for `allowAllOutbound` property.

packages/aws-cdk-lib/aws-lambda/lib/function.ts
ashishdhingra: Updated. Thanks for the suggestion.

packages/aws-cdk-lib/aws-lambda/README.md
ashishdhingra: Updated. Thanks for the suggestion.

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
ashishdhingra: Done

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3649,6 +3649,86 @@ describe('function', () => {
...
+
+      expect(() => new lambda.Function(stack, 'MyLambda', {
+        code: new lambda.InlineCode('foo'),
+        handler: 'index.handler',
+        runtime: lambda.Runtime.NODEJS_LATEST,
+        allowAllIpv6Outbound: true,
+      })).toThrow(/Cannot configure \'allowAllIpv6Outbound\' without configuring a VPC/);
+    });
+
+    test('allowAllIpv6Outbound defined with SecurityGroup also defined', () => {
```
pahud, line 3696:
```suggestion
    test('throws when both allowAllIpv6Outbound and securityGroup are defined', () => {
```

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3649,6 +3649,86 @@ describe('function', () => {
...
+          {
+            CidrIpv6: '::/0',
+            Description: 'Allow all outbound ipv6 traffic by default',
+            IpProtocol: '-1',
+          },
+        ],
+      });
+    });
+
+    test('allowAllIpv6Outbound set to true but no VPC', () => {
```
pahud, line 3685:
```suggestion
    test('throws when allowAllIpv6Outbound is defined without vpc', () => {
```

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3649,6 +3649,86 @@ describe('function', () => {
...
+        code: new lambda.InlineCode('foo'),
+        handler: 'index.handler',
+        runtime: lambda.Runtime.NODEJS_LATEST,
+        allowAllIpv6Outbound: true,
+        vpc: vpc,
+        securityGroup: securityGroup,
+      })).toThrow(/Configure \'allowAllIpv6Outbound\' directly on the supplied SecurityGroup./);
+    });
+
+    test('allowAllIpv6Outbound defined with SecurityGroups also defined', () => {
```
pahud, line 3714:
```suggestion
    test('throws when both allowAllIpv6Outbound and securityGroups are defined', () => {
```

pahud (dismissed):
Just some minor nit

pahud:
Question - what happens if `allowAllOutbound: false` with `allowAllIpv6Outbound: true`? Is it allowed?

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -1143,6 +1143,27 @@ const fn = new lambda.Function(this, 'Lambda_with_IPv6_VPC', {
...
 
+## Outbound traffic
+By default, when creating a Lambda function, it would add a security group outbound rule to allow sending all network traffic (except IPv6). This is controlled by `allowAllOutbound` in function properties, which has a default value of `true`.
+
+To allow outbound IPv6 traffic by default, explicitly set `allowAllIpv6Outbound` to `true` in function properties as shown below (the default value for `allowAllIpv6Outbound` is `false`):
+```ts
+const vpc = new ec2.Vpc(this, 'Vpc', {
+  maxAzs: 2,
+  restrictDefaultSecurityGroup: false,
+});
```
pahud, line 1154:
```suggestion
import * as ec2 from 'aws-cdk-lib/aws-ec2';

const vpc = new ec2.Vpc(this, 'Vpc', {
  maxAzs: 2,
  restrictDefaultSecurityGroup: false,
});
```

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -1143,6 +1143,27 @@ const fn = new lambda.Function(this, 'Lambda_with_IPv6_VPC', {
...
 ```
 
+## Outbound traffic
+By default, when creating a Lambda function, it would add a security group outbound rule to allow sending all network traffic (except IPv6). This is controlled by `allowAllOutbound` in function properties, which has a default value of `true`.
+
+To allow outbound IPv6 traffic by default, explicitly set `allowAllIpv6Outbound` to `true` in function properties as shown below (the default value for `allowAllIpv6Outbound` is `false`):
+```ts
+const vpc = new ec2.Vpc(this, 'Vpc', {
+  maxAzs: 2,
+  restrictDefaultSecurityGroup: false,
```
pahud, line 1153: Is `restrictDefaultSecurityGroup: false` required if we need to allow all outbound traffic? If yes, maybe we should note this in the README? If not, let's remove this from the sample?

packages/aws-cdk-lib/aws-lambda/README.md
ashishdhingra: Fixed

packages/aws-cdk-lib/aws-lambda/README.md
ashishdhingra: Removed

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
ashishdhingra: Updated

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
ashishdhingra: Updated

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
ashishdhingra: Updated

ashishdhingra:
> Question - what happens if `allowAllOutbound: false` with `allowAllIpv6Outbound: true`? Is it allowed?

@pahud I will test it tomorrow with `ec2.SecurityGroup`. As per my understanding, it would not add outbound rule for `IPv4`.

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -1143,6 +1143,28 @@ const fn = new lambda.Function(this, 'Lambda_with_IPv6_VPC', {
...
+## Outbound traffic
+By default, when creating a Lambda function, it would add a security group outbound rule to allow sending all network traffic (except IPv6). This is controlled by `allowAllOutbound` in function properties, which has a default value of `true`.
+
+To allow outbound IPv6 traffic by default, explicitly set `allowAllIpv6Outbound` to `true` in function properties as shown below (the default value for `allowAllIpv6Outbound` is `false`):
+```ts
+import * as ec2 from 'aws-cdk-lib/aws-ec2';
+
+const vpc = new ec2.Vpc(this, 'Vpc', {
+  maxAzs: 2,
+});
```
pahud, line 1155:
```suggestion
const vpc = new ec2.Vpc(this, 'Vpc');
```

Let's just remove optional props and keep the sample as simple as possible.

pahud (changes requested):
generally LGTM except for some nit

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3649,6 +3649,83 @@ describe('function', () => {
         }),
     ).toThrow(/ADOT Lambda layer can't be configured with container image package type/);
   });
+
+  describe('allowAllIpv6Outbound', () => {
+    test('allowAllIpv6Outbound set to true', () => {
+      const stack = new cdk.Stack();
+      const vpc = new ec2.Vpc(stack, 'Vpc', {
+        maxAzs: 2,
+      });
```
pahud, line 3658:
```suggestion
      const vpc = new ec2.Vpc(stack, 'Vpc');
```

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3649,6 +3649,83 @@ describe('function', () => {
...
+
+      new lambda.Function(stack, 'MyLambda', {
+        code: new lambda.InlineCode('foo'),
+        handler: 'index.handler',
+        runtime: lambda.Runtime.NODEJS_LATEST,
+        allowAllIpv6Outbound: true,
+        vpc: vpc,
```
pahud, line 3665:
```suggestion
        vpc,
```

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3649,6 +3649,83 @@ describe('function', () => {
...
+        runtime: lambda.Runtime.NODEJS_LATEST,
+        allowAllIpv6Outbound: true,
+      })).toThrow(/Cannot configure \'allowAllIpv6Outbound\' without configuring a VPC/);
+    });
+
+    test('throws when both allowAllIpv6Outbound and securityGroup are defined', () => {
+      const stack = new cdk.Stack();
+      const vpc = new ec2.Vpc(stack, 'Vpc', {
+        maxAzs: 2,
+      });
```
pahud, line 3699:
```suggestion
      const vpc = new ec2.Vpc(stack, 'Vpc');
```

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3649,6 +3649,83 @@ describe('function', () => {
...
+      const securityGroup = new ec2.SecurityGroup(stack, 'SecurityGroup', { vpc: vpc });
+
+      expect(() => new lambda.Function(stack, 'MyLambda', {
+        code: new lambda.InlineCode('foo'),
+        handler: 'index.handler',
+        runtime: lambda.Runtime.NODEJS_LATEST,
+        allowAllIpv6Outbound: true,
+        vpc: vpc,
```
pahud, line 3707:
```suggestion
        vpc,
```

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3649,6 +3649,83 @@ describe('function', () => {
...
+        securityGroup: securityGroup,
+      })).toThrow(/Configure \'allowAllIpv6Outbound\' directly on the supplied SecurityGroup./);
+    });
+
+    test('throws when both allowAllIpv6Outbound and securityGroups are defined', () => {
+      const stack = new cdk.Stack();
+      const vpc = new ec2.Vpc(stack, 'Vpc', {
+        maxAzs: 2,
+      });
```
pahud, line 3716:
```suggestion
      const vpc = new ec2.Vpc(stack, 'Vpc');
```

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -3649,6 +3649,83 @@ describe('function', () => {
...
+      const securityGroup = new ec2.SecurityGroup(stack, 'SecurityGroup', { vpc: vpc });
+
+      expect(() => new lambda.Function(stack, 'MyLambda', {
+        code: new lambda.InlineCode('foo'),
+        handler: 'index.handler',
+        runtime: lambda.Runtime.NODEJS_LATEST,
+        allowAllIpv6Outbound: true,
+        vpc: vpc,
```
pahud, line 3724:
```suggestion
        vpc,
```

pahud (dismissed):
we are not really deploying vpc in unit tests so I prefer to keep the code as simple as possible

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
ashishdhingra: Done

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
ashishdhingra: Done

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
ashishdhingra: Done

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
ashishdhingra: Done

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
ashishdhingra: Done

packages/aws-cdk-lib/aws-lambda/README.md
ashishdhingra: Done

ashishdhingra:
> > Question - what happens if `allowAllOutbound: false` with `allowAllIpv6Outbound: true`? Is it allowed?
> 
> @pahud I will test it tomorrow with `ec2.SecurityGroup`. As per my understanding, it would not add outbound rule for `IPv4`.

@pahud Using the code `new ec2.SecurityGroup(this, 'mySecurityGroup', { vpc, allowAllOutbound: false, allowAllIpv6Outbound: true});` results below in the CloudFormation template:
```YAML
  mySecurityGroup6B1044D0:
    Type: AWS::EC2::SecurityGroup
    Properties:
      GroupDescription: CdktestStack/mySecurityGroup
      SecurityGroupEgress:
        - CidrIp: 255.255.255.255/32
          Description: Disallow all traffic
          FromPort: 252
          IpProtocol: icmp
          ToPort: 86
        - CidrIpv6: ::/0
          Description: Allow all outbound ipv6 traffic by default
          IpProtocol: "-1"
      VpcId:
        Ref: myVpc3CC7CF9E
    Metadata:
      aws:cdk:path: CdktestStack/mySecurityGroup/Resource
```
Below is deployed Security Group:
Screenshot 2024-08-06 at 10 22 57 AM

Thanks,
Ashish

packages/aws-cdk-lib/aws-lambda/README.md
```diff
@@ -1143,6 +1143,26 @@ const fn = new lambda.Function(this, 'Lambda_with_IPv6_VPC', {
 });
 ```
 
+## Outbound traffic
+By default, when creating a Lambda function, it would add a security group outbound rule to allow sending all network traffic (except IPv6). This is controlled by `allowAllOutbound` in function properties, which has a default value of `true`.
+
+To allow outbound IPv6 traffic by default, explicitly set `allowAllIpv6Outbound` to `true` in function properties as shown below (the default value for `allowAllIpv6Outbound` is `false`):
+```ts
```
lpizzinidev, line 1150:
```suggestion
## Outbound traffic

By default, when creating a Lambda function, it would add a security group outbound rule to allow sending all network traffic (except IPv6).
You can override the default behavior by setting the `allowAllOutbound` property to `false`.

To allow outbound IPv6 traffic by default, explicitly set the `allowAllIpv6Outbound` property to `true`, as shown below.
The default value for `allowAllIpv6Outbound` is `false`.

```ts
```

lpizzinidev (approved):
Thanks 👍

samson-keung (approved):

Issue #17526, opened 2021-11-16
(aws-lambda): Add validation to verify the filesystem configuration mountPath matches the AccessPoint path


gshpychka:
### What is the problem?

When attaching an EFS mount point to a lambda, there is only one correct value for the `mountPath`, and that is `/mnt/{access point root path}`. Any other value will result in an OS-level access denied exception. Since the root path can be accessed from the access point by CDK, why expose this parameter at all?

### Reproduction Steps

Create an EFS Access Point. Attach it to a Lambda with `mountPoint` set to anything other than `/mnt/{access point root path}`. Attempt to write anything to the filesystem from the lambda.

### What did you expect to happen?

File I/O operations work normally.

### What actually happened?

Access denied I/O errors if the mount path differs from the access point's root path.

### CDK CLI Version

1.132

### Node.js Version

16.11.1

### OS

Linux

### Language

Typescript, Python, .NET, Java, Go

### Other information

Either infer the mountPath from the access point, or specify clearly in the documentation that it has to follow this strict pattern and is not customizable.

nija-at:
@gshpychka - 

By `{access point root path}`, do you mean the value passed here - https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-efs.AccessPoint.html#path ?

Per the CFN docs, this is configurable - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-filesystemconfig.html#cfn-lambda-function-filesystemconfig-localmountpath.

I've not tested this myself, but if that's not the access the value is always fixed, we can add a validation that these paths are the same. We'll still need the property in order to support [imported EFS access points](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-efs.AccessPoint.html#static-fromwbraccesswbrpointwbrattributesscope-id-attrs).

gshpychka:
Yes, that's the one. From my testing, it has to match the root path or it doesn't work. Very confusing for sure.

gshpychka:
@nija-at what do you think about potential resolutions? Maybe this is a bug in lambda and should be dealt with upstream?

gshpychka:
Still relevant, I believe

PR #31019, opened 2024-08-04
fix(lambda): validate localMountPath format and length

exoego:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/17526

### Reason for this change

Show user-friendly error message faster (before deployment)

### Description of changes

Added format and length validation according to [AWS::Lambda::Function FileSystemConfig document](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-filesystemconfig.html#cfn-lambda-function-filesystemconfig-localmountpath)

image

- Added unit test
- No integration test because I think it is overkill

### Checklist

----


exoego:
I think integration test is overkill for this PR

Rizxcviii:
Exemption request

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -2923,6 +2923,66 @@ describe('function', () => {
...
+      expect(() => {
+        new lambda.Function(stack, 'MyFunction', {
+          vpc,
+          handler: 'foo',
+          securityGroups: [securityGroup],
+          runtime: lambda.Runtime.NODEJS_LATEST,
+          code: lambda.Code.fromAsset(path.join(__dirname, 'handler.zip')),
+          filesystem: lambda.FileSystem.fromEfsAccessPoint(accessPoint, '/not-mnt/foo-bar'),
+        });
+      }).toThrow();
```
Rizxcviii, line 2953: bit of a nick pick, but do you think it would be a good idea to add the actual error message? We could check that the correct error is being thrown out, rather than just an error being thrown?

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
```diff
@@ -2923,6 +2923,66 @@ describe('function', () => {
...
+      expect(() => {
+        new lambda.Function(stack, 'MyFunction', {
+          vpc,
+          handler: 'foo',
+          securityGroups: [securityGroup],
+          runtime: lambda.Runtime.NODEJS_LATEST,
+          code: lambda.Code.fromAsset(path.join(__dirname, 'handler.zip')),
+          filesystem: lambda.FileSystem.fromEfsAccessPoint(accessPoint, `/mnt/${'a'.repeat(160)}`),
+        });
+      }).toThrow();
```
Rizxcviii, line 2983: Same here, would be a good idea to test against the actual error message.

Rizxcviii (changes requested):

packages/aws-cdk-lib/aws-lambda/test/function.test.ts
exoego: Fixed in https://github.com/aws/aws-cdk/pull/31019/commits/0e9cb467377fccf3f3b944a36dca22f07446e7f8

Rizxcviii (approved):

packages/aws-cdk-lib/aws-lambda/lib/function.ts
```diff
@@ -913,6 +913,12 @@ export class Function extends FunctionBase {
     // add additional managed policies when necessary
     if (props.filesystem) {
       const config = props.filesystem.config;
+      if (!/^\/mnt\/[a-zA-Z0-9-_.]+$/.test(config.localMountPath)) {
+        throw new Error(`Local mount path should match with ^/mnt/[a-zA-Z0-9-_.]+$ but given ${config.localMountPath}.`);
+      }
+      if (config.localMountPath.length > 160) {
+        throw new Error(`Local mount path can not be longer than 160 characters but has ${config.localMountPath.length} characters.`);
+      }
```
moelasmar, line 921: You need to check if the `LocalMountPath` is a resolved token or not (like a CFN Parameter that will be only available during deployment time)

moelasmar (changes requested):

packages/aws-cdk-lib/aws-lambda/lib/function.ts
exoego: OK, but how?
I am not familiar with CDK internal

exoego: `Token.isUnresolved(config.localMountPath)`?

exoego: Changed https://github.com/aws/aws-cdk/pull/31019/commits/bf7911dc4d3bfe1b0b50a91b75aa91cff850e3d2

moelasmar (approved):

moelasmar (approved):

PR #31028, opened 2024-08-05
feat(codepipeline): Add `CodeConnections` source for creating codepipeline

### LINKED ISSUE #30957: codepipeline_actions: No way to add CodeConnection source
### Describe the feature

I'm trying to set up CodePipeline with a GitHub source. I think this is a common flow.
I see that there is a `GitHubSourceAction` construct but it's not what I want - it wants an oauth secret. But now in the console we can go set up a "CodeConnection" to link GItHub to AWS without having to store an oauth token somewhere in secrets manager. I'd like to use that existing connection someone has created.

I found the L1 construct I want, but have no way to use it as a Source Action in CodePipeline: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_codeconnections.CfnConnection.html

Please add support for a CodeConnections Source Action. 

### Use Case

I think my use case is pretty common: using GitHub with CodePipeline.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.150.0

Linux

ishon19:
### Issue 

Closes #30957

### Reason for this change

This change allows the users to create the codepipeline using `CodeConnections` as a source.

### Description of changes
This involves addition of a new source option in the `codepipeline-source` and addition of a new action `CodeConnectionsSourceAction` for triggering the creation.

🚧  WIP

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-codepipeline-actions/test/codepipeline-actions/codepipeline-source-actions.test.ts
+++ b/packages/aws-cdk-lib/aws-codepipeline-actions/test/codepipeline-actions/codepipeline-source-actions.test.ts
@@ -1,0 +1,113 @@
+import { Stack } from 'aws-cdk-lib';
+import { Artifact } from 'aws-cdk-lib/aws-codepipeline';
+import * as iam from 'aws-cdk-lib/aws-iam';
+import * as s3 from 'aws-cdk-lib/aws-s3';
+import { CodeConnectionsSourceAction, CodeConnectionsSourceActionProps } from '../../lib';
+
+describe('CodeConnectionsSourceAction', () => {
+  test('requires connectionArn', () => {
+    const stack = new Stack();
+
+    expect(() => {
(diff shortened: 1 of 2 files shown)


TheRealAmazonKendra:
Creating a duplicate just because of the name change is not something that we want to do here. We can update the doc string and README to clarify that the name has been changed.

ishon19:
> Creating a duplicate just because of the name change is not something that we want to do here. We can update the doc string and README to clarify that the name has been changed.

@TheRealAmazonKendra thanks a lot for the review! I agree, I also don't like the way it is right now, but based on the [issue](https://github.com/aws/aws-cdk/issues/30957), my understanding was to add a new source option, which in our case happens to be quite similar to the existing one. 

Could you please confirm if the intention was to simply update the documentation and README? I can proceed accordingly based on your suggestions. Thanks again :)

PR #31041, opened 2024-08-06
feat: set log retention of all CDK-vended custom resources with CustomResourceConfig

awslukeguan:
### Reason for this change

CDK vended custom resources may or may not have Cloudwatch logGroup set with a retention period. 
Modify all CDK-vended custom resources with `CustomResourceConfig` and to modify retention period with `addLogRetentionLifetime` method.

`CustomResourceConfig.of(app).addLogRetentionLifetime(logs.RetentionDays.TEN_YEARS);`

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
+++ b/packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
@@ -118,7 +118,6 @@ describe('when a singleton-backed custom resource log group is specified', () =>
   });
 });
 
-
 test('addLogRetentionLifetime modifies the retention period of the custom resources in two top-level stacks', () => {
   // GIVEN
   const customResourceLogRetention = logs.RetentionDays.TEN_YEARS;


packages/aws-cdk-lib/custom-resource-config/test/custom-resource-config.test.ts
```diff
@@ -0,0 +1,547 @@
...
+    test('catch thrown error if no retention', ()=> {
+        const app = new cdk.App();
+        try {
+            CustomResourceConfig.of(app).addLogRetentionLifetime()
+        } catch (e) {
+            expect(e.message).toBe("Retention must have a value");
+        }
+    })
+})
+describe('aws-ses drop spams', () => {
```
comcalvi, line 22: Name these according to the use-case of the feature, instead of according to implementation or testing details. 

One way to do it is a `describe` block for each of the three cases; one for when logRetention is set, one for when loggingConfig is set, and one for when both are undefined.

packages/aws-cdk-lib/custom-resource-config/test/custom-resource-config.test.ts
```diff
@@ -0,0 +1,547 @@
...
+    test('Drop spams - UnModified by Aspects', () => {
```
comcalvi, line 23: This test string gives no information about the feature. The use of Aspects is not relevant to the tests; test cases should describe behavior, not implementation details. 

In this case, the behavior is "When `addLogRetentionLifetime` is not called, no LogGroup is created". 

You don't need this case; every test we have for `ReceiptRuleSet` is already confirming that no LogGroup is created when you don't use `addLogRetentionLifetime`.

packages/aws-cdk-lib/custom-resource-config/test/custom-resource-config.test.ts
```diff
@@ -0,0 +1,547 @@
...
+        CustomResourceConfig.of(app).addLogRetentionLifetime(SET_LOG_RETENTION);
+        const stack = new cdk.Stack(app);
+
+        new ses.ReceiptRuleSet(stack, 'RuleSet', {
+            dropSpam: true,
+        });
+
+        const template = Template.fromStack(stack);
+
+        expect(template.toJSON()).toMatchSnapshot();
```
comcalvi, line 73: the snapshot is empty?

packages/aws-cdk-lib/custom-resource-config/test/custom-resource-config.test.ts
```diff
@@ -0,0 +1,547 @@
...
+    })
+})
+describe('aws-events-targets awsapi', () => {
```
comcalvi, line 76: I appreciate the pains you've taken to do this for all these different custom resources, but this doesn't add much to the testing.

You only need to test the different cases that impact your implementation. In this case, there's only two types of custom resources that matter; those backed by Singleton lambdas and those backed by non-Singleton lambdas. 

Choose one of each and use them throughout the entire test file. In the test string, don't refer to them by name; eg, there's nothing gained by saying "S3Deployment lambda log retention set to blah". Instead, say, "singleton-backed custom resource logging configuration set to blah", or something along those lines. That communicates the case that you care about, rather than the arbitrary choice of custom resource.

packages/aws-cdk-lib/custom-resource-config/test/custom-resource-config.test.ts
```diff
@@ -0,0 +1,547 @@
...
+            LoggingConfig: {
+                LogGroup: {
+                    Ref: Match.stringLikeRegexp('LogGroup')
+                }
+            }
+        })
+    });
+
+    test('s3LogGroup - update lambda logGroup to SET_LOG_RETENTION', () => {
+        const SET_LOG_RETENTION = logs.RetentionDays.TEN_YEARS;
```
comcalvi, line 240: Most constant variables should not be all caps. The only constants that we use this naming convention for are variables that hold some magic value we don't want to typo, or that would otherwise be difficult to read.

packages/aws-cdk-lib/custom-resource-config/test/custom-resource-config.test.ts
```diff
@@ -0,0 +1,547 @@
...
+        template.hasResourceProperties('AWS::Lambda::Function', {
+            LoggingConfig: {
+                LogGroup: {
+                    Ref: Match.stringLikeRegexp('LogGroup')
+                }
+            }
+        })
+    });
+
+    test('s3LogGroup - update lambda logGroup to SET_LOG_RETENTION', () => {
```
comcalvi, line 239: Try something like:

Modify a custom resource's log group.

packages/aws-cdk-lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,73 @@
...
+    visit(node: IConstruct) {
+        for (const metaEntry of node.node.metadata as MetadataEntry[]) {
+            // console.log("this is ", this.SET_LOG_RETENTION);
+            if (metaEntry.type == 'aws:cdk:is-custom-resource-handler-logGroup') {
+                console.log("Found a marked logGroup", node.node.path); //
+                const localNode = node.node.defaultChild as logs.CfnLogGroup;
+                // console.log("Modifying logGroup to RetentionInDays"); //
+                localNode.addPropertyOverride("RetentionInDays", this.SET_LOG_RETENTION);
+            }
+            if (metaEntry.type == 'aws:cdk:is-custom-resource-handler-singleton') {
```
comcalvi, line 48: Don't use these `aws:cdk:...` things everywhere. Define a constant, like `CUSTOM_RESOURCE_SINGLETON_MARKER` and use that

packages/aws-cdk-lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,73 @@
...
+                // console.log("Modifying the RetentionInDays"); //
+                node.addPropertyOverride("RetentionInDays", this.SET_LOG_RETENTION);
+            }
+        }
+    }
+
+    createLogGroup(scope: lambda.CfnFunction): string {
+        return new logs.LogGroup(scope, 'logGroup', {
+            retention: this.SET_LOG_RETENTION
+        }).logGroupName
```
comcalvi, line 71: This method name is misleading. A `createLogGroup` should return an `ILogGroup`. There's no harm in making th the caller use `.logGroupName`

packages/aws-cdk-lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
```
comcalvi, line 73:
```suggestion

}
```

packages/aws-cdk-lib/custom-resource-config/test/custom-resource-config.test.ts
```diff
@@ -0,0 +1,547 @@
...
+        });
+        template.resourceCountIs("AWS::Logs::LogGroup", 2);
+        template.hasResourceProperties("AWS::Logs::LogGroup", {
+            RetentionInDays: SET_LOG_RETENTION,
+        });
+        template.hasResourceProperties("AWS::Logs::LogGroup", {
+            RetentionInDays: DEFAULT_LOG_GROUP_RETENTION,
+        });
+    });
+    test("s3None - snapshot", () => {
```
comcalvi, line 530: What's the purpose of these snapshot tests?

packages/aws-cdk-lib/custom-resource-config/test/custom-resource-config.test.ts
```diff
@@ -0,0 +1,547 @@
...
+        new s3deploy.BucketDeployment(stack, "s3deployNone", {
+            sources: [s3deploy.Source.jsonData("file.json", { a: "b" })],
+            destinationBucket: websiteBucket,
+        });
+
+        const template = Template.fromStack(stack);
+
+        expect(template.toJSON()).toMatchSnapshot();
+    });
+});
```
comcalvi, line 547:
```suggestion
});

```

packages/aws-cdk-lib/custom-resource-config/test/custom-resource-config.test.ts
```diff
@@ -0,0 +1,547 @@
...
+                    Ref: Match.stringLikeRegexp("CustomCDKBucketDeployment"),
+                },
+            },
+        });
+        template.resourceCountIs("AWS::Logs::LogGroup", 1);
+        template.hasResourceProperties("AWS::Logs::LogGroup", {
+            RetentionInDays: SET_LOG_RETENTION,
+        });
+    });
+    test("s3None - assert that extra logGroup is unchanged", () => {
```
comcalvi, line 498: This description is relying on context that the reader lacks; what is `s3None`,  why is there an extra log group, and what operation is supposed to leave it unchanged? 

A more informative description would be:

"CustomResourceConfig only affects Log Groups that support custom resources".

packages/aws-cdk-lib/custom-resource-config/test/custom-resource-config.test.ts
```diff
@@ -0,0 +1,547 @@
...
+        });
+        rule.addTarget(awsApi);
+
+        const template = Template.fromStack(stack);
+
+        expect(template.toJSON()).toMatchSnapshot();
+    })
+})
+
+describe('LogGroup', () => {
```
comcalvi, line 208: `LogGroup` does not provide enough information. Are we modifying existing Log Groups? Deleting them? Creating them? Something else entirely?

comcalvi (dismissed):
This is a great start! I have a few comments though 🙂 

Please add an integration test (in `@aws-cdk-testing`), move this code to `aws-cdk-lib/custom-resources` and update that module's README to tell users how to use the new feature.

comcalvi:
Also please update the title to be a `feat`, since this is a feature. The feature should describe the new feature you're adding; for example, "Set log retention of all CDK-vended custom resources with `CustomResourceConfig`"

packages/@aws-cdk-testing/framework-integ/test/custom-resources/test/custom-resource-config/integ.custom-resource-config-logGroup.ts
```diff
@@ -0,0 +1,38 @@
+import * as s3 from 'aws-cdk-lib/aws-s3';
+import * as logs from 'aws-cdk-lib/aws-logs';
+import * as cdk from 'aws-cdk-lib';
+import * as integ from '@aws-cdk/integ-tests-alpha';
+import { Construct } from 'constructs';
+import * as s3deploy from 'aws-cdk-lib/aws-s3-deployment';
+import { CustomResourceConfig } from 'aws-cdk-lib/custom-resources';
+
+class TestCustomResourceConfig extends cdk.Stack {
```
comcalvi, line 9: Don't make a new class here. Integration tests follow this convention:

`const stack = new cdk.Stack(this, 'Stack', ...);`

Then you create your resources in the scope of the stack, with

`new s3deploy.BucketDeployment(stack, 's3Deploy', {...});`

packages/@aws-cdk-testing/framework-integ/test/custom-resources/test/custom-resource-config/integ.custom-resource-config-logRetention.ts
```diff
@@ -0,0 +1,36 @@
...
+const set_log_retention = logs.RetentionDays.TEN_YEARS;
+CustomResourceConfig.of(app).addLogRetentionLifetime(set_log_retention);
+const testCase = new TestCustomResourceConfig(app, 'test-custom-resource-config-logRetention');
+
+new integ.IntegTest(app, 'integ-test-custom-resource-config-logRetention', {
+  testCases: [testCase],
+  diffAssets: false,
+});
+
+app.synth();
```
comcalvi, line 36: you don't need the `app.synth()` calls here. This is a leftover requirement from old-style integration tests, which you do not need when you use `IntegTest`.

packages/@aws-cdk-testing/framework-integ/test/custom-resources/test/custom-resource-config/integ.custom-resource-config-logRetention.ts
```diff
@@ -0,0 +1,36 @@
...
+      sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+      destinationBucket: websiteBucket,
+      logRetention: locally_set_log_retention,
+    });
+
+  }
+}
+
+const app = new cdk.App();
+const set_log_retention = logs.RetentionDays.TEN_YEARS;
```
comcalvi, line 27:
```suggestion
const logRetentionDays = logs.RetentionDays.TEN_YEARS;
```

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -359,6 +384,11 @@ export abstract class HandlerFrameworkClass extends ClassType {
 
     const superInitializerArgs: Expression[] = [scope, id, props.superProps];
     init.addBody(new SuperInitializer(...superInitializerArgs));
+    if (props.statements){
+      for(const statement of props.statements){
```
comcalvi, line 388:
```suggestion
      for (const statement of props.statements){
```

packages/@aws-cdk-testing/framework-integ/test/custom-resources/test/custom-resource-config/integ.custom-resource-config-undefined-log.ts
```diff
@@ -0,0 +1,33 @@
...
+  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    let websiteBucket = new s3.Bucket(this, 'WebsiteBucket', {});
+
+    new s3deploy.BucketDeployment(this, 's3deployNone', {
+      sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+      destinationBucket: websiteBucket,
+    });
+
```
comcalvi, line 19:
```suggestion
```

packages/@aws-cdk-testing/framework-integ/test/custom-resources/test/custom-resource-config/integ.custom-resource-config-undefined-log.ts
```diff
@@ -0,0 +1,33 @@
...
+const set_log_retention = logs.RetentionDays.TEN_YEARS;
+CustomResourceConfig.of(app).addLogRetentionLifetime(set_log_retention);
+const testCase = new TestCustomResourceConfig(app, 'test-custom-resource-config-undefined-log');
+
+new integ.IntegTest(app, 'integ-test-custom-resource-config-undefined-log', {
+  testCases: [testCase],
+  diffAssets: false,
+});
+
+app.synth();
```
comcalvi, line 33:
```suggestion
```

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -56,6 +63,13 @@ interface ConstructorBuildProps {
    * @default MemberVisbility.Public
    */
   readonly constructorVisbility?: MemberVisibility;
+
+    /**
```
comcalvi, line 67:
```suggestion
  /**
```

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -56,6 +63,13 @@ interface ConstructorBuildProps {
...
+   * Statement. Additional statements to constructor are optional.
```
comcalvi, line 68:
```suggestion
   * These statements are added to the constructor body in the order they appear in this property. 
```

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -206,10 +220,17 @@ export abstract class HandlerFrameworkClass extends ClassType {
           ['handler', expr.lit(props.handler)],
           ['runtime', this.buildRuntimeProperty(scope, { runtime: props.runtime, isEvalNodejsProvider })],
         ]);
+        const metadataStatements: Statement[] = [
+          expr.directCode(`this.addMetadata('${CUSTOM_RESOURCE_SINGLETON}', true)`),
+          expr.directCode(`if (props?.logGroup) { this.logGroup.node.addMetadata('${CUSTOM_RESOURCE_SINGLETON_LOG_GROUP}', true) }`),
+          // (this singleton function) as it extends FunctionBase and property lambdaFunction is available to be cast to lambda.Function to access the internal readonly _logRetention
+          expr.directCode(`if (props?.logRetention) { ((this as any).lambdaFunction as lambda.Function)._logRetention?.node.addMetadata('${CUSTOM_RESOURCE_SINGLETON_LOG_RETENTION}', true) }`),
```
comcalvi, line 227: I realized we're missing a case here. Look at this code: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L1331 and https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-lambda/lib/singleton-lambda.ts#L104.

What we're missing is that a user can go

`const func = new BucketDeploymentSingletonFunction(...)`

`func.logGroup // <-- this call creates a LogRetention custom resource!`

can you check for the `_logGroup` property being set on the lambda, and if it is, add the metadata like you do in the `if (props?.logRetention)` case?

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -310,11 +331,15 @@ export abstract class HandlerFrameworkClass extends ClassType {
             isCustomResourceProvider: true,
           })],
         ]);
+        const metadataStatements: Statement[] = [
+          expr.directCode(`this.node.addMetadata('${CUSTOM_RESOURCE_PROVIDER}', true)`),
+        ]
```
comcalvi, line 336:
```suggestion
        const metadataStatements: Statement[] = [expr.directCode(`this.node.addMetadata('${CUSTOM_RESOURCE_PROVIDER}', true)`)];
```

comcalvi (changes requested):

packages/@aws-cdk-testing/framework-integ/test/custom-resources/test/custom-resource-config/integ.custom-resource-config-logRetention.ts
```diff
@@ -0,0 +1,27 @@
...
+const app = new cdk.App();
+const stack = new cdk.Stack(app, 'MyStack');
+const locallySetLogRetentionDays = logs.RetentionDays.ONE_WEEK;
+
+let websiteBucket = new s3.Bucket(stack, 'WebsiteBucket', {});
+
+new s3deploy.BucketDeployment(stack, 's3deployNone', {
+  sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+  destinationBucket: websiteBucket,
+  logRetention: locallySetLogRetentionDays,
```
comcalvi, line 17:
```suggestion
  logRetention: logs.RetentionDays.ONE_WEEK,
```

packages/@aws-cdk-testing/framework-integ/test/custom-resources/test/custom-resource-config/integ.custom-resource-config-logRetention.ts
```diff
@@ -0,0 +1,27 @@
+import * as s3 from 'aws-cdk-lib/aws-s3';
+import * as logs from 'aws-cdk-lib/aws-logs';
+import * as cdk from 'aws-cdk-lib';
+import * as integ from '@aws-cdk/integ-tests-alpha';
+import * as s3deploy from 'aws-cdk-lib/aws-s3-deployment';
+import { CustomResourceConfig } from 'aws-cdk-lib/custom-resources';
+
+const app = new cdk.App();
+const stack = new cdk.Stack(app, 'MyStack');
+const locallySetLogRetentionDays = logs.RetentionDays.ONE_WEEK;
```
comcalvi, line 10:
```suggestion
```

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
```diff
@@ -206,10 +220,17 @@ export abstract class HandlerFrameworkClass extends ClassType {
           ['handler', expr.lit(props.handler)],
           ['runtime', this.buildRuntimeProperty(scope, { runtime: props.runtime, isEvalNodejsProvider })],
         ]);
+        const metadataStatements: Statement[] = [
+          expr.directCode(`this.addMetadata('${CUSTOM_RESOURCE_SINGLETON}', true)`),
+          expr.directCode(`if (props?.logGroup) { this.logGroup.node.addMetadata('${CUSTOM_RESOURCE_SINGLETON_LOG_GROUP}', true) }`),
+          // (this singleton function) as it extends FunctionBase and property lambdaFunction is available to be cast to lambda.Function to access the internal readonly _logRetention
```
comcalvi, line 226:
```suggestion
          // We need to access the private `_logRetention` custom resource, the only public property - `logGroup` - provides an ARN reference to the resource, instead of the resource itself. 
```

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -824,3 +824,91 @@ new cr.AwsCustomResource(this, 'CrossAccount', {
...
+
+    const stack = new cdk.Stack(app);
+
+    let websiteBucket = new s3.Bucket(stack, 'WebsiteBucket', {});
+    new s3deploy.BucketDeployment(stack, 's3deployNone', {
+      sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+      destinationBucket: websiteBucket,
+    });
+```
+The following example shows configuring resources of nested stacks to retain its logs for ten years:
```
comcalvi, line 846:
```suggestion

The following example configures every custom resource in a nested stack to retain its log for ten years:
```

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -824,3 +824,91 @@ new cr.AwsCustomResource(this, 'CrossAccount', {
...
+    }
+}
+export class AlphaStack extends cdk.NestedStack {
+    constructor(scope: Construct) {
+        super(scope, "AlphaStack");
+        let websiteBucket = new s3.Bucket(this, "WebsiteBucket", {});
+        new s3deploy.BucketDeployment(this, "s3deployNone", {
+            sources: [s3deploy.Source.jsonData("file.json", { a: "b" })],
+            destinationBucket: websiteBucket,
+            logRetention: logs.RetentionDays.ONE_DAY,
```
comcalvi, line 863:
```suggestion
            logRetention: logs.RetentionDays.ONE_DAY, // overridden by the `TEN_YEARS` set by `CustomResourceConfig`. 
```

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -824,3 +824,91 @@ new cr.AwsCustomResource(this, 'CrossAccount', {
...
+    }
+}
+export class BetaStack extends cdk.NestedStack {
+    constructor(scope: Construct) {
+        super(scope, "BetaStack");
+        let websiteBucket = new s3.Bucket(this, "WebsiteBucket", {});
+        new s3deploy.BucketDeployment(this, "s3deployNone", {
+            sources: [s3deploy.Source.jsonData("file.json", { a: "b" })],
+            destinationBucket: websiteBucket,
+            logRetention: logs.RetentionDays.ONE_DAY,
```
comcalvi, line 874:
```suggestion
            logRetention: logs.RetentionDays.ONE_DAY, // overridden by the `TEN_YEARS` set by `CustomResourceConfig`.
```

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -824,3 +824,91 @@ new cr.AwsCustomResource(this, 'CrossAccount', {
...
+    constructor(scope: Construct) {
+        super(scope, "rootStack");
+        new AlphaStack(this);
+        new BetaStack(this);
+        CustomResourceConfig.of(this).addLogRetentionLifetime(
+            logs.RetentionDays.TEN_YEARS
+        );
+    }
+}
+export class AlphaStack extends cdk.Stack {
```
comcalvi, line 892: `AlphaStack` and `BetaStack` are not top-level stacks because they're in the scope of `RootStack`. To be top-level, they have to be direct children of an `App`.

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,79 @@
...
+export const CUSTOM_RESOURCE_SINGLETON = 'aws:cdk:is-custom-resource-handler-singleton';
+export const CUSTOM_RESOURCE_SINGLETON_LOG_GROUP = 'aws:cdk:is-custom-resource-handler-logGroup';
+export const CUSTOM_RESOURCE_SINGLETON_LOG_RETENTION = 'aws:cdk:is-custom-resource-handler-logRetention';
+
+/**
+ * Manages AWS vended Custom Resources
+ */
+export class CustomResourceConfig {
+  /**
+   * Returns the tags API for this scope.
```
comcalvi, line 17:
```suggestion
   * Returns the CustomResourceConfig for this scope.
```

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,79 @@
...
+   * @param scope The scope
+   */
+  public static of(scope: IConstruct): CustomResourceConfig {
+    return new CustomResourceConfig(scope);
```
comcalvi, line 21: a user shouldn't be able to add 2 of these on the same scope. Can you see if it's feasible to throw an error if they've already specified a `CustomResourceConfig` on this scope? `Aspects.of(this.scope).all` should give you all the existing aspects.

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,79 @@
...
+  public addLogRetentionLifetime(rentention: logs.RetentionDays) {
+    Aspects.of(this.scope).add(new CustomResourceLogRetention(rentention));
+  }
+}
+
+/**
+ * Manages log retention for AWS vended custom resources within a construct scope.
+ */
+export class CustomResourceLogRetention implements IAspect {
+  private readonly SET_LOG_RETENTION: logs.RetentionDays;
```
comcalvi, line 38: These aren't magic values, so don't use the ALL_CAPS convention. 

```suggestion
  private readonly logRetention: logs.RetentionDays;
```

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  constructor(setLogRetention: logs.RetentionDays) {
+    this.SET_LOG_RETENTION = setLogRetention;
+  }
```
comcalvi, line 42:
```suggestion

  constructor(logRetention: logs.RetentionDays) {
    this.logRetention = logRetention;
  }
```

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,79 @@
...
+        const localNode = node.node.defaultChild as logs.CfnLogGroup;
+        localNode.addPropertyOverride('RetentionInDays', this.SET_LOG_RETENTION);
+      }
+
+      if (metadataEntry.type == CUSTOM_RESOURCE_SINGLETON) {
+        const localNode = node.node.defaultChild as lambda.CfnFunction;
+
+        if (localNode && !localNode.loggingConfig) {
+          const newLogGroup = this.createNewLogGroupForSingletonFunction(localNode);
+          newLogGroup.node.addMetadata(`${CUSTOM_RESOURCE_SINGLETON_LOG_GROUP}`, true);
```
comcalvi, line 55: Why do we add metadata here? I understand that this is logically a log group that should be targeted, but have you tested that this causes the Aspect to run one more time than if it wasn't there? 

Either way move the `addMetadata` to be inside `createNewLogGroup`

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,79 @@
...
+      if (metadataEntry.type == CUSTOM_RESOURCE_SINGLETON_LOG_GROUP) {
+        const localNode = node.node.defaultChild as logs.CfnLogGroup;
+        localNode.addPropertyOverride('RetentionInDays', this.SET_LOG_RETENTION);
+      }
+
+      if (metadataEntry.type == CUSTOM_RESOURCE_SINGLETON) {
+        const localNode = node.node.defaultChild as lambda.CfnFunction;
+
+        if (localNode && !localNode.loggingConfig) {
+          const newLogGroup = this.createNewLogGroupForSingletonFunction(localNode);
```
comcalvi, line 54:
```suggestion
          const newLogGroup = this.createLogGroup(localNode);
```

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    template.hasResourceProperties('AWS::Lambda::Function', {
+      LoggingConfig: {
+        LogGroup: {
+          Ref: Match.stringLikeRegexp('CustomCDKBucketDeployment'),
+        },
+      },
+    });
+    template.resourceCountIs('AWS::Logs::LogGroup', 1);
+    template.hasResourceProperties('AWS::Logs::LogGroup', {
+      RetentionInDays: setLogRetention,
```
comcalvi, line 33:
```suggestion
      RetentionInDays: logRetention,
```

you don't need to prefix the variables with `set`

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    });
+  });
+
+  test('when CustomResourceConfig.addLogRetentionLifetime is called with extra logGroup, extra logGroup remain unmodified by CustomResourceConfig', () => {
```
comcalvi, line 37:
```suggestion
  test('addLogRetentionLifetime only modifies custom resource log groups', () => {
```

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    const setLogRetention = logs.RetentionDays.TEN_YEARS;
```
comcalvi, line 38:
```suggestion
    const customResourceLogRetention = logs.RetentionDays.TEN_YEARS;
```

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    const default_log_group_retention = logs.RetentionDays.TWO_YEARS;
```
comcalvi, line 39:
```suggestion
    const nonCustomResourceLogRetention = logs.RetentionDays.TWO_YEARS;
```

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,158 @@
...
+      RetentionInDays: setLogRetention,
+    });
+    template.hasResourceProperties('AWS::Logs::LogGroup', {
+      RetentionInDays: default_log_group_retention,
+    });
+  });
+});
+
+describe('when logRetention is specified', () => {
+  test('when CustomResourceConfig.addLogRetentionLifetime is called, singleton-backed custom resource logRetention period modified', () => {
```
comcalvi, line 64:
```suggestion
  test('addLogRetentionLifetime overrides log retention', () => {
```

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    const setLogRetention = logs.RetentionDays.TEN_YEARS;
```
comcalvi, line 65:
```suggestion
    const logRetentionOverride = logs.RetentionDays.TEN_YEARS;
```

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    new s3deploy.BucketDeployment(stack, 's3deployLogRetention', {
+      sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+      destinationBucket: websiteBucket,
+      logRetention: locallySetLogRetention,
+    });
+    const template = Template.fromStack(stack);
+
+    const captureRetentionInDays = new Capture();
+    template.hasResourceProperties('Custom::LogRetention', {
+      RetentionInDays: captureRetentionInDays,
```
comcalvi, line 81: `Capture` is a cool feature but you don't need to use it here. It's enough to say:

```suggestion
      RetentionInDays: logRetentionOverride,
```

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    });
+    expect(captureRetentionInDays.asNumber()).toEqual(setLogRetention);
+    expect(captureRetentionInDays.asNumber()).not.toEqual(
```
comcalvi, line 84: you don't need this assertion because `setLogRetention` and `locallySetLogRetention` are not equal

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    template.hasResourceProperties('AWS::Logs::LogGroup', {
+      RetentionInDays: default_log_group_retention,
+    });
+  });
+});
+
+describe('when logRetention is specified', () => {
+  test('when CustomResourceConfig.addLogRetentionLifetime is called, singleton-backed custom resource logRetention period modified', () => {
+    const setLogRetention = logs.RetentionDays.TEN_YEARS;
+    const locallySetLogRetention = logs.RetentionDays.ONE_WEEK;
```
comcalvi, line 66:
```suggestion
    const localLogRetention = logs.RetentionDays.ONE_WEEK;
```

comcalvi (changes requested):

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
awslukeguan: This was marking the new LogGroup for deletionPolicy to recognize it.

packages/@aws-cdk/custom-resource-handlers/lib/custom-resources-framework/classes.ts
awslukeguan: Not able to test BucketDeploymentSingletonFunction as it is not exposed to be tested.

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -824,3 +824,88 @@ new cr.AwsCustomResource(this, 'CrossAccount', {
   })]),
 });
 ```
+
+#### Custom Resource Config
+
+You can configure every CDK-vended custom resource in a given scope with `CustomResourceConfig`. 
+
+*Please understand that there is no guarantee the order in which Custom Resource Config is run. More than one Custom Resource Config of the same scope might result in conflict.
```
comcalvi, line 832:
```suggestion
Note that `CustomResourceConfig` uses Aspects to modify your constructs. There is no guarantee in the order in which Aspects modify the construct tree, which means that adding the same Aspect more than once to a given scope produces undefined behavior. This example guarantees that every affected resource will have a log retention of ten years or one day, but you cannot know which:

```
CustomResourceConfig.of(App).addLogRetentionLifetime(logs.RetentionDays.TEN_YEARS);
CustomResource.of(App).addLogRetentionLifetime(logs.RetentionDays.ONE_DAY);
```

File: packages/aws-cdk-lib/custom-resources/README.md [hunk 47]
```diff
@@ -824,3 +824,88 @@ new cr.AwsCustomResource(this, 'CrossAccount', {
...
+
+    let websiteBucket = new s3.Bucket(stack, 'WebsiteBucket', {});
+    new s3deploy.BucketDeployment(stack, 's3deployNone', {
+      sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+      destinationBucket: websiteBucket,
+    });
+```
+The following example configures every custom resource in a nested stack to retain its log for ten years:
+```ts
+export class RootStack extends cdk.Stack {
```
comcalvi, line 850: Can you avoid making new classes for these? This example never shows these classes being instantiated. You can save a bunch of space by doing:

```
const app = new App();
const stack = new Stack(app, 'Stack');
const nestedStack = new NestedStack(stack, 'NestedStack');

new s3deploy.BucketDeployment(...);

CustomResourceConfig.of(app).addLogRetentionLifetime(...);
```

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -824,3 +824,88 @@ new cr.AwsCustomResource(this, 'CrossAccount', {
...
+        new s3deploy.BucketDeployment(this, "s3deployNone", {
+            sources: [s3deploy.Source.jsonData("file.json", { a: "b" })],
+            destinationBucket: websiteBucket,
+            logRetention: logs.RetentionDays.ONE_DAY, // overridden by the `TEN_YEARS` set by `CustomResourceConfig`.
+        });
+    }
+}
+```
+
+The following example shows configuring resources of two top-level stacks to retain its logs for ten years:
```
comcalvi, line 882:
```suggestion
This also applies to nested stacks:
```

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -824,3 +824,88 @@ new cr.AwsCustomResource(this, 'CrossAccount', {
...
+
+    const stack = new cdk.Stack(app);
+
+    let websiteBucket = new s3.Bucket(stack, 'WebsiteBucket', {});
+    new s3deploy.BucketDeployment(stack, 's3deployNone', {
+      sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+      destinationBucket: websiteBucket,
+    });
+```
+The following example configures every custom resource in a nested stack to retain its log for ten years:
```
comcalvi, line 848:
```suggestion
`CustomResourceConfig` can modify resources across any scope. This example modifies every custom resource in two top-level stacks:
```

(move the top-level stacks example to be first, before the nested stack example) 

Can you also add unit tests for both of these cases?

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,150 @@
+import { Template, Match } from '../../../assertions';
+import * as logs from '../../../aws-logs';
+import * as s3 from '../../../aws-s3';
+import * as s3deploy from '../../../aws-s3-deployment';
+import * as cdk from '../../../core';
+import { CustomResourceConfig } from '../../lib/custom-resource-config/custom-resource-config';
+
+describe('when logging is undefined', () => {
+  test('when CustomResourceConfig.addLogRetentionLifetime is called, singleton-backed custom resource has new logGroup with retention period specified', () => {
```
comcalvi, line 9: This is a bit wordy; it's difficult to understand at a glance. 

```suggestion
  test('addLogRetentionLifetime creates a new log group with the correct retention period if one does not already exist', () => {
```

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+      destinationBucket: websiteBucket,
+    });
+
+    const template = Template.fromStack(stack);
+
+    template.hasResourceProperties('AWS::Lambda::Function', {
+      LoggingConfig: {
+        LogGroup: {
+          Ref: Match.stringLikeRegexp('CustomCDKBucketDeployment'),
```
comcalvi, line 26: can we avoid using `stringLikeRegexp`? The correct result here is not a string containing that thing, it exactly the string 'CustomCDKBucketDeployment'.  You should be able to say:

```
Ref: 'CustomCDKBucketDeployment',
```

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      logGroup: new logs.LogGroup(stack, 'LogGroup2', {
+        retention: logs.RetentionDays.ONE_WEEK,
+      }),
+    });
+
+    const template = Template.fromStack(stack);
+    template.hasResourceProperties('AWS::Lambda::Function', {
+      LoggingConfig: {
+        LogGroup: {
+          Ref: Match.stringLikeRegexp('LogGroup1'),
```
comcalvi, line 137: Same here

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    });
+    template.resourceCountIs('AWS::Logs::LogGroup', 1);
+    template.hasResourceProperties('AWS::Logs::LogGroup', {
+      RetentionInDays: customResourceLogRetention,
+    });
+  });
+});
+
+describe('when logGroup is specified', () => {
+  test('when CustomResourceConfig.addLogRetentionLifetime is called, singleton-backed custom resource logGroup period modified', () => {
```
comcalvi, line 88: "when <something> is called" adds 3 words you don't need. You can improve clarity by always preferring:

"<something> <did something, or performed some action>"

Eg:

```suggestion
  test('addLogRetentionLifetime modifies the retention period of a singleton-backed custom resource log group.', () => {
```

This applies to each test.

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      RetentionInDays: customResourceLogRetention,
+    });
+    template.resourceCountIs('AWS::Logs::LogGroup', 1);
+    template.hasResourceProperties('AWS::Logs::LogGroup', {
+      RetentionInDays: customResourceLogRetention,
+    });
+  });
+});
+
+describe('when logGroup is specified', () => {
```
comcalvi, line 87:
```suggestion
describe('when a singleton-backed custom resource log group already exists', () => {
```

comcalvi (dismissed):

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -824,3 +824,71 @@ new cr.AwsCustomResource(this, 'CrossAccount', {
...
+
+#### Custom Resource Config
+
+You can configure every CDK-vended custom resource in a given scope with `CustomResourceConfig`. 
+
+Note that `CustomResourceConfig` uses Aspects to modify your constructs. There is no guarantee in the order in which Aspects modify the construct tree, which means that adding the same Aspect more than once to a given scope produces undefined behavior. This example guarantees that every affected resource will have a log retention of ten years or one day, but you cannot know which:  
+CustomResourceConfig.of(App).addLogRetentionLifetime(logs.RetentionDays.TEN_YEARS);
+CustomResourceConfig.of(App).addLogRetentionLifetime(logs.RetentionDays.ONE_DAY);
+
+The following example configures every custom resource in this CDK app to retain its logs for ten years:
```
comcalvi, line 836:
```suggestion
The following example configures every custom resource in this CDK app to retain its logs for ten years:

```

packages/aws-cdk-lib/custom-resources/README.md
```diff
@@ -824,3 +824,71 @@ new cr.AwsCustomResource(this, 'CrossAccount', {
...
+CustomResourceConfig.of(app).addLogRetentionLifetime(logs.RetentionDays.TEN_YEARS);
+const stack = new Stack(app, 'Stack');
+
+let websiteBucket = new s3.Bucket(stack, 'WebsiteBucket', {});
+new s3deploy.BucketDeployment(stack, 's3deploy', {
+  sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+  destinationBucket: websiteBucket,
+});
+```
+The following example configures every custom resource in two top-level stacks to retain its log for ten years:
```
comcalvi, line 850:
```suggestion

The following example configures every custom resource in two top-level stacks to retain its log for ten years:
```

comcalvi (changes requested):

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,83 @@
...
+export const CUSTOM_RESOURCE_SINGLETON_LOG_GROUP = 'aws:cdk:is-custom-resource-handler-logGroup';
+export const CUSTOM_RESOURCE_SINGLETON_LOG_RETENTION = 'aws:cdk:is-custom-resource-handler-logRetention';
+
+/**
+ * Manages AWS vended Custom Resources
+ */
+export class CustomResourceConfig {
+  /**
+   * Returns the CustomResourceConfig for this scope.
+   * @param scope The scope
```
comcalvi, line 19:
```suggestion
```

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  public static of(scope: IConstruct): CustomResourceConfig {
+    return new CustomResourceConfig(scope);
+  }
+
+  private constructor(private readonly scope: IConstruct) { }
+
+  /**
+   * Set the log retention specified on AWS vended custom resources that have certain CDK metadata from the CDK library.
```
comcalvi, line 28:
```suggestion
   * Set the log retention of AWS-vended custom resource lambdas.
```

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  public addLogRetentionLifetime(rentention: logs.RetentionDays) {
+    Aspects.of(this.scope).add(new CustomResourceLogRetention(rentention));
+  }
+
+}
+
+/**
+ * Manages log retention for AWS vended custom resources within a construct scope.
```
comcalvi, line 37:
```suggestion
 * Manages log retention for AWS vended custom resources.
```

packages/aws-cdk-lib/custom-resources/lib/custom-resource-config/custom-resource-config.ts
```diff
@@ -0,0 +1,83 @@
...
+        let localNode = node.node.defaultChild as cloudformation.CfnCustomResource;
+        localNode.addPropertyOverride('RetentionInDays', this.logRetention);
+      }
+    }
+  }
+
+  /*
+   * Creates a new logGroup and associates with the singletonLambda
+   * Returns a Cloudwatch LogGroup
+   * @param {lambda.CfnFunction} scope - SingletonFunction
```
comcalvi, line 73:
```suggestion
```

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,230 @@
+import { Template } from '../../../assertions';
+import * as logs from '../../../aws-logs';
+import * as s3 from '../../../aws-s3';
+import * as s3deploy from '../../../aws-s3-deployment';
+import * as cdk from '../../../core';
+import { CustomResourceConfig } from '../../lib/custom-resource-config/custom-resource-config';
+
+describe('when a singleton-backed custom resource does not have logging defined', () => {
+  test('addLogRetentionLifetime creates a new log group with the correct retention period if one does not already exist', () => {
```
comcalvi, line 9: Organize these into:

```
// GIVEN
This is the setup for the test. For this test, this is the creation of the app, stack, the bucket, the `BucketDeployment`, and the `customResourceLogRetention` variable.

// WHEN
This is where the behavior or event you're interested in testing executes. In this case, it's a single event, `CustomResourceConfig.of()....`

// THEN
This is where your assertions are. So in this case, its the `Template.fromStack()` and everything that happens on `template`. 
```

This test file has some good examples of this: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/test/cdk-toolkit.test.ts#L256

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,230 @@
...
+  });
+
+  test('addLogRetentionLifetime only modifies custom resource log groups', () => {
+    const customResourceLogRetention = logs.RetentionDays.TEN_YEARS;
+    const nonCustomResourceLogRetention = logs.RetentionDays.TWO_YEARS;
+    const app = new cdk.App();
+    CustomResourceConfig.of(app).addLogRetentionLifetime(customResourceLogRetention);
+    const stack = new cdk.Stack(app);
+
+    const ignored = new logs.LogGroup(stack, 'ignored', {});
```
comcalvi, line 43: This `ignored` variable seems unused

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,230 @@
...
+
+    const websiteBucket = new s3.Bucket(stack, 'WebsiteBucket', {});
+    new s3deploy.BucketDeployment(stack, 's3deployLogGroup1', {
+      sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+      destinationBucket: websiteBucket,
+      logGroup: new logs.LogGroup(stack, 'LogGroup1', {
+        retention: logs.RetentionDays.ONE_WEEK,
+      }),
+    });
+    new s3deploy.BucketDeployment(stack, 's3deployLogGroup2', {
```
comcalvi, line 125: `s3deployLogGroup2` is not a good name for this construct, because it isn't a log group. Can we call it `BucketDeployment2`?

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,230 @@
...
+      sources: [s3deploy.Source.jsonData('file.json', { a: 'b' })],
+      destinationBucket: websiteBucket,
+      logGroup: new logs.LogGroup(stack, 'LogGroup2', {
+        retention: logs.RetentionDays.ONE_WEEK,
+      }),
+    });
+
+    const template = Template.fromStack(stack);
+    template.hasResourceProperties('AWS::Lambda::Function', {
```
comcalvi, line 134: This assertion doesn't add much value, can we remove it?

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
```diff
@@ -0,0 +1,230 @@
...
+      LoggingConfig: {
+        LogGroup: {
+          Ref: 'LogGroup106AAD846',
+        },
+      },
+    });
+    template.resourceCountIs('AWS::Logs::LogGroup', 2);
+    template.hasResourceProperties('AWS::Logs::LogGroup', {
+      RetentionInDays: customResourceLogRetention,
```
comcalvi, line 143: why is one log group retention supposed to be set, but not another? They're both bucket deployments, and they both have the explicitly created log group.

comcalvi (dismissed):

comcalvi:
This is very close to merging 🙂 we're almost there!

packages/aws-cdk-lib/custom-resources/test/custom-resource-config/custom-resource-config.test.ts
awslukeguan: One singleton (lambda) function from 2 BucketDeployment, that one singleton (lambda) only has one log group, the other log group is ignored as it's not attached to a singleton lambda.

comcalvi: This is an interesting edge case, but this doesn't feel like behavior we want to enforce, because it's confusing and not entirely desirable. Let's remove this test case.

comcalvi (dismissed):
Nice work! 🎉

comcalvi (dismissed):

comcalvi (approved):

PR #29794, opened 2024-04-11
fix(eks): incorrect nodegroupName(under feature flag)

pahud:
### Issue # (if applicable)

As https://github.com/aws/aws-cdk/pull/22910 and https://github.com/aws/aws-cdk/pull/25278 were closed for staleness. I am opening a new PR to fix https://github.com/aws/aws-cdk/issues/22442.

To avoid breaking change, this PR introduces a new feature flag for the bug fix.

Closes https://github.com/aws/aws-cdk/issues/22442

### The Code for verification

```ts
import {
  App, Stack,
  aws_eks as eks,
  aws_ec2 as ec2,
  CfnOutput, cx_api,
} from 'aws-cdk-lib';

const app = new App();
const env = { region: process.env.CDK_DEFAULT_REGION, account: process.env.CDK_DEFAULT_ACCOUNT };
const stack = new Stack(app, 'my-test-stack2', { env });

stack.node.setContext(cx_api.EKS_NODEGROUP_NAME, true);

const vpc = ec2.Vpc.fromLookup(stack, 'Vpc', { isDefault: true });

const cluster = new eks.Cluster(stack, 'Cluster', {
  vpc,
  version: eks.KubernetesVersion.V1_29,
  defaultCapacity: 0,
});

const ng1 = new eks.Nodegroup(stack, 'ng-1', {
  cluster,
});

const ng2 = cluster.addNodegroupCapacity('ng-2', {
  nodegroupName: 'custom-name',
});

new CfnOutput(stack, 'ng1-name', { value: ng1.nodegroupName });
new CfnOutput(stack, 'ng2-name', { value: ng2.nodegroupName });
```

The output is like

```
Outputs:
my-test-stack2.ng1name = ng19F1EA947-sSziNMzPVkik
my-test-stack2.ng2name = custom-name
```

Console:

image

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #31068, opened 2024-08-09
fix(rds): fixed the IAM policy that grantConnect() generates for DatabaseInstanceReadReplica

### LINKED ISSUE #31061: (aws-rds): grantConnect generates incorrect policy for DatabaseInstanceReadReplica
### Describe the bug

Calling `grantConnect` on an instance of `DatabaseInstanceReadReplica` generates an incorrect policy that uses the full ARN of the instance instead of the `instanceResourceId` value.

### Expected Behavior

```json
{
    "Action": "rds-db:connect",
    "Resource": "arn:aws:rds-db:us-east-1:1234567890:dbuser:db-INSTANCE_RESOURCE_ID/user",
    "Effect": "Allow"
}
```

### Current Behavior

```json
{
    "Action": "rds-db:connect",
    "Resource": "arn:aws:rds-db:us-east-1:1234567890:dbuser:arn:aws:rds:us-east-1:1234567890:db:instance-name-wq2y5qzlfdy6/user",
    "Effect": "Allow"
}
```

### Reproduction Steps

1. Create a read replica
2. Call grantConnect on it

### Possible Solution

_No response_

https://github.com/aws/aws-cdk/blob/abc78bfa613453185e59d5d9a17e5c5cfb5437b3/packages/aws-cdk-lib/aws-rds/lib/instance.ts#L201-L206

### CDK CLI Version

2.150.0

### Framework Version

2.150.0

### Node.js Version

v20.14.0

### OS

macOS

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

ashishdhingra:
### Issue # (if applicable)

Closes #31061.

### Reason for this change
Calling `grantConnect()` on an instance of `DatabaseInstanceReadReplica` generates an incorrect policy that uses the full ARN of the instance instead of the instanceResourceId value. It should have created policy with correct resource format `arn:aws:rds-db:region:account-id:dbuser:DbiResourceId/db-user-name` per [Creating and using an IAM policy for IAM database access](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.IAMPolicy.html).

### Description of changes
Fixed the IAM policy that `grantConnect()` generates for `DatabaseInstanceReadReplica`. The change correctly sets the value of `instanceResourceId` to replica instance `attrDbiResourceId`. The value of `instanceResourceId` is used to generate IAM policy.

- Added new unit test.
- Updated existing integration test.

### Checklist

----


packages/aws-cdk-lib/aws-rds/lib/instance.ts
```diff
@@ -1366,7 +1366,7 @@ export class DatabaseInstanceReadReplica extends DatabaseInstanceNew implements
     this.instanceIdentifier = instance.ref;
     this.dbInstanceEndpointAddress = instance.attrEndpointAddress;
     this.dbInstanceEndpointPort = instance.attrEndpointPort;
-    this.instanceResourceId = instance.attrDbInstanceArn;
```
moelasmar, line 1369: I am afraid that this can be a breaking change to customers. I know that we assigned a wrong value, but may be there are some customers that use this property to get the DB Instance Arn and use it some how. So, my recommendation is to deprecate this property, and to add a new property for the instance resource id, and use it in grant method, and also check where else we use the current property.

moelasmar (dismissed):
left one comment

packages/aws-cdk-lib/aws-rds/lib/instance.ts
ashishdhingra: @moelasmar Thanks for your review comment. Would creating a feature flag (example https://github.com/aws/aws-cdk/pull/29794) be the recommended approach instead of adding a new property? Please advise.

moelasmar: I think the feature flag here will not fix the problem, as the flag default value will be false (to avoid breaking customers), so the policy will still referring to the wrong replica instance id. So, I do not prefer feature flag, it add a lot of confusion on us when we write code, so we can use this property in similar cases assuming that we are using the correct property, but because the feature flag is disabled, this may cause more issues for customers.

I prefer to go with deprecation, and adding new property.

ashishdhingra: @moelasmar Thanks for your review comment. I have added a commit to:
- Deprecate `instanceResourceId` property and add new `instanceResourceIdV2` property in `DatabaseInstanceReadReplica`.
- Added a copy of `grantConnect()` in `DatabaseInstanceReadReplica` that makes use of `instanceResourceIdV2`.

Please review.

Thanks,
Ashish

moelasmar: Thanks @ashishdhingra for resolving my comment, and sorry for late response. 

I actually, did a deeper check, and I found that I missed that the property `instanceResourceId` is actually exist in the Interface `IDatabaseInstance` so, adding a new property will actually break the idea of that interface, and also deprecating this property is not a good idea.

So, I think it is better to make this change under feature flag, and to enable the feature flag by default, and the customers who want to keep the current behaviour, they can disable the flag.

moelasmar: I pushed some updates to this change to add the feature flag. Could you please take a look.

ashishdhingra: Looks good reviewing at high level. Needs review from the maintainer.

packages/@aws-cdk/cx-api/FEATURE_FLAGS.md
```diff
@@ -1376,4 +1379,21 @@ When this feature flag is enabled, specify newly introduced props 's3InputUri' a
...
+
+Currently, the value of the property `instanceResourceId` in construct `DatabaseInstanceReadReplica` is not correct, and set to `DbInstanceArn` which is not correct when it is used to create the IAM Policy in the grantConnect method.
+
+When this feature flag is enabled, the value of that property will be as expected set to `DbiResourceId` attribute, and that will fix the grantConnect method.
+
+
+| Since | Default | Recommended |
+| ----- |---------|-------------|
+| (not in v1) |         |             |
+| V2NEXT | `true`  | `true`      |
```
paulhcsun, line 1395: The spacing on some of these seem off, was it manually generated or auto generated by running the build?

packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
```diff
@@ -1378,4 +1381,21 @@ When this feature flag is enabled, specify newly introduced props 's3InputUri' a
...
+
+Currently, the value of the property 'instanceResourceId' in construct 'DatabaseInstanceReadReplica' is not correct, and set to 'DbInstanceArn' which is not correct when it is used to create the IAM Policy in the grantConnect method.
+
+When this feature flag is enabled, the value of that property will be as expected set to 'DbiResourceId' attribute, and that will fix the grantConnect method.
+
+
+| Since | Default | Recommended |
+| ----- | ----- | ----- |
+| (not in v1) |  |  |
+| V2NEXT | `true` | `true` |
```
paulhcsun, line 1396: Do we want this to be `true` by default? Would that not break customers?

paulhcsun (commented):
Changes look good, just a couple comments about how the feature flag is defined and the default value being `true`.

packages/@aws-cdk/cx-api/FEATURE_FLAGS.md
moelasmar: it seems while I merged the conflicts I messed up these files, let me rerun the build and see if it will fix it

packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
moelasmar: my idea was which behaviour should I keep, shall I keep the wrong behaviour, or should I fix it. I chose to fix the wrong behaviour with an option to let people keep the current one. But, honestly you have a point, it is a breaking change

moelasmar: I will change it to false.

moelasmar:
for some reason I am not able to update this PR, and push changes to @ashishdhingra forked repo, so I created this new PR https://github.com/aws/aws-cdk/pull/31579 that includes @ashishdhingra changes, and close this one.

PR #31072, opened 2024-08-09
feat(vpc-v2): add `add` methods to RouteTable and SubnetV2

Leo10Gama:
### Issue # (if applicable)

N/A

### Reason for this change

With this change, routes can be created in a given `RouteTable` without the need to use the `Route` L2 class. This will streamline the process of routing different endpoints together, making for an overall more enjoyable user experience.

### Description of changes

`RouteTable` now has an `addRoute()` method, which creates a `CfnRoute` object in the stack that links the provided IP destination to a provided router target.

Unit tests have been added to validate the behaviour of the new method, and it has been included in one aspect of the existing `integ.route-v2.ts` integration test.

### Checklist

----


packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -569,6 +569,27 @@ export class RouteTable extends Resource implements IRouteTable, IDependable {
...
+   */
+  public addRoute(destination: string, target: RouteTargetType) {
+    if (!target.gateway && !target.endpoint) {
+      throw new Error('Target is defined without a gateway or endpoint.');
+    }
+
+    let routerType: RouterType = target.gateway ? target.gateway.routerType : RouterType.VPC_ENDPOINT;
+    let routerTargetId: string = target.gateway ? target.gateway.routerTargetId : target.endpoint!.vpcEndpointId;
+
+    new CfnRoute(this, 'Route', {
```
shikha372, line 586: can we use Route L2 here?

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -569,6 +569,27 @@ export class RouteTable extends Resource implements IRouteTable, IDependable {
 
     this.routeTableId = this.resource.attrRouteTableId;
   }
+
+  /**
+   * Add a new route to the route table.
```
shikha372, line 574:
```suggestion
   * Adds a new custom route to the route table.
```

packages/@aws-cdk/aws-ec2-alpha/test/integ.route-v2.ts
```diff
@@ -84,9 +80,13 @@ const eigw = new EgressOnlyInternetGateway(stacks.eigw, 'testEOIGW', {
 });
 new Route(stacks.eigw, 'testEIGWRoute', {
   routeTable: routeTables.eigw,
-  destination: '0.0.0.0/0',
+  destination: '10.0.0.0/24',
   target: { gateway: eigw },
 });
+const eigwAddRoute = new EgressOnlyInternetGateway(stacks.eigw, 'testEOIGWviaAddRoute', {
```
shikha372, line 86: i think we don't need to add another eigw, and can just add route with a new destination to earlier eigw. Let me know if this is not correct.

Leo10Gama: Tested it, and it seems like you're right: it doesn't fail if we make a new route with the same EIGW. Pushing new changes now 👍

packages/@aws-cdk/aws-ec2-alpha/lib/subnet-v2.ts
```diff
@@ -250,6 +252,23 @@ export class SubnetV2 extends Resource implements ISubnetV2 {
...
+    return this._routeTable;
+  }
+
+  /**
+   * Associate a Route Table with this subnet.
+   * @param routeTableProps The Route Table to associate with this subnet.
+   */
+  public associateRouteTable(routeTableProps: IRouteTableV2) {
+    this._routeTable = routeTableProps;
+    this.routeTableAssociation.addPropertyOverride('RouteTableId', routeTableProps.routeTableId);
```
Leo10Gama, line 269: Currently we are doing an override rather than creating a new `CfnSubnetRouteTableAssociation` L1. If we create a new one, we would have to find some way to remove the old association L1, otherwise they will both show up in the template, and may confuse users who synthesize and see two associations for one table. The override allows us to use the same L1 object, just alter which route table it points to.

Leo10Gama:
Changes will be applied in [PR31224](https://github.com/aws/aws-cdk/pull/31224)

PR #31076, opened 2024-08-11
feat(ec2): `ipv6AddressCount` property for an instance

### LINKED ISSUE #31075: ec2: support for specify IPv6 address count for an instance
### Describe the feature

Specify [Ipv6AddressCount property](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-instance.html#cfn-ec2-instance-ipv6addresscount) for EC2 instance.

### Use Case

We can not configure `ipv6AddressCount` property through EC2 instance L2 construct.

### Proposed Solution

Add `ipv6AddressCount` to `InstanceProps` and pass it to `CfnInstance.ipv6AddressCount` property.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.150.0

irrelevant

badmintoncryer:
### Issue # (if applicable)

Closes #31075.

### Reason for this change

AWS CDK cannot configure `ipv6AddressCount` property but cloudformation can do that.

### Description of changes

- Add `ipv6AddressCount` to `InstanceProps`
- Check whether `ipv6AddressCount` is a positive integer

Add both unit and integ tests

### Checklist

----


go-to-k (changes requested):
It seems that the `ipv6AddressCount` and network interfaces should not be specified together.

> You cannot specify this option and the network interfaces option in the same request.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-instance.html#cfn-ec2-instance-ipv6addresscount

Also, when the `associatePublicIpAddress` is specified (can be true or false), network interfaces array is automatically specified.

https://github.com/aws/aws-cdk/blob/v2.151.0/packages/aws-cdk-lib/aws-ec2/lib/instance.ts#L499-L507

In other words, specifying the `ipv6AddressCount` and the `associatePublicIpAddress` should result in an error for CFn. Therefore, how about putting in a validation for this? (It would also be good to add that description to the JSDoc.)

P.S.) I actually tried to run the integ test with those properties specified and CFn got an error.

```
Resource handler returned message: "Network interfaces and an instance-level IPv6 address count may not be specified on the same request (Service: Ec2, Status Code: 400, Request ID: ...
```

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.instance-ipv6-address-count.ts
```diff
@@ -0,0 +1,37 @@
...
+      machineImage: ec2.MachineImage.latestAmazonLinux2(),
+      vpc: vpc,
+      vpcSubnets: { subnetType: ec2.SubnetType.PUBLIC },
+      allowAllIpv6Outbound: true,
+      ipv6AddressCount: 2,
+    });
+  }
+}
+
+const testCase = new TestStack(app, 'TestStack');
```
go-to-k, line 33: How about making the stack name specific to this integ test? Because there are often other integ tests with stacks of the same name, which will fail if we try to run them in parallel.

go-to-k (changes requested):

badmintoncryer:
@go-to-k Thank you for your review!! I've addressed your comments.

packages/@aws-cdk-testing/framework-integ/test/aws-ec2/test/integ.instance-ipv6-address-count.ts
badmintoncryer: I've updated it!

go-to-k (approved):
Thanks, LGTM.

badmintoncryer:
Thanks always!

packages/aws-cdk-lib/aws-ec2/lib/instance.ts
```diff
@@ -514,6 +527,20 @@ export class Instance extends Resource implements IInstance {
...
     }
 
+    if (
+      props.ipv6AddressCount !== undefined &&
+      !Token.isUnresolved(props.ipv6AddressCount) &&
+      (props.ipv6AddressCount < 0 || !Number.isInteger(props.ipv6AddressCount))
+    ) {
+      throw new Error(`\'ipv6AddressCount\' must be a non-negative integer, got: ${props.ipv6AddressCount}`);
+    }
+
```
GavinZZ, line 537: Curious that if there's a upper bound value for this property?

badmintoncryer: I tried to create an instance by specifying `ipv6AddressCount` as 2000.

Result:
```sh
Failed resources:
InstanceIpv6AddressCountTestStack | 8:04:58 AM | CREATE_FAILED        | AWS::EC2::Instance                    | Instance (InstanceC1063A87) Resource handler returned message: "Address count 2000 per interface exceeds the limit for m5.large (Service: Ec2, Status Code: 400, Request ID: 6fcb0c3e-a029-4b62-9e31-4452642bede8)" (RequestToken: 498ff895-1532-848d-2a6b-2e2af377efb0, HandlerErrorCode: InvalidRequest)
```

It seems that the upper limit varies depending on the instance type. However, I couldn't find any documentation on this...

GavinZZ: Thanks for giving it a try. To be clear I don't want us to do the validation, but just curious about this.

GavinZZ (approved):
LGTM

Issue #22535, opened 2022-10-17
(cli): does not properly set `externalId` when trying to assume lookupRole


peterwoodworth:
### Describe the bug

The lookup role doesn't seem to be assumable through CDK when requiring an external id. All other roles (deploy, file publishing, image publishing) properly set the external id, however the lookup role isn't working.

### Expected Behavior

I expect the lookup role to be assumable when requiring an external ID

### Current Behavior

Throws error:

```
Could not assume role in target account using current credentials (which are for account 181354823827) 
User: arn:aws:iam::181354823827:user/test is not authorized to perform: sts:AssumeRole 
on resource: arn:aws:iam::676158502875:role/cdk-hnb659fds-lookup-role-676158502875-eu-west-1 . 
Please make sure that this role exists in the account. If it doesn't exist, (re)-bootstrap 
the environment with the right '--trust', using the latest version of the CDK CLI.
```

### Reproduction Steps

Bootstrap an environment in account A with `cdk bootstrap aws://123456789012/eu-west-1 --trust 111122223333 --cloudformation-execution-policies arn:aws:iam::aws:policy/AdministratorAccess --trust-for-lookup 111122223333`. Then modify the trust policy on the four roles described above to require an external id when account B is assuming the role. (Note: `--trust-for-lookup` is redundant if `--trust` is also specified for the same account and doesn't fix the issue whether you use it or not)

In the CDK app, pass in these external ids to the stack to be deployed like so

```ts
new ExternalIdStack(app, 'ExternalIdTestingStack', {
  synthesizer: new DefaultStackSynthesizer({
    lookupRoleExternalId: 'myexternalid',
    deployRoleExternalId: 'myexternalid',
    fileAssetPublishingExternalId: 'myexternalid',
    imageAssetPublishingExternalId: 'myexternalid'
  }),
  env: {
    region: 'eu-west-1', account: testAccountId
  }
})
```

Inside the stack, you can test the various roles by commenting out each section. Both the image role and file role succeed deployment, indicating that the image role, file role, and deploy role all can be assumed through CDK when an external ID is required. However even with the exact same trust policy as the others, the lookup role never succeeds. 

```ts
    // Test Lookup Role
    Vpc.fromLookup(this, 'Vpc', {
      vpcName: 'MyVpc'
    });

    // Test File Role
    // new BucketDeployment(this, 'BucketDeployment', {
    //   sources: [Source.asset('./assets')],
[...]

PR #31089, opened 2024-08-12
feat: allow all `sts` options for roles assumed by the cli 

### LINKED ISSUE #26157: (aws-cdk): Add session tags when CDK assumes an IAM Role.
### Describe the feature

Provide the ability to add session tags when assuming the CDK IAM Roles.

### Use Case

Our team uses a central S3 bucket where all assets are uploaded for CloudFormation to use, as we have a lot of accounts and managing per-account/per-region buckets would be less than ideal. In order to prevent pipeline builds from accidentally overwriting assets uploaded by another, we use bucket policies to restrict the bucket prefix that any given session created by CI/CD server has access to. That way the CI/CD server must use the correct session tag to be granted access to prefix where the assets are uploaded. For example, here is what a simplified statement in the bucket policy looks like:

```JSON
    {
      "Sid": "AllowReadWriteCrossAccountViaPrincipalTag",
      "Effect": "Allow",  
      "Principal": "*",
      "Action": [
        "s3:GetObject",
        "s3:PutObject"
      ],  
      "Resource": "arn:aws:s3:::bucketName/us-east-1/${aws:PrincipalTag/serverId:uniqueId}/*",
      "Condition": {
        "StringLike": {      
          "aws:arn": [
            "arn:aws:sts::<account-id>:assumed-role/cicd-role/*",
           ]
        }
      }
    }
```

We would like to be able to follow this same access pattern when CDK uploads assets to the S3 bucket from our pipeline; however, there does not seem to be any way to configure session tags when using the DefaultStackSynthesizer.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.78.0 (build 8e95c37)

MacOS Ventura 13.4.1 (22F82)

sumupitchayan:
Allow passing [all STS options](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html#API_AssumeRole_RequestParameters) to assume role configuration for various CDK roles. 

> The following PR description focuses on Session Tags because it was originally the only option we wanted to add support to. After some thought, we decided to allow all available STS options via a transparent pass-through.

### Prerequisites

- https://github.com/cdklabs/cloud-assembly-schema/pull/33
- https://github.com/cdklabs/cdk-assets/pull/40

### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/26157 
Fixes https://github.com/aws/aws-cdk/issues/22535

### Reason for this change

Enabling [ABAC](https://docs.aws.amazon.com/IAM/latest/UserGuide/introduction_attribute-based-access-control.html) via STS session tags. From the [AWS docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_session-tags.html):

> _“Session Tags are key-value pair attributes that you pass when you assume an IAM role or federate a user in AWS STS. You do this by making an AWS CLI or AWS API request through AWS STS or through your identity provider (IdP). When you use AWS STS to request temporary security credentials, you generate a session. Sessions expire and have [credentials](https://docs.aws.amazon.com/STS/latest/APIReference/API_Credentials.html), such as an access key pair and a session token. When you use the session credentials to make a subsequent request, the [request context](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition.html#AccessPolicyLanguage_RequestContext) includes the [aws:PrincipalTag](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-principaltag) context key. You can use the aws:PrincipalTag key in the Condition element of your policies to allow or deny access based on those tags”_

### Description of changes

The CDK creates [4 IAM roles](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping-env.html#bootstrapping-env-roles) during bootstrap. It then assumes these roles at various phases of the deployment.

- [DeploymentActionRole](https://github.com/aws/aws-cdk/blob/v2.154.1/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml#L429): Assumed when invoking CloudFormation operations such as _Deploy_ and _DescribeStackEvents_.
- [FilePublishingRole](https://github.com/aws/aws-cdk/blob/v2.154.1/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml#L275): Assumed when file assets are uploaded to the bootstrap bucket.
- [ImagePublishingRole](https://github.com/aws/aws-cdk/blob/v2.154.1/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml#L298): Assumed when docker images are published to the bootstrap ECR registry.
- [LookupRole](https://github.com/aws/aws-cdk/blob/v2.154.1/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml#L321): Assumed while performing context lookups.

Each of these roles should be assumable with their own specific session tags, as they server different purposes.

> Note: The [CloudFormationExecutionRole](https://github.com/aws/aws-cdk/blob/v2.154.1/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml#L536) is also created during bootstrap, but the CLI never assumes it, therefore it doesn't need session tags support.

Session tags for each role will be configurable via synthesizer properties (similarly to how `externalId` is [exposed](https://github.com/aws/aws-cdk/pull/15604)) both for the `DefaultStackSynthesizer`, and for a custom synthesizer extending the `StackSynthesizer` class. The new properties will propagate down and will eventually be written to the cloud assembly.

#### `+ manifest.json`

```json
{
  "version": "36.0.0",
  "artifacts": {
    "MyStack.assets": {
      "type": "cdk:asset-manifest",
      "properties": {
        "file": "SeshTagsManifestStack.assets.json",
        "requiresBootstrapStackVersion": 6,
        "bootstrapStackVersionSsmParameter": "/cdk-bootstrap/hnb659fds/version"
      }
    },
    "MyStack": {
      "type": "aws:cloudformation:stack",
      "environment": "aws://unknown-account/unknown-region",
      "properties": {
        "assumeRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-deploy-role-${AWS::AccountId}-${AWS::Region}",
          "assumeRoleAdditionalOptions": {
            "Tags": < deployRoleSessionTags > 
          }
        "lookupRole": {
          "arn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-lookup-role-${AWS::AccountId}-${AWS::Region}",
          "requiresBootstrapStackVersion": 8,
          "bootstrapStackVersionSsmParameter": "/cdk-bootstrap/hnb659fds/version"
          "assumeRoleAdditionalOptions": {
            "Tags": < lookupRoleSessionTags > 
          }
        }
      },
```

#### `+ assets.json`

```json
{
  "version": "36.0.0",
  "files": {
    "9ebfd704f02f99b2711998e6435822b0dbed6e80dcac7e75f339fe894861ac20": {
      "source": {
        "path": "mystack.template.json",
        "packaging": "file"
      },
      "destinations": {
        "current_account-current_region": {
          "assumeRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-file-publishing-role-${AWS::AccountId}-${AWS::Region}"
          "assumeRoleAdditionalOptions": {
             "Tags": < fileAssetPublishingRoleSessionTags >
           }
        }
      }
    }
  },
  "dockerImages": {
    "dockerHash": {
      "source": {
        "directory": "."
      },
      "destinations": {
        "current_account-current_region": {
          "assumeRoleArn": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/cdk-hnb659fds-image-publishing-role-${AWS::AccountId}-${AWS::Region}"
          "assumeRoleAdditionalOptions": {
            "Tags": < imageAssetPublishingRoleSessionTags >
          }
        }
      }
    }
  }
}
```

- CLI integration tests.
- CLI and framework unit tests.

### Checklist

----

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,75 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  analyze:
+    # Need this because the default runners run out of disk space
+    runs-on: [aws-cdk_ubuntu-latest_16-core]
+    
+    name: Analyze (${{ matrix.language }})
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
+    #   - https://gh.io/supported-runners-and-hardware-resources
+    #   - https://gh.io/using-larger-runners (GitHub.com only)
+    # Consider using larger runners or machines with greater resources for possible analysis time improvements.
+    permissions:
+      # required for all workflows
+      security-events: write
+
+      # required to fetch internal or private CodeQL packs
+      packages: read
+
+      # only required for workflows in private repositories
+      actions: read
+      contents: read
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - language: javascript-typescript
+          build-mode: none
+        - language: python
+          build-mode: none
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v3
+      with:
+        languages: ${{ matrix.language }}
+        build-mode: ${{ matrix.build-mode }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+
+        # For more details on CodeQL's query packs, refer to: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
+        # queries: security-extended,security-and-quality
+
+    # If the analyze step fails for one of the languages you are analyzing with
+    # "We were unable to automatically build your code", modify the matrix above
+    # to set the build mode to "manual" for that language. Then modify this step
+    # to build your code.
+    # ℹ️ Command-line programs to run using the OS shell.
+    # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
+    - if: matrix.build-mode == 'manual'
+      shell: bash
+      run: |
+        echo 'If you are using a "manual" build mode for one or more of the' \
+          'languages you are analyzing, replace this with the commands to build' \
+          'your code, for example:'
+        echo '  make bootstrap'
+        echo '  make release'
+        exit 1
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v3
+      with:
+        category: "/language:${{matrix.language}}"
--- a/.github/workflows/request-cli-integ-test.yml
+++ b/.github/workflows/request-cli-integ-test.yml
@@ -19,7 +19,7 @@ jobs:
           persist-credentials: false
       - name: Find changed cli files
         id: changed-cli-files
-        uses: tj-actions/changed-files@e9772d140489982e0e3704fea5ee93d536f1e275
+        uses: tj-actions/changed-files@48d8f15b2aaa3d255ca5af3eba4870f807ce6b3c
         with:
           base_sha: ${{ github.event.pull_request.base.sha }}
           files_yaml: |
--- a/yarn.lock
+++ b/yarn.lock
@@ -16556,7 +16556,16 @@ string-length@^4.0.1:
     char-regex "^1.0.2"
     strip-ansi "^6.0.0"
 
-"string-width-cjs@npm:string-width@^4.2.0", string-width@*, string-width@^1.0.1, "string-width@^1.0.2 || 2 || 3 || 4", string-width@^4.1.0, string-width@^4.2.0, string-width@^4.2.3, string-width@^5.0.1, string-width@^5.1.2:
+"string-width-cjs@npm:string-width@^4.2.0":
+  version "4.2.3"
+  resolved "https://registry.npmjs.org/string-width/-/string-width-4.2.3.tgz#269c7117d27b05ad2e536830a8ec895ef9c6d010"
+  integrity sha512-wKyQRQpjJ0sIp62ErSZdGsjMJWsap5oRNihHhu6G7JVO/9jIB6UyevL+tXuOqrng8j/cxKTWyWUwvSTriiZz/g==
+  dependencies:
+    emoji-regex "^8.0.0"
+    is-fullwidth-code-point "^3.0.0"
+    strip-ansi "^6.0.1"
+
+string-width@*, string-width@^1.0.1, "string-width@^1.0.2 || 2 || 3 || 4", string-width@^4.1.0, string-width@^4.2.0, string-width@^4.2.3, string-width@^5.0.1, string-width@^5.1.2:
   version "4.2.3"
   resolved "https://registry.npmjs.org/string-width/-/string-width-4.2.3.tgz#269c7117d27b05ad2e536830a8ec895ef9c6d010"
   integrity sha512-wKyQRQpjJ0sIp62ErSZdGsjMJWsap5oRNihHhu6G7JVO/9jIB6UyevL+tXuOqrng8j/cxKTWyWUwvSTriiZz/g==
@@ -16621,7 +16630,7 @@ stringify-package@^1.0.1:
   resolved "https://registry.npmjs.org/stringify-package/-/stringify-package-1.0.1.tgz#e5aa3643e7f74d0f28628b72f3dad5cecfc3ba85"
   integrity sha512-sa4DUQsYciMP1xhKWGuFM04fB0LG/9DlluZoSVywUMRNvzid6XucHK0/90xGxRoHrAaROrcHK1aPKaijCtSrhg==
 
-"strip-ansi-cjs@npm:strip-ansi@^6.0.1", strip-ansi@^6.0.0, strip-ansi@^6.0.1:
+"strip-ansi-cjs@npm:strip-ansi@^6.0.1":
   version "6.0.1"
   resolved "https://registry.npmjs.org/strip-ansi/-/strip-ansi-6.0.1.tgz#9e26c63d30f53443e9489495b2105d37b67a85d9"
   integrity sha512-Y38VPSHcqkFrCpFnQ9vuSXmquuv5oXOKpGeT6aGrr3o3Gc9AlVa6JBfUSOCnbxGGZF+/0ooI7KrPuUSztUdU5A==
@@ -16634,6 +16643,13 @@ strip-ansi@^3.0.1:
   integrity sha512-VhumSSbBqDTP8p2ZLKj40UjBCV4+v8bUSEpUb4KjRgWk9pbqGF4REFj6KEagidb2f/M6AzC0EmFyDNGaw9OCzg==
   dependencies:
     ansi-regex "^2.0.0"
+
+strip-ansi@^6.0.0, strip-ansi@^6.0.1:
+  version "6.0.1"
+  resolved "https://registry.npmjs.org/strip-ansi/-/strip-ansi-6.0.1.tgz#9e26c63d30f53443e9489495b2105d37b67a85d9"
+  integrity sha512-Y38VPSHcqkFrCpFnQ9vuSXmquuv5oXOKpGeT6aGrr3o3Gc9AlVa6JBfUSOCnbxGGZF+/0ooI7KrPuUSztUdU5A==
+  dependencies:
+    ansi-regex "^5.0.1"
 
 strip-ansi@^7.0.1, strip-ansi@^7.1.0:
   version "7.1.0"
@@ -17632,7 +17648,7 @@ workerpool@^6.5.1:
   resolved "https://registry.npmjs.org/workerpool/-/workerpool-6.5.1.tgz#060f73b39d0caf97c6db64da004cd01b4c099544"
   integrity sha512-Fs4dNYcsdpYSAfVxhnl1L5zTksjvOJxtC5hzMNl+1t9B8hTJTdKDyZ5ju7ztgPy+ft9tBFXoOlDNiOT9WUXZlA==
 
-"wrap-ansi-cjs@npm:wrap-ansi@^7.0.0", wrap-ansi@^7.0.0:
+"wrap-ansi-cjs@npm:wrap-ansi@^7.0.0":
   version "7.0.0"
   resolved "https://registry.npmjs.org/wrap-ansi/-/wrap-ansi-7.0.0.tgz#67e145cff510a6a6984bdf1152911d69d2eb9e43"
   integrity sha512-YVGIj2kamLSTxw6NsZjoBxfSwsn0ycdesmc4p+Q21c5zPuZ1pl+NfxVdxPtdHvmNVOQ6XSYG4AUtyt/Fi7D16Q==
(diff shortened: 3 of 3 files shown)


packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
```diff
@@ -620,7 +628,7 @@ export async function installNpmPackages(fixture: TestFixture, packages: Record<
   }, undefined, 2), { encoding: 'utf-8' });
 
   // Now install that `package.json` using NPM7
-  await fixture.shell(['node', require.resolve('npm'), 'install']);
+  await fixture.shell(['node', '/Users/sumughan/Developer/aws-cdk/packages/@aws-cdk-testing/cli-integ/lib/package-sources/repo-tools/npm', 'install']);
```
cgarvis, line 631: you have your personal path hardcoded.

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
```diff
@@ -643,6 +664,9 @@ const app = new cdk.App({
   },
 });
 
+// const sessionTagApp = new cdk.App({
+// });
```
cgarvis, line 668: remove commented out code.

packages/@aws-cdk-testing/cli-integ/lib/with-cdk-app.ts
sumupitchayan: Yup this is temporary to get the cli-integ test to work.

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
```diff
@@ -422,6 +422,29 @@ class LambdaStack extends cdk.Stack {
...
 
+class SessionTagsLambdaStack extends cdk.Stack {
+  constructor(parent, id, props) {
+    super(parent, id, {
+      ...props,
+      synthesizer: new DefaultStackSynthesizer({
+        deployRoleSessionTags: {
+          'Department' : 'Engineering',
+        },
+        fileAssetPublishingRoleSessionTags: {
```
iliapolo, line 433: Add (and assert on) session tags for the docker image upload role and add a docker image asset to this stack.

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
```diff
@@ -422,6 +422,29 @@ class LambdaStack extends cdk.Stack {
   }
 }
 
+class SessionTagsLambdaStack extends cdk.Stack {
```
iliapolo, line 425:
```suggestion
class SessionTagsStack extends cdk.Stack {
```

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
```diff
@@ -672,6 +695,7 @@ switch (stackSet) {
     new MissingSSMParameterStack(app, `${stackPrefix}-missing-ssm-parameter`, { env: defaultEnv });
 
     new LambdaStack(app, `${stackPrefix}-lambda`);
+    new SessionTagsLambdaStack(app, `${stackPrefix}-session-tags-lambda`);
```
iliapolo, line 698:
```suggestion
    new SessionTagsLambdaStack(app, `${stackPrefix}-session-tags`);
```

packages/@aws-cdk-testing/cli-integ/tests/cli-integ-tests/bootstrapping.integtest.ts
```diff
@@ -82,6 +82,24 @@ integTest('can and deploy if omitting execution policies', withoutBootstrap(asyn
   });
 }));
 
+// Custom Bootstrap test with Session Tags on the DeployRole
+integTest('can deploy with session tags on the deploy role', withoutBootstrap(async (fixture) => {
```
iliapolo, line 86: As we discussed, at a test for a custom synthesizer that doesn't define a cloudformation execution role and defines session tags for the deploy role.

iliapolo (changes requested):
Still missing plenty of unit tests as well

packages/@aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js
```diff
@@ -422,6 +425,95 @@ class LambdaStack extends cdk.Stack {
...
+      })
+    });
+
+    // Lambda Function to test AssetPublishingRole
+    const fn = new lambda.Function(this, 'my-function', {
+      code: lambda.Code.asset(path.join(__dirname, 'lambda')),
+      runtime: lambda.Runtime.NODEJS_LATEST,
+      handler: 'index.handler'
+    });
+    new cdk.CfnOutput(this, 'FunctionArn', { value: fn.functionArn });
```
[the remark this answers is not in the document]
sumupitchayan: nope

packages/aws-cdk-lib/core/README.md
```diff
@@ -155,6 +155,31 @@ new MyStack(app, 'MyStack', {
...
+        deployRoleSessionTags: {
+          'Department' : 'Engineering',
+        },
+      })
+    });
+  }
+}
+```
+
+You can also do the same with a custom stack synthesizer.
```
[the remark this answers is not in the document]
sumupitchayan: Added an example using the same `CustomSynthesier` as the unit/integ test which extends the `DefaultStackSynthesizer` and passes `undefined` in for the `cloudFormationExecutionRoleArn`.

packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml
```diff
@@ -323,6 +337,13 @@ Resources:
     Properties:
       AssumeRolePolicyDocument:
         Statement:
+          # allows this role to be assumed with session tags.
+          # see https://docs.aws.amazon.com/IAM/latest/UserGuide/id_session-tags.html#id_session-tags_permissions-required
+          - Action: sts:TagSession
```
rix0rrr, line 342: I'm surprised that this appears in the `AssumeRolePolicyDocument`.

But you've tested this, so this is definitely correct, right?

rix0rrr:
I thought CodeQL just ran automatically...

In any case:

```
  Caused by: Severe disk cache trouble (corruption or out of space) at /home/runner/work/_temp/codeql_databases/javascript/db-javascript/default/cache/pages/ae/64.pack: Failed to write item to disk
  (eventual cause: IOException "No space left on device")
```

build.assume-role-additional-options.sh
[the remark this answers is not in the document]
iliapolo: Yes.

packages/aws-cdk-lib/core/README.md
```diff
@@ -155,6 +155,46 @@ new MyStack(app, 'MyStack', {
...
+}
+```
+
+This will cause the CDK CLI to include session tags when assuming each of these roles during deployment. Note that the trust policy of the role 
+must contain permissions for the `sts:TagSession` action.
+
+> See https://docs.aws.amazon.com/IAM/latest/UserGuide/id_session-tags.html#id_session-tags_permissions-required
+
+- If you are using a custom bootstrap template, make sure the template includes these permissions.
+- If you are using the default bootstrap template, you will need to rebootstrap your enviroment (once).
```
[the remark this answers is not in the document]
iliapolo: This sounds awfully familiar doesn't it? :) 

You just know merge will get delayed and we'll end up putting the wrong version number here...let me think about it.

packages/aws-cdk-lib/core/lib/stack-synthesizers/default-synthesizer.ts
```diff
@@ -429,34 +487,57 @@ export class DefaultStackSynthesizer extends StackSynthesizer implements IReusab
...
+    return MIN_BOOTSTRAP_STACK_VERSION;
+  }
+
+  private get _requiredBootstrapVersionForLookup() {
+    if (this.props.lookupRoleAdditionalOptions?.Tags) {
+      return MIN_SESSION_TAGS_BOOTSTRAP_STACK_VERSION;
+    }
+    return MIN_LOOKUP_ROLE_BOOTSTRAP_STACK_VERSION;
+  }
+
```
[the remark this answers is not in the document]
iliapolo: Hh yeah, but it helped me make sense of whats happening here. I'll leave it in the hopes it will help others who aren't familiar with this enough.

packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml
iliapolo: I was also surprised but yes.

iliapolo:
> I thought CodeQL just ran automatically...
> 
> In any case:
> 
> ```
>   Caused by: Severe disk cache trouble (corruption or out of space) at /home/runner/work/_temp/codeql_databases/javascript/db-javascript/default/cache/pages/ae/64.pack: Failed to write item to disk
>   (eventual cause: IOException "No space left on device")
> ```

What the fork :\

packages/aws-cdk-lib/core/README.md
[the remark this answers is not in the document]
rix0rrr: I'm happy if you put "version XXX" in it. At least it's an obvious placeholder that either you remember to update just before merging, or someone will complain after the fact and then we can update it.

packages/aws-cdk-lib/core/lib/stack.ts
```diff
@@ -1061,7 +1061,10 @@ export class Stack extends Construct implements ITaggable {
    * Synthesizes the cloudformation template into a cloud assembly.
    * @internal
    */
-  public _synthesizeTemplate(session: ISynthesisSession, lookupRoleArn?: string): void {
+  public _synthesizeTemplate(session: ISynthesisSession,
+    lookupRoleArn?: string,
+    lookupRoleExternalId?: string,
```
iliapolo, line 1066: Noticed that we didn't used to pass `lookupRoleExternalId` here, which means it wasn't added to the missing context, which means that it was ignored during lookups. I reckon we didn't see this because its a pretty extreme use case and its likely no one is using it.

packages/aws-cdk-lib/core/lib/stack.ts
```diff
@@ -1104,11 +1107,17 @@ export class Stack extends Construct implements ITaggable {
     fs.writeFileSync(outPath, templateData);
 
     for (const ctx of this._missingContext) {
-      if (lookupRoleArn != null) {
```
iliapolo, line 1107: There is no need for this `if` because undefined properties are filtered anyway.

packages/aws-cdk-lib/core/lib/stack.ts
iliapolo: This is the cause of https://github.com/aws/aws-cdk/issues/22535

packages/@aws-cdk/cx-api/FEATURE_FLAGS.md
iliapolo, line 1: This file is a duplicate of https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md and should never have been committed.

packages/aws-cdk/lib/api/aws-auth/sdk-provider.ts
```diff
@@ -575,3 +592,21 @@ function fmtObtainedCredentials(
       return msg.join('');
   }
 }
+
+/**
+ * Instantiate an SDK for context providers. This function ensures that all
+ * lookup assume role options are used when context providers perform lookups.
+ */
+export async function initContextProviderSdk(aws: SdkProvider, options: cxschema.ContextLookupRoleOptions): Promise<ISDK> {
```
iliapolo, line 600: @rix0rrr Can you have a look at this (and how its being used)? I added it since your last review to fix the lookup issues.

packages/@aws-cdk/cx-api/FEATURE_FLAGS.md
iliapolo, line 1: This file is a copy of https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md and is actually configured to be ignored:

https://github.com/aws/aws-cdk/blob/1593500735e6ddbcc087f005c5124f5ec57aec20/packages/%40aws-cdk/cx-api/.gitignore#L3-L9

But since it was committed before the ignore rules, it still lingers on.

rix0rrr (approved):

Issue #30607, opened 2024-06-21
aws_s3: BucketNotification in owning stack deletes BucketNotifications from other stacks


sebastian-fredriksson-bernholtz:
### Describe the bug

When making changes to the S3 event notifications in the stack that owns an S3 Bucket (on `Bucket`), it deletes event notifications for the bucket that have been configured in other stacks (on `IBucket`).

### Expected Behavior

Event notifications configured in other stacks should not be deleted.

### Current Behavior

Event notifications configured in other stacks are being deleted.

### Reproduction Steps

1. Create and deploy Bucket in stack 1.
```ts
// stack 1
new Bucket(this, 'Bucket', {
    bucketName: 'bucketname',
});
```
2. Add and deploy event notification in stack 2
```ts
// stack 2
Bucket.fromBucketName(this, 'DataBucket', 'bucketName').addEventNotification(
    EventType.OBJECT_CREATED_PUT,
    new LambdaDestination(lambda)
);
```
3. Make a change to event notifications in stack 1 and deploy:
```diff
// stack 1
new Bucket(this, 'Bucket', {
    bucketName: 'bucketname',
+   eventBridgeEnabled: true,
});
```

The event handler configured and deployed in step 2 will be deleted when doing step 3.

### Possible Solution

Use the same logic for handling BucketNotifications in the stack that owns the Bucket as in other stack: https://github.com/aws/aws-cdk/blob/7360a885e6282ad28b4ae72f9ae92a6bcda88b15/packages/%40aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py#L37

### Additional Information/Context

This is happening for Bucket (unlike IBucket) cdk sets `Managed` property on the custom resource that manages event notifications to true.
https://github.com/aws/aws-cdk/blob/71986ff986d13bbb496b33c0554f657e77dbb2d0/packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource.ts#L120
https://github.com/aws/aws-cdk/blob/71986ff986d13bbb496b33c0554f657e77dbb2d0/packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource.ts#L135

And the code in the custom resource lambda handler disregards externally set notifications if `Managed` is set to true.
[...]

PR #31091, opened 2024-08-12
fix(s3): bucket notifications in owning stack deletes bucket notifications from other stacks

xazhao:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/30607.

### Reason for this change

There's a bug reported in the Github issue that bucket notifications in owing stack will remove all notifications added in imported stack. This is because we treated the bucket as `managed` hence we use bucket notifications in that stack as source of truth.

In the `unmanaged` path, we already filtered out external notifications it should handle both scenarios when the bucket is managed or unmanaged.

### Description of changes

Always set `Managed` property to false when the feature flag is enabled. Here we introduce a feature flag to prevent it breaking current customers. 

Added unit tests. Integrations test can't validate this change because we need to deploy twice to actually see the change.
Also tested manually.

### Checklist

----


packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource.ts
```diff
@@ -126,6 +127,13 @@ export class BucketNotifications extends Construct {
         }));
       }
 
+      // The customer resource handles unmanaged bucket is the right way so setting the managed flag to false
```
GavinZZ, line 130: This comment is confusing. Can we phrase it better?

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource.ts
```diff
@@ -126,6 +127,13 @@ export class BucketNotifications extends Construct {
...
+      // Ading a feature flag to prevent it brings unexpected changes to customers
+      // Put it here because we still need to create the permission if it's unmanaged bucket.
+      if (cdk.FeatureFlags.of(this).isEnabled(cxapi.S3_KEEP_NOTIFICATION_IN_IMPORTED_BUCKET)) {
+        managed = false;
+      }
```
GavinZZ, line 135: I still don't get why we put it after the `addToRolePolicy` statement. If the variable `managed` is originally `false` but feature flag is on, then we treated it as unmanaged. It will go through the unmanaged code which will need the role policy, but it's not added.

packages/aws-cdk-lib/aws-s3/test/notification.test.ts
```diff
@@ -226,6 +227,28 @@ describe('notification', () => {
...
+
+    // WHEN
+    new s3.Bucket(stack, 'MyBucket', {
+      eventBridgeEnabled: true,
+    });
+
+    // THEN
+    Template.fromStack(stack).resourceCountIs('AWS::S3::Bucket', 1);
+    Template.fromStack(stack).hasResourceProperties('Custom::S3BucketNotifications', {
+      NotificationConfiguration: {
```
GavinZZ, line 245: Follow up comment with the policy one, let's also add a check to make sure the policy exists.

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource.ts
xazhao: Done.

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource.ts
xazhao: Updated.

GavinZZ (approved):

PR #31092, opened 2024-08-12
feat(CLI): synth displays "AssertDescription: CDK bootstrap stack version 6 required"

### LINKED ISSUE #17942: cdk 2.1.0 synth, diff subcommands fail with "AssertDescription: CDK bootstrap stack version 6 required...". cdk 2.1.0 deploy subcommand succeeds.
### What is the problem?

`cdk synth` and `cdk diff` fail with "AssertDescription: CDK bootstrap stack version 6 required. Please run 'cdk bootstrap' with a recent version of the CDK CLI." after the template checks SSM for version not 1-5. The actual version in SSM is 9.

`cdk deploy` succeeds.

After the successful `cdk deploy`, the synth and diff continue to fail.

### Reproduction Steps

1. Upgrade cdk 2.0.0 to 2.1.0
2. run cdk bootstrap with appropriate credentials
3. create a new folder and cd to it
4. create a new cdk project with `cdk init --language typescript`
5. cdk synth
6. cdk diff
7. cdk deploy

A clean synth with no errors displayed
A clean diff with no errors displayed
A clean deploy with no errors displayed

### What actually happened?

for synth and diff
```
...
Rules:
  CheckBootstrapVersion:
    Assertions:
      - Assert:
          Fn::Not:
            - Fn::Contains:
                - - "1"
                  - "2"
                  - "3"
                  - "4"
                  - "5"
                - Ref: BootstrapVersion
        AssertDescription: CDK bootstrap stack version 6 required. Please run 'cdk bootstrap' with a recent version of the CDK CLI.
```

for deploy - a clean, error-free execution with a CFn template produced as expected.

### CDK CLI Version

2.1.0

### Framework Version

2.0.0

### Node.js Version

16.10.0

### OS

macOS Monterey (12.0.1)

### Language

Typescript

### Language Version

TypeScript 4.2.4

### Other information

The bootstrap CDKToolkit stack looks as expected - all of the modern template trappings (the previous bootstrap was legacy). The value of the SSM parameter /cdk-bootstrap/hnb659fds/version is 9, as expected.  The cdk-hnb659fds-deploy-role-\<account>-\<region> role has the appropriate ssm:GetParameter permission for .../cdk-bootstrap/hnb659fds/version. the cdk-hnb659fds-lookup-role-\<account>-\<region> role has ssm:Get* permission for *.

"old" cdk projects continue to function as expected with `cdk synth` etc.

Hoping you can deduce some bozo thing I must have overlooked or done. Thanks for whatever help you can provide.

comcalvi:
### Issue # (if applicable)

Closes #17942. 

### Reason for this change

The CDK CLI shows the stack template, which includes the CFN Rule `CheckBootstrapVersion`. This rule will fail a deployment if the bootstrap is not right. Customers think this rule is an error message. 

### Description of changes

Obscure this `CheckBootstrapVersion` Rule from the template when we print it, if it exists. If it is the only Rule, remove the `Rules` section entirely. 

Manual testing, unit tests, and CLI integration tests. 

### Checklist

----


mrgrain (changes requested):
This seems like not a great idea to me to obscure things from the template the we later deploy regardless. What other options where considered as a solution?

Feel free to dismiss my review if the has been discussed sufficiently in your team.

comcalvi:
@mrgrain the team has not discussed this yet. I think this is the best path forward, and it's a precedent we've set by obscuring `AWS::CDK::Metadata` from the diff. 

I see the following options:

1. Update our docs to tell users it's expected when you run `synth` or `diff`. 
2. Update the error message, telling people they can ignore it if they see it in CLI output. 
3. Obscure it from the CLI output, a precedent we set with `AWS::CDK::Metadata`. 

I don't think the first two address the issue. 

For 1: People will likely run `synth` once before they read whatever doc tells them about it, and be confused.
For 2: The error message will become more confusing if they get it during deployment and it says "you can maybe ignore this."

3 addresses the issue by ensuring people only see the error message where they need to see it; if a deployment is failing because of it. If people want to see the full, un-altered diff, they can pass `--strict`, which will keep it there.

mrgrain:
Hmm I've forget about metadata, my bad. That's a great argument for it.

--- 

There were a few other ideas in the issue: 
- slightly adjust the message
- move it to the top of the template 
- print "success" to stderr

I also wonder if this check needs to be in the template or if it could be done by the cli.

comcalvi:
@mrgrain that's a good point. The reason is that deployments are sometimes initiated from pipelines, without the CDK CLI; this is documented here: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/stack-synthesizers/stack-synthesizer.ts#L309

mrgrain:
> @mrgrain that's a good point. The reason this is a Rule in the template is that deployments are sometimes initiated from pipelines, without the CDK CLI; this is documented here: https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/stack-synthesizers/stack-synthesizer.ts#L309

Cool! This is probably true, but if you are curious let's dig deeper. How are pipelines deploying stacks? What do they need the Bootstrap stack for? 

My guess here is that it's the `cdk-asset` command that really cares about this. But I'm not sure. 

Either way, this is not a blocker.

packages/aws-cdk/lib/cdk-toolkit.ts
```diff
@@ -1526,3 +1533,21 @@ function buildParameterMap(parameters: {
...
+
+/**
+ * Remove any template elements that we don't want to show users.
+ */
+function obscureTemplate(template: any = {}) {
+  if (template.Rules) {
+    // see https://github.com/aws/aws-cdk/issues/17942
+    if (template.Rules.CheckBootstrapVersion) {
+      if (Object.keys(template.Rules).length > 1) {
+        delete template.Rules.CheckBootstrapVersion;
```
scanlonp, line 1545: I thought you had mentioned that the rules would only be deleted if the check bootstrap version was the only rules. Did I misinterpret that, or did we switch to always removing the bootstrap version check?

scanlonp (approved):
Approving with 1 nit.

We should update the cli option description here: https://github.com/aws/aws-cdk/blob/9acd5288b2b9a4d72dac199d472a42b5590a916e/packages/aws-cdk/lib/cli.ts#L263.

packages/aws-cdk/lib/cdk-toolkit.ts
comcalvi: The intention is to always remove the Rule, but to remove the entire Rules section if that is the only rule.

comcalvi:
@mrgrain I am curious but I'm not able to dig deeper into this right now.

PR #31097, opened 2024-08-13
fix(dynamodb): replication regions are incompatible with resource policies in TableV2

### LINKED ISSUE #30705: dynamodb: Replication regions are incompatible with resource policies in TableV2
### Describe the bug

As stated in the [AWS documentation for resource policies](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/rbac-considerations.html), we cannot create a replica and add a resource policy to that replica in the same stack update.

This means every time we want to add a new replication region in the CDK we need to somehow not add the resource policy to that new region for the initial deployment. TableV2 allows us to customize each replica, including whether or not we want a resource policy, but it too eagerly adds the resource policy under TableV2 construct and we are unable to override it with an `undefined` to not include any resourcePolicy.

### Expected Behavior

When adding a new replica region that specifies an undefined resource policy, TableV2 should not add the resource policy defined within it's construct.

### Current Behavior

TableV2 eagerly adds the resource policy to all replicas, even when specifying it should be undefined, and the deployment fails. The only way to add a new replica without adding a resource policy is to first deploy a stack update that creates the new replica and removes the resource policy from **all** other replicas. Then following up with a second stack update that re-adds the resource policy to all replicas. This means between the first and second update none of the tables will have a resource policy attached.

### Reproduction Steps

Using a TableV2 WITHOUT replicas, but WITH the resource policy and this works
```
new TableV2(this, `MyTable-${stage}`, {
      pointInTimeRecovery: true,
      partitionKey: {
        name: 'key',
        type: AttributeType.STRING,
      },
      tableName: 'MyTable',
      resourcePolicy: tablePolicyDocument,
    });
```
If I then try to deploy WITH a replica (us-east-1) and WITH a resource policy, it fails due same stack update error
```
new TableV2(this, `MyTable-${stage}`, {
      pointInTimeRecovery: true,
      partitionKey: {
        name: 'key',
        type: AttributeType.STRING,
      },
      tableName: 'MyTable',
      replicas: [{ region: 'us-east-1' }],
      resourcePolicy: tablePolicyDocument,
    });
```
When I delete the stack and restart and try to deploy WITH a replica (us-east-1) but WITHOUT the resource policy, this succeeds
```
new TableV2(this, `MyTable-${stage}`, {
      pointInTimeRecovery: true,
      partitionKey: {
        name: 'key',
        type: AttributeType.STRING,
      },
      tableName: 'MyTable',
      replicas: [{ region: 'us-east-1' }]
    });
```
When I then add the resource policy WITHOUT adding a new replica region, this works and the resource policy is added to both the table in us-west-2 and us-east-1
```
new TableV2(this, `MyTable-${stage}`, {
      pointInTimeRecovery: true,
      partitionKey: {
        name: 'key',
        type: AttributeType.STRING,
      },
      tableName: 'MyTable',
      replicas: [{ region: 'us-east-1' }],
      resourcePolicy: tablePolicyDocument,
    });
```
But when I then try to add another region (us-east-2), it then fails with the same stack update error
```
new TableV2(this, `MyTable-${stage}`, {
      pointInTimeRecovery: true,
      partitionKey: {
        name: 'key',
        type: AttributeType.STRING,
      },
      tableName: 'MyTable',
      replicas: [{ region: 'us-east-1' }, { region: 'us-east-2' }],
      resourcePolicy: tablePolicyDocument,
    });
```
If I try to override the resourcePolicy in the new replica to not include one, TableV2 still adds `tablePolicyDocument` to it in the CFN template and the deployment fails with the same stack update error
```
new TableV2(this, `MyTable-${stage}`, {
      pointInTimeRecovery: true,
      partitionKey: {
        name: 'key',
        type: AttributeType.STRING,
      },
      tableName: 'MyTable',
      replicas: [{ region: 'us-east-1' }, { region: 'us-east-2', resourcePolicy: undefined }],
      resourcePolicy: tablePolicyDocument,
    });
```

I've so far been able to work around this by using the L1 escape hatch with `CfnGlobalTable` and manually specifying each replica and selectively adding the resource policy to each replica.

```
    new CfnGlobalTable(this, `MyTable-${stage}`, {
      tableName: 'MyTable',
      attributeDefinitions: [{ attributeName: 'key', attributeType: AttributeType.STRING }],
      keySchema: [{ attributeName: 'key', keyType: 'HASH' }],
      replicas: [
        {
          pointInTimeRecoverySpecification: {
            pointInTimeRecoveryEnabled: true,
          },
          region: 'us-west-2', // main table region
          resourcePolicy: { policyDocument: tablePolicyDocument },
        },
        {
          pointInTimeRecoverySpecification: {
            pointInTimeRecoveryEnabled: true,
          },
          region: 'us-east-1',
        },
      ],
      billingMode: 'PAY_PER_REQUEST',
      streamSpecification: { streamViewType: 'NEW_AND_OLD_IMAGES' },
    });
```

### Possible Solution

The easiest solution is to update [this line](https://github.com/aws/aws-cdk/blob/9437b4dbbea15db18ff2991d410e033cc3437001/packages/aws-cdk-lib/aws-dynamodb/lib/table-v2.ts#L618) from:

>const resourcePolicy = props.resourcePolicy ?? this.tableOptions.resourcePolicy;

to

>const resourcePolicy = props.resourcePolicy

and require all replicas to manually include a resourcePolicy if one is desired.

Ideally, the construct could allow `null` and when `null` is specified in a specific replica, no resourcePolicy is added to that replica even when one is defined in the TableV2 itself

```
new TableV2(this, `MyTable-${stage}`, {
      pointInTimeRecovery: true,
      partitionKey: {
        name: 'key',
        type: AttributeType.STRING,
      },
      tableName: 'MyTable',
      // us-east-2 should not have tablePolicyDocument added in the template
      replicas: [{ region: 'us-east-1' }, { region: 'us-east-2', resourcePolicy: null }], 
      resourcePolicy: tablePolicyDocument,
    });
```

_No response_

### CDK CLI Version

2.136.0

### Framework Version

_No response_

### Node.js Version

v18.18.2

### OS

Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

LeeroyHannigan:
### Issue # (if applicable)

Closes #30705

### Reason for this change

Resource policies were shared across all replicas in a region. 

### Description of changes

Changed the logic to only apply resource policy to the local replica region, or to specific replicas only when defined.

Unit tests were conducted

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/cx-api/lib/features.ts
+++ b/packages/aws-cdk-lib/cx-api/lib/features.ts
@@ -108,7 +108,6 @@ export const LOG_API_RESPONSE_DATA_PROPERTY_TRUE_DEFAULT = '@aws-cdk/custom-reso
 export const DYNAMODB_TABLEV2_RESOURCE_POLICY_PER_REPLICA = '@aws-cdk/aws-dynamodb:resourcePolicyPerReplica';
 export const S3_KEEP_NOTIFICATION_IN_IMPORTED_BUCKET = '@aws-cdk/aws-s3:keepNotificationInImportedBucket';
 
-
 export const FLAGS: Record<string, FlagInfo> = {
   //////////////////////////////////////////////////////////////////////
   [ENABLE_STACK_NAME_DUPLICATES_CONTEXT]: {


GavinZZ (commented):
Can you explain why this changes are needed to help me understand? What's the current behaviour and issue when setting the resoruce policies in the replication regions?

LeeroyHannigan:
The issue is defined in #31097 

Do you require more information than is supplied there?

packages/aws-cdk-lib/aws-dynamodb/lib/table-v2.ts
```diff
@@ -664,8 +664,9 @@ export class TableV2 extends TableBaseV2 {
   private configureReplicaTable(props: ReplicaTableProps): CfnGlobalTable.ReplicaSpecificationProperty {
     const pointInTimeRecovery = props.pointInTimeRecovery ?? this.tableOptions.pointInTimeRecovery;
     const contributorInsights = props.contributorInsights ?? this.tableOptions.contributorInsights;
-    const resourcePolicy = props.resourcePolicy ?? this.tableOptions.resourcePolicy;
-
+    //const resourcePolicy = props.resourcePolicy ?? this.tableOptions.resourcePolicy;
+    //const resourcePolicy = props.resourcePolicy;
```
GavinZZ, line 668: Can we remove the comments?

packages/aws-cdk-lib/aws-dynamodb/lib/table-v2.ts
```diff
@@ -664,8 +664,9 @@ export class TableV2 extends TableBaseV2 {
...
+    const resourcePolicy = (props.region === this.region ? this.tableOptions.resourcePolicy : props.resourcePolicy) || undefined;
```
GavinZZ, line 669: This is a breaking change for existing users. If existing users manage to bypass the CFN error by using escape hatch. They then remove the escape hatch and upgrade to a version that includes this change, their replication region will now have the resource policies removed, thus this is not an acceptable change. I would recommend you to do it following the suggested proposal in the original issue.
```
Ideally, the construct could allow null and when null is specified in a specific replica, no resourcePolicy is added to that replica even when one is defined in the TableV2 itself

new TableV2(this, `MyTable-${stage}`, {
      pointInTimeRecovery: true,
      partitionKey: {
        name: 'key',
        type: AttributeType.STRING,
      },
      tableName: 'MyTable',
      // us-east-2 should not have tablePolicyDocument added in the template
      replicas: [{ region: 'us-east-1' }, { region: 'us-east-2', resourcePolicy: null }], 
      resourcePolicy: tablePolicyDocument,
    });

```

GavinZZ (changes requested):

packages/aws-cdk-lib/aws-dynamodb/lib/table-v2.ts
LeeroyHannigan: This is forcing a user to add null for when they don't want a resource policy. The default implementation should only require them to add a policy to a replica should the want it. It shouldn't default to adding it to all replicas.

packages/@aws-cdk-testing/framework-integ/test/aws-dynamodb/test/integ.dynamodb-v2.featureflag.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    // table with resource policy
+    new dynamodb.TableV2(this, 'TableTestV2', {
+      partitionKey: {
+        name: 'id',
+        type: dynamodb.AttributeType.STRING,
+      },
+      removalPolicy: RemovalPolicy.DESTROY,
+      resourcePolicy: docu,
+    });
```
GavinZZ, line 35: This test doesn't seem to test for replica?

packages/@aws-cdk-testing/framework-integ/test/aws-dynamodb/test/integ.dynamodb-v2.policy.js.snapshot/ResourcePolicyTest-v2.template.json
```diff
@@ -46,7 +49,10 @@
        }
       }
      }
-    ]
+    ],
+    "StreamSpecification": {
+     "StreamViewType": "NEW_AND_OLD_IMAGES"
+    }
```
GavinZZ, line 55: Where is this change from?

packages/@aws-cdk-testing/framework-integ/test/aws-dynamodb/test/integ.dynamodb-v2.policy.ts
```diff
@@ -31,6 +31,9 @@ class TestStack extends Stack {
     });
 
     table.grantReadData(new iam.AccountPrincipal('123456789012'));
+    table.addReplica({
+      region: 'eu-west-2',
+    });
```
GavinZZ, line 36: This test didn't even set the feature flag, so my understanding is that it should use the table's resource policy but doesn't seem to be the case in the output template.json. May I ask how you ran the tests locally and update the snapshot?

GavinZZ (changes requested):
Changes LGTM, some minor feedbacks on the integ tests.

packages/@aws-cdk-testing/framework-integ/test/aws-dynamodb/test/integ.dynamodb-v2.policy.js.snapshot/ResourcePolicyTest-v2.template.json
LeeroyHannigan: The stream came from adding a replica. 
```
    table.addReplica({
      region: 'eu-west-2',
    });
 ```

GavinZZ: Thanks, that's good to know.

Issue #8581, opened 2020-06-16
SNS notifications on stack events


logemann:
## :question: General Issue

I am playing around with different ways to hook into lifecycle events of Stacks / Components. Of course there is the AwsCustomResource or ProviderFramework way. While AwsCustomResource is only viable for a simple SDK calls, ProviderFramework offers a lot more but with more complexity involved.

Now i noticed notificationArns on Stack level. What i recall you can attach them via command line on deployment (not what i am looking for) or programatically via NestedStack Construct. I wonder why this requirement is in place? Why cant we add them on our main stacks? (i think you can only read them)

This way one could easily setup a topic and a consumer and one should be able to hook into those lifecycle events pretty easily. 

### proposed idea

Why not having a method like this on Stack construct:

```
let myStack = new MyStack(app,...);
myStack.addNotificationArns(...);
```

Not sure why this method is not available in the construct when at the end it will synth to AWS::CloudFormation::Stack

eladb:
Relabeled as feature. This is something we should be able to support.

yashghatti:
Hey guys, any update on this?

stephankaag:
@rix0rrr Can you give us a pointer where to implement this? I can look into building a pull-request.

yashghatti:
@stephankaag thanks for looking into this, so right now notification ARNs are only supported as a command-line arg to the cdk command. But it'd be nice (as the first comment mentioned) to have a method in the core.Stack class or even a constructor field to pass in notificationArns

AustinZhu:
+1

domfie:
AFAIK: Currently only possible through the cli:
`cdk deploy stackID --notification-arns='arn:aws:sns:{region}:{aws-id}:{sns-name}'`
(Replace the placeholders with your values)

On subsequent calls the arn can be omitted.

logemann:
Thats why it would make up a good feature Request. Think should be part of the AWS CDK.

kornicameister:
Wouldn't this be a bit of chicken & egg problem? I mean something gotta provide SNS topic?

domfie:
> Wouldn't this be a bit of chicken & egg problem? I mean something gotta provide SNS topic?

Not really since you first need to deploy the sns with another stack or create it by hand. A check if this sns exists and a corresponding error message should not be the problem

kornicameister:
Given that it seems to be already supported on nested stack it shouldn't bee to much of a trouble to add this for top-level stacks, right?

[...]

PR #31107, opened 2024-08-14
feat(core): configure Stack SNS notification ARNs on the Stack construct

comcalvi:
### Issue # (if applicable)

#8581.

### Reason for this change

It is easier and clearer to specify the SNS Topic ARNs on the stack construct itself instead of passing it as a command line argument.

### Description of changes

Added a new optional stack prop, `notificationArns`, that is written to the CloudAssembly and concatenated with the CLI option `--notification-arns`. 

Don't forget to select stacks by hierarchical ID (currently display name, in our tests) when writing certain test code. Otherwise, the tests may not select the stack you expect.

Depends on: https://github.com/cdklabs/cdk-assets/pull/87 and https://github.com/cdklabs/cloud-assembly-schema/pull/58.

Unit tests + CLI integ test. Framework integ tests not included because they would require an externally-created SNS Topic, which is not what we want in integ tests; besides, the case is covered by the CLI integ test. 

### Checklist

----


packages/aws-cdk-lib/core/lib/stack.ts
```diff
@@ -450,6 +464,7 @@ export class Stack extends Construct implements ITaggable {
       throw new Error(`Stack name must be <= 128 characters. Stack name: '${this._stackName}'`);
     }
     this.tags = new TagManager(TagType.KEY_VALUE, 'aws:cdk:stack', props.tags);
+    this._notificationArns = props.notificationArns ?? [];
```
rix0rrr, line 467: We should probably validate that these strings don't contain any tokens.

rix0rrr (commented):
I don't necessarily have a problem with this, but I'm a bit dubious on the practical use cases.

rix0rrr (changes requested):
See previous review

comcalvi:
@rix0rrr could you review: https://github.com/cdklabs/cloud-assembly-schema/pull/29? That needs to merge before this PR can merge.

rix0rrr (approved):

PR #31108, opened 2024-08-14
feat(cognito): add `tags` property for user pool

kdnakt:
### Reason for this change

The userPoolTags property existed in the L1 construct but was not present in the L2 construct

### Description of changes

Add the tags property for UserPool

Added a unit test and an integration test.

### Checklist

----


packages/aws-cdk-lib/aws-cognito/lib/user-pool.ts
```diff
@@ -749,6 +749,12 @@ export interface UserPoolProps {
    * @default - no value
    */
   readonly advancedSecurityMode?: AdvancedSecurityMode;
+
+  /**
+   * The tags applied to the user pool.
+   * @default {} - no tags
+   */
```
badmintoncryer, line 756:
```suggestion
  /**
   * The tags applied to the user pool.
   *
   * @default - no tags
   */
```

packages/@aws-cdk-testing/framework-integ/test/aws-cognito/test/integ.user-pool-tags.ts
```diff
@@ -0,0 +1,26 @@
...
+
+new CfnOutput(stack, 'user-pool-id', {
+  value: userpool.userPoolId,
+});
+
+new IntegTest(app, 'IntegUserPoolTags', {
+  testCases: [stack],
+});
+
+app.synth();
```
badmintoncryer, line 26:
```suggestion
```

badmintoncryer (changes requested):
Thank you for your contribution! I've added 2 minor comments.

kdnakt:
@badmintoncryer Thank you for the suggestion, I've just committed them.

badmintoncryer (approved):
Thanks!

GavinZZ (commented):
In L2 construct, we don't directly provide `tags` property even though it's supported in the relevant L1 construct.

The typical way of adding tag is through `Tags.of(this).add`. Does this method work for you?

kdnakt:
Thank you, the `Tags.of(this).add` solution worked for me.
Would you recommend that I close this PR now?

GavinZZ:
As a side note, if you use `this`, it will default to result in tags being added at stack level this propagating to all resources created in the stack.

If this is not the ideal solution you could also do `Tags.of(resource)`.

Let’s close this PR as I believe this change is not needed anymore.

PR #31129, opened 2024-08-16
fix: allow assumeRoleAction to be set

### LINKED ISSUE #31128: aws-iam: Unable to use sts:AssumeRoleWithWebIdentity to assume a role
### Describe the bug

Due to <https://github.com/aws/aws-cdk/blob/9acd5288b2b9a4d72dac199d472a42b5590a916e/packages/aws-cdk-lib/aws-iam/lib/role.ts#L373> the principal used will always be able to use the sts:AssumeRole action, even if the desired behaviour is only to allow sts:AssumeRoleWithWebIdentity.

This matters when using OIDC to assume a role over on AWS as you should only be able to do sts:AssumeRoleWithWebIdentity

### Regression Issue

_No response_

### Expected Behavior

Users would be able to set which assumeRoleAction they want to use, as documented in https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_iam.Role.html#:~:text=Description-,assumeroleaction,-string

### Current Behavior

assumeRoleAction is always sts:AssumeRole

### Reproduction Steps

Create a role 

```typescript


 const githubActionsRole = new cdk.aws_iam.Role(this, "GithubActionsRole", {
      roleName: "GithubActionsRole", // Must be static to make cross account auth easier
      assumedBy: new cdk.aws_iam.PrincipalWithConditions(
        new cdk.aws_iam.FederatedPrincipal(
          `arn:aws:iam::${this.account}:oidc-provider/token.actions.githubusercontent.com`
        ),
        // It's important that this role is locked down to only our github orgs, as otherwise anyone on github could use permissions on our AWS infrastructure.
        {
          StringEquals: {
            "token.actions.githubusercontent.com:aud": "sts.amazonaws.com",
          },
          StringLike: {
            "token.actions.githubusercontent.com:sub": "repo:example-organisation/*", // This currently allows all repos in the tesslio github org to assume this role.
          },
        }

      ),
    });```

Then look at the AWS console, and see that the assumeRoleAction is "sts:AssumeRole" rather than the required "sts:AssumeRoleWithWebIdentity" and there's no current way to set this.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.147.3 (build 32f0fdb)

### Framework Version

_No response_

### Node.js Version

20.16.0

### OS

macOS 14.3 (23D56)

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

RichardoC:
### Issue #31128

Closes #31128

### Reason for this change

This is required for using OIDC to assume an AWS role, this should be sts:AssumeRoleWithWebIdentity in this case

### Description of changes

Default assumeRoleAction to 'sts:AssumeRole' for backwards compatibility, but allow it to be overridden if needed

Added custom assumeRoleAction to 'can use all role properties' 

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-iam/test/role.test.ts
+++ b/packages/aws-cdk-lib/aws-iam/test/role.test.ts
@@ -219,6 +219,7 @@ describe('customizeRoles', () => {
     // WHEN
     const role = new Role(stack, 'Role', {
       assumedBy: new ServicePrincipal('sns.amazonaws.com'),
+      assumeRoleAction: "sts:AssumeRoleWithWebIdentity";
     });
     const principal = Role.fromRoleName(stack, 'OtherRole', 'OtherRole');
     role.grant(principal, 'sts:AssumeRole');
@@ -288,7 +289,7 @@ describe('customizeRoles', () => {
         },
         conditions: {},
       },
-      AssumeRoleAction: 'sts:AssumeRole',
+      AssumeRoleAction: 'sts:AssumeRoleWithWebIdentity',
       PrincipalAccount: {
         Ref: 'AWS::AccountId',
       },


RichardoC:
Not needed, can be done via <https://github.com/aws/aws-cdk/issues/31128#issuecomment-2293279743>

PR #23590, opened 2023-01-06
chore(cli): move integ tests to @aws-cdk-testing/cli-integ

rix0rrr:
Clean up the CLI integration tests a bit, by moving them to a separate TypeScript package. This has the following benefits:

- More clearly isolates the tests and keeps them separate from the CLI package proper.
- Allows normal specification of dependencies of the integration tests themselves, via a proper `package.json` and Yarn lockfile.
- Makes it more clear what reusable files are part of the CLI, vs what reusable files are part of the integration tests.
- Makes it more clear/easier to download an old copy of the tests.
- Tests that used to be written in `bash` are now written in TypeScript+Jest, making them easier to maintain.

Also makes the following changes:

- Publishes the test artifacts to CodeArtifact, so that we don't need to do crazy shell hijacking shenanigans in order to fake NPM/Maven/NuGet/PyPI repositories from a set of artifacts.
- We needed to change how the alpha packages are numbered in a candidate build:
  * If the current build was `2.3.4`, the candidate build versions would be `2.4.0-rc.0` and `2.4.0-alpha.0`.
  * However, if the pipeline runs with `2.4.0` actually released, **its** alpha packages would also be numbered `2.4.0-alpha.0`.
  * To unambiguously refer to the alpha packages of the candidate version (keeping in mind the naming restrictions of all package managers), we now call those `2.4.0-alpha.999`.

This is not the final form of our testing infrastructure. Right now, a fresh CodeArtifact repo is created for every test with all packages in it. In an actual pipeline, we would be better off publishing to a CodeArtifact repo once and all tests pulling from that.

That will be forthcoming in the future.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #31131, opened 2024-08-16
chore(CLI-integ-tests): cli integ tests cannot use local CDK framework

comcalvi:
### Reason for this change

The cli integration tests cannot use your local version of `aws-cdk-lib`. This can be verified by making your `Stack` construct throw an error upon creation, and watching no CLI integration tests fail, even with `-a`.

### Description of changes

Fixed the CLI integration test framework to correctly link the local packages, like it [used to](https://github.com/aws/aws-cdk/pull/23590).

Manual testing. This isn't something we can add automated tests for. 

### Checklist

----


packages/@aws-cdk-testing/cli-integ/lib/package-sources/repo-source.ts
```diff
@@ -75,13 +75,14 @@ const YARN_MONOREPO_CACHE: Record<string, any> = {};
...
+export async function findYarnPackages(root: string): Promise<Record<string, string>> {
   if (!(root in YARN_MONOREPO_CACHE)) {
-    const output: YarnWorkspacesOutput = JSON.parse(await shell(['yarn', 'workspaces', '--silent', 'info'], {
+    const outputDataString: string = JSON.parse(await shell(['yarn', 'workspaces', '--json', 'info'], {
       captureStderr: false,
       cwd: root,
       show: 'error',
-    }));
+    })).data;
+    const output: YarnWorkspacesOutput = JSON.parse(outputDataString);
```
scanlonp, line 85: Why do we need to parse the shell output again?

packages/@aws-cdk-testing/cli-integ/lib/package-sources/repo-source.ts
```diff
@@ -96,7 +97,7 @@ async function findYarnPackages(root: string): Promise<Record<string, string>> {
  * Find the root directory of the repo from the current directory
  */
 export async function autoFindRoot() {
-  const found = await findUp('release.json');
+  const found = findUp('release.json');
```
scanlonp, line 100: Odd that this was ever awaited. Looking in the `findUp` function, I am not seeing async calls, but is there any chance this await is necessary?

scanlonp (commented):
This essentially skips install of the CDK and uses the local copy, correct? Just getting clarification, then good to approve.

packages/@aws-cdk-testing/cli-integ/lib/package-sources/repo-source.ts
comcalvi: TS compiler said it's not needed, so I removed it. It doesn't return a promise, so there's no need to await anything here.

packages/@aws-cdk-testing/cli-integ/lib/package-sources/repo-source.ts
comcalvi: we first need to parse the output of `yarn workspaces`, which includes a `data` object. That `data` object is an escaped JSON string, which needs to be parsed again.

comcalvi:
Yep, it skips installing the CDK and uses the local version.

scanlonp (approved):

Issue #24507, opened 2023-03-08
(aws-iam): grantAssumeRole does nothing


dkaksl:
### Describe the bug

`Role.grantAssumeRole()` does not make any difference in the generated policy.

### Expected Behavior

`grantAssumeRole` should grant the given princial permission to assume the role.

### Current Behavior

No-op.

### Reproduction Steps

```typescript
const role = new Role(this, 'role', {
  assumedBy: new AccountPrincipal(this.account)
})
role.grantAssumeRole(new AccountPrincipal('123456789')) // no-op
```

### Additional Information/Context

Workaround:

``````typescript
const role = new Role(this, 'role', {
  assumedBy: new AccountPrincipal(this.account)
})

//role.grantAssumeRole(new AccountPrincipal('123456789'))

role.assumeRolePolicy?.addStatements(
  new PolicyStatement({
    effect: Effect.ALLOW,
    actions: ['sts:AssumeRole'],
    principals: [new AccountPrincipal('123456789')]
  })
)
```

### CDK CLI Version

2.67.0

### Framework Version

_No response_

### Node.js Version

v18.14.2

### OS

Ubuntu

### Language

Typescript

### Language Version

_No response_

### Other information

_No response_

pahud:
Hi 

You will need to use `assumeRolePolicy` instead:

```ts
  const role = new iam.Role(this, 'role', {
    assumedBy: new iam.AccountPrincipal(this.account),
  })

  role.assumeRolePolicy?.addStatements(new iam.PolicyStatement({
    actions: ['sts:AssumeRole'],
    principals: [new iam.AccountPrincipal('123456789')],
  }))
```

related to https://github.com/aws/aws-cdk/issues/22550#issuecomment-1283095003

begna112:
What is the purpose of `role.grantAssumeRole` if it does not actually add the AssumeRole statements for you? 

When trying to use this workaround, it can end up trying to create cyclical references. In particular when trying to add RolePrincipals or RoleArns, such as for a specific lambda.

IllarionovDimitri:
have the same problem. spent hours to figure out why grant_assume_role(identity) does nothing. how should one actually find out this workaround to add statement to the role policy... no idea...

lukepafford:
Would like to comment that I just encountered this issue too and it's a bit bewildering. I'm totally fine using `role.assumeRolePolicy?.addStatements` to modify the roles trust policy.

[...]

Issue #23090, opened 2022-11-25
[aws-iam]: Should it warn or fail if using grantAssumeRole on imported role?


asgerjensen:
### Describe the bug

After playing around with an imported eks cluster, trying to get the kubeprovider lambda to be able to assume the role of the kubecontroller, i finally realized that the issue was, that the kubecontroller role was imported, and the code tried to grant the lambda-role permission to run as the kubecontroller role, via `lambdarole.grant(kubecontroller, 'sts::AssumeRole'`.

But since, in my case, the kubecontroller role was also an import (in order to preregister it in aws-auth), this was rendered as

```
  "HandlerServiceRoleDefaultPolicyCBD0CC91": {
   "Type": "AWS::IAM::Policy",
   "Properties": {
    "PolicyDocument": {
     "Statement": [
      {
       "Action": "eks:DescribeCluster",
       "Effect": "Allow",
       "Resource": "arn:aws:eks:eu-north-1:xxxx:cluster/MyCluster"
      },
      {
       "Action": "sts:AssumeRole",
       "Effect": "Allow",
       "Resource": "arn:aws:iam::xxxx:role/KubernetesController"
      }
     ],
     "Version": "2012-10-17"
    },
    "PolicyName": "HandlerServiceRoleDefaultPolicyCBD0CC91",
    "Roles": [
     {
      "Ref": "HandlerServiceRoleFCDC14AE"
     }
    ]
   },
```
Ie a resource policy, and not as part of the KubernetesControllers trust-relationship policy.

I get that it cant just (at least i dont think it can), update the trust-relationship policy of the imported role, but it would have saved me a few days if it had just bailed early, with a "This is not going to do what you think it will do".

Does it make sense to add a check to Grant, that checks if the principal is non-owned, and give a warning during the validation step?

### Expected Behavior

Either grantAssumeRole or grant(..., 'sts:AssumeRole') would work on any kind of principal, owned or not, or it would give a warning that the operation will not end up being functional.

### Current Behavior

Generates resource policy that does not seem to have the desired effect.

### Reproduction Steps

```
const roleA = Role.fromRoleArn('.....');
const roleB = new Role(this, 'newrole', {});


roleA.grantAssumeRole(roleB);
```
Try to run something with roleB, that assumes roleA.

### Possible Solution

Warning or Error in validation step if granter is not owned.

### CDK CLI Version

2.51.1

### Node.js Version

16

### OS

mac

### Language

Typescript

### Language Version

3.9.7

[...]

PR #31135, opened 2024-08-17
fix(iam): adding addPrincipalsToAssumedBy() to allow adding moreprincipals to TrustRelationship of Role to overcome limitation of grantAssumeRole() (#22550)

### LINKED ISSUE #22550: aws-iam: Make setting trust on roles more clear in overview and function descriptions
### Describe the bug

When running the grant_assume_role on a role with a Service Principle as the input, the role's trust policy is not updated.

### Expected Behavior

I expect the role's trust policy to be updated.

### Current Behavior

Nothing happens. If I change the Service Principle in the function and run a cdk diff, there is no difference in deployment suggesting the function is not doing anything.

### Reproduction Steps

In Python: 

```
        # Create new IAM role for DMS access to Redshift
        dmsRedshiftRole = iam.Role(self, "dmsRedshiftRole",
            assumed_by=iam.ServicePrincipal(
                "dms.{}.amazonaws.com".format(self.region)),
            description="IAM role to be used by DMS for access to Redshift",
            managed_policies=[iam.ManagedPolicy.from_aws_managed_policy_name(
                "service-role/AmazonDMSRedshiftS3Role")],
        )

        # Allow DMS role to be assumed by Redshift.
        dmsRedshiftRole.grant_assume_role(iam.ServicePrincipal("redshift.amazonaws.com"))
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.43.1

### Framework Version

_No response_

### Node.js Version

8.5.4

### OS

Mac Monterey 12.5

### Language

Python

### Language Version

3.9.14

### Other information

_No response_

stm29:
### Issue # (if applicable)

Closes #22550

### Reason for this change
- We need to rely on `assumeRolePolicy` property of a `Role` to add more trust relationships to a role, as `grantAssumeRole()` has a limitation mentioned in https://github.com/aws/aws-cdk/issues/22550#issuecomment-1283095003 
- This Following PR - https://github.com/aws/aws-cdk/pull/29452 added,
  - How to use `assumeRolePolicy` in `README.md`
  - Silently Fails, when **Service / Account Principal** is used with **grantAssumeRole()**
  - We still can't add `ArnPrincipal()` using grantAssumeRole(), (Doing this does nothing) as mentioned on below issues,
    - https://github.com/aws/aws-cdk/issues/23090
    - https://github.com/aws/aws-cdk/issues/24507
    - https://github.com/aws/aws-cdk/issues/22550
- It would be **better** to have similar function like grantAssumeRole() to add Principals directly.

### Description of changes
- Added **addPrincipalsToAssumedBy()** function
- Removed Optional property on **assumeRolePolicy** (as I guess, it will never be undefined)
- Changed **README.md** respectively
- Added Integration Tests

- Unit Tests
- Integ Tests
- 
  ```
  $ yarn integ --directory test/aws-iam/test --update-on-failed
  ```

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-iam/README.md
+++ b/packages/aws-cdk-lib/aws-iam/README.md
@@ -396,8 +396,7 @@ const role = new iam.Role(this, 'role', {
 // Using addPrincipalsToAssumedBy()
 role.addPrincipalsToAssumedBy([
   new iam.AccountPrincipal('123456789012'),
-  ...
-])
+]);
 
 // Using assumeRolePolicy
 role.assumeRolePolicy.addStatements(new iam.PolicyStatement({
--- a/packages/aws-cdk-lib/aws-iam/test/role.test.ts
+++ b/packages/aws-cdk-lib/aws-iam/test/role.test.ts
@@ -421,7 +421,6 @@ describe('IAM role', () => {
   test('extend TrustRelationships of a role using addPrincipalsToAssumedBy()', () => {
     // GIVEN
     const stack = new Stack();
-    stack.partition;
 
     // WHEN
     const role = new Role(stack, 'MyRole', {


packages/aws-cdk-lib/aws-iam/lib/role.ts
```diff
@@ -375,7 +375,7 @@ export class Role extends Resource implements IRole {
   /**
    * The assume role policy document associated with this role.
    */
-  public readonly assumeRolePolicy?: PolicyDocument;
+  public readonly assumeRolePolicy: PolicyDocument;
```
shikha372, line 378: we cannot make an optional props to be required as it is a breaking change and can lead to failures for multiple users.

packages/aws-cdk-lib/aws-iam/lib/role.ts
```diff
@@ -697,6 +697,20 @@ export class Role extends Resource implements IRole {
     return getPrecreatedRoleConfig(this);
   }
 
+  /**
+   * Adds given list of Principals to the TrustRelationship.
+   * This helps in overcoming how grantAssumeRole() works for Service / Account Principals.
+   * More Details on how grantAssumeRole() doesn't adds principals to trust relationship on below issue:
+   * https://github.com/aws/aws-cdk/issues/22550#issuecomment-1283095003
+   */
+  public addPrincipalsToAssumedBy(principals: IPrincipal[]): void {
```
shikha372, line 706: i think we can do it currently with the help of `role.assumeRolePolicy?` already, don't see any added value with this function.

shikha372:
Thanks @stm29 for submitting this PR, we're aware of the current limitations of `grantAssumeRole` method and it is documented as well in the docs. While we value our community contributions, the functionality that we're trying to add here can easily be done already with `role.assumeRole`, i don;t see how doing the same thing by do different methods could help.

packages/aws-cdk-lib/aws-iam/lib/role.ts
stm29: This seems to be **public readonly** property. and this is set in constructor (see below) and moreover this is the **only place we set this**. There is no way user will set this. and There is no way it will be undefined, based on how it is being used in constructor IMO. So, ideally it shouldn't break anything. it should be backward compatible.

https://github.com/aws/aws-cdk/blob/4e7d4804093d1a2a54f03830a3770d56c4729bc8/packages/aws-cdk-lib/aws-iam/lib/role.ts#L425

please let me know does this makes sense.

packages/aws-cdk-lib/aws-iam/lib/role.ts
stm29: Correct, but I personally fell rather than relying **role.assumeRolePolicy**, it is better to have similar function like `grant()`, `grantAssume()`. Having function of similar kind **addPrincipalsToAssumedBy()** to give better helper value to users (though it is documented well).
Consider this following scenario, `grant()` is sufficient enough to add a PolicyDocument, but we do have helper functions like **grantSendMessage()** for `sqs`, **grantEncryptDepcrypt()** for `kms` to service for specific purposes right?, so this will also serve similar value is my thought.

Would like you hear from you..

shikha372: grant methods are encapsulation for policy statements, are we encapsulating any policies here(just the assume role), we are using the same object and just modifying it in the function, i'll  put this into perspective with the help of dx to make it more clear and to be sure i'm not missing something here:

Current method:
```
role.assumeRolePolicy?.addStatements(new iam.PolicyStatement({
  actions: ['sts:AssumeRole'],
  principals: [
    new iam.AccountPrincipal('123456789'),
    new iam.ServicePrincipal('beep-boop.amazonaws.com')
    ],
}));
```

Proposed: 

```
role.addPrincipalsToAssumedBy([
  new iam.AccountPrincipal('123456789012'),
  new iam.ServicePrincipal('beep-boop.amazonaws.com')
]);
```

stm29: Exactly, you got my point. that **sts:AssumeRole** is the one i am talking about in terms of easing usage of utilities.
For Example see here how it is used in other utility functions in current scenario,

https://github.com/aws/aws-cdk/blob/975df1f5a17e9a2ff2b34223c84dd43e72057de2/packages/aws-cdk-lib/aws-kms/lib/key.ts#L201-L203

https://github.com/aws/aws-cdk/blob/975df1f5a17e9a2ff2b34223c84dd43e72057de2/packages/aws-cdk-lib/aws-kms/lib/private/perms.ts#L26-L28

This just adds `kms:Decrypt` to the provided `Grantee`.

stm29:
Hi @shikha372 , is there any comments, pending from my end to be addressed?

shikha372:
Hi @stm29,
Thank you for contributing to this, aligned with team as well that this function is not adding any value to the current usage. Hence, will be closing this PR. If we wish to modify this behaviour, right place to address it could be under `assumeRole` function where current behaviour is to throw an error.

PR #31150, opened 2024-08-19
feat(cloudfront): s3 origin access control

gracelu0:
### Issue # (if applicable)

n/a

### Reason for this change

Note: This PR is a WIP, not merging to the main branch. Merging to a feature branch first to facilitate smaller PR reviews.

Support OAC for S3 origins, see [RFC](https://github.com/aws/aws-cdk-rfcs/pull/624) for more details

### Description of changes

- Deprecate `CloudFrontWebDistribution` in favour of existing class `Distribution`
- Add new abstract base class `OriginAccessControlBase` and subclass `S3OriginAccessControl`
- Deprecate `S3Origin` in favour of new classes `S3BucketOrigin` and `S3StaticWebsiteOrigin`
- Add new abstract class `S3BucketOrigin`
    - 3 static functions for instantiating S3 origin with OAI, OAC and no origin access control (using bucket defaults)
- Add new class `S3StaticWebsiteOrigin`
- Update README for `cloudfront` and `cloudfront-origins`

- `integ.s3-origin-oac`: basic use case setting up a distribution with an S3 origin with OAC, using assertions to check the distribution has the oac configured and the oac has the expected default properties
- unit tests for `S3OriginAccessControl`

(more tests to follow in coming PRs)

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudfront/lib/web-distribution.ts
+++ b/packages/aws-cdk-lib/aws-cloudfront/lib/web-distribution.ts
@@ -4,7 +4,6 @@ import { HttpVersion, IDistribution, LambdaEdgeEventType, OriginProtocolPolicy, 
 import { FunctionAssociation } from './function';
 import { GeoRestriction } from './geo-restriction';
 import { IKeyGroup } from './key-group';
-import { IOriginAccessControl } from './origin-access-control';
 import { IOriginAccessIdentity } from './origin-access-identity';
 import { formatDistributionArn } from './private/utils';
 import * as certificatemanager from '../../aws-certificatemanager';
@@ -12,7 +11,6 @@ import * as iam from '../../aws-iam';
 import * as lambda from '../../aws-lambda';
 import * as s3 from '../../aws-s3';
 import * as cdk from '../../core';
-import { Annotations } from '../../core';
 
 /**
  * HTTP status code to failover to second origin
@@ -215,12 +213,6 @@ export interface SourceConfiguration {
    * @default - origin shield not enabled
    */
   readonly originShieldRegion?: string;
-
-  /**
-   * Origin Access Control
-   * @default - No origin access control
-   */
-  readonly originAccessControl?: IOriginAccessControl;
 }
 
 /**
@@ -318,13 +310,6 @@ export interface S3OriginConfig {
    * @default No Origin Access Identity which requires the S3 bucket to be public accessible
    */
   readonly originAccessIdentity?: IOriginAccessIdentity;
-
-  /**
-   * The optional Origin Access Control that Cloudfront will use when accessing your S3 bucket.
-   *
-   * @default - No origin access control
-   */
-  readonly originAccessControl?: IOriginAccessControl;
 
   /**
    * The relative path to the origin root to use for sources.
@@ -1134,9 +1119,6 @@ export class CloudFrontWebDistribution extends cdk.Resource implements IDistribu
 
     let s3OriginConfig: CfnDistribution.S3OriginConfigProperty | undefined;
     if (originConfig.s3OriginSource) {
-      if (originConfig.s3OriginSource.originAccessIdentity && originConfig.s3OriginSource.originAccessControl) {
-        throw Error('Only one of origin access identity or origin access control can be defined.');
-      }
       // first case for backwards compatibility
       if (originConfig.s3OriginSource.originAccessIdentity) {
         // grant CloudFront OriginAccessIdentity read access to S3 bucket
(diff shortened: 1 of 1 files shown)


packages/aws-cdk-lib/aws-cloudfront/lib/web-distribution.ts
```diff
@@ -1118,6 +1134,9 @@ export class CloudFrontWebDistribution extends cdk.Resource implements IDistribu
 
     let s3OriginConfig: CfnDistribution.S3OriginConfigProperty | undefined;
     if (originConfig.s3OriginSource) {
+      if (originConfig.s3OriginSource.originAccessIdentity && originConfig.s3OriginSource.originAccessControl) {
+        throw Error('Only one of origin access identity or origin access control can be defined.');
+      }
```
samson-keung: Would this block users from doing 2-step migration approach where they would have both OAI and OAC?

samson-keung (changes requested):

packages/aws-cdk-lib/aws-cloudfront/lib/origin-access-control.ts
```diff
@@ -0,0 +1,210 @@
...
+ * An Origin Access Control for Amazon S3 origins.
+ * @resource AWS::CloudFront::OriginAccessControl
+ * @see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudfront-originaccesscontrol.html
+ */
+export class S3OriginAccessControl extends OriginAccessControlBase {
+  /**
+   * Imports an S3 origin access control from its id.
+   */
+  public static fromOriginAccessControlId(scope: Construct, id: string, originAccessControlId: string): IOriginAccessControl {
+    class Import extends Resource implements IOriginAccessControl {
```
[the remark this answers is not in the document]
gracelu0: Yep makes sense, I'll update it

packages/aws-cdk-lib/aws-cloudfront/lib/web-distribution.ts
gracelu0: Removed this as we are deprecating `CloudFrontWebDistribution` - was an old change I accidentally pushed.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        const distributionId = options.distributionId;
+        const actions = this.getActions(props?.originAccessLevels ?? [cloudfront.AccessLevel.READ]);
+        const result = this.grantDistributionAccessToBucket(distributionId, actions);
+
+        // Failed to update bucket policy, assume using imported bucket
+        if (!result.statementAdded) {
+          Annotations.of(scope).addWarningV2('@aws-cdk/aws-cloudfront-origins:updateBucketPolicy',
+            'Cannot update bucket policy of an imported bucket. Set overrideImportedBucketPolicy to true or update the policy manually instead.');
```
samson-keung, line 76: I searched and cannot find info on `overrideImportedBucketPolicy`. Was the plan to have it as a feature flag?

Imo, we should just throw the warning as feature flag adds to the code complexity.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        if (bucket.encryptionKey) {
+          // TODO: update this warning
+          Annotations.of(scope).addWarningV2('@aws-cdk/aws-cloudfront-origins:updateKeyPolicy',
+            'If you run into a circular dependency issue during deployment, please refer to README for instructions.');
+        }
```
samson-keung, line 83: This warning probably will leave users scratching their head because when they see this warning, they wouldn't have seen the circular dependency error.

We should update this to change the key policy. It will allow users to have other workarounds.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: Ah yes it was part of the previous design, I'll remove since we are not updating policies of imported buckets.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: Do you mean add to the key policy here and let them run into the circular dependency issue? Then at that time they would see the warning message and follow it to configure a workaround

packages/aws-cdk-lib/aws-cloudfront/lib/origin.ts
```diff
@@ -119,6 +126,11 @@ export interface OriginBindOptions {
    * as assigned by the Distribution this Origin has been used added to.
    */
   readonly originId: string;
+
+  /**
+   * The identifier of the Distribution this Origin is used for.
+   */
+  readonly distributionId: string;
```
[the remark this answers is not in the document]
gracelu0: Yes good catch, will update to be optional

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * Properties for configuring a S3 origin with OAC
+ */
+export interface S3BucketOriginWithOACProps extends S3BucketOriginBaseProps {
+  /**
+  * An optional Origin Access Control
+  * @default - an Origin Access Control will be created.
+  */
+  readonly originAccessControl?: cloudfront.IOriginAccessControl;
```
samson-keung, line 27: I found that this interface follows this "extends path":
`S3BucketOriginWithOACProps -> S3BucketOriginBaseProps -> cloudfront.OriginProps -> OriginOptions`

And in `OriginOptions`, there is a prop `originAccessControlId`.  So the user can do:
```
origins.S3BucketOrigin.withOriginAccessControl(bucket, {
          originAccessControlId: 'oacId123',
          originAccessControl: new cloudfront.S3OriginAccessControl(stack, 'MyOac'),
        })
```
Then in the CFN template, the `originAccessControlId` will be ignored.

gracelu0: Good catch, we should remove `originAccessControlId` from `OriginOptions` and only allow user to configure OAC with object of type `IOriginAccessControl`. This follows the design guidelines to use strong typing when referencing resources: https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md#construct-interface.

PR #18989, opened 2022-02-16
fix(s3-notifications): notifications allowed with imported kms keys

markussiebert:
----
fixes: #18988

If you add an sqs queue as notification target to an s3 bucket, and this sqs queue is encrypted with an imported kms IKey, the stack won't synthesize. Instead of failing, it should warn the user, that it can not ensure the correct kms key policy permissions.

This fix will solve this.

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #31155, opened 2024-08-19
fix(s3-notifications): fixing circular dependency when Bucket and SQS are encrypted by same KMS is used for s3 notification

### LINKED ISSUE #3067: Circular dependency on s3 notification to a destination
**Note: for support questions, please first reference our [documentation](https://docs.aws.amazon.com/cdk/api/latest), then use [Stackoverflow](https://stackoverflow.com/questions/ask?tags=aws-cdk)**. This repository's issues are intended for feature requests and bug reports.

* **I'm submitting a ...**
  - [ ] :beetle: bug report

* **What is the current behavior?**
*If the current behavior is a :beetle:bug:beetle:: Please provide the steps to reproduce*
```
import cdk = require('@aws-cdk/cdk');
import s3 = require('@aws-cdk/aws-s3')
import sqs = require("@aws-cdk/aws-sqs");
import {SqsDestination} from "@aws-cdk/aws-s3-notifications";

export class Test extends cdk.Construct {
    constructor(scope: cdk.Construct, id: string) {
        super(scope, id);
        
        const queue = new sqs.Queue(this, 'testQueue', {queueName: 'testQueue'})
        const bucket = new s3.Bucket(this, 'ankag-tstbuckt', {bucketName: 'ankag-tstbuckt'})

        bucket.addObjectCreatedNotification(new SqsDestination(queue));
    }
}
```

Exception:
Circular dependency between resources: [Testankagtstbuckt88871975, TestankagtstbucktNotifications3AD3AE7F, TesttestQueuePolicy250852E7]

* **What is the expected behavior (or behavior of feature suggested)?**

s3 notification should be created to sqs without circular dependency exception

* **What is the motivation / use case for changing the behavior or adding this feature?**
This is a bug

* **Please tell us about your environment:**
  
  - CDK CLI Version: 0.33
  - Module Version: 0.33
  - OS: [all | Windows 10 | OSX Mojave | Ubuntu | etc... ]
  - Language: [all | TypeScript | Java | Python ] TypeScript

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. associated pull-request, stackoverflow, gitter, etc)

CDK 0.32 is working fine but when upgrading to CDK 0.33 we had to introduce a SqsDestination and thats when we start seeing this issue

stm29:
### Issue #3067

Closes #3067

### Reason for this change
- we can't `cdk deploy`, when we use same same KMS for S3 and SQS.
- This was partially addressed on https://github.com/aws/aws-cdk/pull/18989 , rather than fixing, this PR adds a warning when policy is unable to add to KMS.
  -  But issue arises when the policy is added. It creates **Circular dependency**.
  - When, we didn't add this permission, then we can't send message event to SQS.

So, we need to draw a fair line on how to handle this. approaches considered,
1. Issue arises, when we are constraining the polices in KMS. so what happens is, KMS depends on Bucket, Bucket depends on KMS, exactly explained by https://github.com/aws/aws-cdk/issues/3067#issuecomment-1312156955 
  - We can just expand the scope of Policy to Service level (s3.amazonaws.com), rather than constraining with Resource level (BucketName), this may not be security wise advised, but this better than allowing user to deploy than just showing circular dependency.
2. We can optionally, ask use wether to add this relaxed permission added. If doesn't wish to add this, he can opt out and add his constrained policy.
3. In both the cases we will show FAIR warning on how this will result in Security / Runtime Exception (Circular Dependency).

### Description of changes
- Added optional parameter `shouldAddGlobalS3PermissionToKMSandSQS` , defaulted to `true`, as it will be backward compatible
- Based on this Flag, we are showing warning during `cdk synth`
  -  **Circular Dependency Violation** when `false` is provided, as they will try to add imported Bucket Value to add permission
  - **Security Warning** - as Service Level permission is granted 

- Unit Tests
- Integration Tests

### Checklist

----


packages/aws-cdk-lib/aws-s3-notifications/lib/sqs.ts
```diff
@@ -8,19 +8,36 @@ import { Annotations } from '../../core';
...
+      // If user is okay to add global permission, we will add and show security warning
+      this.queue.grantSendMessages(new iam.ServicePrincipal('s3.amazonaws.com'));
+
+      const warning = 'Consider passing \`shouldAddGlobalS3PermissionToKMSandSQS:false\` and add restricted permission with condition - \'aws:SourceArn\': bucket.bucketArn';
+      Annotations.of(_scope).addWarningV2('@aws-cdk/aws-s3-notifications:securityWarning', warning);
+    } else {
+      // If user didn't want this to be Global, then we will show CRITICAL warning calling out to add
+      // the required permission without using imported values.
+      const warning = 'You have opted out to add global permission for KMS & SQS Key Policy. Consider manually adding kms.grantEncryptDecrypt(), queue.grantSendMessages()';
+      Annotations.of(_scope).addWarningV2('@aws-cdk/aws-s3-notifications:sqsKMSPermissionsNotAdded', warning);
```
laurelmay, line 39: If we're going to tell the user that this is a better way of implementing this, it's probably best if we have a new integration test that does it and ensures that it works. I'm not entirely sure that I understand how moving this condition out of the `SqsDestination.bind` method and making the user do it manually actually resolves the circular dependency while still allowing the more specific policy.

stm29: > I'm not entirely sure that I understand how moving this condition out of the SqsDestination.bind method and making the user do it manually actually resolves the circular dependency while still allowing the more specific policy

So, Given the Condition, we can't depend on Class's members to get the **name**, to add the policy like us do in CDK by itself.
There are 2 ways,
1. User Needs to add the permissions by himself, as a constant defined by himself (**this will relax the Circular dependency between both**), Only thing I am seeing here is, At this moment, we can't make CDK to do this automatically.
1. Allow Global access.

What are we trying to resolve here is, "cdk deploy" is even failing at this moment.
@laurelmay Please let me know wether there is any suggested alternatives.

PR #31168, opened 2024-08-21
More unit tests for S3BucketOrigin

samson-keung:
Adding more unit tests and more verifications. Added unit test to test imported buckets as well.


gracelu0 (approved):
LGTM, I'll update the imported bucket test as part of addressing the comment about the outdated warning message

Issue #31160, opened 2024-08-20
appsync: API level cache config


TonySherman:
### Describe the feature

The `aws_appsync` module does not include any L2 constructs to enable caching at the API level. There is a `CachingConfig` construct but that is specifically for resolvers and cannot be enabled if the API does not have caching enabled itself.

### Use Case

There is no way to enable caching on a graphql API outside of cfn constrcuts. I would like to enable this directly in the GraphqlApi construct or via it's own L2 construct.

### Proposed Solution

Ideally, it could be configured directly in the GraphqlApi construct:
```typescript
const api = new appsync.GraphqlApi(this, 'EventBridgeApi', {
  name: 'EventBridgeApi',
  definition: appsync.Definition.fromFile(path.join(__dirname, 'appsync.eventbridge.graphql')),
  cacheConfig: appsync.ApiCacheConfig(...)
});
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2.153.0

### Environment details (OS name and version, etc.)

MacOS 14.6.1, Node v18

pahud:
Thank you for your feature request.

Looking at the CFN doc for `AWS::AppSync::GraphQLApi`, I didn't see how an API could configure with cache config.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-graphqlapi.html

Can you elaborate more on how a cacheConfig should be passed to a GraphQLApi?

TonySherman:
@pahud - It's possible that it cannot be added directly to the `GraphQLApi` construct, but there still is no L2 construct for `CfnApiCache`. Maybe the usage would actually be:

```typescript
const api = new appsync.GraphqlApi(this, 'EventBridgeApi', {
  name: 'EventBridgeApi',
  definition: appsync.Definition.fromFile(path.join(__dirname, 'appsync.eventbridge.graphql')),
});

const cache = new appsync.ApiCache(
   api_id: api.api_id
   cache_behavior: appsync.CacheBehavior.FULL_REQUEST_CACHING,
   type: appsync.CacheType.LARGE,
   ...
)
```

pahud:
I think we can simply create a `addCache` method for the GraphqlApi L2 so it would be like

```ts
const api = new appsync.GraphqlApi(this, 'EventBridgeApi', {
  name: 'EventBridgeApi',
[...]

PR #31174, opened 2024-08-21
feat(appsync): add ApiCache L2 construct and addCache method

TonySherman:
### Issue # (if applicable)
#31160
Closes #<issue number here>.

### Description of changes
Adds L2 Construct for ApiCache on AppSync graphQL Apis
and a method to add a cache to existing graphQL Api constructs.

### Checklist

----


packages/aws-cdk-lib/aws-appsync/lib/api-cache.ts
```diff
@@ -0,0 +1,136 @@
...
+export interface ApiCacheBaseProps {
+  /**
+   * Caching behavior
+   */
+  readonly apiCachingBehavior: CacheBehavior;
+
+  /**
+   * At-rest encryption flag for cache
+   *
+   * @default - idk
```
TonySherman, line 82: I could not find the default behavior of the various optional properties for this resource

TonySherman: verified these by deploying without setting the values.

packages/aws-cdk-lib/aws-appsync/lib/api-cache.ts
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * The GraphQL API ID.
+   */
+  readonly apiId: string;
+}
+
+/**
+ * Cache for a GraphQL API
+ */
+export class ApiCache extends Construct {
```
pahud, line 121: L2 construct should extends `Resource` implements `IApiCache`. You should export `IApiCache` as well.

packages/aws-cdk-lib/aws-appsync/lib/api-cache.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/**
+ * Properties for an Api Cache
+ */
+export interface ApiCacheProps extends ApiCacheBaseProps {
+  /**
+   * The GraphQL API ID.
+   */
+  readonly apiId: string;
```
pahud, line 115: I think it should be

```
readonly api: IGraphqlApi;
```

pahud (dismissed):
If you are implementing the ApiCache L2, I suggest you first read the design guidelines [here](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md) and read other existing L2 constructs to learn the patterns.

Generally, if you are creating a L2 construct called `Foo`. You should:

1. Define `IFoo` interface for it that extends `IResource`.
2. Define `Foo` construct class that extends `Resources` implements `IFoo`
3. Define `FooProps` as the construct props for `Foo`

In addition to the design guides and existing L2 constructs, I would recommend you read [this blog post](https://community.aws/content/2kcTpCDCTOTYOY8UoWGGpttA6t8/building-agentic-code-interpreter-with-aws-cdk) in which I shared how I built an Agent L2 construct.

pahud:
Some suggested changes here:

- Validation for Ttl property:

1. The CloudFormation documentation specifies that the Ttl property should be a number between 1 and 3,600 (inclusive). You can add some validation to ensure that the provided ttl value is within this range.
You can add a check in the constructor of the ApiCache class to validate the ttl value:

```ts
constructor(scope: Construct, id: string, props: ApiCacheProps) {
  super(scope, id);

  if (props.ttl < 1 || props.ttl > 3600) {
    throw new Error('Ttl value must be between 1 and 3,600 seconds');
  }

  new CfnApiCache(this, 'Resource', {
    // Rest of the code
  });
}
```

TonySherman:
> Some suggested changes here:
> 
> * Validation for Ttl property:
> 
> 1. The CloudFormation documentation specifies that the Ttl property should be a number between 1 and 3,600 (inclusive). You can add some validation to ensure that the provided ttl value is within this range.
>    You can add a check in the constructor of the ApiCache class to validate the ttl value:
> 
> ```ts
> constructor(scope: Construct, id: string, props: ApiCacheProps) {
>   super(scope, id);
> 
>   if (props.ttl < 1 || props.ttl > 3600) {
>     throw new Error('Ttl value must be between 1 and 3,600 seconds');
>   }
> 
>   new CfnApiCache(this, 'Resource', {
>     // Rest of the code
>   });
> }
> ```

Thanks! Sorry for the delay on updates to this work. Got sidetracked with some other stuff and I'll be getting back to this shortly.

PR #31175, opened 2024-08-21
fix(cognito-identitypool): allow default roles to have their rules added

### LINKED ISSUE #27411: aws-cognito-identitypool-alpha: Can't add rules for the default authenticated role
### Describe the bug

I am trying to configure an `IdentityPool` with `google` as an `authenticationProviders` and add some custom `rules` through `roleMappings`.

It's neither possible directly through the initiation of the `IdentityPool`, nor later through the `addRoleMappings()` method.

### Expected Behavior

To be able to add custom `rules` to the default roles during initiation of the `IdentityPool`, or add them later with the `addRoleMappings()` method

### Current Behavior

Getting errors that the `DefaultRoleAttachment` failed to update, so can't create a new one

```
The following resource(s) failed to create: [identitypoolRoleMappingAttachmentXXXXXXXXX]. The following resource(s) failed to update: [identitypoolDefaultRoleAttachmentYYYYYYYY]. 
```

### Reproduction Steps

Since this isn't possible

```javascript
new IdentityPool(stack, 'identity-pool', {
  allowUnauthenticatedIdentities: false,
  authenticationProviders: {
    google: {
      clientId: '12345678012.apps.googleusercontent.com',
    },
  },
  roleMappings: [
    {
      providerUrl: IdentityPoolProviderUrl.GOOGLE,
      rules: [
        {
          claim: 'sub',
          claimValue: '12345678012',
          mappedRole: 'default-auth-role-reference', // 👈 it would be nice if we could reference the default roles directly during initiation of the identitypool
         },
      ],
    },
  ],
})
```

I instead tried this

```javascript
const identityPool = new IdentityPool(stack, 'identity-pool', {
  allowUnauthenticatedIdentities: false,
  authenticationProviders: {
    google: {
      clientId: '12345678012.apps.googleusercontent.com',
    },
  },
})

identityPool.addRoleMappings({
  providerUrl: IdentityPoolProviderUrl.GOOGLE,
  rules: [
    {
      claim: 'sub',
      claimValue: '12345678012',
      mappedRole: identityPool.authenticatedRole,
    },
  ],
})
```

### Possible Solution

I don't have a technical solution to solve this, but I managed to get it working in a not so pretty way

```javascript
// 1. create a new admin role
const adminRole = new Role(stack, 'admin-role', {
  assumedBy: new WebIdentityPrincipal('cognito-identity.amazonaws.com', {
    StringEquals: {
      'cognito-identity.amazonaws.com:aud': 'something-silly', // 👈 we unfortunately don't know the poolId yet
    },
    'ForAnyValue:StringLike': {
      'cognito-identity.amazonaws.com:amr': 'authenticated',
    },
  }),
})

// 2. create the identity pool
const identityPool = new IdentityPool(stack, 'identity-pool', {
  allowUnauthenticatedIdentities: false,
  authenticationProviders: {
    google: {
      clientId: '12345678012.apps.googleusercontent.com',
    },
  },
  roleMappings: [
    {
      providerUrl: IdentityPoolProviderUrl.GOOGLE,
      rules: [
        {
          claim: 'sub',
          claimValue: '12345678012',
          mappedRole: adminRole, // 👈 it is possible to reference the adminRole during initiation of the identityPool
        },
      ],
    },
  ],
})

// 3. create a new policy statement to the admin role, to include the poolId
const adminRolePolicyStatement = new PolicyStatement({
  principals: [new WebIdentityPrincipal('cognito-identity.amazonaws.com')],
  actions: ['sts:AssumeRoleWithWebIdentity'],
  conditions: {
    StringEquals: {
      'cognito-identity.amazonaws.com:aud': identityPool.identityPoolId, // 👈 only now can we reference the poolId
    },
    'ForAnyValue:StringLike': {
      'cognito-identity.amazonaws.com:amr': 'authenticated',
    },
  },
})

// 4. attach the policy statement to the admin role
adminRole.assumeRolePolicy?.addStatements(adminRolePolicyStatement)
```

This approach works, but it's not very intuitive and it does have two undesirable outcomes
1. we get an extra role (`adminRole`) when we could have just used the default `AuthenticatedRole`
2. we get an unnecessary policy statement on the admin role
```json
{
  "Version": "2012-10-17",
  "Statement": [
    // this statement is not needed
    {
      "Effect": "Allow",
      "Principal": {
        "Federated": "cognito-identity.amazonaws.com"
      },
      "Action": "sts:AssumeRoleWithWebIdentity",
      "Condition": {
        "StringEquals": {
          "cognito-identity.amazonaws.com:aud": "something-silly"
        },
        "ForAnyValue:StringLike": {
          "cognito-identity.amazonaws.com:amr": "authenticated"
        }
      }
    },
    {
      "Effect": "Allow",
      "Principal": {
        "Federated": "cognito-identity.amazonaws.com"
      },
      "Action": "sts:AssumeRoleWithWebIdentity",
      "Condition": {
        "StringEquals": {
          "cognito-identity.amazonaws.com:aud": "[AWS:REGION]:[POOLID]"
        },
        "ForAnyValue:StringLike": {
          "cognito-identity.amazonaws.com:amr": "authenticated"
        }
      }
    }
  ]
}

```

_No response_

### CDK CLI Version

2.99.1

### Framework Version

_No response_

### Node.js Version

v20.8.0

### OS

MacOS Ventura 13.5.2

### Language

Typescript

### Language Version

5.2.2

### Other information

_No response_

Leo10Gama:
### Issue # (if applicable)

Closes #27411 

### Reason for this change

Previously, it was not possible to update the default-created roles made when generating a new `IdentityPool` without hacky, unintuitive workarounds. This mainly stems from an error being thrown when there are multiple `CfnIdentityPoolRoleAttachment` resources present in the stack, as the `.addRoleMappings()` method previously created a new one.

### Description of changes

Since having multiple `AWS::Cognito::IdentityPoolRoleAttachment` resources for one identity pool causes errors, the methods that created new ones have been altered in such a way that they only manipulate the existing attachment resource. This is supported by the fact that one identity pool may have two separate role mappings that may conflict with each other, so it is necessary that there is only one attachment. 

Now, the `addRoleMappings()` method will append the role mappings to the `IdentityPool`'s newly-added `roleMappings` attribute. However, this can only be added if a new property, `createDefaultRoleAttachment`, is set to false, preventing a role attachment from being created. Thus, when set to false, this indicates that the user wants to create their own `IdentityPoolRoleAttachment` L2 construct, and will be adding role mappings of their own. If `addRoleMappings()` is called after an attachment has already been created, it will throw an error, prompting the user to prevent creation of a default role attachment and to create their own.

To ensure the issue has been dealt with, the integration test now explicitly makes a call to the `.addRoleMappings()` method, passing in one of the default-created roles to update. This is the exact behaviour described in the issue. In addition, the unit tests have been updated to reflect the new behaviour of the role attachment resource.

### Checklist

----

### DIFF
--- a/packages/@aws-cdk/aws-cognito-identitypool-alpha/lib/identitypool.ts
+++ b/packages/@aws-cdk/aws-cognito-identitypool-alpha/lib/identitypool.ts
@@ -488,6 +488,7 @@ export class IdentityPool extends Resource implements IIdentityPool {
    * Assigns an IdentityPoolRoleAttachment to the IdentityPool, if it does not
    * already have one assigned. In order to be assigned, the attachment must have
    * the same Identity Pool ID.
+   * @internal
    * @param roleAttachment The new IdentityPoolRoleAttachment object to assign to
    * the IdentityPool.
    */
--- a/packages/@aws-cdk/aws-cognito-identitypool-alpha/test/identitypool.test.ts
+++ b/packages/@aws-cdk/aws-cognito-identitypool-alpha/test/identitypool.test.ts
@@ -972,7 +972,7 @@ describe('role mappings', () => {
             claimValue: 'admin',
             mappedRole: idPool.authenticatedRole,
           },
-        ]
+        ],
       });
     }).toThrow(Error);
   });


packages/@aws-cdk/aws-cognito-identitypool-alpha/lib/identitypool.ts
```diff
@@ -388,9 +396,14 @@ export class IdentityPool extends Resource implements IIdentityPool {
...
-   * Running count of added role attachments
+   * The list of role mappings for the Identity Pool
+   */
+  public roleMappings: IdentityPoolRoleMapping[];
+
+  /**
+   * Role attachment L2 construct for the Identity Pool
    */
-  private roleAttachmentCount: number = 0;
+  private roleAttachment: IdentityPoolRoleAttachment | null = null;
```
paulhcsun, line 406: I don't think this union type will be supported by JSII. Just make this optional and set it in the constructor.

paulhcsun (changes requested):

Leo10Gama:
This fix was originally being made under the assumption that only one `AWS::Cognito::IdentityPoolRoleAttachment` resource could exist at a time, as that seemed to be the underlying bug. However, after some investigation, this assumption was wrong. I'll apply a more correct fix in another PR.

PR #31188, opened 2024-08-22
chore(cloudfront-origins): grant key permissions to distribution 

gracelu0:
Changes:
- grant key permissions to distribution and output info message about circular dependency
- make `distributionId` optional
- rename some variables for clarity

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
+++ b/packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
@@ -79,7 +79,8 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
         // Failed to update bucket policy, assume using imported bucket
         if (!bucketPolicyResult.statementAdded) {
           Annotations.of(scope).addWarningV2('@aws-cdk/aws-cloudfront-origins:updateImportedBucketPolicy',
-            'Cannot update bucket policy of an imported bucket. You may need to update the policy manually instead.');
+            'Cannot update bucket policy of an imported bucket. You may need to update the policy manually instead.\n' +
+            'See the "Using pre-existing S3 buckets" section of module\'s README for more info.');
         }
 
         if (bucket.encryptionKey) {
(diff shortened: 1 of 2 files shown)


packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -67,19 +73,37 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
...
-          // TODO: update this warning
-          Annotations.of(scope).addWarningV2('@aws-cdk/aws-cloudfront-origins:updateKeyPolicy',
-            'If you run into a circular dependency issue during deployment, please refer to README for instructions.');
+          let bucketName = bucket.bucketName;
+          if (Token.isUnresolved(bucket.bucketName)) {
+            bucketName = JSON.stringify(Tokenization.resolve(bucket.bucketName, {
+              scope,
+              resolver: new DefaultTokenResolver(new StringConcat()),
+            }));
+          }
```
samson-keung, line 93: Do you have links to examples of this trick when the code is doing `JSON.stringify(Tokenization.resolve ... `? Trying to understand what it does. Does it resolve to the logical ID?

gracelu0: example of usage in iam module: https://github.com/aws/aws-cdk/blob/af50620cdd2eaa87c24bf26d2e3ca3954a1523d6/packages/aws-cdk-lib/aws-iam/lib/private/util.ts#L24-L31

Output from when I tested it locally is like this:
```
[Info at /MigrationOacStack/MyDistribution/Origin2] Granting OAC permissions to access KMS key for S3 bucket origin {"Ref":"MyBucketF68F3FF0"} may cause a circular dependency error when this stack deploys.
The key policy references the distribution's id, the distribution references the bucket, and the bucket references the key.
See the 'Using OAC for a SSE-KMS encrypted S3 origin' section in the module README for more details.
```
So yes it resolves to the logical ID. I thought it would be helpful to include the specific resource involved in the info message, but there might be a cleaner way to do this

samson-keung: I see. Thank you. Looks good to me.

samson-keung (approved):

PR #31190, opened 2024-08-22
Unit test for S3BucketOrigin.withOriginAccessIdentity

samson-keung:
### Changes:
- Added unit tests for `S3BucketOrigin.withOriginAccessIdentity` with different use case
- Added warning inside ` S3BucketOrigin.withOriginAccessIdentity.bind` when bucket is imported. So the behaviour is same as ` S3BucketOrigin.withOriginAccessControl`

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-bucket-origin.test.ts
+++ b/packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-bucket-origin.test.ts
@@ -2,7 +2,7 @@ import { Annotations, Template } from '../../assertions';
 import * as cloudfront from '../../aws-cloudfront/index';
 import * as origins from '../../aws-cloudfront-origins';
 import * as s3 from '../../aws-s3/index';
-import { Duration, Stack } from '../../core';
+import { App, Duration, Stack } from '../../core';
 
 describe('S3BucketOrigin', () => {
   describe('withOriginAccessControl', () => {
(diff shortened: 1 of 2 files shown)


packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-bucket-origin.test.ts
```diff
@@ -467,4 +467,404 @@ describe('S3BucketOrigin', () => {
...
+        });
+      });
+
+      it('should warn user bucket policy is not updated', () => {
+        Annotations.fromStack(distributionStack).hasWarning('/distributionStack/MyDistributionA/Origin1',
+          'Cannot update bucket policy of an imported bucket. You may need to update the policy manually instead.\n' +
+          'See the "Using pre-existing S3 buckets" section of module\'s README for more info. [ack: @aws-cdk/aws-cloudfront-origins:updateImportedBucketPolicy]');
+      });
+
+      it('should create OAI in bucket stack and output it, then reference the output in the distribution stack', () => {
```
samson-keung, line 793: This is specifically for testing the cross stack logic in the OAI bind method: https://github.com/samson-keung/aws-cdk/blob/9f5c1430bec9dd41246b83085cd4350b1ab1dffb/packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts#L195-L199

Such logic is not in `withOriginAccessControl`, do we need it there?

gracelu0: We won't be able to handle cross-stack references as easily for OAC because the bucket policy references the distribution, so the solution is to create the bucket and distribution in the same stack. Versus OAI where it just references the OAI user (so creating the OAI in the bucket stack easily resolves the circular dependency).

gracelu0 (approved):

PR #31200, opened 2024-08-23
feat(scheduler): validate schedule name length

LaurenceWarne:
### Issue # (if applicable)

N/A

### Reason for this change

To catch name too long errors before hitting Cloudformation.

### Description of changes

I've added a length assertion in the `Schedule` constructor against the schedule name length, the documentation on schedule name already makes it clear that there is a limit though AFAICS this isn't validated.

Also see the Cloudformation doc here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-scheduler-schedule.html#cfn-scheduler-schedule-name.

I've added a unit test.

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-synthetics/lib/runtime.ts
+++ b/packages/aws-cdk-lib/aws-synthetics/lib/runtime.ts
@@ -222,6 +222,19 @@ export class Runtime {
   public static readonly SYNTHETICS_NODEJS_PUPPETEER_8_0 = new Runtime('syn-nodejs-puppeteer-8.0', RuntimeFamily.NODEJS);
 
   /**
+   * `syn-nodejs-puppeteer-9.0` includes the following:
+   * - Lambda runtime Node.js 20.x
+   * - Puppeteer-core version 22.12.1
+   * - Chromium version 126.0.6478.126
+   *
+   * New Features:
+   * - **Bug fixes** Bug fix to enable visual monitoring capabilities.
+   *
+   * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_Library_nodejs_puppeteer.html#CloudWatch_Synthetics_runtimeversion-nodejs-puppeteer-9.0
+   */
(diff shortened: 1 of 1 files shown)


LaurenceWarne:
Exemption Request: not really sure how to cover this with an integration test - I thought a unit test might suffice for something this small?

packages/@aws-cdk/aws-scheduler-alpha/lib/schedule.ts
```diff
@@ -324,6 +324,9 @@ export class Schedule extends Resource implements ISchedule {
     const flexibleTimeWindow = props.timeWindow ?? TimeWindow.off();
 
     this.validateTimeFrame(props.start, props.end);
+    if (props.scheduleName && !Token.isUnresolved(props.scheduleName) && props.scheduleName.length > 64) {
+      throw new Error(`schedule name '${props.scheduleName}' exceeds 64 characters in length`);
```
go-to-k, line 328: It would be user-friendly and good to display the length.

```suggestion
      throw new Error(`scheduleName cannot be longer than 64 characters, got: ${props.scheduleName.length}`);
```

packages/@aws-cdk/aws-scheduler-alpha/test/schedule.test.ts
```diff
@@ -214,5 +214,16 @@ describe('Schedule', () => {
         });
       }).toThrow('The provided duration must be between 1 minute and 1440 minutes, got 0');
     });
+
+    test('throw error when schedule name exceeds 64', () => {
```
go-to-k, line 218:
```suggestion
    test('throw error when scheduleName exceeds 64 characters', () => {
```

go-to-k (changes requested):
Thanks for the PR. I just made a few minor comments.

go-to-k:
As this issue is not about a bug, the PR title may be good for `chore`.

LaurenceWarne:
Thanks for the review, I've updated the title and made the requested changes.

go-to-k (approved):
Thanks, LGTM.

kaizencc (approved):
I think this is technically a feat! validation is a feature of CDK so I think this counts. I also want this to show up in the changelog

PR #31207, opened 2024-08-23
add unit tests for s3 static website origin

gracelu0:
Changes:
- unit tests for `S3StaticWebsiteOrigin` with default and custom props

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-static-website-origin.test.ts
+++ b/packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-static-website-origin.test.ts
@@ -5,66 +5,91 @@ import { Duration, Stack } from '../../core';
 
 describe('S3StaticWebsiteOrigin', () => {
   let stack: Stack;
-  let bucket: s3.Bucket;
 
-  beforeEach(() => {
-    stack = new Stack();
-    bucket = new s3.Bucket(stack, 'myStaticWebsiteBucket', {
-      websiteIndexDocument: 'index.html',
+  describe('test setup using new bucket', () => {
+    let bucket: s3.Bucket;
+
+    beforeEach(() => {
+      stack = new Stack();
+      bucket = new s3.Bucket(stack, 'myStaticWebsiteBucket', {
+        websiteIndexDocument: 'index.html',
+      });
     });
-  });
-  test('creates a S3 static website origin with default properties set', () => {
-    const origin = new origins.S3StaticWebsiteOrigin({ bucket });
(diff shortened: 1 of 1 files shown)


packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-static-website-origin.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  });
+  test('creates a S3 static website origin with default properties set', () => {
+    const origin = new origins.S3StaticWebsiteOrigin({ bucket });
+    const originBindConfig = origin.bind(stack, { originId: 'StackOrigin12AB34569' });
+
+    expect(originBindConfig.originProperty).toEqual({
+      id: 'StackOrigin12AB34569',
+      domainName: bucket.bucketWebsiteDomainName,
+      customOriginConfig: {
+        originSslProtocols: ['TLSv1.2'],
```
samson-keung, line 24: If S3 website endpoint support HTTP only and not HTTPS, why do we need this `originSslProtocols` and what would it do?

packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-static-website-origin.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      connectionAttempts: 2,
+      connectionTimeout: 5,
+      originCustomHeaders: [
+        {
+          headerName: 'X-Custom-Header',
+          headerValue: 'custom-value',
+        },
+      ],
+    });
+  });
```
samson-keung, line 69: Should we have a test case for using imported bucket? to make sure the code works, no errors, setting props correctly etc.

samson-keung (changes requested):

packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-static-website-origin.test.ts
gracelu0: Good point. We don't need it but it's currently set by default in the constructor for `HttpOrigin` [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-cloudfront-origins/lib/http-origin.ts#L74). It does not have any effect on the deployment though and the origin protocol policy is correctly set as HTTP only.

packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-static-website-origin.test.ts
gracelu0: I'll add this

samson-keung (approved):

PR #31212, opened 2024-08-24
fix(iam): `Role.addManagedPolicy()` does not work for imported roles `IRole` #8307

### LINKED ISSUE #8307: Role.addManagedPolicy does not work for imported roles
### The Scenario
TLDR; aws_iam.Role.attachManagedPolicy does not attach the specified managed policy to the role, aws_iam.ManagedPolicy.attachToRole does.

I have two stacks: one stack deploys roles another deploys an application stack (both are in the same account).  The roles stack is deployed first.

In the application stack, I would like to attach a managed policy to a role in the first.

Attempting to use aws_iam.Role.attachManagedPolicy does not create the association but aws_iam.ManagedPolicy.attachToRole does.

### Environment

  - **CDK CLI Version:** 1.39.0 (build 5d727c1) 
  - **Module Version:** 1.38.0
  - **Node.js Version:** v14.0.0 
  - **OS:** macOS Mojave 10.14.6 (18G4032) 
  - **Language:** Typescript and Python 

#### Steps to reproduce
1. Deploy one stack with an IAM role
1. In a second stack create a managed policy
1. In the second stack Import the role using aws_iam.Role.fromRoleArn (importedRole)
1. In the second stack attempt to add the managed policy to importedRole via importedRole.attachManagedPolicy

##### Observed
1. cdk synth does not show any associations made between the role and the managed policy
1. using ManagedPolicy.attachToRole works however

##### Expected
1. cdk synth (and subsequently) cdk deploy should associate the role and managed policy when Role.attachManagedPolicy is used

stm29:
### Issue #8307

Closes #8307

### Reason for this change
- This addManagedPolicy() just does nothing, neither throws warning nor fails.

### Description of changes
- This implements **addManagedPolicy()** for **ImportedRole**
- This throws Error, when `IRole` is used with `IManagedPolicy`, due to constrains

### Approaches decided
- we can't change **addManagedPolicy(policy: IManagedPolicy)** to **addManagedPolicy(policy: ManagedPolicy)** (i.e., to ManagedPolicy) , like how **attachInlinePolicy(policy: Policy)**, it will break for almost many customers.
  - https://github.com/aws/aws-cdk/blob/823ff6e03899f790a4cb1c43f92a02cc906ac356/packages/aws-cdk-lib/aws-iam/lib/identity-base.ts#L17-L21
  - So, we need to handle this gracefully when accepting only **ManagedPolicy**.
- We can't use IRole & IManagedPolicy to do the work at the same time, so we need to allow ManagedPolicy in params, and throw Error when IManagedPolicy is being sent, because of how **attachToRole()** will be used by **addManagedPolicy()**
- #### This PR Implements follows,
  - enables **addManagedPolicy()**
  - Accepts **ManagedPolicy**
  - Throws understandable Error, when `IRole` is used with `IManagedPolicy`.

- unit & integration tests

### Checklist

----


epoctic (approved):
Changes seem reasonable to me.

stm29:
Can I get one more community review?

stm29:
Communicated in #contributing channel. waiting for one more community review.

packages/aws-cdk-lib/aws-iam/lib/private/imported-role.ts
```diff
@@ -74,8 +74,15 @@ export class ImportedRole extends Resource implements IRole, IComparablePrincipa
     }
   }
 
-  public addManagedPolicy(policy: IManagedPolicy): void {
-    Annotations.of(this).addWarningV2('@aws-cdk/aws-iam:importedRoleManagedPolicyNotAdded', `Not adding managed policy: ${policy.managedPolicyArn} to imported role: ${this.roleName}`);
+  public addManagedPolicy(policy: ManagedPolicy): void {
+    try {
+      policy.attachToRole(this);
```
tmokmss, line 79: We should not change the signature (IManagedPolicy -> ManagedPolicy) because technically it is a breaking change.

Instead I'd write something like `(policy as ManagedPolicy).attachToRole`.

Or use type predicate:

```ts
    const isManagedPolicy = (x: IManagedPolicy): x is ManagedPolicy => {
      return (x as any).attachToRole !== undefined;
    };
    if (isManagedPolicy(policy)) {
      policy.attachToRole(this);
    }
```

This way we don't need catch the error for type check.

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource-handler.ts
```diff
@@ -63,14 +63,20 @@ export class NotificationsResourceHandler extends Construct {
       assumedBy: new iam.ServicePrincipal('lambda.amazonaws.com'),
     });
 
-    this.role.addManagedPolicy(
-      iam.ManagedPolicy.fromAwsManagedPolicyName('service-role/AWSLambdaBasicExecutionRole'),
-    );
```
tmokmss, line 68: Why are we changing this file?

packages/aws-cdk-lib/aws-iam/lib/private/imported-role.ts
```diff
@@ -74,8 +74,15 @@ export class ImportedRole extends Resource implements IRole, IComparablePrincipa
...
+    } catch (e) {
+      if (e instanceof Error && e.message === 'policy.attachToRole is not a function') {
+        throw new Error('Can\'t Combine IRole with IManagedPolicy. use ManagedPolicy directly.');
```
tmokmss, line 82: Throwing an error from this function is also a breaking change. Can we just add a warning as the previous behavior?

tmokmss (changes requested):
Thank you for the fix! I left some comments :)

packages/aws-cdk-lib/aws-iam/lib/private/imported-role.ts
stm29: Resolved with **type predicate** and changed as **Warning** instead of **Error**.
Removed ```try ... catch```

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource-handler.ts
stm29: because `ManagedPolicy.addManagedPolicy` is being used in that file also, till now which is also **doing Nothing.**
```
      this.role.addManagedPolicy(
        iam.ManagedPolicy.fromAwsManagedPolicyName('service-role/AWSLambdaBasicExecutionRole'),
      );
```

packages/aws-cdk-lib/aws-iam/lib/private/imported-role.ts
stm29: Done.

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource-handler.ts
tmokmss: I see. Maybe the maintainer will request you to move the fix to another PR because it is a related but different issue.

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource-handler.ts
```diff
@@ -63,14 +63,20 @@ export class NotificationsResourceHandler extends Construct {
...
     this.role.addToPrincipalPolicy(new iam.PolicyStatement({
       actions: ['s3:PutBucketNotification'],
       resources: ['*'],
     }));
 
+    if (props.role) {
+      cdk.Annotations.of(this).addWarningV2('@aws-cdk-lib/aws-s3:missingLambdaExecutionRole', 'CDK doesn\'t add service-roles if `notificationsHandlerRole` is provided, Make Sure Imported Role have AWSLambdaBasicExecutionRole added');
+    }
+    if (!props.role) {
```
tmokmss, line 74: I'm not sure whether the conditions are correct. What happens when a user passed a Role construct? e.g. 
```ts
{
    role: new iam.Role(this, 'Role', {
      assumedBy: new iam.ServicePrincipal('lambda.amazonaws.com'),
    })
}
```

This should just add a managed policy to the role.

tmokmss: Thanks for the clarification. Do we really need two warning for this? I feel one warning is enough. Then we can keep the current code as-is for this file.

packages/aws-cdk-lib/aws-iam/lib/private/imported-role.ts
```diff
@@ -75,7 +75,20 @@ export class ImportedRole extends Resource implements IRole, IComparablePrincipa
...
+    const isManagedPolicy = (x: IManagedPolicy): x is ManagedPolicy => {
+      return (x as ManagedPolicy).attachToRole !== undefined;
+    };
+
+    if (isManagedPolicy(policy)) {
+      policy.attachToRole(this);
+    } else {
+      Annotations.of(this).addWarningV2(
+        '@aws-cdk/aws-iam:IRoleCantBeUsedWithIManagedPolicy',
+        'Can\'t combine IRole with IManagedPolicy. Use ManagedPolicy directly.',
```
tmokmss, line 89: We can add more information like `${policy.managedPolicyArn} to imported role: ${this.roleName}` as in the original code to make it clearer why and where the warning is emitted.

stm29: Corrected with One Common message.

packages/aws-cdk-lib/aws-s3/lib/notifications-resource/notifications-resource-handler.ts
tmokmss: Don't forget to revert this change :)

tmokmss (approved):
LGTM, thanks!

Leo10Gama:
Hi @stm29 thanks for your contribution! Reading through the changes, I'm trying to understand the nature of the changes being made. The PR description mentions you're adding functionality to `IRole`, but it looks like you're adding it to `ImportedRole`. Could you please update the description to better reflect this?

I also did some digging through the rest of the codebase, and it seems like some of the other extensions of `IRole` have other implementations of `addManagedPolicy()`, like [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-iam/lib/private/precreated-role.ts#L110-L112):
```typescript
  public addManagedPolicy(policy: IManagedPolicy): void {
    this.managedPolicies.push(policy);
  }
```
Would a change like this work instead of the more complex logic?

It also looks like the changes in the S3 notifications resource are minor, so I would agree that moving that change to another PR and opening a separate issue would be best, as long as the tests that currently exist for it still pass with this change. Though I'm not quite sure what the issue is in this case. It seems like a few other libraries already call on the `addManagedPolicy()` method, so is there something specific to S3 in this regard?

stm29:
@Leo10Gama , Thanks for the review.
>  I'm trying to understand the nature of the changes being made. The PR description mentions you're adding functionality to `IRole`, but it looks like you're adding it to `ImportedRole`. Could you please update the description to better reflect this?

Updated this with correct `ImportedRole`

----
> Would a change like this work instead of the more complex logic?

Indeed Logic is simple, following is the working Logic which adds Policy to the Role. and this is same as how other Implementations are already in place.
```
policy.attachToRole(this);
```
Other `if ... else` and `type predicate` is to make sure Implementation is working as expected.

-----
> It seems like a few other libraries already call on the addManagedPolicy() method, so is there something specific to S3 in this regard?

Sorry for the confusion, There is no change in `notifications-resource-handler.ts` file, I have reverted to `main`.

**Quick Question**: Should we need to add Unit Tests in all the possible occurrence of IManagedPolicy with IRole, or Just UnitTests in `packages/aws-cdk-lib/aws-iam/test/imported-role.test.ts` is sufficient?
If UnitTests in `imported-role.test.ts` is sufficient, then `notification.test.ts` can also be reverted to `main`

----

GavinZZ (approved):

GavinZZ (approved):

Issue #30762, opened 2024-07-05
Tracking: Full Control VPC


shikha372:
### Overview

Full Control VPC provides feasibility to add custom route table and subnets to your network with the support of creating your VPC with Ipv6 Addresses.

The main components of this VPC construct is:

- Add Secondary Address to the VPC
- IP address allocation using AWS IPAM
- Subnets With Custom IP range
- Custom Routing Support with Gateways

### Link to the service’s CDK Construct Library API reference page.

https://github.com/aws/aws-cdk-rfcs/blob/57fd92a7f20e242b96885264c12567493f5e867f/text/0507-subnets.md

### Maturity: CloudFormation Resources Only

Experimental

### Implementation

###Features and Support

### VPC

| Done | Item | Needed for Graduation | Link to issue or docs |
| ---- | ---- | ---- | ---- |
| ✔️| Support for Secondary Address Ipv4 | ✔️ | |
| ✔️| Support for Secondary Address Ipv6 | ✔️ | |
| ✔️| Provision IP address using IPAM IPv4 | ✔️ | |
| ✔️| Provision IP address using IPAM IPv6(Amazon-Owned) Addresses  | ✔️ | |
| ✔️| Validation of Secondary Ipv4 address per RFC1918 range | ✔️ | |
| ✔️| Create VPC using BYOIP Ipv6 | ✔️ | |
|✔️ | Import existing VPC to this construct| ✔️ | |
| ✔️| Add Tagging capabilites to VPC and its components| ✔️| |

### Subnet
| Done | Item | Needed for Graduation | Link to issue or docs |
| ---- | ---- | ---- | ---- |
| ✔️| Support for custom Ipv4 range in subnet  | ✔️ | |
| ✔️| Support for custom Ipv6 range in subnet  | ✔️ | |
| ✔️| (Public Subnet) Construct Implementation of Nat and Internet Gateway| ✔️ | |
| ✔️| Validation for overlapping IP ranges |  ✔️| |
| ✔️| Validation for IP range lies within VPC range | ✔️ | |
| ✔️ | Support for custom Ipv6 range(BYOIP) in subnet  | ✔️ | |
| | Support for direct IPAM allocation in subnet | X | |

### Routing
| Done | Item | Needed for Graduation | Link to issue or docs |
| ---- | ---- | ---- | ---- |
| ✔️ | Route L2 construct | ✔️ | |
| ✔️ | RouteTable L2 construct | ✔️ | |
[...]

PR #31224, opened 2024-08-26
feat(vpcv2): implementation of add gateway method

shikha372:
### Issue # (if applicable)

Tracking  #[30762](https://github.com/aws/aws-cdk/issues/30762).

### Reason for this change

implementing below methods for vpcV2.
`routeTable.addroute(destination, target)`:   Adds a new route to the existing route table of the subnet.

`vpc.enableVpnGatewayV2()`:  added a new function for the customer to add VPNGateway to their VPC. In the options, user can specify list of subnets for VPNRoutePropogation. This is similar to previous implementation, only difference is with VPNGateway L2, it is now creating VPNGatewayV2 which implements IRouteTarget and hence can be used a destination to be set up in route tables.

`addInternetGateway` : adds internetGW to the VPC. 
**Default behaviour:**  add default route with destination set to ‘0.0.0.0’ and ‘::0’(in case of subnet with ipv6). Also a check in place to verify SubnetType is set to public as IGW is meant to be added to public subnets.

`addNatGateway`:  NatGateways are subnet specific and are usually associated with PRIVATE_WITH_EGRESS or PUBLIC subnet. Also, one can’t attach NGW(Public) to subnet if VPC doesn’t have an IGW attached to it. This is validated in method implementation to prevent runtime deployment error.

**No default behaviour** for the routes, it takes in the single subnet option and associates a NATGW with it. 

`vpc.addEgressOnlyInternetGateway()`: Egress Only internet GW are meant for outbound ipv6 traffic which can be custom or all ipv6(::/0).

**Default behaviour:** Associates a EIGW to the vpc and takes optional input for subnets to define a default route in associated route Table, if a destination is not provided, then it is defined as all outbound ipv6 in subnet’s route table.

**Additional changes:**
-> Modify Readme
-> Separate ipam related Tests

### Use Case

Allows user to define gateways in their vpc with a simple method and an optional default route setup on provided subnets.

Note: Breaking change since previously VPNGateway was released under route class,  we’ve modified it to VPNGatewayV2.
`vpc.enableVpnGateway` is marked as deprecated in vpcv2 base class.

Added unit tests and integration tests.

### Checklist

----


packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -569,6 +617,23 @@ export class RouteTable extends Resource implements IRouteTable, IDependable {
 
     this.routeTableId = this.resource.attrRouteTableId;
   }
+
+  /**
+   * Adds a new custom route to the route table.
+   * @param destination The IPv4 or IPv6 CIDR block used for the destination match.
+   * @param target The gateway or endpoint targeted by the route.
```
paulhcsun, line 624:
```suggestion
   * Adds a new custom route to the route table.
   *
   * @param destination The IPv4 or IPv6 CIDR block used for the destination match.
   * @param target The gateway or endpoint targeted by the route.
```

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -286,7 +291,30 @@ export class VPNGateway extends Resource implements IRouteTarget {
...
+
     this.vpcId = props.vpc.vpcId;
+    this._attachment = new CfnVPCGatewayAttachment(this, 'VPCVPNGW', {
+      vpcId: this.vpcId,
+      vpnGatewayId: this.resource.attrVpnGatewayId,
+    });
+
+    // Propagate routes on route tables associated with the right subnets
+    const vpnRoutePropagation = props.vpnRoutePropagation ?? [{}];
+    const routeTableIds = allRouteTableIds(flatten(vpnRoutePropagation.map(s => props.vpc.selectSubnets(s).subnets)));
```
paulhcsun, line 303: What is this line doing?

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -286,7 +291,30 @@ export class VPNGateway extends Resource implements IRouteTarget {
...
+    this._routePropagation = new CfnVPNGatewayRoutePropagation(this, 'RoutePropagation', {
+      routeTableIds,
+      vpnGatewayId: this.routerTargetId,
+    });
+    // The AWS::EC2::VPNGatewayRoutePropagation resource cannot use the VPN gateway
+    // until it has successfully attached to the VPC.
+    // See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpn-gatewayrouteprop.html
+    this._routePropagation.node.addDependency(this._attachment);
+
   }
```
paulhcsun, line 318:
```suggestion
    this._routePropagation.node.addDependency(this._attachment);
  }
```

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -331,6 +359,12 @@ export class NatGateway extends Resource implements IRouteTarget {
     this.connectivityType = props.connectivityType || NatConnectivityType.PUBLIC;
     this.maxDrainDuration = props.maxDrainDuration || Duration.seconds(350);
 
+    if (this.connectivityType == NatConnectivityType.PUBLIC) {
+      if (!props.vpc && !props.allocationId) {
+        throw new Error('Either provide vpc or allocationId');
```
paulhcsun, line 364: Is providing both acceptable?

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -491,39 +525,53 @@ export class Route extends Resource implements IRouteV2 {
...
   constructor(scope: Construct, id: string, props: RouteProps) {
     super(scope, id);
 
-    this.destination = props.destination;
     this.target = props.target;
     this.routeTable = props.routeTable;
+    this.destination = props.destination;
+    const isDestinationIpv4 = NetworkUtils.validIp(props.destination);
+    if (!isDestinationIpv4) {
+      //TODO Validate for IPv6 CIDR range
```
paulhcsun, line 546: is this intentionally left to be addressed later on?

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -569,6 +617,23 @@ export class RouteTable extends Resource implements IRouteTable, IDependable {
...
+
+  /**
+   * Adds a new custom route to the route table.
+   * @param destination The IPv4 or IPv6 CIDR block used for the destination match.
+   * @param target The gateway or endpoint targeted by the route.
+   */
+  public addRoute(id: string, destination: string, target: RouteTargetType) {
+    if (!target.gateway && !target.endpoint) {
+      throw new Error('Target is defined without a gateway or endpoint.');
+    }
```
paulhcsun, line 629: I feel like this check is not necessary if we tweak the implementation of `RouteTargetType` to enforce that exactly one of the two types are specified in the constructor. Something like this:

```ts
    if ((gateway && endpoint) || (!gateway && !endpoint)) {
      throw new Error("Exactly one of 'gateway' or 'endpoint' must be specified.");
    }

    if (gateway) {
      this.gateway = gateway;
    } else {
      this.endpoint = endpoint;
    }
```

packages/@aws-cdk/aws-ec2-alpha/lib/subnet-v2.ts
```diff
@@ -95,6 +96,12 @@ export interface ISubnetV2 extends ISubnet {
    */
   readonly ipv6CidrBlock?: string;
 
+  /**
+   * The type of subnet (public or private) that this subnet represents.
+   * @attribute SubnetType
```
paulhcsun, line 101:
```suggestion
   * The type of subnet (public or private) that this subnet represents.
   *
   * @attribute SubnetType
```

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -1,7 +1,73 @@
...
+import { EgressOnlyInternetGateway, InternetGateway, NatConnectivityType, NatGateway, NatGatewayOptions, Route, VPNGatewayV2 } from './route';
+import { ISubnetV2 } from './subnet-v2';
+
+/**
+ * Options to define EgressOnlyInternetGateway for VPC
+ */
+export interface EgressOnlyInternetGatewayOptions{
+  /**
+   * List of subnets where route to EGW will be added
+   * @default no route created
```
paulhcsun, line 14:
```suggestion
   * List of subnets where route to EGW will be added
   *
   * @default - no route created
```

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -1,7 +1,73 @@
...
+   */
+  readonly subnets?: SubnetSelection[];
+
+  /**
+   * Destination Ipv6 address for EGW route
+   * @default '::/0' all Ipv6 traffic
```
paulhcsun, line 20:
```suggestion
   * Destination Ipv6 address for EGW route
   *
   * @default - '::/0' all Ipv6 traffic
```
Can you clarify what this default value means? It feels a bit confusing when I read it.

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -1,7 +1,73 @@
...
+}
+
+/**
+ * Options to define InternetGateway for VPC
+ */
+export interface InternetGatewayOptions{
+
+  /**
+   * Destination Ipv6 address for EGW route
+   * @default '0.0.0.0' all Ipv4 traffic
```
paulhcsun, line 32:
```suggestion
   * Destination Ipv6 address for EGW route
   *
   * @default - '0.0.0.0' all Ipv4 traffic
```

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -1,7 +1,73 @@
...
+   */
+  readonly ipv4Destination?: string;
+
+  /**
+   * Destination Ipv6 address for EGW route
+   * @default '::/0' all Ipv6 traffic
```
paulhcsun, line 38:
```suggestion
   * Destination Ipv6 address for EGW route
   *
   * @default - '::/0' all Ipv6 traffic
```

paulhcsun: Please add a linebreak between the description and any @ default, @ see, @ attribute tags. Also separate the `@default` from the value with a hyphen `-` if you are describing it with a sentence and not the value itself.

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -1,7 +1,73 @@
...
+  readonly amazonSideAsn?: number;
+
+  /**
+     * The resource name of the VPN gateway.
+     * @default none
+     */
+  readonly vpnGatewayName?: string;
+
+  /**
+     * Provide an array of subnets where the route propagation should be added.
```
paulhcsun, line 66:
```suggestion
     * Subnets where the route propagation should be added.
```

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -196,6 +323,106 @@ export abstract class VpcV2Base extends Resource implements IVpcV2 {
...
+   * to the route table of given subnets
+   * @default in case of no input subnets, no route is created.
+   */
+  public addEgressOnlyInternetGateway(options?: EgressOnlyInternetGatewayOptions): void {
+    const egw = new EgressOnlyInternetGateway(this, 'EgressOnlyGW', {
+      vpc: this,
+    });
+
+    const useIpv6 = (this.secondaryCidrBlock.some((secondaryAddress) => secondaryAddress.amazonProvidedIpv6CidrBlock === true ||
+    secondaryAddress.ipv6IpamPoolId != undefined))? true : false;
```
paulhcsun, line 337: What is this doing?

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -196,6 +323,106 @@ export abstract class VpcV2Base extends Resource implements IVpcV2 {
...
+
+    if (!useIpv6) {
+      throw new Error('Egress only IGW can only be added to Ipv6 enabled VPC');
+    };
+
+    if (options?.subnets) {
+      const subnets = flatten(options.subnets.map(s => this.selectSubnets(s).subnets));
+      subnets.forEach((subnet) => {
+        this.createEgressRoute(subnet, egw, options.destination);
```
paulhcsun, line 346: What is this doing?

paulhcsun (changes requested):

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -174,7 +153,18 @@ export interface NatGatewayProps {
...
+}
 
+/**
+ * Properties to define a NAT gateway.
+ */
+export interface NatGatewayProps extends NatGatewayOptions {
+  /**
+   * The ID of the VPC in which the NAT gateway is located.
+   * Required in case of public connectivity if allocation id is not defined
+   * @default none
```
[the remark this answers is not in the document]
shikha372: I wonder why is this a requirement if its not enforced with the linter, also I see it that this is not consistent in whole library, is this something suggested by the core team.
eg. checking this file https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-apigatewayv2/lib/http/authorizer.ts

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -286,7 +291,30 @@ export class VPNGateway extends Resource implements IRouteTarget {
     this.node.defaultChild = this.resource;
 
     this.routerTargetId = this.resource.attrVpnGatewayId;
+
     this.vpcId = props.vpc.vpcId;
+    this._attachment = new CfnVPCGatewayAttachment(this, 'VPCVPNGW', {
```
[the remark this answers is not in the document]
shikha372: since its a L1, it was suggested in earlier review to store the reference the variable(i suppose which makes it a part of construct tree)

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -174,7 +153,18 @@ export interface NatGatewayProps {
...
   readonly natGatewayName?: string;
+}
 
+/**
+ * Properties to define a NAT gateway.
+ */
+export interface NatGatewayProps extends NatGatewayOptions {
+  /**
+   * The ID of the VPC in which the NAT gateway is located.
+   * Required in case of public connectivity if allocation id is not defined
```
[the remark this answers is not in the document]
shikha372: NATGateway can be `Public` or `Private`, so in case of public connectivity we need allocation id(refers to Elastic IP).

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
shikha372: default means setting up outbound to all IPv6 address for which the range is '::/0'

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
shikha372: Since we can have multiple secondary address attached to VPC, this is to check if any of the secondary address is a IPv6 meaning created using `IPAMIpv6` or is a `AmazonProvidedIpv6`

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
shikha372: this is the default route that we are creating under provided subnet's routeTable. It will add an entry to subnet's routeTable either with the specified destination address or with default(::/0)

packages/@aws-cdk/aws-ec2-alpha/README.md
```diff
@@ -195,9 +213,37 @@ const dynamoEndpoint = new ec2.GatewayVpcEndpoint(this, 'DynamoEndpoint', {
...
+  vpc: myVpc,
+});
+const subnet = new SubnetV2(this, 'Subnet', {  
+  vpc: myVpc,
+  availabilityZone: 'eu-west-2a',
+  ipv4CidrBlock: new IpCidr('10.0.0.0/24'),
+  subnetType: ec2.SubnetType.PRIVATE });
+
+myVpc.addEgressOnlyInternetGateway({
+  subnets: [{SubnetType.PUBLIC}],
```
GavinZZ, line 245: Haven't looked into implementation but this already looks weird to me.

packages/@aws-cdk/aws-ec2-alpha/awslint.json
```diff
@@ -2,6 +2,7 @@
     "exclude": [
       "from-method:@aws-cdk/aws-ec2-alpha.VpcV2",
       "attribute-tag:@aws-cdk/aws-ec2-alpha.RouteTable.routeTableId",
-      "from-method:@aws-cdk/aws-ec2-alpha.SubnetV2"
+      "from-method:@aws-cdk/aws-ec2-alpha.SubnetV2",
+      "from-method:@aws-cdk/aws-ec2-alpha.Route"
```
GavinZZ, line 6: For my knowledge, what does these two lines do?

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -23,7 +25,7 @@ export enum NatConnectivityType {
  * Interface to define a routing target, such as an
  * egress-only internet gateway or VPC endpoint.
  */
-export interface IRouteTarget {
+export interface IRouteTarget extends IDependable{
```
GavinZZ, line 28: needs a space before `{`

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -71,45 +73,22 @@ export interface InternetGatewayProps {
 /**
  * Properties to define a VPN gateway.
  */
-export interface VPNGatewayProps {
```
GavinZZ, line 74: This is worrying. This is an exported interface and users might use it for some purposes.

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -71,45 +73,22 @@ export interface InternetGatewayProps {
...
 /**
- * Properties to define a NAT gateway.
+ * Options to define a NAT gateway.
  */
-export interface NatGatewayProps {
+export interface NatGatewayOptions {
   /**
    * The subnet in which the NAT gateway is located.
    */
-  readonly subnet: ISubnet;
```
GavinZZ, line 106: Isn't this a breaking change for users?

GavinZZ (changes requested):
I'm a bit worried by the number of breaking changes. I understand it is an alpha module and breaking changes are allowed. However, a number of changes are removing an old construct and introduced V2, i.e. `VPNGateway` is removed and introduced `VPNGatewayV2`. It's breaking changes anyway, should we just use the same name?

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
shikha372: If you mean with that they might be extending it somewhere for implementation, we don't bind the contract of keeping the interface aligned even in the main lib(checked with core team for one of the route interfaces), for the existing VPNGW props this remains same overall, only breaking change that we specify is the change from VPNGateway to VPNGatewayV2, explained below why we need to do that.

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
shikha372: if mean changing the ISubnet to ISubnetv2, its not breaking change since ISubnetv2 extends ISubnet

packages/@aws-cdk/aws-ec2-alpha/awslint.json
shikha372: this is enforced from `awslint` to have the import methods in classes extending `Resource` , since we haven't introduced it yet these are part of exceptions

packages/@aws-cdk/aws-ec2-alpha/README.md
shikha372: This is as per the current implementations we have in place for addendpoint and addgateways and provides a better way for the users to select multiple subnets at once without forming an array list first and then pass it. This seems to be the best option for specifying a subnet as an input.

packages/@aws-cdk/aws-ec2-alpha/README.md
```diff
@@ -195,9 +213,37 @@ const dynamoEndpoint = new ec2.GatewayVpcEndpoint(this, 'DynamoEndpoint', {
...
 ```
+
+## Adding Egress-Only Internet Gateway to VPC
+
+An egress-only internet gateway is a horizontally scaled, redundant, and highly available VPC component that allows outbound communication over IPv6 from instances in your VPC to the internet, and prevents the internet from initiating an IPv6 connection with your instances. For more information see@ https://docs.aws.amazon.com/vpc/latest/userguide/egress-only-internet-gateway.html
+
+VPCv2 supports adding an egress only internet gateway to VPC with the help of `addEgressOnlyInternetGateway` method as well.
+
+By Default, it sets up a route to all outbound IPv6 Address ranges unless specified to a specific destination by the user. It can only be set up for IPv6 enabled VPCs.
+`Subnets` takes in value of `SubnetFilter` which can be based on a SubnetType in VPCV2. A new route will be added to route tables of all subnets filtered out with this property.
```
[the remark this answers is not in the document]
shikha372: Selecting a type of subnet is a `SubnetFilter` :) 
`subnets:` [{SubnetType.PUBLIC}]

shikha372:
> Only half way through.
> 
> I'm a bit worried by the number of breaking changes. I understand it is an alpha module and breaking changes are allowed. However, a number of changes are removing an old construct and introduced V2, i.e. `VPNGateway` is removed and introduced `VPNGatewayV2`. It's breaking changes anyway, should we just use the same name?

So one is compatible with `Route` construct and other lives in the main lib `VPNGateway` is not implementing `IRouteTarget` so that can not be set as an target while defining a route. So, the V2 is aligned with features (i.e. `Route` and `RouteTable`) in this new module, it is more clear that this is latest. Also, if there is already a construct with same name in main lib, not sure if its best practice to introduce another with same name as it can create issue with graduation.

packages/@aws-cdk/aws-ec2-alpha/README.md
```diff
@@ -129,47 +129,65 @@ const subnet = new vpc_v2.SubnetV2(this, 'Subnet', {
...
 });
-new vpc_v2.Route(this, 'IgwRoute', {
+new Route(this, 'IgwRoute', {
   routeTable,
   destination: '0.0.0.0/0',
   target:  { gateway: igw },
 });
 ```
 
+Alternatively, `Route`s can be created via a method in the `RouteTable` class. An example using the `EgressOnlyInternetGateway` construct can be seen below:
```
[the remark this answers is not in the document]
shikha372: 1. corrected
2. it does relates with creating the route in line , you cannot create a route without a target so we need a target first(in this case), it is EgressOnlyInternetGateway
routeTable.addRoute('::/0', { gateway: eigw });
To make it more clear mentioned the method name

packages/@aws-cdk/aws-ec2-alpha/README.md
```diff
@@ -195,9 +213,37 @@ const dynamoEndpoint = new ec2.GatewayVpcEndpoint(this, 'DynamoEndpoint', {
...
   destination: '0.0.0.0/0',
   target: { endpoint: dynamoEndpoint },
 });
 ```
+
+## Adding Egress-Only Internet Gateway to VPC
+
+An egress-only internet gateway is a horizontally scaled, redundant, and highly available VPC component that allows outbound communication over IPv6 from instances in your VPC to the internet, and prevents the internet from initiating an IPv6 connection with your instances. For more information see@ https://docs.aws.amazon.com/vpc/latest/userguide/egress-only-internet-gateway.html
+
+VPCv2 supports adding an egress only internet gateway to VPC with the help of `addEgressOnlyInternetGateway` method as well.
```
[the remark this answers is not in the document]
shikha372: checked to confirm i think its aligned everywhere else in the Readme, just not in the two places in this section corrected it.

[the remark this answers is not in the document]
shikha372: Thanks @GavinZZ , I checked and its only in this block that its `VPCv2` but I think Readme had pending updates for some other add methods, have updated it.

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
shikha372: good point, validation can be in `RouteTarget`, moved it in latest rev.

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -331,6 +358,12 @@ export class NatGateway extends Resource implements IRouteTarget {
     this.connectivityType = props.connectivityType || NatConnectivityType.PUBLIC;
     this.maxDrainDuration = props.maxDrainDuration || Duration.seconds(350);
 
+    if (this.connectivityType == NatConnectivityType.PUBLIC) {
```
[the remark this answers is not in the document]
GavinZZ: nvm, I found that `connectivityType` is a string type.

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -196,6 +327,106 @@ export abstract class VpcV2Base extends Resource implements IVpcV2 {
...
+   * to the route table of given subnets
+   * @default in case of no input subnets, no route is created.
+   */
+  public addEgressOnlyInternetGateway(options?: EgressOnlyInternetGatewayOptions): void {
+    const egw = new EgressOnlyInternetGateway(this, 'EgressOnlyGW', {
+      vpc: this,
+    });
+
+    const useIpv6 = (this.secondaryCidrBlock.some((secondaryAddress) => secondaryAddress.amazonProvidedIpv6CidrBlock === true ||
+    secondaryAddress.ipv6IpamPoolId != undefined))? true : false;
```
GavinZZ, line 341: `some` already returns a boolean.
```suggestion
    const useIpv6 = this.secondaryCidrBlock.some((secondaryAddress) => secondaryAddress.amazonProvidedIpv6CidrBlock === true ||
    secondaryAddress.ipv6IpamPoolId != undefined);
```

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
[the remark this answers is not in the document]
GavinZZ: should we make it a Enum typed value?

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -408,7 +441,7 @@ export class RouteTargetType {
 /**
  * Interface to define a route.
  */
-export interface IRouteV2 {
+export interface IRoute extends IConstruct, IResource{
```
GavinZZ, line 444: I see some discrepancy where some changes are from v1 to v2 and some changes are from v2 to v1. Can we make it consistent?

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -1,7 +1,77 @@
...
+     * The resource name of the VPN gateway.
+     * @default none
+     */
+  readonly vpnGatewayName?: string;
+
+  /**
+     * Subnets where the route propagation should be added.
+     * @default noPropagation
+     */
+  readonly vpnRoutePropagation?: SubnetSelection[];
```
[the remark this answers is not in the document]
shikha372: yeah, got to know it from @paulhcsun , so working to reformat this for entire repo in next revision,

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -286,7 +291,29 @@ export class VPNGateway extends Resource implements IRouteTarget {
...
+
     this.vpcId = props.vpc.vpcId;
+    this._attachment = new CfnVPCGatewayAttachment(this, 'VPCVPNGW', {
+      vpcId: this.vpcId,
+      vpnGatewayId: this.resource.attrVpnGatewayId,
+    });
+
+    // Propagate routes on route tables associated with the right subnets
+    const vpnRoutePropagation = props.vpnRoutePropagation ?? [{}];
+    const routeTableIds = allRouteTableIds(flatten(vpnRoutePropagation.map(s => props.vpc.selectSubnets(s).subnets)));
```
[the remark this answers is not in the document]
shikha372: `allRouteTableIDs` signifies that there are multiple because of possible multiple subnets, i feel its more clear than getRoutTableIds

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -286,7 +291,29 @@ export class VPNGateway extends Resource implements IRouteTarget {
...
+
+    if (routeTableIds.length === 0) {
+      Annotations.of(this).addError(`enableVpnGateway: no subnets matching selection: '${JSON.stringify(vpnRoutePropagation)}'. Select other subnets to add routes to.`);
```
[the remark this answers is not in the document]
shikha372: since its optional, i'll modify it be a warning.

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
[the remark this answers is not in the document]
shikha372: it is an enum typed value in props, so the user pass an enum for this value not string.

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
shikha372: i  think its supposed be `IRoute` only, we implemented an idea in between of extending this for introducing one field but went back to the [existing one](https://github.com/aws/aws-cdk/blob/vpcv2-rt-addGateway/packages/%40aws-cdk/aws-ec2-alpha/lib/route.ts#L444C33-L444C54) , this is same as current but i see a nit here `IResource {` that needs to be fixed.
V2 are mainly for the constructs that already exist.

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -462,7 +495,7 @@ export interface RouteProps {
  * Creates a new route with added functionality.
  * @resource AWS::EC2::Route
  */
-export class Route extends Resource implements IRouteV2 {
+export class Route extends Resource implements IRoute, IDependable {
```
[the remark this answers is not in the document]
shikha372: This helps in specifying dependencies within construct. 
So this was needed in order to create a dependency on the gateways and endpoints creation before creation of route. It should start only after your target is created else the deployment will fails. This is handled as below in code.

```
    if (this.target.gateway) {
      this.node.addDependency(this.target.gateway);
    }
    if (this.target.endpoint) {
      this.node.addDependency(this.target.endpoint);
    }
```

packages/@aws-cdk/aws-ec2-alpha/lib/subnet-v2.ts
```diff
@@ -250,6 +257,14 @@ export class SubnetV2 extends Resource implements ISubnetV2 {
       subnet: this,
     });
   }
+
+  /**
+   * Return the Route Table associated with this subnet
+   */
+  public get routeTable(): IRouteTable {
```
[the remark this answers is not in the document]
shikha372: so this was tied to implementation of `[associateRouteTable](https://github.com/aws/aws-cdk/pull/31224/commits/f904a97031a301787b16db7f14619b4e67868a7e#diff-fe8bcfc72ac04e9cea500b3fd101b842e2bb77a648a3e6d0c7fdd2d3cdb16ba0R258)` method in SubnetV2, which we also reverted but we can introduce it in the next release with imports so prefer to keep it as is now currently unless you see any blocking concerns.

packages/@aws-cdk/aws-ec2-alpha/lib/subnet-v2.ts
```diff
@@ -145,19 +153,16 @@ export class SubnetV2 extends Resource implements ISubnetV2 {
...
-   * The route table for this subnet
-   */
-  public readonly routeTable: IRouteTable;
-
   /**
    * The type of subnet (public or private) that this subnet represents.
    * @attribute SubnetType
    */
-  public readonly subnetType: SubnetType;
+  public readonly subnetType?: SubnetType;
```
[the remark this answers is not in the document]
shikha372: This is not optional under props, so user still needs to defined it but it is marked as optional under the new interface `SubnetV2` because if not it will cause issue with the existing property in [`SubnetSelection` for subnets ](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts#L363) which we are leveraging in some of the places [like here.](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/aws-ec2-alpha/lib/subnet-v2.ts#L351)

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
```diff
@@ -331,9 +369,15 @@ export class NatGateway extends Resource implements IRouteTarget {
...
+    if (this.connectivityType === NatConnectivityType.PUBLIC) {
+      if (!props.vpc && !props.allocationId) {
+        throw new Error('Either provide vpc or allocationId');
+      }
+    }
+
     // If user does not provide EIP, generate one for them
     var aId: string | undefined;
-    if (this.connectivityType == 'public') {
+    if (this.connectivityType == NatConnectivityType.PUBLIC) {
```
GavinZZ, line 380: `===`

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -1,7 +1,80 @@
...
 import { allRouteTableIds, flatten, subnetGroupNameFromConstructId } from './util';
-import { IDependable, Dependable, IConstruct } from 'constructs';
+import { IDependable, Dependable, IConstruct, DependencyGroup } from 'constructs';
+import { EgressOnlyInternetGateway, InternetGateway, NatConnectivityType, NatGateway, NatGatewayOptions, Route, VPNGatewayV2 } from './route';
+import { ISubnetV2 } from './subnet-v2';
+
+/**
+ * Options to define EgressOnlyInternetGateway for VPC
+ */
+export interface EgressOnlyInternetGatewayOptions{
```
GavinZZ, line 11:
```suggestion
export interface EgressOnlyInternetGatewayOptions {
```

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -196,6 +333,108 @@ export abstract class VpcV2Base extends Resource implements IVpcV2 {
...
+    }
+
+    const igw = new InternetGateway(this, 'InternetGateway', {
+      vpc: this,
+    });
+
+    this._internetConnectivityEstablished.add(igw);
+    this._internetGatewayId = igw.routerTargetId;
+
+    //If there are no public subnets defined, no default route will be added
```
GavinZZ, line 392:
```suggestion
    // If there are no public subnets defined, no default route will be added
```

packages/@aws-cdk/aws-ec2-alpha/lib/vpc-v2-base.ts
```diff
@@ -196,6 +333,108 @@ export abstract class VpcV2Base extends Resource implements IVpcV2 {
...
+   * Adds default route for the internet gateway
+   * @internal
+   */
+  private addDefaultInternetRoute(subnet: ISubnetV2, igw: InternetGateway, options?: InternetGatewayOptions): void {
+
+    if (subnet.subnetType !== SubnetType.PUBLIC) {
+      throw new Error('No public subnets defined to add route for internet gateway');
+    }
+
+    //Add deffault route to IGW for IPv6
```
GavinZZ, line 408:
```suggestion
    // Add default route to IGW for IPv6
```

packages/@aws-cdk/aws-ec2-alpha/test/integ.ipam.ts
```diff
@@ -0,0 +1,84 @@
+// /*
+//  * Our integration tests act as snapshot tests to make sure the rendered template is stable.
+//  * If any changes to the result are required,
+//  * you need to perform an actual CloudFormation deployment of this application,
+//  * and, if it is successful, a new snapshot will be written out.
+//  *
+//  * For more information on CDK integ tests,
+//  * see the main CONTRIBUTING.md file.
+//  */
```
GavinZZ, line 9: duplicate `//`

GavinZZ (dismissed):
Thanks Shikha for this great work. The code generally looks good to me. I have two high-level thoughts:
1. I know I've mentioned this, but seeing the usage and it's in `alpha` modules, I'm not convinced why we need to remove the original constructs, i.e. `VPNGateway, Subnet` and create a `VPNGatewayV2, SubnetV2`. Either way, it's breaking changes for customers, wouldn't keep using the name `VPNGateway and Subnet` but with new implementation be better if in the future we decide to make it stable and it won't be starting from `V2` by default.
2. Subnets have to be attached with a VPC resource. Can we introduces methods in VPC construct like `vpc.addSubnet(...)` just like how you added `vpc.addNatGateway, vpc.addInternetGateway`. I peronsally think it makes sense to add subnet this way and prefer creating subnets this way instead of `new SubnetV2`. Want to see your thoughts.

packages/@aws-cdk/aws-ec2-alpha/lib/route.ts
shikha372: i see it updated in the last revision, not sure if you;re still seeing the same https://github.com/aws/aws-cdk/pull/31224/files/5c7548c4954c8b3433cef4b45e7f0955825ddf1b#diff-65117dd44ab7b3f0b0f8d0421973a71b104cd76b4dc5147237da60616c718b2fR372

paulhcsun (approved):
Looks good to me, deferring final approval to @GavinZZ. Nice work @shikha372!

shikha372:
@mergifyio update

PR #31225, opened 2024-08-26
add unit test for oac permission levels

gracelu0:
- add unit test to check the correct permissions are being added if user specifies READ/WRITE/DELETE for `originAccessLevels`

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-bucket-origin.test.ts
+++ b/packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-bucket-origin.test.ts
@@ -466,6 +466,72 @@ describe('S3BucketOrigin', () => {
         });
       });
     });
+    describe('when specifying READ, WRITE, and DELETE origin access levels', () => {
+      it('should add the correct permissions to bucket policy', () => {
+        const stack = new Stack();
+        const bucket = new s3.Bucket(stack, 'MyBucket');
+        const origin = origins.S3BucketOrigin.withOriginAccessControl(bucket, {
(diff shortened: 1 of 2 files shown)


packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-bucket-origin.test.ts
```diff
@@ -466,6 +466,72 @@ describe('S3BucketOrigin', () => {
         });
       });
     });
+    describe('when specifying READ, WRITE, and DELETE origin access levels', () => {
+      it('should add the correct permissions to bucket policy', () => {
```
samson-keung, line 470: I expected this case to have been covered by the existing unit test: https://github.com/samson-keung/aws-cdk/blob/4c9d7196370351c9f4f7a61b82e2981520b0f516/packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-bucket-origin.test.ts#L20

Or do we need this one to focus on the security aspect?

gracelu0: Yes I was thinking this is scoped to checking the permissions and covers `DELETE` permissions as well.

samson-keung: I would suggest modifying the existing unit test to cover the DELETE case so we do not have overlapping tests. But with that said, I don't think it is a blocked to have overlapping tests either.

samson-keung (approved):

PR #31227, opened 2024-08-27
Use wildcard in KMS key policy instead of referencing Distribution to avoid circular dependency

samson-keung:
Making the KMS key policy to use wildcard instead on the distribution id part instead of referencing the Distribution resource to get the exact distribution ID. This breaks the circular dependency in the template.

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
+++ b/packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
@@ -84,12 +84,6 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
         }
 
         if (bucket.encryptionKey) {
-          Annotations.of(scope).addInfo(
-            `Granting OAC permissions to access KMS key for S3 bucket origin ${bucket.node.id} may cause a circular dependency error when this stack deploys.\n` +
-            'The key policy references the distribution\'s id, the distribution references the bucket, and the bucket references the key.\n'+
-            'See the "Using OAC for a SSE-KMS encrypted S3 origin" section in the module README for more details.\n',
-          );
-
           const keyPolicyActions = this.getKeyPolicyActions(props?.originAccessLevels ?? [cloudfront.AccessLevel.READ]);
           const keyPolicyResult = this.grantDistributionAccessToKey(keyPolicyActions, bucket.encryptionKey);
           // Failed to update key policy, assume using imported key
@@ -161,9 +155,10 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
             },
           },
         );
-        Annotations.of(key.node.scope!).addWarningV2('@aws-cdk/aws-cloudfront-origins:wildcardKeyPolicy',
-          'Using wildcard to match all Distribution IDs in Key policy condition.\n' +
-          'To further scope down the policy, see the "Using OAC for a SSE-KMS encrypted S3 origin" section in the module README.');
+        Annotations.of(key.node.scope!).addWarningV2('@aws-cdk/aws-cloudfront-origins:wildcardKeyPolicyForOac',
+          'To avoid circular dependency between the KMS key, Bucket, and Distribution,' +
+          'a wildcard is used to match all Distribution IDs in Key policy condition.\n' +
+          'To further scope down the policy for best security practices, see the "Using OAC for a SSE-KMS encrypted S3 origin" section in the module README.');
         const result = key.addToResourcePolicy(oacKeyPolicyStatement);
         return result;
       }
--- a/packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-bucket-origin.test.ts
+++ b/packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-bucket-origin.test.ts
@@ -379,8 +379,9 @@ describe('S3BucketOrigin', () => {
           },
         });
         Annotations.fromStack(stack).hasWarning('/Default',
-          'Using wildcard to match all Distribution IDs in Key policy condition.\n' +
-          'To further scope down the policy, see the "Using OAC for a SSE-KMS encrypted S3 origin" section in the module README. [ack: @aws-cdk/aws-cloudfront-origins:wildcardKeyPolicy]');
+          'To avoid circular dependency between the KMS key, Bucket, and Distribution,' +
+          'a wildcard is used to match all Distribution IDs in Key policy condition.\n' +
+          'To further scope down the policy for best security practices, see the "Using OAC for a SSE-KMS encrypted S3 origin" section in the module README. [ack: @aws-cdk/aws-cloudfront-origins:wildcardKeyPolicyForOac]');
       });
     });
 


packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -84,15 +110,8 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
         }
 
         if (bucket.encryptionKey) {
-          let bucketName = bucket.bucketName;
-          if (Token.isUnresolved(bucket.bucketName)) {
```
samson-keung, line 88: Removed this because turns out `bucket.bucketName` is always unresolved as it returns CFN reference to the bucket (i.e. `!Ref bucket`). It is set [here](https://github.com/samson-keung/aws-cdk/blob/4e7d4804093d1a2a54f03830a3770d56c4729bc8/packages/aws-cdk-lib/aws-s3/lib/bucket.ts#L1962).

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -61,10 +70,27 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
...
+          if (!Token.isUnresolved(bucket.bucketName)) {
+            // this is the case when bucket is from Bucket.fromBucketName
+            bucketName = bucket.bucketName;
+          } else if (!Token.isUnresolved((bucket.node.defaultChild as CfnBucket)?.bucketName)) {
+            // this is the case when bucket is a L2 bucket with bucketName set
+            bucketName = (bucket.node.defaultChild as CfnBucket)?.bucketName;
+          }
+
+          if (!bucketName) {
+            throw new Error(`Cannot assemble static DomainName as bucket ${bucket.node.id} has no bucketName set.`);
```
xazhao, line 91: I see it's getting the bucket name using 
```
${(bucket.node.defaultChild as CfnBucket).bucketName}
```
at Line 132. Do we still need to check `Token.isUnresolved(bucket.bucketName)`?

samson-keung: Good catch! We do. Thank you.

I will cache the `bucketName` in the class and reuse it at Line 132.

I will change this PR to `Draft` as well since this approach is pending security discussion.

packages/aws-cdk-lib/aws-cloudfront-origins/test/s3-bucket-origin.test.ts
```diff
@@ -165,6 +166,221 @@ describe('S3BucketOrigin', () => {
       });
     });
 
+    describe('when using bucket with KMS Customer Managed key', () => {
+      it('should match expected template resource', () => {
```
samson-keung, line 170: One more unit test is needed to test escape hatch. PR to come

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -154,20 +147,23 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
...
             conditions: {
-              StringEquals: {
-                'AWS:SourceArn': `arn:${Aws.PARTITION}:cloudfront::${Aws.ACCOUNT_ID}:distribution/${distributionId}`,
+              ArnLike: {
+                'AWS:SourceArn': `arn:${Aws.PARTITION}:cloudfront::${Aws.ACCOUNT_ID}:distribution/*`,
               },
             },
           },
         );
+        Annotations.of(key.node.scope!).addWarningV2('@aws-cdk/aws-cloudfront-origins:wildcardKeyPolicy',
```
gracelu0, line 164:
```suggestion
        Annotations.of(key.node.scope!).addWarningV2('@aws-cdk/aws-cloudfront-origins:wildcardKeyPolicyForOac',
```

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -84,21 +84,14 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
...
-              scope,
-              resolver: new DefaultTokenResolver(new StringConcat()),
-            }));
-          }
           Annotations.of(scope).addInfo(
-            `Granting OAC permissions to access KMS key for S3 bucket origin ${bucketName} may cause a circular dependency error when this stack deploys.\n` +
+            `Granting OAC permissions to access KMS key for S3 bucket origin ${bucket.node.id} may cause a circular dependency error when this stack deploys.\n` +
             'The key policy references the distribution\'s id, the distribution references the bucket, and the bucket references the key.\n'+
             'See the "Using OAC for a SSE-KMS encrypted S3 origin" section in the module README for more details.\n',
           );
```
gracelu0, line 91: Do we still need this info message or can we combine it with the warning about wildcard? Since with the wildcard policy they most likely will not run into the circular dependency error.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -154,20 +147,23 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
...
-              StringEquals: {
-                'AWS:SourceArn': `arn:${Aws.PARTITION}:cloudfront::${Aws.ACCOUNT_ID}:distribution/${distributionId}`,
+              ArnLike: {
+                'AWS:SourceArn': `arn:${Aws.PARTITION}:cloudfront::${Aws.ACCOUNT_ID}:distribution/*`,
               },
             },
           },
         );
+        Annotations.of(key.node.scope!).addWarningV2('@aws-cdk/aws-cloudfront-origins:wildcardKeyPolicy',
+          'Using wildcard to match all Distribution IDs in Key policy condition.\n' +
```
gracelu0, line 165:
```suggestion
          `A wildcard is used to grant OAC permissions for the KMS key access to the S3 bucket origin ${bucket.node.id} to resolve a circular dependency between the key policy, distribution, and bucket.\n` +
          'To follow best security practices and avoid unintended access, do not rely on this default.\n' +
          'Instead, you scope down the policy condition to the specific distribution. Refer to the "Using OAC for a SSE-KMS encrypted S3 origin" section in the module README for guidance.'
```

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: Suggested a warning message below which combines the two.

gracelu0 (changes requested):

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
samson-keung: Agree this message should not be needed anymore. I have removed it in the latest commit. Also updated the other warning to capture the info in your suggest warning message.

gracelu0 (approved):
LGTM, thank you!

PR #31232, opened 2024-08-27
feat(cloudwatch): anomaly detection alarms

### LINKED ISSUE #10540: [aws-cloudwatch] Missing ThresholdMetricId

scorbiere:
### Issue #10540

Closes #10540.

### Reason for this change

This change adds proper support for creating CloudWatch anomaly detection alarms, fixing a long-standing issue where alarms using anomaly detection with range-based comparison operators would fail at deployment time with the error:

ComparisonOperators for ranges require ThresholdMetricId to be set (Service: AmazonCloudWatch; Status Code: 400; Error 
Code: ValidationError)

### Description of changes

1. **New `AnomalyDetectionAlarm` class**: a dedicated class for anomaly detection alarms that handles all the complexity of setting up the anomaly detection band and configuring the alarm correctly.

2. **Enhanced helper methods**: Added the `anomalyDetectionFor` helper method on `Metric` class to handle the creation of the MathExpression wrapper using the function `ANOMALY_DETECTION_BAND`.

3. **Improved validation**: Added proper validation to ensure users don't mix standard alarm operators with anomaly detection metrics or vice versa.

- Added comprehensive unit tests for the new `AnomalyDetectionAlarm` class
- Executed the integration test (`packages/@aws-cdk-testing/framework-integ/test/aws-cloudwatch/test/integ.anomaly-detection-alarm.ts`) on a test account

### Checklist

----


packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -387,6 +387,25 @@ only supports filtering by `unit` for Alarms, not in Dashboard graphs.
 Please see the following GitHub issue for a discussion on real unit
 calculations in CDK: https://github.com/aws/aws-cdk/issues/5595
 
+## Anomaly Detection Alarms
+
+To create an alarm based on anomaly detection, you can use the `createAnomalyDetectionAlarm` method on a `Metric` or `MathExpression` object:
```
xazhao, line 392: I'm just curious why do you decide to create a function here instead of a construct `new cloudwatch.AnomalyDetectionAlarm(...)`

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -137,6 +144,18 @@ export class Alarm extends AlarmBase {
...
+  /**
+   * Determine whether this operator is an anomaly detection operator.
+   *
+   * @param operator the comparison operator for the alarm.
+   * @returns true if the operator is an anomaly detection operator, false otherwise.
+   */
+  public static isAnomalyDetectionOperator(operator: ComparisonOperator): boolean {
+    return operator === ComparisonOperator.LESS_THAN_LOWER_OR_GREATER_THAN_UPPER_THRESHOLD
+      || operator === ComparisonOperator.GREATER_THAN_UPPER_THRESHOLD
+      || operator === ComparisonOperator.LESS_THAN_LOWER_THRESHOLD;
```
xazhao, line 156: How about 
```
[ComparisonOperator.LESS_THAN_LOWER_OR_GREATER_THAN_UPPER_THRESHOLD, 
ComparisonOperator.GREATER_THAN_UPPER_THRESHOLD, 
ComparisonOperator.LESS_THAN_LOWER_THRESHOLD].includes(operator)
```

packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts
```diff
@@ -723,6 +748,31 @@ export class MathExpression implements IMetric {
...
+    });
+
+    return new Alarm(scope, id, {
+      ...props,
+      metric: anomalyDetectionExpression,
+      threshold: 0, // This value will be ignored for anomaly detection alarms
+      thresholdMetricId: 'expr_1',
+    });
+  }
+
```
xazhao, line 775: Looks like this function is identical to the one in Metric class. Is there a way to re-use the same function?

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -206,12 +227,22 @@ export class Alarm extends AlarmBase {
...
     });
 
+    if (isAnomalyDetection) {
+      // Ensure returnData is set to true for both metrics in anomaly detection
+      const cfnMetrics = alarm.metrics as any[];
+      if (cfnMetrics && cfnMetrics.length >= 2) {
+        cfnMetrics[0].returnData = true;
+        cfnMetrics[1].returnData = true;
+      }
+    }
```
xazhao, line 237: Where is `cfnMetrics` used?

packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts
scorbiere: Correct I will factorize that part :)

packages/aws-cdk-lib/aws-cloudwatch/README.md
scorbiere: Main motivations:
- Less refactoring.
- Less risk of breaking the existing `Alarm` construct.
- The handling of the anomaly detection case require limited amount of changes in the existing `Alarm` construct.

Tradeoff:
- It is still possible to not use the new helper methods and give a wrong metric to the existing `Alarm` construct.

packages/aws-cdk-lib/rosetta/aws_cloudwatch/default.ts-fixture
```diff
@@ -4,6 +4,7 @@ import { Stack, Duration } from 'aws-cdk-lib';
 import * as cloudwatch from 'aws-cdk-lib/aws-cloudwatch';
 import * as route53 from 'aws-cdk-lib/aws-route53';
 import * as sns from 'aws-cdk-lib/aws-sns';
+import * as sqs from 'aws-cdk-lib/aws-sqs';
```
xazhao, line 7: I don't know about `rosetta`. What does this change do?

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
scorbiere: Only in this block. To make sure the `returnData` property is set to true. TBH, the documentation is not super clear about which one should have the property set to true.

packages/aws-cdk-lib/rosetta/aws_cloudwatch/default.ts-fixture
scorbiere: It is used to translate the example in the markdown files from TypeScript to Python, Java, ... 

See: https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#rosetta

packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts
```diff
@@ -259,6 +259,33 @@ export class Metric implements IMetric {
...
+ * @param props Alarm creation properties for anomaly detection
+ * @param metric The metric to create the alarm for
+ * @returns The newly created Alarm
+ */
+  public static createAnomalyDetectionAlarmFromMetric(scope: Construct, id: string, props: CreateAnomalyDetectionAlarmProps, metric: IMetric): Alarm {
+    if (!Alarm.isAnomalyDetectionOperator(props.comparisonOperator)) {
+      throw new Error(`Invalid comparison operator for anomaly detection alarm: ${props.comparisonOperator}`);
+    }
+    const anomalyDetectionExpression = new MathExpression({
+      expression: `ANOMALY_DETECTION_BAND(m0, ${props.bounds ?? 1})`,
```
xazhao, line 276: since `bounds` is a required property, do we need null check here?

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -190,7 +213,8 @@ export class Alarm extends AlarmBase {
 
       // Evaluation
       comparisonOperator,
-      threshold: props.threshold,
+      threshold: isAnomalyDetection ? undefined : props.threshold,
```
xazhao, line 216: What will happen if it's anomaly detection and threshold value is passed in? Does it ignore the value or throw an error?

xazhao:
General question: There's a `CfnAnomalyDetector` L1 construct which is not used here. Could you help me understand the diff between `CfnAnomalyDetector` and what we are achieving in this PR?

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
scorbiere: If the comparison operator is for an anomaly detection alarm, there will be an error when CloudFormation will try to execute the template. Because providing a threshold for anomaly detection is not valid.

packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts
scorbiere: Right, this is useless now that the property is required.

xazhao: Or should we have a default value for it?

packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts
```diff
@@ -259,6 +259,33 @@ export class Metric implements IMetric {
...
+      usingMetrics: { m0: metric },
+      period: props.period,
+    });
+
+    return new Alarm(scope, id, {
+      ...props,
+      metric: anomalyDetectionExpression,
+      threshold: 0, // This value will be ignored for anomaly detection alarms
+      thresholdMetricId: 'expr_1',
```
xazhao, line 285: should we expose the property to customers? Will create 2 anomaly detection alarms with the same `thresholdMetricId` cause conflicts?

scorbiere: `thresholdMetricId` is per alarm and these metric id are also local to the alarm. Which mean that there is no risk of conflict.

Conflict wise, the only risk with this change is if the user defines an expression metric with sub metrics with an `m0` identifier. I think it is fine to keep the complexity as it is now.

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
GavinZZ: Can we do this validation on CDK side to preemptively throw?

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -206,12 +230,22 @@ export class Alarm extends AlarmBase {
       ...metricProps,
     });
 
+    if (isAnomalyDetection) {
+      // Ensure returnData is set to true for both metrics in anomaly detection
+      const cfnMetrics = alarm.metrics as any[];
```
GavinZZ, line 235: can we make it stronger type instead or `any`?

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
GavinZZ: Can you maybe provide the documentation in the comment? This code seems weird and providing the doc in the comment for future reference would help.

GavinZZ (changes requested):

rix0rrr (commented):
API-wise, it seems  that to create an Anomaly detection alarm, 3 things need to happen:

- You must set `thresholdMetricId`
- You must wrap metric in a MetricMath that uses `ANOMALY_DETECTION_BAND`
- You must not set a `threshold`
- For an operator you must pick one of the special 3 comparison operators.

That feels like a lot of work, no?

rix0rrr (changes requested):
See previous review

GavinZZ:
@scorbiere are you still working on this PR or should we close it for now?

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
scorbiere: I changed it, but I will still need to cast as `any` in order to update the value of the `returnData` property.

packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts
```diff
@@ -259,6 +259,33 @@ export class Metric implements IMetric {
...
+    const anomalyDetectionExpression = new MathExpression({
+      expression: `ANOMALY_DETECTION_BAND(m0, ${props.bounds ?? 1})`,
+      usingMetrics: { m0: metric },
+      period: props.period,
+    });
+
+    return new Alarm(scope, id, {
+      ...props,
+      metric: anomalyDetectionExpression,
+      threshold: 0, // This value will be ignored for anomaly detection alarms
```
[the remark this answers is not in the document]
scorbiere: Wouldn't that be a breaking change? In the updated version of this PR I am adding a constant `THRESHOLD_IGNORED_FOR_ANOMALY_DETECTION` to avoid the change of `threshold` to optional.

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -206,12 +230,22 @@ export class Alarm extends AlarmBase {
...
+      if (cfnMetrics && cfnMetrics.length >= 2) {
+        cfnMetrics[0].returnData = true;
+        cfnMetrics[1].returnData = true;
```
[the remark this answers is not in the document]
scorbiere: I updated the logic to make sure we will modify the relevant metrics.

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -190,7 +213,8 @@ export class Alarm extends AlarmBase {
 
       // Evaluation
       comparisonOperator,
-      threshold: props.threshold,
+      threshold: isAnomalyDetection ? undefined : props.threshold,
+      thresholdMetricId: isAnomalyDetection ? (props.thresholdMetricId || 'expr_1') : undefined,
```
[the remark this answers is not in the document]
scorbiere: Most likely given the logic used to create the unique metric id. But I changed the implementation to explicitly look for the metric which uses the `ANOMALY_DETECTION_BAND` in its expression.

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -15,6 +15,13 @@ import { ArnFormat, Lazy, Stack, Token, Annotations } from '../../core';
  * Properties for Alarms
  */
 export interface AlarmProps extends CreateAlarmOptions {
+
+  /**
+   * In an alarm based on an anomaly detection model, this is the ID of the ANOMALY_DETECTION_BAND function used as the threshold for the alarm.
+   *
+   * @default - No metric id.
+   */
+  readonly thresholdMetricId?: string;
```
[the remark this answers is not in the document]
scorbiere: In the new version of the PR, I was really wondering how it would make sense to keep this metric id in the props, as it is making a lot more sense to let our logic find the correct metric id after the internal method `renderMetric` got executed.

packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts
```diff
@@ -259,6 +259,33 @@ export class Metric implements IMetric {
...
+  /**
+ * Make a new Alarm for this metric using anomaly detection
+ *
+ * @param scope The scope in which to create the alarm
+ * @param id The ID of the alarm
+ * @param props Alarm creation properties for anomaly detection
+ * @param metric The metric to create the alarm for
+ * @returns The newly created Alarm
+ */
+  public static createAnomalyDetectionAlarmFromMetric(scope: Construct, id: string, props: CreateAnomalyDetectionAlarmProps, metric: IMetric): Alarm {
```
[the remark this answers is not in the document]
rix0rrr: I think this method makes more sense on `Alarm`.

```ts
Alarm.createAnomalyDetectionAlarm(...)
```

?

rix0rrr (approved):
Conditional approval. I'd definitely like to have some minor things changed.

I'm not entirely happy with the concepts/classes design, but I can't think of anything better myself, really, other than a `new AnomalyDetectionAlarm()` to isolate all the anomaly detection special casing that's now going into `new Alarm()`.

I think I'd strongly suggest that, but I'll leave it to you to decide what to do with.

packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts
```diff
@@ -560,6 +595,17 @@ export class Metric implements IMetric {
...
+  /**
+   * Make a new Alarm for this metric using anomaly detection
+   *
+   * @param scope The scope in which to create the alarm
+   * @param id The ID of the alarm
+   * @param props Alarm creation properties for anomaly detection
+   */
+  public createAnomalyDetectionAlarm(scope: Construct, id: string, props: CreateAnomalyDetectionAlarmProps): Alarm {
+    return Metric.createAnomalyDetectionAlarmFromMetric(scope, id, props, this);
+  }
```
rix0rrr, line 607: Adding these helpers to `IMetric` feels unnecessary. I understand you're probably doing it because `createAlarm()` exists, but I regret that as well :). 

What about 2 static factory functions, and then we're doing:

```ts
Alarm.createAnomalyDetectionAlarm(originalMetric, { stdDevs: 1, ... })

Metric.createAnomalyDetectionBandMetric(originalMetric, { stdDevs: 1, ... })
```

And that's enough?

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -137,6 +163,16 @@ export class Alarm extends AlarmBase {
...
   }
 
+  /**
+   * Determine whether this operator is an anomaly detection operator.
+   *
+   * @param operator the comparison operator for the alarm.
+   * @returns true if the operator is an anomaly detection operator, false otherwise.
+   */
+  public static isAnomalyDetectionOperator(operator?: ComparisonOperator): boolean {
+    return operator !== undefined && ANOMALY_DETECTION_OPERATORS.includes(operator);
```
GavinZZ, line 173: For my understanding, existing alarms can't use one of the new operator?

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -413,6 +498,149 @@ export class Alarm extends AlarmBase {
...
+  constructor(scope: Construct, id: string, props: AnomalyDetectionAlarmProps) {
+    super(scope, id, {
+      ...props,
+      metric: AnomalyDetectionAlarm.createAnomalyDetectionMetric(scope, props),
+      threshold: THRESHOLD_IGNORED_FOR_ANOMALY_DETECTION,
+    });
+  }
+
+  protected override onCfnAlarmCreated(alarm: CfnAlarm): void {
+    super.onCfnAlarmCreated(alarm);
```
GavinZZ, line 545: this, I believe does nothing right?

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -170,6 +206,9 @@ export class Alarm extends AlarmBase {
 
     const comparisonOperator = props.comparisonOperator || ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD;
 
+    this.validateComparisonOperator(comparisonOperator);
+    this.validateThreshold(props.threshold);
```
GavinZZ, line 210: I assume this is to disallow the use of Anomaly detection alarm using the Alarm construct. However, is this really necessary? Would there be a case where existing users use Alarm construct to get the anomaly detection alarm behaviour and this change would break them?

packages/aws-cdk-lib/aws-cloudwatch/lib/metric.ts
```diff
@@ -303,6 +303,23 @@ export class Metric implements IMetric {
...
+  /**
+   * Make a new Alarm for this metric using anomaly detection
+   *
+   * @param scope The scope in which to create the alarm
+   * @param id The ID of the alarm
+   * @param props Alarm creation properties for anomaly detection
+   * @param metric The metric to create the alarm for
+   * @returns The newly created Alarm
+   */
+  public static createAnomalyDetectionAlarmFromMetric(scope: Construct, id: string, props: CreateAnomalyDetectionAlarmProps, metric: IMetric): Alarm {
```
GavinZZ, line 315: nit: I didn't find this meaningful when there's also `createAnomalyDetectionAlarm` method

GavinZZ (commented):
Left some comments to help me understand the changes.

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -387,6 +387,85 @@ only supports filtering by `unit` for Alarms, not in Dashboard graphs.
...
+
+// Create an anomaly detection alarm with default settings
+const alarm = metric.createAnomalyDetectionAlarm(this, 'AnomalyAlarm', {
+  stdDevs: 2, // Number of standard deviations for the band (default: 2)
+  evaluationPeriods: 3,
+  comparisonOperator: cloudwatch.ComparisonOperator.LESS_THAN_LOWER_OR_GREATER_THAN_UPPER_THRESHOLD,
+});
+
+// Or, create an anomaly detection alarm with custom settings
+const customAlarm = metric.createAnomalyDetectionAlarm(this, 'CustomAnomalyAlarm', {
```
rix0rrr, line 448: The only difference between this and the previous example is `alarmDescription`.

Not really that much to customize... 😬

packages/aws-cdk-lib/aws-cloudwatch/README.md
```diff
@@ -387,6 +387,85 @@ only supports filtering by `unit` for Alarms, not in Dashboard graphs.
...
+// Create a metric
+const metric = new cloudwatch.Metric({
+  namespace: 'AWS/EC2',
+  metricName: 'CPUUtilization',
+  statistic: 'Average',
+  period: Duration.minutes(5),
+});
+
+// Create an anomaly detection alarm with default settings
+const alarm = metric.createAnomalyDetectionAlarm(this, 'AnomalyAlarm', {
```
rix0rrr, line 441: I still hate this and regret `metric.createAlarm()`. 

Please consider one more time to not add this? 😉

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
rix0rrr: Here is my understanding, correct me if I'm wrong @scorbiere:

* These operators only make sense in combination with an `ANOMALY_DETECTION_BAND()` function
* Existing Alarms are a superset of Anomaly Detection alarms.

So, if a user puts an `ANOMALY_DETECTION_BAND` function into their metric they can use the operator with the old Alarm, otherwise not.

Just by the fact that the user is using `Alarm` we can't tell whether the operator is correct or not, but we
could conceivably combine it with some sort of validation on the presence of the right function in `Alarm`.

```ts
if (usesAnomalyFunction(props.metric) !== isAnomlyDetectionOperator(props.operator)) {
  throw new Error('oh no');
}
```

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -137,6 +163,16 @@ export class Alarm extends AlarmBase {
     return new Import(scope, id);
   }
 
+  /**
+   * Determine whether this operator is an anomaly detection operator.
+   *
+   * @param operator the comparison operator for the alarm.
+   * @returns true if the operator is an anomaly detection operator, false otherwise.
+   */
+  public static isAnomalyDetectionOperator(operator?: ComparisonOperator): boolean {
```
rix0rrr, line 172: Does this function need to be public?

I would just make it a module-local free-floating function.

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -208,26 +248,71 @@ export class Alarm extends AlarmBase {
...
     }
   }
+  protected validateComparisonOperator(comparisonOperator: ComparisonOperator): void {
+    if (Alarm.isAnomalyDetectionOperator(comparisonOperator)) {
+      throw new ValidationError(
+        'Anomaly detection operators require to use the construct AnomalyDetectionAlarm.',
+        this,
+      );
+    }
+  }
```
rix0rrr, line 276: Yeah see my above comment. I don't think this is a good idea.

People should be allowed to use an `Alarm` to define an anomaly alarm (in fact, they probably already are today, because they have no other choice!)... it's just that the new construct makes it easier.

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -208,26 +248,71 @@ export class Alarm extends AlarmBase {
...
+  protected validateThreshold(threshold?: number): void {
+    // For standard alarms, we need a threshold
+    if (threshold === undefined) {
+      throw new ValidationError('threshold must be specified for standard alarms', this);
+    }
+  }
```
rix0rrr, line 282: Again this is not necessarily a "standard" alarm.

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -413,6 +498,149 @@ export class Alarm extends AlarmBase {
...
+    super(scope, id, {
+      ...props,
+      metric: AnomalyDetectionAlarm.createAnomalyDetectionMetric(scope, props),
+      threshold: THRESHOLD_IGNORED_FOR_ANOMALY_DETECTION,
+    });
+  }
+
+  protected override onCfnAlarmCreated(alarm: CfnAlarm): void {
+    super.onCfnAlarmCreated(alarm);
+    this.ensureAnomalyMetricsReturnData(alarm);
```
rix0rrr, line 546: What's this protection against? Can we not set it up so it definitely works?

In any case, I don't like the pattern of making a new "subclass function hole" in this way.

Use `this.node.addValidation()` instead.

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -208,26 +248,71 @@ export class Alarm extends AlarmBase {
...
+
+  /**
+   * Gets the thresholdMetricId for CloudFormation alarm properties.
+   *
+   * @param _metrics - The array of metric data queries to validate against
+   * @returns The threshold metric ID to use, or undefined if none is applicable
+   */
+  protected getThresholdMetricIdForCfnAlarmProps(_metrics: CfnAlarm.MetricDataQueryProperty[]): string | undefined {
+    return undefined;
+  }
```
rix0rrr, line 306: I don't like any of this. Can we not pass in the properties into the `super()` constructor?

rix0rrr (changes requested):

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
scorbiere: The actual `Alarm` construct do not handle the utilization of an anomaly detection operator. The changes in this PR cannot break existing user experience with anomaly detection scenario.

rix0rrr (approved):
I made some changes to this PR. Have a look through and see if you agree, involve Mohamed as well.

If you agree, you can merge it by removing the **do-not-merge** label (and dismissing Mohamed's review)

packages/aws-cdk-lib/aws-cloudwatch/lib/alarm.ts
```diff
@@ -169,11 +218,48 @@ export class Alarm extends AlarmBase {
...
+    let thresholdMetricId: string | undefined; // For anomaly detection
+    if (isAnomalyDetection) {
+      if (!isAnomalyDetectionMetric(props.metric)) {
+        throw new ValidationError(
+          `Anomaly detection operator ${comparisonOperator} requires an ANOMALY_DETECTION_BAND() metric. Use the construct AnomalyDetectionAlarm or wrap your metric in an ANOMALY_DETECTION_BAND expression.`,
+          this,
+        );
+      }
+
+      // For now the property `threshold` is required, so checking it against 'undefined' is a bit of a stretch, but reasonable.
```
GavinZZ, line 233: Since `threshold` is required, should we have a validation error for missing threshold?

packages/@aws-cdk-testing/framework-integ/test/aws-dynamodb/test/integ.dynamodb.alarm-metrics.js.snapshot/alarm-metrics.template.json
```diff
@@ -32,7 +32,8 @@
      {
       "Expression": "putitem + scan",
       "Id": "expr_1",
-      "Label": "Sum of throttled requests across all operations"
+      "Label": "Sum of throttled requests across all operations",
+      "ReturnData": true
```
GavinZZ, line 36: any way we can change the render function to hide the `ReturnData: true` for current behaviour?

packages/aws-cdk-lib/rosetta/aws_cloudwatch/default.ts-fixture
```diff
@@ -5,6 +5,7 @@ import * as cloudwatch from 'aws-cdk-lib/aws-cloudwatch';
 import * as elbv2 from 'aws-cdk-lib/aws-elasticloadbalancingv2';
 import * as route53 from 'aws-cdk-lib/aws-route53';
 import * as sns from 'aws-cdk-lib/aws-sns';
+import * as sqs from 'aws-cdk-lib/aws-sqs';
```
GavinZZ, line 8: is this still needed?

packages/aws-cdk-lib/aws-cloudwatch/test/alarm.test.ts
```diff
@@ -457,6 +457,27 @@ describe('Alarm', () => {
     expect(alarmFromName.alarmName).toEqual('TestAlarmName');
     expect(alarmFromName.alarmArn).toMatch(/:alarm:TestAlarmName$/);
   });
+
+  describe('Anomaly Detection', () => {
```
GavinZZ, line 461: can we have a unit test to use anomaly detection alarm using alarm construct?

GavinZZ (commented):
Thanks Seb! LGTM, just some minor comments.

packages/@aws-cdk-testing/framework-integ/test/aws-dynamodb/test/integ.dynamodb.alarm-metrics.js.snapshot/alarm-metrics.template.json
scorbiere: According to the documentation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudwatch-alarm-metricdataquery.html#cfn-cloudwatch-alarm-metricdataquery-returndata

> When you create an alarm based on a metric math expression, specify True for this value for only the one math expression that the alarm is based on. You must specify False for ReturnData for all the other metrics and expressions used in the alarm.


Which means that the new version of the `renderMetric` method seems to be fixing a bug where the `ReturnData` of the `MathExpression` of the alarm was set to `false` instead of `true`.

packages/aws-cdk-lib/rosetta/aws_cloudwatch/default.ts-fixture
scorbiere: Good catch. Since I changed the examples in the README, we don't need this anymore.

GavinZZ (approved):
Thank you!

PR #31246, opened 2024-08-28
Unit test for using escape hatch to scope down Key Policy

samson-keung:
Adding a unit test to make sure user is able to scope down the key policy to specific distribution ID via escape hatch.

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
+++ b/packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
@@ -167,6 +167,8 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
 
   /**
    * Create a S3 Origin with Origin Access Identity (OAI) configured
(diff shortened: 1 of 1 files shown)


packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -167,8 +167,6 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
 
   /**
    * Create a S3 Origin with Origin Access Identity (OAI) configured
-   * OAI is a legacy feature and we **strongly** recommend you to use OAC via `withOriginAccessControl()`
```
gracelu0, line 170: Is there a reason for removing this from the docstring?

samson-keung: No. It was a fat finger mistake. Added back now. Thanks for catching it!

gracelu0 (approved):

PR #31254, opened 2024-08-29
feat(cloudfront): s3 origin access control L2 construct

### LINKED ISSUE #21771: cloudfront: Support Origin Access Control

gracelu0:
### Issue # (if applicable)

Closes #21771 .

### Reason for this change

L2 support for setting up OAC for S3 origins.

See [RFC 617](https://github.com/aws/aws-cdk-rfcs/blob/main/text/0617-cloudfront-oac-l2.md) for more details.

### Description of changes

- New L2 construct `OriginAccessControl`
- Deprecation of `CloudFrontWebDistribution` — replaced by existing `Distribution` construct
- Deprecation of `S3Origin` — replaced by new constructs `S3BucketOrigin` and `S3StaticWebsiteOrigin`
- README updates

Added unit tests and integration tests

Internal tracking: V1491594161

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-bedrock/lib/foundation-model.ts
+++ b/packages/aws-cdk-lib/aws-bedrock/lib/foundation-model.ts
@@ -239,6 +239,15 @@ export class FoundationModelIdentifier {
   /** Base model "stability.stable-diffusion-xl-v1:0". */
   public static readonly STABILITY_STABLE_DIFFUSION_XL_V1_0 = new FoundationModelIdentifier('stability.stable-diffusion-xl-v1:0');
 
+  /** Base model "stability.sd3-large-v1:0". */
+  public static readonly STABILITY_SD3_LARGE_V1_0 = new FoundationModelIdentifier('stability.sd3-large-v1:0');
+
+  /** Base model "stability.stable-image-ultra-v1:0". */
+  public static readonly STABILITY_STABLE_IMAGE_ULTRA_V1_0 = new FoundationModelIdentifier('stability.stable-image-ultra-v1:0');
+
+  /** Base model "stability.stable-image-core-v1:0". */
+  public static readonly STABILITY_STABLE_IMAGE_CORE_V1_0 = new FoundationModelIdentifier('stability.stable-image-core-v1:0');
+
   /**
    * Constructor for foundation model identifier
    * @param modelId the model identifier


GavinZZ (commented):
Great work Grace, I just reviewed the changes in `aws-cloudfront-origins`. Left some minor questions. Will continue reviewing the other module.

packages/aws-cdk-lib/aws-cloudfront-origins/README.md
```diff
@@ -1,26 +1,502 @@
...
+and the distribution can use built-in S3 redirects and S3 custom error pages.
+
+```ts
+const myBucket = new s3.Bucket(this, 'myBucket');
+new cloudfront.Distribution(this, 'myDist', {
+  defaultBehavior: { origin: new origins.S3StaticWebsiteOrigin({ bucket: myBucket }) },
+});
+```
+
+> Note: `S3Origin` has been deprecated. Use `S3BucketOrigin` for standard S3 origins and `S3StaticWebsiteOrigin` for static website S3 origins.
```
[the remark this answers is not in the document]
gracelu0: I can move it up

packages/aws-cdk-lib/aws-cloudfront-origins/README.md
```diff
@@ -1,26 +1,502 @@
...
+  objectOwnership: s3.ObjectOwnership.BUCKET_OWNER_ENFORCED,
+});
+new cloudfront.Distribution(this, 'myDist', {
+  defaultBehavior: {
+    origin: origins.S3BucketOrigin.withOriginAccessControl(myBucket) // Automatically grants Distribution access to `myKmsKey`
+  },
+});
+```
+
+I saw this warning message during synth time. What do I do?
```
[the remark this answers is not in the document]
gracelu0: I'll add a sub-header to create a section for this. I do think it should be easier to find since it has steps for user to follow

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,246 @@
...
+
+const KEY_ACTIONS: Record<string, string[]> = {
+  READ: ['kms:Decrypt'],
+  WRITE: ['kms:Encrypt', 'kms:GenerateDataKey*'],
+};
+
+/**
+ * Properties for configuring a origin using a standard S3 bucket
+ */
+export interface S3BucketOriginBaseProps extends cloudfront.OriginProps { }
```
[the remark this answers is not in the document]
gracelu0: This is for future-proofing in the case where there are new S3 origin specific props that are added. Similar props exist currently (see [load-balancer-origin.ts](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-cloudfront-origins/lib/load-balancer-origin.ts#L7))

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,246 @@
...
+  /**
+  * An optional Origin Access Control
+  * @default - an Origin Access Control will be created.
+  */
+  readonly originAccessControl?: cloudfront.IOriginAccessControl;
+
+  /**
+   * The level of permissions granted in the bucket policy and key policy (if applicable)
+   * to the CloudFront distribution.
+   * @default AccessLevel.READ
```
xazhao, line 38: nit: should the default be [AccessLevel.READ]?

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,246 @@
...
+        for (const accessLevel of new Set(accessLevels)) {
+          actions = actions.concat(BUCKET_ACTIONS[accessLevel]);
+        }
+        return actions;
+      }
+
+      getKeyPolicyActions(accessLevels: cloudfront.AccessLevel[]) {
+        let actions: string[] = [];
+        // Remove duplicates and filters out DELETE since delete permissions are not applicable to KMS key actions
+        const keyAccessLevels = [...new Set(accessLevels.filter(level => level !== AccessLevel.DELETE))];
```
xazhao, line 119: Should we move removing duplicates to `bind()` so we only need to do it once? `DELETE` check can be done in the loop below.

xazhao (commented):
Very nice work Grace! Excited to see this feature. Left a few comments and still need to look further into it.

paulhcsun (commented):
Good stuff, just a couple of small changes.

packages/aws-cdk-lib/aws-cloudfront-origins/README.md
```diff
@@ -1,26 +1,502 @@
...
+### S3 Bucket Configured as a Website Endpoint
+
+To set up an origin using a S3 bucket configured as a website endpoint, use the `S3StaticWebsiteOrigin` class. When the bucket is configured as a
+website endpoint, the bucket is treated as an HTTP origin,
+and the distribution can use built-in S3 redirects and S3 custom error pages.
+
+```ts
+const myBucket = new s3.Bucket(this, 'myBucket');
+new cloudfront.Distribution(this, 'myDist', {
+  defaultBehavior: { origin: new origins.S3StaticWebsiteOrigin({ bucket: myBucket }) },
```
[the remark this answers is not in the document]
paulhcsun: From our discussion, there will be a `bucket` parameter added to the `S3StaticWebsiteOrigin` class. The reason it was taking in the object before was because it extends the `HttpOrigin` class which. accepts props as an object. This change will make the definitions more consistent.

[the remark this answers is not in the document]
gracelu0: I added the bucket as a prop to `S3StaticWebsiteOriginProps` which extends `HttpOriginProps` , but I can update the API so the constructor it would look like:
```
origin: origins.S3BucketOrigin.withOriginAccessControl(myBucket)
origin: new origins.S3StaticWebsiteOrigin(myBucket)
```
which is more consistent

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: Good idea, will update to create the `Set` of AccessLevels and pass that into each function for getting the bucket and key policy actions

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: updated!

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -105,20 +106,21 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
         };
       }
 
-      getBucketPolicyActions(accessLevels: cloudfront.AccessLevel[]) {
+      getBucketPolicyActions(accessLevels: Set<cloudfront.AccessLevel>) {
```
samson-keung, line 109: This method and `getKeyPolicyActions` should be private I believe?

gracelu0: Yes, thanks!

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,248 @@
...
+ */
+export interface S3BucketOriginBaseProps extends cloudfront.OriginProps { }
+
+/**
+ * Properties for configuring a S3 origin with OAC
+ */
+export interface S3BucketOriginWithOACProps extends S3BucketOriginBaseProps {
+  /**
+  * An optional Origin Access Control
+  * @default - an Origin Access Control will be created.
```
paulhcsun, line 31:
```suggestion
  * An optional Origin Access Control
  *
  * @default - an Origin Access Control will be created.
```
nit: should leave a line break between the description and the default line

packages/aws-cdk-lib/aws-cloudfront-origins/README.md
```diff
@@ -1,26 +1,514 @@
...
+  defaultBehavior: {
+    origin: s3Origin
+  },
+});
+```
+
+An existing S3 origin access control can be imported using the `fromOriginAccessControlId` method:
+
+```ts
+const importedOAC = cloudfront.S3OriginAccessControl.fromOriginAccessControlId(this, 'myImportedOAC', 'ABC123ABC123AB');
```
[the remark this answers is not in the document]
gracelu0: Yes, typically it starts with `E` but the format is a 14 char string of uppercase letters and numbers

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,248 @@
...
+            'See the "Setting up OAC with imported S3 buckets" section of module\'s README for more info.');
+        }
+
+        if (bucket.encryptionKey) {
+          const keyPolicyActions = this.getKeyPolicyActions(accessLevels);
+          const keyPolicyResult = this.grantDistributionAccessToKey(keyPolicyActions, bucket.encryptionKey);
+          // Failed to update key policy, assume using imported key
+          if (!keyPolicyResult.statementAdded) {
+            Annotations.of(scope).addWarningV2('@aws-cdk/aws-cloudfront-origins:updateImportedKeyPolicy',
+              'Cannot update key policy of an imported key. You may need to update the policy manually instead.');
```
paulhcsun, line 96: Is there some other plausible cause for why a user cant update key policy of an imported key? If so we should list it or just change this wording to say `You need to update...`

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: updated this for all property docstrings with defaults

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,248 @@
...
+              ArnLike: {
+                'AWS:SourceArn': `arn:${Aws.PARTITION}:cloudfront::${Aws.ACCOUNT_ID}:distribution/*`,
+              },
+            },
+          },
+        );
+        Annotations.of(key.node.scope!).addWarningV2('@aws-cdk/aws-cloudfront-origins:wildcardKeyPolicyForOac',
+          'To avoid circular dependency between the KMS key, Bucket, and Distribution, ' +
+          'a wildcard is used to match all Distribution IDs in Key policy condition.\n' +
+          'To further scope down the policy for best security practices, see the "Using OAC for a SSE-KMS encrypted S3 origin" section in the module README.');
```
paulhcsun, line 166:
```suggestion
          'To avoid a circular dependency between the KMS key, Bucket, and Distribution during the initial deployment, ' +
          'a wildcard is used in the Key policy condition to match all Distribution IDs.\n' +
          'After deploying once, it is recommended to further scope down the policy for best security practices by following the guidance in the "Using OAC for a SSE-KMS encrypted S3 origin" section in the module README.');
```

Changing the wording to make it clearer that it is recommended to further scope down the policy after the initial deployment. 

Reminder to also change the warning in the README along with this.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,248 @@
...
+          Annotations.of(scope).addWarningV2('@aws-cdk/aws-cloudfront-origins:updateImportedBucketPolicyOai',
+            'Cannot update bucket policy of an imported bucket. You will need to update the policy manually instead.\n' +
+            'See the "Setting up OAI with imported S3 buckets (legacy)" section of module\'s README for more info.');
+        }
+        return this._bind(scope, options);
+      }
+
+      protected renderS3OriginConfig(): cloudfront.CfnDistribution.S3OriginConfigProperty | undefined {
+        if (!this.originAccessIdentity) {
+          throw new Error('Origin access identity cannot be undefined');
```
paulhcsun, line 221:
```suggestion
          throw new Error('Origin access identity must be defined.');
```
Nit: is `null` an acceptable value? If not then maybe we use the reverse wording for this.

packages/aws-cdk-lib/aws-cloudfront/lib/origin-access-control.ts
```diff
@@ -0,0 +1,208 @@
...
+ */
+export interface OriginAccessControlBaseProps {
+  /**
+   * A description of the origin access control.
+   *
+   * @default - no description
+   */
+  readonly description?: string;
+  /**
+   * A name to identify the origin access control. You can specify up to 64 characters.
```
paulhcsun, line 27:
```suggestion
   * A name to identify the origin access control, with a maximum length of 64 characters.
```

packages/aws-cdk-lib/aws-cloudfront/lib/origin.ts
```diff
@@ -119,6 +126,13 @@ export interface OriginBindOptions {
    * as assigned by the Distribution this Origin has been used added to.
    */
   readonly originId: string;
+
+  /**
+   * The identifier of the Distribution this Origin is used for.
+   * This is used to grant origin access permissions to the distribution for origin access control.
+   * @default - no distribution id
```
paulhcsun, line 133:
```suggestion
   * This is used to grant origin access permissions to the distribution for origin access control.
   *
   * @default - no distribution id
```

paulhcsun:
A few more minor changes, otherwise this looks good to me. 

Amazing work @gracelu0! Can't wait for this to be released!

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: I don't think so, I'll update the wording and point to the README section that describes the policy they need to add

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: The user doesn't have to pass in an OAI, if they don't then one will automatically be created. This is catching the case where somehow an OAI was not created at all, so I think it's ok to leave as is

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,251 @@
...
+const BUCKET_ACTIONS: Record<string, string[]> = {
+  READ: ['s3:GetObject'],
+  WRITE: ['s3:PutObject'],
+  DELETE: ['s3:DeleteObject'],
+};
+
+const KEY_ACTIONS: Record<string, string[]> = {
+  READ: ['kms:Decrypt'],
+  WRITE: ['kms:Encrypt', 'kms:GenerateDataKey*'],
+};
```
xazhao, line 18: Should we have have them defined within `origin-access-controls.ts` so later when we add lambda urls we have a single place to track all permissions?

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,248 @@
...
+  public static withOriginAccessControl(bucket: IBucket, props?: S3BucketOriginWithOACProps): cloudfront.IOrigin {
+    return new class extends S3BucketOrigin {
+      private originAccessControl?: cloudfront.IOriginAccessControl;
+
+      constructor() {
+        super(bucket, { ...props });
+        this.originAccessControl = props?.originAccessControl;
+      }
+
+      public bind(scope: Construct, options: cloudfront.OriginBindOptions): cloudfront.OriginBindConfig {
```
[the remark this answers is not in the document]
gracelu0: Correct, the `bind` method is triggered by the `Distribution` when the origin is passed in for the first time, e.g.
```
const s3Origin = origins.S3BucketOrigin.withOriginAccessControl(myBucket, {
    originAccessControl: oac 
  }
)
new cloudfront.Distribution(this, 'myDist', {
  defaultBehavior: {
    origin: s3Origin // calls bind() on this origin
  },
});
```

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,248 @@
...
+
+/**
+ * A S3 Bucket Origin
+ */
+export abstract class S3BucketOrigin extends cloudfront.OriginBase {
+  /**
+   * Create a S3 Origin with Origin Access Control (OAC) configured
+   */
+  public static withOriginAccessControl(bucket: IBucket, props?: S3BucketOriginWithOACProps): cloudfront.IOrigin {
+    return new class extends S3BucketOrigin {
```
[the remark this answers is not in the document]
GavinZZ: FYI, there's a discussion around using a static method here https://github.com/aws/aws-cdk/pull/30663#discussion_r1653550554

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  public static withBucketDefaults(bucket: IBucket, props?: cloudfront.OriginProps): cloudfront.IOrigin {
+    return new class extends S3BucketOrigin {
+      constructor() {
+        super(bucket, { ...props });
+      }
+    }();
+  }
+
+  constructor(bucket: IBucket, props: S3BucketOriginBaseProps) {
```
GavinZZ, line 239: should the `prop` be an optional? It seems that all properties are optional and `withOriginAccessControl` can take undefined props.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,251 @@
...
+    super(bucket.bucketRegionalDomainName, props);
+  }
+
+  /** @internal */
+  protected _bind(scope: Construct, options: cloudfront.OriginBindOptions): cloudfront.OriginBindConfig {
+    return super.bind(scope, options);
+  }
+
+  protected renderS3OriginConfig(): cloudfront.CfnDistribution.S3OriginConfigProperty | undefined {
+    return { originAccessIdentity: '' };
```
GavinZZ, line 249: Any documentation supporting `''`?

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: Agreed, will update

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,251 @@
...
+          },
+        };
+      }
+
+      private getBucketPolicyActions(accessLevels: Set<cloudfront.AccessLevel>) {
+        let actions: string[] = [];
+        for (const accessLevel of accessLevels) {
+          actions = actions.concat(BUCKET_ACTIONS[accessLevel]);
+        }
+        return actions;
```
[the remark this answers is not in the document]
gracelu0: It's because `BUCKET_ACTIONS` (and `KEY_ACTIONS`) map the keys to an array of permissions:
```
const BUCKET_ACTIONS: Record<string, string[]> = {
  READ: ['s3:GetObject'],
  WRITE: ['s3:PutObject'],
  DELETE: ['s3:DeleteObject'],
};
```
so we need `concat` to merge the arrays

packages/aws-cdk-lib/aws-cloudfront-origins/README.md
```diff
@@ -1,26 +1,514 @@
...
+### S3 Bucket Configured as a Website Endpoint
+
+To set up an origin using an S3 bucket configured as a website endpoint, use the `S3StaticWebsiteOrigin` class. When the bucket is configured as a
+website endpoint, the bucket is treated as an HTTP origin,
+and the distribution can use built-in S3 redirects and S3 custom error pages.
+
+```ts
+const myBucket = new s3.Bucket(this, 'myBucket');
+new cloudfront.Distribution(this, 'myDist', {
+  defaultBehavior: { origin: new origins.S3StaticWebsiteOrigin(myBucket) },
```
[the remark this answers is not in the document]
gracelu0: Since OAC/OAI are not supported for `S3StaticWebsiteOrigin`, it would be superfluous to use a static function to instantiate it because there is only one type of configuration. From my understanding we use static functions in CDK to help users intuitively pick different configuration types for resources.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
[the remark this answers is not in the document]
gracelu0: This is an anonymous class that implements the base class methods according to each configuration (withOAC, withOAI, withBucketDefaults). An example of this pattern can be found in DynamoDb module [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-dynamodb/lib/encryption.ts#L14-L27).

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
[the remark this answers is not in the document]
GavinZZ: You don't have to change it here, but just a FYI, you can also do something like `list.push(...[A,B,C])`.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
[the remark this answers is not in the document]
xazhao: So if we extract the nested class out of `return`, we can still keep the same return type.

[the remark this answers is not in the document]
xazhao: my point is this class is not tiny and defining it as an anonymous class makes it hard to review.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,251 @@
...
+        if (!this.originAccessControl) {
+          this.originAccessControl = new cloudfront.S3OriginAccessControl(scope, 'S3OriginAccessControl');
+        }
+
+        const distributionId = options.distributionId;
+        const accessLevels = new Set(props?.originAccessLevels ?? [cloudfront.AccessLevel.READ]);
+        const bucketPolicyActions = this.getBucketPolicyActions(accessLevels);
+        const bucketPolicyResult = this.grantDistributionAccessToBucket(distributionId!, bucketPolicyActions);
+
+        // Failed to update bucket policy, assume using imported bucket
```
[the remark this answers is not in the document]
gracelu0: I think so yes, looking at the  `[addToResourcePolicy](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-s3/lib/bucket.ts#L651-L662)` method, it will not create a policy (fail) when the `this.policy` is undefined. `this.policy` is undefined when there is no existing policy and `this.autoCreatePolicy` is set to `false`, which is the case for imported buckets (see [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-s3/lib/bucket.ts#L1774)). For new buckets `this.autoCreatePolicy` is set to `true` (see [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-s3/lib/bucket.ts#L1909)), and the property is protected so it cannot be set by the user.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: Yes I think that's a good idea, will update

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: Yes, see [CloudFormation docs for `S3OriginConfig` property](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-s3originconfig.html). 
`If you're using origin access control (OAC) instead of origin access identity, specify an empty OriginAccessIdentity element. `
`If you want viewers to be able to access objects using either the CloudFront URL or the Amazon S3 URL, specify an empty OriginAccessIdentity element.`
`To delete the origin access identity from an existing distribution, update the distribution configuration and include an empty OriginAccessIdentity element.`

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: yes good catch, will update

GavinZZ (approved):
LGTM, thanks Grace for the great work!

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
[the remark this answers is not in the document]
gracelu0: Ohh okay gotcha, I can abstract it out as a subclass and return that. I agree that is more readable

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
samson-keung: > Should we have have them defined within `origin-access-controls.ts` so later when we add lambda urls we have a single place to track all permissions?

Two questions:
1. Does the lambda use case involve KMS key? I thought these KMS permissions only applies on the S3 Bucket use case.
2. If we define these in `origin-access-controls.ts`, does that mean we need to export them? If we export them, users can start using them and changing them may lead to breaking changes. Is my understanding correct?

paulhcsun (approved):
Noice, great work @gracelu0!

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,251 @@
...
+
+      public bind(scope: Construct, options: cloudfront.OriginBindOptions): cloudfront.OriginBindConfig {
+        if (!this.originAccessControl) {
+          this.originAccessControl = new cloudfront.S3OriginAccessControl(scope, 'S3OriginAccessControl');
+        }
+
+        const distributionId = options.distributionId;
+        const accessLevels = new Set(props?.originAccessLevels ?? [cloudfront.AccessLevel.READ]);
+        const bucketPolicyActions = this.getBucketPolicyActions(accessLevels);
+        const bucketPolicyResult = this.grantDistributionAccessToBucket(distributionId!, bucketPolicyActions);
```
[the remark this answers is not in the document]
gracelu0: It will not throw an error but it will be an non-functional bucket policy statement since `undefined` cannot be a valid distribution ID

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -240,3 +87,167 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
     return { originAccessIdentity: '' };
   }
 }
+
+export class S3BucketOriginWithOAC extends S3BucketOrigin {
```
samson-keung, line 91: Let's keep this class private if possible so we do not expose more than needed. I am worried that if we expose this then changing the API of this class will become breaking change.

Same goes to the `S3BucketOriginWithOAI` class below

samson-keung (changes requested):

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: Yes these should definitely be private, thanks!

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
gracelu0: 1. Setting up OAC for Lambda function url origins does not involve KMS key. 
2. Yes we would export them, I think I see your point about possible breaking changes to external consumers if we change it. But wouldn't it be a breaking change if we didn't export and updated the permissions? Since that would affecting any customers using the construct. Also I see we have permission files that export sets of permissions ([example](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-s3/lib/perms.ts)) which are also used for grant methods.
I will move it back to the s3 origin file though to keep it private and encapsulated, thanks for pointing it out.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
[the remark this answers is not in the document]
GavinZZ: I think it will throw an error for `undefined!`. However, seeing [where this is triggered ](https://github.com/aws/aws-cdk/blob/gracelu0/s3-oac-l2/packages/aws-cdk-lib/aws-cloudfront/lib/distribution.ts#L610), I think we're pretty safe to say that it will always be defined.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
GavinZZ: I think that's a fair callout from Samson. While both scenarios can be a breaking changes, they are technically two different breaking changes. Keeping it in this file and don't allow users to access it directly reduce the chance of breaking changes (if we will ever change the value) for future proofing.

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -0,0 +1,264 @@
...
+      'To avoid a circular dependency between the KMS key, Bucket, and Distribution during the initial deployment, ' +
+      'a wildcard is used in the Key policy condition to match all Distribution IDs.\n' +
+      'After deploying once, it is strongly recommended to further scope down the policy for best security practices by ' +
+      'following the guidance in the "Using OAC for a SSE-KMS encrypted S3 origin" section in the module README.');
+    const result = key.addToResourcePolicy(oacKeyPolicyStatement);
+    return result;
+  }
+};
+
+
```
GavinZZ, line 215: This will fail the build due to linting issue.

xazhao (approved):
LGTM! Thanks for the work!

packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
xazhao: That's a good callout. I agree we should keep it private in s3 origin file to minimize breaking changes.

samson-keung (approved):
Excited to see this launch. Thank you for the hard work

gracelu0:
Thank you all for reviewing!! @paulhcsun @GavinZZ @xazhao @samson-keung

gracelu0:
@Mergifyio update

PR #31261, opened 2024-08-30
fix: pipelines ties cli version with cdk-assets version

### LINKED ISSUE #31253: aws-cdk-lib > codepipeline: Setting `cliVersion` can cause an invalid version of `cdk-assets` to attempt to be installed
### Describe the bug

The [property `cliVersion`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.pipelines.CodePipeline.html#cliversion) in `CodePipeline` is used to infer the [version of `cdk-assets` to install in the `publishAssetsAction`](https://github.com/aws/aws-cdk/blame/main/packages/aws-cdk-lib/pipelines/lib/codepipeline/codepipeline.ts#L820C18-L820C18).

Since #31119 was released, this causes the install to fail and the pipeline to break.

### Regression Issue

2.153.0

### Expected Behavior

The pipeline buildspec will be generated with a command to install a version of `cdk-assets` that is appropriate for the given version of `aws-cdk`.

### Current Behavior

The pipeline buildspec will be generated with a command to install `cdk-assets` with an invalid version string. This causes the build to fail with the following error.

```
[Container] 2024/08/29 14:43:38.596103 Running command npm install -g cdk-assets@2.154.1
npm ERR! code ETARGET
npm ERR! notarget No matching version found for cdk-assets@2.154.1.
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.
```

### Reproduction Steps

Building this CDK Stack will generate a buildspec with the invalid `npm install` command. Note that `CrossAccountZoneDelegationRecord` is what triggers the behavior.

The full repo can be found [here](https://github.com/sethmiller/cdk-assets-bug).

```typescript
import { App, Stack, StackProps, Stage, StageProps } from 'aws-cdk-lib';
import { Role } from 'aws-cdk-lib/aws-iam';
import { CrossAccountZoneDelegationRecord, HostedZone } from 'aws-cdk-lib/aws-route53';
import { ShellStep, CodePipeline, CodePipelineSource } from 'aws-cdk-lib/pipelines';

class MyServiceStack extends Stack {
    super(scope, id, props);

    // This Construct adds a node with the type `publish-assets` which triggers the behavior.
    new CrossAccountZoneDelegationRecord(this, 'cross-account-zone', {
      delegatedZone: HostedZone.fromHostedZoneAttributes(this, 'my-hosted-zone', {
        hostedZoneId: 'Z01201272X6Y0ABCDE0FG',
        zoneName: 'zone.name',
      }),
      delegationRole: Role.fromRoleName(this, 'my-role', 'my-role-name'),
      parentHostedZoneId: 'parent-hosted-zone',
    });
  }
}

class MyStage extends Stage {
  constructor(scope: Construct, id: string, props?: StageProps) {
    super(scope, id, props);

    new MyServiceStack(this, 'my-stack');
  }
}

  constructor(scope: Construct, id: string, props: StackProps = {}) {
    super(scope, id, props);

    const pipeline = new CodePipeline(this, 'my-pipeline', {
      synth: new ShellStep('my-build', {
        input: CodePipelineSource.gitHub(
          'my-org/my-repo',
          'main',
        ),
        commands: [
          "echo 'hello world'",
        ],
      }),
      cliVersion: '2.154.1',
    });

    const wave = pipeline.addWave('my-wave');
    wave.addStage(new MyStage(this, 'my-stage'));
  }
}

const devEnv = {
  region: process.env.CDK_DEFAULT_REGION,
};

const app = new App();

new MyStack(app, 'cdk-assets-bug-dev', { env: devEnv });

app.synth();
```

### Possible Solution

Either don't try to infer the version of `cdk-assets` at all or allow it to be set explicitly if `cliVersion` is set.

[This commit](https://github.com/aws/aws-cdk/commit/a31513a4f5ddc8fa79ade1869dfd90a5bde70d40) was the source of this issue (PR #31119).

### CDK CLI Version

2.154.1

### Framework Version

_No response_

### Node.js Version

v20.15.1

### OS

Linux

### Language

TypeScript

### Language Version

5.5.4

### Other information

_No response_

TheRealAmazonKendra:
cdk-assets is no longer being versioned in line with the cli or with aws-cdk-lib. Because every change must be backward compatible, we can just use `latest` for cdk-assets (not that we will soon be releasing a new major version of cdk-assets but we are releasing that with the `v3-latest` tag at present. When we are ready to swap versions, we will begin labeling the v3 line as `latest` and make the v2 line `v2-latest` as it will be in maintenance mode.

Note: the linter and potentially the build will fail for this initially because an integ test needs to be updated.

Closes #31253.

### Checklist

----


packages/aws-cdk-lib/pipelines/lib/codepipeline/codepipeline.ts
```diff
@@ -388,7 +387,7 @@ export class CodePipeline extends PipelineBase {
     this.selfMutationEnabled = props.selfMutation ?? true;
     this.dockerCredentials = props.dockerCredentials ?? [];
     this.singlePublisherPerAssetType = !(props.publishAssetsInParallel ?? true);
-    this.cliVersion = props.cliVersion ?? preferredCliVersion();
+    this.cliVersion = props.cliVersion ?? 'latest';
```
rix0rrr, line 390: If you are sure it is safe to unlock all CLI versions (both `aws-cdk` and `cdk-assets`), then remove the `this.cliVersion` property entirely.

Otherwise, keep the versioning on `aws-cdk` and remove its use only for `cdk-assets`.

rix0rrr (changes requested):

rix0rrr (approved):

PR #31262, opened 2024-08-30
add README section for sse-kms circular dependency workaround

gracelu0:
- add README section with steps for user to follow to use escape hatch to scope down the key policy


packages/aws-cdk-lib/aws-cloudfront-origins/lib/s3-bucket-origin.ts
```diff
@@ -156,7 +156,7 @@ export abstract class S3BucketOrigin extends cloudfront.OriginBase {
           },
         );
         Annotations.of(key.node.scope!).addWarningV2('@aws-cdk/aws-cloudfront-origins:wildcardKeyPolicyForOac',
-          'To avoid circular dependency between the KMS key, Bucket, and Distribution,' +
+          'To avoid circular dependency between the KMS key, Bucket, and Distribution, ' +
```
samson-keung, line 159: Thank you!

samson-keung (approved):

PR #31268, opened 2024-08-30
feat(location): support Tracker and TrackerConsumer

### LINKED ISSUE #30712: aws-location-alpha: support L2 Tracker and Tracker Consumer Construct
### Describe the feature

In [aws-location-alpha](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-location-alpha-readme.html), [tracker](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_location.CfnTracker.html) and [tracker consumer](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_location.CfnTrackerConsumer.html) have not been supported yet.

### Use Case

To use tracker and tracker consumer.

### Proposed Solution

Add `Tracker` and `TrackerConsumer` class.

I intend to implement this after #30710 is closed because `TrackerConsumer` will use `GeofenceCollection`.

### Other Information

_No response_

### Acknowledgements

### CDK version used

2.147.0

all

mazyu36:
### Issue # (if applicable)

Closes #30712.

### Reason for this change
To support tracker and tracker consumer.

### Description of changes
Add `Tracker` and `TrackerConsumer` class.

### Checklist

----


packages/@aws-cdk/aws-location-alpha/lib/tracker.ts
```diff
@@ -0,0 +1,267 @@
...
+      trackerName: this.physicalName,
+      description: props.description,
+      eventBridgeEnabled: props.eventBridgeEnabled,
+      kmsKeyEnableGeospatialQueries: props.kmsKeyEnableGeospatialQueries,
+      kmsKeyId: props.kmsKey?.keyArn,
+      positionFiltering: props.positionFiltering,
+    });
+
+    props.geofenceCollections?.forEach((collection) => {
+      new CfnTrackerConsumer(this, `TrackerConsumer${collection.node.id}`, {
```
mazyu36, line 205: Initially, I had TrackerConsumer as a separate Construct. 
https://github.com/aws/aws-cdk/pull/31268/commits/5fb3cdd5c13b6efda280802bfb316275c407caa3#diff-fc1da6063a6db97ebec475178eaaa40ce1a7d0a56f8c1b5cb249b99f33e9946a

However, since it only involved associations, I decided to integrate it into the Tracker. 
If you have any opinions on this design, please let me know.

packages/@aws-cdk/aws-location-alpha/test/integ.tracker.ts
```diff
@@ -0,0 +1,37 @@
...
+      geofenceCollections: [geofenceCollection1, geofenceCollection2],
+    });
+
+    tracker.addGeofenceCollections(geofenceCollectionForAdd1, geofenceCollectionForAdd2);
+  }
+}
+
+const app = new App();
+
+new integ.IntegTest(app, 'TrackerTest', {
```
sumupitchayan, line 35: Thanks for the PR @mazyu36  - I'm wondering if it's possible to add some integ test assertions here to validate that the construct works beyond deployment

mazyu36: Thank you. I added an assertion to ensure that the tracker consumers are correctly created. 
How does this look?

packages/@aws-cdk/aws-location-alpha/lib/tracker.ts
sumupitchayan: > However, since it only involved associations, I decided to integrate it into the Tracker.
If you have any opinions on this design, please let me know.

@mazyu36 what do you mean by it only involved associations? Just looking to clarify my understanding. Here it seems CDK users would not be able to create TrackerConsumer objects via an L2 - does this make sense?

mazyu36: @sumupitchayan 
TrackerConsumer is just an association between Tracker and Consumer (GeofenceCollection) and has no actual entity.

I thought the association could be auto-generated, similar to how VPC Subnet and route table associations work.
https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-ec2/lib/vpc.ts#L2117

Alternatively, TrackerConsumer could be made into a separate L2 construct.

Which approach would be better do you think?

sumupitchayan: @mazyu36 thanks for the clarification, this design makes sense to me. From my understanding, customers won't directly be creating TrackerConsumer objects and this case is similar to how VPC Subnets/route table associations are designed.

Additionally, since this is an alpha module and it will be easy to add an L2 for TrackerConsumer for this design, I think we can just do that in the future if we find it is needed for a customer use case.

packages/@aws-cdk/aws-location-alpha/test/integ.tracker.ts
sumupitchayan: @mazyu36 where is the assertion? I don't see it in this file

mazyu36: @sumupitchayan 
The following line. Please let me know if this is insufficient.

https://github.com/aws/aws-cdk/blob/a0d0cd39d68377e818256d5723d55bb75973cca8/packages/%40aws-cdk/aws-location-alpha/test/integ.tracker.ts#L52

sumupitchayan: Ah yes sorry I missed this for some reason - thanks for adding this! Approved 👍

sumupitchayan (approved):

PR #31270, opened 2024-08-30
OAC encrypted bucket origin integ test

samson-keung:
Integration test for usecase where Distribution is linked to a encrypted bucket as origin.


gracelu0 (approved):
LGTM, thank you!

PR #31275, opened 2024-08-31
chore: remove uses of aws-sdk outside cli

TheRealAmazonKendra:
Every use of sdkv2, besides what is used inside the cli has been removed as of this change.

### Checklist

----


kaizencc (commented):
will customers who currently use the old generated metadata thing be broken when they upgrade and suddenly be using the v3 version?

TheRealAmazonKendra:
> will customers who currently use the old generated metadata thing be broken when they upgrade and suddenly be using the v3 version?

No, it's actually pretty much the same data being generated (v3 is more up-to-date), we were just parsing it in slightly different ways in different locations, with the slight differences not actually resulting in different output.

rix0rrr (approved):

PR #31280, opened 2024-09-01
feat(rds): enable `grantDataApiAccess` method for imported database cluster

### LINKED ISSUE #31116: aws-rds: Imported Database Cluster cannot grant data API access
### Describe the bug

I am having the same issue as #30676 that was previously closed.

I am looking up a RDS Database Cluster to use as an AppSync datasource. The RDS Cluster does have Data API enabled and I can add it as a datasource in the AppSync console, however when I add the RDS datasource via cdk, I get the error:
`RuntimeError: Error: Cannot grant Data API access when the Data API is disabled`

### Regression Issue

_No response_

### Expected Behavior

I should be able add a RDS datasource to an AppSync API by using the `fromDatabaseClusterAttributes` function.

### Current Behavior

Error is thrown:
`RuntimeError: Error: Cannot grant Data API access when the Data API is disabled`

### Reproduction Steps

``` python

# look up cluster
cluster = aws_rds.DatabaseCluster.from_database_cluster_attributes(
            self,
            'cluster',
            cluster_identifier='my-cluster',
        )

# add cluster as AppSync datasource
graph_api.add_rds_datasource(
            'rds-datasource',
            serverless_cluster=cluster,
            secret_store=my_secret,
        )

```

### Possible Solution

_No response_

#30676 explains this issue very well and what is causing it.

### CDK CLI Version

2.151.0 (build b8289e2

### Framework Version

_No response_

### Node.js Version

v18.17.1

### OS

MacOS

### Language

Python

### Language Version

Python 3.9

### Other information

_No response_

go-to-k:
### Issue # (if applicable)

Closes #31116 .

### Reason for this change

It was not possible to call the `grantDataApiAccess` method for an imported database cluster.

Because the imported database cluster always has the `enableDataApi` with false.

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-rds/lib/cluster.ts#L983

```ts
class ImportedDatabaseCluster extends DatabaseClusterBase implements IDatabaseCluster {
  // ...
  protected readonly enableDataApi = false;
```

But the `grantDataApiAccess` throws an error when the `enableDataApi` is set to false.

https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-rds/lib/cluster.ts#L523-L526

```ts
  public grantDataApiAccess(grantee: iam.IGrantable): iam.Grant {
    if (this.enableDataApi === false) {
      throw new Error('Cannot grant Data API access when the Data API is disabled');
    }
    // ...
}
```

### Description of changes

Added the property `dataApiEnabled` to attributes for the imported database cluster.

If the `dataApiEnabled` to the attributes is set to true, the `enableDataApi` in the imported cluster will be set to true and the `grantDataApiAccess` can be called.

Unit and integ tests.

### Checklist

----


packages/@aws-cdk-testing/framework-integ/test/aws-rds/test/integ.cluster-data-api-to-imported-cluster.ts
```diff
@@ -0,0 +1,37 @@
...
+  dataApiEnabled: true,
+});
+
+importedCluster.grantDataApiAccess(func);
+
+new integ.IntegTest(app, 'integ-cluster-data-api-to-imported-cluster', {
+  testCases: [stack],
+});
+
+app.synth();
```
badmintoncryer, line 37: It's likely that this line is unnecessary.

```suggestion
```

badmintoncryer (approved):
Thank you for the PR! I've added only a nit comment.

I hardcoded `enableDataApi` as `false` for the imported cluster, so I really appreciate the fix!

go-to-k:
Thanks, changed!

packages/@aws-cdk-testing/framework-integ/test/aws-rds/test/integ.cluster-data-api-to-imported-cluster.ts
```diff
@@ -0,0 +1,35 @@
...
+  runtime: lambda.Runtime.NODEJS_18_X,
+  handler: 'index.handler',
+  code: lambda.Code.fromInline('exports.handler = async (event) => { return "hello"; }'),
+});
+
+const cluster = new rds.DatabaseCluster(stack, 'Database', {
+  engine: rds.DatabaseClusterEngine.auroraPostgres({ version: rds.AuroraPostgresEngineVersion.VER_16_1 }),
+  writer: rds.ClusterInstance.serverlessV2('writerInstance'),
+  vpc,
+  enableDataApi: true,
```
GavinZZ, line 22: I wonder what would happen if this is spcified as `false`, but user set it to true when setting `enableDataApi` to `true` on imported interface.

go-to-k: Regardless of the actual configuration, the target (in this case Lambda) will be granted the permission if it is set to true on the imported interface. This is true for all attribute values that are set without a lookup (context) method, regardless of the DataAPI. In this case, `secret`, etc.

Another way is to remove that property (`dataApiEnabled`) from the imported interface and make the `enableDataApi` variable undefined in `ImportedDatabaseCluster` class as follows.

```diff
-    this.enableDataApi = attrs.dataApiEnabled ?? false;
```

https://github.com/go-to-k/aws-cdk/blob/a0d9a65d9c699f5c7e8f53ce49de597ce5352d36/packages/aws-cdk-lib/aws-rds/lib/cluster.ts#L999

Then the following branches can be avoided.

```ts
  public grantDataApiAccess(grantee: iam.IGrantable): iam.Grant {
    if (this.enableDataApi === false) {
      throw new Error('Cannot grant Data API access when the Data API is disabled');
    }
```

This would behave in the same way as above. However, it is a little simpler with fewer properties in the imported interface, but it is a little safer with `dataApiEnabled` on the interface.

GavinZZ (approved):
LGTM, thanks for the quick response to my question.

Issue #26754, opened 2023-08-14
aws_route53: Allow updates to existing records in a single operation


christow-aws:
### Describe the feature

Currently, in order to update a Route53 record's routing policy using CDK (which uses an underlying lambda to process events), we have to do multiple passes at a record's fields over the course of multiple deployments. This operation is supported in the console with 1 click. For example, we have to do the following:

(initial code - uses simple routing policy)
```
const aRecord = new ARecord(this, 'RecordSet', {
    target: RecordTarget.fromAlias(new ApiGatewayDomain(domain)),
    zone: hostedZone,
})
```

(first pass - changing routing policy to weighted)
```
// Delete this for round 2 of deployments
const aRecord = new ARecord(this, 'RecordSet', {
    target: RecordTarget.fromAlias(new ApiGatewayDomain(domain)),
    zone: hostedZone,
})

const aRecord2 = new ARecord(this, 'RecordSet2', {
    target: RecordTarget.fromAlias(new ApiGatewayDomain(domain)),
    zone: hostedZone,
    deleteExisting: true, // Delete this for round 2 of deployments
})
const recordSet = aRecord2.node.defaultChild as CfnRecordSet
recordSet.weight = 100
recordSet.setIdentifier = 'API-Gateway-A-Record-weighted'
aRecord2.node.addDependency(aRecord)
```

(second pass - cleanup)
```
const aRecord2 = new ARecord(this, 'RecordSet2', {
    target: RecordTarget.fromAlias(new ApiGatewayDomain(domain)),
    zone: hostedZone,
})
const recordSet = aRecord2.node.defaultChild as CfnRecordSet
recordSet.weight = 100
recordSet.setIdentifier = 'API-Gateway-A-Record-weighted'
```

### Use Case

Modifying record routing policies.

### Proposed Solution

Involves updating the custom resource lambda, of which I haven't poked around in.

### Other Information

We're also required to use the `deleteExisting` field, which theoretically isn't necessary (but is super dangerous). R53 appears to allow the entire operation to be done at once (rm old, insert new) with validation at API-call-time that the new, to-be-inserted is valid _before_ deleting the old one. This means that there shouldn't be a case where the existing record is deleted and the new one cannot be inserted.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2, TypeScript

### Environment details (OS name and version, etc.)

AL2 on x86

christow-aws:
[...]

PR #31281, opened 2024-09-02
feat(route53): new Route53 CrossAccountRecordSet construct

### LINKED ISSUE #15213: [Route53] cdk-route53  Cross account records.
Extend Cross Acount Records to be more than Just Zone Delegations.     It would be great if out of the box, we could do any kind of record in a zone that was not in the account that the stack is in. 

### Use Case

many times, we want to be able to add records for external things... eg

customerportal.domain.com
www.domain.com

### Other

---

lukepafford:
### Issue #

Closes  #15213 

### Reason for this change

Deploying your APEX Route53 HostedZone (example.com) to a designated AWS account, and delegating child HostedZones (us-east-1.prod.example.com, us-east-2.prod.example.com) to other AWS accounts is a common setup. Managing this setup through CDK becomes impossible once a child account contains RecordSet information that needs to be created in the APEX zone. This scenario is most common when you want to create a RecordSet at the root of your domain that contains a non-simple [routing policy](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy.html), however the information needed to create those RecordSets is distributed between multiple child CloudFormation stacks. 

This `CrossAccountRecordSet` construct bridges this gap by creating the RecordSet in the APEX HostedZone, and it only needs the following props:
  * A `crossAccountRole: iam.IRole` that the custom resource will assume which authorizes the Route53 changes in the APEX HostedZone.
  * The same [RecordSetProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_route53.RecordSetProps.html) that exists for a [RecordSet](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_route53.RecordSet.html).

This issue is quite popular, as #15213 has 36 reactions, and there are already a couple constructs addressing this on ConstructHub:
* [cross-account-record-set](https://constructs.dev/packages/cdk-cross-account-route53/v/1.0.1?lang=typescript)
* [@fallobst22/cross-account-record-set](https://constructs.dev/packages/@fallobst22/cdk-cross-account-route53/v/0.0.35?lang=typescript)

This change was created because I believe this functionality should exist as a core part of the CDK. As for using this implementation over the ConstructHub implementations, I think the implementations are too simple for how many rough edges the Route53 ChangeResourceRecordSets API has, which is explained further in the change. I also think the `CrossAccountRecordSet` construct interface is intuitive, and easy for existing users to adopt.

---
This change doesn't solve all CrossAccount DNS related problems. A specific pain for me is that the [CertificateValidation](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_certificatemanager.CertificateValidation.html) doesn't support cross account updates (see #8934), which diminishes the impact of this change. Still, this change is already quite large and those problems can be addressed in later changes. I first wanted to confirm this change moves in a direction the CDK team finds acceptable considering the [DnsValidatedCertificate deprecation](https://github.com/aws/aws-cdk/pull/21982) has so much back and forth dialogue around it.

Side note: The [CloudFormation AWS::CertificateManager::Certificate](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-certificatemanager-certificate.html) should make the CNAME RecordName and RecordValue a Return Value accessible through [Fn::GetAtt](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html). You could then pass those values to this `CrossAccountRecordSet` construct to enable cross account DNS validation for your ACM certificate!
...One problem at a time.

### Key Design Decisions
* I spent a significant amount of time debating back and forth between the construct accepting `RecordSetProps` versus accepting an array of `RecordSetProps[]` which would fully address the concerns of #26754. The `aws-route53/README.md` update defends why this decision was made.
* The `@aws-cdk/custom-resource-handlers/lib/aws-route53/cross-account-record-set-handler/index.ts` custom resource handler may appear to be overly complicated because the implementation performs DELETE/CREATE ChangeBatches, rather than an UPSERT. I faced so much pain when testing various RecordSet changes relying on only Route53 UPSERT, and I firmly believe this will result in a better user experience for the price of a more complicated implementation.
* The `@aws-cdk/custom-resource-handlers/lib/aws-route53/cross-account-record-set-handler/index.ts` returns its own PhysicalResourceId, which requires sending a [Custom Resource Response Object](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html). The `PhysicalResourceId` class explains why this was necessary. Route53 uniquely identifies a RecordSet in a non obvious way, and this implementation 🪄 just does the right thing 🪄 .
* Up until the very end, the `CrossAccountRecordSet` construct supported an `evaluateTargetHealth` prop, but I decided to remove it because the [RecordSetProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_route53.RecordSetProps.html) doesn't support that prop, and I only wanted this change to match the features for what already exists (See #30664). There are a few parameters to the [ChangeResourceRecordSet API](https://docs.aws.amazon.com/Route53/latest/APIReference/API_ChangeResourceRecordSets.html) that aren't available in the Route53 CDK package, and should probably be addressed in a separate change. I'm most interested in figuring out a sane way to configure health checks, but looks like this is an open issue that is being worked on (#9481 #30739).

* By the time I was mostly done with the implementation, I reread the [NEW_CONSTRUCTS_GUIDE](https://github.com/aws/aws-cdk/blob/main/docs/NEW_CONSTRUCTS_GUIDE.md) and realized this construct should probably go into an alpha package. If reviewers want this to be made an alpha package, I'll do the work of refactoring the code, but I don't want to spend more time on this unless I know the changes would be merged.

### Description of changes

* Creates a new `CrossAccountRecordSet` construct that creates Route53 RecordSet resources in a HostedZone that exists in another account.
* Minor refactors to utility files for code reuse:
  * `route53Region` from `@aws-cdk/custom-resource-handlers/lib/aws-route53/cross-account-zone-delegation-handler/index.ts` -> `@aws-cdk/custom-resource-handlers/lib/aws-route53/sts-util.ts`.
  * `validateRecordSetProps` from `aws-cdk-lib/aws-route53/lib/record-set.ts` -> `aws-cdk-lib/aws-route53/lib/util.ts`.
* `submitResponse` was made a public export from `packages/@aws-cdk/custom-resource-handlers/lib/core/nodejs-entrypoint-handler/index.ts` This seemed like a safe change to me, and it would have been a waste to rewrite this functionality in the custom handler.

* Unit Tests (100% cov)
* Integ Tests

### Personal Issues

* This may be an issue on my part, but I couldn't find any good documentation on creating and validating `.lit.ts` files. I basically had to copy/paste `packages/aws-cdk-lib/aws-route53/test/integ.cross-account-record-set-simple-example.lit.ts` into the integ test at `packages/@aws-cdk-testing/framework-integ/test/aws-route53/test/integ.cross-account-record-set-simple-example.lit.ts`. One file purely exists for README documentation, and the other exists as an IntegTest. From what I can tell, past changes are basically duplicating these files as well. Hopefully I'm missing something.

* In order to get [Istanbul](https://github.com/gotwarlost/istanbul) to report my code coverage correctly, I had to make the following tsconfig.json changes:
```console
diff --git a/packages/@aws-cdk/custom-resource-handlers/tsconfig.json b/packages/@aws-cdk/custom-resource-handlers/tsconfig.json
index 312bd74a61..eb97e8ad3d 100644
--- a/packages/@aws-cdk/custom-resource-handlers/tsconfig.json
+++ b/packages/@aws-cdk/custom-resource-handlers/tsconfig.json
@@ -6,8 +6,7 @@
         "strict": true,
         "alwaysStrict": true,
         "declaration": true,
-        "sourceMap": false,
-        "inlineSourceMap": false,
+        "inlineSourceMap": true,
         "inlineSources": false,
         "noUnusedLocals": true,
         "noUnusedParameters": true,
```
* I spent half my time fighting the build system. `npx lerna run build` never compiled the files like I thought it would, and `yarn build` was inconsistent.  I would frequently deploy changes and be confused as to why it didn't work, only to realize the build commands didn't actually work. I finally realized I should use `yarn watch` as that was mostly consistent for compiling my changes on file save, but even that had to be frequently killed and restarted. I'm not sure if this is a problem with my setup, but I basically learned to not trust that my files were actually built, and always inspect the compiled .js files to confirm they contained the changes that I expected.

### Checklist
- [X ] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----


packages/@aws-cdk/custom-resource-handlers/lib/core/nodejs-entrypoint-handler/index.ts
```diff
@@ -95,7 +96,7 @@ function renderResponse(
   };
 }
 
-async function submitResponse(status: 'SUCCESS' | 'FAILED', event: Response) {
+export async function submitResponse(status: 'SUCCESS' | 'FAILED', event: Response) {
```
lukepafford, line 99: Make `submitResponse` public so we can reuse it in the CrossAccountRecordSet custom resource handler.

lukepafford:
A critical piece of information has to be missing from [INTEGRATION_TESTS.md](https://github.com/aws/aws-cdk/blob/main/INTEGRATION_TESTS.md#running-integration-tests); I have now spent twice as long trying to successfully update snapshots as I did as making the change.

Integ tests can run with 3 different environment variables:
  1. The [default values](https://github.com/aws/aws-cdk/blob/775c58c9b73b877500b739fb1a1e7c21216a834c/packages/%40aws-cdk/integ-runner/lib/runner/runner-base.ts#L425-L432) set by the test runner itself.
  1. Environment variables you set when you run the integ tests.
  1. Default values set by the integ test when no environment variables are found.
  
### Catch 22 Integ Tests
  When the CodeBuild project checks all the snapshots, it won't run with your environment variable values. It will either use  the default values set in the integ runner, or the default values set in the integ test.

These are dummy values that you can't use when you rerun the snapshot tests. If you did try and use them, the tests would surely fail. So you fundamentally have to regenerate the snapshots with valid environment variables that the CodeBuild project will never have access to. Your snapshot, and the CodeBuild snapshot will never align.

So how did past snapshots ever get generated? If you inspect a lot of the existing tests, you can see that the snapshot uses the AWS account `12345678`, indicating someone committed a snapshot using a fake account ID. How did they do this? Was there some script that retroactively found/replaced real environment variables with the dummy values? I've now ran `git blame` on so many past integration tests at this point just to get the tiniest bit of insight into how these past snapshots were generated, but no wisdom seems to have been left behind.

### What's going wrong?
I'm torn between two conclusions which dramatically impacts my mental decision tree going forward. One of the two scenarios is true:
1. I'm doing something stupid, and creating snapshots works fine
    * Past committers literally just run `yarn integ --update-on-failed integ/test.js`. Their snapshots are somehow correctly generated with fake environment values.
    * When I do the same, my real env values are saved into the snapshot.
2. Past committers do something unique in their environment to get their snapshots to align with what the CodeBuild project also expects.

Common sense tells me scenario 2 can't be true, because that would mean there has been lots of past commits where people have been successfully following a magical undocumented process. However I'm quickly running out of other conclusions to reach, since I have tried everything under the sun. I would love to be missing something, but I've thought about this from every possible angle and I'm starting to lose hope.

lukepafford:
I thought I almost found the secret sauce. Perhaps if the explicit `env` declaration was removed from the stack, and the profile was passed to the CDK command, the generated snapshot wouldn't contain real env values:

```console
@@ -158,9 +156,12 @@ const crossAccountRoleStack = new CrossAccountRoleStack(app, 'CrossAccountRoleSt
   authorizedAccounts: [appAccountId],
   changeResourceRecordSetsRecordTypes: ['A'],
   changeResourceRecordSetsNormalizedRecordNames: [`test.${hostedZoneName}`],
-  env: {
-    account: apexAccountId,
-    region: 'us-east-1',
+  cdkCommandOptions: {
+    deploy: {
+      args: {
+        profile: 'cdk_integ_account',
+      },
+    },
   },
 });

@@ -170,14 +171,17 @@ const appStack = new AppStack(app, 'AppStack', {
   apexAccountId: crossAccountRoleStack.account,
   crossAccountRoleName: crossAccountRoleName,
   recordName: 'test',
-  env: {
-    account: appAccountId,
-    region: 'us-west-1',
+  cdkCommandOptions: {
+    deploy: {
+      args: {
+        profile: 'cdk_integ_cross_account',
+      },
+    },
   },
 });
```

This unfortunately did not work, and the generated snapshot still contained real values.

PR #30711, opened 2024-06-30
feat(location): support GeofenceCollection

mazyu36:
### Issue # (if applicable)

Closes #30710.

### Reason for this change
To support L2 level geofence collection.

### Description of changes
Add `Geofence Collection` class.

### Description of how you validated changes
Add unit tests and integ tests.

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #30682, opened 2024-06-26
feat(location): support RouteCalculator

mazyu36:
### Issue # (if applicable)

Closes #30681 .

### Reason for this change
In [aws-location-alpha](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-location-alpha-readme.html), [route calculator](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_location.CfnRouteCalculator.html) has not been supported yet.

### Description of changes
Add `RouteCalculator` class.

### Description of how you validated changes
Add unit tests and integ tests.

### Checklist
- [x] My code adheres to the [CONTRIBUTING GUIDE](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md) and [DESIGN GUIDELINES](https://github.com/aws/aws-cdk/blob/main/docs/DESIGN_GUIDELINES.md)

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #31287, opened 2024-09-02
fix(location): remove base class from PlaceIndex class

mazyu36:
### Issue # (if applicable)

None

### Reason for this change
I implemented the location resources in #30711 and #30682. 

During the review, it was suggested that it would be better not to have a base class.
As a result, the `GeofenceCollection` class and `RouteCalculator` class do not have a base class. 

https://github.com/aws/aws-cdk/pull/30711#discussion_r1734937879

On the other hand, the `PlaceIndex `class has already been implemented with a base class, which is inconsistent with the current approach.

### Description of changes
Remove a unnecessary base class from `PlaceIndex` class.

Re-run unit tests and integ tests.

### Checklist

----


packages/@aws-cdk/aws-location-alpha/lib/place-index.ts
```diff
@@ -74,39 +74,12 @@ export enum IntendedUse {
   STORAGE = 'Storage',
 }
 
-abstract class PlaceIndexBase extends Resource implements IPlaceIndex {
```
sumupitchayan, line 77: @mazyu36 - just to double clarify here, you don't forsee adding an additional class anywhere that would need to extend the base class?

mazyu36: @sumupitchayan 

I don't think we'll need to add other classes, since there are no resources similar to placeindex.

sumupitchayan (approved):

sumupitchayan (approved):

PR #31305, opened 2024-09-03
fix(stepfunctions-tasks): fix bedrock input/output path in step-funct…

### LINKED ISSUE #31302: ❗ NOTICE: stepfunctions-tasks: BedrockInvokeModel outputPath no longer passes output to next state
Please add your +1 👍 to let us know you have encountered this
---

### Overview:
Previously, the BedrockInvokeModel `outputPath` parameter could be used to select a portion of the model output to pass to the next state. This behavior aligns with how the parameter is documented "JSONPath expression to select select a portion of the state output to pass to the next state."
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions_tasks.BedrockInvokeModel.html#outputpath

Now, the outputPath parameter is assumed to be an S3Uri where the model output should be saved. The parameter can no longer be used to select model output and pass it to the next state.

This behavior appears to have changed in https://github.com/aws/aws-cdk/commit/f5dd73bd19f546335531f587f6d7fbde8aff96ce

### Solution:
Revert existing code for encoding the path under S3Uri, and possibly introduce another variable in BedrockTaskProps instead of extending the existing `TaskStateBaseProps`.

### Regression Issue

- [X] Select this option if this issue appears to be a regression.

2.150.0

### Expected Behavior

In a BedrockInvokeModel task, set outputPath:
```python
# Extract the response from the model
output_path="$.Body.content[0].text",
```

The Step Functions workflow definition is synthesized as:
```
"OutputPath": "$.Body.content[0].text",
```

When I execute the deployed Step Functions state machine, the model text output is passed as the output of the task.
```
"Here is a one sentence summary of ...."
```

### Current Behavior

The outputPath parameter is synthesized as:
```
"Output": {
   "S3Uri": "$.Body.content[0].text"
}
```

When I execute the deployed Step Functions state machine, I get a validation exception:
```
The value for the S3Uri field is not recognized as a valid URI: $.Body.content[0].text
```

### Reproduction Steps

This is my stack:
https://github.com/aws-samples/amazon-bedrock-serverless-prompt-chaining/blob/main/techniques/stacks/model_invocation.py

### Possible Solution

I suggest reverting the change above, so that outputPath works as documented and expected. The `input` and `output` parameters were intended to distinguish between state that should be passed between states (inputPath/outputPath) vs destinations where input and output should be sourced from (input/output).

_No response_

### CDK CLI Version

2.155.0

### Framework Version

_No response_

### Node.js Version

v20.17.0

### OS

Linux

### Language

Python

### Language Version

Python 3.12.3

### Other information

_No response_

shikha372:
### Issue # (if applicable)

Closes #31302.

### Reason for this change

PR#30298 introduced a breaking change in the stepfunctions-tasks which was encoding the input/output path from the `TaskStateBaseProps` to be used to input a S3 URI, which is not always the case and customer might be using the JSON string as it is.

### Description of changes

To keep the functionality from the original ask of issue #29229 ,added another props to specify S3 URI as an input to the stepfunctions-tasks.

Introduced a feature flag to keep the existing behaviour for the customers and not introduce a breaking change.

Updated integ test and deployed in account

Integ Test Results:
`aws stepfunctions start-execution --state-machine-arn <deployed state machine arn> `: should return execution arn

A5-zEnXPmmPWTJx
{
    "executionArn": "arn:aws:states:us-east-1:XXXX:execution:StateMachine2E01A3A5-zEnXPmmPWTJx:516f2e60-9507-46cb-95e6-4d9453429b08",
    "startDate": "2024-09-04T15:43:33.200000-07:00"
}

`aws stepfunctions describe-execution --execution-arn <exection-arn generated before> `: should return status as SUCCEEDED

### Checklist

----


packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -108,6 +108,24 @@ export interface BedrockInvokeModelProps extends sfn.TaskStateBaseProps {
...
   readonly output?: BedrockInvokeModelOutputProps;
 
+  /**
+   * The destination location where the API response is written.
+   *
+   * This field can be used to specify s3 URI in the form of token
+   *
+   * @default - The API response body is returned in the result.
+   */
+  readonly s3inputPath?: string;
```
godwingrs22, line 118: nit: will it be better to rename the prop to `inputS3Uri`?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -108,6 +108,24 @@ export interface BedrockInvokeModelProps extends sfn.TaskStateBaseProps {
...
+
+  /**
+   * The source location where the API response is written.
+   *
+   * This field can be used to specify s3 URI in the form of token
+   *
+   * @default - The API response body is returned in the result.
+   */
+  readonly s3outputPath?: string;
```
godwingrs22, line 127: Similarly as `outputS3Uri`

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -271,10 +289,10 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
         Body: this.props.body?.value,
         Input: this.props.input?.s3Location ? {
           S3Uri: `s3://${this.props.input.s3Location.bucketName}/${this.props.input.s3Location.objectKey}`,
-        } : this.props.inputPath ? { S3Uri: this.props.inputPath } : undefined,
+        } : this.props.s3inputPath ? { S3Uri: this.props.s3inputPath } : undefined,
```
godwingrs22, line 292: Can we also have some unit test case?

godwingrs22 (commented):
Overall looks good. Just left few minor comments

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -108,6 +108,24 @@ export interface BedrockInvokeModelProps extends sfn.TaskStateBaseProps {
...
   readonly output?: BedrockInvokeModelOutputProps;
 
+  /**
+   * The destination location where the API response is written.
+   *
+   * This field can be used to specify s3 URI in the form of token
+   *
+   * @default - The API response body is returned in the result.
+   */
+  readonly s3InputUri?: string;
```
godwingrs22, line 118: Can we also update the [Readme](https://github.com/aws/aws-cdk/blob/f3bc16c7b9f1d58318a451d8640a0f8f07f4ffcb/packages/aws-cdk-lib/aws-stepfunctions-tasks/README.md?plain=1#L402-L422)

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -108,6 +108,24 @@ export interface BedrockInvokeModelProps extends sfn.TaskStateBaseProps {
    */
   readonly output?: BedrockInvokeModelOutputProps;
```
clareliguori, line 109: Should the addition of the s3InputUri and s3OutputUri params deprecate the input/output params? These params appear to result in the same functionality.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -108,6 +108,24 @@ export interface BedrockInvokeModelProps extends sfn.TaskStateBaseProps {
...
 
+  /**
+   * The destination location where the API response is written.
```
clareliguori, line 112: s3InputUri should be the source location to retrieve the input data from.

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/bedrock/invoke-model.test.ts
```diff
@@ -234,8 +234,6 @@ describe('Invoke Model', () => {
         ],
       },
       End: true,
-      InputPath: '$.prompt',
-      OutputPath: '$.prompt',
```
clareliguori, line 238: Are there other unit tests that test this functionality?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
shikha372: changes are keeping in mind the original request where user is not able to specify the s3URI in below format, for which they will need something like these new props `s3InputURI`, let me know if you think otherwise

 Input: this.props.input?.s3Location ? { 
   S3Uri: `$prompt.bucket`, 
 } : undefined, 
 Output: this.props.output?.s3Location ? { 
   S3Uri: `$prompt.key`, 
 } : undefined,

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/bedrock/integ.invoke-model.ts
```diff
@@ -52,13 +52,32 @@ const prompt2 = new BedrockInvokeModel(stack, 'Prompt2', {
...
+        'Alphabetize this list of first names:\n{}',
+        sfn.JsonPath.stringAt('$.names'),
+      ),
+      textGenerationConfig: {
+        maxTokenCount: 100,
+        temperature: 1,
+      },
+    },
+  ),
   outputPath: sfn.JsonPath.stringAt('$.names'),
```
godwingrs22, line 70: Can help me to understand why body is required instead of inputPath?

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/bedrock/invoke-model.test.ts
shikha372: There are two unit tests, one to check the expected template and one to check the permissions populated in policy

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/bedrock/integ.invoke-model.ts
shikha372: just added this integ test is to make sure, we are not breaking any existing use case for `outputPath` or `inputPath` from the base class, looking back at why it might have been marked as required could be a decision take while designing this s3 input parameter, i don't see it something in [API doc](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html)

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
clareliguori: Would it be better to add an S3Uri param to BedrockInvokeModelInputProps and BedrockInvokeModelOutputProps.  The input/output params were intended to map directly to the Input and Output fields in Step Functions' Bedrock service integration for InvokeModel:
https://docs.aws.amazon.com/step-functions/latest/dg/connect-bedrock.html#connect-bedrock-custom-apis

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/bedrock/invoke-model.test.ts
clareliguori: The title of this unit test is "invoke model allows input and output json path". I think this test was originally meant to test the original behavior of inputPath and outputPath. So we need an additional new test for these new s3InputUri and s3OutputUri params, AND fix this unit test to go back to testing the original behavior of inputPath and outputPath.

clareliguori: Update: I see in the history now that you had actually added this test in the last change. I do still think we should have both tests (inputPath and s3InputUri)

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/bedrock/integ.invoke-model.ts
clareliguori: In order to test the inputPath field, you can do a Pass state to transform the previous output into the format expected for the InvokeModel body.

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/bedrock/integ.invoke-model.ts
```diff
@@ -52,13 +52,32 @@ const prompt2 = new BedrockInvokeModel(stack, 'Prompt2', {
...
 });
 
-const chain = sfn.Chain.start(prompt1).next(prompt2).next(prompt3);
+/** Test for Bedrock s3 URI Path */
+const prompt4 = new BedrockInvokeModel(stack, 'Prompt4', {
+  model,
+  s3InputUri: sfn.JsonPath.stringAt('$.names'),
+  s3OutputUri: sfn.JsonPath.stringAt('$.names'),
+});
+
```
clareliguori, line 79: Can you run the verification steps listed at the top of this class?
`Stack verification steps`

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
shikha372: yeah, that's a good point, i've modified the PR to add these properties under bedrockInvokeModelProps

packages/aws-cdk-lib/aws-stepfunctions-tasks/test/bedrock/invoke-model.test.ts
shikha372: Thanks @clareliguori , this leads to one question can `body` be skipped in case of an `inputPath` ? since the earlier validation seems to added only for `body` and `input` and didn't point to a definition where we might be using these base props.

packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
```diff
@@ -72,6 +72,7 @@ Flags come in three types:
 | [@aws-cdk/aws-ecs:removeDefaultDeploymentAlarm](#aws-cdkaws-ecsremovedefaultdeploymentalarm) | When enabled, remove default deployment alarm settings | 2.143.0 | (default) |
 | [@aws-cdk/custom-resources:logApiResponseDataPropertyTrueDefault](#aws-cdkcustom-resourceslogapiresponsedatapropertytruedefault) | When enabled, the custom resource used for `AwsCustomResource` will configure the `logApiResponseData` property as true by default | 2.145.0 | (fix) |
 | [@aws-cdk/aws-s3:keepNotificationInImportedBucket](#aws-cdkaws-s3keepnotificationinimportedbucket) | When enabled, Adding notifications to a bucket in the current stack will not remove notification from imported stack. | 2.155.0 | (fix) |
+| [@aws-cdk:aws-stepfunctions-tasks:useNewS3UriParametersForBedrockInvokeModelTask](#aws-cdkaws-stepfunctions-tasksusenews3uriparametersforbedrockinvokemodeltask) | When enabled, use new props for S3 URI field in task definition of state machine for bedrock invoke model. | 2.155.1 | (fix) |
```
moelasmar, line 75: should we fix this version if we are planning to include this fix in the normal release

packages/aws-cdk-lib/cx-api/FEATURE_FLAGS.md
```diff
@@ -1356,4 +1358,21 @@ Other notifications that are not managed by this stack will be kept.
...
+of State Machine Task definition.
+
+When this feature flag is enabled, specify newly introduced props 's3InputUri' and 
+'s3OutputUri' to populate S3 uri under input and output fields in state machine task definition for Bedrock invoker model.
+
+
+| Since | Default | Recommended |
+| ----- | ----- | ----- |
+| (not in v1) |  |  |
+| 2.155.1 | `true` | `true` |
```
moelasmar, line 1375: here also, we need to fix the version

packages/aws-cdk-lib/cx-api/lib/features.ts
```diff
@@ -1107,6 +1108,22 @@ export const FLAGS: Record<string, FlagInfo> = {
...
+    summary: 'When enabled, use new props for S3 URI field in task definition of state machine for bedrock invoke model.',
+    detailsMd: `
+    Currently, 'inputPath' and 'outputPath' from the TaskStateBase Props is being used under BedrockInvokeModelProps to define S3URI under 'input' and 'output' fields
+    of State Machine Task definition.
+
+    When this feature flag is enabled, specify newly introduced props 's3InputUri' and 
+    's3OutputUri' to populate S3 uri under input and output fields in state machine task definition for Bedrock invoker model.  
+
+    `,
+    introducedIn: { v2: '2.155.1' },
```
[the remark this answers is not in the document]
moelasmar: should we set the default value for it to be true ?

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -145,9 +164,17 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
...
     const isBodySpecified = props.body !== undefined;
-    //Either specific props.input with bucket name and object key or input s3 path
-    const isInputSpecified = (props.input !== undefined && props.input.s3Location !== undefined) || (props.inputPath !== undefined);
+
+    let isInputSpecified: boolean;
+    if (!isFeatureFlagEnabled) {
+      isInputSpecified = (props.input !== undefined && props.input.s3Location !== undefined) || (props.inputPath !== undefined);
+    } else {
+      //Either specific props.input with bucket name and object key or input s3 path
+      isInputSpecified = props.input!==undefined ? props.input?.s3Location !== undefined || props.input?.s3InputUri !== undefined : false;
```
moelasmar, line 176: should we add a validation that either s3Location or s3InputUri can be used, and not both of them

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -162,18 +189,25 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
       throw new Error('Output S3 object version is not supported.');
     }
 
-    this.taskPolicies = this.renderPolicyStatements();
+    //Warning to let users know about the newly introduced props
+    if (props.inputPath || props.outputPath && !isFeatureFlagEnabled) {
+      Annotations.of(scope).addWarningV2('aws-cdk-lib/aws-stepfunctions-taks',
+        'These props will set the value of inputPath/outputPath as s3 URI under input/output field in state machine JSON definition. To modify the behaviour set feature flag `@aws-cdk/aws-stepfunctions-tasks:useNewS3UriParametersForBedrockInvokeModelTask": true` and use props s3InputUri/s3OutputUri');
```
moelasmar, line 195: should we change `s3InputUri/s3OutputUri` to be `input.s3InputUri/output.s3OutputUri` to make it clear

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -162,18 +189,25 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
...
+    }
+
+    this.taskPolicies = this.renderPolicyStatements(isFeatureFlagEnabled);
   }
 
-  private renderPolicyStatements(): iam.PolicyStatement[] {
+  private renderPolicyStatements(isFeatureFlagEnabled?: boolean): iam.PolicyStatement[] {
```
moelasmar, line 201: optional, I prefer to recalculate the `isFeatureFlagEnabled` here in this method instead of passing it, so juts in case in the future if we have another problem and we need to add a new feature flag, so we do not confuse different flags.

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -145,9 +164,17 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
 
     validatePatternSupported(this.integrationPattern, BedrockInvokeModel.SUPPORTED_INTEGRATION_PATTERNS);
 
+    const isFeatureFlagEnabled = FeatureFlags.of(this).isEnabled(cxapi.USE_NEW_S3URI_PARAMETERS_FOR_BEDROCK_INVOKE_MODEL_TASK);
```
moelasmar, line 167: could you rename this flag to be more related to this issue

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -285,5 +320,27 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
       }),
     };
   };
+
+  private getInputSource(props?: BedrockInvokeModelInputProps, inputPath?: string, isFeatureFlagEnabled?: boolean): string | undefined {
+    if (props?.s3Location) {
+      return `s3://${props.s3Location.bucketName}/${props.s3Location.objectKey}`;
+    } else if (isFeatureFlagEnabled && props?.s3InputUri) {
+      return props.s3InputUri;
```
moelasmar, line 328: can we make it simple if `isFeatureFlagEnabled` return `props.s3InputUri` so regardless it has undefined value, or any other value that can be translated as false like an empty string

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -285,5 +320,27 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
...
+    } else if (!isFeatureFlagEnabled && inputPath) {
+      return inputPath;
+    }
```
moelasmar, line 331: similar to the previous comment

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -262,19 +297,19 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
...
         ContentType: this.props.contentType,
         Body: this.props.body?.value,
-        Input: this.props.input?.s3Location ? {
-          S3Uri: `s3://${this.props.input.s3Location.bucketName}/${this.props.input.s3Location.objectKey}`,
-        } : this.props.inputPath ? { S3Uri: this.props.inputPath } : undefined,
-        Output: this.props.output?.s3Location ? {
-          S3Uri: `s3://${this.props.output.s3Location.bucketName}/${this.props.output.s3Location.objectKey}`,
-        } : this.props.outputPath ? { S3Uri: this.props.outputPath }: undefined,
+        Input: this.props.input || (this.props.inputPath && !isFeatureFlagEnabled) ? { S3Uri: inputSource } : undefined,
+        Output: this.props.output || ( this.props.outputPath && !isFeatureFlagEnabled) ? { S3Uri: outputSource } : undefined,
```
moelasmar, line 312: should you use the `inputSource` and `outputSource` you defined in lines 302, and 303

moelasmar, line 312: is it correct to make it like this:
```
Input: inputSource?? { S3Uri: inputSource } : undefined,
Output: outputSource?? { S3Uri: outputSource } : undefined,
```

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
shikha372: In that case if we return an empty string, it will populated in the input as { } for s3URI which could lead to failure with invalid URI provided for Input field.

moelasmar: so we need to validate the customer input that the empty `s3InputUri` is not a valid input

packages/aws-cdk-lib/aws-stepfunctions-tasks/lib/bedrock/invoke-model.ts
```diff
@@ -188,17 +188,24 @@ export class BedrockInvokeModel extends sfn.TaskStateBase {
     if (props.output?.s3Location?.objectVersion !== undefined) {
       throw new Error('Output S3 object version is not supported.');
     }
+    if (props.input?.s3InputUri && props.input.s3Location) {
+      throw new Error('Cannot specify both S3 InputUri and S3 location');
+    }
+    if (props.input?.s3InputUri === '') {
+      throw new Error('S3 InputUri cannot be an empty string');
+    }
```
moelasmar, line 196: should we have similar validations for output

packages/aws-cdk-lib/cx-api/lib/features.ts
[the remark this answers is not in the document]
shikha372: default value is `true`, version is updated to `V2Next` as the build was failing if we set it to any other major version than 2.155.0, this seems to be the Magic_Value which will get updated after the release

packages/aws-cdk-lib/aws-stepfunctions-tasks/README.md
```diff
@@ -398,11 +398,45 @@ const task = new tasks.BedrockInvokeModel(this, 'Prompt Model', {
...
+
+```
+### Using Input Path for S3 URI
+
+Provide S3 URI as an input or output path to invoke a model
+
+To specify the S3 URI as JSON path to your input or output fields, use props `s3InputUri` and `s3OutputUri` under BedrockInvokeModelProps and set 
+feature flag `@aws-cdk/aws-stepfunctions-tasks:useNewS3UriParametersForBedrockInvokeModelTask` to true.
+
+If this flag is not set, then the existing behaviour of populating the S3Uri from `InputPath` and `OutputPath` will take effect.
```
moelasmar, line 410: I recommend to remove the existing behaviour, and to add that it is not a recommended behaviour.

packages/aws-cdk-lib/aws-stepfunctions-tasks/README.md
```diff
@@ -398,11 +398,45 @@ const task = new tasks.BedrockInvokeModel(this, 'Prompt Model', {
...
+});
+
 ```
+
 ### Using Input Path
 
 Provide S3 URI as an input or output path to invoke a model
 
+Currently, input and output Path provided in the BedrockInvokeModelProps input is defined as S3URI field under task definition of state machine.
+To modify the existing behaviour, set `@aws-cdk/aws-stepfunctions-tasks:useNewS3UriParametersForBedrockInvokeModelTask` to true.
```
moelasmar, line 435: can you start that if the flag `@aws-cdk/aws-stepfunctions-tasks:useNewS3UriParametersForBedrockInvokeModelTask` is not true, the behaviour will be to ... and also mention that it is not recommended

packages/@aws-cdk-testing/framework-integ/test/aws-stepfunctions-tasks/test/bedrock/integ.invoke-model.ts
```diff
@@ -10,7 +10,11 @@ import { BedrockInvokeModel } from 'aws-cdk-lib/aws-stepfunctions-tasks';
  * * aws stepfunctions describe-execution --execution-arn <exection-arn generated before> : should return status as SUCCEEDED
  * This integ test does not actually verify a Step Functions execution, as not all AWS accounts have Bedrock model access.
  */
-const app = new cdk.App();
+const app = new cdk.App({
+  postCliContext: {
+    '@aws-cdk/aws-cdk.aws-stepfunctions-tasks:useNewS3UriParametersForBedrockInvokeModelTask': true,
+  },
```
moelasmar, line 16: if the default value of the flag is True, so we do not need to set this flag, and instead you need to update the old integration test cases to set the flag to false to test the current behaviour (using Input path, and output path instead of the new S3 uris)

moelasmar (approved):

shikha372:
@mergifyio update

PR #31307, opened 2024-09-03
docs: update note on downtime during migration

gracelu0:
update note on downtime, 2-step deployment should not cause any downtime

### DIFF
--- a/packages/aws-cdk-lib/aws-cloudfront-origins/README.md
+++ b/packages/aws-cdk-lib/aws-cloudfront-origins/README.md
@@ -495,7 +495,7 @@ No, following the migration steps does not cause any replacement of the existing
 
 **Will migrating from OAI to OAC have any availability implications for my application?**
 
-Migrating from OAI to OAC following the steps above (requires a 2-step deployment) should not cause any downtime. However, if you decide to skip Step 1 and migrate to OAC in a single deployment, you should be aware there is a possibility of downtime.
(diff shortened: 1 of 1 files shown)


packages/aws-cdk-lib/aws-cloudfront-origins/README.md
```diff
@@ -494,7 +494,8 @@ The following changes will take place:
 No, following the migration steps does not cause any replacement of the existing `AWS::CloudFront::Distribution`, `AWS::S3::Bucket` nor `AWS::S3::BucketPolicy` resources. It will modify the bucket policy, create a `AWS::CloudFront::OriginAccessControl` resource, and delete the existing `AWS::CloudFront::CloudFrontOriginAccessIdentity`.
 
 **Will migrating from OAI to OAC have any availability implications for my application?**
-While the above steps follow the order recommended by CloudFront, updates to CloudFront distributions and S3 bucket policies can take some time to propagate globally. Bucket configuration updates are eventually consistent. As such, you should be aware there is a possibility of downtime.
+
+Migrating from OAI to OAC following the steps above (requires a 2-step deployment) should not cause any downtime. However, if you decide to skip Step 1 and migrate to OAC in a single deployment, you should be aware there is a possibility of downtime.
```
samson-keung, line 498:
```suggestion
Updates to Bucket Policies are eventually consistent, therefore, removing OAI permissions and adding OAC in the same CloudFormation stack deployment is not recommended as it may introduce downtime where CloudFront loses access to the Bucket. Following the steps above will lower this risk ask the Bucket Policy is updated to have both OAI and OAC permissions, then in a subsequent deployment, the OAI permissions are removed.
```

samson-keung (approved):

packages/aws-cdk-lib/aws-cloudfront-origins/README.md
gracelu0: I like the suggestion is more nuanced - updated the answer :)

PR #31312, opened 2024-09-04
feat(pipes-enrichments): support API destination enrichment

### LINKED ISSUE #29383: (pipes-enrichments): Add API destination enrichment for EventBridge pipes
### Describe the feature

Add a implementation for the API destination enrichment for EventBridge pipes in the `pipes-enrichment` module.

### Use Case

For feature completeness of the L2 pipes module this source is required.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### CDK version used

-

-

mazyu36:
### Issue # (if applicable)

Closes #29383 .

### Reason for this change
To support API destination enrichment for EventBridge pipes.

### Description of changes
Add `ApiDestinationEnrichment` class.

Add unit tests and an integ test.

### Checklist

----


packages/@aws-cdk/aws-pipes-enrichments-alpha/lib/api-destination.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.inputTransformation = props?.inputTransformation;
+    this.headerParameters = props?.headerParameters;
+    this.queryStringParameters = props?.queryStringParameters;
+    this.pathParameterValues = props?.pathParameterValues;
+  }
+
+  bind(pipe: IPipe): EnrichmentParametersConfig {
+
+    const httpParameters: CfnPipe.PipeEnrichmentHttpParametersProperty | undefined =
+      this.headerParameters ??
```
sumupitchayan, line 61: Thanks for the PR @mazyu36 - overall looks good to me. I'm just wondering about the assignment of the `httpParameters` variable here - in practice, does it make sense to pass in HTTP parameters if one of the parameter values is undefined? (It might be nice to add a test for this, too?)

mazyu36: @sumupitchayan 
Yes, I think it has meaning. For example, when calling an API that only requires path parameters. 
This is in line with similar implementations as follows:
https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/aws-pipes-targets-alpha/lib/api-destination.ts
https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-events-targets/lib/api-destination.ts

I think it depends on the design of the API being called, so I think it would be sufficient to confirm that the parameters are set in the unit test. 
https://github.com/aws/aws-cdk/pull/31312/files#diff-0523525438cc22991d3e0555d86f9ec9c8a79cfa4a86599cdb2ea8d4075d4d84R63

I think adding integration tests unnecessary.
What do you think?

sumupitchayan: @mazyu36 thanks for the clarification. Looks good to me then 👍

PR #31324, opened 2024-09-04
feat(iam): validate AccountPrincipal argument to be exactly 12 digits

duranbe:
### Reason for this change

There is no validation and test that the AWS Account Id when creating AccountPrincipal Object. Thus, according to cdk `1234` is a valid AWS Account id.

Context : 
In my case I missed a digit when copy pasting an account id and the build still passed, the typo has been caught only during the code review process.
### Description of changes

Adding simple regex to check that AWS Id is 12 digits long & update error message

- Update existing tests
- Add 2 units ones

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-codepipeline/test/cross-env.test.ts
+++ b/packages/aws-cdk-lib/aws-codepipeline/test/cross-env.test.ts
@@ -201,7 +201,7 @@ describe('cross-environment CodePipeline', function () {
     const asm = app.synth();
     const supportStack = asm.getStackByName(`${pipelineStack.stackName}-support-234567890123`);
     Template.fromJSON(supportStack.template).hasResourceProperties('AWS::IAM::Role', {
-      RoleName: 'pipelinestack-support-456dbuildactionrolebf03938f0cc9a2599be5',
+      RoleName: 'pipelinestack-support-234dbuildactionrolec4ab6bf1cf3a58f0a3ba',
     });
 
     Template.fromStack(pipelineStack).hasResourceProperties('AWS::CodePipeline::Pipeline', {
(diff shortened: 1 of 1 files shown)


duranbe:
Exemption Request

packages/aws-cdk-lib/aws-cloudfront/test/distribution.test.ts
```diff
@@ -469,14 +469,14 @@ describe('certificates', () => {
       DistributionConfig: {
         Aliases: Match.absent(),
         ViewerCertificate: {
-          AcmCertificateArn: 'arn:aws:acm:us-east-1:123456789012:certificate/12345678-1234-1234-1234-123456789012',
+          AcmCertificateArn: 'arn:aws:acm:us-east-1:123456789012:certificate/12345678-123456789012-123456789012-123456789012-123456789012',
```
rix0rrr, line 472: The `1234` here was not an account ID, so I'm not sure why it is file is changed.

Did you do a mass search-replace over the entire codebase?

packages/aws-cdk/test/context-providers/security-groups.test.ts
```diff
@@ -266,11 +266,11 @@ describe('security group context provider plugin', () => {
     const provider = new SecurityGroupContextProviderPlugin(mockSDK);
 
     AWS.mock('EC2', 'describeSecurityGroups', (_params: aws.EC2.DescribeSecurityGroupsRequest, cb: AwsCallback<aws.EC2.DescribeSecurityGroupsResult>) => {
-      expect(_params).toEqual({ GroupIds: ['sg-1234'] });
+      expect(_params).toEqual({ GroupIds: ['sg-123456789012'] });
```
rix0rrr, line 269: Security group IDs are also not account IDs.

packages/aws-cdk-lib/aws-iam/lib/principals.ts
```diff
@@ -481,6 +481,9 @@ export class AccountPrincipal extends ArnPrincipal {
     if (!cdk.Token.isUnresolved(accountId) && typeof accountId !== 'string') {
       throw new Error('accountId should be of type string');
     }
+    if (!cdk.Token.isUnresolved(accountId) && !this.accountIdRegExp.test(accountId)) {
+      throw new Error('accountId should be composed of 12 digits');
+    }
```
rix0rrr, line 486: I wonder if there might be use cases where people are passing valid strings that aren't exactly 12 digits in length.

I'd feel more comfortable making this a warning, rather than an exception.

rix0rrr (changes requested):

PR #31328, opened 2024-09-05
fix(pipelines): "Node with duplicate id" on duplicate stack names

### LINKED ISSUE #30960: pipelines: Runtime error when a Stage has multiple Stacks with the same stack name
### Describe the bug

This is a duplicate of issue https://github.com/aws/aws-cdk/issues/30449 which should not have been closed. 

To provide additional context, if you have an app that you are deploying without pipelines, this behavior works, because the logical ids are different. Using waves only solves the case where you have two stacks with no other dependencies, or dependencies that are in the same region. If you have stacks which consume variables or exports from other stacks cross region by passing `crossRegionReferences: true` it is not possible to split these stacks into separate stages, because it is not possible to consume variables from stage to stage. This causes working cdk apps to be unable to be adopted by cdk pipelines.

### Expected Behavior

App is able to synth and deploy

### Current Behavior

Stacks with the same name but different IDs in the same stage cause a runtime error:

### Reproduction Steps

See duplicate issue

### Possible Solution

Add an explicit override parameter that can be used to bypass this check (which I'm sure does save a lot of headaches if code gets mistakenly copy pasted)

_No response_

### CDK CLI Version

2.150.0 (build 3f93027)

### Framework Version

_No response_

### Node.js Version

v18.20.4

### OS

Mac

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

rix0rrr:
When having 2 stacks with the same name in the same stage (which makes sense when deploying them to different environments), the CodePipeline Action name is derived from the stack name, and will be duplicated.

Detect if an graph node name is already being used and if so, use environment information to try and make the name unique.

Closes #30960.

### Checklist

----


packages/aws-cdk-lib/pipelines/lib/helpers-internal/pipeline-graph.ts
```diff
@@ -134,7 +134,12 @@ export class PipelineGraph {
     const stackGraphs = new Map<StackDeployment, AGraph>();
 
     for (const stack of stage.stacks) {
-      const stackGraph: AGraph = Graph.of(this.simpleStackName(stack.stackName, stage.stageName), { type: 'stack-group', stack });
+      const stackGraphName = findUniqueName(retGraph, [
+        this.simpleStackName(stack.stackName, stage.stageName),
+        ...stack.account ? [stack.account] : [],
+        ...stack.region ? [stack.region] : [],
+      ]);
```
GavinZZ, line 141: nit picking: Just thinking of an edge case when 4 stack names are the same, then this won't fix it right?

rix0rrr: But a different stack couldn't have the same name AND be deployed to both the same region AND account. Then it would just be the same stack.

GavinZZ (approved):

PR #31340, opened 2024-09-06
fix(lambda): configuring log retention fails on 70+ Lambdas

### LINKED ISSUE #31338: aws-lambda: Log retention gives rate exceeded error
### Describe the bug

[Legacy log retention](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.Function.html#logretention) in Lambda gives a rate limit exceeded error. 

We are in the process of upgrading our app from CDK v1 to v2. To test this, we created a new env in a new account and redeployed the configuration using CDK v1. 

We are creating 70-80 lambdas with log retention enabled. The legacy log retention creates a custom lambda resource to create log group and set log retention. CDK V1 used to create Node 14 lambas for this purpose (for which the creation is blocked in AWS). Hence, we disabled the log retention and upgraded the stack to 2.151.0 and then enabled the log retention.

While doing so, our stack is failing with the error: 
```
Received response status [FAILED] from custom resource. Message returned: Out of attempts to change log group
```

Initially we thought this is an issue with the `“CreateLogGroup throttle limit in transactions per second” quota`. We increased it to 80 from 10 but the issue still exists.

On exploring the cloudwatch logs for the custom lambda resource, we found: 
```
2024-09-06T05:23:33.260Z	06a9833f-0ad3-4faf-8f94-aa78dd49d0ec	ERROR	{
  clientName: 'CloudWatchLogsClient',
  commandName: 'PutRetentionPolicyCommand',
  input: {
    logGroupName: '/aws/lambda/LogRetentionaae0aa3c5b4d-mE6Tt6xks1CB',
    retentionInDays: 1
  },
  error: ThrottlingException: Rate exceeded
      at de_ThrottlingExceptionRes (/var/runtime/node_modules/@aws-sdk/client-cloudwatch-logs/dist-cjs/index.js:2321:21)
      at de_CommandError (/var/runtime/node_modules/@aws-sdk/client-cloudwatch-logs/dist-cjs/index.js:2167:19)
      at async /var/runtime/node_modules/@aws-sdk/node_modules/@smithy/middleware-serde/dist-cjs/index.js:35:20
      at async /var/runtime/node_modules/@aws-sdk/node_modules/@smithy/core/dist-cjs/index.js:165:18
      at async /var/runtime/node_modules/@aws-sdk/node_modules/@smithy/middleware-retry/dist-cjs/index.js:320:38
      at async /var/runtime/node_modules/@aws-sdk/middleware-logger/dist-cjs/index.js:34:22
      at async /var/task/index.js:1:1148
      at async /var/task/index.js:1:2728
      at async y (/var/task/index.js:1:1046) {
    '$fault': 'client',
    '$metadata': {
      httpStatusCode: 400,
      requestId: 'e247739c-8ebb-40d3-b85e-293802a87e24',
      extendedRequestId: undefined,
      cfId: undefined,
      attempts: 3,
      totalRetryDelay: 466
    },
    __type: 'ThrottlingException'
  },
  metadata: {
    httpStatusCode: 400,
    requestId: 'e247739c-8ebb-40d3-b85e-293802a87e24',
    extendedRequestId: undefined,
    cfId: undefined,
    attempts: 3,
    totalRetryDelay: 466
  }
}

```

Looks like an issue with the rate limit for `PutRetentionPolicyCommand`. The service quota for the same cannot be changed. Our earlier implementation had one difference in how log retention was implemented.
The `base` property was enabled to apply a exponential backoff (probably to handle such cases). This is now [deprecated](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.LogRetentionRetryOptions.html) and hence we removed it during our upgrade from CDK v1 to v2. The documentation for [LogRetentionRetryOptions](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.LogRetentionRetryOptions.html) says that this was removed as it is handled differently in AWS SDK v3. Is this what is causing the issue? Should't CDK/ SDK handle the backoff in this case? 

### Regression Issue

- [X] Select this option if this issue appears to be a regression.

1.204.0

### Expected Behavior

Log retention backoff should be handled internally

### Current Behavior

Creating legacy log retention for multiple lambas together gives a rate limit exceeded error.

### Reproduction Steps

-

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.108.1

### Framework Version

_No response_

### Node.js Version

v22.1.0

### OS

MacOs

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

rix0rrr:
When the Log Retention Lambda runs massively parallel (on 70+ Lambdas at the same time), it can run into throttling problems and fail.

Raise the retry count and delays:

- Raise the default amount of retries from 5 -> 10
- Raise the sleep base from 100ms to 1s.
- Change the sleep calculation to apply the 60s limit *after* jitter instead of before (previously, we would take a fraction of 60s; now we're taking a fraction of the accumulated wait time, and after calculating that limit it to 60s).

Fixes #31338.

### Checklist

----

### DIFF
--- a/CHANGELOG.v2.alpha.md
+++ b/CHANGELOG.v2.alpha.md
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [2.160.0-alpha.0](https://github.com/aws/aws-cdk/compare/v2.159.1-alpha.0...v2.160.0-alpha.0) (2024-09-24)
+
+
+### Features
+
+* **kinesisanalytics-flink:** support Apache Flink 1.20 ([#31349](https://github.com/aws/aws-cdk/issues/31349)) ([b3b9aa8](https://github.com/aws/aws-cdk/commit/b3b9aa8a4584e178808f5babe0583749d1a87da5))
+
+
+### Bug Fixes
+
+* **cognito-identitypool-alpha:** cannot configure roleMappings with imported userPool and client ([#30421](https://github.com/aws/aws-cdk/issues/30421)) ([0fdd6a9](https://github.com/aws/aws-cdk/commit/0fdd6a92ac7196e5498969be2743019f825b9262)), closes [#30304](https://github.com/aws/aws-cdk/issues/30304) [/github.com/aws/aws-cdk/blob/c3003ab41f0efc763f39eb2cab490c8a005e146b/packages/aws-cdk-lib/aws-cognito/lib/user-pool.ts#L902](https://github.com/aws//github.com/aws/aws-cdk/blob/c3003ab41f0efc763f39eb2cab490c8a005e146b/packages/aws-cdk-lib/aws-cognito/lib/user-pool.ts/issues/L902)
+* **ec2:** instance resourceSignalTimeout overwrites initOptions.timeout ([#31446](https://github.com/aws/aws-cdk/issues/31446)) ([a29bf19](https://github.com/aws/aws-cdk/commit/a29bf19be1e17c13b85f6edd45c382c1f0d89702)), closes [#30052](https://github.com/aws/aws-cdk/issues/30052)
+
 ## [2.159.1-alpha.0](https://github.com/aws/aws-cdk/compare/v2.159.0-alpha.0...v2.159.1-alpha.0) (2024-09-19)
 
 ## [2.159.0-alpha.0](https://github.com/aws/aws-cdk/compare/v2.158.0-alpha.0...v2.159.0-alpha.0) (2024-09-18)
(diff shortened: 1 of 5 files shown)


mrgrain (approved):

mrgrain (changes requested):
PR not completed. Will review again once complete.

packages/@aws-cdk/custom-resource-handlers/lib/aws-logs/log-retention-handler/index.ts
```diff
@@ -188,8 +193,8 @@ function parseIntOptional(value?: string, base = 10): number | undefined {
 
 function makeWithDelay(
   maxRetries: number,
-  delayBase: number = 100,
-  delayCap = 10 * 1000, // 10s
+  delayBase: number = 1_000,
+  delayCap = 60_000, // 60s
```
iliapolo, line 197: Nit: PR description says you cap at 10s, but this says 60s.

PR #31380, opened 2024-09-09
feat(ivs): support RTMP ingest for IVS channel

mazyu36:
### Issue # (if applicable)
N/A

### Reason for this change
Add missing property.

### Description of changes
Add `insecureIngest`property.

Add unit tests and integ test.

### Checklist

----


packages/@aws-cdk/aws-ivs-alpha/README.md
```diff
@@ -55,6 +55,18 @@ const myChannel = new ivs.Channel(this, 'myChannel', {
 });
 ```
 
+If you want to use RTMP ingest, set `ingestInsecure` property to `true`.
+
+By default, `ingestInsecure` is `false` which means using RTMPS ingest.
```
GavinZZ, line 60: I think the property name is `insecureIngest`

GavinZZ (dismissed):
Minor change

packages/@aws-cdk/aws-ivs-alpha/README.md
mazyu36: Thanks. I’ve fixed.

> Also by default the value is false, so should means not using the RTMP ingest right?

Basically yes, AWS also recommends using RTMP only for specific use cases. The relationship between RTMPS and RTMP is similar to that of HTTPS and HTTP.

RTMP Ingest was added on 2023/3/20. (Ref: https://docs.aws.amazon.com/ivs/latest/LowLatencyUserGuide/doc-history.html)
IMG_3374

Below is the explanation from the management console. I've added a NOTE to the README based on this information.

IMG_3373

If this property should not be added, please let me know.
There are cases where the software only supports RTMP, and I want this property. 
However, since it might weaken security, I also think it might be an option to not add this property to L2 and instead use escape hatches if someone wants to use RTMP.

mazyu36:
@GavinZZ 
Thank you for always reviewing. I’ve addressed the comments.

packages/@aws-cdk/aws-ivs-alpha/README.md
```diff
@@ -55,6 +55,17 @@ const myChannel = new ivs.Channel(this, 'myChannel', {
 });
 ```
 
+If you want to use RTMP ingest, set `insecureIngest` property to `true`.
+By default, `insecureIngest` is `false` which means using RTMPS ingest.
```
GavinZZ, line 59: > `insecureIngest` is `false` which means using RTMPS ingest

Is this right?

GavinZZ (approved):

PR #31385, opened 2024-09-10
feat(rds): support performance insights configuration at cluster level

### LINKED ISSUE #31375: (rds): add `performanceInisghts` configuration at cluster level
### Describe the feature

Add the performance insights configuration options at cluster level:
```ts
/**
 * Whether to enable Performance Insights for the DB instance.
 *
 * @default - false, unless `performanceInsightRetention` or `performanceInsightEncryptionKey` is set.
 */
readonly enablePerformanceInsights?: boolean;

/**
 * The amount of time, in days, to retain Performance Insights data.
 *
 * @default 7
 */
readonly performanceInsightRetention?: PerformanceInsightRetention;

/**
 * The AWS KMS key for encryption of Performance Insights data.
 *
 * @default - default master key
 */
readonly performanceInsightEncryptionKey?: kms.IKey;
```

### Use Case

So we don't need to configure it to all instances. Also, because CloufFormation allows it.

### Proposed Solution

Add the same 3 configuration options present in `DatabaseInstance` class, to the `DatabaseCluster` class: `enablePerformanceInsights`, `performanceInsightRetention` and `performanceInsightEncryptionKey`.

### Other Information

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-performanceinsightsenabled

### Acknowledgements

### CDK version used

2.256

macOS 14.6.1

go-to-k:
### Issue # (if applicable)

Closes #31375 .

### Reason for this change

Properties for Performance Insights at cluster level are supported in L1, but not in L2.

### Description of changes

Added the properties in props for Database Cluster.

Both unit tests and integ tests.

### Checklist

----


packages/aws-cdk-lib/aws-rds/README.md
```diff
@@ -476,14 +476,14 @@ to use for the instance:
 declare const vpc: ec2.Vpc;
 
 const iopsInstance = new rds.DatabaseInstance(this, 'IopsInstance', {
-  engine: rds.DatabaseInstanceEngine.mysql({ version: rds.MysqlEngineVersion.VER_8_0_30 }),
```
go-to-k, line 479: VER_8_0_30 is deprecated.

packages/aws-cdk-lib/aws-rds/README.md
```diff
@@ -1242,3 +1242,87 @@ new rds.DatabaseCluster(this, 'DatabaseCluster', {
...
+  }),
+});
+```
+
+To enable Performance Insights at the instance level, set the same properties for each instance of the `writer` and the `readers`.
+
+In this way, different settings can be applied to different instances in a cluster.
+
+However, if Performance Insights is enabled at the cluster level, it is not possible to specify a different value for the instance
+level than the cluster level.
```
go-to-k, line 1279: If Performance Insights was enabled at cluster and instance level, and the values of the `performanceInsightRetention` or `performanceInsightEncryptionKey` were different for the cluster and the instance, CFn failed.

If `performanceInsightRetention`s are different, CFn occurred:

```
Resource handler returned message: "PerformanceInsightsRetentionPeriod conflicts with cluster level parameter. ...“
```

If `performanceInsightEncryptionKey`s are different, CFn occurred:

```
Resource handler returned message: "PerformanceInsightsKMSKeyId conflicts with cluster level parameter. …”
```

However, if Performance Insights was enabled at cluster and instance level and the values were **the same** for the cluster and the instance, **CFn did not fail.**

If the value of `enablePerformanceInsights` was different for the cluster and the instance, it **did not result in an error**. If it was true in the cluster and **false in the instance**, it did not result in an error but was also **enabled in the instance**.

===

I wanted to implement validation of this behaviour. However, the instance does not have Performance Insights settings publicly, so it is not accessible from the cluster. Therefore, to implement validation, the structure of the instances had to be changed significantly, so I decided not to do it in this PR.

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
```diff
@@ -730,6 +757,11 @@ abstract class DatabaseClusterNew extends DatabaseClusterBase {
       copyTagsToSnapshot: props.copyTagsToSnapshot ?? true,
       domain: this.domainId,
       domainIamRoleName: this.domainRole?.roleName,
+      performanceInsightsEnabled: enablePerformanceInsights || props.enablePerformanceInsights, // fall back to undefined if not set
```
lpizzinidev, line 760:
```suggestion
      performanceInsightsEnabled: enablePerformanceInsights,
```
Isn't the second part of the condition redundant?

packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
```diff
@@ -1713,7 +1713,66 @@ describe('cluster', () => {
     });
   });
 
-  describe('performance insights', () => {
+  describe('performance insights for cluster', () => {
+    test('cluster with all performance insights properties', () => {
```
lpizzinidev, line 1717: Can you add coverage for the scenario where `enablePerformanceInsights: true` and `performanceInsightRetention` / `performanceInsightEncryptionKey` are not specified?

packages/aws-cdk-lib/aws-rds/README.md
lpizzinidev, line 1279:
```suggestion
**Note:** If Performance Insights is enabled at the cluster level, it is not possible to specify a different value for the instance
level than the cluster level.
```
Let's highlight this behavior.

lpizzinidev: > the structure of the instances had to be changed significantly,

Isn't exposing the props in [`AuroraClusterInstance`](https://github.com/aws/aws-cdk/blob/1726abddbe10a22c092de7bf0b566710d13f5efd/packages/aws-cdk-lib/aws-rds/lib/aurora-cluster-instance.ts#L444) and adding validation in [`validateClusterInstances`](https://github.com/aws/aws-cdk/blob/1726abddbe10a22c092de7bf0b566710d13f5efd/packages/aws-cdk-lib/aws-rds/lib/cluster.ts#L795) enough? 🤔

lpizzinidev (changes requested):

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
go-to-k: In that case, `performanceInsightsEnabled` in `this.newCfnProps` will be `false` if all `props.enablePerformanceInsights`, `props.performanceInsightRetention` and `props.performanceInsightEncryptionKey` are undefined.

```ts
    const enablePerformanceInsights = props.enablePerformanceInsights
      || props.performanceInsightRetention !== undefined || props.performanceInsightEncryptionKey !== undefined;
```

However, if the aws-cdk-lib is only upgraded after this PR has been merged, the CFn template will be changed with `PerformanceInsightsEnabled: false`, even though the user has not specified anything additional. Therefore, we should fall back to undefined in this case.

packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
go-to-k: Yeah, I added: https://github.com/aws/aws-cdk/pull/31385/commits/217d94552e3f5953fe1588e7ea751385879c17ea

packages/aws-cdk-lib/aws-rds/README.md
go-to-k: Added: https://github.com/aws/aws-cdk/pull/31385/commits/217d94552e3f5953fe1588e7ea751385879c17ea

go-to-k: I added the validation, what do you think? 

The cluster has a writer and readers as `IAuroraClusterInstance` not `AuroraClusterInstance`, so I added the variable `performanceInsightsEnabled` in `IAuroraClusterInstance` interface. 

And the `legacyCreateInstances` function for the database cluster is not a method, I also added a public variable `performanceInsightsEnabled` in `DatabaseClusterNew`.

https://github.com/aws/aws-cdk/pull/31385/commits/cf568902909eb3e0ee5ec2abed91623d08bc7363

go-to-k:
@lpizzinidev

Thanks for your review.
I have made the changes, so could you please take a look at them?

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
lpizzinidev: Thanks for clarifying 👍

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
```diff
@@ -793,9 +831,14 @@ abstract class DatabaseClusterNew extends DatabaseClusterBase {
    * Perform validations on the cluster instances
    */
   private validateClusterInstances(writer: IAuroraClusterInstance, readers: IAuroraClusterInstance[]): void {
+    let performanceInsightsDuplicate = false;
```
lpizzinidev, line 834: Shouldn't we validate that `performanceInsightEncryptionKey` and `performanceInsightRetention` match between the cluster and instances instead? We're blocking valid deployments with this setup.

> If the value of enablePerformanceInsights was different for the cluster and the instance, it did not result in an error. If it was true in the cluster and false in the instance, it did not result in an error but was also enabled in the instance.

Should we display a warning for this scenario?

packages/aws-cdk-lib/aws-rds/README.md
```diff
@@ -1242,3 +1242,87 @@ new rds.DatabaseCluster(this, 'DatabaseCluster', {
...
+  }),
+});
+```
+
+To enable Performance Insights at the instance level, set the same properties for each instance of the `writer` and the `readers`.
+
+In this way, different settings can be applied to different instances in a cluster.
+
+**Note:** If Performance Insights is enabled at the cluster level, it is not possible to specify a different value for the instance
+level than the cluster level.
```
lpizzinidev, line 1279:
```suggestion
**Note:** If Performance Insights is enabled at the cluster level, it is also automatically enabled at the instance level. If specified, Performance Insights at the instance level require the same retention period and encryption key as the cluster level.
```

lpizzinidev (changes requested):

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
go-to-k: I added the validations for `performanceInsightEncryptionKey` and `performanceInsightRetention` and the warning . The implementation has become a bit complicated, what do you think?

packages/aws-cdk-lib/aws-rds/README.md
go-to-k: I changed, but slightly reworded. Is there a problem? (The reason is that "cluster level" and "instance level" are exclusive settings, so I thought using the word "instance level" when enabled on the cluster level might be confusing to users.)

```
**Note:** If Performance Insights is enabled at the cluster level, it is also automatically enabled for each instance. If specified, Performance Insights
for each instance require the same retention period and encryption key as the cluster level.
```

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
```diff
@@ -905,9 +912,45 @@ abstract class DatabaseClusterNew extends DatabaseClusterBase {
...
+    if (this.performanceInsightEncryptionKey && instance.performanceInsightEncryptionKey) {
+      const clusterKeyArn = this.performanceInsightEncryptionKey.keyArn;
+      const instanceKeyArn = instance.performanceInsightEncryptionKey.keyArn;
+      const compared = Token.compareStrings(clusterKeyArn, instanceKeyArn);
+
+      if (compared === TokenComparison.DIFFERENT) {
+        throw new Error(`\`performanceInsightEncryptionKey\` for each instance must be the same as the one at cluster level, got instance '${instance.node.id}': '${instance.performanceInsightEncryptionKey.keyArn}', cluster: '${this.performanceInsightEncryptionKey.keyArn}'`);
+      }
+      // Even if both of cluster and instance keys are unresolved, check if they are the same token.
+      if (compared === TokenComparison.BOTH_UNRESOLVED && clusterKeyArn !== instanceKeyArn) {
```
lpizzinidev, line 951: Not sure about the validity of this condition 🤔
Given that we're already ignoring the `ONE_UNRESOLVED` case, we could maybe skip the check for `BOTH_UNRESOLVED` as well

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
```diff
@@ -1514,3 +1561,48 @@ function legacyCreateInstances(cluster: DatabaseClusterNew, props: DatabaseClust
 function databaseInstanceType(instanceType: ec2.InstanceType) {
   return 'db.' + instanceType.toString();
 }
+
+/**
+ * Validate Performance Insights settings for legacy instanceProps with cluster level settings
+ */
+function legacyValidatePerformanceInsightsSettingsWithCluster(
```
lpizzinidev, line 1568: Can we create a single method using params? Idea
```ts
private validatePerformanceInsightsSettings(cluster: { 
    performanceInsightsEnabled: boolean, 
    performanceInsightRetention?: PerformanceInsightRetention, 
    performanceInsightEncryptionKey?: kms.IKey 
  }, instance: { 
    performanceInsightsEnabled: boolean, 
    performanceInsightRetention?: PerformanceInsightRetention, 
    performanceInsightEncryptionKey?: kms.IKey 
  }): void {
```

packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
```diff
@@ -1798,15 +1819,139 @@ describe('cluster', () => {
...
+      });
+
+      // THEN
+      Annotations.fromStack(stack).hasWarning('*',
+        'Performance Insights is enabled on cluster \'Database\' at cluster level, but disabled for instance \'writer\'. '+
+        'However, Performance Insights for this instance will also be automatically enabled if enabled at cluster level. [ack: @aws-cdk/aws-rds:instancePerformanceInsightsOverridden]',
+      );
+    });
+
+    test('warn if performance insights is enabled at cluster level but disabled on reader instance', () => {
```
lpizzinidev, line 1851: We could remove this as redundant with `warn if performance insights is enabled at cluster level but disabled on writer and reader instances`

packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
```diff
@@ -1798,15 +1819,139 @@ describe('cluster', () => {
...
           vpc,
-          writer: ClusterInstance.provisioned('writer', {
+          writer: ClusterInstance.provisioned('writer'),
+          enablePerformanceInsights: false,
+          performanceInsightRetention: PerformanceInsightRetention.DEFAULT,
+        });
+      }).toThrow(/`enablePerformanceInsights` disabled, but `performanceInsightRetention` or `performanceInsightEncryptionKey` was set/);
+    });
+
+    test('warn if performance insights is enabled at cluster level but disabled on writer instance', () => {
```
lpizzinidev, line 1829: We could remove this as redundant with `warn if performance insights is enabled at cluster level but disabled on writer and reader instances`

packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
```diff
@@ -1815,30 +1960,371 @@ describe('cluster', () => {
...
+          enablePerformanceInsights: true, // default period is set by the construct if the `enablePerformanceInsights` is enabled
+          instanceProps: {
+            vpc,
+            performanceInsightRetention: PerformanceInsightRetention.MONTHS_12,
+          },
+        });
+      }).toThrow(/`performanceInsightRetention` for each instance must be the same as the one at cluster level, got `instanceProps`: 372, cluster: 7/);
+    });
+
+    test('does not throw if performanceInsightRetention is the same value on cluster and instanceProps', () => {
```
lpizzinidev, line 2206: We could simplify the implementation by merging the `not.toThrow();` cases. Idea
```ts
test.each([
  [
    true, PerformanceInsightRetention.LONG_TERM, new kms.Key(stack, 'Key1'), // cluster props
    true, PerformanceInsightRetention.LONG_TERM, new kms.Key(stack, 'Key1'), // instance props
 ],
])('does not throw if ...', (clusterPerformanceInsightsEnabled: boolean, clusterPerformanceInsightRetention: ...) => {
```

lpizzinidev (changes requested):

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
go-to-k: I was certainly wondering about this myself. However, I thought that if only one of the values is a token, it is impossible to compare them in any way, but if both are tokens, they can be compared by check whether both are the same token.

I thought that excluding this case would diminish the validation as it would mean skipping the most common cases, such as the following. What do you think? (I still wonder...)

```ts
new DatabaseCluster(stack, 'Database', {
  engine: DatabaseClusterEngine.AURORA,
  vpc,
  performanceInsightEncryptionKey: new kms.Key(stack, 'Key1'),
  writer: ClusterInstance.provisioned('writer', {
    performanceInsightEncryptionKey: new kms.Key(stack, 'Key2'),
  }),
});
```

packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
go-to-k: If we turn them off, we will not be able to see whether the validation process has been done for both writer and reader. Therefore, may I suggest that we turn the test (`warn if performance insights is enabled at cluster level but disabled on writer and reader instances`) off?

go-to-k: Sorry, I misunderstood. This is about the warning, so we could see it. I'll take it.

go-to-k: done: https://github.com/aws/aws-cdk/pull/31385/commits/a167d1f12169f219eca03b811b50e70b7563566a

packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
go-to-k: done: https://github.com/aws/aws-cdk/pull/31385/commits/a167d1f12169f219eca03b811b50e70b7563566a

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
go-to-k: done: https://github.com/aws/aws-cdk/pull/31385/commits/c0c937f6ef062c3c46df34087250f0a6070c377b

- `IConstruct` is needed for annotations, so `DatabaseClusterNew` was passed to cluster.
- Since `${instance.node.id}` is required in the error message, I added `nodeId` to the `instance`.
  - It is not needed in `instanceProps`, so it is left as an option.

```ts
 */
function validatePerformanceInsightsSettings(
  cluster: DatabaseClusterNew,
  instance: {
    nodeId?: string;
    performanceInsightsEnabled?: boolean;
    performanceInsightRetention?: PerformanceInsightRetention;
    performanceInsightEncryptionKey?: kms.IKey;
  },
): void {
```

packages/aws-cdk-lib/aws-rds/test/cluster.test.ts
go-to-k: done: https://github.com/aws/aws-cdk/pull/31385/commits/e5d546385a887ea622bb8227829ee659c9db1fd8

```ts
    test.each([
      [
        ...
        .., key, // <- here
```

If a Construct instance is passed as an argument to `test.each` in this way, the resource must be created not only in `beforeEach`, but also at the beginning. For this reason, the following is used.

```ts
  describe('performance insights for cluster', () => {
    function setTestStack() {
      const stack = testStack();
      const vpc = new ec2.Vpc(stack, 'VPC');
      const key = new kms.Key(stack, 'Key');
      const importedKey = kms.Key.fromKeyArn(stack, 'ImportedKey', 'arn:aws:kms:us-east-1:123456789012:key/imported');
      return { stack, vpc, key, importedKey };
    }
    // Needs to be declared first, not just beforeEach, for use in `test.each` arguments
    let { stack, vpc, key, importedKey } = setTestStack();

    beforeEach(() => {
      ({ stack, vpc, key, importedKey } = setTestStack());
    });
```

go-to-k:
@lpizzinidev 

I have reflected on your points and replied to them, could you take another look at them?

lpizzinidev (approved):
Thanks 👍

packages/aws-cdk-lib/aws-rds/lib/aurora-cluster-instance.ts
```diff
@@ -436,9 +436,24 @@ export interface IAuroraClusterInstance extends IResource {
...
+
+  /**
+   * The amount of time, in days, to retain Performance Insights data.
+   */
+  readonly performanceInsightRetention?: PerformanceInsightRetention;
+
+  /**
+   * The AWS KMS key for encryption of Performance Insights data.
+   */
+  readonly performanceInsightEncryptionKey?: kms.IKey;
```
GavinZZ, line 456: Why do they need to be introduced in the interface?

packages/aws-cdk-lib/aws-rds/lib/aurora-cluster-instance.ts
```diff
@@ -484,6 +502,12 @@ class AuroraClusterInstance extends Resource implements IAuroraClusterInstance {
       throw new Error('`enablePerformanceInsights` disabled, but `performanceInsightRetention` or `performanceInsightEncryptionKey` was set');
     }
 
+    this.performanceInsightsEnabled = enablePerformanceInsights;
+    this.performanceInsightRetention = enablePerformanceInsights
+      ? (props.performanceInsightRetention || PerformanceInsightRetention.DEFAULT)
+      : undefined;
+    this.performanceInsightEncryptionKey = props.performanceInsightEncryptionKey;
```
GavinZZ, line 509: Why is this needed instead of the old way

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
```diff
@@ -1477,3 +1551,53 @@ function legacyCreateInstances(cluster: DatabaseClusterNew, props: DatabaseClust
...
+  },
+): void {
+  const target = instance.nodeId ? `instance \'${instance.nodeId}\'` : '`instanceProps`';
+
+  // If Performance Insights is enabled on the cluster, the one for each instance will be enabled as well.
+  if (cluster.performanceInsightsEnabled && instance.performanceInsightsEnabled === false) {
+    Annotations.of(cluster).addWarningV2(
+      '@aws-cdk/aws-rds:instancePerformanceInsightsOverridden',
+      `Performance Insights is enabled on cluster '${cluster.node.id}' at cluster level, but disabled for ${target}. `+
+      'However, Performance Insights for this instance will also be automatically enabled if enabled at cluster level.',
```
GavinZZ, line 1574: should we just throw an error here to match the bebhaviour with the bottom checks?

GavinZZ (changes requested):
Thanks for contributing! Left some minor feedbacks

packages/aws-cdk-lib/aws-rds/lib/aurora-cluster-instance.ts
go-to-k: The reason for this is that in order to validate instance's settings in the `DatabaseClusterNew` class, properties need to be added to the interface. Otherwise, the class cannot access the properties.

see: https://github.com/aws/aws-cdk/pull/31385#discussion_r1760052070

```ts
  /**
   * Perform validations on the cluster instances
   */
  private validateClusterInstances(writer: IAuroraClusterInstance, readers: IAuroraClusterInstance[]): void {
    if (writer.type === InstanceType.SERVERLESS_V2) {
      this.hasServerlessInstance = true;
    }
    validatePerformanceInsightsSettings(this, {
      nodeId: writer.node.id,
      performanceInsightsEnabled: writer.performanceInsightsEnabled,
      performanceInsightRetention: writer.performanceInsightRetention,
      performanceInsightEncryptionKey: writer.performanceInsightEncryptionKey,
    });
```

packages/aws-cdk-lib/aws-rds/lib/aurora-cluster-instance.ts
go-to-k: This is just to set it to new variables added to the `AuroraClusterInstance` instance variables.

L467-469
```ts
  public readonly performanceInsightsEnabled: boolean;
  public readonly performanceInsightRetention?: PerformanceInsightRetention;
  public readonly performanceInsightEncryptionKey?: kms.IKey;
```

packages/aws-cdk-lib/aws-rds/lib/cluster.ts
go-to-k: CFn can also deploy with this configuration. It just means that the instance configuration is overridden by the Aurora layer.

If it can be deployed as the correct behaviour in CFn but not in CDK, it will block valid deployments. This is why we have made it a warning here.

What do you think?

GavinZZ: That's good to know. I agree with your approach instead if it's deployable.

GavinZZ (approved):
LGTM

PR #31402, opened 2024-09-11
feat(cognito): new `cloudFrontEndpoint` method for user pool domain without custom resource

go-to-k:
### Issue # (if applicable)

Closes #31342.

### Reason for this change

`UserPoolDomain` creates a custom resource to get CloudFront endpoint. However, CFn exposes the attribute natively now (see the issue). No custom resource is better if it is not needed.

### Description of changes

I propose a new method `cloudFrontEndpoint` without a custom resource.

Three ways were originally considered. This method was chosen as it was the most reasonable of all.

#### 1. Create a new method

This is the method submitted in this PR.

#### 2. Rewrite an existing method directly

This causes destructive changes. Custom resources are not directly used in the user's application. However, this change will result in resource deletion in the user's CDK stack. This causes confusion for users and should be avoided.

Also, the existing integ tests that use the method will fail because the changes are considered as destructive changes.

```
Tests:    1 failed, 1 total
Failed: /aws-cdk/packages/@aws-cdk-testing/framework-integ/test/aws-cognito/test/integ.user-pool-domain-cfdist.js
!!! This test contains destructive changes !!!
    Stack: integ-user-pool-domain-cfdist - Resource: UserPoolDomainCloudFrontDomainNameE213E594 - Impact: WILL_DESTROY
    Stack: integ-user-pool-domain-cfdist - Resource: UserPoolDomainCloudFrontDomainNameCustomResourcePolicy7DE54188 - Impact: WILL_DESTROY
    Stack: integ-user-pool-domain-cfdist - Resource: AWS679f53fac002430cb0da5b7982bd2287ServiceRoleC1EA0FF2 - Impact: WILL_DESTROY
    Stack: integ-user-pool-domain-cfdist - Resource: AWS679f53fac002430cb0da5b7982bd22872D164C4C - Impact: WILL_DESTROY
!!! If these destructive changes are necessary, please indicate this on the PR !!!
Error: Some changes were destructive!
    at main (/aws-cdk/packages/@aws-cdk/integ-runner/lib/cli.js:183:15)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

#### 3. Rewrite an existing method with a feature flag

An alternative to way 2, but a feature flag was avoided in this case as it leads to complexity. The [design guidelines](https://github.com/aws/aws-cdk/blob/main/CONTRIBUTING.md#dealing-with-breaking-behavior-changes) also recommend a new method.

### Additional information

[...]

PR #31403, opened 2024-09-11
fix(route53-targets): deprecated method for dns name is used in userpool domain target (under feature flag)

go-to-k:
### Reason for this change

The [PR](https://github.com/aws/aws-cdk/pull/31402) created a new method to get CloudFront DNS name in Cognito user pool domain. A custom resource is created in the old method, but is not in the new method.

However, the `UserPoolDomainTarget` in the `route53-targets` module continues to use the old method. So we should change to use the new method.

### Description of changes

The `bind` method in the `UserPoolDomainTarget` implements `IAliasRecordTarget` interface, so a new method instead of the `bind` cannot be created.

Therefore, I take it using a feature flag.

Both unit and integ tests.

### Checklist

----


nmussy:
I didn't have a long enough look to give a meaningful review, but I'm not sure "new method" is descriptive enough. Not to mention that we might add "newer" method flags in the future

go-to-k:
@nmussy 

You're absolutely right (I hadn't really thought about it properly...). I have corrected it as such.

packages/@aws-cdk-testing/framework-integ/test/aws-route53-targets/test/integ.userpool-domain-alias-target.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+
+const app = new App({
+  context: { [USER_POOL_DOMAIN_NAME_METHOD_WITHOUT_CUSTOM_RESOURCE]: true },
+});
+const stack = new TestStack(app, 'userpool-domain-alias-target');
+new IntegTest(app, 'userpool-domain-alias-target-integ', {
+  testCases: [stack],
+});
```
nmussy, line 34: I was going to ask to add a test to make sure the CR wasn't created, but you're already checking for it in the unit test, so it might be overkill

https://github.com/aws/aws-cdk/pull/31403/files#diff-d39a13fcbdb9a308f6017d9800b5f4244d21103d76b708c68d84226d0172faeaR78-R79

nmussy (approved):
LGTM. It might be better to rename the feature flag to say what it does vs what it doesn't do anymore, ie. attribute referencing vs CR lookup, but I'll let maintainers weigh in on this

GavinZZ (approved):

PR #31406, opened 2024-09-11
fix(cli): hotswapping appsync functions fails when API does not return function on the first page

dpilch:
### Issue # (if applicable)

https://github.com/aws-amplify/amplify-category-api/issues/2307

### Reason for this change

Hotswap for AppSync functions sometimes fails with `Deployment failed: MissingRequiredParameter: Missing required key 'functionId' in params`. This is because the SDK list function only retrieves the first page. In APIs where there are many functions it is possible for the hotswapped function to not be contained in the first page. This results in the previously mentioned error because the `functionId` is never retrieved.

### Description of changes

List all AppSync functions for a given API by iterating through the pages with the `nextToken`. The change was modeled after a similar request [here](https://github.com/aws/aws-cdk/blob/1e203753519e10e19ef0db87e1382377b609bcaa/packages/aws-cdk/lib/api/evaluate-cloudformation-template.ts#L23-L36).

* Unit tests
* Integ tests
* Manual testing in app from https://github.com/aws-amplify/amplify-category-api/issues/2307

### Checklist

----


phani-srikar (approved):

kaizencc (changes requested):
Hi! The cli integ test you wrote is not succeeding. Did you run this locally to ensure it worked? If not, the instructions for how to do so are [here](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk-testing/cli-integ/README.md).

Specifically the failure is this:

```
const api = new appsync.GraphqlApi(this, "Api", {
--

ReferenceError: appsync is not defined
```

Looks like you need to add a line like [this](https://github.com/aws/aws-cdk/blob/bf3764fe6046799a2a6e074b39b9a5f4474c603c/packages/%40aws-cdk-testing/cli-integ/resources/cdk-apps/app/app.js#L15) to the top of `app.js`. But please run the cli integ test on your own to ensure it does as its supposed to still!

Once that has been fixed i'll run it through the pipeline again.

sundersc (approved):

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -155,3 +155,18 @@ async function simpleRetry(fn: () => Promise<any>, numOfRetries: number, errorCo
 async function sleep(ms: number) {
   return new Promise(ok => setTimeout(ok, ms));
 }
+
+/**
+ * Get all functions for a given AppSync API by iterating through the paginated list of functions
+ */
+async function getAppSyncFunctions(sdk: ISDK, apiId: string, nextToken: string | undefined): Promise<FunctionConfiguration[]> {
```
kaizencc, line 162:
```suggestion
async function getAppSyncFunctions(sdk: ISDK, apiId: string, nextToken?: string): Promise<FunctionConfiguration[]> {
```

packages/aws-cdk/lib/api/hotswap/appsync-mapping-templates.ts
```diff
@@ -96,7 +96,7 @@ export async function isHotswappableAppSyncChange(
             delete sdkRequestObject.runtime;
           }
 
-          const { functions } = await sdk.appsync().listFunctions({ apiId: sdkRequestObject.apiId }).promise();
+          const functions = await getAppSyncFunctions(sdk, sdkRequestObject.apiId, undefined);
```
kaizencc, line 99:
```suggestion
          const functions = await getAppSyncFunctions(sdk, sdkRequestObject.apiId);
```

kaizencc (approved):

Issue #31378, opened 2024-09-09
cognito: Using a token for UserPoolIdentityProviderApple doesn't resolve at deploy


Someone5328546:
### Describe the bug

I'm attempting to create a Cognito user pool using Apple as an identity provider. I created the secret key and manually saved it in Secrets Manager using the AWS console.

Originally, I attempted this:

```typescript
const applePrivateKeySecret = Secret.fromSecretNameV2(this, 'ApplePrivateKey', 'TheKeyNameInSecretsManager')

const appleProvider = new UserPoolIdentityProviderApple(this, 'UserPoolIdentityProviderApple',
{
    clientId: 'com.myapp',
    teamId: '123456',
    keyId: '123456',
    privateKey: applePrivateKeySecret.secretValue.toString(),
    userPool: this.userPool
})
```

It didn't work on deploy since the secret is exposed as a string.

I also tried saving the value as a `SecureString` in AWS Systems Manager and retrieving it like this:

```typescript
const applePrivateKeyParameter = SecretValue.ssmSecure('AppleSignInPrivateKey').toString()
```

Which didn't give the same error regarding the value being exposed, but results in a `TypeError` (described just below), since apparently this token is unresolved when I attempt to deploy.

So I attempted to create a lambda that could be called by my Cognito stack to retrieve the token from Secrets Manager:

The lambda code:

```typescript
import {
    GetSecretValueCommand,
    SecretsManagerClient
} from '@aws-sdk/client-secrets-manager'

export const handler = async (event: any = {}) : Promise <any> =>
{
    try
    {
        const client = new SecretsManagerClient()
        const command = new GetSecretValueCommand({ SecretId: 'TheKeyNameInSecretsManager' })
        const response = await client.send(command)

        if (response.SecretString)
        {
            return {
                Status: 'SUCCESS',
                PhysicalResourceId: 'ApplePrivateKey',
                Data:
                {
                    SecretValue: response.SecretString
                }
            }
        }
        else
        {
            return {
                Status: 'FAILED',
                Reason: 'SecretString is empty.',
            }
        }
    }
    catch (error: any)
    {
        return {
            status: 'FAILED',
            message: error?.message || 'Unknown Failure.'
        }
    }
}
```

In my Cognito Stack, I retrieve the token and attempt to use it, but `UserPoolIdentityProviderApple` is seemingly `undefined` since the token never resolves.

This line will always throw an error:

```typescript
this.userPool.registerIdentityProvider(appleProvider)
```

[...]

PR #31409, opened 2024-09-11
fix(cognito): deprecate privateKey and add privateKeyValue as typed SecureValue

pahud:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/31378

### Reason for this change

1. `privateKey` was typed `string` which should be `SecureValue` just as [clientSecretValue](https://github.com/aws/aws-cdk/blob/1e203753519e10e19ef0db87e1382377b609bcaa/packages/aws-cdk-lib/aws-cognito/lib/user-pool-idps/google.ts#L28) in Google IdP. This PR deprecates `privateKey` and adds `privateKeyValue` with correct type.
2. `apple.ts` was named by mistake and it won't be unit tested. This PR renames it to `apple.test.ts` so it would be covered. Figured an existing test was failed, just fixed that failed one as well.

### Description of changes

- Add `privateKeyValue` property of type SecretValue to UserPoolIdentityProviderAppleProps
- Deprecate the existing `privateKey` string property
- Implement logic to ensure exactly one of `privateKey` or `privateKeyValue` is provided
- Update UserPoolIdentityProviderApple to use the new `privateKeyValue` when available
- Rename apple.ts test file to apple.test.ts for consistency
- Add new test case to verify mutually exclusive properties

Users must now provide either `privateKey` or `privateKeyValue`,
but not both. This change enhances security by allowing the use of SecretValue
for the Apple IDP private key.

### Checklist

----


packages/aws-cdk-lib/aws-cognito/test/user-pool-idps/apple.test.ts
```diff
@@ -90,8 +90,8 @@ describe('UserPoolIdentityProvider', () => {
         keyId: 'CDKKEYCDK1',
         privateKey: 'PRIV_KEY_CDK',
         attributeMapping: {
-          familyName: ProviderAttribute.APPLE_LAST_NAME,
-          givenName: ProviderAttribute.APPLE_FIRST_NAME,
+          familyName: ProviderAttribute.APPLE_FIRST_NAME,
```
pahud, line 93: This is not part of this PR but it won't pass if unchanged so we have to update it. I am not sure if relevant code has to be fixed but we don't fix that in this PR.

packages/aws-cdk-lib/aws-cognito/lib/user-pool-idps/apple.ts
```diff
@@ -44,6 +52,12 @@ export class UserPoolIdentityProviderApple extends UserPoolIdentityProviderBase
 
     const scopes = props.scopes ?? ['name'];
 
+    //at least one of the properties must be configured
```
GavinZZ, line 55:
```suggestion
    // Exactly one of the properties must be configured
```

packages/aws-cdk-lib/aws-cognito/test/user-pool-idps/apple.test.ts
GavinZZ: Instead of changing this, I think you should change the verification step. Family name is last name.

pahud: yes you are right!!

pahud:
Exemption Request

pahud:
@GavinZZ 

Unfortunately I don't have a valid private_key for that. And there's no existing integ test for that as well so I am offering unit tests only.

>8:07:08 PM | CREATE_FAILED        | AWS::Cognito::UserPoolIdentityProvider | AppleIdentityProvider6271AB3B
Provided private key cannot be used for Sign in with Apple. (Service: AWSCognitoIdentityProviderService; Status Code: 400; Error Code: InvalidParameterException;
Request ID: b28d3b10-14ca-42bc-a288-e1db34e7d6e3; Proxy: null)

GavinZZ (approved):

PR #31410, opened 2024-09-11
fix(cli): bootstrap respects qualifier from cdk.json

### LINKED ISSUE #28249: core: cdk.json "qualifier" ignored
### Describe the bug

Setting `qualifier` attribute in `cdk.json` appears to not have any impact on running `cdk bootstrap` as it continues to use the default qualifier.

### Expected Behavior

Setting the `qualifier` attribute in `cdk.json` should behave the same as running `cdk bootstrap --qualifier <QUALIFIER_VALUE>`.

### Current Behavior

Running `cdk bootstrap` does not inherit the `qualifier` value stored in `cdk.json`

### Reproduction Steps

Add a `qualifier` value to `cdk.json`:
```
{
  "app": "npx ts-node --prefer-ts-exts bin/app.ts",
  "versionReporting": false,
  "watch": {
    "include": [
      "**"
    ],
    "exclude": [
      "README.md",
      "cdk*.json",
      "**/*.d.ts",
      "**/*.js",
      "tsconfig.json",
      "package*.json",
      "yarn.lock",
      "node_modules",
      "test"
    ]
  },
  "context": {
    "@aws-cdk/core:bootstrapQualifier": "sample-app",
    "@aws-cdk/aws-apigateway:usagePlanKeyOrderInsensitiveId": true,
    "@aws-cdk/core:stackRelativeExports": true,
    "@aws-cdk/aws-rds:lowercaseDbIdentifier": true,
    "@aws-cdk/aws-lambda:recognizeVersionProps": true,
    "@aws-cdk/aws-cloudfront:defaultSecurityPolicyTLSv1.2_2021": true,
    "@aws-cdk-containers/ecs-service-extensions:enableDefaultLogDriver": true,
    "@aws-cdk/aws-ec2:uniqueImdsv2TemplateName": true,
    "@aws-cdk/aws-iam:minimizePolicies": true,
    "@aws-cdk/core:target-partitions": [
      "aws",
      "aws-cn"
    ],
    "@aws-cdk/core:newStyleStackSynthesis": true
  },
  "qualifier": "sample-app",
  "toolkitStackName": "sample-app-cdk-toolkit"
}
```

Run `cdk bootstrap` and observe resources being  created using the default qualifier.

### Possible Solution

Unkown, I'd assume the CLI should plug in the qualifier value from `cdk.json` if not explicitly provided as an option to the command.

_No response_

### CDK CLI Version

2.113.0 (build ccd534a)

### Framework Version

_No response_

### Node.js Version

v20.9.0

### OS

OSX 13.6 (22G120)

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

kaizencc:
closes #28249. 

The qualifier property can be set via the context key "@aws-cdk/core:bootstrapQualifier" and if omitted, the arbitrary default is `hnb659fds`. In the case of `cdk bootstrap`, there is an additional way to set the qualifier via the `--qualifier` CLI option. Specific to the `cdk bootstrap` logic, we currently only honor the command line argument, which is an error.

Ultimately, the following `cdk bootstrap` calls should be identical:

- `cdk bootstrap` with the following `cdk.json` file:

```json
{
  "@aws-cdk/core:bootstrapQualifier": "abcde",
}
```

- `cdk bootstrap --qualifier="abcde"`

I've made the decision that the `--qualifier` parameter takes precedent over the `cdk.json` context if they are both set.


mrgrain (changes requested):
> I've made the decision that the --qualifier parameter takes precedent over the cdk.json context if they are both set but I can see a reasonable argument for throwing an error too

@kaizencc That's usually how it's done. But to confirm, how do other flags work?

Also the issue expects `cdk.json` to work. But you say you are fixing it for `context.json`? Can you clarify please.

kaizencc:
I mean `cdk.json`, sorry. Although all places we add context goes into the same `Configuration` object so in theory it should work for both `cdk.json` and `cdk.context.json`.

kaizencc:
The only other place in cli.ts where I see two ways of describing the same thing is here:

```ts
case 'synth':
        const quiet = configuration.settings.get(['quiet']) ?? args.quiet;
```

and it is in fact defaulting to the opposite. But this line was written by a community contributor and it doesn't look like that decision was made with any intentionality (and in fact maybe was done to make testing easier). I think we should default to command line arguments before context values and would support switching the `quiet` logic around also.

mrgrain (approved):

Issue #22064, opened 2022-09-15
SSM: StringParameter.valueFromLookup should return undefined if SSM parameter not available


astef:
### Describe the feature

When using `ssm.StringParameter.valueFromLookup`, it crashes (not sure, but error is not catchable), when the parameter doesn't exist. It makes it unusable for scenarios with a fallback parameter.

### Use Case

For a short-lived infrastructure it's both important to be configurable (create a parameter & deploy), and have a nice fallback value (just deploy). SSM parameter is a perfect solution, since it can be resolved at synth time. It's very unfortunate, that currently it's impossible not to crash, when the value doesn't exist.

### Proposed Solution

Return undefined, instead of crashing, or expose a better API, with multiple fallback parameter names to check.

### Other Information

This issue was also discussed here: https://github.com/aws/aws-cdk/issues/7259

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### CDK version used

2

### Environment details (OS name and version, etc.)

Any

mascur:
Thank you for the ticket, I've labeled this as both a bug and feature-request since it seems to be a bit of both. One approach you could take would be to use a custom resource as mentioned here: https://github.com/aws/aws-cdk/issues/7259#issuecomment-612198184 to handle fallback values.

sammcj:
We just got hit by this as well, it certainly seems like a bug.

otaviomacedo:
@astef Just to clarify, when you say "it crashes", do you mean that the CLI prints an error message like this:

```
[Error at /MyStack] SSM parameter not available in account 12345689, region us-east-1: foo

Found errors
```

and returns exit code 1?

If that's the case, the template is still generated. To know whether the parameter exists at runtime, check whether the value starts with "dummy-value-for-".

andrew-glenn:
@otaviomacedo Thanks - but that doesn't solve the `exit code 1` problem when dealing with automation.

tyrauber:
This `exit code 1` is a problem for SST. 

It prevents conditionally creating resources by testing whether an ssm value exists.

Take a look as the [aws-cdk-lib ssm documentation #lookup-existing-parameters](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ssm-readme.html#lookup-existing-parameters).

```
const arnLookup = ssm.StringParameter.valueFromLookup(this, '/my/topic/arn');
[...]

Issue #7259, opened 2020-04-08
SSM StringParameter should query multiple paths


brettswift:
The ability to look up multiple SSM paths to satisfy a construct. 

### Use Case

In our applications we have a `../common/..` pathing and `../<namespace>/..`  pathing in ssm values, where `<namespace>` comes from a cdk context variable `-c namespace=bswift`. 

The pattern is pretty common, grab all the defaults for that account and if I want something special on my deploy - then just set that one value. 

This doesn't work with CDK as ValidationErrors are thrown.

### Proposed Solution

```
const value = ssm.StringParameter.valueFromLookup(this, 
     [
        '/service/common/value', 
        '/service/bswift/value'
    ]
  )
```
Or something to that effect. 

Currently neither `valueFromLookup` or `fromStringParameterName` allow me to do my own null checking so I'm not sure if there's a workaround internal to CDK.  The only thing I can think of is to do it outside and use a context parameter.  It would be nice if I could do this within the Construct.

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

MrArnoldPalmer:
@brettswift to help understand the use case better. You want the construct to accept a list of names and it looks for a parameter using each name in order until it finds one that is defined?

Do you use this pattern in cloudformation or other IAC tools currently? Would love to see some examples if so.

brettswift:
Hey, 
Your rephrase of my question is correct, however there might be a lower effort higher value solution. 

We do use this lookup strategy in our service API's so far.    I used to use puppet, where there was a hierarchy tool called hiera.  It's way too in depth to go into here, but pretty powerful.   Something similar to our puppet hierarchies was default behaviour based on linux vs windows.   In AWS that might be account level defaults for ASG size.  In our dev account that would be 1.   But if I want to do some testing on rolling updates, I might update my instance config that way.    It's a different mechanism than context parameters or hard coding it.    With context parameters it's hard to control all of our services at a global level - we'd have to go change defaults in every service. 


I could write a library to do this but think it'd be nice to have in CDK.  The tough part is we'd need to use the SDK, which is async.   So CDK helps us synchronize calls to that by giving us a .stringValue option.  (Actually I'm not sure.. maybe CDK uses the API? I haven't dug that deep yet). 

[...]

PR #31415, opened 2024-09-11
fix(ssm): update ssm-context to prevent raising an error on missing parameter

### LINKED ISSUE #7051: (aws-ssm): configure a default value instead of dummy-value-for-parameterName
My feature request is that it is useful to be able to return a configured default value from StringParameter.valueFromLookup() instead of dummy-value-for-parameterName.

### Use Case

CDK can lookup ssm parameters like this:

`    const value:string = ssm.StringParameter.valueFromLookup(this, 'keyName');`

It usually works well. However, the returned value may be like _dummy-value-for-keyName_ when no context value exist. In this case, if the returned value is used by some other functions like  Bucket.fromBucketArn, Topic.fromTopicArn, or etc, it causes an error like this:

`  ARNs must have at least 6 components: dummy-value-for-keyName`

It prevents from working cdk command well.
Based on the above case, I think it is useful to be able to configure the default value or it is good to return the actual ssm parameter value instead of a dummy value.

### Other

The testcase to reproduce this issue is like this:

1. Init cdk app.

```
 mkdir foo
 cd foo
 cdk init sample-app --language=typescript
```

2. Configure a default region and account.

Set env in bin/foo.ts 

```
#!/usr/bin/env node
import * as cdk from '@aws-cdk/core';
import { FooStack } from '../lib/foo-stack';

const myEnv = { account: 'aws account', region: 'a region'};

const app = new cdk.App();
new FooStack(app, 'FooStack', { env: myEnv });
```

3. And then edit lib/foo-stack.ts like this.

```
import * as cdk from '@aws-cdk/core';
import * as ssm from '@aws-cdk/aws-ssm';
import * as s3 from '@aws-cdk/aws-s3';

export class FooStack extends cdk.Stack {
  constructor(scope: cdk.App, id: string, props?: cdk.StackProps) {
    super(scope, id, props);

    const arnValue:string = ssm.StringParameter.valueFromLookup(this, 'keyName');
    const bucket = s3.Bucket.fromBucketArn(this, 'FooBucket', arnValue);
  }
}
```

4. Then run cdk synth, the application failed even though keyName exists.

```
$ cdk synth
ARNs must have at least 6 components: dummy-value-for-keyName
```

I tested it with cdk 1.31.0 .

The workaround of this issue is to comment out fromXxxxArn code in the above step3 to avoid the error of parsing the returned parameter(arnValue). After cdk synth command finish successfully, cdk.context.json is set. So, cdk synth command can success after uncomment the code because the context value is returned instead of a dummy value.
However, the issue reproduces again if you run cdk context --clear true .

This may be similar to https://github.com/aws/aws-cdk/issues/3654 though the issue was closed.
And I think dummy-value-for relates to the following code:
https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-ssm/lib/parameter.ts#L367

---

dsilbergleithcu-godaddy:
Updates StringParameter.valueFromLookup with an optional "defaultValue" When specified this value will be used:
 - in place of the standard dummyValue
 - to signal that an Error should not be raised during synthesis if the parameter is not found in the account.

Test are updated to prove that this works

### Issue # (if applicable)

Resolves #7051 

There are some closed issues which also benefit from this change:
- #22064 
- #7259

### Reason for this change

We have a library which has a fixed set of SSM parameters on which it depends.  The values from those parameters are made available as attributes of a custom Stack.  We have many users in many different AWS accounts, and not all of the parameters are guaranteed to exist.  This is okay.  In general, teams would simply not use those values and be happy with that outcome.  Unfortunately, CDK crashes when you look up an SSM parameter that does not exist in the account.  This is unacceptable.

### Description of changes

To address the issue described above, I implemented an optional parameter on the `valueFromLookup` method: `defaultValue`.  The idea is that if this value is specified, and we fail to look up a parameter in the account, we will return this value and suppress the Error that is currently raised when a parameter is not found.

To implement that functionality, I added a field to the `GetContextValueOptions` interface which is used to flag that we're not going to raise the error.  Then, in `valueFromLookup`, I set that flag to `true` if the `dummyValue` is specified.  `valueFromLookup` then calls `ContextProvider.getValue` passing along those values.

`ContextProvider.getValue` is modified so that when it calls `stack.reportMissingContextKey` it passes a modified set of `props` which include the `defaultValue` and the `ignoreErrorOnMissingContext` flag.

These finally land in the `aws-cdk` context provider for `ssm-parameter`.  That code has been updated so that if the value is not found in SSM, and we're told to suppress the error, then we'll simply return the `defaultValue` that was passed in. 

I added a unit tests which covers when the default value is set.  I also updated the original unit test as the `props` now contain some additional field.

I added an integration test which calls `valueFromLookup` with a `defaultValue` set and then confirms that no exception is raised and that `valueFromLookup` returned the `defaultValue`

**NOTE**
I considered that the changes made _might_ need to be a part of the `cloud-assembly-schema` but chose to work around that for now.  I'm open to incorporating them there if that's a more correct path.

**NOTE 2**
I'm unsure about how to update API documentation for this change.  This does alter the public API for `valueFromLookup` and the function doesn't appear to have a proper `TSDoc` header on it.  Please let me know if there's a proper way for me to update the documentation.

### Checklist

----


packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
```diff
@@ -574,12 +574,17 @@ export class StringParameter extends ParameterBase implements IStringParameter {
...
+   * if the SSM Parameter is not found in the account at synth time.
    */
-  public static valueFromLookup(scope: Construct, parameterName: string): string {
+  public static valueFromLookup(scope: Construct, parameterName: string, defaultValue?: string): string {
     const value = ContextProvider.getValue(scope, {
       provider: cxschema.ContextProvider.SSM_PARAMETER_PROVIDER,
       props: { parameterName },
-      dummyValue: `dummy-value-for-${parameterName}`,
+      dummyValue: defaultValue || `dummy-value-for-${parameterName}`,
+      ignoreErrorOnMissingContext: !!defaultValue,
```
rix0rrr, line 587:
```suggestion
      ignoreErrorOnMissingContext: defaultValue !== undefined,
```

If the default value is the empty string, this will turn it into `false`, but you probably expect it to be `true`.

rix0rrr (dismissed):
Looks good! One small thing...

packages/aws-cdk-lib/aws-ssm/lib/parameter.ts
dsilbergleithcu-godaddy: Good call.  And you're right, I do.

kaizencc:
sending thru our test pipeline will add the necessary labels + approve when its done

kaizencc (approved):

PR #31430, opened 2024-09-12
chore(kinesisfirehose-alpha): refactor encryption property to combine encryptionKey

paulhcsun:
### Reason for this change

The previous `encryption` and `encryptionKey` properties required error handling to enforce when an `encryptionKey` could be specified and when it was invalid (only valid when using `CUSTOMER_MANAGED_KEY`).

The properties should be combined to make this user experience more straightforward and only allow a KMS key to be passed in when using a customer-managed key. 

### Description of changes

BREAKING CHANGE: `encryptionKey` property is removed and `encryption` property type has changed from the `StreamEncryption` enum to the `StreamEncryption` class. 

To pass in a KMS key for the customer managed key case, use `StreamEncryption.customerManagedKey(key)`

#### Details
Replaced `encryption` and `encryptionKey` properties with a single property `encryption` of type `StreamEncryption` and is used by calling one of the 3 methods:
```ts
SreamEncryption.unencrypted()
StreamEncryption.awsOwnedKey()
StreamEncryption.customerManagedKey(key?: IKey)
```

This makes it so it's not longer possible to pass in a key when the encryption type is AWS owned or unencrypted. The `key` is an optional parameter in `StreamEncryption.customerManagedKey(key?: IKey)` so following the previous behaviour, if a key is provided it will be used, otherwise a key will be created for the user. 

Generated templates do not change so behaviour remains the same. 

Updated integ/unit tests. 

### Checklist

----


packages/@aws-cdk/aws-kinesisfirehose-alpha/lib/encryption.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    }) (StreamEncryptionType.CUSTOMER_MANAGED, encryptionKey);
+  }
+
+  /**
+   * Constructor for StreamEncryption.
+   *
+   * @param type The type of server-side encryption for the Kinesis Firehose delivery stream.
+   * @param encryptionKey Optional KMS key used for customer managed encryption.
+   */
```
paulhcsun, line 40: Added to deal with awslint errors:
```
error: [awslint:docs-public-apis:@aws-cdk/aws-kinesisfirehose-alpha.StreamEncryption.type] Public API element must have a docstring
error: [awslint:docs-public-apis:@aws-cdk/aws-kinesisfirehose-alpha.StreamEncryption.encryptionKey] Public API element must have a docstring
```

packages/@aws-cdk/aws-kinesisfirehose-alpha/lib/delivery-stream.ts
```diff
@@ -216,16 +217,16 @@ export interface DeliveryStreamProps {
...
-   *
-   * @default - no KMS key will be used; if `encryption` is set to `CUSTOMER_MANAGED`, a KMS key will be created for you
-   */
-  readonly encryptionKey?: kms.IKey;
+  // /**
+  //  * Customer managed key to server-side encrypt data in the stream.
+  //  *
+  //  * @default - no KMS key will be used; if `encryption` is set to `CUSTOMER_MANAGED`, a KMS key will be created for you
+  //  */
+  // readonly encryptionKey?: kms.IKey;
```
GavinZZ, line 229: Is this intended (commented out)?

packages/@aws-cdk/aws-kinesisfirehose-alpha/lib/encryption.ts
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Configure server-side encryption using customer managed keys.
+   *
+   * @param encryptionKey the KMS key for the delivery stream.
+   */
+  public static customerManagedKey(encryptionKey?: IKey): StreamEncryption {
+    return new (class extends StreamEncryption {
+
+    }) (StreamEncryptionType.CUSTOMER_MANAGED, encryptionKey);
+  }
```
GavinZZ, line 33: Why not something like this?
```
export interface StreamEncryption {
  encryptionKey?: string;
  type: StreamEncryptionType
}

/**
 * Options for server-side encryption of a delivery stream.
 */
export enum StreamEncryptionType {
  /**
   * Data in the stream is stored unencrypted.
   */
  UNENCRYPTED,

  /**
   * Data in the stream is stored encrypted by a KMS key managed by the customer.
   */
  CUSTOMER_MANAGED,

  /**
   * Data in the stream is stored encrypted by a KMS key owned by AWS and managed for use in multiple AWS accounts.
   */
  AWS_OWNED,
}
```

packages/@aws-cdk/aws-kinesisfirehose-alpha/test/delivery-stream.test.ts
```diff
@@ -151,12 +152,12 @@ describe('delivery stream', () => {
...
   });
 
-  test('not providing role but specifying encryptionKey creates two roles', () => {
+  test('not providing role but using customerManagedKey encryption with a key creates two roles', () => {
     const key = new kms.Key(stack, 'Key');
 
     new firehose.DeliveryStream(stack, 'Delivery Stream', {
       destinations: [mockS3Destination],
-      encryptionKey: key,
+      encryption: StreamEncryption.customerManagedKey(key),
```
GavinZZ, line 160: This pattern feels a bit off to me. `StreamEncryption` usage sounds like a ENUM to me but the actual usage is calling a statis method.

GavinZZ (commented):
What's the purpose for this change? Is there an issue about the current usage?

packages/@aws-cdk/aws-kinesisfirehose-alpha/lib/delivery-stream.ts
paulhcsun: Forgot to remove that, will do that now.

packages/@aws-cdk/aws-kinesisfirehose-alpha/test/delivery-stream.test.ts
paulhcsun: This is the same pattern that's being used for TableV2's encryption [here](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-dynamodb/README.md#encryption).

paulhcsun:
@GavinZZ the main issue with the current usage is that it requires the user to know when it is appropriate to pass in an `encryptionKey` depending on which `encryptionType` they used. 

i.e. passing in an `encryptionKey` is only valid when the `encryptionType === CUSTOMER_MANAGED_KEY` and would throw an error if the key were passed in with `encryptionType` set to `AWS_OWNED` or `UNENCRYPTED`. This change makes it so that then the `encryptionType` is  `AWS_OWNED` or `UNENCRYPTED` the user does not have an option to pass in a key which will remove the need for that error handling and remove the need to the user to know the restrictions.

packages/@aws-cdk/aws-kinesisfirehose-alpha/lib/encryption.ts
paulhcsun: That would also work but is slightly less preferred because of it's nested nature (although at one level of nesting it's still fine). Using this class design follows the existing pattern in [TableV2's encryption](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-dynamodb/lib/encryption.ts) and also opens up the option to add private methods to set additional values in the future.

GavinZZ (approved):

PR #31431, opened 2024-09-12
fix(s3): unable to update the s3 event notifications on an existing S3 bucket

### LINKED ISSUE #31303: s3-notifications: Unable to update the s3 event notifications on an existing S3 bucket.
### Describe the bug

I am facing an error when trying to update the s3 event notification configuration on an existing S3 bucket using CDK.
Created a stack to add s3 event notification on an existing s3 bucket. Create operation goes through successfully.
When I update the cdk stack to add event notifications on the same s3 bucket, the Update operation fails with below error.

**Error**:
```
Received response status [FAILED] from custom resource. Message returned: Error: An error occurred (InvalidArgument) when calling the PutBucketNotificationConfiguration operation: Configuration is ambiguously defined. Cannot have overlapping suffixes in two rules if the prefixes are overlapping for the same event type.. See the details in CloudWatch Log Stream:
```

Here the [lambda code](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py#L41) backing the BucketNotifications custom resource is unable to identify that the existing event notifications were also managed by the same cdk stack thereby creating duplicate notifications resulting in error.

### Regression Issue

_No response_

### Expected Behavior

Should be able to add additional notification configuration on the existing s3 bucket.

### Current Behavior

**Issue**
Created a cdk stack and added an event notification on an existing s3 bucket. Create operation is successful.
Updated the same stack to add a new event notification along with the existing ones. The update operation fails with below error.

**Error**:
```
Received response status [FAILED] from custom resource. Message returned: Error: An error occurred (InvalidArgument) when calling the PutBucketNotificationConfiguration operation: Configuration is ambiguously defined. Cannot have overlapping suffixes in two rules if the prefixes are overlapping for the same event type.. See the details in CloudWatch Log Stream:
```

### Reproduction Steps

Steps:
    1. Create a stack by commenting out the 2nd event notification (notificationfilter2)
    2. once stack is created, update the stack after uncommenting the 2nd event notification.
```


 my_lambda = _lambda.Function(
          self, 'HelloHandler',
            runtime=_lambda.Runtime.PYTHON_3_8,
            code=_lambda.Code.from_asset('lambda'),
            handler='hello.handler',
        )  


        bucket = _s3.Bucket.from_bucket_name(self, "Bucket", "bucketname")
        notification = _s3_notify.LambdaDestination(my_lambda)

        notificationfilter1 = _s3.NotificationKeyFilter(prefix="foo/", suffix="bar/")
        bucket.add_event_notification(_s3.EventType.OBJECT_CREATED, notification, notificationfilter1
        )

       # notificationfilter2 = _s3.NotificationKeyFilter(prefix="fo1/",suffix="ba1/",)
      # bucket.add_event_notification(_s3.EventType.OBJECT_CREATED, notification, notificationfilter2
      #  )
```

### Possible Solution

**Analysis details**:
Create a cdk stack to add below event notification to an existing s3 bucket. The Create operation is successful.
```
'Events': ['s3:ObjectCreated:*'], 'Filter': {'Key': {'FilterRules': [{'Name': 'Prefix', 'Value': 'bar/'}, {'Name': 'Suffix', 'Value': 'foo/'}]}}}
```
In the [ notifications-resource-handler](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/custom-resource-handlers/lib/aws-s3/notifications-resource-handler/index.py#L41) code

For unmanaged buckets, there is a get_id function used to evaluate the hash for each event notification to confirm if this is created by the stack or is an existing external configuration:
```
def get_id(n):
    n['Id'] = ''
    strToHash=json.dumps(n, sort_keys=True).replace('"Name": "prefix"', '"Name": "Prefix"').replace('"Name": "suffix"', '"Name": "Suffix"')
    return f"{stack_id}-{hash(strToHash)}"
```
During creation, this goes fine as it treats all existing configurations as external.
Then it appends incoming + external and creates the final notification configuration as expected.

During an update operation, the lambda code will first get the existing event notifications on the s3 bucket.
Then it validates if the existing event notifications matches the notification in the incoming request from Cloudformation.
It does this by evaluating the hash for each existing event notification and validating if this matches with the hash of the incoming event notifications from Cloudformation. 
When there is a match, it identifies this existing event configuration as managed by the stack. 
This helps in eliminating the duplicates.

There is an issue with this hash evaluation, due to a change in order of the prefix and suffix within the filter rules between the existing notification and the notification coming from Cloudformation.

Sample event notification from Cloudformation request
```
{
    "Events": [
        "s3:ObjectCreated:*"
    ],
    "Filter": {
        "Key": {
            "FilterRules": [
                {
                    "Name": "Suffix",
                    "Value": "foo/"
                },
                {
                    "Name": "Prefix",
                    "Value": "bar/"
                }
            ]
        }
    },
    "Id": "",
    "LambdaFunctionArn": "arn:aws:lambda:<aws-region>:<aws-account-id>:function:<FunctionName>"
}
```

Sample existing S3 event notification:
```
    "Events": [
        "s3:ObjectCreated:*"
    ],
    "Filter": {
        "Key": {
            "FilterRules": [
                {
                    "Name": "Prefix",
                    "Value": "bar/"
                },
                {
                    "Name": "Suffix",
                    "Value": "foo/"
                }
            ]
        }
    },
    "Id": "",
    "LambdaFunctionArn": "arn:aws:lambda:<aws-region>:<aws-account-id>:function:<FunctionName>"
}
```

Note: Even though the content of the above jsons are the same, the order of the filter rules (prefix and suffix) are different. 
Due to this, the hash evaluated will also be different. So, it is unable to identify that the existing configuration was also added by the stack itself.
This ends up in a configuration which includes the duplicates (existing configurations on s3 + incoming event configuration from Cloudformation stack)
resulting in below error:
```
Received response status [FAILED] from custom resource. Message returned: Error: An error occurred (InvalidArgument) when calling the PutBucketNotificationConfiguration operation: Configuration is ambiguously defined. Cannot have overlapping suffixes in two rules if the prefixes are overlapping for the same event type.
```
This can be workaround by changing the order of filters in the cdk synthesized stack template, by passing the prefix first and then suffix within the filter rules array.

_No response_

### CDK CLI Version

2.152.0

### Framework Version

_No response_

### Node.js Version

v20.16.0

### OS

Linux

### Language

Python

### Language Version

_No response_

### Other information

_No response_

xazhao:
### Issue # (if applicable)

Closes #31303.

### Reason for this change

During the update S3 bucket notifications, if there are multiple filter rules involved, the order of filter rules is different from CFN event notifications and S3 get notifications call. This will result different hash value which makes CDK unable to recognize s3 notifications created by the stack itself.

### Description of changes
Sort the notifications filter rules by key when calculating the hash value.

Manually tested and integration tests.

### Checklist

----


xenonlcs:
Hello everyone, here are my 2 cents. THIS IS NOT RELATED TO SORT PROBLEMS (or not only)

In CDK v2.3 the method used to verify if a notification configuration was external was based on the following check (this is from the Custom Resource behind aws_s3_notifications.LambdaDestination resource):
external_notifications[t] = [n for n in existing_notifications.get(t, []) if not n['Id'].startswith(f"{stack_id}-")]

In CDK v2.136 the method is this one (not checked when this change was introduced):
ids = [with_id(n) for n in old.get(t, [])] old_incoming_ids = [n['Id'] for n in ids] external_notifications[t] = [n for n in existing_notifications.get(t, []) if not n['Id'] in old_incoming_ids] 

The function that with_id(n) is this one:
def with_id(n): n['Id'] = f"{stack_id}-{hash(json.dumps(n, sort_keys=True))}" return n

The built-in function hash does not return the same value across different instances of python even if the input is the same.

The Custom Resource will treat as external_notifications even the ones that are not, leading to duplicates

GavinZZ (approved):
I think the changes LGTM, thanks for fixing the issue.

GavinZZ:
I approved the PR but added a `do-not-merge` as I think @shikha372  is also reviewing the PR. Would wait for her reviews.

shikha372:
Thanks @GavinZZ , discussed with @xazhao , we would like to confirm one that its not affecting anything in custom resources that are related to this and also test out a successful update scenario before merging.
Also, from the analysis, i really think we should check there will be no impact of this sorting to the external customers anything that causes replacement

Here's an example to illustrate a potential regression:

```typescript
const bucket = new s3.Bucket(this, 'MyBucket');

// Existing notification configuration
bucket.addEventNotification(
  s3.EventType.OBJECT_CREATED,
  new s3n.LambdaDestination(myLambda),
  { prefix: 'uploads/', suffix: '.jpg' }
);

// After deploying with the old logic, the notification has a specific hash identifier

// When updating the stack with the new logic:
// The same notification might be perceived as different due to the new sorting method
// CDK might attempt to remove the "old" notification and create a "new" one
// This could lead to temporary loss of notifications or duplicate notifications
```

shikha372 (approved):

PR #31432, opened 2024-09-12
chore(dynamodb): update waitForReplicationToFinish flag summary doc to highlight that it may cause deletion for existing replicas.

moelasmar:
### Description of changes

Add a warning message to the summary doc of the flag `waitForReplicationToFinish` to highlight that misusing this flag may cause deletion for existing replicas. This change is required to resolve some internal feedback as some people may miss reading the detail documentation of that flag, and so they will may misuse it and then have some serious issues.

see the below image for the updated doc:

Screenshot 2024-09-12 at 3 42 22 PM

### Checklist

----

### DIFF
--- a/packages/aws-cdk-lib/aws-dynamodb/lib/table.ts
+++ b/packages/aws-cdk-lib/aws-dynamodb/lib/table.ts
@@ -347,6 +347,7 @@ export interface TableOptions extends SchemaOptions {
   readonly replicationTimeout?: Duration;
 
   /**
+   * [WARNING: Use this flag with caution, misusing this flag may cause deleting existing replicas, refer to the detailed documentation for more information]
    * Indicates whether CloudFormation stack waits for replication to finish.
    * If set to false, the CloudFormation resource will mark the resource as
    * created and replication will be completed asynchronously. This property is


packages/aws-cdk-lib/aws-dynamodb/lib/table.ts
```diff
@@ -347,6 +347,7 @@ export interface TableOptions extends SchemaOptions {
   readonly replicationTimeout?: Duration;
 
   /**
+   * [WARNING: Use this flag with caution, misusing this flag may cause deleting existing replicas, refer to the detailed documentation for more information]
```
paulhcsun, line 350: Do the square brackets do something to format this in the dosctring? If not should we remove them like on like 356?

moelasmar: to update the summary doc, either we need a paragraph of 20 words, or one sentence that follow this pattern https://github.com/aws/jsii/blob/37082de53576b9c807e83506af336458c6f2c6cb/packages/jsii/lib/docs.ts#L259-L261

that is why I added square brackets here to separate between the warning message, and the actual definition of this flag. I am open for better ideas :)

paulhcsun: Discussed with @moelasmar and the updated summary line in the table looks good. This is just a workaround to ensure that the warning and the original summary line both show up without the use of any punctuation which would cause the original summary line to be cut off.

paulhcsun (approved):

PR #31433, opened 2024-09-12
fix(lambda): invalid Version object created from Version.fromVersionArn

GavinZZ:
### Issue # (if applicable)

Closes https://github.com/aws/aws-cdk/issues/29813

### Reason for this change

The current `lambda.Version.fromVersionArn` will incorrectly create the lambda function arn with the version qualifier. This is incorrect behaviour and cause CFN deployment failures.

### Description of changes

If the version arn is an unresolved token, use intrinsics to join the first 7 components. Otherwise, split the array and join the first 7 components directly.

All new and existing tests should pass. New integration tests added with assertions.

### Checklist

----


paulhcsun (approved):
Changes here look good to me and should not run into the same issue as the previous fix. Will approve for now but wait for a second opinion before we merge.

moelasmar (approved):

PR #31434, opened 2024-09-12
fix(ec2): fixing vpc endpoint pattern for ecr and ecr docker

shikha372:
### Issue # (if applicable)

Closes #.
Internal Tracking ticket P150271569

### Reason for this change

Expected Endpoints for ECR in some isolated regions are as below

gov.ic.hci.csp.us-isof-name.ecr.api,
gov.ic.hci.csp.us-isof-name.ecr.dkr,
uk.adc-e.cloud.eu-isoe-name.ecr.api,
 uk.adc-e.cloud.eu-isoe-name.ecr.dkr,

### Description of changes

As discussed with the ECR Service team, endpoints for the service are being generated in reverse order of the domain suffix.
Since some of the endpoints for other services are still using `com.amazonaws` , added fix only for the partitions and service(ECR) flagged.
 
Cannot do for cn regions on the basis of suffix as both regions have different services under exceptions.

Added unit test for validation of endpoint, keeping the region names as `us-isoe-test-1` as the regions are in build stage and could be confidential.

No changes to integration test as the fix is for isolated regions.

### Checklist

----


shikha372:
Exemption Request : No changes to integration test as the fix is for isolated regions.

packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts
```diff
@@ -660,8 +661,21 @@ export class InterfaceVpcEndpointAwsService implements IInterfaceVpcEndpointServ
         'redshift', 'redshift-data', 's3', 'sagemaker.api', 'sagemaker.featurestore-runtime', 'sagemaker.runtime', 'securityhub',
         'servicecatalog', 'sms', 'sqs', 'states', 'sts', 'sync-states', 'synthetics', 'transcribe', 'transcribestreaming', 'transfer',
         'workspaces', 'xray'],
+      'us-isof-': ['ecr.api', 'ecr.dkr'],
+      'eu-isoe-': ['ecr.api', 'ecr.dkr'],
     };
-    if (VPC_ENDPOINT_SERVICE_EXCEPTIONS[region]?.includes(name)) {
+
+    const regionPartition = region.split('-').slice(0, 2).join('-');
+    const partitionDetails = PARTITION_MAP[`${regionPartition}-`];
```
GavinZZ, line 669: Does this work for commercial regions? The map seems to use `default` for cmmercial regions, but your `regionPartition` won't return `default` for commercial regions.

packages/aws-cdk-lib/aws-ec2/test/vpc-endpoint.test.ts
```diff
@@ -694,6 +694,33 @@ describe('vpc endpoint', () => {
       });
 
     });
+
+    test.each([
+      ['us-isof-test-1', 'gov.ic.hci.csp'],
+      ['eu-isoe-test-1', 'uk.adc-e.cloud'],
+    ])('test vpc interface endpoint for ECR can be created correctly in isolated regions', (region : string, domain: string) => {
```
GavinZZ, line 701: can we add tests for commercial regions, cn regions, govcloud regions?

GavinZZ (changes requested):

packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts
shikha372: this is not for commercial regions, only for isolated regions that too only for the ones raised in the request with expected endpoint, as don't want to impact any other region . For regions/services other than the one defined in exception list it will be `com.amazonaws` i.e commercial region too

packages/aws-cdk-lib/aws-ec2/test/vpc-endpoint.test.ts
shikha372: sure, though we already had tests for cn regions below this, added it here too.

packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts
```diff
@@ -660,8 +661,21 @@ export class InterfaceVpcEndpointAwsService implements IInterfaceVpcEndpointServ
         'redshift', 'redshift-data', 's3', 'sagemaker.api', 'sagemaker.featurestore-runtime', 'sagemaker.runtime', 'securityhub',
         'servicecatalog', 'sms', 'sqs', 'states', 'sts', 'sync-states', 'synthetics', 'transcribe', 'transcribestreaming', 'transfer',
         'workspaces', 'xray'],
+      'us-isof-': ['ecr.api', 'ecr.dkr'],
+      'eu-isoe-': ['ecr.api', 'ecr.dkr'],
```
GavinZZ, line 665: Do we only need this two here? What about other us gov regions?

packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts
GavinZZ: I see, that makes sense. THanks

packages/aws-cdk-lib/aws-ec2/lib/vpc-endpoint.ts
shikha372: these two are the only ones flagged in ticket, so for now only these two. These are isolated, for US Gov region i think it starts with `us-gov`

GavinZZ (approved):

PR #31439, opened 2024-09-13
feat(lambda-event-sources): starting position timestamp for kafka

### LINKED ISSUE #31808: (lambda-event-sources): (starting position timestamp for kafka)
### Describe the feature

Allow user to start consuming Kafka Event Sources from a specific point in time: https://aws.amazon.com/about-aws/whats-new/2023/06/aws-lambda-starting-timestamp-kafka-sources/

### Use Case

???

### Proposed Solution

Add the `startingPositionTimestamp` to `ManagedKafkaEventSource` and `SelfManagedKafkaEventSource`.

### Other Information

_No response_

### Acknowledgements

### CDK version used

v2.162.0

macOS 15.0 (24A335)

nikovirtala:
### Issue # (if applicable)

Closes #31808

### Reason for this change

It was impossible to start consuming a Kafka topic from a specific point in time.

### Description of changes

The user may now set `startingPositionTimestamp` to `ManagedKafkaEventSource` and `SelfManagedKafkaEventSource` to start consuming a Kafka topic from a specific point in time. 

[Lambda and CloudFormation have supported the functionality for a while](https://aws.amazon.com/about-aws/whats-new/2023/06/aws-lambda-starting-timestamp-kafka-sources/), and other stream event sources like `KinesisEventSource` already had a CDK implementation for it. So, technically, this change is doing nothing new; it is only repeating the pattern that has already been proven to work on other sources.

The change is tested by adding similar unit tests to other event sources supporting this functionality.

### Checklist

----


moelasmar:
could you please add a new GH issue as a Feature request, and link it to this issue.

moelasmar (dismissed):
IMO, this is a feature and not a chore, and it needs update in Readme, and new integration test case

sumupitchayan:
> IMO, this is a feature and not a chore, and it needs update in Readme, and new integration test case

Agreed with @moelasmar, this seems like a `feat` to me - @nikovirtala can you please add an integ test and README documentation on this?

sumupitchayan (commented):
Agreed with @moelasmar, this seems like a feat to me - @nikovirtala can you please add an integ test and README documentation on this?

nikovirtala:
> 2. testing Kafka Event Source Mapping requires creating an MSK cluster or such, which is not as trivial as creating a simple resource like DynamoDB Table (I hope there is something I can borrow from the MSK module, if one exist)

Ok, [the MSK module is in alpha state](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-msk-alpha-readme.html), and even that doesn't have an integration test that would create a topic and messages to the topic — which would be required to test this feature 😞 

I doubt that creating all that is too much for me.

nikovirtala:
It seems there are is zero integration tests in this module — so no one can change the given module before these is one?

packages/aws-cdk-lib/aws-lambda-event-sources/lib/kafka.ts
```diff
@@ -55,6 +55,13 @@ export interface KafkaEventSourceProps extends BaseStreamEventSourceProps {
    * @default - discarded records are ignored
    */
   readonly onFailure?: lambda.IEventSourceDlq;
+
+  /**
+   * The time from which to start reading, in Unix time seconds.
+   *
+   * @default - no timestamp
+   */
+  readonly startingPositionTimestamp?: number;
```
moelasmar, line 64: a minor comment, I know that we already use number as a type for this property in Kinesis, but I was thinking that it is better for customers to use a real time stamp here, and we do calculate the epoch number on behalf of the customers.

I can accept to have this property to keep it aligned with Kinesis, but I prefer to add another property that use Time stamp instead of number .. what do you think?

moelasmar:
sorry @nikovirtala for not coming back to this PR. I see that you already handled the integ testing point. I left one minor comment.

packages/aws-cdk-lib/aws-lambda-event-sources/lib/kafka.ts
nikovirtala: > a minor comment, I know that we already use number as a type for this property in Kinesis, but I was thinking that it is better for customers to use a real time stamp here, and we do calculate the epoch number on behalf of the customers.
> 
> I can accept to have this property to keep it aligned with Kinesis, but I prefer to add another property that use Time stamp instead of number .. what do you think?

Well, our current internal implementation of this very same functionality is using `Date` but I chose to use the number here to keep the consistency. I will take a look if I can find a nice way to get them both and we can then choose which looks/feels the best. 

*If* we end up to improve the API here, I can then implement those changes also to the Kinesis source (naturally on another pull request).

nikovirtala:
> sorry @nikovirtala for not coming back to this PR. I see that you already handled the integ testing point. I left one minor comment.

No worries! I was stuck in thinking that the integration test requirement meant something else than it actually does — I had set the bar "a little" too high in my mind 😅

packages/aws-cdk-lib/aws-lambda-event-sources/lib/kafka.ts
moelasmar: thanks @nikovirtala

packages/aws-cdk-lib/aws-lambda-event-sources/lib/kafka.ts
```diff
@@ -55,6 +55,13 @@ export interface KafkaEventSourceProps extends BaseStreamEventSourceProps {
    * @default - discarded records are ignored
    */
   readonly onFailure?: lambda.IEventSourceDlq;
+
+  /**
+   * The time from which to start reading. Unix time seconds or `Date`
+   *
+   * @default - no timestamp
+   */
+  readonly startingPositionTimestamp?: number | Date;
```
nikovirtala, line 64: @moelasmar, here is one way to achieve it but is it very idiomatic to do it that way in CDK?

moelasmar: in CDK we do not accept the union types in L2 any more. Personally I was thinking on replacing the type to be TimeStamp instead of number, but to be align with kinesis, I think we can add another property, and let customers choose which one to use, and for sure they can only use one of them.

nikovirtala: After digging into how this kind of property type change should happen in CDK, I found out that there is no clearly defined convention for it. There is a long discussion, e.g., on [Enhanced L1s RFC](https://github.com/aws/aws-cdk-rfcs/pull/657) about using the `_v2` pattern, but no consensus on going with it.

So, I suggest that we merge this like this:

> I can accept to have this property to keep it aligned with Kinesis

and worry about the API later once there is a clear convention on how to proceed.

The changes I made to test the union type are now reverted.

packages/aws-cdk-lib/aws-lambda-event-sources/lib/kafka.ts
```diff
@@ -148,6 +155,15 @@ export class ManagedKafkaEventSource extends StreamEventSource {
...
   constructor(props: ManagedKafkaEventSourceProps) {
     super(props);
+
+    if (props.startingPosition === lambda.StartingPosition.AT_TIMESTAMP && !props.startingPositionTimestamp) {
+      throw new Error('startingPositionTimestamp must be provided when startingPosition is AT_TIMESTAMP');
+    }
+
+    if (props.startingPosition !== lambda.StartingPosition.AT_TIMESTAMP && props.startingPositionTimestamp) {
+      throw new Error('startingPositionTimestamp can only be used when startingPosition is AT_TIMESTAMP');
+    }
```
samson-keung, line 165: I'd suggest giving warnings instead of throw errors on these validations so that the following use cases become possible:
- when customers use Tokens for `props.startingPosition` and/or `props.startingPosition` (although probably a rare case)
- when customers use escape hatches (e.g. using `addPropertyOverride` method) to set either of the props after initializing the construct

Same comment on the other validations that throws error (line 260 - 268).

samson-keung (changes requested):
Left a small comment, then I think it will be good to merge!

packages/aws-cdk-lib/aws-lambda-event-sources/test/kafka.test.ts
```diff
@@ -407,6 +407,48 @@ describe('KafkaEventSource', () => {
           },
         }))).toThrow(/Minimum provisioned pollers must be less than or equal to maximum provisioned pollers/);
     });
+    test('AT_TIMESTAMP starting position', () => {
```
shikha372, line 410:
```suggestion
    test('Setting startingPositionTimestamp for kafka event ', () => {
```

packages/aws-cdk-lib/aws-lambda-event-sources/test/kafka.test.ts
```diff
@@ -1186,5 +1228,56 @@ describe('KafkaEventSource', () => {
           },
         }))).toThrow(/Minimum provisioned pollers must be less than or equal to maximum provisioned pollers/);
     });
+
+    test('AT_TIMESTAMP starting position', () => {
```
shikha372, line 1232:
```suggestion
    test('Setting startingPositionTimestamp for kafka event source ', () => {
```

packages/aws-cdk-lib/aws-lambda-event-sources/test/kafka.test.ts
```diff
@@ -1186,5 +1228,56 @@ describe('KafkaEventSource', () => {
...
+
+      expect(() => new sources.SelfManagedKafkaEventSource({
+        bootstrapServers,
+        secret: secret,
+        topic: kafkaTopic,
+        startingPosition: lambda.StartingPosition.AT_TIMESTAMP,
+      })).toThrow(/startingPositionTimestamp must be provided when startingPosition is AT_TIMESTAMP/);
+    });
+
+    test('startingPositionTimestamp without AT_TIMESTAMP throws error', () => {
```
shikha372, line 1268:
```suggestion
    test('setting timestamp without  startingPosition as AT_TIMESTAMP throws error', () => {
```

packages/aws-cdk-lib/aws-lambda-event-sources/test/kafka.test.ts
```diff
@@ -407,6 +407,48 @@ describe('KafkaEventSource', () => {
...
+      }),
+      );
+
+      Template.fromStack(stack).hasResourceProperties('AWS::Lambda::EventSourceMapping', {
+        StartingPosition: 'AT_TIMESTAMP',
+        StartingPositionTimestamp: 1640995200,
+      });
+    });
+
+    test('startingPositionTimestamp missing throws error', () => {
```
shikha372, line 430:
```suggestion
    test('missing startingPositionTimestamp with startingPosition set to AT_TIMESTAMP throws error', () => {
```

packages/aws-cdk-lib/aws-lambda-event-sources/test/kafka.test.ts
```diff
@@ -407,6 +407,48 @@ describe('KafkaEventSource', () => {
...
+      const kafkaTopic = 'some-topic';
+
+      expect(() => new sources.ManagedKafkaEventSource({
+        clusterArn,
+        topic: kafkaTopic,
+        startingPosition: lambda.StartingPosition.AT_TIMESTAMP,
+      })).toThrow(/startingPositionTimestamp must be provided when startingPosition is AT_TIMESTAMP/);
+    });
+
+    test('startingPositionTimestamp without AT_TIMESTAMP throws error', () => {
```
shikha372, line 441:
```suggestion
    test('setting timestamp without setting startingPosition to AT_TIMESTAMP throws error', () => {
```

shikha372:
Thank you @nikovirtala for contributing this PR, please let us know if you're still interested continue with this feature implementation, to prevent auto closing this PR.

packages/aws-cdk-lib/aws-lambda-event-sources/lib/kafka.ts
nikovirtala: Warning? Can you show me what is the convention of warning in CDK? — I think I have never dealt with one here.

It is throwing an error because the resulting CloudFormation template won't work if these conditions are not met.

nikovirtala:
> Thank you @nikovirtala for contributing this PR, please let us know if you're still interested continue with this feature implementation, to prevent auto closure.

I have a lot of interest in getting this feature available but I don't know if I have time to deal with the constantly changing "requirements" from changing reviewers — it should be enough to get things merged if currently used conventions are followed.

packages/aws-cdk-lib/aws-lambda-event-sources/lib/kafka.ts
shikha372: Hi @nikovirtala , warnings can be added to constructs using [Annotations](https://github.com/aws/aws-cdk/blob/b9cb47c4dbaacd839ddfdc38febdc58260d20c57/packages/aws-cdk-lib/core/lib/annotations.ts#L9)

shikha372:
Thank you @nikovirtala for your contribution, all comments are addressed now, will work on getting this merged in.

shikha372 (approved):

Issue #15947, opened 2021-08-09
(core): cloudformation spec adds a Tags property to an existing resource, this can break customer deployments or cause unexpected changes


madeline-k:
If a CDK user has added Tags to a construct, those tags will be added to all cloudformation resources within that construct's scope. If the Tags property was added to a resource via a cloudformation spec update, then when the CDK user updates their CDK version, a new Tags property will be added to that resource. This might not be what the user intended, and can cause deployment failures in some cases. 

### Reproduction Steps
Repro steps for one incarnation of this issue.

1. Deploy below sample with CDK version 1.113.0 or below.
```
export class CodeDeploy extends cdk.Construct {
  constructor(scope: cdk.Stack, id: string, props: cdk.StackProps) {
    super(scope, id)
    
    const fn = new lambda.Function(this, 'MyLambda', {
      code: new lambda.InlineCode('foo'),
      handler: 'index.handler',
      runtime: lambda.Runtime.NODEJS_10_X,
    });
    const alias = new lambda.Alias(this, "alias", { aliasName: "Prod", version: fn.currentVersion })

    const deployment = new codedeploy.LambdaDeploymentGroup(this, "deployment-group", {
      alias,
    })

    // This adds a 'Name' tag to all cloudformation resources in the `deployment` construct's scope.
    cdk.Tags.of(deployment).add('Name', `buffer-${props.environment}`)
  }
}
```

2. Upgrade to CDK version 1.114.0 or greater.

### What did you expect to happen?

Deployment succeeds after upgrade.

### What actually happened?

Deployment failed with error "Update to resource type AWS::CodeDeploy::Application is not supported" after upgrade, because the "Name" tag was added to the CodeDeploy::Application resource inside the `LambdaDeploymentGroup` construct.

### Workaround

Explicitly remove tags from being added to specific resource types within your constructs.

```
const tagOptions = {
  excludeResourceTypes: ['AWS::CodeDeploy::Application'],
};
cdk.Tags.of(deployment).add('Name', `buffer-${props.environment}`, tagOptions);
```

--- 

This is :bug: Bug Report

rix0rrr:
The solution to this will probably involve some feature flag. I don't see how else it could work.  I'm thinking:

```
{
  "context": {
    "@aws-cdk/core.taggableResources": "1.115.0"
  }
}
```

Which will only apply tags to resources that were considered taggable in 1.115.0, or something like that. 

[...]

PR #31443, opened 2024-09-13
fix(tags): excludeResourceTypes does not work for tags applied at stack level (under feature flag)

### LINKED ISSUE #28017: Trigger: Token strings not resolving
### Describe the bug

Hi Team,

Just opening this ticket where the "Token" strings are not resolving during deployment and stays as "Token" strings `${Token[TOKEN.664]}`. This results in the error below:

```
 "3 validation errors detected: Value '${Token[TOKEN.664]}' at 'tags.1.member.value' failed to satisfy constraint: Member must satisfy regular expression pattern: [\p{L}\p{Z}\p{N}_.:/=+\-@]*;
```

The error above is thrown against the "AWS::IAM::Role" which the `Trigger` construct creates because the value of its tag is a token string `${Token[TOKEN.664]}` which did not resolve during deployment and is not a valid value for a Tag.

For context, the setup here is that stack-level tags are being applied to the resources in the CDK stack (see code below) where its Tag value is retrieved from a resolvable token SSM parameter.

```
    const stringValue = ssm.StringParameter.fromStringParameterAttributes(this, 'MyValue', {
      parameterName: 'StringParameter',
      // 'version' can be specified but is optional.
    }).stringValue;

    const func = new lambda.Function(this, 'MyFunction', {
      handler: 'index.handler',
      code: lambda.Code.fromInline('foo'),
    });
    
    const trigger = new triggers.Trigger(this, 'MyTrigger', {
      handler: func,
      timeout: cdk.Duration.minutes(10),
      invocationType: triggers.InvocationType.EVENT,
    });

    # Apply stack-level tags where value come from the SSM parameter
    cdk.Tags.of(this).add("test", stringValue)

```

The value of these tags come from an SSM parameter `stringValue` which is a resolvable token. This error seems to occur when creating the IAM role that the parent construct `Trigger` creates:

        * AWS::IAM::Role
        * AWS::Lambda::Version
        * AWS::Lambda::Function
        * Custom::Trigger

Comparing this with a similar parent construct `AwsCustomResource`, the token strings are resolving during deployment and the tags are appropriately propagating to its child resources (AWS::IAM::Role, AWS::Lambda::Function).

```
    const stringValue = ssm.StringParameter.fromStringParameterAttributes(this, 'MyValue', {
      parameterName: 'StringParameter',
      // 'version' can be specified but is optional.
    }).stringValue;

    const getParameter = new cr.AwsCustomResource(this, 'GetParameter', {
      onUpdate: { // will also be called for a CREATE event
        service: 'SSM',
        action: 'getParameter',
        parameters: {
          Name: 'my-parameter',
          WithDecryption: true,
        },
        physicalResourceId: cr.PhysicalResourceId.of(Date.now().toString()), // Update physical id to always fetch the latest version
      },
      policy: cr.AwsCustomResourcePolicy.fromSdkCalls({
        resources: cr.AwsCustomResourcePolicy.ANY_RESOURCE,
      }),
    });

    cdk.Tags.of(this).add("test", stringValue)
```

Hence, I'm presuming there is an issue with the `Trigger` construct such that token strings passed to its child resources are not resolving during deployment. I also noticed within the template that the "Tags" are not propagating to the child resources (Role, Lambda Function) during synthesis when compared to `AwsCustomResources` where "Tags" are visible in the synthesized template.

I also did some additional testing and this time just passing tag values as a simple string instead of SSM parameters and the deployment went through for `Trigger` and propagated the tags to the child resources but this wasn't still visible in the template.

Are you able to give me an idea to why this happens, perhaps there is a gap with the `Trigger` construct? Is this a bug?

### Expected Behavior

Token strings of stack-level tags applied on child resources of Trigger not resolved during deployment

### Current Behavior

Token strings of stack-level tags applied on child resources of Trigger to be resolved during deployment

### Reproduction Steps

-

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.88.0

### Framework Version

_No response_

### Node.js Version

v18.7.0

### OS

Windows

### Language

TypeScript

### Language Version

TypeScript

### Other information

_No response_

### LINKED ISSUE #30055: (core/stack): stack.tags does not do anything
### Describe the bug

Possible regression. Creating this for a customer and they had this working. Waiting on them to give me the cdk version where this was working.

[stack.tags](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.Stack.html#tags) does not do anything. I spot checked 2.1.0, 2.50.0, 2.100.0 cdk version and I cannot see where it worked. 

### Expected Behavior

When specifing stack.tags it will go out and tag the constructs that support tagging.

### Current Behavior

stack.tags seemingly does nothing.

### Reproduction Steps

```
#!/usr/bin/env node
import { Bucket } from 'aws-cdk-lib/aws-s3';

const app = new cdk.App();
class TagsStack extends cdk.Stack {
    super(scope, id, props);

    new Bucket(this, "bucket");

    cdk.Tags.of(this).add('tester','value'); //This works adding the tester tag to the s3 bucket

  }
}
new TagsStack(app, 'TagsStack', {
  tags: {
    'test': 'value' // This does not add the any tags to the s3 bucket.
  }
});
```

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.140.0

### Framework Version

_No response_

### Node.js Version

v18.17.1

### OS

Ubuntu 22.04 - Linux

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

### LINKED ISSUE #33945: Stack: Tag propagation
### Describe the bug

When Tags are applied at a stack level - they propagate to all children, which is fine, but in some case some children must be excluded. E.g. some resources may not be taggable in Govcloud, attempt to tag them will throw an exception. Or adding tags will force resource to delete an re-create.

Using `excludeResourceTypes` while adding tags, or `includeResaourceTypes` while removing  has no effect. Neither does `addPropertyDeletionOverride('Tags')` at the resource level. By "no effect" I mean tags are removed from the resource in synthesized template, but will still appear in the deployed resource.

Is there any way to skip some resources from being tagged when tags are applied at the stack level?

### Regression Issue

_No response_

### Expected Behavior

Specified resource should not be tagged when tags are applied at stack level.

### Current Behavior

All stack's resources are tagged when tags are applied at stack level.

### Reproduction Steps

Try something like

```typescript
Tags.of(stack).add('key', 'value, {
      excludeResourceTypes: ['AWS::S3::Bucket']
 });
```
on a stack that has buckets. Tag will be applied anyway.

### Possible Solution

_No response_

_No response_

### CDK CLI Version

2.178.1 (build ae342cb)

### Framework Version

_No response_

### Node.js Version

v20.12.2

### OS

MacOS Sonoma 14.7.3

### Language

TypeScript

### Language Version

_No response_

### Other information

_No response_

rix0rrr:
Stacks are considered taggable, and so `Tags.of(this).add('key', 'value')` used to add tags to Stacks in scope. Usually this happens if `this` is an instance of `Stack`, which it commonly is in user code.

Since `Tags.of(...)` walks the construct tree, it will add tags to the stack *and* to all the resources in the stack. Then, come deploy time, CloudFormation will also try and apply all the stack tags to the resources again. This is both unnecessary, as well as leads to loss of control: `excludeResourceTypes` appears to not work, since it will lead to resources not being tagged in the template (good) but then the resources will still be tagged by CloudFormation because the stack itself is tagged (bad).

Also, if the tags applied this way contain intrinsics, they will contain nonsense because they are applied in a context where CloudFormation expressions don't work.

## In this change

There is way to prevent Stacks from being tagged, by including `aws:cdk:stack` in the list of `excludeResourceTypes` (this is a fake resource type that Stack tags respect).

Under a feature flag, `@aws-cdk/core:explicitStackTags`, this is now the default behavior. That resource type will be excluded by default, unless it is listed in the `includeResourceTypes` list. However, doing `includeResourceTypes` is still not desirable: stack tags should be applied directly on the `Stack` object if desired.

This requires a user to make a conscious decision between resource-level and stack-level tagging: either apply tags to the stack, which will apply it to all resources but remove the ability to do `excludeResourceTypes`; or apply tags to (groups of) resources inside the template.

Another benefit is that for tags applied at the stack level, this will resolve the following issue: https://github.com/aws/aws-cdk/issues/15947, as resources "becoming" taggable all of a sudden will not affect the template anymore.

Closes #28017. Closes #33945. Closes #30055.

----


mrgrain:
> The duality of Stack tags vs resource tags, and the fact that we effectively try and tag all resources twice, has been a thorn in my side for a while. I took this issue as a chance to address that long-standing design flaw.
> 
> I can be convinced that it's too messy to explain, and not worth fixing (we haven't had complaints about this explicitly broken behavior in over 5 years), and we should just error out on the unresolvable tag and that's it.

I think this makes a lot of sense and it's the right think to address the fundamental flaw. My concern is the implementation is using a feature flag and otherwise sticks to the very cool but also super vague `Tags.of(this).add('key', 'value')` naming. There will be many docs, articles and tutorials out there that reference this Aspect with the current behavior.

Should we instead take this as an opportunity to deprecate the current aspect (and maybe add a warning/error to catch the unsupported use case) and implement the new feature as `TagAllResources.of(this).add('key', 'value')` or similar.

comcalvi:
There's two other options:

1. (under feature flag) Make `Tags.of(this).add(key, value)` silently drop unresolved tokens, if it detects any; otherwise, it behaves as it does today.
2. Implement a post-deploy token resolver. This is probably too much effort for too little payoff, but we could put special markers in unresolved stack tags that we look up in AWS (presumably, a resource in the stack itself) and fill in after the deployment succeeds. We'd to store a token map in the cloud assembly.

> I didn't think that was that bad of a behavior. It's mostly what you want.

I agree; it seems the only time you don't want it is when that tag contains a Token, so dropping the unresolved tokens would be nice; is that too much hidden behavior from users though? 

I don't think the `Tags.of()` naming is too vague, CDK users have become familiar with it and it behaves as you would expect, bar this unresolved tag issue.

rix0rrr:
> (under feature flag) Make Tags.of(this).add(key, value) silently drop unresolved tokens, if it detects any; otherwise, it behaves as it does today.

I think silently not doing what you asked for is not a great choice.

> Implement a post-deploy token resolver. This is probably too much effort for too little payoff,

Exactly 😉

rix0rrr:
As I thought, this would be harder to close.

So I split off the most important bit to another PR: https://github.com/aws/aws-cdk/pull/31457

cjnoname:
Our application is breaking. We're trying to add tags to all resources within the CloudFormation template by reading the values from the SSM Parameter Store. However, during deployment with the latest version, we're encountering the following error:

Error: Stack tags may not contain deploy-time values (tag: awsApplication=${Token[TOKEN.16]}). Apply tags containing deploy-time values to resources only, avoid tagging stacks.

moelasmar:
@rix0rrr, then should we close this PR ?

packages/aws-cdk-lib/core/lib/tag-aspect.ts
```diff
@@ -155,12 +158,41 @@ export class Tags {
...
+   * leads to tags being applied twice (once in the template, and then once
+   * again automatically by CloudFormation as part of the stack deployment).
+   * That behavior leads to loss of control as `excludeResourceTypes` will
+   * prevent tags from appearing in the template, but they will still be
+   * applied to the Stack and hence CloudFormation will still apply them
+   * to the resource.
    */
   public add(key: string, value: string, props: TagProps = {}) {
+    // Implicitly add `aws:cdk:stack` to the `excludeResourceTypes` array in modern behavior
+    if (this.explicitStackTags && !props.includeResourceTypes?.includes('aws:cdk:stack')) {
```
otaviomacedo, line 189: If `props.includeResourceTypes?.includes('aws:cdk:stack')` is true, is there still a point in having the aspect traverse the tree? Can't we call `addStackTag()` to the stack and stop?

rix0rrr: Fair point, but if the user wanted exactly that behavior they would just set the stack tags directly.

I'm including this as an easy way of restoring the old behavior locally, for those people who might want it (for whatever ill-conceived reason). I don't seriously think it will see big adoption because it shouldn't, but I'm hesitant to take it away alltogether as well.

PR #31444, opened 2024-09-13
chore: clarify `cloud-assembly-schema` contribution flow

comcalvi:
### Reason for this change

The process for merging changes to the new `cloud-assembly-schema` repository is unclear. This clarifies the process.

### Checklist

----

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1227,11 +1227,11 @@ $ yarn test
 If you plan on making changes to the `cloud-assembly-schema` package, make sure you familiarize yourself with
 its own [contribution guide](https://github.com/cdklabs/cloud-assembly-schema/blob/main/CONTRIBUTING.md).
 
-The team will not actively monitor the `cloud-assembly-schema` repository for contributions. If you have a PR
-that needs to modify `cloud-assembly-schema`, open both PRs and ensure that you link the `cloud-assembly-schema`
-PR in the one opened against this repository. The `cloud-assembly-schema` changes must be merged (and released)
-before the PR in this repository can be merged. The PR in this repository must be updated to use the new release
-of the `cloud-assembly-schema` before it can be merged.
+The team will not actively monitor the [cloud-assembly-schema](https://github.com/cdklabs/cloud-assembly-schema)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1225,7 +1225,13 @@ $ yarn test
 ## Changing Cloud Assembly Schema
 
 If you plan on making changes to the `cloud-assembly-schema` package, make sure you familiarize yourself with
-its own [contribution guide](./packages/@aws-cdk/cloud-assembly-schema/CONTRIBUTING.md)
+its own [contribution guide](https://github.com/cdklabs/cloud-assembly-schema/blob/main/CONTRIBUTING.md).
+
+The team will not actively monitor the [cloud-assembly-schema](https://github.com/cdklabs/cloud-assembly-schema)
+repository for contributions. If you have a PR that needs to modify `cloud-assembly-schema`, open both PRs and
+ensure that you link the `cloud-assembly-schema` PR in the one opened against this repository.
```
kaizencc, line 1232: i don't understand what PR would be opened against aws-cdk's cloud-assembly-schema given that its just a shell for the external version.

CONTRIBUTING.md
```diff
@@ -1225,7 +1225,13 @@ $ yarn test
...
+The `cloud-assembly-schema` changes must be merged (and released) before the PR in this repository can be merged.
+The PR in this repository must be updated to use the new release of the `cloud-assembly-schema` before it can be merged.
```
kaizencc, line 1234: oh wait we don't automatically grab new versions of cloud-assembly-schema when they're released? we have to manually update the version?

a) if thats the case, is that safe from a security perspective
b) we should probably have an example/instruction for what the PR against aws-cdk repo looks like then

kaizencc:
also if we don't monitory cdklabs/cloud-assembly-schema we should at least remove the ability to add issues

CONTRIBUTING.md
comcalvi: We do grab new ones, it's just reliant on dependabot. I can clarify that we either have to wait for the version bump or manually bump it, but the point was that it's been approved, but it can't merge until dependabot decides to update it.

